Issue #1300, opened 2021-02-03
Code action to remove redundant constraints can't deal with type operators properly


berberman:
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
haskell-language-server version: 0.9.0.0 (GHC: 8.10.3) (PATH: /home/berberman/.cabal/store/ghc-8.10.3/haskell-language-server-0.9.0.0-81f61f357a41d74a3d6d524b9f8f4949ec5bbcb5dd08161f78f8859c0cd951c1/bin/haskell-language-server)
Tool versions found on the $PATH
cabal:          3.2.0.0
stack:          Not found
ghc:            8.10.3
```
Which lsp-client do you use:
vscode

### Steps to reproduce

Create a haskell file like this, then try applying code actions to remove redundant constraints:

```haskell

{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TypeOperators #-}
{-# OPTIONS_GHC -Wredundant-constraints #-}

class a ** b

f :: (**) a b => a -> b -> a
f = const

g :: ((**) a b) => a -> b -> a
g = const

h :: ((**) a b, Eq a) => a -> b -> a
h = const

```

### Expected behaviour

It results in:

```haskell

{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TypeOperators #-}
{-# OPTIONS_GHC -Wredundant-constraints #-}

class a ** b

f :: a -> b -> a
f = const

g :: a -> b -> a
g = const

h :: a -> b -> a
h = const
```

### Actual behaviour

* `f`'s type signature became `f :: **) a => a -> b -> a` 
* no code action populated for `g`
* `h`'s type signature became `h :: (**) a b => a -> b -> a`

PR #1578, opened 2021-03-15
Fix remove constraint

kderme:
This pr refactors the code actions which remove redundant constraints from function signatures. The new approach traverses the ast to find the related signature and uses the `ExactPrint` functionality to create a new signature. The old approach did some text manipulation of the module, which was not very reliable. This caused a number of issues which are now fixed:

- code actions for signatures which are not top level were wrong:
```
f :: Int -> ()
f = g
  where
    g :: Eq a => a -> ()
  -- abcdefghijklmn
    g _ = ()
```
resulted in
```
f :: Int -> ()
f = g
  where
    g :: Eq a => a -> ()
  -- ijklmn
    g _ = ()
```
- signatures with an explicit `forall` were not fixed at all 
```
g :: forall a. Show a => a -> ()
g _ = ()
```
created no visible change
- signatures with spaces created no code action
```
g ::  Show a => a -> ()
g _ = ()
```

Tests are failing (hence the draft) because `ExactPrint` creates an unnecessary space. I'm still trying to understand why and exactly how `TransformT` works

I think the pr also helps fixing https://github.com/haskell/haskell-language-server/issues/1300, but doesn't really fix it.


kderme:
Test failures. Not sure where this space comes from :thinking: 
```
    remove redundant function constraints
      Remove redundant constraint `Eq a` from the context of the type signature for `foo`:                FAIL (1.64s)
        test/exe/Main.hs:2681:
        expected: "{-# OPTIONS_GHC -Wredundant-constraints #-}\nmodule Testing where\n\nfoo :: a -> a\nfoo = id\n"
         but got: "{-# OPTIONS_GHC -Wredundant-constraints #-}\nmodule Testing where\n\nfoo ::  a -> a\nfoo = id\n"
      Remove redundant constraints `(Eq a, Monoid a)` from the context of the type signature for `foo`:   FAIL (1.82s)
        test/exe/Main.hs:2681:
        expected: "{-# OPTIONS_GHC -Wredundant-constraints #-}\nmodule Testing where\n\nfoo :: a -> a\nfoo = id\n"
         but got: "{-# OPTIONS_GHC -Wredundant-constraints #-}\nmodule Testing where\n\nfoo ::  a -> a\nfoo = id\n"
```

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
```diff
@@ -118,6 +119,19 @@ fixParens openDP closeDP ctxt@(L _ elems) = do
...
+  LHsType GhcPs ->
+  Rewrite
+removeConstraint pred = go
+  where
+    go (L l it@HsQualTy{hst_ctxt = L l' ctxt}) = Rewrite l $ \_ -> do
+      return $ L l $ it{hst_ctxt = L l' $ filter (not . pred) ctxt}
+    go (L _ (HsParTy _ ty)) = go ty
+    go (L _ HsForAllTy{hst_body}) = go hst_body
+    go (L l other) = Rewrite l $ \_ -> return $ L l other
+
```
berberman, line 134: Is it enough to only remove the node from AST without adjusting annotations? I didn't look into it, but I guess the unexpected space might come from the `DeltaPos` of `HsFunTy` , which determines preceding spaces of the first `a`. Printing `Anns` to see what's going on may help.

Also, `pred` is a function in Prelude, so I would change the predication another name to avoid confusion.

isovector: This is likely the issue; getting annotations right is the single hardest part of HLS in my experience.

kderme: This helped thanks!

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -162,6 +162,61 @@ findSigOfDecl pred decls =
...
+        case unLoc <$> findDeclContainingLocE (_start range) lsigs of
+          Right sig' -> Right sig'
+          Left _ -> do
+            lHsBindLR <- findDeclContainingLocE (_start range) (bagToList binds)
+            findSigOfBind range (unLoc lHsBindLR)
+    go _ = Left "findSigOfBinds"
+
+findDeclContainingLocE :: Position -> [Located a] -> Either String (Located a)
+findDeclContainingLocE loc ls =
+  maybe (Left "findDeclContainingLoc") Right $ findDeclContainingLoc loc ls
```
berberman, line 218: Why not use `Maybe`? Those strings in `Left` are unused, right?

berberman (dismissed):
LGTM, and tests are great! Thank you!

ghcide/src/Development/IDE/Plugin/CodeAction.hs
kderme: I've used them for debugging. There are too many places where a signature can appear and my `findSig`s functions are not complete. These messages have helped me locate the dead ends and extend to different kind of ast nodes. If we want to catch more cases in the future these messages can be very helpful I think.

berberman: Well, I don't think it would be great to include these debug stuffs. Perhaps you could add some comments instead.

kderme: I guess you're right. I'll take a look at one last thing: signatures in multiple lines and then I'll turn these `Either`\`s into `Maybe`\`s

jneira: I've re-requested the review to wait for this change

jneira: @kderme have you got the opprtunity of removing Eithers?

kderme: Sorry this tool so long to address :flushed:

jneira: No worries, thanks for taking care of requests
@berberman would it ready to go?

berberman (approved):

PR #1580, opened 2021-03-16
Update homepage and other urls for ghcide

felixonmars:
It is currently referring to a deprecated location.


berberman (approved):

jneira:
@felixonmars good catch! thanks for the patch

PR #1587, opened 2021-03-17
Ensure eval plugin Print class doesn't rely on Prelude being in scope

akrmn:
Hi,

This PR is pretty much what it says on the tin :)

The reason I need this is because the codebase at work uses `{-# NoImplicitPrelude #-}`, together with a custom prelude that includes `show :: Show a => a -> Text`, which doesn't play well with the Eval plugin. Currently, we get errors like:

```
Request workspace/executeCommand failed.
  Message: Couldn't match type ‘Text’ with ‘[Char]’
Expected type: IO String
  Actual type: IO Text

  Code: -32603 
```

This is because the `Print` class, as defined on `Ide.Plugin.Eval.Code`, uses whatever `show` is in scope. Having a qualified Prelude in scope when `Print` is defined should be enough to avoid that, and I can't think of any drawbacks.

Thanks for your consideration and for all the great work on HLS so far!


pepeiborra (approved):
Thank you señor Ackermann! And best wishes

PR #1585, opened 2021-03-17
Add ability for plugins to handle file change notifications

alanz:
This allows alternate parsers to be used, such as tree-sitter.

Issue #594, opened 2020-11-12
Changes in hlint config file should be applied on the fly


jneira:
(from twitter: https://twitter.com/korayal/status/1326963429257711623)
* I describe that behaviour prematurely in docs but it does not work
* We should change `getHlintSettingsRule` from `defineNoFile` to a definition that takes in account possible configuration files 

https://github.com/haskell/haskell-language-server/blob/61c69fd387281c2222f75f13542eaa2e91cbc1b1/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs#L218-L223

Profpatsch:
Seconded, I run into this every time I change the hlint file.

PR #524, opened 2020-10-21
[WIP] Metaprogramming interface to tactics

isovector:
This PR introduces a tactic metaprogramming language for creating new code actions and extending `Attempt to fill hole`. It will automatically find scripts under `$XDG_CONFIG_HOME/hls-tactics-plugin` and load them into the tactics plugin.

PR #1588, opened 2021-03-17
Add ability for plugins to handle file change notifications

pepeiborra:
Rebased #1585 and extracted ghcide notification handlers

### DIFF
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -194,39 +194,6 @@ instance PluginMethod TextDocumentRangeFormatting where
 
 -- ---------------------------------------------------------------------
 
--- | Notifications that can be handled by plugins.
--- 'ExtraParams' captures any extra data the IDE passes to the handlers for this method
-class HasTracing (MessageParams m) => PluginNotification m where
-
-  -- | Parse the configuration to check if this plugin is enabled
-  pluginEnabledNotification :: SMethod m -> PluginId -> Config -> Bool
-
-instance PluginNotification TextDocumentDidOpen where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification TextDocumentDidChange where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification TextDocumentDidSave where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification TextDocumentDidClose where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification WorkspaceDidChangeWatchedFiles where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification WorkspaceDidChangeWorkspaceFolders where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification WorkspaceDidChangeConfiguration where
-    pluginEnabledNotification _ _ _ = True
-
-instance PluginNotification Initialized where
-    pluginEnabledNotification _ _ _ = True
-
--- ---------------------------------------------------------------------
-
 -- | Methods which have a PluginMethod instance
 data IdeMethod (m :: Method FromClient Request) = PluginMethod m => IdeMethod (SMethod m)
 instance GEq IdeMethod where
@@ -235,7 +202,7 @@ instance GCompare IdeMethod where
   gcompare (IdeMethod a) (IdeMethod b) = gcompare a b
 
 -- | Methods which have a PluginMethod instance
-data IdeNotification (m :: Method FromClient Notification) = PluginNotification m => IdeNotification (SMethod m)
+data IdeNotification (m :: Method FromClient Notification) = HasTracing (MessageParams m) => IdeNotification (SMethod m)
 instance GEq IdeNotification where
   geq (IdeNotification a) (IdeNotification b) = geq a b
 instance GCompare IdeNotification where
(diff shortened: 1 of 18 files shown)


hls-plugin-api/src/Ide/Types.hs
```diff
@@ -180,18 +181,74 @@ instance PluginMethod TextDocumentRangeFormatting where
...
+--   = PluginNotificationHandler (PluginNotificationMethodHandler a m)`
+
+{-
+From Zubin
+alanz_: I would say `newtype PluginNotificationHandler a (m :: Method FromClient Notification) = PluginNotificationHandler (PluginNotificationMethodHandler a m)`
+16:28 and `newtype PluginNotificationHandlers a = PluginNotificationHandlers (DMap SMethod (PluginNotificationHandler a))
+
+-}
+
+newtype PluginHandlers a             = PluginHandlers             (DMap IdeMethod       (PluginHandler a))
```
alanz, line 249: This section from here down is a repeat of what is in  ghcide/src/Development/IDE/Plugin/HLS.hs

alanz (commented):
There seems to be some duplication between the HLS and ghcide side.

Otherwise seems to be what I had before.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -180,18 +181,74 @@ instance PluginMethod TextDocumentRangeFormatting where
...
+instance GCompare IdeNotification where
+  gcompare (IdeNotification a) (IdeNotification b) = gcompare a b
+
 -- | Combine handlers for the
 newtype PluginHandler a (m :: Method FromClient Request)
   = PluginHandler (PluginId -> a -> MessageParams m -> LspM Config (NonEmpty (Either ResponseError (ResponseResult m))))
 
-newtype PluginHandlers a = PluginHandlers (DMap IdeMethod (PluginHandler a))
+newtype PluginNotificationHandler a (m :: Method FromClient Notification)
+  = PluginNotificationHandler (PluginId -> a -> MessageParams m -> LspM Config (NonEmpty ()))
```
wz1000, line 238:
```suggestion
  = PluginNotificationHandler (PluginId -> a -> MessageParams m -> LspM Config ())
```

hls-plugin-api/src/Ide/Types.hs
pepeiborra: Which one is redundant?

pepeiborra: Just to clarify, I haven't added anything. If there is any duplication, it was already present in your branch.

alanz: fair enough :)

isovector:
Is this capable of watching arbitrary files on the filesystem? I wanted such a thing in #524, in order to watch `$XDG_CONFIG/wingman`.

pepeiborra:
> Is this capable of watching arbitrary directories on the filesystem? I wanted such a thing in #524, in order to watch `$XDG_CONFIG/wingman`. 

No, only workspace folders. Outside the workspace, you would need to implement your own watching

jneira:
> 
> 
> > Is this capable of watching arbitrary directories on the filesystem? I wanted such a thing in #524, in order to watch `$XDG_CONFIG/wingman`.
> 
> No, only workspace folders. Outside the workspace, you would need to implement your own watching

I would need it for hlint config files for #594, maybe it worths to implement it in some common place (hls-plugin-api?)

pepeiborra:
> > > Is this capable of watching arbitrary directories on the filesystem? I wanted such a thing in #524, in order to watch `$XDG_CONFIG/wingman`.
> > 
> > 
> > No, only workspace folders. Outside the workspace, you would need to implement your own watching
> 
> I would need it for hlint config files for #594, maybe it worths to implement it in some common place (hls-plugin-api?)

Rather than turning HLS into an operating system, I would push back on these feature requests and ask people to reload

alanz (approved):

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -191,19 +192,64 @@ instance PluginMethod TextDocumentRangeFormatting where
   pluginEnabled _ pid conf = (PluginId $ formattingProvider conf) == pid
   combineResponses _ _ _ _ (x :| _) = x
 
+-- ---------------------------------------------------------------------
+
+-- | Notifications that can be handled by plugins.
+-- 'ExtraParams' captures any extra data the IDE passes to the handlers for this method
+class HasTracing (MessageParams m) => PluginNotification m where
```
wz1000, line 199: I don't think we really need this class. Isn't `HasTracing (MessageParams m)` sufficient?

wz1000: It doesn't really make sense to disable notifications for a particular plugin via the config I think. Its more of an internal implementation detail that shouldn't be visible to the user.

We could disable notifications from a plugin if the plugin itself is disabled.

pepeiborra: Isn't this class also used to restrict what `Method`s are allowed to be used in `IdeNotification`?

wz1000: Yes, but is there any reason to have such a restriction?

pepeiborra: I'll remove it

wz1000 (approved):

PR #1590, opened 2021-03-18
Avoid duplicating known targets and import paths

pepeiborra:
We currently duplicate all the known targets on every cradle reload, which leaks space and leads to slower performance in `GetLocatedImports`. 
The trivial fix is to use a `Set` instead of a list, since the order doesn't matter here.


jneira (approved):
Nice catch, thanks!

PR #1592, opened 2021-03-18
Use CiInterface/SkInterface for typeclass symbols

### LINKED ISSUE #1581: Suggestion: Use CiInterface as completion item kind for typeclasses
Typeclasses currently use `CiClass` as completion item kinds. My impression, however, is that most LSP clients (e.g. VSCode) interpret the completion item kind `Class` to be referring to OOP-style classes, as VSCode's icon suggests:

image

Since Haskell's typeclasses are semantically much closer to OOP interfaces, my suggestion would be to use `CiInterface` as completion item kind instead.

fwcd:
### Fixes #1581

This branch replaces the `CiClass`/`SkClass` item kinds for completions and document symbols with `CiInterface`/`SkInterface` (see the linked issue for the motivation behind this change).

Additionally, the symbol icons for type families which were previously `SkClass` have been replaced with `SkFunction`.

### DIFF
--- a/ghcide/src/Development/IDE/LSP/Outline.hs
+++ b/ghcide/src/Development/IDE/LSP/Outline.hs
@@ -69,7 +69,7 @@ documentSymbolForDecl (L (RealSrcSpan l) (TyClD _ FamDecl { tcdFam = FamilyDecl 
                        t  -> " " <> t
                      )
     , _detail = Just $ pprText fdInfo
-    , _kind   = SkClass
+    , _kind   = SkFunction
     }
 documentSymbolForDecl (L (RealSrcSpan l) (TyClD _ ClassDecl { tcdLName = L _ name, tcdSigs, tcdTyVars }))
   = Just (defDocumentSymbol l :: DocumentSymbol)
--- a/ghcide/src/Development/IDE/LSP/Outline.hs
+++ b/ghcide/src/Development/IDE/LSP/Outline.hs
@@ -69,7 +69,7 @@ documentSymbolForDecl (L (RealSrcSpan l) (TyClD _ FamDecl { tcdFam = FamilyDecl 
                        t  -> " " <> t
                      )
     , _detail = Just $ pprText fdInfo
-    , _kind   = SkClass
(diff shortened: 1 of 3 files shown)


Ailrun:
Just a question: does this affect anything other than icons in auto-completion popup?

Ailrun:
Oh, right. So it's all about icons, right?

fwcd:
Yup.

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -69,7 +69,7 @@ documentSymbolForDecl (L (RealSrcSpan l) (TyClD _ FamDecl { tcdFam = FamilyDecl
                        t  -> " " <> t
                      )
     , _detail = Just $ pprText fdInfo
-    , _kind   = SkClass
+    , _kind   = SkInterface
```
Ailrun, line 72: This one is actually for type families, right? I think the PR description is a bit misleading...

jneira (approved):
It seems to me that the interface icon is more appropriate, many thanks!

Ailrun (approved):
Oops, I meant to click this one. Thank you for the PR!

ghcide/src/Development/IDE/LSP/Outline.hs
fwcd: Whoops, you're right, thanks!

Ailrun: Is there any reason to leave them as `SkClass`?

fwcd: Wasn't entirely sure about type families, but in any case we should do it consistently. Thinking about it, `SkInterface` seems more appropriate than `SkClass` for type families too, since type families are somewhat analogous to type classes, but over data instead of functions.

Should I change it again?

michaelpj: :hot_pepper:  suggestion: `SkFunction`. Isn't a type family *really* a type-level function? :D

PR #1595, opened 2021-03-19
Relax ghcides upper bound on base16-bytestring

maralorn:
base16-bytestring 1.0.1.0 has recently entered stackage nightly.
I have build haskell-language-server with that version and it seems to work fine.

Note: As of now implicit-hie-cradle and retrie also need this bump.


berberman (approved):
LGTM. Thanks!

PR #1596, opened 2021-03-19
Make the eval plugin use the same default language extensions as ghci.

### LINKED ISSUE #1591: Eval plugin reports "Ambiguous type variable" where ghci does not.
Which lsp-client do you use:
VS Code

### Steps to reproduce
Type `-- >>> head []`in the editor and click on `Evaluate...`

### Expected behaviour
Output: 
`*** Exception: Prelude.head: empty list`
as in ghci.

### Actual behaviour
Output: 
````
-- Ambiguous type variable ‘f0’ 
-- prevents the constraint ‘(Show f0)’ from being solved.
-- Probable fix: use a type annotation to specify what ‘f0’ should be.
-- These potential instances exist:
--   instance Show a => Show (ZipList a)
--     -- Defined in ‘Control.Applicative’
--   instance Show NestedAtomically
--     -- Defined in ‘Control.Exception.Base’
--   instance Show NoMethodError -- Defined in ‘Control.Exception.Base’
--   ...plus 215 others
--   (use -fprint-potential-instances to see them all)
````

### Workaround
Specifying a concrete type, such as  `-- >>> head ([]::[Int])` does not result in this error, but is not pretty.

### PR DESCRIPTION
**Pull request by fmehta**
This pull request aims to fix #1591

### CODE REVIEW

**Comment by jneira** [comment]
Many thanks!, a test case would be great to make visibly eval plugin is able to handle it and to avoid future changes breaks it

**Comment by jneira** [comment]
@fmehta feel free to ask for help if you need some guidance to write the test
The test suite for eval plugin is here: https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/test/Eval.hs

**Comment by fmehta** [comment]
@jneira : Thanks a lot for following up on my progress and pointing me in the right direction. And yes, of course I need to include a test for this change (what was I thinking 🙄).

The reason for my delay: I was just wondering whether it even makes sense to enable this language extension at all. Maybe it makes sense to use exactly the same language extensions as ghc, since the generated comments will be part of the source code. Since there is no specification as to what the correct behaviour of the evals should be, I was wondering if you, or anyone else, could comment on what behaviour would be desirable.

I also notice that I cannot enable this language extension within the eval plugin for some reason (I guess only those that are listed in https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/README.md  `>>> :set -XScopedTypeVariables -XStandaloneDeriving -XDataKinds -XTypeOperators -XExplicitNamespaces` ) can be set. I also see that there are other language extensions that are enabled for ghci (`:showi language` within ghci lists`NoDatatypeContexts, ExtendedDefaultRules, NoMonomorphismRestriction, NondecreasingIndentation`), whereas the following are enabled by default in the eval plugin:

````
evalExtensions =
    [ OverlappingInstances
    , UndecidableInstances
    , FlexibleInstances
    , IncoherentInstances
    , TupleSections
    ]
````

I wonder whether they should not be the same?

Do you (or anyone else) have any thoughts on all of this, or should I just move on and add the tests for this commit without getting too deep here?

Thanks, FM

**Comment by pepeiborra** [comment]
I don't see anything wrong with enabilng `ExtendedDefaultRules` by default, and all the other ones ghci enables by default.

The set of extensions that are enabled currently for the eval plugin seem arbitrary and should be removed. /cc @tittoassini

**Comment by Ailrun** [comment]
I believe some of the extensions are used printing code of the Eval plugin, so I guess it's not so trivial to remove....

**Comment by pepeiborra** [comment]
> I believe some of the extensions are used in the printing code (which is executed in Eval plugin session) of Eval plugin, so I guess it's not so trivial to remove....

@berberman removed that custom printing code recently

**Comment by Ailrun** [comment]
Oh, I missed that, then probably we are good to go.

**Comment by jneira** [comment]
@fmehta hi, do you still plan to continue working on this? thanks!

**Comment by fmehta** [comment]
Hi @jneira 
Thanks for your help. I have committed the tests. 
Sorry for the delay. I was trying to run them through the top level contribution instructions at <https://github.com/haskell/haskell-language-server/blob/master/CONTRIBUTING.md> which did not work for me. I only later realised that I could just run `cabal test` within the eval plugin.

Could you please review and merge my changes in case they are OK. I will then create and work on a second issue for the rest of the default ghci extensions as mentioned in  <https://github.com/haskell/haskell-language-server/pull/1596#issuecomment-804770960>.

Thanks,
FM

**Comment by berberman** [comment]
Could you change the purpose of this PR, resolving #1954 together?

**Comment by fmehta** [comment]
Hi @berberman 

> Could you change the purpose of this PR, resolving #1954 together?

#1954 needs some more work. In particular:
- I am trying to see if it is possible to get the defaults from the ghci code programatically, instead of manually.
- I am currently unsure if and how I can specify `NoDatatypeContexts` and `NoMonomorphismRestriction`, given that I can only find `DatatypeContexts` and `MonomorphismRestriction` within `GHC.LanguageExtensions.Type.Extension`.

(In case you have any pointers, please let me know.)

I therefore thought that I should start a new PR for the rest. But in case you think it is better to merge both together, I can do that as well. Just let me know.

Thanks 
FM

**Comment by berberman** [comment]
There is no negative type of extension, so If you want to disable one, you should *unset* it. As far as I can tell, `NoMonomorphismRestriction` and `ExtendedDefaultRules` are configured manually in GHCi: https://github.com/ghc/ghc/blob/5abf59976c7335df760e5d8609d9488489478173/ghc/GHCi/UI.hs#L473-L483.

And `evalExtensions` finally comes to here:

https://github.com/haskell/haskell-language-server/blob/0b3bb10cd22cb265b7cbab84514178bcf4ae4f92/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs#L304-L306

**Comment by fmehta** [comment]
Dear @berberman 

Thanks for your input. I have added the fix for #1954 to this pull request as well.

A third thing that I noticed and did not get my head around was that it seems that `ExplicitForAll` is enabled by default in the eval plugin, but not in ghci. 

```
-- >>> plus = (+)
-- >>> :t plus
-- plus :: forall a. Num a => a -> a -> a
```

I cannot find out where this is added. Do you have any idea?

FM

**Comment by fmehta** [comment]
Hi @berberman 
Hi @jneira 

It should work now. Maybe one of you need to kickstart the CI.

Best,
FM

**Review by jneira — APPROVED** [review]
Looks great (love the comments in tests), many thanks

**Comment by fmehta** [comment]
Thanks @jneira!

File: plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs [hunk 1]
```diff
@@ -81,15 +80,17 @@ asStmts (Example e _ _) = NE.toList e
...
+-- evalExtensions :: [Extension]
+-- evalExtensions =
+--     [ NondecreasingIndentation
+--     -- , OverlappingInstances
+--     -- , UndecidableInstances
+--     -- , FlexibleInstances
+--     -- , IncoherentInstances
+--     -- , TupleSections
+--     , ExtendedDefaultRules
+--     ]
```
berberman, line 93:
```suggestion
```

fmehta: Hi @berberman 
Sorry, I forgot to delete the comment. Thanks for spotting this!
FM

PR #1598, opened 2021-03-19
Bump haddock-library to 1.10.0

berberman:


jneira (approved):
lgtm, do you want to get some bug fix fromthe new version?

berberman:
We have packaged ghcide to Arch Linux: https://archlinux.org/packages/community/x86_64/haskell-ghcide/, but with a temporary package https://archlinux.org/packages/community/x86_64/haskell-haddock-library1.9/, since haddock-library on Arch Linux is 1.10.0. We could get rid of the 0.9 one if no packages depend on it.

berberman:
stack is annoying :(

PR #1603, opened 2021-03-21
Eval plugin: evaluate expressions as statements

### LINKED ISSUE #1577: Inline eval no longer capable of using "it" to reference the output of the previous inline
Obviously the inline code evaluation is amazing, but in previous versions I used to do this:

```
-- >>> example
-- >>> used it
-- >>> fresh it
-- \a. \x. (\y. a) x b
-- ["a","b","x","y"]
-- "c"
```

So I would use ```it``` to refer to the previous output, the same as in a ghci session,

Now the following happens:

```
-- >>> example
-- >>> used it
-- >>> fresh it
-- \a. \x. (\y. a) x b
-- Variable not in scope: it :: Term
-- Variable not in scope: it :: [Var]
```

it can still be sorted in many ways, a very ugly one with variables:

```
-- >>> example
-- >>> it1 = example
-- >>> used it1
-- >>> it2 = used it1
-- >>> fresh it2
-- \a. \x. (\y. a) x b
-- ["a","b","x","y"]
-- "c"
```

But it is really a pity because removes the versatility of ```it``` that just brings the previous output whatever it was.

Any chance to bring it back?

Thanks a lot

berberman:
Closes #1577

#438 introduced an alternative way to evaluate expressions: we wrapped expressions with `asPrint`, and then used `compileExpr` to get the coerced result `IO String`. However, it broke functionality of `it` variable -- referencing the last evaluated result.

This PR restores the old eval plugin's behaviour: there is no special treatment for expressions anymore; instead, expressions are just statements, and we redirect the output of `execStmt` to a temporary file to get the same result as `compileExpr`.

### DIFF
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
@@ -24,7 +24,7 @@ import           Ide.Plugin.Eval.Types          (Language (Plain), Loc,
 import           InteractiveEval                (getContext, parseImportDecl,
                                                  runDecls, setContext)
 import           Language.LSP.Types.Lens        (line, start)
-import           System.IO.Extra                (newTempFile)
+import           System.IO.Extra                (newTempFile, readFile')
 
 -- | Return the ranges of the expression and result parts of the given test
 testRanges :: Test -> (Range, Range)
@@ -106,7 +106,7 @@ myExecStmt stmt opts = do
     modifySession $ \hsc -> hsc {hsc_IC = setInteractivePrintName (hsc_IC hsc) evalPrint}
     result <- execStmt stmt opts >>= \case
               ExecComplete (Left err) _ -> pure $ Left $ show err
-              ExecComplete (Right _) _ -> liftIO $ Right . (\x -> if null x then Nothing else Just x ) <$> readFile temp
+              ExecComplete (Right _) _ -> liftIO $ Right . (\x -> if null x then Nothing else Just x ) <$> readFile' temp
               ExecBreak{} -> pure $ Right $ Just "breakpoints are not supported"
     liftIO purge
     pure result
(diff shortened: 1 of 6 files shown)


plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
```diff
@@ -99,12 +97,19 @@ evalSetup = do
...
+
+-- | A wrapper of 'InteractiveEval.execStmt', capturing the execution result
+myExecStmt :: String -> ExecOptions -> Ghc (Either String (Maybe String))
+myExecStmt stmt opts = do
+    (temp, purge) <- liftIO newTempFile
+    evalPrint <- head <$> runDecls ("evalPrint x = P.writeFile \"" <> temp <> "\" (P.show x)")
+    modifySession $ \hsc -> hsc {hsc_IC = setInteractivePrintName (hsc_IC hsc) evalPrint}
+    result <- execStmt stmt opts >>= \case
+              ExecComplete (Left err) _ -> pure $ Left $ show err
+              ExecComplete (Right _) _ -> liftIO $ Right . (\x -> if null x then Nothing else Just x ) <$> readFile temp
```
pepeiborra, line 109: use the strict [`readFile'`](https://hackage.haskell.org/package/extra-1.7.9/docs/Extra.html#v:readFile-39-)

pepeiborra (approved):
Very nice!

berberman:
All test cases on Windows are failing:

```
lexical error in string/character literal at character 'U'
```

Perhaps an encoding-related issue?

jneira:
@berberman mmm, i think you will have to save the file using utf-8, not sure how being in a evaluated expression, see

https://github.com/haskell/haskell-language-server/blob/b9c6e6cf4de5a0778c5ddb5d5e62fc8d1e0d78ef/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs#L399-L403

(maybe you will not need the newline thing)

There is a function [in the `extra` package `writeUtf8`](http://hackage.haskell.org/package/extra-1.2/docs/System-IO-Extra.html) but i guess maybe you cant use it in that context.

berberman:
Never mind, I found it might be caused by the backslash was not escaped properly

PR #1422, opened 2021-02-21
Extract an hls-brittany-plugin package

pepeiborra:
This also shows how to set up an isolated test suite /cc @konn @isovector @tittoassini  see 4c2011a15f7ca59e2071f1d2d7fd2db4cbfd4893

PR #1604, opened 2021-03-22
Extract stylish-haskell plugin into a standalone package

berberman:
Following #1422


pepeiborra (approved):
Nice, thank you!

prikhi:
This commit broke formatting w/ stylish-haskell for me(via coc.nvim). 

On this commit, launching HLS w/ `-d` doesn't show `PluginId "stylish-haskell"` in `LspArguments` plugin list. On the commit before this one, it does.

berberman:
Hi, thanks for the bug report! I just realized that `cpp-options: -D` was removed accidently from common stanza, so the plugin was not actually wired in HLS. I'm afraid brittany plugin had the same problem. /cc @pepeiborra

prikhi:
hehe, I just submitted a PR w/ the same fix :stuck_out_tongue: 
https://github.com/haskell/haskell-language-server/pull/1620

PR #1606, opened 2021-03-22
Extract hls-test-utils

berberman:
With the migration of plugins' tests from HLS's `func-test` to their own packages, the common stanza `hls-test-utils` will no loger be available in the pattern that tests of plugins were imported as sources into `func-test`. But there are still many common functions need to be shared. Therefore, this PR extracts `Test.Hls.Util` into a new package `hls-test-utils`, with some re-exported modules. Plugin test executables now can depend on this new package, rather than `data-default`, `lsp-test`, `tasty-*`, etc.

Moreover, this PR finishes the migration of eval plugin tests.


jneira (approved):
look really great, this was a necessary step, many thanks
we will have another package to maintain in hackage but I hope it will not change frequently (maybe if re-exported packages change?)

jneira:
eval tests are failing in windows with `Exception: Timed out waiting to receive a message from the server.`
rerunning just if case (but i dont have many hopes)

jneira:
I've tried the hls-eval-plugin locally with no luck: `40 out of 50 tests failed (2379.28s)`
Lots of timeouts, but 
```
  Multi line comments, with the last test line ends without newline:                                  FAIL (7.11s)
    test\Main.hs:186:
    expected: "module TEndingMulti where\n\n-- Now trailing doctest is allowed:\n\n{- >>> 42\n>>> 54\n42\n54\n-}\n"
     but got: "module TEndingMulti where\n\n-- Now trailing doctest is allowed:\n\n{- >>> 42\n>>> 54\n<stderr>: hPutChar: invalid argument (invalid character)\n-}\n"
  Evaluate expressions in Plain comments in both single line and multi line format:                   FAIL
    Last message received:
    {
        "id": 2,
        "jsonrpc": "2.0",
        "error": {
            "message": "C:\\TEMP\\TPl7BFB.hs: hPutChar: invalid argument (invalid character)",
            "code": -32603
        }
    }
```

ghcide/src/Development/IDE/Main.hs
```diff
@@ -116,6 +117,7 @@ stderrLogger = do
 
 defaultMain :: Arguments -> IO ()
 defaultMain Arguments{..} = do
+    setLocaleEncoding utf8
```
berberman, line 120: @jneira In `defaultMain` we've already set the encoding of `stdout` and `stderr` to UTF-8; plus `setLocaleEncoding utf8`, I think we could remove `withUtf8` from here:
https://github.com/haskell/haskell-language-server/blob/5724343ee2f861cba26fe430ff6413ff5f2f9909/exe/Main.hs#L13

and get rid of this dependency.

jneira: The package has some utility functions but it seems `withUft8` does esentially that (also for stdin?)

https://github.com/serokell/haskell-with-utf8/blob/b60991303f9cf3f2891479c581549c2d54f45a0e/lib/Main/Utf8.hs#L78-L82

If we are not going to use its utility functions we can take ride of, yeah

berberman: No worry, `stdin` is used in lsp mode, where [`runServerWithHandles`](https://hackage.haskell.org/package/lsp-1.1.0.0/docs/src/Language.LSP.Server.Control.html#runServerWithHandles) will set it to utf8

PR #1608, opened 2021-03-23
Allow for customizable Haskell views of Property types

isovector:
This PR lets consumers of the new `Property` machinery customize the type they end up with for the `TEnum`, `TArray` and `TObject` properties.


berberman (approved):
Many thanks, I love it!

Issue #681, opened 2020-12-17
A language pragma snippet would be very useful


expipiplus1:
A snippet which matches on `language` and expands to `{-# LANGUAGE <> #-}` would be very handy

jneira:
Maybe it could be added to the completion plugin that @gdevanla is moving to hls from ghcide (it does not have to be in the initital pr but afterwards)

July541:
To make it more clearly, for the code
```haskell
{-# LANGUAGE OverloadedStrings #-}

import Data.Text as T

foo :: T.Text
foo = "abc"
```

does this look like the image?
![image](https://user-images.githubusercontent.com/16057632/116085738-b16cb800-a6d1-11eb-8954-1c98e42ac32f.png)

Issue #729, opened 2020-10-16
My pragmas keep shouting at me


emilypi:
Hey all, you have a great project here, and I'm liking the progress. My only gripe so far is that whenever `ghcide` prompts me to add an extension, it inserts a very loud pragma LANGUAGE or INLINE or MINIMAL etc. This is a very low priority feature request, but you could add the ability to write lower-case extensions so that I can have an indoor conversation with my compiler? 

Thanks

michaelpj:
Generally I think we're trying to avoid configuration, and nobody else has asked for this, so closing.

georgefst:
> Generally I think we're trying to avoid configuration

Why exactly? I don't think I've seen any discussion around that. I am aware that adding options is currently quite onerous, from how long it took to resolve https://github.com/haskell/haskell-language-server/issues/2827. But there are various suggestions in linked threads about how that situation could be improved.

In general, I'm not sure about the wisdom of closing unresolved issues like this, as opposed to just marking them low-priority. I do still care slightly about this, and presumably so does @emilypi. (This is essentially the "stale bot" debate, where I can see the logic in tagging "stale" issues, but not in auto-closing them as some do.)

michaelpj:
> Why exactly? I don't think I've seen any discussion around that.

Perhaps it's folklore. But generally: configuration means a bigger state space, means more to test, and gives users more choices (which is not necessarily good for them). If it's possible to get something that's good enough and doesn't require configuration, I think that's usually better (sometimes it's not possible!).

> In general, I'm not sure about the wisdom of closing unresolved issues like this, as opposed to just marking them low-priority.

Okay, we can keep it open! But I've generally found that issues that people care about even slightly tend to have _some_ additional comments on them. This one is very old, and my reading of it was that nobody except @emilypi cared about it (you commented but I wasn't sure if you were in favour!).

The reality is that we have a _lot_ of issues for things that somebody maybe wanted once. I'm leaving a lot of them, but I am trying to cut down on ones that are particularly dead or don't seem popular.

PR #1609, opened 2021-03-23
Makes pragma completions case-insensitive

gdevanla:
Address the case-sensitive completions request in #681


jneira:
is this related with https://github.com/haskell/haskell-language-server/issues/729?

pepeiborra:
Looks plausible, but a test would be useful to verify that it works

jneira:
Superseded by #2142

PR #1611, opened 2021-03-24
CPS tactics

isovector:
I wanted to understand how refinery worked, so I decided to implement it myself. Some research goals: I thought we might be running into issues reusing `ProofState` for rule and that `proofs` likely had a type that was too specialized.

This PR implements `proofs` as a big monstrous CPS (but right now it only gives back one result --- presumably that part isn't too hard?) As best I can tell, it comes with the correct `commit` and `catch` semantics.

Open questions: 

* what needs to change here in order to get back ALL of the extracts?
* how do we implement `(<%>)`?

### DIFF
--- a/plugins/hls-tactics-plugin/hls-tactics-plugin.cabal
+++ b/plugins/hls-tactics-plugin/hls-tactics-plugin.cabal
@@ -77,6 +77,8 @@ library
     , text
     , transformers
     , deepseq
+    , QuickCheck
+    , checkers
 
   default-language: Haskell2010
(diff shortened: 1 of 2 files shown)


isovector:
@TOTBWF i'd love your eyes on this, if you've got some spare cycles

isovector:
I think accumulation can happen by adding a `m r -> m r -> m r` continuation to `kill`, and then having `Alt` and friends manipulate their success and failure continuations to always run both arguments.

PR #1612, opened 2021-03-24
Migrate tests of plugins

berberman:
This PR moves the tests of
* Class plugin
* Splice plugin
* Haddock comments plugin

from HLS's func-test to their packages.


berberman:
It seems that circle ci failed due to the lack of memory... 
```
--  While building package hls-eval-plugin-1.0.0.0 (scroll up to its section to see the error) using:
      /root/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_3.2.1.0_ghc-8.10.4 --builddir=.stack-work/dist/x86_64-linux/Cabal-3.2.1.0 build lib:hls-eval-plugin exe:test-server test:tests --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure (-9) (THIS MAY INDICATE OUT OF MEMORY)

Exited with code exit status 1
CircleCI received exit code 1

```
Perhaps we could decrease the parallelism (the number of packages we build at the same time), such as running `stack build -j3` in the job `Build Testsuite without running it`

jneira (approved):
Many thanks again, it is nice to see how plugins are being emancipated 🙂

PR #1617, opened 2021-03-24
Configurable I/O handles

pepeiborra:
Aimed at use in test suites with http://hackage.haskell.org/package/lsp-test-0.13.0.0/docs/Language-LSP-Test.html#v:runSessionWithHandles


berberman (approved):
LGTM

PR #1576, opened 2021-03-15
Regularize custom config of plugins

berberman:
With #691, each plugin is able to have a generic config:

```json
{
   "pluginName":{
      "globalOn":false,
      "codeActionsOn":false,
      "codeLensOn":false,
      "diagnosticsOn":false,
      "hoverOn":false,
      "symbolsOn":false,
      "formattingOn":false,
      "completionOn":false,
      "renameOn":false,
      "config":{
         
      }
 }
``` 
where particularly, `config` is an arbitrary JSON object representing the dedicated, or the extra custom config, which can be parsed and used by the plugin. However, this is hard to be documented and used from the clinet side -- I don't know much about other lsp clients, but for vscode, we have to write the [schema](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema) manually, like @jneira did in https://github.com/haskell/vscode-haskell/pull/361, which is kind of cumbersome. So it would be better to automate this process.

This PR includes mainly two changes:

* Defines a new data type `Properties` to let plugin type-safely describe what custom config it use
* Adds `--vscode-extension-schema` to ghcide and hls executables, which prints the generated schema for vscode
* Adds `--generate-default-config` to ghcide and hls executables, which prints the generated lsp json config with default values

## Properties

In order to make plugins' custom config compatible with vscode, i.e. it's expressible in vscode settings UI, a property is defiend as a direct child of `config` JSON object, where the value has five potential types (taken from references of vscode):

* `Number`
* `String`
* `Boolean`
* `Object`
* `Array`
* `Enum`

`Properties` is a set of *descriptions* of properties used by the plugin. It has a phantom type `r :: [*]` to make sure the name, or the key of the property is unique, and we can derive the corresponding haskell type of the property once we have its name. Given a property name and `Properties`, we also can derive a JSON parser of this proeprty, which acts on the `config` object. Overall, `Properties` is used in two places: vscode schema generation and parser derivation at runtime. Currently type lenses plugin and tactics plugin have their own custom configs, and let's use type lenses plugin as an exmaple:

```haskell
properties = emptyProperties
  & defineEnumProperty #mode "Control how type lenses are shown"
    [ ("always", "Always displays type lenses of global bindings")
    , ("exported", "Only display type lenses of exported global bindings")
    , ("diagnostics", "Follows error messages produced by GHC about missing signatures")
    ] "always"
```
[...]

PR #1619, opened 2021-03-25
Use custom config for completions plugin

berberman:
* Remove `haskell.completionSnippetsOn`
* Use `Properties` (#1576) to define the custom config, including switches of whether to enable snippets and auto extending import lists.

Clients now shoud provide JSON value like:
```json
"ghcide-completions":{
      "globalOn":true,
      "config":{
         "autoExtendOn":true,
         "snippetsOn":true
      },
      "completionOn":true
}
```

in the section `haskell.plugin` to control completions plugin.

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -35,7 +35,7 @@ import           Development.Shake
 import           Development.Shake.Classes
 import           GHC.Exts                                     (toList)
 import           GHC.Generics
-import           Ide.Plugin.Config                            (Config (completionSnippetsOn))
+import           Ide.Plugin.Config                            (Config)
 import           Ide.Types
 import qualified Language.LSP.Server                          as LSP
 import           Language.LSP.Types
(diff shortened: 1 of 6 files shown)


michaelpj:
Do we want to call it "ghcide-completions"? `ghcide` is more-or-less not user-facing any more, perhaps we should try and avoid mentioning the name in user-visible config? Could we just call it "completions"?

berberman:
The name "ghcide-completions" comes from its descriptor:
https://github.com/haskell/haskell-language-server/blob/b9c6e6cf4de5a0778c5ddb5d5e62fc8d1e0d78ef/ghcide/src/Development/IDE/Plugin/HLS/GhcIde.hs#L22-L29
In my opnion we'd better not introduce new names for the key, but use descriptors to keep things simple. Besides, I don't think it would be great to call the key "completions", because there might be other plugins are providing completion functionality as well.

jneira (approved):
lgtm

pepeiborra:
One option is to merge all the "ghcide-" descriptors into a single one and call it simply "HLS"

berberman:
Currently `CustomConfig` blocks defining the merge operation of descriptors, which means we need define the custom config for type lenses plugin and completions plugin in one `Properties` if we merge ghcide descriptors together. This would bring unnecessary couplings, imho.

PR #1620, opened 2021-03-25
Add CPP Options for Stylish Haskell & Brittany Formatters

prikhi:
This fixes an issue where the `stylish-haskell` & `brittany` plugins
weren't being added to the enabled plugins list, even if HLS was
compiled with their specific flags or the `all-formatters` flag.

Also removed the AGPL flag checking when adding the `brittany` plugin,
since that is no longer defined anywhere.


berberman (approved):
Thanks for the quick fix, you are faster than me :)

pepeiborra (approved):

PR #1621, opened 2021-03-25
Fix cpp flags of brittany plugin and stylish haskell plugin

berberman:
#1422 and #1604 extracted the two plugins into standalone packages respectively, whereas they missed setting corresponding cpp flags.


berberman:
Duplicate of #1620

PR #1623, opened 2021-03-25
Remove ignored paths

jneira:
* `paths-ignore` doesnt work with required checks branch protection rules:  https://github.community/t/path-filtering-on-required-pull-request-checks/18402/3
* so we have to remove it, fortunately `skip-duplicate-actions` which was being already used can be used to get the same functionality: https://github.com/marketplace/actions/skip-duplicate-actions#option-1-skip-entire-jobs

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -92,7 +92,6 @@ jobs:
   bench-example:
     needs: pre_job
     if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
-    needs: [bench-init]
     runs-on: ${{ matrix.os }}
 
     strategy:
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -4,18 +4,22 @@ on:
   pull_request:
     branches:
       - '**'
-    paths-ignore:
-      - '**/docs/**'
-      - '**.md'
-      - '**/LICENSE'
-      - '.circleci/**'
-      - 'install/**'
-      - 'nix/**'
-      - 'test/**'
-      - 'ghcide/test/**'
 
 jobs:
+  pre_job:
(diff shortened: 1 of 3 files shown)


.github/workflows/bench.yml
```diff
@@ -86,6 +90,8 @@ jobs:
         path: ~/.cabal/cabal.tar.gz
 
   bench-example:
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
     needs: [bench-init]
```
Ailrun, line 95: Is it possible to have two `needs` keys? I thought yaml is just more elaborated json...

berberman: In the case that `bench-init` is skipped, is `bench-example` still reachable if we don't add this?

Ailrun: It will skip its computation but from the GitHub's point of view I think the jot itself is not skipped, and will execute following jobs...

jneira: Even if the action lets you duplicate keys it is better to have only one, will change

.github/workflows/bench.yml
```diff
@@ -86,6 +90,8 @@ jobs:
         path: ~/.cabal/cabal.tar.gz
 
   bench-example:
+    needs: pre_job
```
jneira, line 93:
```suggestion
    needs: [pre_job, bench-init]
```

.github/workflows/bench.yml
jneira, line 95:
```suggestion
```

jneira (commented):
Remove duplicate `needs`

Ailrun (approved):
LGTM :)

Issue #1627, opened 2021-03-27
CI jobs of plugins' test suites got stuck randomly


berberman:
#1606 and #1612 moved plugins' test suites and created standalone CI jobs for them. However, according to the runs for recent PRs, it seemd that they got stuck randomly:

https://github.com/haskell/haskell-language-server/runs/2197372198
https://github.com/haskell/haskell-language-server/runs/2206602816
https://github.com/haskell/haskell-language-server/runs/2206604842

pepeiborra:
You need to make sure to build with `--threaded` and have 2 or more capabilities, or we could just fix #727

PR #1626, opened 2021-03-26
Display package names of external libraries on hover

berberman:
Resolves part of #686

PR #1628, opened 2021-03-27
Run plugins' test suites with server in the same process

berberman:
Following #1617

Hope this would make tests more stable


berberman:
Tests on Windows seemd got stuck again:
https://github.com/haskell/haskell-language-server/pull/1628/checks?check_run_id=2211603744
https://github.com/haskell/haskell-language-server/pull/1628/checks?check_run_id=2211603749

Isn't it enough to set `shakeThreads` to `2`?

hls-test-utils/src/Test/Hls.hs
```diff
@@ -9,38 +9,108 @@ module Test.Hls
...
+              argsDefaultHlsConfig = conf,
+              argsIdeOptions = \config sessionLoader ->
+                let ideOptions = (argsIdeOptions def config sessionLoader) {optTesting = IdeTesting True}
+                 in ideOptions {optShakeOptions = (optShakeOptions ideOptions) {shakeThreads = 2}},
+              argsHlsPlugins = pluginDescToIdePlugins $ plugin ++ Ghcide.descriptors
+            }
+  x <- withAsync server $ \_ ->
+    runSessionWithHandles inW outR sconf caps root s
+      `finally` setCurrentDirectory cwd
+  sleep 0.5
```
berberman, line 115: `0.5` is an arbitrary value. Perhaps we could use barrier here. But how do we know that all services are really shut down?

pepeiborra: Don't use `withAsync`. Instead, start the server using a normal `async` and then, once `runSession...` is over, wait on the server (with a timeout)

pepeiborra:
> Tests on Windows seemd got stuck again:
> https://github.com/haskell/haskell-language-server/pull/1628/checks?check_run_id=2211603744
> https://github.com/haskell/haskell-language-server/pull/1628/checks?check_run_id=2211603749
> 
> Isn't it enough to set `shakeThreads` to `2`?

You also need to build with `--threaded`

berberman:
> > Tests on Windows seemd got stuck again:
> > https://github.com/haskell/haskell-language-server/pull/1628/checks?check_run_id=2211603744
> > https://github.com/haskell/haskell-language-server/pull/1628/checks?check_run_id=2211603749
> > Isn't it enough to set `shakeThreads` to `2`?
> 
> You also need to build with `--threaded`

Mmm, `-threaded` it's already here:
https://github.com/haskell/haskell-language-server/blob/1f12a613ca43a9631bc35c4c719c7be393382163/plugins/hls-eval-plugin/hls-eval-plugin.cabal#L95-L100

berberman:
So far, tests of plugins on linux and macOS look great, but on Windows, they seem to hang100%. I don't have Windows machine with development environment setup, no idea how to debug this.

jneira:
I'll try to take a look in my local win 10

jneira:
Unfortunately tests pass in my local windows 10, after a first try with some `resource vanished` but no stuck. :worried:

berberman:
Ugh, it's still getting stuck... BTW definition and hover tests of ghcide is fairly unstable, and it even fails on my machine constantly: https://github.com/haskell/haskell-language-server/pull/1626#issuecomment-808875761

hls-test-utils/src/Test/Hls.hs
```diff
@@ -9,38 +9,116 @@ module Test.Hls
...
+  -- restore cwd after running the session; otherwise the path to test data will be invalid
+  cwd <- getCurrentDirectory
+  server <-
+    async $
+      Ghcide.defaultMain
+        def
+          { argsHandleIn = pure inR,
+            argsHandleOut = pure outW,
+            argsDefaultHlsConfig = conf,
+            argsLogger = pure noLogging,
```
pepeiborra, line 112: Why `noLogging`?

pepeiborra:
> Unfortunately tests pass in my local windows 10, after a first try with some `resource vanished` but no stuck. 😟

But this is the same problem that CI shows, isn't it? Tests are not stuck, they are failing in Windows GHC 8.10.4. "resource vanished - broken pipe" means that HLS stopped running. Either it was killed, it crashed, or it decided to exit on its own. 

Why is HLS crashing in Windows 8.10.4 all of a sudden? Has anything changed?

jneira:
> But this is the same problem that CI shows, isn't it?

Yeah, i thought they stuck without no response until a ci timeout, without reviewing the logs,thanks for pointing it out.
The symtomps are similar locally but the situation is much worse in ci, locally i got three errors in a cold first run and it passed in the second one.

> Why is HLS crashing in Windows 8.10.4 all of a sudden? Has anything changed?

It is failing since some time ago, although it seems in this pr is worse. I have been reviewing the gh actions and dont see a pattern. There are recent builds with succesful test run at first like https://github.com/haskell/haskell-language-server/runs/2197782538?check_suite_focus=true
Being transient errors and the differences between ci and my local machine make a bisect is hard 😟

hls-test-utils/src/Test/Hls.hs
berberman: I think we need suppress the output from server otherwise the test report would be unreadable. And `silenceStderr` is used together with it, since there are many `hPutStrLn stderr` which don't comply with the logger.

berberman:
Sometimes a dying server (after the test session finished successfully) throws this error:
```
tests.exe: Control.Concurrent.Extra.signalBarrier, attempt to signal a barrier that has already been signaled
  errorIO, called at src\Control\Concurrent\Extra.hs:196:16 in extra-1.7.9-1fb0534e9d12add2b941eefeb5d3e9bbdea5a70c:Control.Concurrent.Extra
  signalBarrier, called at src\Development\IDE\LSP\LanguageServer.hs:64:16 in ghcide-1.1.0.0-inplace:Development.IDE.LSP.LanguageServer
```
https://github.com/haskell/haskell-language-server/blob/d60dee03aca30aa4c04db2824a113a4006f95e13/ghcide/src/Development/IDE/LSP/LanguageServer.hs#L59-L64

Does it mean the server received two exit messages?

pepeiborra:
> > But this is the same problem that CI shows, isn't it?
> 
> Yeah, i thought they stuck without no response until a ci timeout, without reviewing the logs,thanks for pointing it out.
> The symtomps are similar locally but the situation is much worse in ci, locally i got three errors in a cold first run and it passed in the second one.
> 
> > Why is HLS crashing in Windows 8.10.4 all of a sudden? Has anything changed?
> 
> It is failing since some time ago, although it seems in this pr is worse. I have been reviewing the gh actions and dont see a pattern. There are recent builds with succesful test run at first like https://github.com/haskell/haskell-language-server/runs/2197782538?check_suite_focus=true
> Being transient errors and the differences between ci and my local machine make a bisect is hard 😟

This PR didn't make any changes to the ghcide test suite, so I don't see the connection. Unfortunately master has the getting-stuck problem which means we cannot see if the Windows issue is specific to this PR or not

berberman:
One difference between local machine and Windows CI I noticed (by checking lsp messages log one by one) was that in CI, there were no hiedb related messages sent from the server, such as  `ghcide/reference/ready`, `Finished indexing 1 files`.

berberman:
> > Sometimes a dying server (after the test session finished successfully) throws this error:
> > ```
> > tests.exe: Control.Concurrent.Extra.signalBarrier, attempt to signal a barrier that has already been signaled
> > CallStack (from HasCallStack):
> >   errorIO, called at src\Control\Concurrent\Extra.hs:196:16 in extra-1.7.9-1fb0534e9d12add2b941eefeb5d3e9bbdea5a70c:Control.Concurrent.Extra
> >   signalBarrier, called at src\Development\IDE\LSP\LanguageServer.hs:64:16 in ghcide-1.1.0.0-inplace:Development.IDE.LSP.LanguageServer
> > ```
> > 
> > 
> > https://github.com/haskell/haskell-language-server/blob/d60dee03aca30aa4c04db2824a113a4006f95e13/ghcide/src/Development/IDE/LSP/LanguageServer.hs#L59-L64
> > 
> > Does it mean the server received two exit messages?
> 
> Yes, I think so. Perhaps this is a new feature of lsp-test.
> 
> Great find! Please fix the crash by replacing the `Barrier` with a plain `MVar` and let's see if that unblocks the test suite!

But according to the lsp messages log, the client sends exactly one exit message 🤔

wz1000:
You might want to investigate/patch lsp-test. I believe the resource vanished error comes from this line: https://github.com/haskell/lsp/blob/6dd6ad630ddb4a08a2bbcd6f71588f0864a555b6/lsp-test/src/Language/LSP/Test/Session.hs#L440

It might be good to add a catch there, or at least print out more information for debugging (like the message we are trying to send).

hls-test-utils/src/Test/Hls.hs
```diff
@@ -9,38 +10,117 @@ module Test.Hls
...
+               in ideOptions {optShakeOptions = (optShakeOptions ideOptions) {shakeThreads = 2}},
+            argsHlsPlugins = pluginDescToIdePlugins $ plugin ++ Ghcide.descriptors
+          }
+
+  x <-
+    runSessionWithHandles inW outR sconf caps root s
+      `finally` setCurrentDirectory cwd
+  timeout 3 (wait server) >>= \case
+    Just () -> pure ()
+    Nothing -> putStrLn "Server does not exit on time, canceling the async task..." >> cancel server
```
berberman, line 125: Tests on Windows passed with this. We should investigate why the server didn't exit properly.

pepeiborra: I'm just looking at the exit code in ghcide and I think it might be time to update it. I'm not sure that the `waitAnyCancel` construction is needed any more, and it could be hiding exceptions coming from the `lsp` library:

https://github.com/haskell/haskell-language-server/blob/a4dee2d078dc85fb450ec9b35851d3ba3d84789d/ghcide/src/Development/IDE/LSP/LanguageServer.hs#L114-L120

pepeiborra: Another problem is that `defaultMain` does not call [`setupLogger`](https://hackage.haskell.org/package/lsp-1.2.0.0/docs/Language-LSP-Server.html#v:setupLogger) so we might be missing lsp errors

berberman: https://github.com/haskell/haskell-language-server/blob/6d1f1a55e39a9fd905bea13d6144125fcb42a594/ghcide/src/Development/IDE/LSP/LanguageServer.hs#L59-L62

These comments match the old code:

https://github.com/haskell/haskell-language-server/blob/fa671cbe1ee8f161df8d79d7a23962239d4d9e75/ghcide/src/Development/IDE/LSP/Server.hs#L73-L99

but now this barrier will be signalled when the server receives exit message instead.

I think we could try removing the `waitAnyCancel`.

pepeiborra: Yes, now we signal the barrier from two places - the exit handler and also if the initialisation fails, which probably explains the error that you were seeing earlier.

I'm not convinced that removing it will actually fix anything. `runServerWithHandles` doesn't actually exit until the input stream is closed, and I don't know whether `lsp-test` closes the stream until the server exits

berberman: It seems that `lsp-test` won't close the two handles for us (in our case `mServerProc` is `Nothing`):
https://github.com/haskell/haskell-language-server/blob/7a2ff3ad87b75812880bd580a98b120615bd1ef4/lsp-test/src/Language/LSP/Test/Session.hs#L282-L294

But I didn't find any manual close in the test of `lsp-test`. Maybe @wz1000 could help?

wz1000: I'm pretty sure the handle closes when the server exits.

pepeiborra: The handle closes when the server exits, sure, but as I said `runServerWithHandles` doesn't exit until the stream ends, so it won't exit.

berberman:
I observed some snippets from debug log, like this:
```
[Debug] Finishing build session(exception: Error when running Shake build system:
thread blocked indefinitely in an STM transaction
)
[Debug] Finishing build session(exception: Error when running Shake build system:
[Debug] Finishing build session(exception: Error when running Shake build system:
[Debug] Finishing build session(exception: Error when running Shake build system:
thread blocked indefinitely in an STM transaction
thread blocked indefinitely in an STM transaction
thread blocked indefinitely in an STM transaction
)))
```

pepeiborra:
> I observed some snippets from debug log, like this:
> 
> ```
> [Debug] Finishing build session(exception: Error when running Shake build system:
> thread blocked indefinitely in an STM transaction
> )
> [Debug] Finishing build session(exception: Error when running Shake build system:
> [Debug] Finishing build session(exception: Error when running Shake build system:
> [Debug] Finishing build session(exception: Error when running Shake build system:
> thread blocked indefinitely in an STM transaction
> thread blocked indefinitely in an STM transaction
> thread blocked indefinitely in an STM transaction
> )))
> ```

That is not necessarily a bug, helper threads like the reactor thread or the hiedb thread would fail like that when the main thread exits the lsp loop and no longer keeps a reference to the shared state

hls-test-utils/src/Test/Hls.hs
```diff
@@ -9,38 +10,126 @@ module Test.Hls
...
+          { argsHandleIn = pure inR,
+            argsHandleOut = pure outW,
+            argsDefaultHlsConfig = conf,
+            argsIdeOptions = \config sessionLoader ->
+              let ideOptions = (argsIdeOptions def config sessionLoader) {optTesting = IdeTesting True}
+               in ideOptions {optShakeOptions = (optShakeOptions ideOptions) {shakeThreads = 2}},
+            argsHlsPlugins = pluginDescToIdePlugins $ plugin ++ Ghcide.descriptors
+          }
+  x <- runSessionWithHandles inW outR sconf caps root s
+  hClose inW
```
berberman, line 127: In Windows CI, with `hClose inW` the server can exit spontaneously. But
```
tests.exe: <file descriptor: 11>: hPutBuf: resource vanished (Broken pipe)
```
arose after running each test case. The server tried to send something while the stream was closed.

berberman: According to the log, the server kept working after the input stream was closed, sending `window/logMessage`, `textDocument/publishDiagnostics`, and `$/progress` etc, and then crashed. This only happened on Windows CI. I'm sure the server had received exit message from client, with shake session restarted.

pepeiborra: That suggests restarting the shake session in the exit handler might be causing issues. `shakeShut` would be a better fit then

berberman:
The server sends "Finishing build session..." using `notifyTestingLogMessage` when a worker exits, but in our case, the input stream has been closed. Thus, calling `shakeShut` in exit message handler results in `hPutBuf: resource vanished (Broken pipe)`.

https://github.com/haskell/haskell-language-server/blob/252c500812e2309f80f39225ccf2e1d3cfda54a0/ghcide/src/Development/IDE/Core/Shake.hs#L726-L735

https://github.com/haskell/haskell-language-server/blob/252c500812e2309f80f39225ccf2e1d3cfda54a0/ghcide/src/Development/IDE/Core/Shake.hs#L673-L677

And it seems that the server still reports progress after shutting down the shake. Should we call `stopProgressReporting` before `cancelShakeSession` in `shakeShut`?
https://github.com/haskell/haskell-language-server/blob/252c500812e2309f80f39225ccf2e1d3cfda54a0/ghcide/src/Development/IDE/Core/Shake.hs#L618-L624

Also, is it possible to disable lsp capabilities when we want server die?

A piece of log from Windows CI:

```
brittany
Starting LSP server...
If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!
Started LSP server in 0.08s
setInitialDynFlags cradle: Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
Output from setting up the cradle Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
lsp:Got EOF, exiting 1 ...

warning: LF will be replaced by CRLF in C:/Users/runneradmin/AppData/Local/Temp/BriA841.actual.
The file will have its original line endings in your working directory
Starting LSP server...
  formats a document with LF endings:   OK (0.79s)
If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!
Started LSP server in 0.00s
  formats a document with CRLF endings: IO Exception: clientOut:
<file descriptor: 6>: hPutBuf: resource vanished (Broken pipe)
{"method":"window/logMessage","params":{"message":"Finishing build session(exception: AsyncCancelled)","type":4},"jsonrpc":"2.0"}
IO Exception: clientOut:
<file descriptor: 6>: hPutBuf: resource vanished (Broken pipe)
{"method":"window/logMessage","params":{"message":"Restarting build session (aborting the previous one took 0.00s)","type":4},"jsonrpc":"2.0"}
IO Exception: clientOut:
<file descriptor: 6>: hPutBuf: resource vanished (Broken pipe)
{"method":"window/logMessage","params":{"message":"Finishing build session(exception: AsyncCancelled)","type":4},"jsonrpc":"2.0"}
IO Exception: clientOut:
<file descriptor: 6>: hPutBuf: resource vanished (Broken pipe)
{"method":"$/progress","params":{"token":"14","value":{"message":"2/2","kind":"report"}},"jsonrpc":"2.0"}
setInitialDynFlags cradle: Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
Output from setting up the cradle Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
lsp:Got EOF, exiting 1 ...

warning: LF will be replaced by CRLF in C:/Users/runneradmin/AppData/Local/Temp/BriB274.actual.
The file will have its original line endings in your working directory
Starting LSP server...
OK (2.61s)
If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!
Started LSP server in 0.00s
setInitialDynFlags cradle: Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
Output from setting up the cradle Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
  formats a range with LF endings:      IO Exception: clientOut:
<file descriptor: 9>: hPutBuf: resource vanished (Broken pipe)
{"method":"window/logMessage","params":{"message":"Finishing build session(exception: AsyncCancelled)","type":4},"jsonrpc":"2.0"}
IO Exception: clientOut:
<file descriptor: 9>: hPutBuf: resource vanished (Broken pipe)
{"method":"window/logMessage","params":{"message":"Restarting build session (aborting the previous one took 2.03s)","type":4},"jsonrpc":"2.0"}
IO Exception: clientOut:
<file descriptor: 9>: hPutBuf: resource vanished (Broken pipe)
{"method":"window/logMessage","params":{"message":"Finishing build session(exception: AsyncCancelled)","type":4},"jsonrpc":"2.0"}
IO Exception: clientOut:
<file descriptor: 9>: hPutBuf: resource vanished (Broken pipe)
{"method":"$/progress","params":{"token":"37","value":{"message":"2/2","kind":"report"}},"jsonrpc":"2.0"}
lsp:Got EOF, exiting 1 ...

warning: LF will be replaced by CRLF in C:/Users/runneradmin/AppData/Local/Temp/BriB4B9.actual.
The file will have its original line endings in your working directory
Starting LSP server...
OK (0.57s)
If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!
Started LSP server in 0.00s
setInitialDynFlags cradle: Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
Output from setting up the cradle Cradle {cradleRootDir = "D:\\a\\haskell-language-server\\haskell-language-server\\plugins\\hls-brittany-plugin\\test\\testdata", cradleOptsProg = CradleAction: Default}
lsp:Got EOF, exiting 1 ...

warning: LF will be replaced by CRLF in plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.formatted_range.hs.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in C:/Users/runneradmin/AppData/Local/Temp/BriB6AF.actual.
The file will have its original line endings in your working directory
  formats a range with CRLF endings:    OK (0.51s)

All 4 tests passed (4.47s)
Test suite tests: PASS
Test suite logged to:
D:\a\haskell-language-server\haskell-language-server\dist-newstyle\build\x86_64-windows\ghc-8.10.4\hls-brittany-plugin-1.0.0.0\t\tests\test\hls-brittany-plugin-1.0.0.0-tests.log
1 of 1 test suites (1 of 1 test cases) passed.
```

pepeiborra:
`shakeShut` already calls progress reporting - could you try to reorder the statements so that it calls it first?

This output is caused by `lsp-test` closing the stream immediately in Windows, whereas it waits for the server to exit in Linux:

https://github.com/haskell/lsp/blob/6dd6ad630ddb4a08a2bbcd6f71588f0864a555b6/lsp-test/src/Language/LSP/Test/Session.hs#L283-L288

So it's probably not a big deal

berberman:
But as I said here: https://github.com/haskell/haskell-language-server/pull/1628#discussion_r606144937, if we start test session using `runSessionWithHandles`, `mServerProc` should be `Nothing` -- there will be no cleanup

pepeiborra:
> But as I said here: [#1628 (comment)](https://github.com/haskell/haskell-language-server/pull/1628#discussion_r606144937), if we start test session using `runSessionWithHandles`, `mServerProc` should be `Nothing` -- there will be no cleanup

Ok, fair enough, so then I suppose the garbage collector eventually closes it when `runSessionWithHandles` returns. Maybe we should keep both sides of the handle alive in the server thread, using e.g. [`#touch`](https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Exts.html#v:touch-35-).

hls-test-utils/src/Test/Hls.hs
```diff
@@ -126,6 +127,8 @@ runSessionWithServer' plugin conf sconf caps root s = withLock lock $ keepCurren
           }
   x <- runSessionWithHandles inW outR sconf caps root s
   hClose inW
+  touch outW
```
pepeiborra, line 130: does it work? If it stays, please add a comment explaining the rationale for future maintainers

berberman: unfortunately it doesn't :(
I'll try again putting it before `sleep 0.2`. If it still doesn't work, I think we could ignore this minor issue first -- which seems to occur only on Windows CI, having no effect on test result -- and then cleanup, merge this PR to address #1627

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -126,10 +126,9 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             registerIdeConfiguration (shakeExtras ide) initConfig
 
 
-            let handleServerException (Left e) = do
+            let handleServerException (Left e) =
                     logError (ideLogger ide) $
                         T.pack $ "Fatal error in server thread: " <> show e
-                    exitClientMsg
```
pepeiborra, line 147: Did you really want to remove this and why?

berberman: nvm, I'm not going to remove that, just for resolving the merge conflicts online to run CI

hls-test-utils/src/Test/Hls.hs
```diff
@@ -9,38 +10,128 @@ module Test.Hls
...
+          }
+  x <- runSessionWithHandles inW outR sconf caps root s
+  hClose inW
+  timeout 3 (wait server) >>= \case
+    Just () -> pure ()
+    Nothing -> do
+      putStrLn "Server does not exit in 3s, canceling the async task..."
+      (t, _) <- duration $ cancel server
+      putStrLn $ "Finishing canceling (took " <> showDuration t <> "s)"
+  sleep 0.2
```
pepeiborra, line 136: can you add a comment on the purpose of this `sleep`?

pepeiborra (approved):

hls-test-utils/src/Test/Hls.hs
berberman: Maybe tests are stable enough without this. Let's have a try.

berberman:
The barrier did be signaled two times. Is this exit what we expected?
```
[Error] Fatal error in server thread: thread blocked indefinitely in an MVar operation
tests: Control.Concurrent.Extra.signalBarrier, attempt to signal a barrier that has already been signaled
  errorIO, called at src/Control/Concurrent/Extra.hs:196:16 in extra-1.7.9-B289Y5Rzkww81ONKh8ABWZ:Control.Concurrent.Extra
  signalBarrier, called at src/Development/IDE/LSP/LanguageServer.hs:64:16 in ghcide-1.1.0.0-inplace:Development.IDE.LSP.LanguageServer
```

Full test log:
```
  formats a range with CRLF endings:    Starting LSP server...
If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!
Started LSP server in 0.00s
setInitialDynFlags cradle: Cradle {cradleRootDir = "/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata", cradleOptsProg = CradleAction: Default}
[Info] Registering ide configuration: IdeConfiguration {workspaceFolders = fromList [NormalizedUri (-6923107000992859315) "file:///home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata"], clientSettings = hashed Nothing}
[Debug] Set files of interest to: [(NormalizedFilePath "/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.hs",Modified {firstOpen = True})]
[Debug] Finishing build session(exception: AsyncCancelled)
[Debug] Restarting build session (aborting the previous one took 0.00s)
[Debug] Opened text document: file:///home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.hs
[Info] Consulting the cradle for "BrittanyCRLF.hs"
[Warning] No [cradle](https://github.com/mpickering/hie-bios#hie-bios) found for BrittanyCRLF.hs.
 Proceeding with [implicit cradle](https://hackage.haskell.org/package/implicit-hie).
You should ignore this message, unless you see a 'Multi Cradle: No prefixes matched' error.
Output from setting up the cradle Cradle {cradleRootDir = "/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata", cradleOptsProg = CradleAction: Default}
[Debug] Session loading result: Right (ComponentOptions {componentOptions = ["-dynamic"], componentRoot = "/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata", componentDependencies = []},"/usr/lib/ghc-8.10.4")
[Info] Using interface files cache dir: /home/berberman/.cache/ghcide/main-1a596a151463f2c53ee4feb14ecd276a1ccebfda
[Info] Making new HscEnv[main]
[Debug] New Component Cache HscEnvEq: (([],Just HscEnvEq 87),fromList [])
[Debug] Known files updated: fromList [(TargetFile NormalizedFilePath "/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.hs",fromList ["/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.hs"])]
[Debug] Restarting build session (aborting the previous one took 0.00s)
[Debug] Finishing build session(exception: AsyncCancelled)
[Info] finish: brittany (took 0.00s)
[Error] Fatal error in server thread: thread blocked indefinitely in an MVar operation
tests: Control.Concurrent.Extra.signalBarrier, attempt to signal a barrier that has already been signaled
  errorIO, called at src/Control/Concurrent/Extra.hs:196:16 in extra-1.7.9-B289Y5Rzkww81ONKh8ABWZ:Control.Concurrent.Extra
  signalBarrier, called at src/Development/IDE/LSP/LanguageServer.hs:64:16 in ghcide-1.1.0.0-inplace:Development.IDE.LSP.LanguageServer
[Debug] finish: InitialLoad (took 0.10s)
[Debug] Set files of interest to: [(NormalizedFilePath "/home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.hs",Modified {firstOpen = False})]
[Debug] Finishing build session(exception: AsyncCancelled)
[Debug] Restarting build session (aborting the previous one took 0.00s)
[Debug] Modified text document: file:///home/berberman/Desktop/haskell/haskell-language-server/plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.hs
[Debug] Finishing build session(exception: AsyncCancelled)
OK (0.45s)
```

pepeiborra:
I thought you replaced this barrier with an MVar at some point - that change makes sense to me, since we call it from two places and therefore it shouldn't be a barrier.

The "thread blocked indefinitely in an MVar operation"  error is ugly - it means that the main loop has finished and no longer holds a reference to the shared state, but there are background threads still waiting on MVars. It would be good to identify which threads and clean them up properly, or just make sure we use `withAsync` instead of `async` where possible

plugins/hls-tactics-plugin/hls-tactics-plugin.cabal
```diff
@@ -107,22 +107,6 @@ library
     TypeOperators,
     ViewPatterns
 
-
-executable test-server
```
isovector, line 111: This is an excellent change. Thanks!

jneira (approved):
Great work, thanks!

berberman:
Branch protection feels unhappy with the skipped nix jobs :(

jneira:
> 
> 
> Branch protection feels unhappy with the skipped nix jobs :(

Umm, i have to change it to skip steps instead the entire job, like https://github.com/haskell/haskell-language-server/pull/1656

hls-test-utils/src/Test/Hls.hs
```diff
@@ -9,38 +10,127 @@ module Test.Hls
...
+silenceStderr :: IO a -> IO a
+silenceStderr action = withTempFile $ \temp ->
+  bracket (openFile temp ReadWriteMode) hClose $ \h -> do
+    old <- hDuplicate stderr
+    buf <- hGetBuffering stderr
+    h `hDuplicateTo'` stderr
+    action `finally` do
+      old `hDuplicateTo'` stderr
+      hSetBuffering stderr buf
+      hClose old
```
anka-213, line 88: Does this mean that we can no longer use `LSP_TEST_LOG_STDERR=1` to get more info about a failing test, since silencing stderr is now hard coded?

jneira: //cc @berberman

pepeiborra: When I attempted this for all the ghcide testsuite, I installed a logger that sends output to the LSP channel. 

https://github.com/haskell/haskell-language-server/pull/1752/files?file-filters%5B%5D=.hs&file-filters%5B%5D=.project#diff-46837fe41900682e7ad33f73eee0194cde2e864e3aebd7775d412660b61029fcR1

anka-213: We could probably check for the environment variable manually and pretend that it was lsp-test that did it, to minimize disturbance of people's workflows.

(Here's lsp-test's code for it: https://github.com/haskell/lsp/blob/e707cbf5ca7077f70884ae0d2a8d016aa30ced5a/lsp-test/src/Language/LSP/Test.hs#L267-L275)

That way the current uses, like in [ci](https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/test.yml#L161) doesn't have to change. Regardless of what we do, it should probably be documented together with the `LSP_TEST_LOG_MESSAGES` env variable in the [contributing](https://github.com/haskell/haskell-language-server/blob/master/docs/contributing/contributing.md) section, which it is currently not.

jneira: agree, pr's welcome!

berberman: > Does this mean that we can no longer use `LSP_TEST_LOG_STDERR=1` to get more info about a failing test, since silencing stderr is now hard coded?

Yes, output from lsp server is completely removed in test suites which use this function since that change, because there were too many logs were printed not through the logger, but were directly written to stderr, messing up testing status. So setting environment variable LSP_TEST_LOG_MESSAGES actually won't work. However, this function is a temporary workaround, and we should cleanup such code in server.

PR #1629, opened 2021-03-27
Avoid reordering plugins

pepeiborra:
Order of execution matters for notification plugins, so best to avoid unnecessary reorderings

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -36,6 +36,7 @@ import           Development.IDE.Core.PositionMapping     (PositionResult (..),
                                                            positionResultToMaybe,
                                                            toCurrent)
 import           Development.IDE.Core.Shake               (Q (..))
+import           Development.IDE.Main                     as IDE
 import           Development.IDE.GHC.Util
 import           Development.IDE.Plugin.Completions.Types (extendImportCommandId)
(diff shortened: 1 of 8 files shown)


ghcide/test/exe/Main.hs
```diff
@@ -5143,21 +5152,26 @@ runInDir' dir startExeIn startSessionIn extraOptions s = do
...
+
+getConfigFromEnv :: IO SessionConfig
+getConfigFromEnv = do
   logColor <- fromMaybe True <$> checkEnv "LSP_TEST_LOG_COLOR"
   timeoutOverride <- fmap read <$> getEnv "LSP_TIMEOUT"
-  let conf = defaultConfig{messageTimeout = fromMaybe (messageTimeout defaultConfig) timeoutOverride}
-            -- uncomment this or set LSP_TEST_LOG_STDERR=1 to see all logging
-            --   { logStdErr = True }
-            --   uncomment this or set LSP_TEST_LOG_MESSAGES=1 to see all messages
-            --   { logMessages = True }
```
alanz, line 5153: Are logMessages/logStdErr no longer used, or this documented elsewhere?

alanz (approved):
LGTM

ghcide/test/exe/Main.hs
pepeiborra: It's much better to use the environment variables (documented in the lsp-test Haddocks)

PR #1631, opened 2021-03-28
Update to lsp-1.2

wz1000:


wz1000:
A lot of code action tests are failing because they use incorrect ranges and `lsp-test` recently became more strict about this: https://github.com/haskell/lsp/commit/f90b00266ea4c750df913fbfa6802f5eb24bb426

:( It is a lot of work to fix all of them.

jneira:
MAybe we could restore the old function in lsp, make it deprecated and translate tests progressively?

pepeiborra (approved):
Very nice!

wz1000:
Benchmarks seem to show some very minor improvements in allocations, possibly because `lsp-types` started using `StrictData`.

PR #1633, opened 2021-03-28
Civilized indexing progress reporting

pepeiborra:
https://user-images.githubusercontent.com/26626/112764331-e23ccd00-8fff-11eb-96fe-7149223b50b6.mov

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -545,7 +545,7 @@ indexHieFile se mod_summary srcPath hash hf = atomically $ do
               _ <- LSP.sendRequest LSP.SWindowWorkDoneProgressCreate (LSP.WorkDoneProgressCreateParams u) (const $ pure ())
               LSP.sendNotification LSP.SProgress $ LSP.ProgressParams u $
                 LSP.Begin $ LSP.WorkDoneProgressBeginParams
-                  { _title = "Indexing references from:"
+                  { _title = "Indexing references"
                   , _cancellable = Nothing
                   , _message = Nothing
                   , _percentage = Nothing
@@ -557,14 +557,12 @@ indexHieFile se mod_summary srcPath hash hf = atomically $ do
         remaining <- HashMap.size <$> readTVar indexPending
         pure (done, remaining)
 
-      let progress = " (" <> T.pack (show done) <> "/" <> T.pack (show $ done + remaining) <> ")..."
-
       whenJust (lspEnv se) $ \env -> whenJust tok $ \tok -> LSP.runLspT env $
         LSP.sendNotification LSP.SProgress $ LSP.ProgressParams tok $
           LSP.Report $ LSP.WorkDoneProgressReportParams
             { _cancellable = Nothing
-            , _message = Just $ T.pack (fromNormalizedFilePath srcPath) <> progress
-            , _percentage = Nothing
+            , _message = Nothing
+            , _percentage = Just (100 * fromIntegral done / fromIntegral (done + remaining) )
             }
 
     -- Report the progress once we are done indexing this file
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -507,7 +507,9 @@ spliceExpresions Splices{..} =
 -- can just increment the 'indexCompleted' TVar and exit.
 --
 indexHieFile :: ShakeExtras -> ModSummary -> NormalizedFilePath -> Fingerprint -> Compat.HieFile -> IO ()
-indexHieFile se mod_summary srcPath hash hf = atomically $ do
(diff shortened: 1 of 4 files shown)


ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -557,14 +557,12 @@ indexHieFile se mod_summary srcPath hash hf = atomically $ do
...
-      let progress = " (" <> T.pack (show done) <> "/" <> T.pack (show $ done + remaining) <> ")..."
-
       whenJust (lspEnv se) $ \env -> whenJust tok $ \tok -> LSP.runLspT env $
         LSP.sendNotification LSP.SProgress $ LSP.ProgressParams tok $
           LSP.Report $ LSP.WorkDoneProgressReportParams
             { _cancellable = Nothing
-            , _message = Just $ T.pack (fromNormalizedFilePath srcPath) <> progress
-            , _percentage = Nothing
+            , _message = Nothing
+            , _percentage = Just (100 * fromIntegral done / fromIntegral (done + remaining) )
```
wz1000, line 565: LSP says the `_percentage` has to be monotonic, but this may not be. That is why I chose not to fill this field in.

pepeiborra: It says "steadily rising", and "clients are free to ignore values that do not follow this rule". 
Sounds like we should fill it in, and let clients decide what to do. 

But I'm equally happy to just leave it empty

wz1000: Can we keep the progress messages in the message?

Also, is there any particular reason why you want this change? I find the messages quite useful.

pepeiborra: It's very distracting on large repos. I will add an option to choose the style

ndmitchell: An option seems like it might not be worth it - its another choice a user has to make, and the value of making the choice seems outweighed by the existence of the choice.

FWIW, in Rust, the indexing shows indexing by package name. If we did that, would there be any problem on large repos? Or do we usually only index one package?

wz1000: I don't have strong opinions about keeping module names/files in the message, though I think keeping explicit progress numbers would be good.

For me, having the module names in there acts as a proxy for figuring out what files HLS is recompiling, which I find useful to track the state of the process and keep an eye out for in case I think its recompiling something I think it shouldn't.

I agree that an option doesn't seem ideal.

ndmitchell: Is the issue for large repos that it changes very often, to the point of almost flickering? If so, one way would be to say that we only show a module name after it's been taking > 0.3s to compile (and continue displaying it for 0.3s after it completes). That let's you see painful files, and if those painful files shouldn't be compiling people can act, but in many cases it won't say anything?

wz1000: Yes, something like that was on my TODO list as well. We should have a separate thread for reporting progress, and it can be responsible for debouncing as well.

That would also allow us to respect the spec in terms of progress messages. We aren't supposed to start sending progress events until we get a reply to the `progressCreate` message. I had implemented this (it is just a couple of lines waiting for a barrier), but removed it so as not to block the indexing thread on a response from the server. (also, all the other progress messages in ghcide also violate the spec in the same way, and I didn't have the stamina to fix those too).

pepeiborra: > An option seems like it might not be worth it - its another choice a user has to make, and the value of making the choice seems outweighed by the existence of the choice.
> I agree that an option doesn't seem ideal.
A user option wouldn't be ideal, agreed, however I wasn't talking about a user option, but a library option (`IdeOptions`) to control how progress is reported: via the `_percentage` field, via explicit numbers ("25/70") or none at all. HLS would then fix this to textual numbers to preserve the existing behaviour. 

> For me, having the module names in there acts as a proxy for figuring out what files HLS is recompiling, which I find useful to track the state of the process and keep an eye out for in case I think its recompiling something I think it shouldn't.
This information belongs in the diagnostics logs.

> Is the issue for large repos that it changes very often, to the point of almost flickering?
Yes, it refreshes several times a second and includes absolute paths, making it unreadable. Even if it was readable it conveys no useful information anyway, since indexing is an implementation detail.

>  all the other progress messages in ghcide also violate the spec in the same way, 

A simple improvement for this is to report progress only for a subset of rules, namely `GenerateCore`, `Typecheck` and `GetModIface`. The current behaviour when it progresses to the end, then back to 0, then back to the end is the result of reporting progress for all the rules.

But to make it truly monotonous we would need additional state. Something for another day though.

pepeiborra:
Updated with an option to choose the progress reporting style which defaults to the explicit style but allows to select percentage.

wz1000 (approved):

Issue #297, opened 2020-08-08
Support for GHC-9.0


mouse07410:
## State of support

* The core functionality of haskell-language-server is supported. 
* Only some plugins and formatters needs to be updated to make it work with ghc-9.0.1 so they are not included until then.
* So you should consider use other ghc version only if you are specially interested in those plugins o formatters 

## Has ghc-9.0 support
- [x] `ghcide`
- [x] `haskell-language-server`
- [x] `hls-haddock-comments-plugin`
- [x] `hls-hlint-plugin`
- [x] `hls-importlens-plugin`
- [x] `hls-pragmas-plugin`
- [x] `hls-retrie-plugin`
- [x] `hls-eval-plugin` #1997
- [x] `hls-module-name-plugin` - Actually builds just fine. Was accidentally disabled
- [x] `hls-class-plugin` - #2475
- [x] `hls-refine-imports-plugin` - All deps build
- [x] `hls-splice-plugin` - All deps build
- [x] `hls-tactic-plugin` 
  - Blocked upstream by ghc-source-gen (https://github.com/google/ghc-source-gen/issues/77 / https://github.com/google/ghc-source-gen/pull/84)
  - Disabled some tests in GHC9 because they expect Wingman to build. See #1873 and the `expectFailIfGhc9` function
  - Pull request in progress: #2202
  - Finally added with #2582, follow up of the previous one

### Formatters
- [x] `hls-floskell-plugin`
- [x] `hls-ormolu-plugin`  - Support added with https://github.com/tweag/ormolu/issues/688
- [x] `hls-brittany-plugin` - Blocked upstream by brittany:ghc-lib-parser (https://github.com/lspitzner/brittany/issues/352)
- [X] `hls-fourmolu-plugin` - #2254
- [x] `hls-stylish-haskell-plugin` blocked upstream by 
  - stylish-haskell:ghc-lib-parser (https://github.com/haskell/stylish-haskell/issues/378)
  - https://github.com/haskell/stylish-haskell/issues/387

### Build system and CI
- [x] `stack-9.0.1.yaml` - Disables not-yet-working plugins and fetches upstream patches
- [x] `cabal-ghc901.project` - Used for disabling the not yet working plugins and fetching upstream patches
- [x] `flake.nix` - Needs to fetch upstream patches and disable not-yet-working plugins (#1995)
- [x] `.circleci/config.yml`
[...]

PR #1635, opened 2021-03-29
hie-compat: Add basic support for ghc 9.0.1

anka-213:
I made a three way merge between the original in GHC-8.10, GHC-9.0 and the custom version for GHC-8.10 from this package, in order to figure out what should be changed.

Perhaps we don't actually need the full fork of `HieAst.hs` for ghc-9.0, since the difference from the original is so small, so we could just reexport the original and replace `mkHieFile` with the custom version.

(Tiny step towards #297)


jneira:
Many thanks for working on ghc-9.0.1 support 🙏

anka-213:
All of these changes are in the new PR: https://github.com/haskell/haskell-language-server/pull/1649, so I'll just close this.

hie-compat/src-ghc901/Compat/HieBin.hs
```diff
@@ -0,0 +1,371 @@
+{-
```
wz1000, line 1: Does this need to be duplicated? I think it should be possible to just reexport `GHC.Iface.Ext.Binary`

hie-compat/src-ghc901/Compat/HieAst.hs
```diff
@@ -0,0 +1,2031 @@
+{-
+Forked from GHC v9.0.1 to work around the readFile side effect in mkHiefile
```
wz1000, line 2: I also don't think we need to duplicate this. The "side effect free" `mkHieFile` is called `mkHieFileWithSource` in the GHC 9.0 module. A rename/re-export should suffice.

hie-compat/src-ghc901/Compat/HieBin.hs
anka-213: Not sure. I just copied whatever was in the ghc-810 versions and updated it to the ghc-901 version.

hie-compat/src-ghc901/Compat/HieAst.hs
anka-213: Ok, that's great

hie-compat/src-ghc901/Compat/HieBin.hs
```diff
@@ -0,0 +1,371 @@
...
+fromHieName :: NameCache -> HieName -> (NameCache, Name)
+fromHieName nc (ExternalName mod occ span) =
+    let cache = nsNames nc
+    in case lookupOrigNameCache cache mod occ of
+         Just name
+           | nameSrcSpan name == span -> (nc, name)
+           | otherwise ->
+             let name' = setNameLoc name span
+                 new_cache = extendNameCache cache mod occ name'
+             in ( nc{ nsNames = new_cache }, name' )
```
anka-213, line 330: @wz1000 
This code seems to be changed in the hie-compat version. Not sure if this is intentional.

In the version in ghc, these lines are just
```
 Just name -> (nc, name)
 ```
 
 I have no clue what this code does and which version should be used  (if it matters at all).

wz1000: I think we don't need this change any longer since the `hiedb` stuff landed. It was originally from this commit: https://github.com/haskell/ghcide/commit/96d1f30ff84dc87d59118630d6b6dbfb

If you just re-export GHCs module and the tests pass, we should be good to go.

wz1000: Even if it does turn out to be needed, I think we can directly modify the `NameCacheUpdater` we pass to `readHieFile` instead of duplicating the code.

hie-compat/src-ghc901/Compat/HieAst.hs
anka-213: I've removed everything but the modified version of `mkHieFile`, which calls `mkHieFileWithSource` internally now.

wz1000 (approved):

hie-compat/src-ghc901/Compat/HieBin.hs
anka-213: Do you know which test would test for this?

wz1000: All the definition tests: see https://github.com/haskell/ghcide/pull/677

anka-213:
Is there a reason why the circleci tests are not running? Is it waiting for something else first?

Ailrun:
There's nothing on the queue (even this one), so it's an actual problem. I don't know whether it's a CircleCI bug or our setting's. One possible thing to try is add one commit (or rebase this PR) so that CI check reactivated.

jneira:
The circleci job for ead87d8 finished succesfully but the status was not changed: https://app.circleci.com/pipelines/github/haskell/haskell-language-server/3004/workflows/999e4192-a140-4e0f-a99f-ffeea46d5778
I've reran it, lets see what happens: https://app.circleci.com/pipelines/github/haskell/haskell-language-server/3004/workflows/ecd57e2f-3184-4032-87cf-63320e5854b6

Ailrun:
It looks like the linked CI run is for ead87d8 and the problematic one is for 47a7b44, so unfortunately I don't think rerun will help...

jneira:
Mmm you are right, the merge was clean and maybe circleci has some trouble with that this time

anka-213:
Is it possible to manually trigger circleci, or should I rebase, so it can have a new chance to trigger?

anka-213:
I'll do that and then it should hopefully work as soon as someone re-approves the changes

anka-213:
Maybe it didn't think it was worth re-running the CI because the merge only changed the README?

jneira:
> 
> 
> Maybe it didn't think it was worth re-running the CI because the merge only changed the README?
> 
> It seems to run the circleci again now.

Unfortunately circleci dont support skip builds by file globs so not sure what happened (otoh checks would ne not required in that case)
Afaik it does not support neither new manual builds chosing a commit of a branch out of the box.  :neutral_face:

PR #1640, opened 2021-03-30
Update ISSUE_TEMPLATE.md

Ailrun:
To include OS

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -9,6 +9,8 @@ Output of `haskell-language-server --probe-tools` or `haskell-language-server-wr
 ```sh
 
 ```
+Which OS do you use:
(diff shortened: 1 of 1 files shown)


jneira (approved):
Wow, that was fast, many thanks

Ailrun:
@jneira Unfortunately, it looks like CI skipping still does not work... Maybe we are skipping things too early?

Ailrun:
Anyway, I will merge this manually for now

jneira:
mmm in the report I see 12/16 successful checks and there are 4 circleci unfinished jobs (they are not skipped only gha ones)

Ailrun:
@jneira The report says that, but the GitHub branch rule does not work with that 12 successful checks.

jneira:
mmm after circleci jobs has finished all checks are green and I think you had been able to merge it
as commented above circleci jobs are not affected by any skip rule so they are executed and handled like in any other pr (so you can't merge if they fail or are unfinished)

PR #1644, opened 2021-03-31
Fix importing type operators

### LINKED ISSUE #1597: Importing type (synonym/family??) operators doesn't work properly
version:
```
tauren :: ~ » haskell-language-server-wrapper --version
haskell-language-server version: 1.0.0.0 (GHC: 8.10.4) (PATH: /home/googleson78/.ghcup/bin/haskell-language-server-wrapper-1.0.0) (GIT hash: 4cd1cf934638881e52b3eba9f70157a4b799c0e9)
```

When trying to import a type synonym operator (possibly also for families), you need to use (e.g.) `import Fcf (type (@@))`. `hls` however only generates `import Fcf ((@@))`

berberman:
Closes #1597

I don't think we could know if a thing is pattern synonym given `AvailInfo`, so this PR assumes that all parenthesized parents must be type operators.


googleson78:
What is a "parenthesized parent"?

berberman:
> What is a "parenthesized parent"?

It's a type constructor which is an operator, such as `:~:`, `~~`, etc. It's so called because it's wrapped with parentheses, stored into `ExportsMap`.

pepeiborra:
What is the behaviour for pattern synonyms? Could you add a test showing that

pepeiborra (approved):

PR #1645, opened 2021-03-31
Make Wingman generate numbered holes

isovector:
Some of the upcoming changes for refinery v4 necessitate giving unique identifiers to each hole generated by Wingman. This PR makes that happen in anticipation.


jneira (approved):

isovector:
Accidentally closed this

Ailrun:
Does this work when there exists a numbered hole already? Like
```
f :: Bool -> Bool -> Int
f True _ = _0
f False _ = _ {- can we split the second argument here? -}
```

isovector:
Good catch. That particular case works, but if the hole exists in the same expression it might cause some naughtiness.

Ailrun:
Then it's might be better to add some test cases for those

Issue #32, opened 2020-02-02
hlint plugin


alanz:
Use the same approach as for DAML, which re-uses the parsed AST.

From IRC: 

> \<cocreature\> alanz: our dlint integration takes the AST instead of reparsing the module which is a bit more efficient, see https://github.com/digital-asset/daml/blob/master/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs#L1081

alanz:
@cocreature @ndmitchell 
 
I am trying the DAML approach for hls, but getting a mismatch on ghc-lib equivalents for the GHC types. Does this approach require the `ghc-lib` flag to be activated on `ghcide`? And what are the implications of doing that?  Can hlint be run on straight GHC modules?

ndmitchell:
HLint only supports one major version of GHC at a time, namely 8.8 at the moment. For every other version it uses ghc-lib-8.8. For DAML there is only one version of GHC and it's not a problem. My guess is that for hls we'll have to use the AST when you are on 8.8, and on older versions, try passing equivalent flags to the parse bit of hlint itself. I'm not sure if there are any negative interactions of putting ghc and ghc-lib in the same process - @cocreature and @shayne-fletcher-da will know.

shayne-fletcher:
In general using  ghc and `ghc-lib` in the same process we expect to work. Note though, that usage of `ghc-lib` in that context will require to use  package import syntax. HLint does not use package import syntax; it calls onto either ghc or `ghc-lib` depending on the compiler it's built with.

alanz:
> HLint does not use package import syntax; it calls onto either ghc or ghc-lib depending on the compiler it's built with.

So perhaps the hlint plugin for `hls` should be in a sublibrary, with the appropriate conditionals in the cabal definition to achieve the same effect. Would that work?

shayne-fletcher:
That sounds eminently plausible.

Here's how it's done in `hlint.cabal`:
- Define a flag
```
flag ghc-lib
  default: False
  manual: True
  description: Force dependency on ghc-lib-parser even if GHC API in the ghc package is supported
```
-  Then in the `library` stanza
```
    if !flag(ghc-lib) && impl(ghc >= 8.8.0) && impl(ghc < 8.9.0)
        build-depends:
          ghc == 8.8.*,
          ghc-boot-th,
          ghc-boot
    else
        build-depends:
          ghc-lib-parser == 8.8.*
```

[...]

PR #1649, opened 2021-04-02
Ghc 9.0.1 support for ghcide

anka-213:
This builds on top of #1635

Significant progress on #297

It includes a bunch of upstream fixes via `source-repository-package` in `cabal.project`:
- https://github.com/jwaldmann/blaze-textual/commit/d8ee6cf80e27f9619d621c936bb4bda4b99a183f
- https://github.com/lspitzner/czipwith/pull/2
- https://github.com/mpickering/hie-bios/pull/285
- https://github.com/mokus0/th-extras/pull/8
- https://github.com/pepeiborra/ghc-check/pull/12
- https://github.com/obsidiansystems/dependent-sum/pull/57
- https://github.com/wz1000/HieDb/pull/31
- https://github.com/haskell/lsp/pull/312
- https://github.com/mpickering/apply-refact/issues/107

It also sets `allow-newer: *` in `cabal.project`, since that was the fastest way to get things going and the process to figure out the exact bounds is tedious.

I've added a whole bunch of compatibility functions to `Development.IDE.GHC.Compat`, many of which are probably useful for ghc9 compatibility outside ghcide as well. Maybe they should be extracted?

<s>Currently, some of the `ghcide` tests fail for ghc-9.0.1 and I haven't fully figured out why yet.</s>


ghcide/src/Development/IDE/GHC/Warnings.hs
```diff
@@ -28,9 +29,16 @@ import           Language.LSP.Types                (type (|?) (..))
...
+#if __GLASGOW_HASKELL__ >= 900
+  let newAction :: DynFlags -> WarnReason -> Severity -> SrcSpan -> MsgDoc -> IO ()
+      newAction dynFlags wr _ loc msg = do
+        let prUnqual = alwaysQualify :: PrintUnqualified -- TODO: Do something proper here
+#else
   let newAction :: DynFlags -> WarnReason -> Severity -> SrcSpan -> PprStyle -> SDoc -> IO ()
       newAction dynFlags wr _ loc style msg = do
-        let wr_d = map ((wr,) . third3 (attachReason wr)) $ diagFromErrMsg diagSource dynFlags $ mkWarnMsg dynFlags loc (queryQual style) msg
+        let prUnqual = queryQual style
+#endif
```
anka-213, line 40: Log actions no longer gets a `PprStyle` in ghc9, so I'm not sure what the proper thing to do here is. Especially if I am to avoid using `CPP`.

jneira:
Still a WIP but just in case but it seems default project formatting style is being changed: see https://github.com/haskell/haskell-language-server/blob/master/CONTRIBUTING.md

anka-213:
There are still a few test cases that fail on ghc-9.0.1 for me, most notably the benchmark tests. I don't really understand how they work.

I'm also considering adding ghc-9.0.1 to the CI, but in that case, we would only want to test ghcide, since that is the only thing that is currently supported.

anka-213:
@jneira This is ghcide and not just hiecompat, so it's probably still relevant.

anka-213:
Does the CI machines have [too little disk space](https://app.circleci.com/pipelines/github/haskell/haskell-language-server/2599/workflows/db14939e-bca0-483c-9db0-10c34ac1df03/jobs/21528/parallel-runs/0/steps/0-109)?

```
> /tmp/ghc5331_0/ghc_48.s: hPutBuf: resource exhausted (No space left on device)
```

jneira:
> There are still a few test cases that fail on ghc-9.0.1 for me, most notably the benchmark tests. I don't really understand how they work.

Sure @pepeiborra could help us in that side

> I'm also considering adding ghc-9.0.1 to the CI, but in that case, we would only want to test ghcide, since that is the only thing that is currently supported.

github job steps can be skipped adding conditions to `if:` field, using the ghc version

pepeiborra:
> > There are still a few test cases that fail on ghc-9.0.1 for me, most notably the benchmark tests. I don't really understand how they work.
> 
> Sure @pepeiborra could help us in that side
> 
> > I'm also considering adding ghc-9.0.1 to the CI, but in that case, we would only want to test ghcide, since that is the only thing that is currently supported.
> 
> github job steps can be skipped adding conditions to `if:` field, using the ghc version

Ignore local failures in benchmark tests - they can time out fairly easily. As long as they work in CI...

anka-213:
@pepeiborra It seemed like they didn't time out, but rather didn't run at all. Every single benchmark test failed in less than 10 seconds with `did not successfully complete 5 repetitions`. It took longer time and more succeeded on older ghc versions. But maybe I'm misinterpreting the error message?

pepeiborra:
Run them with LSP_TEST_LOG_MESSAGES=1 LSP_TEST_LOG_STDERR=1 to find out what's going on

anka-213:
The [CI failure](https://app.circleci.com/pipelines/github/haskell/haskell-language-server/2612/workflows/2fa547fb-c0bb-453a-9248-413430d3ee2a/jobs/21650) `Package index change detected, that's pretty unusual` does indeed seem unusual. The only mention I can find of it on the internet is from the source code of stack:
https://github.com/commercialhaskell/pantry/blob/0a04a97561b3f292e66b285b41f75f082294ac67/src/Pantry/Hackage.hs#L242

anka-213:
@pepeiborra Ah, I see, thanks. The problem was that the benchmark suite is trying to build `Cabal-3.2.0.0`, which doesn't support `base-4.15`.

pepeiborra:
You can change that in ghcide/bench/config.yaml

anka-213:
The tests `reloading-th-test-unboxed` and `th-linking-test-unboxed` are unreliable for me. They randomly fail or succeed depending on (I assume) timing.
Sometimes it gets back an empty list of diagnostics first and then the real list, sometimes it gets the correct diagnostics right away.

anka-213:
How do I make the github actions CI run again? Do I need to resolve the merge conflicts first? Currently only the Circle CI seems to run.

berberman:
Yeah, you need resolve conflicts to trigger github actions.

anka-213:
I did resolve the merge conflicts, but it didn't seem to help. (There are new merge conflicts now, but those happened after the CI)

pepeiborra:
@anka-213 are you still working on this, and is there anything we can do to unblock you?

anka-213:
@pepeiborra I have not had much time to work on this lately. The biggest problem I encountered last I worked on this (which shouldn't really matter though) is that it tries to build all the dependencies for everything, even if I disabled the corresponding plugins through flags. My workaround locally has been to comment out the packages where dependencies doesn't build in `cabal.project`, but I can't do that if I want it to keep building the plugins on older versions of GHC.

Current status is that `ghcide` and a few plugins works, but none of the formatters, since none of them supported ghc-9 when I last worked on this.

anka-213:
There are also a lot of merge conflicts that has accumulated over time, which means it takes quite a bit of time just to get back to where I was before, even before doing new stuff.

I guess I could comment out the CI for ghc-9.0 so this can be merged, and keep working on making it better in a new PR, to avoid all the bitrot of merge conflicts.

pepeiborra:
> I guess I could comment out the CI for ghc-9.0 so this can be merged, and keep working on making it better in a new PR, to avoid all the bitrot of merge conflicts.

I think this is an excellent idea, please do as soon as you can!

anka-213:
I'm not looking forward to resolving all these merge conflicts :/

Perhaps someone can start making a review of the changes so I can know if there is something major that needs to be fixed (other than the merge conflicts) before this can be merged?

wz1000:
@anka-213 would it be possible to separate out the hie-compat changes into another PR? Then I can make a new release and also update hiedb.

mpickering:
I think we should use some of the open collective money to push this PR over the line? Would you be willing to do that @anka-213?

anka-213:
@wz1000 I've reopened and rebased that pull request now!
https://github.com/haskell/haskell-language-server/pull/1635

anka-213:
Looks like the merge conflicts weren't as bad as I thought. Should I try to split this in more pull requests, in addition to splitting out ` hie-compat`?

wz1000:
> Looks like the merge conflicts weren't as bad as I thought. Should I try to split this in more pull requests, in addition to splitting out hie-compat?

That could be a good idea. Makes reviewing easier.

anka-213:
Can someone re-enable CI as well?

anka-213:
> >  Should I try to split this in more pull requests?
> That could be a good idea. Makes reviewing easier.

Is there a way for pull requests to depend on each other, so the later pull requests won't contain a copy of all the earlier pull requests?

wz1000:
> > > Should I try to split this in more pull requests?
> > > That could be a good idea. Makes reviewing easier.
> 
> Is there a way for pull requests to depend on each other, so the later pull requests won't contain a copy of all the earlier pull requests?

I don't think so, you can mention the dependency in the description and we can take care to merge them in order.

anka-213:
I've pushed a build that should hopefully (successfully) run the tests on ghc9 in CI now.

I'm not sure how to best split this up. I guess I can extract most of the changes to extensions in a separate pull request.
Maybe I can extract the changes to `Development.IDE.GHC.Compat` and the changes that are strictly necessary for it to build on older ghc versions with those changes, but most of those changes doesn't make much sense without the context of ghc9 and the places where they are used.
I could separate out the changes to CI, stack files, cabal.project etc, but that prevents running the CI to check that it works with ghc9.

wz1000:
There is no need to split it up if you don't think it would be helpful.

anka-213:
It would probably be helpful to split it up, it's just that it's difficult to do it, especially doing it well.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -527,7 +526,10 @@ cradleToOptsAndLibDir cradle file = do
 emptyHscEnv :: IORef NameCache -> FilePath -> IO HscEnv
 emptyHscEnv nc libDir = do
     env <- runGhc (Just libDir) getSession
+#if !MIN_VERSION_ghc(9,0,0)
+    -- This causes ghc9 to crash
```
wz1000, line 530: Are you calling `initLoaderState? Is there a GHC bug report?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -463,7 +477,15 @@ generateHieAsts hscEnv tcm =
     -- don't export an interface which allows for additional information to be added to hie files.
     let fake_splice_binds = listToBag (map (mkVarBind unitDataConId) (spliceExpresions $ tmrTopLevelSplices tcm))
         real_binds = tcg_binds $ tmrTypechecked tcm
+#if MIN_VERSION_ghc(9,0,1)
+        -- TODO: Use some proper values here!
+        evBinds = emptyBag @EvBind :: Bag EvBind
```
wz1000, line 482: Please fix this, all the values come from the typechecked source:

```haskell
      top_ev_binds = tcg_ev_binds ts
      insts = tcg_insts ts
      tcs = tcg_tcs ts


```

wz1000:
Very nice work, just a few small details to take care of.

ghcide/session-loader/Development/IDE/Session.hs
anka-213: No, there is no GHC bug report (that I know of). I don't really know what's wrong. But it seems to work when I don't call it.

anka-213:
Oh, I realized that I forgot to enable the (reduced) test-suite for ghc9

ghcide/src/Development/IDE/Core/Compile.hs
anka-213: Thanks!

ghcide/session-loader/Development/IDE/Session.hs
anka-213: Can you explain what that function does and why we are calling it?

If I remove the conditional so it is called on ghc9 as well, I get this error:
```
 ghcide: Couldn't find a target code interpreter. Try with -fexternal-interpreter
 ```
 
 It seems like the `hsc_env` variable was not set in the `HscEnv` we get from `getSession`. I don't really know why or really what any of this means.
 
 I'm guessing that this issue was introduced by this commit: https://gitlab.haskell.org/ghc/ghc/commit/18757cab04c5c5c48eaceea19469d4811c5d0371
 
 For some reason, `hsc_interp` is only set when you call `setSessionDynFlags`, despite that seeming like a relatively unrelated function.

Comment (anka-213) [reply ghcide/session-loader/Development/IDE/Session.hs:530]: Maybe I should do this instead, so I can call `initDynLinker`?
```
    env <- runGhc (Just libDir) $ getSessionDynFlags >>= setSessionDynFlags >> getSession
```
The tests seem to succeed both with that change as well as when not calling `initDynLinker` at all.

File: cabal.project [hunk 4]
```diff
@@ -28,20 +28,129 @@ package *
...
+  -- https://github.com/mpickering/hie-bios/pull/285
+
+source-repository-package
+  type: git
+  location: https://github.com/hsyl20/ghc-api-compat
+  tag: 6178d75772c7d923918dfffa0b1f503dfb36d0a6
+
+source-repository-package
+  type: git
+  location: https://github.com/anka-213/th-extras
```
pepeiborra, line 58: are all these `source-repository-package` entries needed here, or can they be moved to the `cabal-ghc901.project` file?

ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -67,12 +67,11 @@ import           Lexer
...
-                                                 mkUserStyle, neverQualify, ppr,
-                                                 renderWithStyle,
+                                                 neverQualify, ppr,
                                                  showSDocUnsafe)
-import           PackageConfig                  (PackageConfig)
-import           Packages                       (getPackageConfigMap,
-                                                 lookupPackage')
+-- import           PackageConfig                  (PackageConfig)
+-- import           Packages                       ( -- getPackageConfigMap,
+--                                                  lookupPackage')
```
pepeiborra, line 74:
```suggestion
```

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -266,7 +266,8 @@ mkNameCompItem doc thingParent origName origMod thingType isInfix docs !imp = CI
             let (args, ret) = splitFunTys t
               in if isForAllTy ret
                   then getArgs ret
-                  else Prelude.filter (not . isDictTy) args
+                  -- TODO: Do we want to use multiplicity here?
+                  else Prelude.filter (not . isDictTy) $ map scaledThing args
```
pepeiborra, line 270: This code is only concerned with extracting argument names, so I don't see how multiplicity would be relevant here

ghcide/src/Development/IDE/Spans/Documentation.hs
```diff
@@ -36,7 +36,7 @@ import           HscTypes                       (HscEnv (hsc_dflags))
 import           Language.LSP.Types             (filePathToUri, getUri)
 import           Name
 import           NameEnv
-import           Packages
+-- import           Packages
```
pepeiborra, line 39:
```suggestion
```

ghcide/src/Development/IDE/Types/Options.hs
```diff
@@ -22,11 +22,12 @@ module Development.IDE.Types.Options
...
 import           Development.IDE.Core.RuleTypes
+import           Development.IDE.GHC.Compat        as GHC
 import           Development.IDE.Graph
 import           Development.IDE.Types.Diagnostics
-import           GHC                               hiding (parseModule,
-                                                    typecheckModule)
-import           GhcPlugins                        as GHC hiding (fst3, (<>))
+-- import           GHC                               hiding (parseModule,
+--                                                     typecheckModule)
+-- import           GhcPlugins                        as GHC hiding (fst3, (<>))
```
pepeiborra, line 30:
```suggestion
```

ghcide/test/exe/Main.hs
```diff
@@ -538,12 +538,17 @@ diagnosticTests = testGroup "diagnostics"
             , "foo = 1 {-|-}"
             ]
       _ <- createDoc "Foo.hs" "haskell" fooContent
+#if MIN_VERSION_ghc(9,0,1)
+      -- Haddock parse errors are ignored on ghc-9.0.1
```
pepeiborra, line 542: This means that we no longer need to parse twice (todo in a separate PR).

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -24,6 +24,12 @@ flag ghc-lib
   description:
     Force dependency on ghc-lib-parser even if GHC API in the ghc package is supported
 
+flag hlint33
+  default:     True
+  manual:      False
+  description:
+    Hlint-3.3 doesn't support older ghc/ghc-lib versions, so we can use hlint-3.2 for backwards compat
+
```
pepeiborra, line 32: I presume this is required in order to support hlint on ghc. 
@jneira what are the use cases for that, can we drop it?

jneira: Yeah hlint-3.3+ghc-lib-9.0 should work for all ghc versions (like hlint<3.2+ghc-lib-8.1.0 worked for 8.8 and 8.6)

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -7,10 +7,17 @@
...
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE TypeFamilies          #-}
 {-# OPTIONS_GHC -Wno-orphans   #-}
 
+#ifdef HLINT_ON_GHC_LIB
+#define MIN_GHC_API_VERSION(x,y,z) MIN_VERSION_ghc_lib(x,y,z)
+#else
+#define MIN_GHC_API_VERSION(x,y,z) MIN_VERSION_ghc(x,y,z)
+#endif
+
```
pepeiborra, line 20: Is it possible to drop this and simply assume that hlint is always on ghc-lib?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
anka-213: Ok, that's good. I wasn't really familiar with what the code is doing.

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
anka-213: The reason I added this flag is that none of the formatters support ghc-lib-9.0 yet, so we would get conflicts if we would try using hlint-3.3 together with any of the formatting plugins. As soon as they support ghc9, this flag can be removed.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
anka-213: Probably. I just followed the pattern from other parts of the code.

cabal.project
anka-213: Most or all of them can probably be removed. They are a remnant from before I separated the `cabal.project` file for ghc9.

ghcide/src/Development/IDE/GHC/Util.hs
anka-213: I think the reason I left them commented out was that they were (incidentally) re-exported from the `.Compat` module and I was a bit unsure if that made it a bit more opaque and if I should import it qualified instead. But yes, I should just remove the commented out code.

ghcide/session-loader/Development/IDE/Session.hs
anka-213: That was apparently premature, some tests fail (e.g.  the `ghcide.cradle.multi` tests) with the
```
env <- runGhc (Just libDir) $ getSessionDynFlags >>= setSessionDynFlags >> getSession
```
version, but all tests succeed when I simply don't run `initLoaderState` at all.

What is the harm of not running it/what is it needed for?

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
jneira: just in case hlint is on ghc-lib except for the last ghc version (9.0 with this pr)

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
jneira: thanks, could we add the note about why it is needed and when we could remove it in the description?

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -54,14 +60,28 @@ library
...
     , unordered-containers
 
-  if ((!flag(ghc-lib) && impl(ghc >=8.10.1)) && impl(ghc <9.0.0))
-    build-depends: ghc ^>=8.10
+  if (flag(hlint33))
+    build-depends: hlint ^>=3.3
+  else
+    build-depends: hlint ^>=3.2
+
+  if (!flag(ghc-lib) && (!flag(hlint33) && impl(ghc >=8.10.1) || flag(hlint33) && impl(ghc >=9.0.1)) && impl(ghc <9.1.0))
```
jneira, line 68: this condition is quite hard to follow, i wonder if it could be simplified

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
jneira: Out of curiosity, have you tried to use `hlint ^> 3.2 || hlint ^> 3.3` to see if cabal is able to choose the right version based in other constraints?

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
anka-213: I think some of the complexity is needed, since we need to do different things depending on what libraries hlint is using. This depends on both the version of hlint and the version of ghc and what flags hlint has set.

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
anka-213: Probably. This was just the first/most direct way I could come up with that worked.
I wanted to try to mirror the behaviour of hlint as closely as possible, so it's based on hlint's cabal file, but for two different versions of hlint at the same time: https://github.com/ndmitchell/hlint/commit/e2c0833c4a337f78f2cee58ef59d318d26f1666b#diff-3ab702ffba9192c29e2d49f82e65c36af604ea811daed8644bf3037fb6f4fc11.

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
anka-213: There's also the problem that hlint has dropped support for ghc-8.6:
https://github.com/ndmitchell/hlint/commit/4df9a6fdaca867d82e46e57d5ddce1a577447d86#diff-3ab702ffba9192c29e2d49f82e65c36af604ea811daed8644bf3037fb6f4fc11

anka-213: If we could find a good way to set the flag:
```
#ifdef HLINT_ON_GHC_LIB
```
we could probably remove most/all of this complexity.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
anka-213: I guess we could force hlint to always use ghc-lib, but that might be a bit redundant when the ghc-lib version is the same as the ghc version.

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
anka-213: How about this version? It's a lot more direct and mirrors the intent of supporting two hlint versions more clearly, but has (very slightly) more code duplication
```
  if (flag(hlint33))
    build-depends: hlint ^>=3.3
    if (!flag(ghc-lib) && impl(ghc >=9.0.1) && impl(ghc <9.1.0))
      build-depends: ghc ==9.0.*

    else
        build-depends:
          , ghc
          , ghc-lib            == 9.0.*
          , ghc-lib-parser-ex  == 9.0.*

      cpp-options:   -DHLINT_ON_GHC_LIB

  else
    build-depends: hlint ^>=3.2
    if (!flag(ghc-lib) && impl(ghc >=8.10.1) && impl(ghc < 8.11.0))
        build-depends: ghc >=8.10 && < 9.0

    else
        build-depends:
          , ghc
          , ghc-lib            ^>= 8.10.4.20210206
          , ghc-lib-parser-ex  ^>= 8.10

      cpp-options:   -DHLINT_ON_GHC_LIB
```

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
pepeiborra: I would prioritise the ease of maintaining here and drop support for hlint-on-ghc

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
jneira: > How about this version? It's a lot more direct and mirrors the intent of supporting two hlint versions more clearly, but has (very slightly) more code duplication

I like it, it is more readable so the slight code duplication worths it imo

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
jneira: The reason to have two code paths (iirc from a discussion with @ndmitchell i dont find right now) is we have three components insteracting:
* ghcide: uses raw ghc
* hlint: by default uses raw ghc for last version and ghc-lib for previous ones
* apply-refact: uses raw ghcs

So we have a code path for the last ghc version which can get the parsed module from ghcide and use it directly for hlint and apply-refact. It is more precise and performant.
For ghc-lib we have to parse the file manually to give it to hlint and then write to a temp dir on disk the contents to call to apply-refact, which parses the file again, etc.

jneira: The  discussion was here:
https://github.com/haskell/haskell-language-server/issues/32#issuecomment-637458673
and
https://github.com/haskell/haskell-language-server/issues/32#issuecomment-637587961

pepeiborra: Thanks for the clear explanation. I still think that supporting hlint-on-ghc is suicidal, given that hlint only supports 1 ghc version whereas we need to support a much wider range. Maintainability trumps performance. But this is just my opinion

cabal.project
pepeiborra: Please remove before merging

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
jneira: Not only performance, quoting the mentioned discussion:

> Better to share the AST, and likely to reduce the potential bugs (as things like the extensions will be precisely known), but otherwise it degrades reasonably.

But maybe we could reconsider the decision @ndmitchell?

jneira: otoh once we have full ghc 9 support and drop ggc 8.6 we will only use one version of hlint, 3.3

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
anka-213: Is the new description here clear enough?

ghcide/test/exe/Main.hs
anka-213: Are the hlint related problems mentioned here resolved as well? https://github.com/haskell/haskell-language-server/blob/e0a464282db913b6c7e0f2ec4504d24e2bcfb4b1/ghcide/src/Development/IDE/Core/Rules.hs#L198-L206

pepeiborra: Ah, good question! I think it might, but I'm not sure. We can test it separately

cabal.project
pepeiborra: About these package overrides: does the 8.10 build of ghcide work without them? 

If the answer to 1 is Yes, then I think we should just remove them. 
If not, then I think we need to look into adding more `MIN_PACKAGE_VERSION` pragmas until the answer is Yes. I'm happy to help with that if you give me access to your fork.

anka-213: I think so. I'll try removing them and check that if still builds.

anka-213: In fact, that's already the case for the stack builds, so removing them should be perfectly safe. I just didn't know how to make a separate `cabal.project` file for ghc9. (We'll probably want to merge them together again as soon as all the patches have been merged upstream, so the overrides are no longer needed)

ghcide/session-loader/Development/IDE/Session.hs
pepeiborra: Ping @wz1000

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
pepeiborra: Looks good to me

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
pepeiborra: But we will need to maintain this hlint duality for the foreseeable future, which is not free. Anyways, we can always remove it later, resolving this.

anka-213:
I may have accidentally created a bunch of duplicate jobs in the github CI by pushing many times in quick succession.  We can probably cancel the older ones so the newest ones get in front of the queue more quickly.

berberman:
> We can probably cancel the older ones so the newest ones get in front of the queue more quickly.

I've cancelled them

anka-213:
I think the build failed on ghc9 because I rebased to master on the PR. Let me update the reference

anka-213:
Future work: (Could be included in this pr though)

Update the nix files to use the patched versions of all the upstream packages when building on ghc9.

The sad thing is that now we would have three copies of all those upstream patches, one for each method for building hls.

anka-213:
Oh, wait, why did stack successfully build it with the old commit hash for lsp, while cabal failed? It must be in a cache, so it didn't notice that the commit had disappeared.

anka-213:
Oh, no! Test failures again. I thought I have fixed those, but maybe there's a regression after upgrading a package or something.

anka-213:
Here are the four that are currently consistently failing on my local machine:
```
haskell-language-server
  code actions
    hlint suggestions
      hlint diagnostics ignore hints honouring .hlint.yaml: FAIL (3.59s)
        src/Test/Hls/Util.hs:389:
        Got unexpected diagnostics for Uri {getUri = "file:///Users/anka/projekt/not-mine/haskell/haskell-language-server/test/testdata/hlint/ignore/ApplyRefact.hs"} got [Diagnostic {_range = Range {_start = Position {_line = 2, _character = 4}, _end = Position {_line = 2, _character = 7}}, _severity = Just DsInfo, _code = Just (InR "refact:Redundant bracket"), _source = Just "hlint", _message = "Redundant bracket\nFound:\n  (1)\nWhy not:\n  1\n", _tags = Nothing, _relatedInformation = Nothing},Diagnostic {_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 22}}, _severity = Just DsInfo, _code = Just (InR "Use camelCase"), _source = Just "hlint", _message = "Use camelCase\nFound:\n  camel_case = ...\nWhy not:\n  camelCase = ...\n", _tags = Nothing, _relatedInformation = Nothing}]
    redundant import code actions
      remove solitary redundant imports:                    FAIL
        Last message received:
        {
            "method": "$/progress",
            "params": {
                "token": "5",
                "value": {
                    "kind": "end"
                }
            },
            "jsonrpc": "2.0"
        }
      doesn't touch other imports:                          FAIL
        Last message received:
        {
            "method": "textDocument/publishDiagnostics",
            "params": {
                "diagnostics": [
                    {
                        "range": {
                            "end": {
                                "character": 0,
                                "line": 1
                            },
                            "start": {
                                "character": 0,
                                "line": 0
                            }
                        },
                        "message": "target â is not a module name or a source file",
                        "severity": 1,
                        "source": "compiler"
                    }
                ],
                "uri": "file:///Users/anka/projekt/not-mine/haskell/haskell-language-server/test/testdata/redundantImportTest/src/MultipleImports.hs"
            },
            "jsonrpc": "2.0"
        }
  window/workDoneProgress
    sends indefinite progress notifications:                FAIL
      Last message received:
      {
          "method": "$/progress",
          "params": {
              "token": "19",
              "value": {
                  "message": "Finished indexing 1 files",
                  "kind": "end"
              }
          },
          "jsonrpc": "2.0"
      }

4 out of 4 tests failed (61.11s)
```

It might be something silly like a currently disabled plugin being required for the test. I'll look a bit more at it.

anka-213:
Ok, more debugging: It looks like it's one of the upstream changes is what triggers these errors, since they also occur on ghc-8.10.4, even if I enable all plugins, when I run it with `--project-file=cabal-ghc9.project`

After testing some more: More specifically, it's with the newest version of `lsp` that it's broken. Removing just that patch and using the version from hackage fixes these four errors (on ghc-8.10.4). Not at all obvious why.

anka-213:
A git bisect points me to this PR: https://github.com/haskell/lsp/pull/326

This might also be an answer to the question posed in that PR:
> Why is lsp changing the current directory? I could not find any good reason.

pepeiborra:
> A git bisect points me to this PR: https://github.com/haskell/lsp/pull/326
> 
> 
> 
> This might also be an answer to the question posed in that PR:
> 
> > Why is lsp changing the current directory? I could not find any good reason.

Lsp changing the current directory was causing other problems and had to be removed. It's unfortunate that this is blocking the 9.0 upgrade, but it will need to be fixed on the hls plugins side. It should be a simple matter of making sure all relative paths to config files are made absolute using the workspace root

anka-213:
> > A git bisect points me to this PR: [haskell/lsp#326](https://github.com/haskell/lsp/pull/326)
> > This might also be an answer to the question posed in that PR:
> > > Why is lsp changing the current directory? I could not find any good reason.
> 
> Lsp changing the current directory was causing other problems and had to be removed. It's unfortunate that this is blocking the 9.0 upgrade, but it will need to be fixed on the hls plugins side. It should be a simple matter of making sure all relative paths to config files are made absolute using the workspace root

Ah, I see. It's not really blocking the 9.0 upgrade, it's blocking the `lsp` package upgrade in general, which the 9.0 upgrade happens to do for 9.0 specifically. The test failures appear on all ghc versions, not just ghc-9 if I use that `lsp` version. There are 3 paths forward I can think of:

1. Temporarily revert that change in the branch of `lsp` that 9.0 uses. This shouldn't cause any regressions, since hls currently still uses the old version without that patch
2. Ignore those test failures on ghc-9 and mark them as broken on ghc-9 for now. I don't know how serious those problems are.
3. Try to fix the problems in the plugins. I don't know how to do that or how to figure out what caused the failures.

I tried building `hls/master` with the current `lsp/master` to verify if these test failures are caused by a combination of my changes and that patch or if that patch alone is sufficient to cause these test failures. And it did give the same test failures.

anka-213:
I opted for option 1 for now, but if someone knows how to fix the test failures or judges that those test failures are a smaller problem than those caused by lsp changing the current directory, we can change that.

anka-213:
I've created issue #1895 for the problems related to haskell/lsp#326 and a draft PR #1894 for testing it.

pepeiborra (approved):
Good to merge!

PR #1650, opened 2021-04-02
Do not override custom commands

pepeiborra:
For users of ghcide as a library - right now it is impossible to register custom commands when using `defaultMain`


berberman (approved):
LGTM

PR #1651, opened 2021-04-02
log exceptions before killing the server

pepeiborra:
I believe that this will help understand the "resource vanished - broken pipe" test failures

### DIFF
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -141,9 +141,11 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             logInfo (ideLogger ide) $ T.pack $ "Registering ide configuration: " <> show initConfig
             registerIdeConfiguration (shakeExtras ide) initConfig
 
-            let handleServerException e = do
-                    logError (ideLogger ide) (T.pack $ "Unexpected exception in server thread: " <> show e)
+            let handleServerException (Left e) = do
+                    logError (ideLogger ide) $
+                        T.pack $ "Fatal error in server thread: " <> show e
                     exitClientMsg
+                handleServerException _ = pure ()
             _ <- flip forkFinally handleServerException $ runWithDb dbLoc $ \hiedb hieChan -> do
               putMVar dbMVar (hiedb,hieChan)
               forever $ do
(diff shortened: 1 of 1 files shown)


ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -141,7 +141,10 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             logInfo (ideLogger ide) $ T.pack $ "Registering ide configuration: " <> show initConfig
             registerIdeConfiguration (shakeExtras ide) initConfig
 
-            _ <- flip forkFinally (const exitClientMsg) $ runWithDb dbLoc $ \hiedb hieChan -> do
+            let handleHieDbException e = do
+                    logError (ideLogger ide) (T.pack $ "Unexpected exception in hiedb thread: " <> show e)
+                    exitClientMsg
+            _ <- flip forkFinally handleHieDbException $ runWithDb dbLoc $ \hiedb hieChan -> do
```
wz1000, line 147: This isn't the hiedb thread, but the server thread. The hiedb thread is forked again in `runWithDb` and shouldn't bring down the main server thread.

pepeiborra: Thanks, I didn't look closely enough. I have fixed the log message to reflect that this is the server thread

jneira (approved):
thanks, lets see what can it tell us

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -141,7 +141,10 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             logInfo (ideLogger ide) $ T.pack $ "Registering ide configuration: " <> show initConfig
             registerIdeConfiguration (shakeExtras ide) initConfig
 
-            _ <- flip forkFinally (const exitClientMsg) $ runWithDb dbLoc $ \hiedb hieChan -> do
+            let handleServerException e = do
+                    logError (ideLogger ide) (T.pack $ "Unexpected exception in server thread: " <> show e)
+                    exitClientMsg
+            _ <- flip forkFinally handleServerException $ runWithDb dbLoc $ \hiedb hieChan -> do
```
berberman, line 147: `handleServerException` always runs after the action. We should check `e` is `Left SomeException` or `Right ()`.

PR #1373, opened 2021-02-15
Add test for multi-component goto def and make runLanguageServer responsible for hiedb

wz1000:
Test for https://github.com/haskell/haskell-language-server/pull/1357

Involved re-jigging 'runWithDb' to be within the callback to 'runLanguageServer' so that the database location
is derived from the 'rootPath' of the LSP session, and not the starting CWD of the ghcide process.

I have verified that the test fails without the mentioned commit.

PR #1655, opened 2021-04-02
Shut the Shake session on exit, instead of restarting it

pepeiborra:
Restarting the session will result in progress reporting and other messages
being sent to the client, which might have already closed the stream

Result of @berberman investigation in #1628

### DIFF
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -188,7 +188,7 @@ exitHandler :: IO () -> LSP.Handlers (ServerM c)
 exitHandler exit = LSP.notificationHandler SExit $ const $ do
     (_, ide) <- ask
     -- flush out the Shake session to record a Shake profile if applicable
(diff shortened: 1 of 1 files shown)


isovector:
Does this fix the mysterious `resource vanished` errors that happen when running tests in parallel?

pepeiborra:
> Does this fix the mysterious `resource vanished` errors that happen when running tests in parallel? <3

Doesn't seem to fix that

pepeiborra:
> > Does this fix the mysterious `resource vanished` errors that happen when running tests in parallel? <3
> 
> Doesn't seem to fix that

I just gave up and bisected it to 132193e1 (https://github.com/haskell/haskell-language-server/pull/1373). It must be an issue with opening the mysql db before it has been released or similar. /cc @wz1000

jneira (approved):
LGTM, the change makes totally sense, not sure why `restartShakeSession` was even used at first.

wz1000:
I will try to take a look when I find the time.

berberman:
Hmm, I cherry-picked this to #1628, but test didn't show the fix

PR #1656, opened 2021-04-02
Fix ignore paths

jneira:
* Attempt to fix properly the combination of ignore paths and required build checks
* Unfortunately the unique solution i have found is make all steps of the build dedpend on the `skip-duplicate-action` outcome
* There is a feature request to make jobs exit early with a specific  status: https://github.com/actions/runner/issues/662 but is is not being implemented

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,10 +21,7 @@ jobs:
           cancel_others: true
           paths_ignore: '["**/docs/**", "**.md", "**/LICENSE", ".circleci/**", "install/**", "nix/**"]'
 
-
   test:
-    needs: pre_job
-    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -58,33 +55,34 @@ jobs:
           #   ghc: '8.6.4'
 
     steps:
-      # Cancel queued workflows from earlier commits in this branch
-      - uses: fkirc/skip-duplicate-actions@master
```
Ailrun, line 62: I think we still need this step to skip unnecessary jobs started by outdated commits

jneira: mmm does not the same action in the `pre_job` do all the needed work? Afaiu it does all the things and no depends on the params: https://github.com/fkirc/skip-duplicate-actions#usage-examples

jneira:
Marked as a draft as i want to get a pr in my repo cancelled properly before merge this

.github/workflows/test.yml
jneira: The log of a build in my branch can provide some evidence:

```
Did not find other workflow-runs to be cancelled
Stop backtracking at commit https://github.com/jneira/haskell-language-server/commit/3eddabf804b4a09ea07cff26f5470514b1639f50 because '.github/workflows/test.yml' are not skippable against paths '' or paths_ignore '**/docs/**,**.md,**/LICENSE,.circleci/**,install/**,nix/**'
Do not skip execution because we did not find a transferable run
```

https://github.com/jneira/haskell-language-server/pull/22/checks?check_run_id=2256857741

Ailrun: Oh, does it work per workflow? I thought it's per job...

jneira:
I think it had worked here: https://github.com/jneira/haskell-language-server/pull/23 so we could merge it
//cc @Ailrun

Ailrun (approved):

PR #1657, opened 2021-04-03
Fix a wingman bug caused by mismanaged stale data

### LINKED ISSUE #1602: Wingman gets slow on big modules
Even for things like "refine hole" which should be instant. My guess is that waiting for a fresh type-checking result is to blame. We could probably get away with using stale data instead.

isovector:
Wingman uses a lot of potentially stale rules, and each comes with its own position mapping. But I was stupid when I first wrote this, and didn't realize just because there results were stale didn't mean they refer to the same document.

This PR adds some type-level information, allowing me to track the age of results from Shake. By throwing an untouchable variable inside of the stale tag, the compiler makes sure I haven't accidentally scrambled ranges between different versions of the document.


jneira (approved):

pepeiborra:
This looks like it could be generally useful. @isovector might be worth putting the tracked `use` combinators inside ghcide, in a module `Development.IDE.Core.UseStale` or similar?

isovector:
@pepeiborra . Done. PTAL

PR #1661, opened 2021-04-03
Avoid creating IsFileOfInterest keys for non workspace files

pepeiborra:
There's no need to create `IsFileOfInterest` keys for every interface file


ndmitchell (commented):
In VS Code you can open individual files that aren't in the workspace. If you have goto definition, then often that is super easy. What happens then?

pepeiborra:
> In VS Code you can open individual files that aren't in the workspace. If you have goto definition, then often that is super easy. What happens then?

* The file is not in the workspace, so it doesn't get an `IsFileOfInterest` dependency
* The file will be VFS, so it gets an `alwaysRerun` dependency

Now the `GetModificationTime` for this file has an `alwaysRerun` dependency forever, i.e. it will never transition to the non `alwaysRerun` state. This is exactly what we want for non-workspace files, since they are not watched. So all is good.

ndmitchell (commented):
So it sounds like your argument is if we hit the alwaysRerun path than isFileOfInterest is not interesting. If so, why both with the isWorkspaceFile check? Why not move it down to the branches where alwaysRerun isn't hit, and save ourself the dependency in a more?

I note there are now two local definitions for `isWF` being whether the file is a workspace file, or a watched file. Would be good to use distinct names.

pepeiborra:
> So it sounds like your argument is if we hit the alwaysRerun path than isFileOfInterest is not interesting. If so, why both with the isWorkspaceFile check? Why not move it down to the branches where alwaysRerun isn't hit, and save ourself the dependency in a more?
 
Right, that's a good point. I'll move the `IsFileOfInterest` dependency down to the branch and remove the `isWorkspaceFile` check.

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -127,16 +128,20 @@ getModificationTimeImpl vfs isWatched missingFileDiags file = do
...
-        -- dependency explicit by using the IsFileOfInterest rule
-        _ <- use_ IsFileOfInterest file
         case mbVirtual of
             Just (virtualFileVersion -> ver) -> do
                 alwaysRerun
                 pure (Just $ LBS.toStrict $ B.encode ver, ([], Just $ VFSVersion ver))
             Nothing -> do
                 isWF <- isWatched file
-                unless (isWF || isInterface file) alwaysRerun
+                unless (isWF || isInterface file)
```
ndmitchell, line 137: I find this logical really confusing. I think it's equivalent to:

```
if isInterfaceFile then do
    return () -- nothing to do
else if isWF then do
    alwaysRerun
else
    use_ IsFileOfInterest file
```

But I don't guarantee it - there's way too much negation, negation of negation etc.  Could we use that structure, and explain the three branches separately?

pepeiborra: I've done that, and it's much better now. Thanks!

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -127,16 +128,26 @@ getModificationTimeImpl vfs isWatched missingFileDiags file = do
...
+                        void (use_ IsFileOfInterest file)
+                    else if isInterface file
+                        then -- interface files are tracked specially using the closed world assumption
+                            pure ()
+                        else -- in all other cases we will need to freshly check the file system
+                            alwaysRerun
+
+                when (isWF && not (isInterface file)) $
+                    -- we use 'getVirtualFile' to discriminate FOIs so make that dependency explicit
+                    void $ use_ IsFileOfInterest file
```
ndmitchell, line 150: These two lines look redundant now?

pepeiborra: They are, forgot to remove them. Gone now

ndmitchell (approved):
Thanks - this code is now beautifully understandable - I learnt something about the old code from your rewrite :)

PR #1664, opened 2021-04-04
Update shake bounds of install script

berberman:


install/hls-install.cabal
```diff
@@ -19,7 +19,7 @@ library
                      , Env
                      , Help
   build-depends:       base >= 4.9 && < 5
-                     , shake >= 0.16.4 && < 0.19
+                     , shake >=0.17.5
```
ndmitchell, line 22:
```suggestion
                     , shake >= 0.17.5
```

We usually have a space after `>=`

ndmitchell (approved):

install/hls-install.cabal
berberman: Okay... but it seems that `cabal-fmt` will remove this space

ndmitchell: Meh, if there's a formatter removing it, don't worry. Although it feels awfully inconsistent that it removes the space here and adds it on the line above. This is a very light suggested change, and if it's not the right thing, don't bother.

berberman: Ah, I see. It's inconsistent because I copied  `shake >=0.17.5` from another cabal file in our codebase, whereas this cabal file haven't gotten formatted yet.

PR #1671, opened 2021-04-05
Intelligent derivations of Semigroup and Monoid for Wingman

isovector:
This PR gives Wingman new special "known tactics" for synthesizing `Semigroup` and `Monoid` instances. They do exactly what you'd expect --- passing the buck to the underlying instances for every value inside. Special to this tactic, Wingman will leave holes for any argument that it can't work out.

Behind the scenes, there's quite a lot going on here. Lots of machinery needed to be moved around to get access to the instance environment, and to make sure we can find the right things in scope. Nothing too interesting, but it is a little janky --- I had to rewrite a chunk of the instance lookup machinery (see `Wingman.Context`).


isovector:
Looks like the HsExpr pretty printer changed in 8.10, which makes one of the tests due to different spacing. Grrrrr

jneira (approved):
LGTM

PR #1672, opened 2021-04-06
Ignore nix job steps by path

jneira:
* Follow up of #1656
* The nix job is required by branch protection rules so the change is needed


berberman (approved):
LGTM

PR #1673, opened 2021-04-06
Ignore ghcide tests by paths

jneira:
* If we only change ghcide downstream packages we have not test ghcide itself
* Leveraging existing `skip-duplicate-actions` machinery


Ailrun (approved):
LGTM

PR #1675, opened 2021-04-06
Wingman: Use infix notation for operator applications

isovector:
#1671 makes it clear that using prefix notation for operators is a horrendous atrocity. This PR makes Wingman generate infix notation whenever applying the first two arguments of an operator.


isovector:
User error. I forgot to update my featureset.

jneira:
Lets see if removing ghcide tests speed up ci

plugins/hls-tactics-plugin/test/golden/AutoInfixApplyMany.hs.expected
```diff
@@ -0,0 +1,3 @@
+test :: (a -> b -> x -> c) -> a -> (a -> b) -> x -> c
+test (/:) a f x = (a /: f a) x
+
```
Ailrun, line 3: Just an idea: it's maybe good to add a nested operators case, like
```Haskell
test :: (a -> b -> c) -> (c -> d -> e) -> a -> (a -> b) -> d -> e
test (/:) (//:) a f x = (a /: f a) //: d
```

plugins/hls-tactics-plugin/src/Wingman/CodeGen.hs
```diff
@@ -203,3 +206,12 @@ buildDataCon should_blacklist jdg dc tyapps = do
     & #syn_trace %~ rose (show dc) . pure
     & #syn_val   %~ mkCon dc tyapps
 
+
+------------------------------------------------------------------------------
+-- | Make a function application, correctly handling the infix case.
+mkApply :: OccName -> [HsExpr GhcPs] -> LHsExpr GhcPs
+mkApply occ (lhs : rhs : more)
+  | all (liftA2 (||) isSymbol isPunctuation) (occNameString occ)
```
wz1000, line 214: Use `isSymOcc`

wz1000 (approved):

PR #1685, opened 2021-04-07
Regenerate ModDetails from iface instead of using result of TidyProgram

wz1000:
@mpickering claims the details from `tidyProgram` leak memory, see https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5478

Let's see if we can detect an effect in the `lsp` benchmarks

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/1685"></a>


mpickering:
I will try this locally.

jneira:
did we get to examine the differences in benchmarks?

PR #1688, opened 2021-04-07
Decrease Wingman timeout from 3.3 minutes to 2 seconds (configurable)

isovector:
At some point I made a boneheaded move, thinking that `1e8` was the number of microseconds in a second. I'm not sure where this number came from, but it's 2 orders of magnitude too large. As a result, the timeout feature in Wingman didn't work very well.

This PR changes the tactic timeout to be a more reasonable 2 seconds, down from 3+ minutes.

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/Judgements/Theta.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/Judgements/Theta.hs
@@ -127,10 +127,23 @@ excludeForbiddenMethods = filter (not . flip S.member forbiddenMethods . hi_name
       [ -- monadfail
         "fail"
         -- show
-      , "showsPrec"
-      , "showList"
+      , "showsPrec", "showList"
+        -- functor
+      , "<$"
+        -- applicative
+      , "liftA2", "<*", "*>"
         -- monad
-      , "return"
+      , "return", ">>"
+        -- alternative
+      , "some", "many"
+        -- foldable
+      , "foldr1", "foldl1", "elem", "maximum", "minimum", "sum", "product"
+        -- traversable
+      , "sequenceA", "mapM", "sequence"
+        -- semigroup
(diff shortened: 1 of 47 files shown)


pepeiborra:
Why 2 seconds though. Seems a bit short. Make it configurable ?

isovector:
Now officially configurable!

plugins/hls-tactics-plugin/src/Wingman/Judgements/Theta.hs
```diff
@@ -127,10 +127,23 @@ excludeForbiddenMethods = filter (not . flip S.member forbiddenMethods . hi_name
...
+        -- alternative
+      , "some", "many"
+        -- foldable
+      , "foldr1", "foldl1", "elem", "maximum", "minimum", "sum", "product"
+        -- traversable
+      , "sequenceA", "mapM", "sequence"
+        -- semigroup
+      , "sconcat", "stimes"
+        -- monoid
+      , "mconcat"
```
Ailrun, line 146: Why are these changes bundled together?

Ailrun (approved):
LGTM

plugins/hls-tactics-plugin/src/Wingman/Judgements/Theta.hs
isovector: I'm an absolute madman

ndmitchell:
My rule is always make timeouts in seconds, and Double (or even better, the Seconds alias https://hackage.haskell.org/package/extra-1.7.9/docs/Extra.html#t:Seconds). Then also use the timeout from Extra (https://hackage.haskell.org/package/extra-1.7.9/docs/Extra.html#v:timeout) and everything works properly. Having timeouts in milliseconds, microseconds or whatever is just too confusing.

PR #1692, opened 2021-04-08
Fix ghcide and HLS enter lsp mode by default

berberman:
We can't use `switch` with alternative combinators, since it never fails.


pepeiborra (approved):
Good catch, thanks! I always make the same mistake with these combinators that never fail

PR #1694, opened 2021-04-08
fix fill hole lost expected backquote

### LINKED ISSUE #1690: Infix backticked holes get their backticks deleted once filled
### Steps to reproduce

Write this:

```
data A = A

foo :: A -> A -> A
foo A A = A

test :: A -> A -> A
test a1 a2 = a1 `_` a2
```

Open up the context menu on the hole and select "replace with `foo`".

### Expected behaviour

The implementation should now look like:

```
test :: A -> A -> A
test a1 a2 = a1 `foo` a2
```

### Actual behaviour

The implementation now looks like:

```
test :: A -> A -> A
test a1 a2 = a1 foo a2
```

July541:
Fix #1690.
I match the backquote as the solution, **but** I am not sure that is an appropriate way to update the range at this function.

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -678,15 +678,21 @@ suggestFillHole :: Diagnostic -> [(T.Text, TextEdit)]
 suggestFillHole Diagnostic{_range=_range,..}
     | Just holeName <- extractHoleName _message
     , (holeFits, refFits) <- processHoleSuggestions (T.lines _message)
-    = map (proposeHoleFit holeName False) holeFits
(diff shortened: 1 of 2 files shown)


OliverMadine (changes requested):
Hello,

This has an issue when the hole needs to be filled with an infix operator since in this case it should not have any backquotes. For example, the action "replace _ with (+)"  will fill the hole with \`(+)\` rather than +

July541:
> Hello,
> 
> This has an issue when the hole needs to be filled with an infix operator since in this case, it should not have any backquotes. For example, the action "replace _ with (+)" will fill the hole with `(+)` rather than +.

Thanks,
I made a silly mistake.

PR #1696, opened 2021-04-08
Fix: Infix holes no longer filled using postfix notation

OliverMadine:
fix: #1690

In follow up to my response to @July541's PR (#1694), I have written a different solution to cover the cases of filling holes with infix operators.

applying the "replace _ with (+)" code action to
```haskell
test :: Int -> Int -> Int
test a1 a2 = a1 `_` a2
```

now results in 

```haskell
test :: Int -> Int -> Int
test a1 a2 = a1 + a2
```

rather than 

```haskell
test :: Int -> Int -> Int
test a1 a2 = a1 `(+)` a2
```

Note: This solution depends on `processHoleSuggestions` returning infix functions in their postfix form (e.g. `(+)`) to determine if a function infix/postfix. I imagine there might be a nicer way of doing this.


isovector:
Sorry, merged into the wrong PR! My bad!

OliverMadine:
Closed due to GitHub issues (committed to branch master by mistake). Will reopen shortly

PR #1697, opened 2021-04-09
Wingman: maintain user-defined fixity for definitions

### LINKED ISSUE #1610: Wingman agda-split transforms infix into prefix
Trying to split `ma >>= f = _` results in matches of the form `(>>=) Foo f = _`

isovector:
This PR makes Wingman respect user-defined fixities when performing case splits. However, if the user didn't explicitly bind any arguments, it will choose to use infix for a symbolically-named functions.

Fixes #1610


isovector:
This is now fixed upstream. I plan to tackle it later tonight if time permits.

isovector:
@berberman nix builds seem to be failing --- any idea what's up?

berberman:
> @berberman nix builds seem to be failing --- any idea what's up?

Don't worry, `nix/build` jobs are broken because this PR introduces `ghc-source-gen-0.4.1.0` which nix doesn't know. The jobs are *not* in branch protection rules and can be fixed later (which only affect creating binaries for nix).

For people who are interested this, the fix would be just overriding `ghc-source-gen-0.4.1.0` here: https://github.com/haskell/haskell-language-server/blob/b8bb06eb1b117943f4436a6fdafe5c09e76cac1c/flake.nix#L71-L86

isovector:
Tests (minus nix) are green!

Ailrun (approved):
Thank you, LGTM.

isovector:
CI is hitting this https://github.com/actions/checkout/issues/23, which claims to be fixed but clearly isn't.

isovector:
@berberman , it appears CI for 9.0.1 is broken. The error I get is this:

```
Cloning into '/home/runner/work/haskell-language-server/haskell-language-server/dist-newstyle/src/operation_-2cf41f8926150174'...
fatal: reference is not a tree: 0e062895678f49fd673ae493371262cfb8c5ab56
```

but only on 9.0.1. So I dug in, and noticed that commit referenced in `stack-9.0.1.yaml`. But it no longer appears to exist on your branch, I'm guessing because it was deleted after being merged here: https://github.com/HeinrichApfelmus/operational/pull/26

isovector:
Blocked by https://github.com/haskell/haskell-language-server/issues/2016

isovector:
\o/

PR #1708, opened 2021-04-11
Fix: #1690 - Infix typed holes are now filled using infix notation

OliverMadine:
fix: #1690

In follow up to my response to @July541's PR (#1694), I have written a different solution to cover the cases of filling holes with infix operators.

applying the "replace _ with (+)" code action to
```haskell
test :: Int -> Int -> Int
test a1 a2 = a1 `_` a2
```

now results in 

```haskell
test :: Int -> Int -> Int
test a1 a2 = a1 + a2
```

rather than 

```haskell
test :: Int -> Int -> Int
test a1 a2 = a1 `(+)` a2
```

Note: This solution depends on `processHoleSuggestions` returning infix functions in their prefix form (e.g. `(+)`) to determine if a function infix/prefix. I imagine there might be a nicer way of doing this.

Edit: In some commit messages I have written "postfix" by mistake instead of "prefix"

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -677,16 +677,24 @@ suggestModuleTypo Diagnostic{_range=_range,..}
 suggestFillHole :: Diagnostic -> [(T.Text, TextEdit)]
 suggestFillHole Diagnostic{_range=_range,..}
     | Just holeName <- extractHoleName _message
-    , (holeFits, refFits) <- processHoleSuggestions (T.lines _message)
-    = map (proposeHoleFit holeName False) holeFits
-    ++ map (proposeHoleFit holeName True) refFits
+    , (holeFits, refFits) <- processHoleSuggestions (T.lines _message) =
+      let isInfixHole = _message =~ addBackticks holeName :: Bool in
+        map (proposeHoleFit holeName False isInfixHole) holeFits
+        ++ map (proposeHoleFit holeName True isInfixHole) refFits
     | otherwise = []
     where
       extractHoleName = fmap head . flip matchRegexUnifySpaces "Found hole: ([^ ]*)"
-      proposeHoleFit holeName parenthise name =
+      addBackticks text = "`" <> text <> "`"
+      addParens text = "(" <> text <> ")"
+      proposeHoleFit holeName parenthise isInfixHole name =
+        let isInfixOperator = T.head name == '('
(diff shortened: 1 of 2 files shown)


jneira (approved):
LGTM, thanks for the fix, tests cases looks great, many thanks!

jneira:
> This solution depends on processHoleSuggestions returning infix functions in their postfix form (e.g. (+)) to determine if a function infix/postfix. I imagine there might be a nicer way of doing this.

The unique caveat i can think of is somebody is using already postfix notation and we turn it in infix one, is there some test case to check we will not do it?

OliverMadine:
> > This solution depends on processHoleSuggestions returning infix functions in their postfix form (e.g. (+)) to determine if a function infix/postfix. I imagine there might be a nicer way of doing this.
> 
> The unique caveat i can think of is somebody is using already postfix notation and we turn it in infix one, is there some test case to check we will not do it?

Do you mean that we should check that infix form is not used when postfix is required?

So we should make sure 

```Haskell
foo :: Int -> Int -> Int
foo = _ a1 a2
```

is filled to

```haskell
foo :: Int -> Int -> Int 
foo a1 a2 = (+) a1 a2
```
?

jneira:
yeah, that was my idea, many thanks

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -677,16 +677,24 @@ suggestModuleTypo Diagnostic{_range=_range,..}
...
+        map (proposeHoleFit holeName False isInfixHole) holeFits
+        ++ map (proposeHoleFit holeName True isInfixHole) refFits
     | otherwise = []
     where
       extractHoleName = fmap head . flip matchRegexUnifySpaces "Found hole: ([^ ]*)"
-      proposeHoleFit holeName parenthise name =
+      addBackticks text = "`" <> text <> "`"
+      addParens text = "(" <> text <> ")"
+      proposeHoleFit holeName parenthise isInfixHole name =
+        let isInfixOperator = T.head name == '('
```
Ailrun, line 690: Does this condition work even when the hole is of type `()`?

OliverMadine: I'm not sure I understand what you mean. Please could you give an example?

The "name" is referring to the name of the proposed fit to the hole, which should not be dependent on the type of the hole.

Ailrun: Nvm, I somehow thought that `name` can be `()`.

Issue #1700, opened 2021-04-10
Enhancement: parenthesise compound completion snippets


konn:
Currently, HLS's completion snippet inserts types or kinds as-is.
For example,

```haskell
foldl
```

gets completed to:

```haskell
foldl b -> a -> b b t a
```

where `b -> a -> b`, `b` and `t a` are three separate snippet placeholders.
This makes it harder to detect boundaries of generated "placeholders" after a snippet is completed tentatively.
It might be good to parenthesise each placeholder when it is compound.
For example, the above must be at least:

```hs
foldl (b -> a -> b) b (t a)
```

or:
```hs
foldl (_ :: b -> a -> b) (_ :: b) (_ :: t a)
```

Hole-based approach works better as it generates well-formed programs; however, holes can make GHC significantly slower, it might be good to parenthesise plain placeholders (with types or kinds alone) could be enough.

pepeiborra:
This is a bit tricky:  

1. Typed holes will only generate well-formed programs when ScopedTypeVariables is on and all the inserted types are in scope. 
2. The type will be uninstantiated, and cannot be instantiated since we collect completions for a module, not for a call site. So the type will possibly be too general and either lead to a type error or make the hole suggestions less useful.
3. If I want to use the hole suggestions to fill in the snippet, I will have to clean up the parenthesis and the type annotation afterwards, leading to way more typing noise than right now.

Given 2 and 3, I don't think this will work so well in practice. But I might be wrong! The only way to find out is prototype it and have a play with some real code

konn:
I agree that we dont need to generate well-formed programs and the first simply parenthesised solution would do.

> the user is expected to rewrite the snippets immediately and editors optimise for this by deleting the whole snippet as soon as the user starts typing, and keybinding TAB to move to the next snippet

I'm slightly against on this point from my experience. We might ponder what to fill in the placeholders for a while, and browses to other location or modules to get insights. Or, some placeholder gets somewhat complicated during the edition. We  can easilly end up with exitting placeholders in such cases.
So, I still strongly think that it really makes sense simply to parenthesise each placeholders (w/o holes) to make argument boundary clear.

konn:
To make my point clearer: the intention of this proposal is to make the argument boundaries in snippets clear. Well-formedness is just a bonus.
[...]

PR #1709, opened 2021-04-11
Bracketing for snippet completions

OliverMadine:
Fix: #1700

This can easily be changed to the first variation suggested in the issue (without typed holes).
I'm not sure how to write a test case for this.


ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -247,7 +247,7 @@ mkNameCompItem doc thingParent origName origMod thingType isInfix docs !imp = CI
       where
         argTypes = getArgs typ
         argText :: T.Text
-        argText =  mconcat $ List.intersperse " " $ zipWithFrom snippet 1 argTypes
+        argText = mconcat $ List.intersperse " " $ zipWithFrom (\i t -> T.pack "(_ :: " <> snippet i t <> T.pack ")") 1 argTypes
         snippet :: Int -> Type -> T.Text
         snippet i t = "${" <> T.pack (show i) <> ":" <> showGhc t <> "}"
```
pepeiborra, line 252: I think that the `(_ :: ..)` text should be part of the snippet and not part of the completion, since snippets are an optional feature and can be disabled in config.

pepeiborra (changes requested):
I left a comment in the original issue on why I am skeptical that this is an improvement.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
konn: Yes, my first intention in #1700  is to include brackets into snippets, not outside.

OliverMadine: Thanks for the feedback,

Is the latest commit more what you had in mind? 

Note:
Including the brackets inside the code snippet means that the brackets are also included in the preselected code after the completion. I think that this is desirable since they might sometimes be redundant.

OliverMadine: Thanks for the feedback,

Is the latest commit more like what you had in mind? 

Note:
Including the brackets inside the code snippet means that the brackets are also included in the preselected code after the completion. I think that this is desirable since they might sometimes be redundant.

pepeiborra: Yes, that makes sense

ghcide/bench/example/HLS
```diff
@@ -1 +0,0 @@
-../../..
```
pepeiborra, line 1: could you restore this file?

pepeiborra (changes requested):
Nice job! 

It looks like a symbolic link was accidentally deleted - please restore it before merging.

ghcide/bench/example/HLS
OliverMadine: Restored. I'm not sure how this went missing.

Thanks for all the feedback.

pepeiborra (approved):

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -247,9 +248,16 @@ mkNameCompItem doc thingParent origName origMod thingType isInfix docs !imp = CI
...
         argText :: T.Text
-        argText =  mconcat $ List.intersperse " " $ zipWithFrom snippet 1 argTypes
+        argText = mconcat $ List.intersperse " " $ zipWithFrom snippet 1 argTypes
         snippet :: Int -> Type -> T.Text
-        snippet i t = "${" <> T.pack (show i) <> ":" <> showGhc t <> "}"
+        snippet i t = case t of
+            (TyVarTy _)     -> noParensSnippet
+            (LitTy _)       -> noParensSnippet
+            (TyConApp _ []) -> noParensSnippet
+            _               -> snippetText i (T.pack "(" <> showGhc t <> T.pack ")")
```
Ailrun, line 257: BTW, you can use just `"("` since it's under OverloadedStrings.

Ailrun, line 257: BTW, you can use just `"("` since it's under OverloadedStrings.

Ailrun (approved):
Awesome, thank you!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
OliverMadine: Done, thanks for the feedback!

konn (approved):
LGTM 👍 
Sorry for late reaction, and thank you for the implementation!

PR #1715, opened 2021-04-12
Don't suggest destruct actions for already-destructed terms

### LINKED ISSUE #1711: Wingman shouldn't suggest destruct actions for things that are already destructed
Related to #1704, if we had a consistent hypothesis throughout a definition, Wingman could notice that there is already provenance from a given value. Therefore, it must have already been destructed, and we shouldn't suggest it. That would clean up a lot of the spurious suggestions.

isovector:
This PR does an extra pass over the existing function definition, looking for `case` expressions that scrutinize a single variable. Every one it finds gets removed from Wingman's allowed hypothesis set, which effectively removes it from participating in anything that Wingman does. This includes the "destruct" code action.

Fixes #1711


jneira (approved):
lgtm

PR #1717, opened 2021-04-12
Restore compat. with haddock-library 1.8

pepeiborra:
Will need a Hackage release after landing


pepeiborra:
8.6.* stack descriptiors remain on haddock-library-1.10 to exercise both CPP paths

berberman (approved):
Thanks, I missed this

PR #1702, opened 2021-04-10
Prepare 1.1.0 release

berberman:
I think we should release ghcide 1.2 first, no?

Issue #1434, opened 2021-02-24
Document the Haskell Language Server versioning policy


pepeiborra:
It's clearly not the PVP, so it ought to be documented somewhere for the sake of future contributors and maintainers.

The documentation should include answers for the following questions:

1. Motivation for deviating from the PVP, and tradeoffs
2. Specification: lower bounds, upper bounds, and version bumps, for all the categories of packages. 
  There are 3 categories imho: 
    1. core packages 
    2. plugin packages
    3. the HLS package
3. Examples of releases that involve interactions with Hackage: 
    1. Major release 
    2. Bug fix release for core packages
    3. Bug fix release for plugin packages
    4. Bug fix release for HLS

jneira:
We are following pvp almost strictly in practice, not sure if it still is needed to document it somewhere

pepeiborra:
That's great news, will close then

Issue #1310, opened 2021-02-05
Automate hackage releases


jneira:
* From the recent experience we know that:
  * Users needs a hackage release for every github release we do: #1282
  * The hackage release is cumbersome and error prone and involve the coordination of several contributors, see #1287 and #1175
* So we *have* to automate the process as much as we can, making the automating process itself lightweight if possible

The steps needed to make the release are:

1. Identify changes of each subpackage and check if they can break downstream ones
2. Bump up the package version accordingly and propagate it to the bounds of downstream packages
3. Make pr(o several prs) with those changes, ensuring the order of dependencies
4. Cherry pick those changes onto the github release commit as a branch
5. Check the subpackage builds succesfully in that branch (and test it?)
6. Do a `cabal check && cabal haddock subpackage && cabal sdist subpackage`
7. Do a `cabal upload subpackage` (we can use https://github.com/haskell-actions/hackage-publish)

I think the steps 5, 6 and 7 could be automated in ci, uploading package candidates to do a final review and publish all of them.

Not sure how could we automate 1-4 though. Maybe doing a `git diff` per subpackage and update version bounds of .cabal files with some form of grep&sed (bumping out version as always breaking with no fine grained pvp considerations as suggested by @michaelpj)

There are projects with lot of subpackages and maybe they have been some work to automate this, as pointed by @wz1000.
For example amazonka. Afaics they have a Makefile that includes upload of packages: https://github.com/brendanhay/amazonka/blob/develop/Makefile

ndmitchell:
Note that step 1 is optional. The PVP says you MUST increase the bounds IF you have incompatible changes. There's nothing to stop you increasing the bounds ALWAYS. Moreover, I don't see a huge heap of value in keeping the bounds as broad as possible, given we do a monolithic release, and people should always use the latest stuff together.

I'd simplify by writing tooling to bump all bounds, including propagating downstream dependency bounds, in one atomic go. 

What special testing should a release get? Ideally none, because we're so confident in our tests :)

michaelpj:
Neil said exactly what I tried to say on the other issue but much better :)

As another angle: we don't *gain* anything much from fine-grained bounds wrangling between plugins. At best, that might allow someone to use a newer version of a plugin with an old version of HLS. But do we actually care much about this usecase? If the versions of HLS and the plugins move in lockstep (as if they were one package!)... would anyone actually mind? I don't think so. Maybe later in the future we might care about this.

[...]

PR #1718, opened 2021-04-12
Bump plugins in preparation for Hackage upload

pepeiborra:
Please check that I haven't missed any

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -187,12 +187,12 @@ common example-plugins
 
 common class
   if flag(class) || flag(all-plugins)
-    build-depends: hls-class-plugin ^>= 1.0.0.1
+    build-depends: hls-class-plugin ^>= 1.1.0.0
     cpp-options: -Dclass
 
 common haddockComments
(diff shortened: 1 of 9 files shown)


berberman:
Plugins' version have already been bumped in #1702. According to PVP, except eval plugin and tactics plugin, which have api-breaking changes so their major version should be bumped, we should bump ohter plugins' minor version. But all of these assume that we are using PVP -- do we have something like "HLSVP" that says plugins' releasing version should be synchronized with which of HLS?

pepeiborra:
Much better, let's stick to the PVP then!

jneira:
Mmm, well, we talked about set (at least) hls and plugin versions in sync, following loosely pvp: 
* https://github.com/haskell/haskell-language-server/issues/1310#issuecomment-778352533

There is an issue about establishing a policy about versioning but we it still is open:
* https://github.com/haskell/haskell-language-server/issues/1434

I am not against following pvp, only want to provide some context from older releases. If we decided to follow strictly pvp, i would document it as suggested in #1434, to try to be consistent.

PR #1721, opened 2021-04-12
Wingman: Destruct on empty case

### LINKED ISSUE #1716: case split on case of statement with no destructor cases declared
This isn't exactly Wingman, but I think @isovector is probably the man for the job (?) given what he's been working on...

```
case my expr of
```

In this case, we'll get some errors about allow "empty-case" and add "EmptyCase" etc. 

My suggestion: to add a code action that inserts empty case destructors:

```
case myMaybeValue of
```

becomes

```
case myMaybeValue of
   Nothing -> _
   Just val -> _
```

and now the same Wingman workflows are recovered.

isovector:
This PR adds a code lens to empty `case x of` expressions, offering to expand out their pattern matches. It works by looking for empty case matches, typechecking their scrutinee, and then grafting newly generated `Match` es back into the AST. Seems to work pretty well.

destruct on empty case

Fixes #1716

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/Types.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/Types.hs
@@ -117,6 +117,9 @@ instance Show Class where
 instance Show (HsExpr GhcPs) where
   show  = unsafeRender
 
+instance Show (HsExpr GhcTc) where
+  show  = unsafeRender
+
 instance Show (HsDecl GhcPs) where
   show  = unsafeRender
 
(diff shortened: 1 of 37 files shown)


wz1000:
I wonder if this would be better to implement as a completion rather than a code action?

plugins/hls-tactics-plugin/src/Wingman/Types.hs
```diff
@@ -117,6 +117,9 @@ instance Show Class where
 instance Show (HsExpr GhcPs) where
   show  = unsafeRender
 
+instance Show (HsExpr GhcTc) where
+  show  = unsafeRender
+
 instance Show (HsDecl GhcPs) where
   show  = unsafeRender
```
pepeiborra, line 122: Please consider moving these to the Orphans module

pepeiborra (approved):
This is awesome! I have wanted this feature so many times!

plugins/hls-tactics-plugin/src/Wingman/Types.hs
isovector: Happy to --- followup PR ok?

PR #1724, opened 2021-04-14
HLint: Pass options through user config

### LINKED ISSUE #1509: Possible to pass options to HLint?
I have my HLint config in a non-standard location. Is it possible to pass `--hint my-hlint-config.yaml` through to HLint? Without it I get only the default hints.

rmehri01:
Closes #1509.

I actually got it to work on a sample file using:
```json
"flags": [
    "-iUnused LANGUAGE pragma",
    "-h/Users/ryanmehri/my-hlint-config.yaml"
]
```

I think the fact that the command line arguments were different than the library ones just threw me off! I'll be working on adding the tests soon.


plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -125,7 +129,13 @@ rules plugin = do
     ideas <- if hlintOn' then getIdeas file else return (Right [])
     return (diagnostics file ideas, Just ())
 
-  getHlintSettingsRule (HlintEnabled [])
+  defineNoFile $ \GetHlintSettings -> do
+    config <- getClientConfigAction def
+    let pluginConfig = configForPlugin config plugin
+    getHlintSettingsRule $ HlintEnabled $
+      case fromJSON $ Object $ plcConfig pluginConfig of
```
berberman, line 136: Please use Ide.Plugin.Properties to define this custom config, instead of parsing json object directly. You may want to take a look at type lenses plugin or tactics plugin to see how to use it.

ndmitchell:
I'd have expected that all the config options are documented somewhere. If so, this should be added.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -258,15 +265,40 @@ instance Binary GetHlintSettings
...
+usePropertyAction ::
+  (HasProperty s k t r) =>
+  KeyNameProxy s ->
+  PluginId ->
+  Properties r ->
+  Action (ToHsType t)
+usePropertyAction kn pId p = do
+  config <- getClientConfigAction def
+  let pluginConfig = configForPlugin config pId
+  return $ useProperty kn p $ Just (plcConfig pluginConfig)
```
berberman: This function is already in ghcide:
https://github.com/haskell/haskell-language-server/blob/0c5a317560c4ee7257ddf03de117868cb7a056c2/ghcide/src/Development/IDE/Core/Rules.hs#L949-L958

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -258,15 +265,40 @@ instance Binary GetHlintSettings
 
 type instance RuleResult GetHlintSettings = (ParseFlags, [Classify], Hint)
 
-getHlintSettingsRule :: HlintUsage -> Rules ()
+getHlintSettingsRule :: HlintUsage -> Action (ParseFlags, [Classify], Hint)
 getHlintSettingsRule usage =
-    defineNoFile $ \GetHlintSettings ->
       liftIO $ case usage of
           HlintEnabled cmdArgs -> argsSettings cmdArgs
           HlintDisabled        -> fail "hlint configuration unspecified"
```
berberman, line 272: It's not a rule, but an action. I would inline this function.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -125,7 +130,9 @@ rules plugin = do
     ideas <- if hlintOn' then getIdeas file else return (Right [])
     return (diagnostics file ideas, Just ())
 
-  getHlintSettingsRule (HlintEnabled [])
+  defineNoFile $ \GetHlintSettings -> do
+      (Config flags) <- getHlintConfig plugin
+      getHlintSettingsRule (HlintEnabled flags)
```
berberman, line 135: Are we using `HlintDisabled` somewhere? If no, can we remove `HlintUsage`?

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -258,15 +265,40 @@ instance Binary GetHlintSettings
...
-    defineNoFile $ \GetHlintSettings ->
       liftIO $ case usage of
           HlintEnabled cmdArgs -> argsSettings cmdArgs
           HlintDisabled        -> fail "hlint configuration unspecified"
 
 -- ---------------------------------------------------------------------
 
+newtype Config = Config [String]
+
+properties :: Properties '[ 'PropertyKey "flags" ('TArray String)]
```
berberman, line 278: Please add `properties` to plugin descriptor, for example:
https://github.com/haskell/haskell-language-server/blob/2683a92af0ce043254d7ba8b4d379d8e7d9075fb/ghcide/src/Development/IDE/Plugin/TypeLenses.hs#L93

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
rmehri01: I was also wondering this, I don't think it is used anywhere else.

jneira:
Hi, many thanks for your work, the plugin will have a very handy and flexible config option (maybe it will not need no one more :smile:)
I would like to have a positive test: maybe starting with no hlint hints, change the config and then check the expected new hint is present, @rmehri01 what do you think?

rmehri01:
> Maybe to have a positive test would be interesting: starting with no hlint hints, change the config and then check the expected new hint is present

Sounds good, I'll add it!

jneira (approved):
This looks great, many thanks again

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -95,8 +101,8 @@ descriptor plId = (defaultPluginDescriptor plId)
       [ PluginCommand "applyOne" "Apply a single hint" applyOneCmd
       , PluginCommand "applyAll" "Apply all hints to the file" applyAllCmd
       ]
-    , pluginHandlers = mkPluginHandler STextDocumentCodeAction codeActionProvider
-    , pluginConfigDescriptor = defaultConfigDescriptor {configHasDiagnostics = True}
+  , pluginHandlers = mkPluginHandler STextDocumentCodeAction codeActionProvider
+  , pluginConfigDescriptor = defaultConfigDescriptor {configCustomConfig = mkCustomConfig properties}
```
berberman, line 105: Why removing `configHasDiagnostics = True`?

jneira: Ugh good catch

rmehri01: Oops, sorry about that!

jneira:
Well i guess this is good to go now
//cc @berberman

PR #1727, opened 2021-04-14
Set TextDocumentEdit version to Nothing instead of Just 0

mfussenegger:
Clients like the one built into neovim reject edits where the version is
lower than the expected document version.

This was always the case with hls because the version is hardcoded to 0.
vscode seems to accept that, but based on the specification it is more
accurate to set the version to `null` if it is not know.

The `textDocument` of `TextDocumentEdit` has the type
`OptionalVersionedTextDocumentIdentifier`

---

Update: Figured out that in 3.15 the `textDocument` was a `VersionedTextDocumentIdentifier`. We merged a change in neovim to allow 0 versions as well so I guess this PR is obsolete.


Ailrun:
If this is obsolete, could you close this?

mfussenegger:
> If this is obsolete, could you please close this?

Sure

PR #1729, opened 2021-04-15
Add installation instructions for FreeBSD.

arrowd:


jneira (approved):
Many thanks for add support for freebsd and include it here!

hasufell:
The docs might need adjustment after https://github.com/haskell/haskell-language-server/pull/2200

arrowd:
From what I can tell, FreeBSD binaries are being produced for FreeBSD 13. They should be produced on FreeBSD 12.2 to be compatible with both 12 and 13.

I also prefer our Ports for producing packages, but if this works out - sure, why not.

hasufell:
@arrowd https://gitlab.haskell.org/ghc/ghc/-/issues/20390

PR #1730, opened 2021-04-15
Add executable stanza in hls-install.cabal.

arrowd:
This helps packaging on FreeBSD where we build installer executable as a
separate package.


jneira (approved):
lgtm, thanks for the change

Issue #1726, opened 2021-04-14
Language Pragmas Auto-Insert After Module Docs 


emilypi:
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
haskell-language-server version: 1.1.0.0 (GHC: 8.10.4) (PATH: /Users/emilypi/.ghcup/bin/haskell-language-server-wrapper-1.1.0) (GIT hash: f1c096927186a93d8e3ccd4fe8385cc1b070350b)
Tool versions found on the $PATH
cabal:		3.4.0.0
stack:		2.5.1
ghc:		8.10.4
```
Which OS do you use:
Big Sur OSX

Which lsp-client do you use:
Emacs lsp-haskell
Describe your project (alternative: link to the project):
Cabal project
Contents of `hie.yaml`:
```yaml
cradle:
  cabal:
    - path: "Cabal/src"
      component: "lib:Cabal"

    - path: "cabal-install/src"
      component: "lib:cabal-install"

    - path: "cabal-install/main/Main.hs"
      component: "cabal-install:exe:cabal"

    - path: "cabal-install/tests"
      component: "cabal-install:test:unit-tests"

    - path: "cabal-install/tests"
      component: "cabal-install:test:memory-usage-tests"

    - path: "cabal-install/tests"
      component: "cabal-install:test:integration-tests2"

    - path: "cabal-install/tests"
      component: "cabal-install:test:long-tests"

    - path: "cabal-install-solver/src"
      component: "lib:cabal-install-solver"

    - path: "cabal-install-solver/src-assertion"
      component: "lib:cabal-install-solver"
```

### Steps to reproduce

On HLS 1.1, auto-insert any language pragma in a module with header documentation

### Expected behaviour
Pragmas should be inserted before docs

### Actual behaviour
Pragmas are inserted *after* docs: 

```haskell
{-# LANGUAGE BangPatterns #-}
{-# LANGUAGE DeriveFunctor #-}
{-# LANGUAGE DeriveFoldable #-}
{-# LANGUAGE DeriveTraversable #-}
-- | Cabal-like file AST types: 'Field', 'Section' etc,
--
-- This (intermediate) data type is used for pretty-printing.
--
-- @since 3.0.0.0
--
{-# LANGUAGE LambdaCase #-}
module Distribution.Fields.Pretty 
```

OliverMadine:
[...]

PR #1417, opened 2021-02-21
Refactor pragmas plugin

berberman:
Closes #1413

~~BTW I still couldn't run pre-commit hook successfully...~~
```
$ git commit -m "Refactor pragmas plugin"
stylish-haskell..........................................................Failed
- hook id: stylish-haskell
- exit code: 1

stylish-haskell: 0:0:Language.Haskell.Stylish.Config.loadConfig: fromJSON: expected Bool, but encountered String

CallStack (from HasCallStack):
  error, called at lib/Language/Haskell/Stylish/Config.hs:107:32 in stylish-haskell-0.11.0.3-F2KP2B6JiKw5YfjeWoLVgI:Language.Haskell.Stylish.Config
stylish-haskell: 0:0:Language.Haskell.Stylish.Config.loadConfig: fromJSON: expected Bool, but encountered String

CallStack (from HasCallStack):
  error, called at lib/Language/Haskell/Stylish/Config.hs:107:32 in stylish-haskell-0.11.0.3-F2KP2B6JiKw5YfjeWoLVgI:Language.Haskell.Stylish.Config
```

pre-commit hook purposed in #1384 enforced formatting files edited in this PR, a good sign!

Issue #555, opened 2020-10-30
Language pragmas are inserted incorrectly at the top of file (before shebangs)


expipiplus1:
- With the following file

    ```haskell
    #! /usr/bin/env nix-shell
    #! nix-shell --pure -i runghc -p "haskellPackages.ghcWithPackages (hp: with hp; [ turtle ])"

    foo = id @()
    ```

- Accept the suggestion to add the `TypeApplications` extension.

- Observe that the language extension is incorrectly inserted above the shebangs

I'm pretty sure I saw `{-# LANGUAGE TypeApplications #-}` appear at the end of the `#! nix-shell` line at one point, but I can't reproduce that

PRs welcome!

expipiplus1:
Thanks!

PR #1731, opened 2021-04-15
Insert pragmas after shebang or to existing pragma list

OliverMadine:
Fix: #1726

Pragmas are now inserted before comments.

Updated the behaviour to insert directly after existing pragmas, else directly after shebangs, else at the top of the file. 

Refactored to use the file contents rather than a parsed module so that missing pragmas, which cause parse errors, are still inserted into the correct position.


jneira:
I commented too late :worried: : https://github.com/haskell/haskell-language-server/issues/1726#issuecomment-820237109

OliverMadine:
> I commented too late 😟 : [#1726 (comment)](https://github.com/haskell/haskell-language-server/issues/1726#issuecomment-820237109)
> Not sure if your final solution is affected by though
> 
> > I will add test cases but in either case
> 
> Tests are definitely needed, to ensure there is no regression and to see examples about how it works

If we were to enable the option to insert at the top of the file, I suppose we would just have #555 again. 
Also since we are working with a ParsedModule, it would be difficult to get the span of pragma blocks? 
So perhaps I should just close this PR, but then I'm also not sure how else we could solve this issue?

I suspected this may be the case, hence I waited before writing the tests 😔

berberman:
I would say adding an option is definitely not ideal. Perhaps we could fix the real issue by finding the first non-empty line after a possible shebang, using file contents rather than parsed module.

jneira:
> Also since we are working with a ParsedModule, it would be difficult to get the span of existing pragma blocks?

No an expert in the ghc parsed module but i think they would be present as annotations. We have to use the version of the parsed module that includes such annotations (and no the default one)

@emilypi could be a little bit more adhoc but could work too. In fact i think #555 tried to do it but was superseded by #1417 which established the actual behaviour.

jneira:
#555 added a regression test over the shebang case:

https://github.com/haskell/haskell-language-server/blob/c117657c05327b25f82b66e1bcade517cfe6614a/test/functional/FunctionalCodeAction.hs#L558-L588

And this pr didnt break it (and should afaiu), i suppose the option is disabled by default

michaelpj:
Perhaps I missed some discussion, but the issue was that we were inserting pragmas after module haddock, right? So could we just extend the property to be "insert just before the first declaration, import, module header, _or haddock comment_"?

OliverMadine:
> Perhaps I missed some discussion, but the issue was that we were inserting pragmas after module haddock, right? So could we just extend the property to be "insert just before the first declaration, import, module header, _or haddock comment_"?

Is it necessary to insert before _haddock comments_ or could we just insert before any comments? I've just pushed a commit to do the latter but I can change it to specifically haddock comments if necessary.

jneira:
before any comments sound good to me too 🤔

test/functional/FunctionalCodeAction.hs
```diff
@@ -513,8 +513,8 @@ missingPragmaTests = testGroup "missing pragma warning code actions" [
             let expected =
 -- TODO: Why CPP???
 #if __GLASGOW_HASKELL__ < 810
-                    [ "{-# LANGUAGE ScopedTypeVariables #-}"
-                    , "{-# LANGUAGE TypeApplications #-}"
+                    [ "{-# LANGUAGE TypeApplications #-}"
+                    , "{-# LANGUAGE ScopedTypeVariables #-}"
 #else
                     [ "{-# LANGUAGE TypeApplications #-}"
                     , "{-# LANGUAGE ScopedTypeVariables #-}"
```
berberman, line 520: This macro seems to be redundant now

berberman (approved):
Looks great, thanks!

michaelpj:
Let me be annoying, what about this?

```
{-# LANGUAGE ... -#}
-- this is here for reason X
{-# LANGUAGE ... -#}
```

I think it's pretty common for people to put comments inside their pragma blocks, which will result in slightly weird behaviour. But nobody puts _haddock_ comments in their pragma blocks!

pepeiborra:
Can we also consider copyright comments? E.g.

```
-- Copyright (c) 2019 The DAML Authors. All rights reserved.
-- SPDX-License-Identifier: Apache-2.0
{-# OPTIONS_GHC -Wno-dodgy-imports #-} -- GHC no longer exports def in GHC 8.6 and above
{-# LANGUAGE TemplateHaskell #-}
```

pepeiborra:
Personally, I think the property should be: 

> Append at the end of the existing list of pragmas, if any, otherwise insert just before the first declaration, import, module header, or haddock comment

OliverMadine:
> Let me be annoying, what about this?
> 
> ```
> {-# LANGUAGE ... -#}
> -- this is here for reason X
> {-# LANGUAGE ... -#}
> ```
> 
> I think it's pretty common for people to put comments inside their pragma blocks, which will result in slightly weird behaviour. But nobody puts _haddock_ comments in their pragma blocks!

Thanks, I hadn't thought about this.



> Personally, I think the property should be as below, even if this doesn't deal properly with copyright headers when there is no existing pragmas:
> 
> > Append at the end of the existing list of pragmas, if any, otherwise insert just before the first declaration, import, module header, or haddock comment

This should now be the behaviour of the code.

test/functional/FunctionalCodeAction.hs
OliverMadine: Thanks, I missed this but since adding pragmas to the end of the list of pragmas, it is necessary again anyway.

berberman:
> Append at the end of the existing list of pragmas, if any, otherwise insert just before the first declaration, import, module header, or haddock comment

Can we reach this by merely looking at file contents instead of parsed module? If there is an existing pragma, we insert the new one below it; otherwise we insert the new one after shebang meanwhile before the first non-empty line. Not sure if this will work or I missed something, but I believe getting rid of parsed module would avoid mysterious CPP macro in test.

OliverMadine:
> > Append at the end of the existing list of pragmas, if any, otherwise insert just before the first declaration, import, module header, or haddock comment
> 
> Can we reach this by merely looking at file contents instead of parsed module? If there is an existing pragma, we insert the new one below it; otherwise we insert the new one after shebang meanwhile before the first non-empty line. Not sure if this will work or I missed something, but I believe getting rid of parsed module would avoid mysterious CPP macro in test.

I can't see any immediate reason why this wouldn't work. What would be the performance difference with this solution? I might be able to refactor it to take this approach at some point, but I have exams at the moment 😔. In the meantime, since we don't guarantee any order of inserting pragmas, are there any practical issues with the current solution?

sorry for the late response

berberman:
> are there any practical issues with the current solution?

No, I think it's good enough :)

I suggest not using parsed module because missing language extensions like TypeApplications will result in a parse error, rather than a typecheck error - in such case our strategy will no loger work, and the pragma will be inserted to the top of file.

michaelpj:
Ugh, that is a bit annoying. Maybe we should add a paragraph to the known limitations section about that, since I bet people will hit it (I think inserting `TypeApplications` is one of my top uses!).

But yeah, I don't think fixing it should block this PR.

OliverMadine:
> I suggest not using parsed module

> Ugh, that is a bit annoying. Maybe we should add a paragraph to the known limitations section about that, since I bet people will hit it (I think inserting `TypeApplications` is one of my top uses!).

I decided it'd be best I to just fix this issue now so I've pushed a new commit which hopefully should.

Issue #703, opened 2020-12-27
Distribute binaries via Homebrew


ad-si:
It would be cool if I could install hls it with a simple:
```sh
brew install haskell-language-server haskell-language-server-wrapper
```
Here is the documentation on how to add it to Homebrew:
https://docs.brew.sh/Adding-Software-to-Homebrew

jneira:
More installation methods supposes to reach more users and that is good. However they should be set up and more importantly, updated. IDeally it should be done in ci automatically.
So we need someone willing to accomplish the task 😄

jneira:
Some investigation has been done by @ELLIOTTCABLE: https://github.com/Homebrew/discussions/discussions/386
Thanks for it!

jneira:
haskell-language-server is on hackage now and can be installed in a more friendly way with cabal install: https://hackage.haskell.org/package/haskell-language-server

kret:
`haskell-language-server` is now available as a Homebrew formula (`brew install haskell-language-server`). This is a version compiled with current GHC from Homebrew (8.10.4). I hope it works 😅 

I guess this is not ideal (single GHC version support), but I think it is a reasonable compromise between Haskellers' needs and Homebrew's rules.

IMO it would be great to distribute the whole release. As I see it we could create a "Haskell tap" similar to this: https://github.com/kret/homebrew-tools, but in an official Haskell repo. There we could distribute binaries without need to strictly conform to Homebrew rules, but at the cost of having to occasionally update it ourselves.
I was thinking about distributing the wrapper and language server binaries per GHC version so that users could pick whatever they need, perhaps also with a single command full release installation. I can prepare such version on my private tap as an example and then it could be moved over to official repo, if accepted as a solution.

kret:
Updated my private tap (https://github.com/kret/homebrew-tools).
It now contains 5 casks:
- `kret/tools/haskell-language-server-full`
- `kret/tools/haskell-language-server-wrapper`
- `kret/tools/haskell-language-server-8104`
- `kret/tools/haskell-language-server-884`
- `kret/tools/haskell-language-server-865`

The `full` contains the full release of course, the rest contains single binaries (8104 - because dots are not allowed in names, hyphens could be there if you think they look better).

You can play around with un/installing and using to verify if that is usable.
All comments are welcome, I can make updates, or you can open PRs of course.

If this becomes satisfactory, we can decide next steps, like:
- moving to an official Haskell/HLS tap,
- which binaries to include like this,
[...]

PR #1734, opened 2021-04-15
Add haskell-language-server Homebrew installation instructions

kret:
As suggested in https://github.com/haskell/haskell-language-server/issues/703#issuecomment-815464067, suggesting an update to README Installation Instructions - adding Homebrew installation section.


jneira (approved):
Many thanks (again), nice doc addition for macos users

PR #1739, opened 2021-04-16
Wingman: case split on punned record fields

### LINKED ISSUE #1736: Splitting a variable bound in a `NamedFieldPun` duplicates record match
Given the following code:

```haskell
data Bar = Bar { ax :: Int, bax :: Bool }

bar :: Bar -> Int
bar Bar {ax, bax} = _
```

executing the `case split on bax` code action results in 

```haskell
data Bar = Bar { ax :: Int, bax :: Bool }

bar :: Bar -> Int
bar Bar {ax, bax} = _
bar Bar {ax, bax} = _
```

a workaround is turning the pattern in question into a normal record field binding first:

```haskell
bar :: Bar -> Int
bar Bar {ax, bax = bax} = _
```

then executing the same code action results in the expected output:

```haskell
bar :: Bar -> Int
bar Bar {ax, bax = False} = _
bar Bar {ax, bax = True} = _
```

I've tried this on both the latest master (9ae20922) and the latest release (`1.1`) and both exhibit this behaviour.

isovector:
Fixes #1736

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/CaseSplit.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/CaseSplit.hs
@@ -64,8 +64,8 @@ rewriteVarPat name rep = everywhere $
   `extT` \case
     HsRecField lbl _ True
       | eqRdrName name $ unLoc $ rdrNameFieldOcc $ unLoc lbl
-          -> HsRecField lbl rep False
-    (x :: HsRecField' (FieldOcc GhcPs) (LPat GhcPs)) -> x
+          -> HsRecField lbl (toPatCompat rep) False
+    (x :: HsRecField' (FieldOcc GhcPs) (PatCompat GhcPs)) -> x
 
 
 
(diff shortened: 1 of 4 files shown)


Ailrun (approved):
Awesome

jneira (approved):

plugins/hls-tactics-plugin/src/Wingman/CaseSplit.hs
```diff
@@ -56,9 +56,17 @@ wildifyT (S.map occNameString -> used) = everywhere $ mkT $ \case
...
+rewriteVarPat name rep = everywhere $
+  mkT (\case
+    VarPat _ (L _ var) | eqRdrName name var -> rep
+    (x :: Pat GhcPs)                        -> x
+      )
+  `extT` \case
+    HsRecField lbl _ True
+      | eqRdrName name $ unLoc $ rdrNameFieldOcc $ unLoc lbl
+          -> HsRecField lbl rep False
+    (x :: HsRecField' (FieldOcc GhcPs) (LPat GhcPs)) -> x
```
Ailrun, line 68: There's extra `L` here

isovector: Good catch. This is what I get for developing on 8.8 :(

PR #1741, opened 2021-04-17
Fix class method completion

### LINKED ISSUE #1738: Completions for methods of the current class in instance declarations
Consider the following example:

```haskell
class Test a where
    test :: a -> ()

instance Test Int where
    test = _
```

At `_`, it would be nice to have the method `test` in the completion list, e.g. for recursive invocations, this is not the case currently, however:

image

July541:
Fix #1738.

From [docs](https://hackage.haskell.org/package/ghc-8.6.5/docs/HsBinds.html#t:Sig), the signature of a class method should be `ClassOpSig`.


pepeiborra (approved):
Nice catch, and thanks for the test!

PR #1742, opened 2021-04-17
Improve vscode extension schema generation

berberman:
As discussed in https://github.com/haskell/vscode-haskell/issues/378#issuecomment-820711751, in order to make the generated vscode extension schema more handy, this PR introduces a new data type taking place of the old `pluginCustomConfig` in `PluginDescriptor`:

```haskell
data ConfigDescriptor = ConfigDescriptor {
  configEnableGenericConfig :: Bool,
  configHasDiagnostics      :: Bool,
  configCustomConfig        :: CustomConfig
}
```

* We shouold set `configHasDiagnostics` to `True` explicitly in plugin like hlint, so that we could generate `diagnosticsOn` option for it.
* Plugin will have single `globalOn` option only if it has one capability; otherwise we use specific capabilities such as `codeActionOn`, `codeLensOn`, etc. as options.
* Normally, we won't hope disable capabilities of ghcide-hover-and-symbols plugin. So we set `configEnableGenericConfig` to `False` to avoid generating options of this plugin.

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -47,7 +47,7 @@ descriptor plId = (defaultPluginDescriptor plId)
   { pluginRules = produceCompletions
   , pluginHandlers = mkPluginHandler STextDocumentCompletion getCompletionsLSP
   , pluginCommands = [extendImportCommand]
-  , pluginCustomConfig = mkCustomConfig properties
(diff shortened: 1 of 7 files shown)


jneira (approved):
Thanks for this changes, i agree they are sensible, only i am little bit worried that are things we have to (and remember to!) set manually (like we have to use properties). But i guess it is not other way to encode the needed info.

PR #1744, opened 2021-04-17
Delete an unused import

ndmitchell:
Not sure why we don't have a warning check for unused imports? But this one seems to have slipped through.


pepeiborra (approved):

pepeiborra:
This may not be the only redundant import.

ndmitchell:
It was the only one I saw in Ghcide, as the scrawl was flying past. There were a few in other plugins. I'm not a fan of `-Wall`, but I do enable this warning in all my projects, since it can help with build times.

PR #1751, opened 2021-04-18
Fix reduction depth

### LINKED ISSUE #1743: `Reduction stack overflow` error loading ghcide code base
### Your environment

```sh
haskell-language-server version: 1.1.0.0 (GHC: 8.10.4) (PATH: /home/berberman/.cabal/store/ghc-8.10.4/haskell-language-server-1.1.0.0-e-haskell-language-server-eb05d8a7d7a91e13040e70afeca109451ba08bb704fbbea7e3581a1a6a883d57/bin/haskell-language-server)
cabal:          3.4.0.0
stack:          Not found
ghc:            8.10.4
```
Which OS do you use:
NixOS
Which lsp-client do you use:
VSCode
ghcide

### Steps to reproduce
Open `Development.IDE.GHC.ExactPrint` module

### Expected behaviour
The module can be loaded

### Actual behaviour
Typecheck failed at line 298:

```
• Reduction stack overflow; size = 201
  When simplifying the following type: Outputable SrcSpan
  Use -freduction-depth=0 to disable this check
  (any upper bound you could choose might fail unpredictably with
   minor updates to GHC, so disabling the check is recommended if
   you're sure that type checking should terminate)
• In the expression: annotate dflags needs_space
  In the second argument of ‘($)’, namely
    ‘annotate dflags needs_space $ mk_parens val'’
  In a stmt of a 'do' block:
    (anns, val'') <- hoistTransform (either fail pure)
                       $ annotate dflags needs_space $ mk_parens val'
```

This function was added in #1721

pepeiborra:
Fixes #1743

### DIFF
--- a/ghcide/src/Development/IDE/GHC/ExactPrint.hs
+++ b/ghcide/src/Development/IDE/GHC/ExactPrint.hs
@@ -292,8 +292,8 @@ graftExprWithM dst trans = Graft $ \dflags a -> do
                         case mval of
                             Just val' -> do
                                 (anns, val'') <-
-                                    hoistTransform (either Fail.fail pure) $
-                                        annotate dflags needs_space $ mk_parens val'
+                                    hoistTransform (either Fail.fail pure)
+                                        (annotate @(HsExpr GhcPs) dflags needs_space (mk_parens val'))
                                 modifyAnnsT $ mappend anns
                                 pure val''
                             Nothing -> pure val


ghcide/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -292,8 +296,8 @@ graftExprWithM dst trans = Graft $ \dflags a -> do
                         case mval of
                             Just val' -> do
                                 (anns, val'') <-
-                                    hoistTransform (either Fail.fail pure) $
-                                        annotate dflags needs_space $ mk_parens val'
+                                    hoistTransform (either Fail.fail pure)
+                                        (annotate @(HsExpr GhcPs) dflags needs_space (mk_parens val'))
```
pepeiborra, line 300: A type annotation is needed here to prevent the type checker from diverging, but no idea why

isovector: Weird. Sounds like a bug in GHC to me; since `mval` is very definitely an `LHsExpr GhcPs`.

isovector (approved):

PR #1753, opened 2021-04-18
Wrap the Shake functions with newtypes

ndmitchell:
To ensure that we really are using our own types, and our own library, it's useful to newtype wrap the Action, ShakeOptions and Rules types. That means that if we accidentally use Shake, it's a type error. It's also step 1 towards replacing Shake, but even if we don't, it makes the abstraction more robust.

In order to further enhance the abstraction, I've changed `addBuiltinRule` for `addRule` which doesn't do the linting and tracking stuff, and simplified the "extra" feature in Shake to only allow a single value, as that's all we need. Fairly minor changes, but cut down on the complexity of the interface to Shake, where we're only using a subset.

There's a potential performance cost to newtype wrapping when it's a list, but if we had to we could switch to the newtype coercion stuff, and these aren't performance sensitive bits, so it's will be negligible. (I ran a benchmark, and saw no change beyond random variation.)

I haven't tried to wrap all Shake types, as that doesn't really give any benefits, but that might be a reasonable end goal in the future.

### DIFF
--- a/hls-graph/src/Development/IDE/Graph/Internal/Options.hs
+++ b/hls-graph/src/Development/IDE/Graph/Internal/Options.hs
@@ -1,0 +1,33 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module Development.IDE.Graph.Internal.Options where
+
+import qualified Development.Shake as Shake
+import qualified Data.HashMap.Strict as Map
+import Development.IDE.Graph.Internal.Action
+import Development.IDE.Graph.Internal.Rules
+import Data.Dynamic
+import Data.Typeable
+
+data ShakeOptions = ShakeOptions {
+    shakeThreads :: Int,
+    shakeFiles :: FilePath,
+    shakeExtra :: Map.HashMap TypeRep Dynamic
+    }
+
+shakeOptions :: ShakeOptions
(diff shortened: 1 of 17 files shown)


hls-graph/src/Development/IDE/Graph/Internal/Options.hs
```diff
@@ -0,0 +1,36 @@
...
+import qualified Data.HashMap.Strict as Map
+import Development.IDE.Graph.Internal.Action
+import Development.IDE.Graph.Internal.Rules
+import Data.Dynamic
+
+data ShakeOptions = ShakeOptions {
+    shakeThreads :: Int,
+    shakeFiles :: FilePath,
+    shakeExtra :: Maybe Dynamic
+    }
```
pepeiborra, line 15: Users of ghcide as a library will want to customise shake options more extensively. 
At the very least, could expose `shakeAllowRedefineRules` too?

ndmitchell: Sure. I went through the full list, but only found two others seem seem remotely plausible to customise - shakeAllowRedefineRules and shakeTimings. So I've added those.

pepeiborra (approved):

Issue #796, opened 2020-03-16
Consider using Reflex backend


ndmitchell:
@mpickering wrote a port to a Reflex backend: https://mpickering.github.io/posts/2020-03-16-ghcide-reflex.html. See https://github.com/mpickering/ghcide/tree/reflex

I'm sceptical that the 1s slowdown on hover is due to Shake overheads, so it would be good to get profile reports showing that. I suspect there's something else we're doing that is taking the 1s, and that should probably be fixed - but the profile will tell us all.

In the longer run, we all knew a well-tested FRP would be a better fit. @mpickering, is this a path towards Ghcide 1.1, or an experiment that we should use to guide future FRP/Ghcide evolution so they collide at a future date?

pepeiborra:
Now that we have reactive builds via reverse dependencies in Shake, is there any reason to consider Reflex?

jneira:
will close soon if nobody disagree

jneira:
closing, @ndmitchell feel free to reopen if you think the suggestion is still sensible after last changes in ghcide

PR #1759, opened 2021-04-19
Rewrite hls-graph to not use the Shake code

ndmitchell:
This PR reimplements the key algorithms from Shake directly. Most of the tests still work, but there are some failures relating to progress messages. Unfortunately, those failures also stop the benchmark from working, and thus I have no idea of the relative performance. In order to simplify relative to Shake, and make it more suitable for HLS:

* No serialisation logic at all.
* A lot less focus on error handling and fast exception propagation, since we never have exceptions.
* No continuation monad or thread pool, everything is separate threads (managed by async) and direct ReaderT IO.
* The Ids and Intern modules are lifted from Shake unchanged.

All the interesting code is in the Database module. Everything else is just wrapping and implementing required type classes on the underlying types.

Things that don't work:

* The benchmarks, which seem to have race conditions around progress messages.
* Most tests work, but the "fix syntax error" ones don't. Unfortunately, they also fail with race conditions around progress messages, but sometimes I won those races and they still failed.
* All profiling is stubbed out. If we consider that important, I think we can get it back. I think reusing the Shake progress reporting HTML is even feasible, with a bit of modularisation in Shake. It wouldn't be unreasonable for Shake to make that a separate package.

How we move forward is probably heavily influenced by what the benchmarks say. If this new Shake is either within the ballpark of same speed or faster, I suggest we move entirely to this new one, at which point things like @pepeiborra's reverse dependencies are way easier to integrate.

### DIFF
--- a/hls-graph/src/Development/IDE/Graph/Database.hs
+++ b/hls-graph/src/Development/IDE/Graph/Database.hs
@@ -1,18 +1,49 @@
 
 module Development.IDE.Graph.Database(
-    Shake.ShakeDatabase,
+    ShakeDatabase,
     shakeOpenDatabase,
     shakeRunDatabase,
-    Shake.shakeProfileDatabase,
+    shakeProfileDatabase,
     ) where
 
-import           Development.IDE.Graph.Internal.Action
-import           Development.IDE.Graph.Internal.Options
-import           Development.IDE.Graph.Internal.Rules
-import qualified Development.Shake.Database             as Shake
+import Development.IDE.Graph.Internal.Action
+import Development.IDE.Graph.Internal.Options
+import Development.IDE.Graph.Internal.Rules
+import Development.IDE.Graph.Internal.Types
+import Data.Maybe
+import Data.Dynamic
+import Development.IDE.Graph.Internal.Database
+import GHC.Conc
+import Control.Concurrent.Extra
 
-shakeOpenDatabase :: ShakeOptions -> Rules () -> IO (IO Shake.ShakeDatabase, IO ())
-shakeOpenDatabase a b = Shake.shakeOpenDatabase (fromShakeOptions a) (fromRules b)
 
-shakeRunDatabase :: Shake.ShakeDatabase -> [Action a] -> IO ([a], [IO ()])
-shakeRunDatabase a b = Shake.shakeRunDatabase a (map fromAction b)
+data ShakeDatabase = ShakeDatabase !Int !Int [Action ()] Database
+
+-- Placeholder to be the 'extra' if the user doesn't set it
+data NonExportedType = NonExportedType
+
+shakeOpenDatabase :: ShakeOptions -> Rules () -> IO (IO ShakeDatabase, IO ())
+shakeOpenDatabase opts rules = pure (shakeNewDatabase opts rules, pure ())
+
+shakeNewDatabase :: ShakeOptions -> Rules () -> IO ShakeDatabase
+shakeNewDatabase opts rules = do
+    let extra = fromMaybe (toDyn NonExportedType) $ shakeExtra opts
+    (theRules, actions) <- runRules extra rules
+    db <- newDatabase extra theRules
+    let threads = shakeThreads opts
+    threads <- if threads /= 0 then pure threads else getNumProcessors
+    pure $ ShakeDatabase threads (length actions) actions db
+
+shakeRunDatabase :: ShakeDatabase -> [Action a] -> IO ([a], [IO ()])
+shakeRunDatabase (ShakeDatabase threads lenAs1 as1 db) as2 = withNumCapabilities threads $ do
+    incDatabase db
+    as <- fmap (drop lenAs1) $ runActions db $ map unvoid as1 ++ as2
+    return (as, [])
+
+-- Only valid if we never pull on the results, which we don't
+unvoid :: Functor m => m () -> m a
+unvoid = fmap undefined
+
+-- Noop
+shakeProfileDatabase :: ShakeDatabase -> FilePath -> IO ()
+shakeProfileDatabase _ file = writeFile file ""
--- a/hls-graph/src/Development/IDE/Graph/Internal/Action.hs
+++ b/hls-graph/src/Development/IDE/Graph/Internal/Action.hs
@@ -1,38 +1,100 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Development.IDE.Graph.Internal.Action where
 
-import           Control.Exception
-import           Control.Monad.Fail
-import           Control.Monad.IO.Class
-import qualified Development.Shake         as Shake
-import           Development.Shake.Classes
-import qualified Development.Shake.Rule    as Shake
+import qualified Development.Shake as Shake
+import Development.Shake.Classes
+import Control.Exception
+import Control.Concurrent.Async
+import System.Exit
+import Control.Monad.IO.Class
+import Control.Monad.Trans.Reader
+import Data.IORef
+import Development.IDE.Graph.Internal.Database
+import Development.IDE.Graph.Internal.Types
+import Control.Monad.Extra
+import Control.Monad.Trans.Class
+import Control.Concurrent.Async
 
-newtype Action a = Action {fromAction :: Shake.Action a}
-    deriving (Monad, Applicative, Functor, MonadIO, MonadFail)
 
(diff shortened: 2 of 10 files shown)


jneira:
Sorry if it is not related at all but would be #796 relevant? this could enable the change suggested there?

ndmitchell:
The work already merged enables #796. This would be going in a different direction with our own implementation

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -0,0 +1,107 @@
...
+newtype Value = Value Dynamic
+
+data Database = Database {
+    databaseExtra :: Dynamic,
+    databaseRules :: TheRules,
+    databaseStep :: !(IORef Step),
+    -- Hold the lock while mutating Ids/Values
+    databaseLock :: !Lock,
+    databaseIds :: !(IORef (Intern Key)),
+    databaseValues :: !(Ids (Key, Status))
```
pepeiborra, line 83:
```suggestion
    -- TODO (someday) Replace the 3 fields below by a concurrent (mutable) Hashtable 
    -- Hold the lock while mutating Ids/Values
    databaseLock :: !Lock,
    databaseIds :: !(IORef (Intern Key)),
    databaseValues :: !(Ids (Key, Status))
```

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -0,0 +1,107 @@
...
+    hashWithSalt i (Key x) = hashWithSalt i (typeOf x, x)
+
+instance Show Key where
+    show (Key x) = show x
+
+newtype Value = Value Dynamic
+
+data Database = Database {
+    databaseExtra :: Dynamic,
+    databaseRules :: TheRules,
```
pepeiborra, line 78: are these fields lazy on purpose?

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -0,0 +1,107 @@
...
+    -- Hold the lock while mutating Ids/Values
+    databaseLock :: !Lock,
+    databaseIds :: !(IORef (Intern Key)),
+    databaseValues :: !(Ids (Key, Status))
+    }
+
+data Status
+    = Clean Result
+    | Dirty (Maybe Result)
+    | Running (IO Result) (Maybe Result)
```
pepeiborra, line 89:
```suggestion
    = Clean !Result
    | Dirty !(Maybe Result)
    | Running !(IO Result) !(Maybe Result)
```

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -0,0 +1,107 @@
...
+
+data Result = Result {
+    resultValue :: !Value,
+    resultBuilt :: !Step,
+    resultChanged :: !Step,
+    resultDeps :: !(Maybe [Id]), -- Nothing = alwaysRerun
+    resultData :: BS.ByteString
```
pepeiborra, line 96:
```suggestion
    resultData :: !BS.ByteString
```

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -0,0 +1,144 @@
...
+        Ids.insert databaseValues id (key, Clean res)
+    pure res
+
+data Box a = Box {fromBox :: a}
+
+splitIO :: IO a -> (IO (), a)
+splitIO act = do
+    let act2 = Box <$> act
+    let res = unsafePerformIO act2
+    (void $ evaluate res, fromBox res)
```
pepeiborra, line 144: I'm not familiar with this trick. What is the `Box` for?

hls-graph/src/Development/IDE/Graph/Internal/Action.hs
```diff
@@ -1,38 +1,100 @@
...
+actionCatch a b = do
+    v <- Action ask
+    Action $ lift $ catchJust f (runReaderT (fromAction a) v) (\x -> runReaderT (fromAction (b x)) v)
+    where
+        -- Catch only catches exceptions that were caused by this code, not those that
+        -- are a result of program termination
+        f e | isAsyncException e = Nothing
+            | otherwise = fromException e
 
 actionBracket :: IO a -> (a -> IO b) -> (a -> Action c) -> Action c
```
pepeiborra, line 74:
```suggestion
-- TODO Generalize type signatures and implement exception classes
actionBracket :: IO a -> (a -> IO b) -> (a -> Action c) -> Action c
```

hls-graph/src/Development/IDE/Graph/Database.hs
```diff
@@ -1,18 +1,49 @@
...
+shakeNewDatabase opts rules = do
+    let extra = fromMaybe (toDyn NonExportedType) $ shakeExtra opts
+    (theRules, actions) <- runRules extra rules
+    db <- newDatabase extra theRules
+    let threads = shakeThreads opts
+    threads <- if threads /= 0 then pure threads else getNumProcessors
+    pure $ ShakeDatabase threads (length actions) actions db
+
+shakeRunDatabase :: ShakeDatabase -> [Action a] -> IO ([a], [IO ()])
+shakeRunDatabase (ShakeDatabase threads lenAs1 as1 db) as2 = withNumCapabilities threads $ do
```
pepeiborra, line 38:
```suggestion
-- TODO move capability bits out of hls-graph
shakeRunDatabase (ShakeDatabase threads lenAs1 as1 db) as2 = withNumCapabilities threads $ do
```

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -0,0 +1,107 @@
...
+
+---------------------------------------------------------------------
+-- ACTIONS
+
+newtype Action a = Action {fromAction :: ReaderT SAction IO a}
+    deriving (Monad, Applicative, Functor, MonadIO, MonadFail)
+
+data SAction = SAction {
+    actionDatabase :: !Database,
+    actionDeps :: !(IORef (Maybe [Id])) -- Nothing means always rerun
```
pepeiborra, line 53:
```suggestion
    actionDeps :: !(IORef (Maybe [Id])) -- ^ Nothing means always rerun
```

pepeiborra (commented):
Very impressive! I read through the code and was able to understand most of it. Planning to play with it this weekend and make it work with the benchmarks.

pepeiborra:
After rebasing on top of #1754 benchmarks are failing with out of range errors:
```
╰─>$ tail -n 20 ghcide/bench-results/unprofiled/Cabal-3.0.0.0/HEAD/completions.output.log
                "method": "workspace/didChangeWatchedFiles",
                "id": "globalFileWatches"
            }
        ]
    },
    "method": "client/registerCapability",
    "id": 0
}
ghcide: Ix{Int}.index: Index (43) out of range ((0,0))
[DEBUG] Set files of
--> {
    "jsonrpc": "2.0",
    "params": null,
    "method": "shutdown",
    "id": 0
}
fd:12: hPutBuf: resource vanished (Broken pipe)
name        | success | samples | startup | setup | userTime | delayedTime | totalTime
----------- | ------- | ------- | ------- | ----- | -------- | ----------- | ---------
completions | False   | 50      | 0.00s   | 0.00s | 0.00s    | 0.00s       | 0.00s
```

ndmitchell:
That's a different error to what I was seeing. I'll double check tomorrow morning that I didn't forget to send up any of my changes.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -0,0 +1,144 @@
...
+                        -- so the following unsafePerformIO isn't actually unsafe
+                        let (force, val) = splitIO act
+                        modifyIORef toForce (force:)
+                        pure val
+                    (key, Dirty s) -> do
+                        -- Important we don't lose any Async things we create
+                        act <- uninterruptibleMask $ \restore -> do
+                            -- the child actions should always be spawned unmasked
+                            -- or they can't be killed
+                            async <- async $ restore $ check db key id s
```
wz1000, line 97: On IRC, we were discussing if it was possible to make all the book-keeping contention free, by eliminating the locks and running in STM (and using a concurrent map like [stm-containers](https://hackage.haskell.org/package/stm-containers-1.2)?).

This bit seems to be the problematic part, the rest of the computation seem like it can easily be moved into STM. But it might just be possible to make something work using judicious use of `unsafePerformIO`, `unsafeIOToSTM` and the proper care. Do you think this would be possible?

pepeiborra:
Continued in https://github.com/haskell/haskell-language-server/pull/2060

PR #1760, opened 2021-04-19
Prevent accidental Cthulhu summons

isovector:
This PR makes Wingman pick more idiomatic identifiers when generating new variables.

Before, the situation was quite bad for complicated types:

2021-03-23-233539_1104x328_scrot

But now, Wingman uses a bunch of rules to find acceptable names. It categorizes variables by their _purpose_, correlating roughly with intended semantic meaning. Right now, that means we can find things like monadic actions, functions, and predicates. For each purpose, we give a preferred ordering of variable names, and Wingman uses the first one that is available.

For long type constructors, Wingman will attempt to split them by camel case, turning eg `HelloWorld` into `hw`. if that fails, it uses the first syllable of the constructor.

This is not yet a perfect solution to naming, which should constrain names based on the relationship. But it's a significant step forward.


plugins/hls-tactics-plugin/test/golden/GoldenListFmap.hs.expected
```diff
@@ -1,3 +1,3 @@
 fmapList :: (a -> b) -> [a] -> [b]
 fmapList _ [] = []
-fmapList fab (a : l_a3) = fab a : fmapList fab l_a3
+fmapList fab (a : as') = fab a : fmapList fab as'
```
isovector, line 3: This should be `as`, not `as'` :confused:

isovector: Ahh, `as` is the name being given the `[a]` parameter introduced by the lambda, which immediately gets split away. Fixing this is going to require a more global view of names than we currently have, so it wont happen today.

jneira (approved):
ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn

plugins/hls-tactics-plugin/test/golden/DestructCthulhu.hs.expected
```diff
@@ -0,0 +1,54 @@
...
+cthulhu (Fanout fp' fp_rcab') = _
+cthulhu (Left' fp') = _
+cthulhu (Right' fp') = _
+cthulhu (EitherOf fp' fp_rca'b') = _
+cthulhu (Fanin fp' fp_rca'b) = _
+cthulhu (LiftC cab) = _
+cthulhu Zero = _
+cthulhu (Plus fp' fp_rcab) = _
+cthulhu (Unleft fp') = _
+cthulhu (Unright fp') = _
```
Ailrun, line 54: Now it's sleeping well

plugins/hls-tactics-plugin/test/golden/DestructCthulhu.hs.expected
```diff
@@ -0,0 +1,54 @@
...
+cthulhu SwapE = _
+cthulhu Fst = _
+cthulhu Snd = _
+cthulhu InjectL = _
+cthulhu InjectR = _
+cthulhu Unify = _
+cthulhu (First fp') = _
+cthulhu (Second fp') = _
+cthulhu (Alongside fp' fp_rca'b') = _
+cthulhu (Fanout fp' fp_rcab') = _
```
Ailrun, line 45: This means if we have a constructor like `R'lyeh :: FreePro r c x y -> FreePro r c (Either a x) (Either b y) -> FreePro r c a b` (or more complex second argument) then we can still summon one, right?

isovector: The `_<blah>` bit only shows up if all of your type arguments are variables. I think we're still safe.

Issue #1749, opened 2021-04-18
Progress messages logic might have race conditions


ndmitchell:
I'm not familiar with the current progress code, so the below is somewhat guesswork and assumption - happy to understand why I'm wrong. There seem to be two invariants around progress messages:

1. During testing we expect progress messages to be sent. If we don't, the benchmarking fails.
2. The LSP spec requires that no progress messages be sent between initialise request/response, as per https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#initialize

As far as I can tell, the way we ensure invariant 1 is by forking a progress thread and not doing a sleep at the beginning. Unfortunately, if you make the operations fast enough or lose a bunch of races, it doesn't get a progress notification, and we get failures. I'm seeing a failure with the ghcide benchmarking stuff I think comes from this.

As far as I can tell, the way we ensure invariant 2 is by hoping the initialise gets handled super quickly, and that the timer gets killed. Again, that's using sleeps and races to ensure the property.

I've changed a lot of timing properties around the Shake graph, and am suddenly getting a lot of errors from progress stuff. While these might be due to my changes, my debugging suggests progress itself might be at fault.

pepeiborra:
Invariant 1 is indeed an assumption for many tests. It would make sense to have an entirely separate codepath for testing that doesn't spawn a separate thread and instead sends progress notifications directly. Let me look into that.

The proper way to enforce invariant 2 would be to make the request handler synchronous. We already handle notifications synchronously, for instance. I'll look into that too.

PR #1770, opened 2021-04-24
Synchronous progress reporting in tests

pepeiborra:
The WorkDoneBegin and WorkDoneEnd notifications are now sent synchronously for tests, to avoid timing issues reported in #1749

### DIFF
--- a/ghcide/src/Development/IDE/Core/ProgressReporting.hs
+++ b/ghcide/src/Development/IDE/Core/ProgressReporting.hs
@@ -1,0 +1,168 @@
+{-# LANGUAGE RankNTypes #-}
+module Development.IDE.Core.ProgressReporting
+  ( ProgressEvent(..)
+  , ProgressReporting(..)
+  , noProgressReporting
+  , delayedProgressReporting
+  -- utilities, reexported for use in Core.Shake
+  , mRunLspT
+  , mRunLspTCallback
+  )
+   where
+
+import           Control.Concurrent.Async
+import           Control.Concurrent.STM
+import           Control.Concurrent.Strict
+import           Control.Monad.Extra
+import           Control.Monad.IO.Class
+import qualified Control.Monad.STM              as STM
+import           Control.Monad.Trans.Class      (lift)
+import qualified Data.HashMap.Strict            as HMap
+import qualified Data.Text                      as T
+import           Data.Unique
+import           Development.IDE.GHC.Orphans    ()
+import           Development.IDE.Graph          hiding (ShakeValue)
+import           Development.IDE.Types.Location
+import           Development.IDE.Types.Options
+import qualified Language.LSP.Server            as LSP
+import           Language.LSP.Types
+import qualified Language.LSP.Types             as LSP
+import           System.Time.Extra
+import           UnliftIO.Exception             (bracket_)
+
+data ProgressEvent
+    = KickStarted
+    | KickCompleted
+
+data ProgressReporting  = ProgressReporting
+  { progressUpdate :: ProgressEvent -> IO ()
+  , inProgress     :: forall a. NormalizedFilePath -> Action a -> Action a
+  , progressStop   :: IO ()
+  }
+
+noProgressReporting :: IO ProgressReporting
+noProgressReporting = return $ ProgressReporting
+  { progressUpdate = const $ pure ()
+  , inProgress = const id
+  , progressStop   = pure ()
+  }
+
+delayedProgressReporting
+  :: Maybe (LSP.LanguageContextEnv c)
+  -> ProgressReportingStyle
+  -> IO ProgressReporting
+delayedProgressReporting lspEnv optProgressStyle = do
+    inProgressVar <- newVar (HMap.empty @NormalizedFilePath @Int)
+    mostRecentProgressEvent <- newTVarIO KickCompleted
+    progressAsync <- async $
+            progressThread optProgressStyle mostRecentProgressEvent inProgressVar
+    let progressUpdate = atomically . writeTVar mostRecentProgressEvent
+        progressStop   = cancel progressAsync
+        inProgress :: NormalizedFilePath -> Action a -> Action a
+        inProgress = withProgressVar inProgressVar
+    return ProgressReporting{..}
+    where
+        -- The progress thread is a state machine with two states:
+        --   1. Idle
+        --   2. Reporting a kick event
+        -- And two transitions, modelled by 'ProgressEvent':
+        --   1. KickCompleted - transitions from Reporting into Idle
+        --   2. KickStarted - transitions from Idle into Reporting
+        progressThread style mostRecentProgressEvent inProgress = progressLoopIdle
+          where
+            progressLoopIdle = do
+                atomically $ do
+                    v <- readTVar mostRecentProgressEvent
+                    case v of
+                        KickCompleted -> STM.retry
+                        KickStarted   -> return ()
+                asyncReporter <- async $ mRunLspT lspEnv lspShakeProgress
+                progressLoopReporting asyncReporter
+            progressLoopReporting asyncReporter = do
+                atomically $ do
+                    v <- readTVar mostRecentProgressEvent
+                    case v of
+                        KickStarted   -> STM.retry
+                        KickCompleted -> return ()
+                cancel asyncReporter
+                progressLoopIdle
+
+            lspShakeProgress :: LSP.LspM config ()
+            lspShakeProgress = do
+                -- first sleep a bit, so we only show progress messages if it's going to take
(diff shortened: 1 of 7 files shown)


ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -183,8 +183,8 @@ data ShakeExtras = ShakeExtras
     -- positions in a version of that document to positions in the latest version
     -- First mapping is delta from previous version and second one is an
     -- accumlation of all previous mappings.
-    ,inProgress :: Var (HMap.HashMap NormalizedFilePath Int)
-    -- ^ How many rules are running for each file
+    ,inProgress :: forall a . NormalizedFilePath -> Action a -> Action a
+    -- ^ Report progress for a rule
     ,progressUpdate :: ProgressEvent -> IO ()
```
ndmitchell, line 188: Did you consider just putting a whole ProgressReporting in ShakeExtras, so you can keep it a bit more abstract?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,229 @@
+{-# LANGUAGE RankNTypes #-}
+module Development.IDE.Core.ProgressReporting
```
ndmitchell, line 2: Great call to pull this out as a module

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,229 @@
...
+
+noProgressReporting :: IO ProgressReporting
+noProgressReporting = return $ ProgressReporting
+  { progressUpdate = const $ pure ()
+  , inProgress = const id
+  , progressStop   = pure ()
+  }
+
+-- | A 'ProgressReporting' that sends the WorkDone Begin and End notifications
+--   synchronously. Progress notifications are sent from a sampling thread.
```
ndmitchell, line 55: Perhaps a line of comment saying why we want this (namely testing), and the guarantee it providers (you will always get the messages)

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,229 @@
...
+                mRunLspT env $ stop u
+
+        inProgress file = actionBracket (f file succ) (const $ f file pred) . const
+        -- This function is deliberately eta-expanded to avoid space leaks.
+        -- Do not remove the eta-expansion without profiling a session with at
+        -- least 1000 modifications.
+        f file shift = atomicModifyIORef'_ inProgressVar $
+                HMap.insertWith (\_ x -> shift x) file (shift 0)
+
+        progressLoop :: Double -> LSP.LspM a ()
```
ndmitchell, line 81: Maybe use `Seconds` from the extra library here?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,229 @@
...
+        progressLoop prev = do
+            mbToken <- liftIO $ readIORef st
+            case mbToken of
+                Nothing ->
+                    liftIO (sleep sample) >> progressLoop 0
+                Just t -> do
+                    current <- liftIO $ readIORef inProgressVar
+                    prev <- progress style prev current t
+                    liftIO $ sleep sample
+                    progressLoop prev
```
ndmitchell, line 91: It might be clearer if you calculated the value inside the case, then had the `sleep` and `processLoop` at the end of the case, so its easier to see the control flow is just an uninterupted loop.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,229 @@
...
+  where
+    -- The progress thread is a state machine with two states:
+    --   1. Idle
+    --   2. Reporting a kick event
+    -- And two transitions, modelled by 'ProgressEvent':
+    --   1. KickCompleted - transitions from Reporting into Idle
+    --   2. KickStarted - transitions from Idle into Reporting
+    progressThread mostRecentProgressEvent inProgress = progressLoopIdle
+      where
+        progressLoopIdle = do
```
ndmitchell, line 126: I think this does what you are describing, but it's really hard to see with the retries and the running of the reporter async with a kill. Is there a more direct encoding? Like when we get a `progressUpdate Started` we sleep, then spawn the thread, and when we get `Completed` we kill it? I spent a long time trying to find the loop, before realising the loop isn't even in `progressLoop`, it's in `lspShakeProgress`.

ndmitchell (approved):

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,228 @@
...
+                    KickStarted   -> STM.retry
+                    KickCompleted -> return ()
+            cancel asyncReporter
+            progressLoopIdle
+
+    lspShakeProgress style inProgress = do
+        u <- liftIO newProgressToken
+
+        void $ LSP.sendRequest LSP.SWindowWorkDoneProgressCreate
+            LSP.WorkDoneProgressCreateParams { _token = u } $ const (pure ())
```
wz1000, line 151: According to the spec, we should wait for the response to this request before proceeding. This could be done by using `signalBarrier` instead of `const (pure ())` as the response callback.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: I'm confused too. There's actually two loops, outer one in the mutually recursive functions forming the state machine, and inner one in `lspShakeProgress`. 

I think the code works fine and I don't see how to make it less confusing, so I've just added more comments.

pepeiborra:
Thank you both for the careful review.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
ndmitchell: I was expecting the code to read something along the lines of:

```
state <- newVar Nothing
let progressUpdate KickCompleted =
        join $ modifyVar state $ \case
             Nothing -> pure (Nothing, pure ())
             Just thread -> pure (Nothing, cancel thread)
    progressUpdate KickStarted =
        modifyVar_ state $ \case
             Nothing -> Just <$> async (mRunLspT ...)
             Just thread -> pure $ Just thread
```

So you don't need the outer loop. You'd also make progressStop be the same as KickCompleted, simplifying things further.

pepeiborra: That code is pretty much equivalent to `directProgressReporting` introduced in this PR for tests. And this is a good things, because in hindsight it seems a terrible idea to have one implementation for testing and a different one for production. So the path is clear - just use `directProgressReporting` for both and remove the old version.

pepeiborra:
This PR introduces a couple of new bugs: 

1. I moved the sleep call in `lspShakeProgress` into the `loop`, so now it no longer delays the Begin message and it happens unconditionally. I believe our tests will not catch this.
2. I added a barrier to wait for the Begin response, but it doesn't look like lsp-test sends this response. It breaks a few tests, and if we use this in prod it could make the IDE less responsive unless done in a background thread.

Having one `ProgressReporting` for testing and another one for production is not really a good idea. I'll need to take another look, so moving to draft for now.

wz1000:
> I added a barrier to wait for the Begin response, but it doesn't look like lsp-test sends this response. It breaks a few tests, and if we use this in prod it could make the IDE less responsive unless done in a background thread.

I'm pretty sure I fixed this in https://github.com/haskell/lsp/commit/5fa33f89d06a5623a4fd06fd2c0fdd4dbbc9c32d#diff-8e90cf08428164fe5d7ef53f4d8e7900169c97415b3198f3040e75b96803e009

pepeiborra:
> > I added a barrier to wait for the Begin response, but it doesn't look like lsp-test sends this response. It breaks a few tests, and if we use this in prod it could make the IDE less responsive unless done in a background thread.
> 
> I'm pretty sure I fixed this in [haskell/lsp@5fa33f8#diff-8e90cf08428164fe5d7ef53f4d8e7900169c97415b3198f3040e75b96803e009](https://github.com/haskell/lsp/commit/5fa33f89d06a5623a4fd06fd2c0fdd4dbbc9c32d#diff-8e90cf08428164fe5d7ef53f4d8e7900169c97415b3198f3040e75b96803e009)

My bad, it's not lsp-test fault - it was a division by zero bug in the `progress` loop when `todo` is 0. This was likely the real issue identified in #1749 - I'll send a minimal fix in a separate PR

pepeiborra:
Replaced by #1770

PR #1771, opened 2021-04-24
Wingman: configurable auto search depth

### LINKED ISSUE #1355: Add a config option to control tactics' auto's gas
It would allow us to solve things like https://github.com/haskell/haskell-language-server/issues/562

isovector:
Fixes #1355

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/Types.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/Types.hs
@@ -80,6 +80,16 @@ data Config = Config
   { cfg_feature_set          :: FeatureSet
   , cfg_max_use_ctor_actions :: Int
   , cfg_timeout_seconds      :: Int
+  , cfg_auto_gas             :: Int
+  }
+  deriving (Eq, Ord, Show)
+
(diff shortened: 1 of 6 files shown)


plugins/hls-tactics-plugin/src/Wingman/Types.hs
```diff
@@ -80,6 +80,16 @@ data Config = Config
   { cfg_feature_set          :: FeatureSet
   , cfg_max_use_ctor_actions :: Int
   , cfg_timeout_seconds      :: Int
+  , cfg_auto_gas             :: Int
```
Ailrun, line 83: Out of curiosity: Why it's not "depth" instead of "gas"? I mean, even your description says that this is a kind of depth.

Ailrun (approved):
Great, thanks!

plugins/hls-tactics-plugin/src/Wingman/Types.hs
isovector: I don't know; this is just what @TOTBWF and I have been calling it!

PR #1772, opened 2021-04-24
Add codetriage badge

jneira:
* Add a badge to encourage issue triaging via https://www.codetriage.com/haskell/haskell-language-server
* It will send you a selection of random (but uniformly distributed on time) issues to take a look on


Ailrun (approved):
Great, thank you for adding this service.

PR #1776, opened 2021-04-26
Tactic metaprogramming

isovector:
This PR exposes Wingman's underlying tactics machinery to end users. By introducing a `[wingman| |]` expression, users can write custom tactics metaprograms inside, and then use the `Run custom tactic` code action to splice the resulting expression. The tactic state is exposed via a hover action, showing what's in scope and what the current goals are.

metaprogram


jneira:
This would supersed #524? or do you want to keep both open for now?

isovector:
> This would supersed #524? or do you want to keep both open for now?

It will, but that branch has a bunch of functionality that isn't going to land here, and I'd like to keep it open because otherwise I'll lose it :)

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -817,7 +850,13 @@ parseFileContents
        -> ExceptT [FileDiagnostic] IO ([FileDiagnostic], ParsedModule)
 parseFileContents env customPreprocessor filename ms = do
    let loc  = mkRealSrcLoc (mkFastString filename) 1 1
-       dflags = ms_hspp_opts ms
+       dflags' = ms_hspp_opts ms
+       dflags =
+         dflags'
+            { staticPlugins =
+                staticPlugins dflags' <> [wingmanMetaprogrammingPlugin]
```
isovector, line 857: @wz1000 is there some way I can hook into ghcide from Wingman to install this plugin, rather than doing it directly in ghcide?

wz1000: Add a new field to `PluginDescriptor` in `hls-plugin-api`.

isovector: cheers

pepeiborra (approved):

PR #1778, opened 2021-04-29
Replace the unsafe getmodtime with safe posix calls

pepeiborra:
We don't need the 2X faster but unsafe getmodtime anymore since GetModificationTime is no longer called O(N) times but only O(FOI) times, where N is the number of known targets and FOI is the number of files of interest.

The main problem with the internal implementation of `getModTime` is that it's not cross platform. The `unix` package implements support for more platforms: https://github.com/haskell/unix/blob/master/System/Posix/Files/Common.hsc#L344-L365

This change switches to the safer `unix` implementation, which provides nanosecond resolution in systems that support it.

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -102,6 +102,9 @@ import           Data.IORef.Extra                         (atomicModifyIORef_)
 import qualified Development.IDE.Plugin.HLS.GhcIde        as Ghcide
 import           Text.Regex.TDFA                          ((=~))
 import qualified Progress
+import Development.IDE.Core.FileStore (getModTime)
+import Control.Concurrent (threadDelay)
+import Text.Printf (printf)
 
 waitForProgressBegin :: Session ()
 waitForProgressBegin = skipManyTill anyMessage $ satisfyMaybe $ \case
@@ -5492,8 +5495,23 @@ unitTests = do
             actualOrder <- liftIO $ readIORef orderRef
 
             liftIO $ actualOrder @?= reverse [(1::Int)..20]
+     , testCase "timestamps have millisecond resolution" $ do
+         resolution_us <- findResolution_us 1
+         let msg = printf "Timestamps do not have millisecond resolution: %dus" resolution_us
+         assertBool msg (resolution_us <= 1000)
      , Progress.tests
      ]
+
+findResolution_us :: Int -> IO Int
+findResolution_us delay_us | delay_us >= 1000000 = error "Unable to compute timestamp resolution"
+findResolution_us delay_us = withTempFile $ \f -> withTempFile $ \f' -> do
+    writeFile f ""
+    threadDelay delay_us
+    writeFile f' ""
+    t <- getModTime f
+    t' <- getModTime f'
+    if t /= t' then return delay_us else findResolution_us (delay_us * 10)
+
 
 testIde :: IDE.Arguments -> Session () -> IO ()
 testIde arguments session = do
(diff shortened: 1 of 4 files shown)


ghcide/test/exe/Main.hs
```diff
@@ -5464,8 +5467,23 @@ unitTests = do
             actualOrder <- liftIO $ readIORef orderRef
 
             liftIO $ actualOrder @?= reverse [(1::Int)..20]
+     , testCase "timestamps have millisecond resolution" $ do
+         resolution_us <- findResolution_us 1
+         let msg = printf "Timestamps do not have millisecond resolution: %dus" resolution_us
+         assertBool msg (resolution_us <= 1000)
```
ndmitchell, line 5473: This is very much OS and file format specific. Docker containers and Mac machines often don't have this, even when using precise functions.

ndmitchell (commented):
Getting rid of custom unsafe code seems good. But given this is no longer a bottleneck, why not go the full way and use https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:getModificationTime on both Windows and Linux? For editors which don't support subscriptions, I guess this code path will still be very hot - but are we assuming those editors are going to have a bad time anyway? And I guess it doesn't include Vim/Emacs/VS Code?

ghcide/test/exe/Main.hs
pepeiborra: This is a test that whatever we are using, file timestamps can be read with a certain precision. The precision will be OS specific, but without millisecond precision I expect that there will be trouble - other tests will fail mysteriously.
The test passes in all the platforms so far, including my Mac.

ndmitchell: Fair enough - I guess this is a canary for the test suite, rather than a requirement. In Shake I often have to introduce file system modification resolution sleeps to get the tests to pass.

pepeiborra:
> Getting rid of custom unsafe code seems good. But given this is no longer a bottleneck, why not go the full way and use https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:getModificationTime on both Windows and Linux? For editors which don't support subscriptions, I guess this code path will still be very hot - but are we assuming those editors are going to have a bad time anyway? And I guess it doesn't include Vim/Emacs/VS Code?

I want to fix the unsafe code but I don't want to give up *all* the performance, specially in large projects with thousands of files, given that we still read timestamps at least once for every file (or multiple times if the client doesn't support file watches) and multiple times for interface files.

But to be fair I haven't benchmarked - I have no idea if the cost of the [conversions](https://hackage.haskell.org/package/time-1.11.1.2/docs/src/Data.Time.Clock.POSIX.html#posixSecondsToUTCTime), one call to `divMod`, is that expensive.

ndmitchell:
The code in Shake has been well tested and doesn't use any C - https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Internal/FileInfo.hs. Any reason we're not using that?

pepeiborra:
> The code in Shake has been well tested and doesn't use any C - https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Internal/FileInfo.hs. Any reason we're not using that?

It does use `foreign import` calls on the Windows path, and the Unix path is essentially the same as this PR with additional conversions that we don't need

pepeiborra:
@ndmitchell do you have any other comments?

wz1000 (approved):

PR #1779, opened 2021-04-29
Drop the nix pre-commit-hook temporarily

pepeiborra:
The pre-commit-hook wasn't pinned, so it has broken. Let's unblock the CI until @Ailrun has a chance to look at it.


pepeiborra:
Replaced by #1780

PR #1784, opened 2021-05-02
Progress reporting improvements

### LINKED ISSUE #1749: Progress messages logic might have race conditions

pepeiborra:
Rework of #1770. Summary of changes:

1. progress reporting logic extracted into its own module
2. Removed one layer of async handling, reducing the chances of race conditions
3. Correctly implement the LSP spec for `window/workDoneProgress/create` by waiting until the client has responded
4. Fix bugs: division by zero (when `todo` is 0) and handling when `NoProgress` 
5. Performance: the reporting loop is now O(1) instead of O(N) in the number of files

I think this is the best we can do to close #1749 while sharing the same logic for both tests and app code

### DIFF
--- a/ghcide/src/Development/IDE/Core/ProgressReporting.hs
+++ b/ghcide/src/Development/IDE/Core/ProgressReporting.hs
@@ -1,0 +1,168 @@
+{-# LANGUAGE RankNTypes #-}
+module Development.IDE.Core.ProgressReporting
+  ( ProgressEvent(..)
+  , ProgressReporting(..)
+  , noProgressReporting
+  , delayedProgressReporting
+  -- utilities, reexported for use in Core.Shake
+  , mRunLspT
+  , mRunLspTCallback
+  )
+   where
+
+import           Control.Concurrent.Async
+import           Control.Concurrent.STM
+import           Control.Concurrent.Strict
+import           Control.Monad.Extra
+import           Control.Monad.IO.Class
+import qualified Control.Monad.STM              as STM
+import           Control.Monad.Trans.Class      (lift)
+import qualified Data.HashMap.Strict            as HMap
+import qualified Data.Text                      as T
+import           Data.Unique
+import           Development.IDE.GHC.Orphans    ()
+import           Development.IDE.Graph          hiding (ShakeValue)
+import           Development.IDE.Types.Location
+import           Development.IDE.Types.Options
+import qualified Language.LSP.Server            as LSP
+import           Language.LSP.Types
+import qualified Language.LSP.Types             as LSP
+import           System.Time.Extra
+import           UnliftIO.Exception             (bracket_)
+
+data ProgressEvent
+    = KickStarted
+    | KickCompleted
+
+data ProgressReporting  = ProgressReporting
+  { progressUpdate :: ProgressEvent -> IO ()
+  , inProgress     :: forall a. NormalizedFilePath -> Action a -> Action a
+  , progressStop   :: IO ()
+  }
+
+noProgressReporting :: IO ProgressReporting
+noProgressReporting = return $ ProgressReporting
+  { progressUpdate = const $ pure ()
+  , inProgress = const id
+  , progressStop   = pure ()
+  }
+
+delayedProgressReporting
+  :: Maybe (LSP.LanguageContextEnv c)
+  -> ProgressReportingStyle
+  -> IO ProgressReporting
+delayedProgressReporting lspEnv optProgressStyle = do
+    inProgressVar <- newVar (HMap.empty @NormalizedFilePath @Int)
+    mostRecentProgressEvent <- newTVarIO KickCompleted
+    progressAsync <- async $
+            progressThread optProgressStyle mostRecentProgressEvent inProgressVar
+    let progressUpdate = atomically . writeTVar mostRecentProgressEvent
+        progressStop   = cancel progressAsync
+        inProgress :: NormalizedFilePath -> Action a -> Action a
+        inProgress = withProgressVar inProgressVar
+    return ProgressReporting{..}
+    where
+        -- The progress thread is a state machine with two states:
+        --   1. Idle
+        --   2. Reporting a kick event
+        -- And two transitions, modelled by 'ProgressEvent':
+        --   1. KickCompleted - transitions from Reporting into Idle
+        --   2. KickStarted - transitions from Idle into Reporting
+        progressThread style mostRecentProgressEvent inProgress = progressLoopIdle
+          where
+            progressLoopIdle = do
+                atomically $ do
+                    v <- readTVar mostRecentProgressEvent
+                    case v of
+                        KickCompleted -> STM.retry
+                        KickStarted   -> return ()
+                asyncReporter <- async $ mRunLspT lspEnv lspShakeProgress
+                progressLoopReporting asyncReporter
+            progressLoopReporting asyncReporter = do
+                atomically $ do
+                    v <- readTVar mostRecentProgressEvent
+                    case v of
+                        KickStarted   -> STM.retry
+                        KickCompleted -> return ()
+                cancel asyncReporter
+                progressLoopIdle
+
+            lspShakeProgress :: LSP.LspM config ()
+            lspShakeProgress = do
+                -- first sleep a bit, so we only show progress messages if it's going to take
+                -- a "noticable amount of time" (we often expect a thread kill to arrive before the sleep finishes)
+                liftIO $ sleep 0.1
+                u <- ProgressTextToken . T.pack . show . hashUnique <$> liftIO newUnique
+
+                void $ LSP.sendRequest LSP.SWindowWorkDoneProgressCreate
+                    LSP.WorkDoneProgressCreateParams { _token = u } $ const (pure ())
+
+                bracket_
+                  (start u)
+                  (stop u)
+                  (loop u 0)
+                where
+                    start id = LSP.sendNotification LSP.SProgress $
+                        LSP.ProgressParams
+                            { _token = id
+                            , _value = LSP.Begin $ WorkDoneProgressBeginParams
+                              { _title = "Processing"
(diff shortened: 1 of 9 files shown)


hls-test-utils/src/Test/Hls.hs
```diff
@@ -134,3 +137,15 @@ runSessionWithServer' plugin conf sconf caps root s = withLock lock $ keepCurren
...
       putStrLn $ "Finishing canceling (took " <> showDuration t <> "s)"
   pure x
+
+-- | Wait for all progress to be done
+-- Needs at least one progress done notification to return
+waitForProgressDone :: Session ()
+waitForProgressDone = loop
+  where
+    loop = do
+      ~() <- skipManyTill anyMessage $ satisfyMaybe $ \case
```
ndmitchell, line 147: What does `~()` mean? A lazy pattern match on `()`? How is that different from just ignoring the result?

hls-test-utils/src/Test/Hls.hs
```diff
@@ -134,3 +137,15 @@ runSessionWithServer' plugin conf sconf caps root s = withLock lock $ keepCurren
       (t, _) <- duration $ cancel server
       putStrLn $ "Finishing canceling (took " <> showDuration t <> "s)"
   pure x
+
+-- | Wait for all progress to be done
+-- Needs at least one progress done notification to return
+waitForProgressDone :: Session ()
+waitForProgressDone = loop
+  where
+    loop = do
```
ndmitchell, line 146: `whileM` from extra can be used to do this - you just return True to keep going.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,176 @@
...
+  :: Seconds  -- ^ Grace period before starting
+  -> Seconds  -- ^ sampling delay
+  -> Maybe (LSP.LanguageContextEnv c)
+  -> ProgressReportingStyle
+  -> IO ProgressReporting
+delayedProgressReporting before after lspEnv optProgressStyle = do
+    inProgressVar <- newVar $ InProgress 0 0 mempty
+    progressState <- newVar NotStarted
+    let progressUpdate event = modifyVar_ progressState $ updateState (mRunLspT lspEnv $ lspShakeProgress inProgressVar) (Event event)
+        progressStop   = modifyVar_ progressState $ updateState (mRunLspT lspEnv $ lspShakeProgress inProgressVar) StopProgress
```
ndmitchell, line 100: Worth defining something that abstracts over most of these lines? So:

````
let f = modifyVar_ progressState $ updateState (mRunLspT lspEnv $ lspShakeProgress inProgressVar)
```

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,176 @@
...
+                stop id = LSP.sendNotification LSP.SProgress
+                    LSP.ProgressParams
+                        { _token = id
+                        , _value = LSP.End WorkDoneProgressEndParams
+                          { _message = Nothing
+                          }
+                        }
+                loop id prev = do
+                    InProgress{..} <- liftIO $ readVar inProgress
+                    if todo == 0 then loop id 0 else do
```
ndmitchell, line 137: If we hit loop id 0, then we spin at 100% CPU. That path seems like it should still have a sleep on it.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,176 @@
...
+                        let next = 100 * fromIntegral done / fromIntegral todo
+                        liftIO $ sleep after
+                        when (optProgressStyle /= NoProgress && next /= prev) $
```
ndmitchell, line 140: What's the point of running this loop at all if progress style is NoProgress?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,176 @@
...
+        lspShakeProgress inProgress = do
+            -- first sleep a bit, so we only show progress messages if it's going to take
+            -- a "noticable amount of time" (we often expect a thread kill to arrive before the sleep finishes)
+            liftIO $ sleep before
+            u <- ProgressTextToken . T.pack . show . hashUnique <$> liftIO newUnique
+
+            b <- liftIO newBarrier
+            void $ LSP.sendRequest LSP.SWindowWorkDoneProgressCreate
+                LSP.WorkDoneProgressCreateParams { _token = u } $ liftIO . signalBarrier b
+            ready <- liftIO $ waitBarrier b
```
ndmitchell, line 114: If we send a WorkDoneProgressCreate, do we need to clean it up ever, or does it just expire on its own?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -487,23 +479,23 @@ shakeOpen lspEnv defaultConfig logger debouncer
...
         indexCompleted <- newTVarIO 0
         indexProgressToken <- newVar Nothing
         let hiedbWriter = HieDbWriter{..}
-        progressAsync <- async $
-            when reportProgress $
-                progressThread optProgressStyle mostRecentProgressEvent inProgress
         exportsMap <- newVar mempty
 
+        progress <- do
+            let (before, after) = if testing then (0,0.1) else (0.1,0.1)
```
ndmitchell, line 490: At a delay of 0, we've definitely reduced the window of the race condition, but it's not gone entirely.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -0,0 +1,176 @@
...
+noProgressReporting = return $ ProgressReporting
+  { progressUpdate = const $ pure ()
+  , inProgress = const id
+  , progressStop   = pure ()
+  }
+
+-- | State used in 'delayedProgressReporting'
+data State
+    = NotStarted
+    | Completed
```
ndmitchell, line 54: I can't find a case where NotStarted and Completed don't have the same behaviour. Could we collapse them?

ndmitchell (approved):
I see a few updates to the test - are these fixing race conditions that this patch showed up?

This code is way clearer!

hls-test-utils/src/Test/Hls.hs
pepeiborra: I have no idea - it's part of the code that was already there. I can't think of any other meaning

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: Good tip, thanks

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: Agreed - but it's the best I can think of without introducing a separate code path for tests

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: As far as  I know it doesn't require any clean up

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: Good point

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: Good catch!

hls-test-utils/src/Test/Hls.hs
pepeiborra: The codebase consistenly uses recursive functions for looping this way, so switching to `whileM` would be inconsistent.

ndmitchell: Fair enough

hls-test-utils/src/Test/Hls.hs
ndmitchell: Maybe it enforces the type in a way that used to be ambiguous? I'd be OK with `_ <-` or `() <-`.

ghcide/src/Development/IDE/Core/Shake.hs
ndmitchell: That seems pretty reasonable. Could we leave a comment in the code here that for testing we'd like to guarantee always sending messages, but instead we do a sleep of 0 and a dose of hope?

pepeiborra:
> I see a few updates to the test - are these fixing race conditions that this patch showed up?

I fixed a bunch of tests in the previous version of this PR and decided to preserve the fixes. Not necessarily race conditions: often tests rely on sketchy invariants, e.g. a test that waits for *three* logging notifications before checking if the build session is restarted is going to break for reasons unrelated to the property it's checking.

PR #1366, opened 2021-02-14
Lose the ghc-lib flag

pepeiborra:
@cocreature confirmed that DA are not depending on this anymore and we are free to drop it. 

Closes #784

PR #1785, opened 2021-05-02
clean up ghc-api pragmas

pepeiborra:
We no longer depend on ghc-lib so it's cleanup time


anka-213:
Oh, we don't? That'll make a huge difference for this PR https://github.com/haskell/haskell-language-server/pull/1649.

What was it used for before and what has changed?

pepeiborra:
> Oh, we don't? Then I should probably also remove all the new references to it in this PR #1649.
> 
> What was it used for before and what has changed?

We dropped the dependency in https://github.com/haskell/haskell-language-server/pull/1366

anka-213:
Ah, thanks!

ndmitchell (approved):
Great idea

plugins/hls-explicit-imports-plugin/hls-explicit-imports-plugin.cabal
```diff
@@ -9,7 +9,6 @@ maintainer:         pepeiborra@gmail.com
 category:           Development
 build-type:         Simple
 extra-source-files:
-  include/ghc-api-version.h
```
Ailrun, line 12: It looks like actual file is not deleted?

Issue #1725, opened 2021-04-14
Inconsistent ghcide bounds in plugins


berberman:
In #1722, I bumped ghcide to 1.2.0.2, and changed ghcide bounds in plugins to `^>=1.2.0.2`. But at that time I didn't realize that @pepeiborra had uploaded `hls-retrie-plugin-1.0.0.2`, `hls-explicit-imports-plugin-1.0.0.2` , and `hls-brittany-plugin-1.0.0.1` to hackage, whereas I made changes in cabal file without bumping their versions. That said we need revise these three plugins in hackage to make the bounds of ghcide in them consistent with https://github.com/haskell/haskell-language-server/tree/1.1.0.

pepeiborra:
Hey @berberman I am sorry for the early upload, I thought that the release was done so I uploaded my plugins. 

I have given you powers in Hackage to all my plugins in case you need to make a revision or upload a new version. But I confess I don't understand the problem. The plugins are compatible with ghcide `>= 1.2` so I would think that the current bounds are fine according to the PVP. 

I would say that the problem here is the `^>=1.2.0.2` bound in HEAD. What does that even mean? I know what `>=1.2.0.2` mean, but I am not sure about `^>=1.2.0.2`. I think the bound should be `>1.2`.

berberman:
`ghcide ^>=1.2.0.2` is semantically identical to `ghcide >=1.2.0.2 && <1.3`, which says "we know that ghcide 1.2.0.2 is compatible." If we obey PVP, then we will know that it is compatible till the 1.3 version, indicating that the lower bound can be relexed to  `>=1.2`, i.e. `ghcide ^>=1.2`. 

Yeah, the bounds in Hackage now are the relexed ones, which in fact are more reasonable according to the PVP. My bad -- I shouldn't bump up the lower bounds. But for now, I think we'd better synchronize those in Hackage with in HEAD by revisions.

pepeiborra:
Thanks for explaining `^>=` to me! I always struggle with those and the Cabal docs are not very clear.

PR #1787, opened 2021-05-02
Loosen dependency bounds

berberman:
Since we've decided to stick to the PVP, we are able to relax the lower bounds of `hls-plugin-api`, `ghcide`, and `hls-test-utils`. We should also bp this PR to [1.1.0-hackage](https://github.com/haskell/haskell-language-server/tree/1.1.0-hackage), the branch where we are going to upload hls-1.1.0 to Hackage, so that #1725 can be closed legitimately without doing Hackage revisions.


pepeiborra (approved):
I hope this works. I wish we had some way of testing these changes in CI

berberman:

PR #1790, opened 2021-05-04
Fix backport conflict of refine-import plugin

Ailrun:
Fix the mistake from #1788


jneira (approved):
thanks for the fix

berberman:
Thank you!

PR #1792, opened 2021-05-04
Replace faulty signature test

kderme:
Before https://github.com/haskell/haskell-language-server/pull/1578 signatures in multiple lines didn't create any suggestions. The deleted test intended to showcase this, however the actual reason it didn't create any suggestion was that it had a parser error: note that `=> a -> Bool` starts at the begining of the line. 

After https://github.com/haskell/haskell-language-server/pull/1578 multiline signatures create suggestions, so new tests are added instead.


jneira (approved):
Good catch, many thanks for fixing tests

Issue #1774, opened 2021-04-24
Hackage release of 1.1.0


jneira:
Hi, afaik the last hls version is not in hackage yet. Some users (particularly nix ones) can access it yet.
Iirc some version bumps were done in the github release and i hope the hackahe one would be more straightforward thanks to that work already done
//cc @berberman @pepeiborra

berberman:
Do we have any meta changes that need to be back ported to 1.1.0-hackage? If no, I guess we are good to uoload all of the rest packages to Hackage now.

berberman:
Packages we need to upload, based on [1.1.0-hackage](https://github.com/haskell/haskell-language-server/tree/1.1.0-hackage):

* [x] `hls-plugin-api` - [1.1.0.0](https://hackage.haskell.org/package/hls-plugin-api-1.1.0.0)
* [x] `hls-test-utils` - [1.0.0.0](https://hackage.haskell.org/package/hls-test-utils-1.0.0.0)
* [x] `ghcide` - [1.2.0.2](https://hackage.haskell.org/package/ghcide-1.2.0.2)
* [x] `hls-brittany-plugin` - [1.0.0.1](https://hackage.haskell.org/package/hls-brittany-plugin-1.0.0.1)
* [ ] `hls-class-plugin` - 1.0.0.1
* [ ] `hls-eval-plugin` - 1.1.0.0
* [x] `hls-explicit-imports-plugin` - [1.0.0.1](https://hackage.haskell.org/package/hls-explicit-imports-plugin-1.0.0.1)
* [ ] `hls-haddock-comments-plugin` - 1.0.0.1
* [ ] `hls-hlint-plugin` - 1.0.0.2
* [x] `hls-retrie-plugin` - [1.0.0.2](https://hackage.haskell.org/package/hls-retrie-plugin-1.0.0.2)
* [ ] `hls-splice-plugin` - 1.0.0.2
* [ ] `hls-stylish-haskell-plugin` - 1.0.0.0
* [ ] `hls-tactics-plugin` - 1.1.0.0
* [ ] `haskell-language-server` - 1.1.0.0

jneira:
Uploading `hls-hlint-plugin`

Ailrun:
Uploaded `hls-class-plugin`, `hls-haddock-comments-plugin`, and `hls-tactic-plugin`.

jneira:
Uploaded `hls-hlint-plugin`

jneira:
[...]

PR #1795, opened 2021-05-04
Add upper bound to hlint

jneira:
* Without the upper bound, cabal picks hlint-3.* which brings ghc-lib-9.* causing compile errors (what this has not been catched by ci before?)
* This has to be backported to hackage-1.1.0 to make progress towards #1774


jneira:

Ailrun (approved):

PR #1805, opened 2021-05-07
Remove hls-ghc-x.y from install script and wrapper

### LINKED ISSUE #1728: Remove handling of ghc-x.y hls executables
(from a question in irc by arrowsvc)
* In several places the code is assuming a hls version compiled with ghc-x.y should (or could) work with all ghc-x.y.z runtime versions
  * hls-wrapper tries to use hls-x.y.z, hls-x.y and hls inthat order
  * the install script copies an hls version as hls-x.y
  * that assumption was inherited from hie but i think it worked in hls at some point (???)
* But you cant use that version with ghcide: `ghcide compiled against GHC x.y.z but currently using x.y.w ...`

So we should either make it work (again?) or remove its existence, cause it causes confusion
/cc @pepeiborra

berberman:
Closes #1728

As discussed in the issue, executables like `haskell-language-server-8.10` are really confusing - they could be the symlink of `haskell-language-server-8.10.2`, `haskell-language-server-8.10.3`, or even `haskell-language-server-8.10.4`, but those are definitely not interchangeable. So it's time to remove them. Now running `./cabal-hls-install hls-8.10.4` will create `haskell-language-server-wrapper`, `haskell-language-server`, and `haskell-language-server-8.10.4`.

### DIFF
--- a/install/src/Cabal.hs
+++ b/install/src/Cabal.hs
@@ -72,20 +72,17 @@ cabalInstallHls versionNumber args = do
     ++ installMethod
     ++ args
 
-  let minorVerExe = "haskell-language-server-" ++ versionNumber <.> exe
-      majorVerExe = "haskell-language-server-" ++ dropExtension versionNumber <.> exe
+  let verExe = "haskell-language-server-" ++ versionNumber <.> exe
 
   let copyCmd old new = if os == "mingw32"
                         then liftIO $ copyFile old new
                         else command [] "ln" ["-f", old, new]
-  copyCmd (localBin </> "haskell-language-server" <.> exe) (localBin </> minorVerExe)
-  copyCmd (localBin </> "haskell-language-server" <.> exe) (localBin </> majorVerExe)
+  copyCmd (localBin </> "haskell-language-server" <.> exe) (localBin </> verExe)
 
   printLine $   "Copied executables "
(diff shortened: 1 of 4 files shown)


Ailrun (approved):
Thank you for the update!

install/src/Cabal.hs
```diff
@@ -72,20 +72,17 @@ cabalInstallHls versionNumber args = do
...
     ++ args
 
-  let minorVerExe = "haskell-language-server-" ++ versionNumber <.> exe
-      majorVerExe = "haskell-language-server-" ++ dropExtension versionNumber <.> exe
+  let verExe = "haskell-language-server-" ++ versionNumber <.> exe
 
   let copyCmd old new = if os == "mingw32"
                         then liftIO $ copyFile old new
                         else command [] "ln" ["-f", old, new]
-  copyCmd (localBin </> "haskell-language-server" <.> exe) (localBin </> minorVerExe)
```
jneira, line 81: Jumm, i think the ghc-x.y.z (`minorVerExe` in the code) is still needed, to have several hls versions and let the wrapper choose the correct one (the `hlsBin` in the wrapper code)
We should remove only the ghc-x.y

jneira: Ah sorry you have renamed it to verExe, nvm

jneira (approved):
thanks for cleaning it out

PR #1814, opened 2021-05-09
Add a hook for modifying the dynflags from a plugin

isovector:
This PR changes the hls-plugin-api, adding a new `pluginModifyDynflags` hook to the `PluginDescriptor`. This field has type:

```haskell

data DynFlagsModifications =
  DynFlagsModifications { dynFlagsModifyGlobal :: DynFlags -> DynFlags
                        , dynFlagsModifyParser :: DynFlags -> DynFlags
                        }
```

which are `DynFlags` endos for things you'd like to enable globally, and only for the parsing step. This PR uses the `modifyGlobal` field to pull some Wingman-specific functionality out of `ghcide`. The upcoming #1776 uses `modifyParser` to steal some Source Haskell syntax and allow inline tactic metaprogramming.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -111,7 +111,7 @@ import           Development.IDE.GHC.Compat                   hiding
                                                                writeHieFile)
 import           Development.IDE.GHC.Error
 import           Development.IDE.GHC.ExactPrint
-import           Development.IDE.GHC.Util
+import           Development.IDE.GHC.Util hiding (modifyDynFlags)
 import           Development.IDE.Import.DependencyInformation
 import           Development.IDE.Import.FindImports
 import qualified Development.IDE.Spans.AtPoint                as AtPoint
@@ -140,7 +140,7 @@ import           Module
 import           TcRnMonad                                    (tcg_dependent_files)
 
 import           Ide.Plugin.Properties (HasProperty, KeyNameProxy, Properties, ToHsType, useProperty)
-import           Ide.Types (PluginId)
+import           Ide.Types (PluginId, DynFlagsModifications(dynFlagsModifyGlobal, dynFlagsModifyParser))
 import           Data.Default (def)
 import           Ide.PluginUtils (configForPlugin)
 import           Control.Applicative
@@ -211,10 +211,12 @@ getParsedModuleRule :: Rules ()
 getParsedModuleRule =
   -- this rule does not have early cutoff since all its dependencies already have it
   define $ \GetParsedModule file -> do
-    ModSummaryResult{msrModSummary = ms} <- use_ GetModSummary file
+    ModSummaryResult{msrModSummary = ms'} <- use_ GetModSummary file
     sess <- use_ GhcSession file
     let hsc = hscEnv sess
     opt <- getIdeOptions
+    modify_dflags <- getModifyDynFlags id dynFlagsModifyParser
+    let ms = ms' { ms_hspp_opts = modify_dflags $ ms_hspp_opts ms' }
 
     let dflags    = ms_hspp_opts ms
         mainParse = getParsedModuleDefinition hsc opt file ms
@@ -284,8 +286,14 @@ getParsedModuleWithCommentsRule =
     opt <- getIdeOptions
 
     let ms' = withoutOption Opt_Haddock $ withOption Opt_KeepRawTokenStream ms
+    modify_dflags <- getModifyDynFlags id dynFlagsModifyParser
+    let ms = ms' { ms_hspp_opts = modify_dflags $ ms_hspp_opts ms' }
 
-    liftIO $ snd <$> getParsedModuleDefinition (hscEnv sess) opt file ms'
+    liftIO $ snd <$> getParsedModuleDefinition (hscEnv sess) opt file ms
+
+getModifyDynFlags :: a -> (DynFlagsModifications -> a) -> Action a
+getModifyDynFlags a f = maybe a (f . dynFlagsMods) <$> getShakeExtra
+
 
 getParsedModuleDefinition
     :: HscEnv
@@ -782,7 +790,9 @@ isHiFileStableRule = defineEarlyCutoff $ RuleNoDiagnostics $ \IsHiFileStable f -
 getModSummaryRule :: Rules ()
 getModSummaryRule = do
     defineEarlyCutoff $ Rule $ \GetModSummary f -> do
-        session <- hscEnv <$> use_ GhcSession f
+        session' <- hscEnv <$> use_ GhcSession f
+        modify_dflags <- getModifyDynFlags id dynFlagsModifyGlobal
+        let session = session' { hsc_dflags = modify_dflags $ hsc_dflags session' }
         (modTime, mFileContent) <- getFileContents f
         let fp = fromNormalizedFilePath f
         modS <- liftIO $ runExceptT $
--- a/ghcide/src/Development/IDE/Core/Service.hs
+++ b/ghcide/src/Development/IDE/Core/Service.hs
@@ -30,6 +30,8 @@ import qualified Language.LSP.Types              as LSP
 
 import           Control.Monad
 import           Development.IDE.Core.Shake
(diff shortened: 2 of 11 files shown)


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -211,10 +211,12 @@ getParsedModuleRule :: Rules ()
 getParsedModuleRule =
   -- this rule does not have early cutoff since all its dependencies already have it
   define $ \GetParsedModule file -> do
-    ModSummaryResult{msrModSummary = ms} <- use_ GetModSummary file
+    ModSummaryResult{msrModSummary = ms'} <- use_ GetModSummary file
     sess <- use_ GhcSession file
     let hsc = hscEnv sess
     opt <- getIdeOptions
+    modify_dflags <- getModifyDynFlags id dynFlagsModifyParser
+    let ms = ms' { ms_hspp_opts = modify_dflags $ ms_hspp_opts ms' }
```
pepeiborra, line 219: Isn't this `DynFlags` already modified in the `GetModSummary` rule?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -284,8 +286,14 @@ getParsedModuleWithCommentsRule =
     opt <- getIdeOptions
 
     let ms' = withoutOption Opt_Haddock $ withOption Opt_KeepRawTokenStream ms
+    modify_dflags <- getModifyDynFlags id dynFlagsModifyParser
+    let ms = ms' { ms_hspp_opts = modify_dflags $ ms_hspp_opts ms' }
```
pepeiborra, line 290: Same, isn't this `DynFlags` already modified?

ghcide/src/Development/IDE/Core/Service.hs
```diff
@@ -46,10 +48,11 @@ initialise :: Config
            -> HieDb
            -> IndexQueue
            -> IO IdeState
-initialise defaultConfig mainRule lspEnv logger debouncer options vfs hiedb hiedbChan =
+initialise defaultConfig mainRule dynFlagsMods lspEnv logger debouncer options vfs hiedb hiedbChan =
```
pepeiborra, line 51: Instead of an additional argument here, can we have this as a field in the `Options`?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -171,6 +171,7 @@ data ShakeExtras = ShakeExtras
      lspEnv :: Maybe (LSP.LanguageContextEnv Config)
     ,debouncer :: Debouncer NormalizedUri
     ,logger :: Logger
+    ,dynFlagsMods :: DynFlagsModifications
```
pepeiborra, line 174: we don't need this if we store them as a field in the `Options`

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -56,6 +57,19 @@ import           Text.Regex.TDFA.Text            ()
 newtype IdePlugins ideState = IdePlugins
   { ipMap :: [(PluginId, PluginDescriptor ideState)]}
 
+data DynFlagsModifications =
```
pepeiborra, line 60: This would be a good place to add documentation.

ghcide/src/Development/IDE/Core/Rules.hs
isovector: It's already modified with the `dynFlagsModifyGlobal`, but not the `*ModifyParser`

ghcide/src/Development/IDE/Core/Service.hs
isovector: Options already has a `optCustomDynFlags` field (added by you in b57a942e2). As best I can tell, today it's only ever `id`, but this seems like a good field to replace with the `DynFlagsModifications`.

ghcide/src/Development/IDE/Core/Rules.hs
pepeiborra: Can we do away with the global one? Why do we need the parser variant?

pepeiborra: NVM, I just read the description

ghcide/src/Development/IDE/Core/Service.hs
pepeiborra: Sounds good

wz1000:
I have to admit, I'm a bit concerned about how this interacts with recompilation checking.

pepeiborra:
> I have to admit, I'm a bit concerned about how this interacts with recompilation checking.

Why? As long as the choice of plugins doesn't change, I think that existing interface files should satisfy the recompilation checks.

wz1000:
> Why? As long as the choice of plugins doesn't change, I think that existing interface files should satisfy the recompilation checks.

The modsummary is changed using `dynFlagsModifyParser` before parsing, and this is the modsummary that is saved in the `ParsedModule`. This means it might diverge from the ModSummary returned by `GetModSummary`.

pepeiborra:
> > Why? As long as the choice of plugins doesn't change, I think that existing interface files should satisfy the recompilation checks.
> 
> The modsummary is changed using `dynFlagsModifyParser` before parsing, and this is the modsummary that is saved in the `ParsedModule`. This means it might diverge from the ModSummary returned by `GetModSummary`.

Ok, so the risk is that the "parsed" DynFlags are used to produce the interface file, whereas the "modSummary" `DynFlags` are the ones used to check for recompilation, which could lead to false negative checks. Agreed, this would be nasty.

@isovector do we really need two modification functions?

isovector:
> @isovector do we really need two modification functions?

Well, that's up to yall! All I care about is getting a parse with `-XQuasiQuotes` enabled, but @wz1000 was concerned about that doing all sorts of horrendous things down the line. I'm happy to take directions here, but am far enough out of my expertise that I'm not sure how much I can contribute other than code.

isovector:
@wz1000 PTAL at d452740 
@pepeiborra likewise, at 1170c2b

ghcide/src/Development/IDE/Main.hs
```diff
@@ -218,6 +218,7 @@ defaultMain Arguments{..} = do
                 config <- LSP.runLspT env LSP.getConfig
                 let options = (argsIdeOptions config sessionLoader)
                             { optReportProgress = clientSupportsProgress caps
+                            , optModifyDynFlags = pluginModifyDynflags plugins
```
pepeiborra, line 221: This throws away the callers `optModifyDynFlags`, which is no good. Please append to whatever is there instead.

ghcide/src/Development/IDE/Main.hs
```diff
@@ -306,8 +308,9 @@ defaultMain Arguments{..} = do
             sessionLoader <- loadSessionWithOptions argsSessionLoadingOptions "."
             let options =
                   (argsIdeOptions argsDefaultHlsConfig sessionLoader)
-                    { optCheckParents = pure NeverCheck,
-                      optCheckProject = pure False
+                    { optCheckParents = pure NeverCheck
+                    , optCheckProject = pure False
+                    , optModifyDynFlags = pluginModifyDynflags plugins
                     }
```
pepeiborra, line 314: I cannot remember any valid reason for these settings, so let's just get rid of them
```suggestion
```

pepeiborra (changes requested):

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -234,7 +235,7 @@ getParsedModuleRule =
             -- If we can parse Haddocks, might as well use them
             --
             -- HLINT INTEGRATION: might need to save the other parsed module too
-            ((diags,res),(diagsh,resh)) <- liftIO $ concurrently mainParse haddockParse
+            ((diags,res),(diagsh,resh)) <- liftIO $ fmap (fmap (fmap (fmap reset_ms))) $ concurrently mainParse haddockParse
```
jneira, line 238: time to `fmap⁴ = fmap . fmap .fmap . fmap` 🤣

PR #1752, opened 2021-04-18
Retry tests on SQLError 

pepeiborra:
The PR does a few things: 

- All the ghcide tests now run in-process using `runServerWithHandles`
- The tests now send logs to the LSP stream. This seems the only reasonable thing to do, but I'm open to other ideas
- The tests now retry once on `SQLError`, after a 1s sleep. This should hopefully help with #1430 
- The ghcide binary is no longer tested. This is fine - the ghcide README states that the ghcide binary is unsupported

PR #1823, opened 2021-05-12
Extract fourmolu plugin into a standalone package

gustavoavena:
## What
Extract fourmolu plugin to a standalone package. 
Similar to what was done for stylish-haskell (#1604) and brittany (#1422) plugins.

#### RFC: Test case in `Progress.hs`
**TL;DR:** is it ok to leave the progress tests case from fourmolu plugin where it is?

I'm basing this PR on the ones that did the same for stylish-haskell (#1604) and brittany (#1422). I moved the functional tests from the `Format` test suite the same way, but unlike stylish-haskell and brittany, fourmolu has a test case in `window/workDoneProgress` test suite (in `test/functional/Progress.hs`). 

My initial plan was to extract it to the fourmolu test suite, but it depends on some things from utility modules that are not visible to the new fourmolu test suite (e.g. `hlsCommand` from `Test.Hls.Command`).

It seems that to extract it, I would need to move some things around to avoid adding a dependency from the new test suite to the functional test suite. Since I wanted to avoid making bigger changes on my first PR, I wanted to double check if extracting that one test case is necessary, **if it's passing as is**.


nix/default.nix
```diff
@@ -32,6 +32,7 @@ let
             hls-test-utils = gitignoreSource ../hls-test-utils;
             hls-brittany-plugin = gitignoreSource ../plugins/hls-brittany-plugin;
             hls-stylish-haskell-plugin = gitignoreSource ../plugins/hls-stylish-haskell-plugin;
+            hls-stylish-haskell-plugin = gitignoreSource ../plugins/hls-fourmolul-plugin;
```
berberman, line 35:
```suggestion
            hls-fourmolul-plugin = gitignoreSource ../plugins/hls-fourmolul-plugin;
```

haskell-language-server.cabal
```diff
@@ -258,10 +258,7 @@ common floskell
 
 common fourmolu
   if flag(fourmolu) || flag(all-formatters)
-    hs-source-dirs: plugins/default/src
-    other-modules: Ide.Plugin.Fourmolu
-    build-depends: fourmolu                 ^>=0.3
-    cpp-options: -Dfourmolu
```
berberman, line 264: We still need `cpp-options: -Dfourmolu` to control whether HLS wires this plugin in Main

pepeiborra:
> My initial plan was to extract it to the fourmolu test suite, but it depends on some things from utility modules that are not visible to the new fourmolu test suite (e.g. hlsCommand from Test.Hls.Command). It seems that to extract it, I would need to move some things around to avoid adding a dependency from the new test suite to the functional test suite. Since I wanted to avoid making bigger changes on my first PR, I wanted to double check if extracting that one test case is necessary, if it's passing as is.

@gustavoavena can you specify what needs to be moved? Maybe some tests need to be refactored instead.
@berberman how difficult is it to extract the tests? I would think it should be pretty easy after your recent changes introducing his-test-utils.

gustavoavena:
> @gustavoavena can you specify what needs to be moved? Maybe some tests need to be refactored instead.
One thing would be `hlsCommand` (from `Test.Hls.Command`). `formatLspConfig` and `progressCaps` are also  used by the test case.

I'm not very familiar with the code, but maybe refactoring the `runServer*` functions in `Test.Hls` to provide one that we could use for this test case would also work.

berberman:
The purpose of `hls-test-utils` is to decouple plugins test suite with HLS executable, and we should never use `hlsCommand` in plugin packages. `hlsCommand` will be removed finally when we migrate all HLS tests (run them in the same process, without building server executable first), but it seems to take a bit more work to do (#1752). 

Moreover, now there are some test suites in HLS referencing plugins' test data, e.g. `Prograss` mentioned above:

https://github.com/haskell/haskell-language-server/blob/bb999054b1703b196f05f8d0cddcc23709a02d86/test/functional/Progress.hs#L32-L53

Which is a bit hacky but works. Unfortunately, I don't have a good idea so far.

@gustavoavena Could you try changing the path `"test/testdata/format"` to  your extracted test data, like eval plugin, and leaving other things in `Prograss` unchanged? P.S. There might be other HLS test suites like `Prograss` which reference fourmolu test data, and you'll need change them as well.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -0,0 +1,60 @@
...
+    , base
+    , bytestring
+    , hls-fourmolu-plugin
+    , text
+    , data-default
+    , hspec-expectations
+    , lens
+    , ghcide
+    , filepath
+    , hls-test-utils ^>= 1.0.0.0
```
berberman, line 54:
```suggestion
    , hls-test-utils ^>= 1.0
```

berberman (approved):
LGTM, thank you!

test/functional/Format.hs
```diff
@@ -16,46 +16,45 @@ import           Test.Hls.Command
...
 providerTests :: TestTree
 providerTests = testGroup "formatting provider" [
-    testCase "respects none" $ runSessionWithConfig (formatConfig "none") hlsCommand fullCaps "test/testdata/format" $ do
+    testCase "respects none" $ runSessionWithConfig (formatConfig "none") hlsCommand fullCaps fourmoluTestDataDir $ do
         doc <- openDoc "Format.hs" "haskell"
         resp <- request STextDocumentFormatting $ DocumentFormattingParams Nothing doc (FormattingOptions 2 True Nothing Nothing Nothing)
         liftIO $ resp ^. LSP.result @?= Left (ResponseError InvalidRequest "No plugin enabled for STextDocumentFormatting, available: []" Nothing)
 
-    ,  testCase "respects initial" $ runSessionWithConfig (formatConfig "floskell") hlsCommand fullCaps "test/testdata/format" $ do
+    ,  testCase "respects initial" $ runSessionWithConfig (formatConfig "floskell") hlsCommand fullCaps fourmoluTestDataDir $ do
```
Ailrun, line 51: Since this `Format.hs` used for Floskell and Ormolu tests, how about duplicating the file to the plugin and integrated test directory for these tests?

It looks a bit strange that this kind of test data dependency exists

gustavoavena: Sure. I see that you pushed a commit extracting the floskell plugin as well, so I'll do the same and copy the test files to the fourmolu test directory.

Ailrun (approved):
LGTM :)

PR #1824, opened 2021-05-12
Apply formats again

Ailrun:
I'm now considering adding CI check for styling, but one big issue is that Haskell formatters (including Stylish-haskell, Ormolu, and Brittany) are not able to parse some specific usage of CPP pragmas.
For example, formatters cannot parse this code:
```Haskell
  test
#if defined(SOME_MACRO)
    = definition1
#else
    = definition2
#endif
```
and give an error. (However, if we move `test` to both branches they do not give an error)

We have quite a lot modules using pragmas in this way, so adding CI check will cause an unnecessary problems to modifying these modules.

Any idea will be appreciated.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -86,7 +86,14 @@ in (import sources.nixpkgs
       # default_stages = ["manual" "push"];
       hooks = {
         stylish-haskell.enable = true;
-        stylish-haskell.excludes = [ "^Setup.hs$" "test/testdata/.*$" "test/data/.*$" "^hie-compat/.*$" "^plugins/hls-tactics-plugin/.*$" ];
+        stylish-haskell.excludes = [
+          # Ignored files
+          "^Setup.hs$" "test/testdata/.*$" "test/data/.*$" "test/manual/lhs/.*$" "^hie-compat/.*$" "^plugins/hls-tactics-plugin/.*$"
+
+          # Temporarily ignored files
+          # Stylish-haskell (and other formatters) does not work well with some CPP usages in these files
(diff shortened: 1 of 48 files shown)


pepeiborra:
Could you disable the formatted in certain modules using a pragma?

Ailrun:
@pepeiborra Yes, that's a way, but will the list maintenance cost be OK?

Maybe it's not that expensive though...

pepeiborra (approved):

jneira (approved):

jneira:
Fixing upstream our actual formatter? :-P

nix/default.nix
```diff
@@ -86,7 +86,14 @@ in (import sources.nixpkgs
...
       hooks = {
         stylish-haskell.enable = true;
-        stylish-haskell.excludes = [ "^Setup.hs$" "test/testdata/.*$" "test/data/.*$" "^hie-compat/.*$" "^plugins/hls-tactics-plugin/.*$" ];
+        stylish-haskell.excludes = [
+          # Ignored files
+          "^Setup.hs$" "test/testdata/.*$" "test/data/.*$" "test/manual/lhs/.*$" "^hie-compat/.*$" "^plugins/hls-tactics-plugin/.*$"
+
+          # Temporarily ignored files
+          # Stylish-haskell (and other formatters) does not work well with some CPP usages in these files
+          "^ghcide/src/Development/IDE/GHC/Compat.hs$" "^plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs$" "^ghcide/test/exe/Main.hs$" "ghcide/src/Development/IDE/Core/Rules.hs" "^hls-test-utils/src/Test/Hls/Util.hs$"
```
Ailrun, line 95: Now these are added

Issue #1808, opened 2021-05-07
Failing test with whitespace mismatches for hls-splice-plugin


maralorn:
### Your environment

Trying to build hls-splice-plugin-1.0.0.1.

Which OS do you use: Nixos

### Steps to reproduce

I ran the tests.

### Expected behaviour

Tests pass.

### Actual behaviour
```
  TSimpleDecl.hs (golden):      FAIL (111.47s)
    test/Main.hs:87:
    expected: "{-# LANGUAGE TemplateHaskell #-}\r\n{-# LANGUAGE QuasiQuotes #-}\r\nmodule TSimpleDecl where\r\nimport Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )\r\n\r\n-- Foo\r\n--  Bar\r\nfoo :: Int\r\nfoo = 42\r\n-- Bar\r\n-- ee\r\n-- dddd\r\n"
     but got: "{-# LANGUAGE TemplateHaskell #-}\r\n{-# LANGUAGE QuasiQuotes #-}\r\nmodule TSimpleDecl where\r\nimport Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )\r\n\r\n-- Foo\r\n--  Bar\r\nfoo :: Int\nfoo = 42\r\n-- Bar\r\n-- ee\r\n-- dddd\r\n"
  TQQDecl.hs (golden):          FAIL (133.83s)
    test/Main.hs:87:
    expected: "{-# LANGUAGE QuasiQuotes #-}\r\nmodule TQQDecl where\r\nimport QQ (str)\r\n\r\nfoo :: String\r\nfoo = \"foo\"\r\n"
     but got: "{-# LANGUAGE QuasiQuotes #-}\r\nmodule TQQDecl where\r\nimport QQ (str)\r\n\r\nfoo :: String\nfoo = \"foo\"\r\n"
  TTypeKindError.hs (golden):   OK (125.31s)
  TDeclKindError.hs (golden):   FAIL (118.40s)
    test/Main.hs:120:
[...]

PR #1825, opened 2021-05-12
Bump up hls-splice-plugin to 1.0.0.2

jneira:
* As commented in #1808 this version is needed to upload a new hackage tarball with lf line endings


jneira:

Ailrun (approved):
Thank you :)

berberman:
We might want to mark `hls-splice-plugin-1.0.0.1` as a deprecated version in Hackage, so that cabal solver would be more likely to pick this new version.

PR #1829, opened 2021-05-13
Extract floskell plugin as a standalone plugin

Ailrun:
Its basic structure is copied from `hls-brittany-plugin`.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -55,8 +55,10 @@ jobs:
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
       run: cabal update
 
+    # max-backjumps is increased as a temporary solution
+    # for dependency resolution failure
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
-      run: cabal configure --enable-benchmarks
+      run: cabal configure --enable-benchmarks --max-backjumps 10000
 
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
       name: Build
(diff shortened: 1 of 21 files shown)


Ailrun:
Cabal configuration fails for benchmark with the following message:
```
 Warning: Requested index-state2021-03-29T21:23:14Z is newer than
'hackage.haskell.org'! Falling back to older state (2021-03-29T13:57:59Z).
Resolving dependencies...
[__0] trying: ghcide-1.3.0.0 (user goal)
[__1] rejecting: ghcide:!test (constraint from config file, command line flag,
or user target requires opposite flag selection)
[__1] trying: ghcide:*test
[__2] trying: tasty-1.4.1 (dependency of ghcide *test)
[__3] trying: ansi-terminal-0.11 (dependency of tasty)
[__4] next goal: retrie (dependency of ghcide)
[__4] rejecting: retrie-0.1.1.1 (conflict: ansi-terminal==0.11, retrie =>
ansi-terminal>=0.10.3 && <0.11)
[__4] skipping: retrie-0.1.1.0, retrie-0.1.0.1, retrie-0.1.0.0 (has the same
characteristics that caused the previous version to fail: excludes
'ansi-terminal' version 0.11)
[__4] fail (backjumping, conflict set: ansi-terminal, ghcide, retrie)
Backjump limit reached (currently 4000, change with --max-backjumps or try to
run with --reorder-goals).
```

Maybe we need to increase the max backjump limit? (I tried 5000, and that at least works)

jneira (approved):
It looks really great (including improvements in tests), many thanks

PR #1836, opened 2021-05-16
Extract last 2 plugins and clean up others

Ailrun:
Extract
- ormolu plugin
- module name plugin

(last 2 plugins reside in `default` directory) and clean up cabal files & tests of other plugins.

Although I think it's generally better to use `.hs` extensions even for golden files so that we can have syntax highlight and others, I didn't touch Wingman's tests as I am not sure whether any of @isovector's tools depend on golden test extensions or not. Is it OK to update Wingman's tests so that we have `.expected.hs` or similar.


hls-test-utils/src/Test/Hls.hs
```diff
@@ -67,6 +72,24 @@ gitDiff fRef fNew = ["git", "-c", "core.fileMode=false", "diff", "--no-index", "
...
+  -> TestName
+  -> FilePath
+  -> FilePath
+  -> FilePath
+  -> FilePath
+  -> (TextDocumentIdentifier -> Session ())
+  -> TestTree
+goldenWithHaskellDoc plugin title testDataDir path desc ext act =
+  goldenGitDiff title (testDataDir </> path <.> desc <.> ext)
+  $ runSessionWithServer plugin testDataDir
```
Ailrun, line 86: there should be a version using `runSessionWithServerFormatter`.

jneira (approved):

isovector:
@Ailrun completely fine to rename the Wingman tests! thanks for tackling this!

plugins/hls-tactics-plugin/test/Utils.hs
```diff
@@ -61,7 +61,7 @@ codeActionTitle (InR(CodeAction title _ _ _ _ _ _ _)) = Just title
 mkTest
     :: Foldable t
     => String  -- ^ The test name
-    -> FilePath  -- ^ The file to load
+    -> FilePath  -- ^ The file basename to load
```
Ailrun, line 64: It's not "basename", it's a stem.

plugins/hls-eval-plugin/test/testdata/hie.yaml
```diff
@@ -1,4 +1,3 @@
 cradle:
-  cabal:
-    - path: "././"
-      component: "lib:test"
+  direct:
+    arguments: []
```
Ailrun, line 3: Should list `Util.hs`

PR #1845, opened 2021-05-19
Skip cachix jobs if token is unset

berberman:
We should skip running cachix jobs on PRs opend from forked repos

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -20,6 +20,8 @@ jobs:
   nix:
     needs: pre_job
     runs-on: ${{ matrix.os }}
+    env:
+      HAS_TOKEN: ${{ secrets.HLS_CACHIX_AUTH_TOKEN != '' }}
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


Ailrun (approved):

.github/workflows/nix.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         authToken: ${{ secrets.HLS_CACHIX_AUTH_TOKEN }}
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
       run: nix build
-    - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+    - if: ${{ needs.pre_job.outputs.should_skip != 'true' && github.repository_owner == 'haskell' }}
       run: nix develop --profile dev && cachix push haskell-language-server dev
-    - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+    - if: ${{ needs.pre_job.outputs.should_skip != 'true' && github.repository_owner == 'haskell' }}
```
berberman, line 51: mmm, this doesn't seem to work at all

Ailrun: How about explicitly check whether the secret is defined or not?

berberman: actions/runner/issues/520

Ailrun: I guess https://github.com/actions/runner/issues/520#issuecomment-700579336 suggests a way?

berberman: I'll try it later. We have to resolve this first, since the nix CI keeps failing, blocking other PRs.

PR #1848, opened 2021-05-20
Avoid package-qualified import in Fourmolu plugin

georgefst:
Nobody likes `PackageImports`. And it's unnecessary since #1823.

### DIFF
--- a/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
+++ b/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE LambdaCase               #-}
 {-# LANGUAGE OverloadedStrings        #-}
-{-# LANGUAGE PackageImports           #-}
 {-# LANGUAGE TypeApplications         #-}
 
 module Ide.Plugin.Fourmolu (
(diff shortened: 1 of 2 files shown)


jneira (approved):
thanks for cleaning it out

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -25,7 +24,7 @@ import           Ide.Types
 import           Language.LSP.Server         hiding (defaultConfig)
 import           Language.LSP.Types
 import           Language.LSP.Types.Lens
-import           "fourmolu" Ormolu
+import           Ormolu
```
Ailrun, line 27: I once removed it in #1836, but I thought that this may confuse people...

Ailrun: BTW, you can also remove it from Ormolu if you prefer

georgefst: Sure: 7bb5a9c

georgefst: > I once removed it in #1836, but I thought that this may confuse people...

My thinking is that with the new plugin architecture, those two plugins no longer know anything about each other, so it doesn't really make sense to keep the qualified imports.

In fact, has there been any discussion about the long-term plan for where the plugins live? I wouldn't mind if the Fourmolu plugin was moved to [the Fourmolu organisation](https://github.com/fourmolu), for example.

Ailrun (approved):

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
Ailrun: AFAIK there has been no such discussion, and I'm afraid that would increase maintenance cost significantly... but it is definitely worthwhile to discuss that with other maintainers.

PR #1852, opened 2021-05-21
Fix flake compat

### LINKED ISSUE #1846: Contributing via nix-env doesn't work, zlib seems to be missing from derivation
Using NixOS 20.09

### Steps to reproduce
Clone repo then follow steps of using-nix
https://github.com/haskell/haskell-language-server#using-nix
cabal build

### Actual behaviour
cabal: Failed to build zlib-0.6.2.3
https://gist.github.com/ErnestKz/f0441e3ce1d55cf22331ac849bfeafc5

berberman:
Closes #1846

We should enter `devShell` rather than the shell of `defaultPackage`.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -8,4 +8,4 @@
     sha256 = lock.nodes.flake-compat.locked.narHash; }
 ) {
   src =  ./.;
-}).defaultNix
+}).shellNix


ErnestKz (approved):
Hey, just some questions to help me learn:

In theory would keeping the file named `default.nix`  still work? What does the `.shellNix` do on the flake import? Is the `.shellNix` something that's defined outside of this repo?

Also thank you for fixing this.

berberman:
> In theory would keeping the file named default.nix still work?

Yes, `nix-shell` will use `default.nix` if there is no `shell.nix`.

> What does the `.shellNix` do on the flake import? Is the `.shellNix` something that's defined outside of this repo?

It's defined in https://github.com/edolstra/flake-compat/blob/master/default.nix, which basically allows you use flakes with the legacy nix commands (`nix-build` and `nix-shell`). You can find more details in "Using flakes project from a legacy Nix" of https://nixos.wiki/wiki/Flakes.

Ailrun (approved):
Awesome, thanks!

PR #1827, opened 2021-05-13
nix: refactor with flakes

berberman:
[Flakes](https://www.tweag.io/blog/2020-05-25-flakes/) are the new feature of nix package manager that solves pain points of pinning nix dependencies and makes nix parts more composable. Recently, [pre-commit-hooks.nix](https://github.com/cachix/pre-commit-hooks.nix) adds flake supports, but now we use it via `builtins.fetchTarball`, which is not good for maintenance. Currently we use [niv](https://github.com/nmattia/niv) to manage nix dependencies, but the generated files and inconsistency are annoying. So I think it's time to migrate to flakes. New commands:

* `nix flake update` - update all nix dependencies
* `nix develop` - start a HLS development environment, like the legacy `nix-shell`
* `nix build` - build HLS without `cabal-install`
* `nix run` - run HLS executable
* ...

If a downstream package using flakes wants to include HLS to its dev shell, it can directly add `"github:haskell/haskell-language-server"` to the `inputs`, and the latest git version of HLS will be available.

I keep only the support of the default GHC version in nixpkgs, because I don't think people will use other GHC versions to develop HLS with nix - not using binary caches is painful. I also add [flake-compat](https://github.com/edolstra/flake-compat) to make legacy `shell.nix` still available.

Now running `nix flake check` or `nix flake show` will cause the evaluation of each system's expression, so I got :

```
error: a 'x86_64-darwin' with features {} is required to build '/nix/store/jxm5zc2s1w6qzfw28j0fdijq2q66d8py-cabal2nix-haskell-language-server.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```
on my NixOS machine, which will not affect the development, but I don't know where the problem is, and I'd appreciate if anyone could help...

UPDATE: I think the working principle of `callCabal2nix` might have some problems with flakes, and other Haskell projects using flakes seem to have the similar issue as well.

PR #1853, opened 2021-05-21
Update mergify.yml

Ailrun:
This updates mergify setting to make it work with #1827

Maybe we can entirely remove "status-success" part, because, as shown in https://github.com/haskell/haskell-language-server/pull/1852/checks?check_run_id=2641030076, it looks like mergify bot now follows the branch protection rule to find necessary checks.


jneira (approved):
👍

PR #1858, opened 2021-05-22
Use last apply-refact and several stack.yaml updates

### LINKED ISSUE #1844: Syntax error after hlint refactor
Applying a refactoring from hlint produces a syntactically wrong result.

### Your environment
```sh
myproject$ haskell-language-server --probe-tools
haskell-language-server version: 1.1.0.0 (GHC: 8.10.4) (PATH: C:\Users\lennart.augustsson\AppData\Roaming\cabal\bin\haskell-language-server.exe)
cabal:          3.4.0.0
stack:          Not found
ghc:            8.10.4
```
Which OS do you use:
Windows

Which lsp-client do you use:
emacs

### Steps to reproduce
Use this file:
```haskell
module M where
import GHC.Stack
foo :: (HasCallStack) => Int
foo = undefined
```
Apply the suggest refactoring to remove redundant parens.

### Expected behaviour
The result should be
```
foo :: HasCallStack => Int
```

### Actual behaviour
The result is
```
foo :: (HasCallStack => Int
```

jneira:
* cabal side: update hackage index and bump up min apply-refact version
  * fixes #1844
* stack side:
  * set the new apply-refact version (it needed bump up brittany version too)
  * drop stack.yaml for 8.10.2 and 8.8.2
  * make stack-8.10.4.yaml the default one copying into the stack.yaml
  * bump up hie-bios to 0.7.5


.circleci/config.yml
```diff
@@ -88,11 +83,6 @@ jobs:
       - STACK_FILE: "stack-8.8.4.yaml"
     <<: *defaults
 
-  ghc-8.10.2:
-    environment:
-      - STACK_FILE: "stack-8.10.2.yaml"
-    <<: *defaults
-
```
Ailrun, line 95: Will our future plan be supporting the latest two minor versions per major one?

stack.yaml
```diff
@@ -1,10 +1,10 @@
-resolver: lts-14.27 # Last 8.6.5
+resolver: lts-17.12
```
Ailrun, line 1: I love this update :)

stack-8.10.3.yaml
```diff
@@ -29,7 +29,7 @@ ghc-options:
   "$everything": -haddock
 
 extra-deps:
-  - brittany-0.13.1.0
```
Ailrun, line 32: We don't need to add apply-refactor here?

Ailrun (approved):
Thanks!

stack-8.10.3.yaml
jneira: yeah, good catch, ci should fail for that stack.yaml

stack.yaml
jneira: yeah, 8.10.4 windows version seems to be reliable enough to make it the default

.circleci/config.yml
jneira: I was not thinking in make it a hard rule for future, but seemed to be time to do it.
Maybe we should make explicit some policy, although it should be flexible, not all versions are equally reliable, popular, etc.

jneira: @ailrThinking it twice we have to remove them from elsewhere and those changes maybe should too big to be included herem so i will open an issue and a pr.
That said, do you think we should revert the removing of stack.yaml's here and do it in that future pr?

Ailrun: Yes, IMO `stack-<version>.yaml`s are one of the immediate hints whether we support some version or not.

jneira: I ve reverted that change and make them work as well, opening a dedicated issue to discuss drop support for 8.8.2 and 8.10.2

jneira:
* benchamark is failing in the solver phase with:
```
Resolving dependencies...
[__0] trying: ghcide-1.3.0.0 (user goal)
[__1] trying: lsp-test-0.14.0.0 (dependency of ghcide)
[__2] trying: some-1.0.3 (dependency of lsp-test)
[__3] trying: lsp-types-1.2.0.0 (dependency of ghcide)
[__4] next goal: dependent-sum (dependency of ghcide)
[__4] rejecting: dependent-sum-0.7.1.0 (conflict: some==1.0.3, dependent-sum
=> some>=1.0.1 && <1.0.3)
[__4] skipping: dependent-sum-0.7.0.0 (has the same characteristics that
caused the previous version to fail: excludes 'some' version 1.0.3)
[__4] rejecting: dependent-sum-0.6.2.0 (conflict: lsp-types =>
dependent-sum>=0.6.2.2)
[__4] skipping: dependent-sum-0.5, dependent-sum-0.4, dependent-sum-0.3.2.2,
dependent-sum-0.3.2.1, dependent-sum-0.3.2.0, dependent-sum-0.3.1.0,
dependent-sum-0.2.1.0, dependent-sum-0.2.0.2, dependent-sum-0.2.0.1,
dependent-sum-0.2, dependent-sum-0.1 (has the same characteristics that caused
the previous version to fail: excluded by constraint '>=0.6.2.2' from
'lsp-types')
[__4] rejecting: dependent-sum-0.6.2.2 (conflict: some==1.0.3, dependent-sum
=> some>=1.0.0.3 && <1.0.1)
[__4] rejecting: dependent-sum-0.6.1 (conflict: lsp-types =>
dependent-sum>=0.6.2.2)
[__4] skipping: dependent-sum-0.6 (has the same characteristics that caused
the previous version to fail: excluded by constraint '>=0.6.2.2' from
'lsp-types')
[__4] fail (backjumping, conflict set: dependent-sum, ghcide, lsp-types, some)
Backjump limit reached (currently 10000, change with --max-backjumps or try to
run with --reorder-goals).
```
i will try to increase the limit (already increased recently)

* nix build with

```
3.2.1.0:Distribution.Simple.UserHooks
       >   configureAction, called at libraries/Cabal/Cabal/Distribution/Simple.hs:180:19 in Cabal-3.2.1.0:Distribution.Simple
       >   defaultMainHelper, called at libraries/Cabal/Cabal/Distribution/Simple.hs:116:27 in Cabal-3.2.1.0:Distribution.Simple
       >   defaultMain, called at /nix/store/4mdp8nhyfddh7bllbi7xszz7k9955n79-Setup.hs:2:8 in main:Main
       > Setup: Encountered missing or private dependencies:
       > apply-refact >=0.9.3.0
       >
```

I hope #1859 could help to workareound this, not sure why it does not work though, bumping up versions in .cabal file to get fixes is a very common task 🤔

.github/workflows/bench.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     # max-backjumps is increased as a temporary solution
     # for dependency resolution failure
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
-      run: cabal configure --enable-benchmarks --max-backjumps 10000
+      run: cabal configure --enable-benchmarks
```
jneira, line 61:
```suggestion
      run: cabal configure --enable-benchmarks --max-backjumps 12000
```

Locally it works for me without `--max-backjumps` and `--max-backjumps 12000` but no with `--max-backjumps 10000`, unfathomable mysteries of cabal solver

PR #1859, opened 2021-05-23
Use cabal-install if nix is failing in CI

berberman:
We have decided to maintain the nix packaging (building HLS binaries in nix) in #1827. Ideally, we are always have HLS binaries prepared without `cabal-install`. However, PRs that update project dependencies like #1858 may be blocked by nix ci, since nixpkgs does not have the new dependency yet, and even if it has, a bump of nixpkgs is required. It's unreasonable to require every contributor to handle nix things, so this PR add a fallback in CI, which uses `cabal-install` to build HLS in case some required packages are missing in nixpkgs.

Also adds `default.nix`, so that the legacy `nix-build` can work.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -47,7 +47,7 @@ jobs:
         name: haskell-language-server
         authToken: ${{ secrets.HLS_CACHIX_AUTH_TOKEN }}
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
-      run: nix build
+      run: nix build || (nix develop --command cabal update && nix develop --command cabal build)
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' && env.HAS_TOKEN == 'true' }}
       run: nix develop --profile dev && cachix push haskell-language-server dev
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' && env.HAS_TOKEN == 'true' }}
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches:
       - '**'
+  push:
+    branches:
+      - master
 
 jobs:
   pre_job:
@@ -17,11 +20,10 @@ jobs:
           cancel_others: true
           paths_ignore: '["**/docs/**", "**.md", "**/LICENSE", ".circleci/**", "install/**"]'
 
-  nix:
+  # Enter the development shell and run `cabal build`
+  develop:
     needs: pre_job
     runs-on: ${{ matrix.os }}
-    env:
-      HAS_TOKEN: ${{ secrets.HLS_CACHIX_AUTH_TOKEN != '' }}
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/nix.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         name: haskell-language-server
         authToken: ${{ secrets.HLS_CACHIX_AUTH_TOKEN }}
     - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
-      run: nix build
+      run: nix build || (nix develop --command cabal update && nix develop --command cabal build)
```
pepeiborra, line 50: Can we split this in two jobs instead? One for the `nix build` and one for the `cabal build`. The first one would not be required to merge.

.github/workflows/nix.yml
```diff
@@ -41,14 +43,43 @@ jobs:
...
         nix_path: nixpkgs=channel:nixos-unstable
-    - if: ${{ needs.pre_job.outputs.should_skip != 'true' && env.HAS_TOKEN == 'true' }}
-      uses: cachix/cachix-action@v8
+    - if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+      uses: cachix/cachix-action@v10
       with:
         name: haskell-language-server
-        authToken: ${{ secrets.HLS_CACHIX_AUTH_TOKEN }}
+        # Disable pushing, we will do that in job `build`
+        skipPush: true
```
pepeiborra, line 51: is this a good idea? I expect that the `nix build` job will be failing most of the time.

berberman: I think we could prioritize pushing the cache of the development shell in the job `build`, so that the push will take place even if `nix build` and the job fail. If I understand correctly, the cache of development shell is very small, where most of drvs are provided by https://cache.nixos.org/. And I believe that `nix develop` now works smoothly without setting up our cachix.

jneira (approved):
I hope this would unblock #1858 🙂

berberman:
@jneira Branch protection rules need updates correspondingly

We should remove:

* nix (macOS-latest)
* nix (ubuntu-latest) 

And add:

* Nix / develop (ubuntu-latest)
* Nix / develop (macOS-latest)

jneira:
> @jneira Branch protection rules need updates correspondingly

you can do it no yourself right now :slightly_smiling_face:

.github/workflows/nix.yml
```diff
@@ -41,14 +43,43 @@ jobs:
...
-      run: nix path-info --json | jq -r '.[].path' | cachix push haskell-language-server
+      run: |
+        nix develop --command cabal update
+        nix develop --command cabal build
+
+  # Build and then push HLS binaries with developmet shell to cachix
+  # This job runs when PRs are merged to master, and should be excluded from branch protections
+  build:
+    runs-on: ${{ matrix.os }}
+    if: ${{ github.repository_owner == 'haskell' && github.ref == 'ref/heads/master' }}
```
berberman, line 61: It didn't run as expected :(

https://github.com/haskell/haskell-language-server/actions/runs/870310690

PR #1861, opened 2021-05-23
Trace more Shake evaluation details

pepeiborra:
Opentelemetry traces have proven very useful for understanding the evaluation and performance of the Shake rules. This change adds more information to those traces:

- What mode a rule is running in - is this a full rerun (dependencies changed) or just a lookup in the values state?
- Whether the rule value changed after running

image

### DIFF
--- a/ghcide/src/Development/IDE/Core/Tracing.hs
+++ b/ghcide/src/Development/IDE/Core/Tracing.hs
@@ -96,9 +96,7 @@ otTracedAction key file mode success act
             res <- act
             unless (success $ runValue res) $ setTag sp "error" "1"
             setTag sp "changed" $ case res of
-              RunResult ChangedRecomputeSame _ _ -> "0"
-              RunResult ChangedNothing _ _       -> "0"
-              RunResult ChangedRecomputeDiff _ _ -> "0"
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/Core/Tracing.hs
```diff
@@ -77,21 +78,27 @@ otTracedAction
...
         )
         endSpan
         (\sp -> do
             res <- act
-            unless (success res) $ setTag sp "error" "1"
+            unless (success $ runValue res) $ setTag sp "error" "1"
+            setTag sp "changed" $ case res of
+              RunResult ChangedRecomputeSame _ _ -> "0"
+              RunResult ChangedNothing _ _       -> "0"
+              RunResult ChangedRecomputeDiff _ _ -> "0"
```
berberman, line 101: This pattern match is non-exhaustive, though we never use `ChangedStore`. BTW why the result of these three cases is constant `"0"`?

pepeiborra: That was a mistake during cleanup, corrected now. Thanks!

berberman (approved):
LGTM, thanks!

PR #1862, opened 2021-05-23
Technology preview: Keep track of changes to minimize rebuilds

pepeiborra:
_NOTE: This is a technology preview of something I have been working on for several months. It requires extending `his-graph` to accept the set of keys that have changed since the last build as an input, of which there isn't a final implementation yet. There is a Shake implementation (https://github.com/ndmitchell/shake/pull/802), but it might end up being reimplemented  in #1759 instead. Details still being worked out._

This PR extends ghcide to keep track of the set of rules that have changed since the last successful build. This information is then shard with his-graph, which can then use it to minimize the amount of build rules visited.

## Motivation

When his-graph runs a rule, it first tells it whether its dependencies have changed. Ghcide uses this information to skip evaluation and perform a lookup on the Values state when the dependencies have not changed. Even though this lookup is fast, the constant overheads are significant and the cost can add up. Currently, in order to evaluate a `TypeCheck Some.Module` rule, Ghcide needs to perform a number of lookups proportional to the transitive imports  of `Some.Module`. This PR makes it technically possible to make the number of lookups independent of the number of transitive imports.

## Example

Consider the "edit" experiment on the Cabal 3.0.0.0 example of the benchmark suite, which performs a whitespace edit on the `Distribution.Simple` module and measures the typecheck time. Ghcide installs an LSP handler for the `TextDocumentDidChange` event that triggers a rebuild of the `TypeCheck Distribution.Simple` rule. However, this doesn't tell his-graph that the only thing that has changed since the last build is the `GetModificationTime Distribution.Simple` rule. Without this information, hls-graph needs to evaluate all the dependencies (collected in the previous run) of the `TypeCheck Distribution.Simple` rule to find out if they have changed since last run, and only then it can run the `TypeCheck` rule.

The `TypeCheck Distribution.Simple` rule depends on two other rules:
- `GhcSessionDeps Distribution.Simple`
- `GetParsedModule Distribution.Simple`

Again, before evaluating either of these two other rules, his-graph needs to run their dependencies to find out whether they have changed. In the case of `GhcSessionDeps`, this includes:
1. `GetTransitiveDependencies D.S`
2. `GetModIface X` for all X in the transitive dependencies of D.S
3. `GhcSession D.S`

Because of 2, his-graph needs to run a number of build rules proportional to the transitive deps of the module being type checked. We know that these runs will be just lookups, since nothing has changed in their dependencies all the way to the leaf rules, but still it's a significant overhead that could be avoided if we were to tell his-graph that the only rule that has changed is the file contents of `Distribution.Simple`. That's what this PR does.

Visually, the open telemetry traces below show the before and after (using https://github.com/ndmitchell/shake/pull/802) of the example. The traces show all the rules evaluated by Shake during the evaluation of `TypeCheck Distribution.Simple`. 

### Before
Takes around 700ms to do a TypeCheck of the module and the statistics table shows all the other rules visited during the evaluation.
Before detail

### After
Typecheck completes in only 70ms and the number of rules visited, while still proportional to the count of transitive imports, is much lower. Why is it still proportional? I haven't investigated but I suspect it's related to the `NeedsCompilation` rules depending on the whole project via the `ModuleGraph` rule. 
After detail

## Implementation Details

There is a new `dirtyKeys` mutable set to keep track of world changes, which are collected from LSP notifications. We already register file watchers for haskell source files, but this needs to be extended to any other files that we care about, e.g. cradle dependencies. Need to check whether this is already done.

- [x] File watches for cradle dependencies and TH dependencies

The set of dirty keys is handed over to his-graph in the call to `shakeRunDatabase`. his-graph is responsible for keeping track of the transitive dirty set across builds. For example, if `GetModificationTime D.S` has become dirty, all its reverse dependencies must be tagged as dirty too. his-graph can remove a rule from the dirty set once it's been evaluated, which may not happen in this build.

Other bits: 
- Rules that are not in the dirty set will not be evaluated, even if they are `alwaysRerun`. Missing rules will lead to stale results and hard to diagnose bugs. There is a flag `optRunSubset` that controls whether we call his-graph in the old way (everything has potentially changed) or in the new way (only these rules have changed) intended to debug these issues. 
- Rules cannot be removed from the dirty set until positive that they have been evaluated by hls-graph, specially since builds can be interrupted at any time. 
- The tracking of files of interest now becomes much more precise: currently everything depends on `GetFilesOfInterest`, this rule is now gone. 

## Merging

This work has been sitting in my fork for too long. Before I forget the details, I want to get it reviewed and possibly merged too even if there isn't a working hls-graph implementation available yet. The changes should be a no-op in terms of functionality and the costs of tracking dirty rules are negligible, so I think it should be safe to merge.

A version of these changes running against https://github.com/ndmitchell/shake/pull/802 and exercising the test and bench suites is available at https://github.com/pepeiborra/ide/pull/9

### DIFF
--- a/ghcide/src/Development/IDE/Core/Actions.hs
+++ b/ghcide/src/Development/IDE/Core/Actions.hs
@@ -116,7 +116,7 @@ highlightAtPoint file pos = runMaybeT $ do
 refsAtPoint :: NormalizedFilePath -> Position -> Action [Location]
 refsAtPoint file pos = do
     ShakeExtras{hiedb} <- getShakeExtras
-    fs <- HM.keys <$> getFilesOfInterest
+    fs <- HM.keys <$> getFilesOfInterestUntracked
     asts <- HM.fromList . mapMaybe sequence . zip fs <$> usesWithStale GetHieAst fs
     AtPoint.referencesAtPoint hiedb file pos (AtPoint.FOIReferences asts)
 
--- a/ghcide/src/Development/IDE/Core/OfInterest.hs
+++ b/ghcide/src/Development/IDE/Core/OfInterest.hs
@@ -5,10 +5,13 @@
 {-# LANGUAGE TypeFamilies      #-}
 
 -- | Utilities and state for the files of interest - those which are currently
---   open in the editor. The useful function is 'getFilesOfInterest'.
+--   open in the editor. The rule is 'IsFileOfInterest'
 module Development.IDE.Core.OfInterest(
     ofInterestRules,
-    getFilesOfInterest, setFilesOfInterest, modifyFilesOfInterest,
+    getFilesOfInterestUntracked,
+    addFileOfInterest,
+    deleteFileOfInterest,
+    setFilesOfInterest,
     kick, FileOfInterestStatus(..),
     OfInterestVar(..)
     ) where
@@ -16,7 +19,6 @@ module Development.IDE.Core.OfInterest(
 import           Control.Concurrent.Strict
 import           Control.Monad
 import           Control.Monad.IO.Class
-import           Data.Binary
 import           Data.HashMap.Strict                          (HashMap)
 import qualified Data.HashMap.Strict                          as HashMap
 import qualified Data.Text                                    as T
@@ -24,7 +26,7 @@ import           Development.IDE.Graph
 
 import           Control.Monad.Trans.Class
 import           Control.Monad.Trans.Maybe
-import qualified Data.ByteString.Lazy                         as LBS
+import qualified Data.ByteString                              as BS
 import           Data.List.Extra                              (nubOrd)
 import           Data.Maybe                                   (catMaybes)
 import           Development.IDE.Core.ProgressReporting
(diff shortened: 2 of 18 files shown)


ghcide/src/Development/IDE/Core/Actions.hs
```diff
@@ -116,7 +116,7 @@ highlightAtPoint file pos = runMaybeT $ do
 refsAtPoint :: NormalizedFilePath -> Position -> Action [Location]
 refsAtPoint file pos = do
     ShakeExtras{hiedb} <- getShakeExtras
-    fs <- HM.keys <$> getFilesOfInterest
+    fs <- HM.keys <$> getFilesOfInterestUntracked
```
wz1000, line 119: What is the reason for this change?

ghcide/src/Development/IDE/Core/OfInterest.hs
```diff
@@ -43,45 +45,59 @@ instance IsIdeGlobal OfInterestVar
 ofInterestRules :: Rules ()
 ofInterestRules = do
     addIdeGlobal . OfInterestVar =<< liftIO (newVar HashMap.empty)
-    defineEarlyCutOffNoFile $ \GetFilesOfInterest -> do
+    defineEarlyCutoff $ RuleNoDiagnostics $ \IsFileOfInterest f -> do
```
wz1000, line 48: Could we perhaps have both `IsFileOfInterest` (early cutoff, not always rerun) and `GetFilesOfInterest`(early cutoff, always rerun). Otherwise would could have subtle inconsistencies due to the file of interest set changing.

ghcide/src/Development/IDE/Core/Actions.hs
pepeiborra: You mean, why is it renamed?

ghcide/src/Development/IDE/Core/OfInterest.hs
pepeiborra: The only genuine use case for `GetFilesOfInterest` is early cutoff inside a rule definition. I looked around and that use case doesn't arise anywhere. All the calls to `getFilesOfInterest` are either in handlers or in `action` clauses. 

Given that `GetFilesOfInterest` is strictly more general than `IsFileOfInterest` but has catastrophic performance, I think it's best not to include it. The inconsistency that you describe cannot arise realistically since whenever the set is mutated, the Shake session is restarted.

mpickering:
Did you fix the bug I reported on IRC about stale diagnostics in parent files?

pepeiborra:
> Did you fix the bug I reported on IRC about stale diagnostics in parent files?

That bug is still there. There's probably more bugs, and it will take some effort to weed them out. But right now there isn't a convenient way to test the new functionality in CI, so I intend to wait until hls-graph supports reverse dependencies

pepeiborra:
Rebased on top of #1880 which fixes all the remaining test failures with `optRunSubset` enabled (see https://github.com/pepeiborra/ide/pull/9 )

Ready to merge, can I have a stamp @wz1000 ?

wz1000 (approved):

PR #1867, opened 2021-05-25
Wingman metaprogram command documentation

isovector:
This PR annotates every command exposed via the metaprogram parser with documentation. It also includes a little pretty printer to generate a markdown file of that documentation.

### DIFF
--- a/plugins/hls-tactics-plugin/COMMANDS.md
+++ b/plugins/hls-tactics-plugin/COMMANDS.md
@@ -2,7 +2,7 @@
 
 ## application
 
-arguments: none.
+arguments: none.  
 non-deterministic.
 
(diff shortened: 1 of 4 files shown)


plugins/hls-tactics-plugin/COMMANDS.md
```diff
@@ -0,0 +1,448 @@
+# Wingman Metaprogram Command Reference
+
+## application
+
+arguments: none.
+non-deterministic.
```
Ailrun, line 6: Maybe better to have two spaces after `none.` (so that there will be two separate lines)?
I mean like
```suggestion
arguments: none.  
non-deterministic.
```

Ailrun (approved):
Awesome! I was looking for this :)

PR #391, opened 2020-09-09
Case splitting and lambda introduction

isovector:
@TOTBWF and I had a mini hackathon last weekend and this is the results. It provides case splitting, homomorphic case splitting[^1], and lambda introduction. It does a reasonable job of coming up with good names, and implements a lot of general-purpose machinery for other plugins.

[^1] Case splitting to the same datacons --- useful for writing things like functor instances.

It looks like this:

Peek 2020-09-09 14-31

Fixes #323 --- although that usecase could be improved with a separate action that flattens `foo = \x y z -> bar` into `foo x y z = bar`.

We have plans to follow this up with more type-directed code generation; a great deal of it is already implemented, but not yet polished. Most of `Ide.Plugin.Tactic.Tactics` is towards that purpose.

PR #1873, opened 2021-05-28
Disable hole fit suggestions when running Wingman

### LINKED ISSUE #1602: Wingman gets slow on big modules

isovector:
HLS by default enables hole fit suggestions, and subsumption hole fits. These work nicely if you only ever have one hole, but perform superlinear amounts of work, and get extremely slow when multiple holes are present. This is a particular dealbreaker for Wingman, which introduces lots of holes, all the time.

This PR disables hole fit suggestions whenever Wingman is enabled. As a result, Wingman feels extremely snappy, even when there are 50+ holes present. Users who preferred the default hole-fit suggestions can disable Wingman to maintain the old behavior.

Fixes #1602


pepeiborra (approved):
This is reasonable, the only thing missing is documentation. Does Wingman have a README or similar?

isovector:
It does. Are you looking for documentation just about the functionality disabling?

pepeiborra:
Yes, the fact that hole suggestions will not be available when using Wingman

isovector:
Not sure what to do about the broken tests here. Just remove them? Don't run Wingman for them? Mark them as failing?

pepeiborra:
Probably disable Wingman using LSP configs. 
Also, you may want to add a test to check that the hole quick fixes are really disabled when Wingman is on.

isovector:
The LSP configs don't disable the `DynFlags` endos fortunately, so that approach doesn't work. I've disabled the tests for now.

test/functional/FunctionalCodeAction.hs
```diff
@@ -397,8 +397,10 @@ redundantImportTests = testGroup "redundant import code actions" [
                 ]
     ]
 
+
 typedHoleTests :: TestTree
 typedHoleTests = testGroup "typed hole code actions" [
+    ignoreTestBecause "Wingman changes the result of this test and I don't know how to disable Wingman" $
```
berberman, line 403: I think setting the lsp config to
```json
{
  "haskell": {
    "plugin": {
      "tactics": {
        "globalOn": false
      }
    }
  }
}
```
could disable Wingman

anka-213:
Does wingman replace the basic functionality of hole-fits where it fills in a hole with some variable of the correct type? E.g.
```
f :: Foo -> Bar
x :: Foo

g :: Bar
g = _ x
```
would fill in `f` here.

If not, I will sadly need to disable wingman on my computer if this change lands, but I don't want to lose case splitting either. I've never used subsumption hole fits though.

-----

Would it be possible to make ghc only compute hole fits when you explicitly ask for them, e.g. when you're at a hole and ask for completions/code actions?

test/functional/FunctionalCodeAction.hs
isovector: It might, but it doesn't seem to disable the `DynFlagsModifications` present in the descriptor, which is where this happens.

https://github.com/haskell/haskell-language-server/blob/41bf8b85ed40499d0a0b524277d51f470ec95e94/hls-plugin-api/src/Ide/Types.hs#L93

isovector:
@anka-213 I don't know how feasible that is. My guess is that ghcide implements these by looking at the global diagnostics that come back from GHC, and assigning them to code actions at the given srcspans. One way or another, you're going to need to wait for GHC to compute the holes, and that is always going to be superlinear in the number of holes.

In my eyes, the real bug here is that GHC is hella slow in computing these things. https://gitlab.haskell.org/ghc/ghc/-/issues/16875

As for the more general problem, @anka-213, you're the first person I've heard from who will actively miss this feature. Would you mind describing what a typical workflow looks like here? Are there hole fits you always ignore? Do you primarily use it for library combinators? Things defined in the same module? Things in local scope? Do you use it primarily for functions or values or both? How polymorphic are the hole fits you end up using?

Although I'd be sad to lose you as a user, the performance implications of this feature are IMO far too burdensome in an interactive application.

test/functional/FunctionalCodeAction.hs
pepeiborra: How difficult would it be to fix that?

anka-213:
As I said, I've never found the subsumption hole fits like `id _` useful at all, so I think we should disable those globally, regardless of if Wingman is enabled at all. They can always be simulated manually with `_ undefined`, `_ _` or `_ . _` if you expect a hole fit like that.

Most of the time, I think, they are not very polymorphic. I kind of use it as a type based autocomplete, so I don't have to remember what the function is called.
I think I've used all three variants of library combinators, definitions from the same module and values in local scope, but I'm not sure what's most important. I'll have to experiment a bit more to get a better appreciation for what my workflow is.

(I did once figure out that the answer was, as always, `traverse` by using hole fits.)

There are hole fits that I always ignore, e.g. #1804 (which is useful for creating a less polymorphic version of an existing function, like `traverse`). I think I always ignore functions like `id`, `$`, `const`, etc., but those mostly show up in subsumption hole fits. (The subsumption hole fits have made the feature worse)


----

What I think I would prefer for now is to allow enabling both at the same time with an option, with a warning that it will likely cause serious slowdown. When/if https://gitlab.haskell.org/ghc/ghc/-/issues/16875#note_210045 is implemented, we could switch to asking for them interactively.

If it would not require recompilation of more than the current module, another option could be to have an interactive command for saying "Compute/Enable hole fits once for this module". It does still have diminishing returns though, since if I have to wait a long time, I might as well figure out and the name of the function/value manually.

But yes, solving the slowness of hole completions would be by far the best solution and I wouldn't mind a slight reduction of functionality for that. One of the easiest (partial) solutions for that would be to completely disable hole fits on unconstrained holes like the one mentioned in https://gitlab.haskell.org/ghc/ghc/-/issues/16875#note_329814.

test/functional/FunctionalCodeAction.hs
isovector: I have no idea --- this stuff is deep in the bowels of ghcide.

pepeiborra: I think you only need to add a filter here: 

https://github.com/haskell/haskell-language-server/blob/f5bc01a5965f760452e6b957786ffa7bbe75e28e/ghcide/src/Development/IDE/Main.hs#L226

isovector: Yeah, I was thinking about that too. I'll give it a go today!

isovector:
I plan to merge this as-is later today, unless anyone objects particularly hard.

anka-213:
Here's a twitter thread of people praising the feature: https://twitter.com/kerckhove_ts/status/1403031394234818562
I'm not sure if this makes any difference, but at least I'm not the only one using it.

plugins/hls-tactics-plugin/README.md
```diff
@@ -34,6 +34,19 @@ fill hole" code action, *et voila!*
...
+
+When enabled, Wingman for Haskell will remove HLS support for hole-fit code
+actions. These code actions are provided by GHC and make typechecking extremely
+slow in the presence of typed holes. Because Wingman relies so heavily on typed
+holes, these features are in great tension.
+
+The solution: we just remove the hole-fit actions. They are mostly subsumed by
+Wingman, and usually make extremely unhelpful suggestions, such as `replace _
+with id _`. If you'd prefer to use these actions, you can get them back by
+disabling Wingman.
```
anka-213, line 47: The "usually make extremely unhelpful suggestions" is mostly caused by the refinement hole fits (and the lack of filtering out `id`, `const`, `$`, from the list, but that wouldn't even be necessary most of the time without refinement hole fits), so that piece documentation would be pretty misleading. If we just remove the refinement holes, the feature becomes 1000% better.

isovector: Yeah, but unfortunately removing just refinement holes doesn't solve the underlying issue. I suggested a feature in https://github.com/haskell/haskell-language-server/issues/1602#issuecomment-857819575 that would give us most of the best of both worlds.

anka-213: Losing the ability to discover things from external modules would severely cripple the feature, so I wouldn't quite see it as the best of both worlds, but at least it's better than nothing. :/

anka-213:
I'd argue fairly strongly for *allowing* the user to enable both at the same time, but not doing it by default (and having a big warning about it in the setting). But maybe that could be done in a later PR?

I just want the ability to do both hole fits and case splitting at the same time. I've never used any other feature of Wingman, so I don't know how useful they are. (I've tried "try to fill hole", but it never produced anything useful at all for me)

isovector:
I'd love some PRs improving the situation!

test/functional/FunctionalCodeAction.hs
pepeiborra: Did you try this ?

pepeiborra:
The main point of contention here is whether the disabling behaviour can be coupled to the Wingman plugin.

Wingman is enabled by default with HLS but we cannot presume that all the HLS users will prefer the Wingman features over the hole fit suggestions, so the least we can do is give them a choice. Until this is possible, I don't think we can merge this change.

isovector:
Apologies if this comes off as curt --- it's not intended to be --- but I'm feeling extremely frustrated here, for several reasons. 

1. There is an existing performance issue in the language server today, completely independent of Wingman (type holes are slow). The `TypeCheck` rule completely freezes up for _several seconds_ per hole, meaning no further actions can run.
2. Wingman exacerbates this issue by producing holes (eg, when case splitting), but short of emitting `undefined` instead, there is no alternative.
3. The existence of holes in any modules with 20+ imports make any sort of interactive flow with HLS untenably slow. 
4. Case splitting was an [_extremely well received feature_](https://github.com/haskell/haskell-language-server/pull/391)
5. In a universe where case splitting came first, it seems unlikely that we would have ever considered enabling hole fits due to the extreme performance regression.
6. Resolving the tension here seems to have fallen on my lap, but
7. This is outside of my domain expertise, and
8. Nobody else seems to be willing to contribute to help

My thought in getting this merged is that it resolves the tension --- if someone is particularly invested in getting hole-fits working, the burden can fall on them to get it working in a tenable fashion.

pepeiborra:
But you didn't address my point: can we couple this change to Wingman, in a way that disabling Wingman also disables this change? 

I suggested a way to address this here: https://github.com/haskell/haskell-language-server/pull/1873#discussion_r650420293

pepeiborra (changes requested):
Cannot merge until the change in behaviour is coupled to Wingman

test/functional/FunctionalCodeAction.hs
isovector: Gave this a go today, but ran up against #1980, which prevents us from using the config to filter this.

jneira (approved):

isovector:
The necessary changes are implemented in #2029

isovector:
@pepeiborra PTAL

plugins/hls-tactics-plugin/src/Wingman/StaticPlugin.hs
```diff
@@ -19,9 +19,16 @@ staticPlugin :: DynFlagsModifications
 staticPlugin = mempty
   { dynFlagsModifyGlobal =
       \df -> allowEmptyCaseButWithWarning
+           $ flip gopt_unset Opt_SortBySubsumHoleFits
            $ df
+             { refLevelHoleFits = Just 0
+             , maxRefHoleFits   = Just 0
+             , maxValidHoleFits = Just 0
```
pepeiborra, line 26: is this the correct set of options? I would think what you need is just:
```
disableHoleFits :: DynFlags -> DynFlags
disableHoleFits = flip gopt_unset Opt_ShowValidHoleFits
```

test/functional/FunctionalCodeAction.hs
```diff
@@ -421,6 +424,7 @@ typedHoleTests = testGroup "typed hole code actions" [
 
       , testCase "shows more suggestions" $
             runSession hlsCommand fullCaps "test/testdata" $ do
+                disableWingman
```
pepeiborra, line 427: Can we have a new test to check that Wingman really disables hole fits? 

Otherwise we will not know if we accidentally break the dynflags config.

pepeiborra (changes requested):

plugins/hls-tactics-plugin/src/Wingman/StaticPlugin.hs
isovector: That would be the equivalent of saying we don't have a preference, right? Perhaps that's equivalent to getting zero fits, but if so, it's only by accident.

isovector: Maybe not; I don't know. Whatever the situation, this gives the right result on my machine.

pepeiborra: I believe it means "disable all hole fit suggestions"

pepeiborra (approved):

isovector:
CI is OOMing. I don't know why, and I've spent enough time fixing broken CI this week. Any ideas @wz1000 @berberman @Ailrun @pepeiborra ?

pepeiborra:
@isovector I sent https://github.com/haskell/haskell-language-server/pull/2031 yesterday.

CI is very expensive to maintain, we support Cabal and Stack builds, a huge range of GHC versions, on 3 different platforms. It is a team effort and every help is welcome.

isovector:
@pepeiborra thanks for taking a look, I really appreciate it! Apologies if I came across as grumpy in my last message!

PR #1878, opened 2021-05-30
Rename implementation

OliverMadine:
A plugin to implement workspace-wide renaming of symbols as my [GSoC 2021 project](https://summerofcode.withgoogle.com/projects/#5971774971314176)!

### Todo List
- [x] Initial package structure
- [x] Test cases 
- [x] Find basic reference locations
- [x] Find correct references of type constructors
- [x] Find correct references of data constructors 
- [ ] Write any additional test cases (type families, record wildcards)
- [ ] Validate names
- [ ] [Index all files (hie-bios)](https://github.com/mpickering/hie-bios/issues/269)
- [x] Use retrie to rewrites right-hand sides
- [x] Limit retrie rewrite locations to references locations
- [x] Parse AST to rename left-hand sides
- [x] Type rewrites
- [x] Declaration rewrite
- [x] Qualified name rewrites
- [x] Update import lists
- [x] Disable plugin for ghc < 8.8
---

Example:
example


plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -0,0 +1,43 @@
...
+        goldenWithRename "type constructor" "TypeConstructor" $ \doc -> do
+            rename doc (Position 2 15) "BinaryTree" -- rotateRight :: Tree a -> Tree a
+    , goldenWithRename "data constructor" "DataConstructor" $ \doc -> do
+        rename doc (Position 0 13) "Apply" -- data Expr = Op Int Int
+    ,  ignoreTestBecause "not yet implemented" $
+        goldenWithRename "type variable" "TypeVariable" $ \doc -> do
+            rename doc (Position 0 13) "b" -- bar :: Maybe b -> Maybe b
+    , goldenWithRename "imported function" "ImportedFunction" $ \doc -> do
+        rename doc (Position 0 35) "baz" -- import           FunctionArgument (foo)
+    ]
```
Ailrun, line 37: It will be great if there are tests for GADTs, Type families, etc.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -0,0 +1,42 @@
...
+        _changeAnnotations = Nothing
+    })
+
+getTextEdits :: IdeState
+    -> TextDocumentIdentifier
+    -> Position
+    -> T.Text
+    -> LspT c IO [(Uri, List TextEdit)]
+getTextEdits state tdi pos name = do
+    mbLocs <- references state $ ReferenceParams tdi pos Nothing Nothing (ReferenceContext False)
```
pepeiborra, line 36: In a multi-project cradle, `references` only returns source locations for the projects that have been indexed. If not all the projects have been indexed, this will lead to incomplete renames. 

How can we address this? 

One way might be to grep the workspace for occurrences of the identifier and `uses GetHieAst` them or similar before calling `references` to ensure that they get indexed.

wz1000, line 36: References are reported for things like qualified names (the entire token `Data.Map.insert` is a reference to `insert`), are implicitly added on to record wildcards(the `{..}` in `Foo{..}` references all the fields of `Foo`) and perhaps more edge cases I'm missing. These will all need to be handled carefully.

OliverMadine: My initial idea was to use the `refMap` from the `HieAstResult` for the files of interest and reference from the DB otherwise. I've done this in my most recent push. Would this be sufficient?

OliverMadine: Currently, my solution to the qualified name problem is to use the length of the occurrence name to set the correct length of the `textEdit`https://github.com/haskell/haskell-language-server/blob/a80bf10ac07f2d804ce281831c9ebe3c163f9a58/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs#L94-L98
I hadn't thought about record wildcards yet thought, so I'll make sure to add a test for that shortly.
Thanks 🙏

pepeiborra: I suspect it won't, because it doesn't address my point - the DB gets populated when files are type checked, and in a multi-project cradle this only happens for the projects that are loaded.

Take the `GhcSessionDeps` data constructor, defined in `Development.IDE.Core.RuleTypes` and used in a bunch of HLS plugins. Can you rename all the occurrences successfully? If yes, then close your editor, wipe `.cache/ghcide`, restart your editor and try the rename again - does it rename all the occurrences this time?

wz1000: I think this is a limitation we will just have to live with. I've been asking @fendor for a long time on a way to enumerate all the components in a cradle.

wz1000: > One way might be to grep the workspace for occurrences of the identifier and uses GetHieAst them or similar before calling references to ensure that they get indexed.

I think this is not a very nice solution at all. For example, naively implemented it will raise all sorts of when we try to `GetHieAST` on files which can't be loaded due to not having a proper crade.

The real solution would be to work on hie-bios and define an interface to enumerate all the components and subprojects in a particular cradle.

wz1000:
Have to thought about augmenting something like `retrie` with the reference data to get semantics preserving transformations using exactprint?

Here is another case for you to consider(the language is full of these, you have to use exactprint to have any hope of being correct):

```haskell
foo = do bar
         baz
```

naivelly substituting `quxx` for `foo` breaks the parse:

```haskell
quux = do bar
         baz

```

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
pepeiborra: > I think this is a limitation we will just have to live with. I've been asking @fendor for a long time on a way to enumerate all the components in a cradle.

A feature that only does X% of the job, X% of the time is bound to generate a lot of support traffic, and I wouldn't want to be on the maintainer side of it.

> The real solution would be to work on hie-bios and define an interface to enumerate all the components and subprojects in a particular cradle.

Can you and @fendor draft the work needed here? Maybe @OliverMadine could take a look at it as part of the GSoc project.

pepeiborra:
> Have you thought about augmenting something like `retrie` with the reference data to get semantics preserving transformations using exactprint?
> 
> Here is another case for you to consider(the language is full of these, you have to use exactprint to have any hope of being correct):
> 
> ```haskell
> foo = do bar
>          baz
> ```
> 
> naively substituting `quxx` for `foo` breaks the parse:
> 
> ```haskell
> quux = do bar
>          baz
> ```

Yes, I agree, please consider hooking up retrie with hiedb. @xich has given powers to maintain and release retrie, so I can help review and land the changes

OliverMadine:
> Have you thought about augmenting something like `retrie` with the reference data to get semantics preserving transformations using exactprint?
> 
> Here is another case for you to consider(the language is full of these, you have to use exactprint to have any hope of being correct):
> 
> ```haskell
> foo = do bar
>          baz
> ```
> 
> naively substituting `quxx` for `foo` breaks the parse:
> 
> ```haskell
> quux = do bar
>          baz
> ```

> Yes, I agree, please consider hooking up retrie with hiedb. @xich has given powers to maintain and release retrie, so I can help review and land the changes

ok, I can see why it’s important that I work with retrie. I’ll look into doing this. Thank you both for the help getting started!

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
OliverMadine: I’d be happy to do work on this. A draft of the work would be much appreciated.

expipiplus1:
An alternative if the user is using a formatting program, would be to modify the AST and have the formatter reprint it. I assume that retrie is doing something similar under the hood.

pepeiborra:
@OliverMadine do how is this going? Do you mind sharing what you are working on?

OliverMadine:
> @OliverMadine do how is this going? Do you mind sharing what you are working on?

Hello,

I've been working on using retrie to perform the rewrites. I've pushed a basic version the manages limiting the rewrites to references (using `MatchResultTransformer`) and should allow rewrites that cause self-recursion (since this is not an issue with rename rewrites)

The main problem now is renaming declarations themselves.
For example:

when renaming `foo` to `bar`
`foo n = foo (n-1) + foo (n-2)`
we get
`foo n = bar (n-1) + bar (n-2)`

I'm looking at modifying the ast directly to do this.

As for hie-bios, I had a hacky approach that assumes the hie.yaml lists all components in a multi-cradle then `getCompilerOptions` can load them all simultaneously, however it looks like @fendor has been working on a[ much nicer solution](https://github.com/haskell/haskell-language-server/pull/2009#issue-686714343).

fendor:
@OliverMadine Your hacky solution is still welcome for supporting older cabal and stack versions!

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -0,0 +1,87 @@
...
+            nfp
+            True
+    return edits
+
+isReference :: [Location] -> MatchResultTransformer
+isReference refs _ctxt match
+  | MatchResult substitution template <- match
+  , Just loc <- srcSpanToLocation $ getOrigin $ astA $ tTemplate template -- Bug: incorrect loc
+  , loc `elem` refs = return match
+  | otherwise = return NoMatch
```
pepeiborra, line 66: This should return `match` if either `loc `elem` refs` **or** if the `loc` is inside the span of the declaration that we are trying to rename right?

pepeiborra:
> > @OliverMadine do how is this going? Do you mind sharing what you are working on?
> 
> Hello,
> 
> I've been working on using retrie to perform the rewrites. I've pushed a basic version that manages limiting the rewrites to references (using `MatchResultTransformer`) and allows rewrites of recursive calls (since this is not an issue with rename rewrites)
> 
> The main problem now is renaming declarations themselves.
> For example:
> 
> when renaming `foo` to `bar` in
> 
> ```haskell
> foo n = foo (n-1) + foo (n-2)
> ```
> 
> we get
> 
> ```haskell
> foo n = bar (n-1) + bar (n-2)
> ```
> 
> but we expect
> 
> ```haskell
> bar n = bar (n-1) + bar (n-2)
> ```
> 
> I'm looking at modifying the ast directly to do this.
> 
> As for hie-bios, I had a hacky approach that assumes the hie.yaml lists all components in a multi-cradle then `getCompilerOptions` can return all components to be loaded simultaneously, however, it looks like @fendor has been working on a[ much nicer solution](https://github.com/haskell/haskell-language-server/pull/2009#issue-686714343).
> 
> Sorry for the late response. I'll also try to update you more frequently from now on.

Fantastic, this is good progress. I left a comment on how to deal with renaming the declaration. 

I think that it would be a good idea to land this early version of the feature as soon as possible. If you don't want people to use it yet, change the flag in `haskell-language-server.cabal`to be disabled by default.

pepeiborra:
The lack of updates here is getting worrying. @OliverMadine Can you share the latest status?

OliverMadine:
> The lack of updates here is getting worrying. @OliverMadine Can you share the latest status?

I’ve had covid for the past week so it hasn’t been my most productive… I pushed a few small changes but my main issue is still renaming declarations.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
OliverMadine: I don’t think this works since the declaration isn’t initially matched. For example, even if we return match in every situation, the declaration would still not be rewritten. I’m not sure if there is any way of renaming the declaration with retrie?

pepeiborra:
> > The lack of updates here is getting worrying. @OliverMadine Can you share the latest status?
> 
> I’ve had covid for the past week so it hasn’t been my most productive… I pushed a few small changes but my main issue is still renaming declarations (I've explained why I don't think your previously suggested solution will work above).

Hope you are all right, and thanks for the update! I'll reply inline. 

While we look for a solution, it would be a good idea to rebase on top of master and fix conflicts so that we can get CI running and a green set of lights.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
pepeiborra: If the declaration isn't initially matched, then this means that Retrie does not alter left-hand-sides, only rhs's. Then the only solution is to rewrite the AST manually using exactprint, but that should be relatively easy.

pepeiborra:
I think you need to bump the hackage snapshot to include the newer version of retrie. @jneira whats the best way to do that?

jneira:
update the hackage index in both cabal.projects and update the retrie version in all stack.yaml's should do the work (and bump up version in .cabal files to ensure we got the right version)

OliverMadine:
> update the hackage index

The timestamp for the latest version of retrie is 2021-04-09T19:38:54Z so I shouldn't change the hackage index right? (since I would be decreasing it)

jneira:
yeah, if the index already set is newer it should allow get last retrie version

pepeiborra:
Hmm, the problem is that retrie 1.0.0.0 no longer supports 8.6.5:
> [__4] rejecting: retrie-1.0.0.0 (conflict: ghc==8.6.5/installed-8.6.5, retrie
53
=> ghc>=8.8 && <9.2)

You can either relax the retrie min bound or disable the plugin in 8.6 (both in the corresponding stack.yaml and in the cabal.project)

jneira:
Feeling a bit worried cause this uses retrie and it is not reliable in windows (as reminded by #2051)
I think that should not block this and the gsoc work but using it for a such basic feature make more prioritary fix retrie for windows.

haskell-language-server.cabal
```diff
@@ -447,6 +457,8 @@ test-suite func-test
     cpp-options: -Deval
   if flag(importLens) || flag(all-plugins)
     cpp-options: -DimportLens
+  if flag(rename) || flag(all-plugins)
```
pepeiborra, line 460: do you need an `impl(ghc > 8.8)` condition here ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -0,0 +1,319 @@
...
+
+descriptor :: PluginId -> PluginDescriptor IdeState
+descriptor pluginId = (defaultPluginDescriptor pluginId) {
+    pluginHandlers = mkPluginHandler STextDocumentRename renameProvider
+}
+
+renameProvider :: PluginMethodHandler IdeState TextDocumentRename
+renameProvider state pluginId (RenameParams (TextDocumentIdentifier uri) pos _prog newNameText) = response $ do
+    nfp <- safeGetNfp uri
+    oldName <- head <$> getNamesAtPos state pos nfp
```
pepeiborra, line 43: `head` will crash if there are no names at pos, please fix

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -355,7 +368,21 @@ callRetrie ::
   NormalizedFilePath ->
   Bool ->
   IO ([CallRetrieError], WorkspaceEdit)
-callRetrie state session rewrites origin restrictToOriginatingFile = do
+callRetrie =
+  callRetrieWithTransformerAndUpdates defaultTransformer updateContext
+
+-- allows custom 'ContextUpdater' to be given to 'applyWithUpdates'
```
pepeiborra, line 374:
```suggestion
-- | allows custom 'ContextUpdater' to be given to 'applyWithUpdates'
```

stack-8.6.4.yaml
```diff
@@ -16,6 +16,7 @@ packages:
   - ./plugins/hls-explicit-imports-plugin
   - ./plugins/hls-refine-imports-plugin
   - ./plugins/hls-hlint-plugin
+  - ./plugins/hls-rename-plugin
```
pepeiborra, line 19: Since we don't expect the plugin to build with 8.6.x, should it be removed from this stack.yaml file?

stack-8.6.5.yaml
```diff
@@ -15,6 +15,7 @@ packages:
   - ./plugins/hls-explicit-imports-plugin
   - ./plugins/hls-refine-imports-plugin
   - ./plugins/hls-hlint-plugin
+  - ./plugins/hls-rename-plugin
```
pepeiborra, line 18: Since we don't expect the plugin to build with 8.6.x, should it be removed from this stack.yaml file?

stack-9.0.1.yaml
```diff
@@ -16,6 +16,7 @@ packages:
   - ./plugins/hls-explicit-imports-plugin
   # - ./plugins/hls-refine-imports-plugin
   - ./plugins/hls-hlint-plugin
+  # - ./plugins/hls-rename-plugin
```
pepeiborra, line 19: is the rename plugin not compatible with 9.0? What's blocking it?

pepeiborra (changes requested):
Looks good! I have left a couple of comments requesting minor changes. 

In addition to those, I think you need to register the plugin test suite in the GitHub CI explicitly, see https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/test.yml

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -0,0 +1,318 @@
...
+                }
+        renameLhsDecl (TyClD xTy dataDecl@DataDecl{tcdLName, tcdDataDefn = hsDataDefn@HsDataDefn{dd_cons}})
+            = TyClD xTy $ dataDecl {
+                    tcdLName = renameRdrName' tcdLName
+                }
+        renameLhsDecl (TyClD xTy synDecl@SynDecl{tcdLName})
+            = TyClD xTy $ synDecl {
+                    tcdLName = renameRdrName' tcdLName
+                }
+        renameLhsDecl decl = decl
```
pepeiborra, line 138: Did you consider SYB for doing this rewrite, just like wingman does? 

That way, if newer GHC versions change the AST, this code will still work.

stack-9.0.1.yaml
OliverMadine: There were just some small issues. It is now working

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
OliverMadine: I think this would be a little difficult since there are a lot of names in the decls that we do not want to rename (either because retrie is already renaming the given reference or we don't support renaming for the given type of name). 

However, I could take this approach (i.e. modifying the entire source directly and not using retrie). See my comment below.

OliverMadine:
@pepeiborra 

Any thoughts on modifying the entire parsed source directly (using SYB rather than using Retrie)?

[Here’s a branch that takes this approach.](https://github.com/OliverMadine/haskell-language-server/blob/rename-plugin-directly-parse-source/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs)
It is passing all of the tests locally.

The advantages seem to be:
- Compatibility with earlier ghc versions
- No issues with retrie compatibility on windows
- We can rename non top-level names
- Easier solution to maintain
 
The disadvantage would be that it is slower. Possibly significantly?
Maybe I’m missing something else?

plugins/hls-rename-plugin/hls-rename-plugin.cabal
```diff
@@ -0,0 +1,53 @@
...
+    , hls-retrie-plugin >= 1.0.1.1
+    , text
+    , unordered-containers
+    , containers
+    , ghc
+    , hiedb
+    , ghc-exactprint
+
+  if impl(ghc >= 8.8)
+    build-depends: retrie >=1.0.0.0
```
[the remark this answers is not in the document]
OliverMadine: I forgot to remove this from an [old commit](03693f2), thanks.

pepeiborra:
> @pepeiborra 
> 
> 
> 
> Any thoughts on modifying the entire parsed source directly (using SYB rather than using Retrie)?
> 
> 
> 
> [Here’s a branch that takes this approach.](https://github.com/OliverMadine/haskell-language-server/blob/rename-plugin-directly-parse-source/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs)
> 
> It is passing all of the tests locally.
> 
> 
> 
> The advantages seem to be:
> 
> - Compatibility with earlier ghc versions
> 
> - No issues with retrie compatibility on windows
> 
> - We can rename non top-level names
> 
> - Easier solution to maintain
> 
>  
> 
> The disadvantage would be that it is slower. Possibly significantly?
> 
> Maybe I’m missing something else?
> 
> 

I like it. Retrie uses SYB as well under the hood, so I don't expect this solution to be slower. A rename is the trivial rewrite where both the lhs and the rhs are a simple name, so we don't really need the full power of retrie.

OliverMadine:
I'm opening a [new PR](#2108) due to the change of approach (and to clean the mess of commits I made while learning how to make the test cases consistent).

samhh:
Fixed link^: https://github.com/haskell/haskell-language-server/pull/2108

PR #1880, opened 2021-05-31
Use file watches for all workspace files

pepeiborra:
This PR hooks up the `GetModificationTime` rule to install new file watchers for workspace files that are not already covered by the initial glob that we install, consolidating the tracking of workspace changes. There are two main use cases for this:

1. Cradle dependencies, including the `hie.yaml` file and the cabal/stack descriptors
2. TH dependencies, defined using `addDependentFile`.

The motivation for this change is not just consolidation or performance. This is required in #1862 to be able to track changes in the use cases listed above.


ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -290,3 +309,43 @@ setSomethingModified state = do
...
+          watchKind = WatchKind { _watchCreate = True, _watchChange = True, _watchDelete = True}
+          -- See Note [Which files should we watch?] for an explanation of why the pattern is the way that it is
+          -- The patterns will be something like "**/.hs", i.e. "any number of directory segments,
+          -- followed by a file with an extension 'hs'.
+          watcher glob = FileSystemWatcher { _globPattern = glob, _kind = Just watchKind }
+          -- We use multiple watchers instead of one using '{}' because lsp-test doesn't
+          -- support that: https://github.com/bubba/lsp-test/issues/77
+          watchers = [ watcher (Text.pack glob) | glob <- globs ]
+
+        void $ LSP.sendRequest LSP.SClientRegisterCapability regParams (const $ pure ()) -- TODO handle response
```
wz1000, line 338: Ideally we would block waiting for the response return and return a result depending on whether it was successful.

wz1000 (approved):

ghcide/src/Development/IDE/Core/FileStore.hs
pepeiborra: This is an old TODO, it's not something added in this PR. I agree that we should wait for the response, but maybe not for this PR

pepeiborra: BTW I checked this and lsp-test doesn't seem to implement this part of the protocol, opened https://github.com/haskell/lsp/issues/333

PR #1881, opened 2021-06-03
Let Wingman peek through type families

### LINKED ISSUE #1874: Wingman can't split type family
```haskell
{-# LANGUAGE TypeFamilies #-}

module Lib where

type family Yo where
    Yo = Bool

test :: Yo -> Int
test b = _
```

Running `Destruct b` at the hole prints an internal error, I guess because there are no data constructors. Pretty sure this used to work, so not sure when it regressed.

isovector:
This PR fixes a bug where type families would be opaque to Wingman, even if they weren't stuck. Now, we normalize any types coming into the judgment.

Fixes #1874

### DIFF
--- a/plugins/hls-tactics-plugin/test/golden/DestructTyFam.hs
+++ b/plugins/hls-tactics-plugin/test/golden/DestructTyFam.hs
@@ -1,0 +1,8 @@
+{-# LANGUAGE TypeFamilies #-}
+
+type family Yo where
+  Yo = Bool
+
+test :: Yo -> Int
+test b = _
+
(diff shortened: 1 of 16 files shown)


plugins/hls-tactics-plugin/test/golden/DestructTyFam.hs
```diff
@@ -0,0 +1,8 @@
+{-# LANGUAGE TypeFamilies #-}
+
+type family Yo where
+  Yo = Bool
+
+test :: Yo -> Int
+test b = _
+
```
Ailrun, line 8: Does this work with open type family too?

isovector: Added some tests

Ailrun: Awesome, thanks!

Ailrun (approved):
LGTM :)

PR #1888, opened 2021-06-04
[explicit-imports] Take in a predicate to filter modules

pepeiborra:
This is intended for custom LSP servers and generalises the current filtering of Prelude


Ailrun (approved):
Thanks :)

PR #1889, opened 2021-06-04
Wingman: FIx evidence when using GADT constructors

### LINKED ISSUE #1689: Wingman doesn't introduce new skolems for existentially bound variables
@masaeedu points out:

```haskell
{-# LANGUAGE DataKinds, GADTs #-}

data Z ab where
  Z :: (a -> b) -> Z '(a, b)

test :: Z ab
test = _
```

produces

```haskell
test = Z id
```

But it doesn't typecheck.

What's going wrong here? When we introduce the `Z` constructor, we should generate two new skolems `a` and `b` such that `'(a, b) ~ ab`. But instead we generate two new unification variables, which are thus free to unify with one another, and thus produce the bogus `id`.

isovector:
This PR fixes #1689, by creating new skolems and using created type equality evidence when building GADT data cons.


plugins/hls-tactics-plugin/test/golden/MessageCantUnify.hs
```diff
@@ -0,0 +1,8 @@
+{-# LANGUAGE DataKinds, GADTs #-}
+
+data Z ab where
+  Z :: (a -> b) -> Z '(a, b)
+
+test :: Z ab
+test = _
+
```
Ailrun, line 8: It looks like the golden file for this test is missing

Ailrun (approved):

PR #1894, opened 2021-06-05
Try using haskell/lsp:master

anka-213:
This PR is testing/demonstrating the test failures caused by: https://github.com/haskell/lsp/pull/326
Issue #1895 is for fixing them.

See also https://github.com/haskell/haskell-language-server/pull/1649#issuecomment-855247288

PR #1897, opened 2021-06-06
Fix getCurrentDirectory calls in ghcide

pepeiborra:
lsp HEAD no longer sets the current directory to the workspace root (https://github.com/haskell/lsp/pull/326) so we need to remove this assumption from the ghcide codebase. There are 3 places where this is relevant:

1. In the call to `loadCradle`, or rather, in the GHC session returned by it, where any relative paths in the flags returned by the cradle will expect to be resolved from the directory where the `hie.yaml` file lives - should change the working directory to the hie.yaml location
2. In the logic to `setInitialDynFlags`, where we search for a `hie.yaml` file in the current directory - should use the workspace root instead
3. In the LSP initialisation handler, where the current directory is included in the hiedb cache dir hash - should use the workspace root instead

### DIFF
--- a/ghcide/src/Development/IDE/Main.hs
+++ b/ghcide/src/Development/IDE/Main.hs
@@ -18,8 +18,7 @@ import qualified Data.HashMap.Strict                   as HashMap
 import           Data.Hashable                         (hashed)
 import           Data.List.Extra                       (intercalate, isPrefixOf,
                                                         nub, nubOrd, partition)
-import           Data.Maybe                            (catMaybes, fromMaybe,
-                                                        isJust)
+import           Data.Maybe                            (catMaybes, isJust)
 import qualified Data.Text                             as T
 import qualified Data.Text.IO                          as T
 import           Development.IDE                       (Action, Rules,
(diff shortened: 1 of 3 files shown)


pepeiborra:
#1894 is open to test these changes with lsp HEAD

ghcide/src/Development/IDE/Main.hs
```diff
@@ -206,14 +206,14 @@ defaultMain Arguments{..} = do
                 t <- t
                 hPutStrLn stderr $ "Started LSP server in " ++ showDuration t
 
-                dir <- IO.getCurrentDirectory
+                dir <- maybe IO.getCurrentDirectory return rootPath
 
                 -- We want to set the global DynFlags right now, so that we can use
                 -- `unsafeGlobalDynFlags` even before the project is configured
                 -- We do it here since haskell-lsp changes our working directory to the correct place ('rootPath')
                 -- before calling this function
```
berberman, line 214: I think this comment needs to be updated, though it still will be true until we update `lsp`.

Also, `fromMaybe` in L220 can be simplified.

berberman (approved):
LGTM

ghcide/src/Development/IDE/Main.hs
pepeiborra: good catch, thanks!

Issue #1898, opened 2021-06-06
Support GHC 8.10.5


Ailrun:
[GHC 8.10.5 is out](https://www.haskell.org/ghc/download_ghc_8_10_5.html), and it fixed [the thread-safety issue](https://gitlab.haskell.org/ghc/ghc/-/issues/19417) mentioned in https://github.com/haskell/haskell-language-server/issues/1863#issuecomment-846755647~~, so I think now we can replace 8.10.2 with 8.10.5~~.

Unfortunately, we cannot remove 8.10.2 because of https://gitlab.haskell.org/ghc/ghc/-/issues/19968. I will open a separate issue for 8.10.6 support and 8.10.2 removal.

Ailrun:
We cannot remove 8.10.2 until 8.10.6 is out, as MacOS has a build problem with GHC 8.10.5

jneira:
@Ailrun is there some ghc ticket about? it would be great to trackit, to ensure 8.10.6 will have a fix for macos

Ailrun:
Here is the link: https://gitlab.haskell.org/ghc/ghc/-/issues/19968

Ailrun:
Closed by #1899. For the GHC 8.10.2 issue and GHC 8.10.6 support, please see #1913.

mouse07410:
Could you clarify, please? It looks like HLS-1.2.0 does **not** support GHC-8.10.5 on MacOS yet?

<img width="564" alt="Screen Shot 2021-06-15 at 10 44 10" src="https://user-images.githubusercontent.com/5923577/122074541-4e40fd00-cdc7-11eb-8024-41b241ea7fa5.png">

Ailrun:
@mouse07410 
From 1.2.0 release note:
> - Support for GHC 8.10.5 is added.
Note that macOS version is unfortunately not included in this release because of [a GHC issue with network package](https://gitlab.haskell.org/ghc/ghc/-/issues/19968).

mouse07410:
Please see https://github.com/haskell/haskell-language-server/issues/1913

Issue #1465, opened 2021-03-01
Frequent crashes loading Hasktorch


tscholak:
This problem is new for me. I just finished updating my Hasktorch dev environment by bumping the version of ghc to 8.10.4. Before I was using hls 0.7 with Hasktorch, now I've tried both 0.9 and 1.0, but both are suddenly very unstable.

### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
haskell-language-server version: 1.0.0.0 (GHC: 8.10.4) (PATH: /nix/store/wsrhqfn07rif6n27qglc3p68kjaflyax-haskell-language-server-exe-haskell-language-server-1.0.0.0/bin/haskell-language-server)
Tool versions found on the $PATH
cabal:		3.2.0.0
stack:		2.5.1
ghc:		8.10.4
```

Which lsp-client do you use:
vs code

Describe your project (alternative: link to the project):
https://github.com/hasktorch/hasktorch/blob/master/cabal.project

Contents of `hie.yaml`:
https://github.com/hasktorch/hasktorch/blob/master/hie.yaml

### Steps to reproduce
start vscode from nix shell, work for a few seconds on random things, crash occurs

### Expected behaviour
no crash

### Actual behaviour
crash

### Include debug information
Execute in the root of your project the command `haskell-language-server --debug .` and paste the logs here:

<details>
<summary>
Debug output:
</summary>

the full output is too long to post here. it finishes after a while with:
```
Completed (36 files worked, 298 files failed)
```
</details>

Paste the logs from the lsp-client, e.g. for [VS Code](https://github.com/haskell/vscode-haskell#troubleshooting)

<details>
<summary>
LSP logs:
</summary>

likewise, this log is too long. I'll upload a file...
</details>

konn:
I encountered this issue several times in a closed in-house codebase, although I couldn't make minimal repro.
Telling from situational evidence in our project, it seems that modules with type-checker plugins can make HLS 1.0 crash, but no concrete evidence.

tscholak:
Thanks for these pointers.
It would be a surprise to me, though. Have hls 0.9 and 1.0 become less stable under the presence of these plugins? I'm a longtime user of hls with hasktorch, and with 0.7 and ghc 8.10.3 I did not have such severe stability problems.

wz1000:
[...]

PR #1899, opened 2021-06-06
Add GHC 8.10.5 support

Ailrun:
This PR
- ~~drops 8.10.2 support~~ (We need to wait until 8.10.6)
- adds 8.10.5 support

~~Resolves~~ Related to #1898, ~~probably resolves~~ related to #1465

~~CI fails until https://github.com/hsyl20/ghc-api-compat/pull/11 is resolved (If it takes too long, I will redirect the dependency to PR head)~~ I redirected the dependency


.github/workflows/build.yml
```diff
@@ -81,7 +75,6 @@ jobs:
     - name: Compress Server Binary
       id: compress_server_binary
       run: |
-        # We normalize windows+choco ghc version 8.10.2.2
```
Ailrun, line 84: Which part was the normalization? Was it `GHC_VER=$(echo $GHC_VER | sed 's/8.10.2.2/8.10.2/g')` part?

jneira: yep, not sure if 8.10.5 is in the choco package yet 🤔

Ailrun: @jneira it's unfortunately not in the list... I added a comment on the package

jneira: it still is not

Ailrun: I also sent a mail to the maintainer. I hope this will be resolved soon...

jneira: There is a [github repo](https://github.com/Mistuke/GhcChoco), i usually open issues about there

Ailrun: Oh I didn't know that, I will make a PR

Ailrun: @jneira Actually the maintainer was faster than my PR :) Now there's GHC 8.10.5

Ailrun:
Maybe blocked by upstream? https://gitlab.haskell.org/ghc/ghc/-/issues/19763#note_357220

.github/workflows/test.yml
```diff
@@ -33,18 +33,15 @@ jobs:
...
         include:
           # only test supported ghc major versions
           - os: ubuntu-latest
             ghc: '9.0.1'
             test: true
           - os: ubuntu-latest
-            ghc: '8.10.4'
-            test: true
-          - os: windows-latest
-            ghc: '8.10.4'
```
Ailrun, line 47: Windows 8.10.4 should not be removed

jneira:
We should estimate the impact of the bug, but i suppose we will have to add support for 8.10.5, as people is gonna use it anyways :face_with_head_bandage: 
It would suppose mark tests as broken for ghc 8.10.5, it will let us make progress and help us to estimate the impact at same time.

jneira:
To try to avoid this scenario i think we should build and test hls against ghc branches that will serve as base for minor releases (8.10 and 9.0). This will help ghc developers to catch them earlier too.

Ailrun:
@jneira It looks like that GHC plugin is the only case that causes the problem in our test set, so I think it's not the worst situation.

Ailrun:
I don't know why but MacOS build stucks after completed Brittany install...

konn:
It gets stuck also on my laptop (MacBook Pro), preprocessing the `network` package forever...

Ailrun:
@konn so GHC 8.10.5 on Mac is kinda... broken? (as it does not work with both Cabal and Stack)

konn:
@Ailrun I haven't done any deep investigation, but it seems there is a problem, I'm afraid...

Ailrun:
@konn Could you check whether the network package alone causes the same issue? If so we can report a more specific issue on GHC

konn:
@Ailrun Ah, I've just confirmed that building the network package alone causes the same issue, and reported it to GHC GitLab: https://gitlab.haskell.org/ghc/ghc/-/issues/19968.

Ailrun:
Awesome, thank you for the fast process!

Ailrun:
Should we leave 8.10.2 until GHC 8.10.6 is out? Especially for MacOS... I disabled MacOS CI and build, and I don't know it's possible to generate a binary in MacOS with GHC 8.10.5

jneira:
> Should we leave 8.10.2 until GHC 8.10.6 is out

I think we should be cautious and keep it some time more

jneira (approved):
I think we could merge this to make it available to user, which are being used it already.
Great work as usual, many thanks 🙂

Ailrun:
@jneira Yes, now we should change the required test to Windows 8.10.5 with test and Windows 8.10.4 without test. I will change them.

PR #1905, opened 2021-06-09
Cleanup stack build output in circleci

jhrcek:
Figuring out stuff from circleci logs is currently pretty annoying due to stack's ansi terminal progress output:

Screenshot from 2021-06-09 19-05-53

This are great when running stack interactively, but they clutter output in CI.
This behavior can be disabled by using `--no-terminal`. From `stack build --help`:

```
  --[no-]terminal          Enable/disable overriding terminal detection in the
                           case of running in a false terminal
```


Ailrun (approved):
Great, thank you for the update!

Issue #1890, opened 2021-06-05
June 2021 Monthly Release


Ailrun:
I think we have quite a number of PRs merged, bugs fixed, and features improved, so it will be great to have this month's official release.
Do you think any PR should be merged before the release? If so, please put it in a comment.

pepeiborra:
Let's try to land #1649 (GHC 9.0 support) this weekend. @anka-213 has put in a lot of work and it's almost ready to merge.

Ailrun:
I'm waiting for GHC 8.10.5 PR (#1899)

Ailrun:
With #1912 and #1919, the release is done.

PR #1912, opened 2021-06-11
Prepare 1.2.0 June release

Ailrun:
Resolves partially #1890


ChangeLog.md
```diff
@@ -1,5 +1,212 @@
 # Changelog for haskell-language-server
 
+## 1.2.0
+
+We finally release a new version of Haskell Language Server!
```
isovector, line 5: This sentence threw me for a loop the first few times I read it. Maybe something like this instead?

```suggestion
We have finally released a new version of Haskell Language Server!
```

isovector: or if you want something more active

```suggestion
We are finally releasing a new version of Haskell Language Server!
```

ChangeLog.md
```diff
@@ -12,9 +12,14 @@ Here are the summary of changes:
...
-  ![https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-tactics-plugin/COMMANDS.md#wingman-metaprogram-command-reference](https://user-images.githubusercontent.com/307223/118190278-bdf24f80-b3f7-11eb-8838-b08a2582d7f1.gif)
+- Wingman plugin has added numerous features and fixed many bugs:
+  - It now supports tactic metaprogramming!  
+    For list of commands, see [this document](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-tactics-plugin/COMMANDS.md#wingman-metaprogram-command-reference).  
+    ![https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-tactics-plugin/COMMANDS.md#wingman-metaprogram-command-reference](https://user-images.githubusercontent.com/307223/118190278-bdf24f80-b3f7-11eb-8838-b08a2582d7f1.gif)
+  - "Refine hole" and "Split all function arguments" code actions are publicly opened.
+  - "Empty case split" code lens is added.
+  - The name generator is fixed [to avoid dangerous summon rituals](https://github.com/haskell/haskell-language-server/pull/1771).
```
isovector, line 21: :+1:

isovector (approved):
:shipit:

pepeiborra:
Should bump the versions of ghcide, hls-graph and his-plugin-api as well in preparation for the Hackage upload?

Ailrun:
@pepeiborra Oh, right. I will bump those too.

hls-test-utils/hls-test-utils.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 2.4
 name:          hls-test-utils
-version:       1.0.0.1
+version:       1.0.1.0
```
berberman, line 3: For the version `a.b.c.d`, in which case should we bump `c` instead of `d`?

Ailrun: For this package, I bump `c` as this exposes a new binding, following the PVP:
> *Non-breaking change.* Otherwise, if only new bindings, types, classes, non-orphan instances or modules (but see below) were added to the interface, then *A.B* **MAY** remain the same but the new *C* **MUST** be greater than the old *C*. Note that modifying imports or depending on a newer version of another package may cause extra non-orphan instances to be exported and thus force a minor version change.

For other cases, I believe I bump `c` part only for packages that expose different behaviors. If there is any mistake, please let me know.

plugins/hls-module-name-plugin/hls-module-name-plugin.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               hls-module-name-plugin
-version:            1.0.0.0
+version:            1.0.0.1
```
berberman, line 3: We haven't had `1.0.0.0` on Hackage, do we need this bump?

plugins/hls-ormolu-plugin/hls-ormolu-plugin.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               hls-ormolu-plugin
-version:            1.0.0.0
+version:            1.0.0.1
```
berberman, line 3: Same here

plugins/hls-refine-imports-plugin/hls-refine-imports-plugin.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      2.2
 name:               hls-refine-imports-plugin
-version:            1.0.0.0
```
berberman, line 3: Same here

berberman (commented):
Also, I think we should bump `hls-brittany-plugin` to `1.0.0.2`

plugins/hls-module-name-plugin/hls-module-name-plugin.cabal
Ailrun: Oh, really? I thought I released one... I will revert these.

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -48,7 +48,7 @@ library
     , extra
     , filepath
     , ghc-exactprint        >=0.6.3.4
-    , ghcide                >=1.2     && <1.4
+    , ghcide                ^>=1.4
```
Ailrun, line 51: This changes because hlint plugin probably won't work with ghcide <= 1.3 after #1862

pepeiborra (approved):

hasufell:
I don't see the `.tar.gz` files that are usually used by ghcup.

Ailrun:
@hasufell It's not yet built. It will be built after all other version-specific `.gz` files are done.

hasufell:
the tar job seems to be failing: https://github.com/haskell/haskell-language-server/runs/2813407586?check_suite_focus=true

Ailrun:
Yes, I'm resolving the issue.

hasufell:
Also... since ghcup now supports windows too, we'd need a  `.tar.gz` for windows as well.

Ailrun:
@hasufell Ah... true. @jneira Could you generate it manually? I will add a CI entry.

jneira:
sure, will do this night gmt

jneira:
As usual i have to generate manually the version for 8.8.4, the only one we support for windows and it will take some time in my poor laptop

anka-213:
Should we add the ghc-9.0 build as well to the release, so users doesn't have to build manually from source to get the basic ghc-9.0 support? Or is it to late now?

jneira:
It is a good idea, if the executable can be used without tweaking too much configuration. A note should be added in the release, with a warning for the limited support and linking to the list of working components.
Not sure if we should change the file name to signal it is experimental and with limited support and avoid its automatic download by the vscode extension.
As stated in the [new release doc version](https://github.com/haskell/haskell-language-server/pull/1935) we can add new artifacts but no update old ones.

anka-213:
No, there shouldn't be much configuration tweaking needed. Adding the flag `--cabal-project=cabal-ghc901.project` or copying the `cabal-ghc901.project` to `cabal.project` should be all that's needed.

I don't think disabling automatic download should be needed, as long as it is clear to the user that it is experimental, since partial support is better than no support at all.

PR #1916, opened 2021-06-13
Fix `build` action for macOS

Ailrun:
We don't have the macOS version for HLS 8.10.5, which broke the assumption of `build` action for releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,11 +115,11 @@ jobs:
         path: ${{ steps.compress_server_binary.outputs.path }}
 
     - name: Build Wrapper
-      if: matrix.ghc == '8.10.5'
+      if: matrix.ghc == '8.10.5' && matrix.os != 'macOS-latest' || matrix.ghc == '8.10.4' && matrix.os == 'macOS-latest'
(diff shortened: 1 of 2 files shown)


jneira (approved):
lgtm

Ailrun:
Currently benchmark fails for some unknown reason, so I will merge this for sake of the release.

.github/workflows/build.yml
```diff
@@ -115,11 +115,11 @@ jobs:
         path: ${{ steps.compress_server_binary.outputs.path }}
 
     - name: Build Wrapper
-      if: matrix.ghc == '8.10.5'
+      if: matrix.ghc == '8.10.5' && matrix.os != 'macOS-latest' || matrix.ghc == '8.10.4' && matrix.os == 'macOS-latest'
```
berberman, line 118: Could we simply use GHC 8.10.4 to build the wrapper for all platforms until the next stable GHC version released?

Ailrun: Oops, thank you for the review. Maybe that's better as GHC 8.10.5 also does not expose `allocateWrite` symbol (which causes some GHC-plugin failures).

PR #1918, opened 2021-06-13
Fall back to hiedb for invalid srcspan paths

pepeiborra:
When reusing .hie files from a cloud cache, the paths may not match the local file system. 

Let's fall back to the hiedb in case it contains local paths


wz1000 (approved):
Makes sense.

PR #1920, opened 2021-06-14
Allow symbol identifiers in tactics

isovector:
The metaprogram parser used to choke on symbol identifiers, which meant it was impossible to ever reference things like `(<>)`. This works correctly now.


Ailrun (approved):
LGTM

PR #1921, opened 2021-06-14
Add pointwise command to the metaprogram parser

isovector:
This PR adds the `pointwise` command, which restricts the hypotheses available to holes so that they line up with top-level destructed terms.


jneira (approved):
lgtm, thanks for your tactical maneuvers

Issue #1913, opened 2021-06-12
Remove GHC 8.10.2, 8.10.3, 8.10.4 and 8.10.5 (after the hls release including GHC 8.10.7)


Ailrun:
Currently, GHC 8.10.5 in macOS experiences https://gitlab.haskell.org/ghc/ghc/-/issues/19968 in our CI (and in @konn's Mac), so we cannot replace GHC 8.10.2 with GHC 8.10.5 as planned in #1898.

We may be able to remove GHC 8.10.2 after GHC 8.10.6 with the fix above.

mouse07410:
The mentioned issue only surfaces if Xcode **Command Line Tools** package is installed. On machines with Xcode.app only (which also provides command line tools!) this issue is not present.

Ailrun:
@mouse07410 Thank you for the information! @konn, could you check you can build HLS without that package? Or is that package necessary for some other things?

mouse07410:
> .  .  .  build HLS without that package?

Trivial - with one caveat. On MacOS you need *either* `Xcode.app` *or* CLT installed. 

> Or is that package necessary for some other things?

No, CLT package is not necessary - unless one insists on **not** installing `Xcode.app`.

Something has to provide Clang, header files, and libraries - they come either from Xcode, or. from CLT. It turns out that CLT (for reasons beyond my understanding) is often (always?) more buggy and less usable that the "main" app that also adds GUI IDE...

jneira:
Mmm, we are using github action envs to build the release and it failed so we could check if that library is installed there.
In that case we could:
* try to change the libraries installed in the images used by gha
* build locally the version in some macos and upload it manually to the release (like we do with 8.8.4 and windows)

mouse07410:
>  we are using github action envs to build the release

Unfortunately, I've zero experience/expertise on GHA. But with some digging, it seems that GHA provides
```
Xcode Command Line Tools 12.4.0.0.1.1610135815
```

But it also seems to have Homebrew Clang-12 and Xcode_12.4.app - see the screenshot below.


![Screen Shot 2021-06-15 at 12 50 32](https://user-images.githubusercontent.com/5923577/122092755-4b023d00-cdd8-11eb-8b4a-7f1f479b9ca7.png)

![Screen Shot 2021-06-15 at 12 48 30](https://user-images.githubusercontent.com/5923577/122092523-04acde00-cdd8-11eb-86d5-240ef7b03ffa.png)

[...]

PR #1931, opened 2021-06-16
Workaround for GHC 8.10.5 on macOS

konn:
As pointed out by @mouse07410 in #1913, we can build HLS with GHC 8.10.5 on macOS just by removing CommandLineTools from the environment.
This PR tries to adopt this workaround in a rather ad-hoc manner.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,8 +74,6 @@ jobs:
 
     - name: Tentative Workaround for GHC 8.10.5 on macOS
       if: matrix.os == 'macOS-latest' && matrix.ghc == '8.10.5'
-      env:
-        GHC_VER: ${{ matrix.ghc }}
       run: |
         echo "# uninstalling CommandLineTools (see https://github.com/haskell/haskell-language-server/issues/1913#issuecomment-861667786)"
         sudo rm -rf /Library/Developer/CommandLineTools
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,8 +27,6 @@ jobs:
             ghc: '8.8.3' # fails due to segfault
           - os: windows-latest
             ghc: '8.8.2' # fails due to error with Cabal
-          - os: macOS-latest
-            ghc: '8.10.5' # https://gitlab.haskell.org/ghc/ghc/-/issues/19968
         include:
           - os: windows-latest
             ghc: '8.10.2.2' # only available for windows and choco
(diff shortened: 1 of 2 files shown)


jneira:
> 
> 
> I want to test build action for GHC 8.10.5 + macOS - is there any way to invoke the build workflow manually?

It is triggered on releases so i usually test them in my repo, creating a fake release there and delete it after the test.
(We could change the trigger in the script and restore it after the test but i would recommend test it in your repo, you could link to your succesful build for completeness)

konn:
Here it goes! :rocket: https://github.com/konn/haskell-language-server/runs/2837481049?check_suite_focus=true

konn:
Hurray! It at least builds the `network` with GHC 8.10.5 on macOS 🎉

スクリーンショット 2021-06-16 18 02 31

Ailrun:
Awesome! Now we can have GHC 8.10.5 on macOS for future versions too :)

.github/workflows/build.yml
```diff
@@ -74,6 +72,12 @@ jobs:
       run: |
         echo "GHC_VERSION=$GHC_VER" >> $GITHUB_ENV
 
+    - name: Tentative Workaround for GHC 8.10.5 on macOS
+      if: matrix.os == 'macOS-latest' && matrix.ghc == '8.10.5'
+      run: |
+        echo "# uninstalling CommandLineTools (see https://github.com/haskell/haskell-language-server/issues/1913#issuecomment-861667786)"
+        sudo rm -rf /Library/Developer/CommandLineTools
+
```
Ailrun, line 80: Could you also apply the similar trick to test CI so that we can test 8.10.5 build on macOS for each PR?

konn: Good point! Added to test.yml.

konn:
OK, it finally generated HLS binary for GHC 8.10.5 on macOS 🎉

konn:
I will re-upload the HLS exes generated by GHA for macOS to haskell/haskell-language-server after the test CI successfully passes.

jneira (approved):
Amazing work, many thanks, macOS users will be grateful

konn:
The hack incoporated in this pull request is essentially due to @mouse07410.
Credit goes to them, great discovery!

konn:
Oops, tests got cancelled by ghc 9 failure -- is this intended behaviour?

konn:
Ah, I used the wrong commit to build 1.2.0 on my fork repository. This Pull Request itself is correct, but I will re-run the dummy release on my fork to generate macOS binaries with the correct commit.

konn:
I've just re-uploaded gzip and tarballs, as it seems all the tests pass now.

PR #1940, opened 2021-06-17
Add ghc-9.0.1 to the build release script

anka-213:
See also discussion here: https://github.com/haskell/haskell-language-server/pull/1912#issuecomment-862982762

Is this sensible, @jneira?

### DIFF
--- a/cabal-ghc901.project
+++ b/cabal-ghc901.project
@@ -84,7 +84,7 @@ source-repository-package
 
 write-ghc-environment-files: never
 
-index-state: 2021-06-17T10:08:24Z
+index-state: 2021-06-21T19:57:32Z
 
 constraints:
     -- These plugins doesn't work on GHC9 yet
--- a/cabal-ghc901.project
+++ b/cabal-ghc901.project
@@ -82,21 +82,13 @@ source-repository-package
   subdir: lsp-test
 -- https://github.com/haskell/lsp/pull/312
 
-source-repository-package
-  type: git
-  location: https://github.com/diagrams/active
-  tag: ca23431a8dfa013992f9164ccc882a3277361f17
--- https://github.com/diagrams/active/pull/36
-
 write-ghc-environment-files: never
 
-index-state: 2021-05-21T05:01:41Z
+index-state: 2021-06-17T10:08:24Z
 
 constraints:
-    -- Diagrams doesn't support optparse-applicative >= 0.16 yet
-    optparse-applicative < 0.16
     -- These plugins doesn't work on GHC9 yet
-    , haskell-language-server -brittany -class -eval -fourmolu -ormolu -splice -stylishhaskell -tactic -refineImports
+    haskell-language-server -brittany -class -eval -fourmolu -ormolu -splice -stylishhaskell -tactic -refineImports
 
 
 allow-newer:
@@ -112,7 +104,6 @@ allow-newer:
     -- multistate:base,
     -- ghc-source-gen:ghc,
 
-    active:base,
     assoc:base,
     cryptohash-md5:base,
     cryptohash-sha1:base,
@@ -121,17 +112,13 @@ allow-newer:
     deepseq:base,
     dependent-sum:some,
     dependent-sum:constraints,
-    diagrams-contrib:base,
-    diagrams-contrib:lens,
-    diagrams-contrib:random,
-    diagrams-core:base,
-    diagrams-core:lens,
-    diagrams-lib:base,
-    diagrams-lib:lens,
     diagrams-postscript:base,
     diagrams-postscript:lens,
-    diagrams-svg:base,
-    diagrams-svg:lens,
+    diagrams-postscript:diagrams-core,
+    diagrams-postscript:monoid-extras,
+    diagrams:diagrams-core,
+    Chart-diagrams:diagrams-core,
+    SVGFonts:diagrams-core,
     dual-tree:base,
     -- Does this make any sense?
     entropy:Cabal,
--- a/cabal-ghc901.project
+++ b/cabal-ghc901.project
@@ -96,7 +96,7 @@ constraints:
     -- Diagrams doesn't support optparse-applicative >= 0.16 yet
     optparse-applicative < 0.16
     -- These plugins doesn't work on GHC9 yet
-    , haskell-language-server -brittany -class -eval -fourmolu -modulename -ormolu -splice -stylishhaskell -tactic -refineImports
+    , haskell-language-server -brittany -class -eval -fourmolu -ormolu -splice -stylishhaskell -tactic -refineImports
 
 
 allow-newer:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ['8.10.5', '8.10.4', '8.10.3', '8.10.2', '8.8.4', '8.8.3', '8.6.5', '8.6.4']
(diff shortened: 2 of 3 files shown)


jneira:
looks good, yeah, could you test the build script creating a fake release in your repo? It is the default trigger of the script

anka-213:
Oh, I should also enable the `module-name` plugin, since that is working but was accidentally disabled for ghc-9.0.1.

anka-213:
Here's the build for the test-release: https://github.com/anka-213/haskell-language-server/actions/runs/945876049
And there's the release itself: https://github.com/anka-213/haskell-language-server/releases/tag/test-release-ghc9

cabal-ghc901.project
```diff
@@ -96,7 +96,7 @@ constraints:
     -- Diagrams doesn't support optparse-applicative >= 0.16 yet
     optparse-applicative < 0.16
     -- These plugins doesn't work on GHC9 yet
-    , haskell-language-server -brittany -class -eval -fourmolu -modulename -ormolu -splice -stylishhaskell -tactic -refineImports
+    , haskell-language-server -brittany -class -eval -fourmolu -ormolu -splice -stylishhaskell -tactic -refineImports
```
jneira, line 99: should we also enable it in the test script? or tests are gonna fail?

anka-213: Oh, right! Thanks!

anka-213: Where is it disabled in the tests?

anka-213: Or did no-one add it to `test.yaml` at all?

anka-213: The later case seems to be the case. https://github.com/haskell/haskell-language-server/pull/1836 didn't modify `test.yaml`

jneira: wow, module name plugin tests are not there! i guess we forgot to add it and we should do it and check if they work with 9.0.1 (to disable it or not)
But that is matter of other pr :wink:

anka-213: I might as well add them in this PR while I'm at it, or would that be confusing?

jneira:
> 
> 
> Here's the build for the test-release: https://github.com/anka-213/haskell-language-server/actions/runs/945876049
> And there's the release itself: https://github.com/anka-213/haskell-language-server/releases/tag/test-release-ghc9

just in case, i am afraid the build failed

cabal-ghc901.project
jneira: i think we should add tests for module plugin in another one, dont worry

anka-213:
> just in case, i am afraid the build failed

Yes, I noticed, thanks! I'm confused about why. Good thing we tested!

cabal-ghc901.project
anka-213: There are a lot of plugins that aren't tested:
```
$ for p in $(ls plugins); do grep -q $p .github/workflows/test.yml || echo "$p doesn't exist";  done
hls-explicit-imports-plugin doesn't exist
hls-floskell-plugin doesn't exist
hls-hlint-plugin doesn't exist
hls-module-name-plugin doesn't exist
hls-ormolu-plugin doesn't exist
hls-pragmas-plugin doesn't exist
hls-retrie-plugin doesn't exist
```

anka-213: Maybe we should add a test for this in a github action, to guarantee that people won't forget to add the test-suite to `test.yml` when they add a new plugin?

jneira: not all plugins have the test suite in its own package, some of them are being tested in the the general hls test suite (under `./test`). Plugins packages with its own test suite have their own test subdir (f.e. hlint does not have it but pragmas does)

> Maybe we should add a test for this in a github action, to guarantee that people won't forget to add the test-suite to test.yml when they add a new plugin?

that is a nice idea

//cc @Ailrun

anka-213: Good point, here's the list with those without an internal testsuite filtered out:

```
$ for p in $(ls plugins); do [[ ! -d plugins/$p/test ]] || grep -q $p .github/workflows/test.yml || echo "$p doesn't exist";  done
hls-floskell-plugin doesn't exist
hls-module-name-plugin doesn't exist
hls-ormolu-plugin doesn't exist
hls-pragmas-plugin doesn't exist
```

And here is the list of plugins without an internal test-suite, for completeness:
```
grep -L test-suite plugins/*/*.cabal
plugins/hls-explicit-imports-plugin/hls-explicit-imports-plugin.cabal
plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
plugins/hls-retrie-plugin/hls-retrie-plugin.cabal
```

Ailrun: Good point. I will update the CI to check those missing tests.

anka-213:
Hmm, does anyone know why the tests on ghc-8.10.5, windows fails with
```
ghcide
  cradle
    loading
      retry failed: FAIL (0.62s)
        test\exe\Main.hs:4743:
        No joy after fixing the cradle
```
?

https://github.com/haskell/haskell-language-server/runs/2847357199?check_suite_focus=true

anka-213:
Here's a newer attempt: https://github.com/anka-213/haskell-language-server/runs/2849221500?check_suite_focus=true
With the release here: https://github.com/anka-213/haskell-language-server/releases/tag/test-release-ghc9-4

anka-213:
Why does `test.yml` [fail on ghc-8.8.3](https://github.com/haskell/haskell-language-server/pull/1940/checks?check_run_id=2849375473)? I haven't modified anything that can affect that.

```
session-loader/Development/IDE/Session/VersionCheck.hs:17:24: error:
Error:     • Exception when trying to run compile-time code:
        I could not find a GHC installation at /home/runner/.ghcup/ghc/8.8.3/lib/ghc-8.8.3. Please do a clean rebuild and/or reinstall GHC.
  error, called at src/GHC/Check.hs:188:7 in ghc-check-0.5.0.4-44e59734b78ba49fb418eea95f220fef65c9786bf80ec362d3c2dc41c1e8ee52:GHC.Check
      Code: makeGhcVersionChecker
              (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR")
    • In the Template Haskell splice
        $$(makeGhcVersionChecker
             (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
      In the expression:
        $$(makeGhcVersionChecker
             (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
      In an equation for ‘ghcVersionChecker’:
          ghcVersionChecker
            = $$(makeGhcVersionChecker
                   (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
   |
17 | ghcVersionChecker = $$(makeGhcVersionChecker (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Failed to build ghcide-1.4.0.1 (which is required by test:wrapper-test
from hls-1.2.0.0 and test:func-test from hls-1.2.0.0).
```

jneira:
I think we need invalidate the cache, as the location of ghc have changed (due to some ghc vm update)

jneira:
@anka-213 many thanks again, this could be merged

jneira (approved):

jneira:
@mergify backport 1.2.0-hackage

anka-213:
Seems like this requires a manual merge, because workflows were modified.

PR #1948, opened 2021-06-20
Add :info command in Eval plugin

### LINKED ISSUE #402: Add support for displaying :info
Another useful feature I miss from dante / intero.

Like haskell/ghcide#70, this might be difficult to implement, because it might not fit within the constraints of LSP?

akrmn:
99% of the code is copied from [`GHCi.UI.info`](https://gitlab.haskell.org/ghc/ghc/-/blob/master/ghc/GHCi/UI.hs#L1529) and the functions called from there. Ideally this would be exported from `GHCi.UI` but sadly that's not the case.

Note that the tests commit is marked as [WIP] because, when running the tests on my machine, I keep getting the instances in different orders. I haven't been able to figure out the source of nondeterminism.

This should eventually close #402

### DIFF
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
@@ -40,7 +40,7 @@ import           Data.Char                            (isSpace)
 import qualified Data.DList                           as DL
 import qualified Data.HashMap.Strict                  as HashMap
 import           Data.List                            (dropWhileEnd, find,
-                                                       intercalate)
+                                                       intercalate, intersperse)
 import qualified Data.Map.Strict                      as Map
 import           Data.Maybe                           (catMaybes, fromMaybe)
 import           Data.String                          (IsString)
@@ -84,7 +84,9 @@ import qualified Development.IDE.GHC.Compat           as SrcLoc
 import           Development.IDE.Types.Options
 import           DynamicLoading                       (initializePlugins)
 import           FastString                           (unpackFS)
-import           GHC                                  (ExecOptions (execLineNumber, execSourceFile),
+import           GHC                                  (ClsInst,
+                                                       ExecOptions (execLineNumber, execSourceFile),
+                                                       FamInst, Fixity,
                                                        GeneralFlag (..), Ghc,
                                                        GhcLink (LinkInMemory),
                                                        GhcMode (CompManager),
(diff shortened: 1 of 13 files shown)


pepeiborra (approved):
Lgtm, thanks Moises!

plugins/hls-eval-plugin/test/testdata/TInfo.expected.hs
```diff
@@ -0,0 +1,10 @@
+module TInfo (Eq, Ord, Foo) where
+
+import InfoUtil (Eq, Ord, Foo)
+
+-- >>> :info Foo
```
akrmn, line 5: I ended up changing these tests to work on a custom `Foo` (and `Bar`) datatype defined in `InfoUtil`

plugins/hls-eval-plugin/test/cabal.project
```diff
@@ -0,0 +1,3 @@
+packages:
```
akrmn, line 1: I moved the `test/testdata/cabal.project` file one directory up, so I could define another library `info-util`. This was because `:info` prints the origin comment (`-- defined in "..."`) using the full path to the file for modules in the same package, which wouldn't be portable.

plugins/hls-eval-plugin/test/info-util/InfoUtil.hs
```diff
@@ -0,0 +1,20 @@
...
+  , Foo (..)
+  , Bar (..)
+  , Baz
+  )
+where
+
+import Prelude (Eq, Ord)
+
+data Foo = Foo1 | Foo2
+  deriving (Eq, Ord)
```
akrmn, line 13: I ended up dropping the `Show` instances since they would still come out in different orders sometimes. Thankfully `Eq`, `Ord` and `Baz` are enough to show the difference between `:info` and `:info!`

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
```diff
@@ -665,6 +684,51 @@ evalGhciLikeCmd cmd arg = do
...
+    sdocs <- mapM infoThing (T.words s)
+    pure $ Just $ T.pack $ showSDoc dflags (vcat sdocs)
+    where
+        infoThing :: GHC.GhcMonad m => Text -> m SDoc
+        infoThing (T.unpack -> str) = do
+            names     <- GHC.parseName str
+            mb_stuffs <- mapM (GHC.getInfo allInfo) names
+            let filtered = filterOutChildren (\(t,_f,_ci,_fi,_sd) -> t)
+                                            (catMaybes mb_stuffs)
+            return $ vcat (intersperse (text "") $ map pprInfo filtered)
```
pepeiborra, line 698: You could always sort the `filtered` children before printing to ensure that the output is deterministic

akrmn: yeah, the problem is the `[ClsInst]` part of the tuple returned by `GHC.getInfo`. Sadly, `ClsInst` doesn't have an `Ord` instance, and I didn't want to impose any custom ordering

akrmn:
hmm, seems like `:info` outputs a different format on GHC 8.8.4 vs 8.10.5. Perhaps a golden test isn't the best way to test that command? any ideas @pepeiborra @jneira?

jneira:
mmm maybe different files for ghc version (I hope only will be two)?
or switch the test to look for the significant strings

plugins/hls-eval-plugin/test/Main.hs
```diff
@@ -107,11 +108,56 @@ tests =
     ]
   , goldenWithEval "Works with NoImplicitPrelude" "TNoImplicitPrelude" "hs"
   , goldenWithEval "Variable 'it' works" "TIt" "hs"
+
+  , testGroup ":info command"
+    [ testCase ":info reports type, constructors and instances" $ do
```
akrmn, line 113: As suggested by @jneira, these tests now only look for the relevant infix strings (or their absence)

jneira:
> 99% of the code is copied from GHCi.UI.info and the functions called from there. Ideally this would be exported from GHCi.UI but sadly that's not the case.

maybe it worths open a ticket in ghc asking for make it public?

jneira:
@akrmn many thanks for this useful feature

jneira:
> 
> 
> > 99% of the code is copied from GHCi.UI.info and the functions called from there. Ideally this would be exported from GHCi.UI but sadly that's not the case.
> 
> maybe it worths open a ticket in ghc asking for make it public?

@akrmn i was about to open an isuue in the ghc project but is not obvious for me what definitions from `GHCi.UI.Info` could be exported to avoid the duplication in our hanlding here

PR #1966, opened 2021-06-22
Wingman: Config option to suppress proofstate styling

### LINKED ISSUE #1965: Wingman proofstates don't look good outside of VSCode
The proofstate contains HTML for styling different terms. But these aren't supported in nvim, for example, so it would be nice to have an option to turn off the HTML generation. Ideally we could check the client capabilities, but `lsp-types` doesn't currently have support. https://github.com/haskell/lsp/issues/329

isovector:
Fixes #1965

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/LanguageServer.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/LanguageServer.hs
@@ -136,8 +136,11 @@ properties :: Properties
    , 'PropertyKey "max_use_ctor_actions" 'TInteger
    , 'PropertyKey "timeout_duration" 'TInteger
    , 'PropertyKey "auto_gas" 'TInteger
+   , 'PropertyKey "proofstate_styling" 'TBoolean
    ]
 properties = emptyProperties
+  & defineBooleanProperty #proofstate_styling
+    "Should Wingman emit styling markup when showing metaprogram proof states?" True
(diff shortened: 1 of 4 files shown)


plugins/hls-tactics-plugin/src/Wingman/LanguageServer.hs
```diff
@@ -136,8 +136,11 @@ properties :: Properties
    , 'PropertyKey "max_use_ctor_actions" 'TInteger
    , 'PropertyKey "timeout_duration" 'TInteger
    , 'PropertyKey "auto_gas" 'TInteger
+   , 'PropertyKey "proofstate_styling" 'TBoolean
    ]
 properties = emptyProperties
+  & defineBooleanProperty #proofstate_styling
```
isovector, line 142: @berberman it's a little annoying that these properties need to be added in the opposite order as in the type. Is there a better pattern here?

berberman: Hmm I think the order only affects in the type signature of `properties`, no?

isovector: Yes, but in the opposite order in which they're defined. Notice that I'm adding to the top of `emptyProperties` here, but the resulting type is added to the _end_ of the type :(

Ailrun (approved):
LGTM

PR #1969, opened 2021-06-23
Clear the ms_hspp_buf for non FOIs

pepeiborra:
We parse the mod summary for all the project targets at startup in order to build the module graph, which means that we store the preprocessed source of the entire project in memory for the duration of the session. 

This is wasteful, we should release those preprocessed sources and compute them again if needed later. But what's the best way to do this? 

One option would be to add a new `GetModSummaryWithoutPreprocessedSource` rule, but we already have `GetModSummary` and `GetModSummaryWithoutTimestamps`, so where does this end?

A simpler option is adopted in this PR. Since we only ever use the preprocessed buffer stored in the `ModSummary` when parsing, and (I believe) we only ever parse FOIs, it should be safe to clear them out for non FOIs.

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -2,7 +2,7 @@ cabal-version:      2.4
 build-type:         Simple
 category:           Development
 name:               ghcide
-version:            1.4.0.2
+version:            1.4.0.3
 license:            Apache-2.0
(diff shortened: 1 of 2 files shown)


pepeiborra:
This won't work, we use the preprocessed source to regenerate interfaces in non FOIs

PR #1978, opened 2021-06-25
Do not request argument docs - we don't use them

pepeiborra:
I believe that `mi_arg_docs` is lazy. Since we don't use argument docs anywhere, this is causing unnecessary work and heap allocations.


pepeiborra:
This doesn't save any heap space, so perhaps `mi_arg_docs` is not lazy after all

PR #1983, opened 2021-06-26
Jump to instance definition and explain typeclass evidence for GHC 9

wz1000:
Now that we have GHC 9, it is easy to implement "Jump to instance definition" as well as show exactly
the instances being used at a point on hover.

Hover

Jump to definition

TODO

- [ ] Extend to references - UI is a bit tricky here
- [ ] Improve markdown rendering - @isovector did you figure out how to have inline syntax highlighting yet?


pepeiborra:
Not strictly related to this change, but I wonder if it's possible to show instances in scope when hovering on a type?

isovector:
Inline syntax highlighting should just work in code fences --- but you'll need to make sure you're pushing a Markdown text kind: https://hackage.haskell.org/package/lsp-types-1.2.0.0/docs/Language-LSP-Types.html#t:MarkupKind

wz1000:
> Not strictly related to this change, but I wonder if it's possible to show the instances in scope when hovering on a type?

Yes it is. In fact, local completions will already (incorrectly) suggest local instances in scope (i.e. those arising from type signature contexts and case splits) if you start completing using `$d...`. I will fix this now.

For top level instances, we can use the same mechanism used by GHCi for `:instances`.

wz1000:
> Inline syntax highlighting should just work in code fences --- but you'll need to make sure you're pushing a Markdown text kind: https://hackage.haskell.org/package/lsp-types-1.2.0.0/docs/Language-LSP-Types.html#t:MarkupKind

What is the syntax? Is it like pandoc?

```
evidence for constraint `Show a`{.haskell}
```

wz1000:
Some opportunities for bikeshedding:

- We can probably do away with the top level bullet point
- "defined using"/"constructed from"?
- "evidence of constraint" or simply "evidence" or even "dictionary of type"?
  - Maybe omit  the words "evidence of constraint" etc. from all the lower level bullet points as it is just noise at that point

pepeiborra:
> Some opportunities for bikeshedding:
> 
> * We can probably do away with the top level bullet point
> * "defined using"/"constructed from"?
> * "evidence of constraint" or simply "evidence" or even "dictionary of type"?

"from instance", "via instance" or simply "instance"?

>   
>   * Maybe omit  the words "evidence of constraint" etc. from all the lower level bullet points as it is just noise at that point

+1

wz1000:
> "from instance", "via instance" or simply "instance"?

It is not always an instance, see for example:

2021-06-26-210629_grab

isovector:
@wz1000 sorry, I misunderstood. I'm not sure about inline syntax highlighting --- was thinking about block-level.  AFAIK there is no way to say you only want the type-level highlighting though.

isovector:
Your CI appears to be blocked by https://github.com/haskell/haskell-language-server/issues/2016

pepeiborra:
@wz1000 are you planning to land this soon or could someone else work on it? 
@July541 would you be interested in wrapping this up?

July541:
It's a very practical function! I'm glad to push it far more with @wz1000' willing.

BTW, have this feature been integrated into `hiedb`? Call hierarchy can also jump to the corresponding instance with this.

jneira:
Needs a non trivial (for me :-P ) rebase

wz1000:
@July541 yes, feel free to pick this up. `hiedb` to some extent will "just work" with these changes, but it may also need some special support somewhere depending on what exactly you need to do.

July541:
@wz1000 Hello, do you mind explaining some details?
1. For `Extend to references`, what should it be? Should it remove irrelative instances?
2. What should improve in markdown?

Anton-Latukha:
Rebased.

That was not a pretty way to rebase, but it does the job while preserving commit authorships.

Anton-Latukha:
(well, *"this works on my machine"* in GHC 8.10.7)

Anton-Latukha:
`nameNameSpace` & stuff is already a programming of a provisioning of a GHC compatibility (`Compat.Plugins` etc) & through the web rebase (without accessing pushing to fork) - it is slow to do.

Rebase is done.

PR #1984, opened 2021-06-27
Don't suggest import an unnecessary data constructor.

### LINKED ISSUE #1932: Adding types to import list includes identically named constructors.
### Steps to reproduce
```haskell
module A where

data Foo = Foo
```
```haskell
module B where

foo :: Foo
foo = undefined
```
Pressing `Ctrl+.` on `Foo` in Module `B` and adding the missing import
adds `import A (Foo(Foo))` to the import list.

### Expected behaviour
Only the type name gets imported, not the constructor.

### Actual behaviour
The constructor gets imported as well, even though it is not required.

### Your environment

```sh
haskell-language-server version: 1.2.0.0 (GHC: 8.10.4) (PATH: /home/finn/.config/Code/User/globalStorage/haskell.haskell/haskell-language-server-1.2.0-linux-8.10.4) (GIT hash: be2071e985cb417f984ab00a1aad76dee02d6d0b)
cabal:		3.0.0.0
stack:		2.5.1
ghc:		8.8.4
```
Which OS do you use:
Fedora 34

Which lsp-client do you use:
VS Code

peterwicksstringfield:
Usually, when the type constructor is used but the data constructor isn't, we
only suggest importing the type constructor. When the data constructor and type
constructor share the same name we get mixed up and suggest importing both.

The solution is to do a better job filtering out the data cons when we know we
are looking for a type con.

Closes #1932 .

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -804,6 +804,10 @@ suggestExtendImport exportsMap (L _ HsModule {hsmodImports}) Diagnostic{_range=_
     = mod_srcspan >>= uncurry (suggestions hsmodImports binding)
     | otherwise = []
     where
+        canUseDatacon = case extractNotInScopeName _message of
+                            Just NotInScopeTypeConstructorOrClass{} -> False
+                            _                                       -> True
+
         suggestions decls binding mod srcspan
           | range <- case [ x | (x,"") <- readSrcSpan (T.unpack srcspan)] of
                 [s] -> let x = realSrcSpanToRange s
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1318,8 +1322,9 @@ hideImplicitPreludeSymbol :: T.Text -> NewImport
 hideImplicitPreludeSymbol symbol = newUnqualImport "Prelude" symbol True
 
 canUseIdent :: NotInScope -> IdentInfo -> Bool
-canUseIdent NotInScopeDataConstructor{} = isDatacon
-canUseIdent _                           = const True
+canUseIdent NotInScopeDataConstructor{}        = isDatacon
+canUseIdent NotInScopeTypeConstructorOrClass{} = not . isDatacon
+canUseIdent _                                  = const True
```
peterwicksstringfield, line 1328: FYI: canUseIdent is called in two places now. suggestExtendImport and suggestNewImport.

jneira (approved):
Many thanks for refine this

PR #1988, opened 2021-06-28
Wingman: add emacs example config to Readme

stuebinm:
This adds a barebones example config for emacs, demonstrating how to bind wingman actions to specific keybindings. Most things work out of the box anyways (i.e. code actions just show up in the relevant menus), but I found the `lsp-make-interactive-code-action` function somewhat hard to find, so I think the pointer to it could probably benefit other people.


isovector (approved):
Love it! Thanks so much!

Issue #1990, opened 2021-06-30
July 2021 monthly release (1.3.0) with limited ghc 9.0 support (needs hackage release)


jneira:
The release also would include, among other things:
* The lately amazing work done in the wingman plugin by @isovector 
* Binaries for ghc-9.0.1 (thanks to @anka-213)
  * Those binaries dont include all plugins and did not pass the entire test suite but i think we could inlcude it with noting they are in beta state (would love to know your opinion about this)
* Artifacts for ghcup for the three os's and a SHA256SUMS with the checksums from all artifacts which could be gpg signed (not sure if we will able to sign them in this one though)
* Checklist of upstream packages we need up to date to make the hackage release:
  * [ ] ghc-api-compat:
    * https://github.com/hsyl20/ghc-api-compat/pull/11
    * https://gitlab.haskell.org/haskell/ghc-api-compat/-/issues/1#note_371055
    * https://github.com/haskell/cabal/issues/4629
    * (pending pr to remove it from hls)
  * [x] remove `allow-newer`: #2097

konn:
I think we could send to lsp clients, when called for ghc 9, some "warning" that noting GHC 9.0.1 support is in the beta stage with partial functionalities and pointing to GHC 9 tracking issue #297.
I think it is likely that some people might not be aware of release note (e.g. automaticaljy updated/installed  by lsp client), and sporadically report it as independent issues - so some auxiliary pointer at the editor level could prevent such cases.

jneira:
> I think we could send to lsp clients, when called for ghc 9, some "warning" that noting GHC 9.0.1 support is in the beta stage with partial functionalities and pointing to GHC 9 tracking issue #297.

Good idea, a warning popup would be great

anka-213:
Oh, does it no longer pass the test suite? What broke?

jneira:
> Oh, does it no longer pass the test suite? What broke?

Dont worry enabled tests are passing but afaics ghc-9.0 is not tested at all in Windows and macOS and the tests of all plugins, including the enabled ones for ghc-9.0 are not being executed.

jneira:
In fact we are not testing macOS at all :worried:

anka-213:
I have run all the tests manually on MacOS on my machine and they worked fine there last I checked. I don't know about windows. But yes, we could enable the tests for ghc9 on windows and macos at the cost of more CI time.

anka-213:
[...]

PR #1992, opened 2021-07-01
Warn GHC 9 Compatibility to LSP Client

konn:
As I proposed in #1990, this PR makes ghcide to send Warning to LSP clients if GHC >= 9.0.1.
It sends the message right before the initialisation; if it's inappropriate, please fix me.

It also adds README to the notes on GHC 9 compatibility.

When invoked on GHC 9 projects, ghcide warns like this:

スクリーンショット 2021-07-01 10 47 27

expanded:

スクリーンショット 2021-07-01 10 47 31


ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -280,3 +281,10 @@ ioe_dupHandlesNotCompatible :: Handle -> IO a
 ioe_dupHandlesNotCompatible h =
    ioException (IOError (Just h) IllegalOperation "hDuplicateTo"
                 "handles are incompatible" Nothing Nothing)
+
+isOverGhc9 :: Bool
```
jneira, line 285: We have something similar in hls-test-utils but more general, maybe it worths move the test code to check ghc version to Development.IDE.GHC.Compat and make it available for the entire codebase:

https://github.com/haskell/haskell-language-server/blob/a0def3c4236b12105970b59f791bf86df666fbcc/hls-test-utils/src/Test/Hls/Util.hs#L110-L126

wz1000:
I think this is not worth the annoyance on every start up. I would consider it acceptable if it was limited to the first start.

jneira:
> 
> 
> I think this is not worth the annoyance on every start up. I would consider it acceptable if it was limited to the first start.

Is it possible using only the lsp spec? If it is not maybe the simpler thing is add the warning in the vscode editor, when it downlods hls for ghc-9.0.1. It should be done for each editor though.

anka-213:
vscode is the only one that downloads the language server automatically, so it's probably not as important for the other editors.

jneira:
maybe we could convert the alert in a log output, like the warning about the implicit cradle?

konn:
> maybe we could convert the alert in a log output, like the warning about the implicit cradle?

Hmm, log output cannot be seen unless a user actively opens and navigate it so it makes less sense, I think...
Well, if there is not much consensus on adding this warning, I would close this.

jneira:
> Hmm, log output cannot be seen unless a user actively opens and navigate it so it makes less sense, I think...
Well, if there is not much consensus on adding this warning, I would close this.

Well i can imagine a situation where users not using tha automatic download of vscode get an error or detect a plugin is not enabled (and they did not know they are disabled for ghc 9) and examine and/or post the logs in an issue. The message will help them or us to detect quickly the cause.

jneira:
@konn do you mind i take over the pr to convert the alert in a los message? fine with closing it too if you think it is better

konn:
Sorry for leaving this PR idle (I've been a bit busy these days...). 

> @konn do you mind i take over the pr to convert the alert in a los message?

Ah, yes please! And thanks for promoting GhcVer things to project-level.

jneira (approved):
lgtm 😉, thanks for adding the warning

Issue #1871, opened 2021-05-27
haskell-language-server-8104 doesn't build in flake


teto:
Thanks for adding a flake to this repo, it makes things so much easier !

I just wanted to warn that `nix build .#haskell-language-server-8104 ` was broken for 204fdcb22842ebc1a03120da94db2942e4f412f8 and that I had to pin it to
`    hls.url = "github:haskell/haskell-language-server?rev=37e0551b0d411e345e250a6ed287103640524a62";`

Would be cool to only update the flake if all targets compile.

NB: Also old channel is mentioned at https://github.com/haskell/haskell-language-server#contributing

berberman:
Hi, thanks for the report, I believe this issue was fixed in #1870, and `nix build` sholud work for master. Feel free to re-open the issue if you still encounter this :)

teto:
cool thanks

teto:
`nix build .#haskell-language-server-901`
fails with 
```
   > Setup: Encountered missing or private dependencies:
       > base >=4.5 && <4.15
```

`nix build .#haskell-language-server-8105` fail too. The advantage of flake is the lock file to guarantee that things work. Maybe mark those targets as broken until they are really available ?

anka-213:
Yes, that's a good point. There is (basic) support for ghc-9.0.1 in HLS now, but the nix-expressions hasn't yet been updated with the info from the `cabal-ghc901.project` or `stack-9.0.1.yaml` files, so until that happens, they should be marked as broken.

teto:
I've been playing with haskell.nix recently and it seems more powerful than nixpkgs' ecosystem (where you contantly have to fiddle with everything). Any thought ?

anka-213:
I'm not too familiar with the `haskell.nix` so I can't really say anything. The only thing I know is that you can't use nixpkgs official binary cache if you use `haskell.nix`, but I don't know if that makes any practical difference here.

berberman:
```
$ nix build .#haskell-language-server-901
error: builder for '/nix/store/d1i4yrpiyxgqmspdr4kb4s1sldb07rf7-cryptohash-md5-0.11.100.1.drv' failed with exit code 1;
       last 10 log lines:
       > [1 of 1] Compiling Main             ( Setup.hs, /build/Main.o )
       > Linking Setup ...
       > configuring
[...]

PR #1995, opened 2021-07-02
nix: add support for ghc 9.0.1

berberman:
As discussed in https://github.com/haskell/haskell-language-server/issues/1871#issuecomment-859990650, this PR attempts to maintain a nix configuration that works for ghc 9.0.1. It's still work in progress, see the CI for results.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -91,16 +91,16 @@ jobs:
         name: haskell-language-server
         authToken: ${{ secrets.HLS_CACHIX_AUTH_TOKEN }}
     - name: Build development shell
-      run: |
-        nix develop --profile dev
-        nix develop .#haskell-language-server-901-dev --profile dev
+      run: nix develop --profile dev
+    - name: Build development shell (GHC 9.0.1)
+      run: nix develop .#haskell-language-server-901-dev --profile dev
     - name: Push development shell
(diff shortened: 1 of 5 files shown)


anka-213:
It looks like it is trying to build `ghc-lib-parser-8.10.5` when compiling for ghc-9. I'm guessing that is the default version in nixpkgs. Maybe overriding with something like
```
ghc-lib-parser = hsuper.ghc-lib-parser_9_0_1_20210324;
```
would work?

maralorn:
> (Or `hself` instead of `hsuper`? I never learned when to use which.)

My rule of thumb is this: Always use `hself` when you can, only use `hsuper` to break recursion cycles.

That mostly means use `foo = f hsuper.foo` but `bar = f hself.foo`.

Reasoning is that when you do `bar = f hself.foo`, there could be other overrides for `foo` in this or a subsequent overlay and you most likely want those overrides to apply to your new `bar`. You wouldn‘t get them from `hsuper.foo`.

berberman:
Now both `nix build .#haskell-language-server-901` and `nix develop .#haskell-language-server-901-dev` work; I'm able to use HLS binary created by nix build to load HLS project in nix develop shell!

jneira (approved):
I am not a nix user but lgtm

PR #1996, opened 2021-07-03
Maximize sharing of NormalizedFilePath values in getLocatedImports

pepeiborra:
The goal is to avoid having multiple copies of the same `NormalizedFilePath` value in the heap. 

A previous attempt was using hash consing (https://github.com/haskell/lsp/pull/340) but I think this small change should be enough.


pepeiborra:
The benchmarks show consistent heap savings of at least 5MB for the Cabal example (around 2.5%). That should scale up nicely in larger projects with thousands of modules.

jneira (approved):
lgtm, thank for this nice improvement

PR #1997, opened 2021-07-03
Eval plugin: support ghc 9.0.1

berberman:
Tests are broken due to:
* the type of strings is displayed as `String` in GHC 9, rather than `[Char]`
* according to https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0179-printing-foralls.rst, inferred type variables are enclosed with braces in GHC 9

These are minor differences, but I'm not sure if it's a good idea to have separate golden test files for GHC 9...


pepeiborra:
Awesome! 

Re the test files, I think that we will need to update them for 9.0 and disable the eval plugin tests for 8.x to avoid maintaining two sets of golden tests files.

jneira:
We are gonna support ghc 8 for a while, so I would wait to a stable and widely used ghc 9 to remove tests for older versions
the ghc golden tests file will not need maintenance effort, as they are not gonna change (probably)

pepeiborra:
> We are gonna support ghc 8 for a while, so I would wait to a stable and widely used ghc 9 to remove tests for older versions
> The ghc golden tests file will not need maintenance effort, as they are not gonna change (probably) and we could only add new tests for 9.

Removing the tests for ghc 8.x does not mean that we don't support the plugin in that version, but simply that we only test it with ghc 9.

Either now or in the future we are going to need to migrate the test suite to ghc 9. My view is that it's better to do it now that it's cheap rather than wait until later when it's more expensive.

plugins/hls-eval-plugin/hls-eval-plugin.cabal
```diff
@@ -94,19 +94,21 @@ library
...
-    , directory
-    , extra
-    , filepath
-    , hls-eval-plugin
-    , hls-test-utils   ^>=1.0
-    , lens
-    , lsp-types
+-- test suites are only available in GHC 9
+if impl(ghc >=9.0)
+  test-suite tests
```
berberman, line 99: Hmm this doesn't seem to work

jneira:
> 
> Removing the tests for ghc 8.x does not mean that we don't support the plugin in that version, but simply that we only test it with ghc 9.

Sure but new bugs could be introduced and in that case, depending in their severity, it will be worse than remove support for the plugin
 
> Either now or in the future we are going to need to migrate the test suite to ghc 9. My view is that it's better to do it now that it's cheap rather than wait until later when it's more expensive.

Mmm we will have to 
# copy golden tests and change them and
# add and cpp (or runtime) if for load files from one location or another

When we will drop support for ghc <= 8, we will have to remove old golden files and the if (like the rest of dozens of similar ifs everywhere).
The extra effort (create the if now and remove later) worths it imo.

pepeiborra:
> > Removing the tests for ghc 8.x does not mean that we don't support the plugin in that version, but simply that we only test it with ghc 9.
> 
> Sure but new bugs could be introduced and in that case, depending on their severity, it will be worse than remove support for the plugin
> 

That's true. As maintainers, we always to balance the cost of duplication versus the perfect coverage. What I am saying is that, as a maintainer, I believe that it is not worth to duplicate these tests and that right now is the optimal moment to upgrade them. But I can also see your point: we don't want to accidentally break GHC 8.x support.

> > Either now or in the future we are going to need to migrate the test suite to ghc 9. My view is that it's better to do it now that it's cheap rather than wait until later when it's more expensive.
> 
> Mmm we will have to
> 
> * copy golden tests and change them and
> * add and cpp (or runtime) if for load files from one location or another
> 
> When we will drop support for ghc <= 8, we will have to remove old golden files and the if (like the rest of dozens of similar ifs everywhere).
> The extra effort (create the if now and remove later) worths it imo.

CPP in our test suites is not as pervasive as it is in our library code, and as far as I know we don't have duplicate golden tests anywhere else. But I'm not against adding them, if @berberman is willing to contribute them.

berberman:
I'm fine with maintaining test cases for multi ghc versions, but having many duplicate golden tests files is a bit tedious, so I extracted tests pieces which need variation with ghc versions to the haskell source file for applying conditions.

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -112,10 +112,13 @@ data GhcVersion
   | GHC88
   | GHC86
   | GHC84
+  | GHC901
```
jneira, line 115: Hi, dont want to make you add more changes but i am preparing a pr which promotes those data types to ghcide to being able to use them in all the codebase. In that pr i am using `GHC90` to be consistent with the rest of constructors: https://github.com/jneira/haskell-language-server/pull/24/files#diff-a54f7055c79019e6a0bfbac7b8dcce0048de756ad4ea47d02a161177dd57d9acR593-R611

Will change this to match my changes after this pr so dont worry if you dont want/have time to make the GHC901 -> GHC90 change.

berberman: OK. As far as i know there are many changes since ghc 9.1, so I just used `GHC901` here, though I'm not sure if the incoming changes will break something.

jneira: ghc 9.1 will have its own `GHC91` and we could change cpp conditions to make `GHC90` and `GHC91` incompatible (with `MIN_VERSION_GLASGOW_HASKELL(9,0,0,0) && !MIN_VERSION_GLASGOW_HASKELL(9,1,0,0)`) so it will throw a runtime (i am afraid) error

jneira (approved):
lgtm, thanks for include the eval plugin for ghc-9

PR #1998, opened 2021-07-04
Remove >= from cabal-version

felixonmars:
Fixes warnings like:

```
Warning: test/testdata/testdata.cabal:3:27: Packages with
'cabal-version:
1.12' or later should specify a specific version of the Cabal spec of
  the form
  'cabal-version: x.y'. Use 'cabal-version: 2.0'.
```


jneira (approved):
lgtm thanks for the fix

PR #1999, opened 2021-07-04
Allow HLS plugins to declare cli commands

pepeiborra:
I want to add a command line retrie subcommand provided by `hls-retrie-plugin`. Generally, this PR enables any `PluginDescriptor` to declare new cli commands.

For consistency, I moved `--vscode-extension-schema` and `--generate-default-config` to be subcommands as well. Let me know if anything else needs changing; where are these commands called from?

How it looks like:
```
Available commands:
  typecheck                Used as a test bed to check your IDE will work
  hiedb                    Query .hie files
  lsp                      Start talking to an LSP client
  vscode-extension-schema  Print generic config schema for plugins (used in the
                           package.json of haskell vscode extension)
  generate-default-config  Print config supported by the server with default
                           values
  retrie                   Retrie commands
```


berberman (approved):
LGTM. As far as I know we use `--vscode-extension-schema` and `--generate-default-config` manually to update the `package.json` of our vscode extension when needed. I think we could create a command `config` and let these two commands be subcommands, grouping them together.

PR #2001, opened 2021-07-06
Enable tests for ghc 9 and promote `ghcVersion` check

jneira:
* Promote `GhcVersion` datatype and `ghcVersion` function to ghcide to make it available to all the codebase
* Use them to replace CPP for code which paths compile for several ghc versions
* Add `GHC90` version and remove `GHC84` (as we dont support it anymore)
* Include test execution for ghc-9.0 and windows and macos (we could remove macos after a succesful execution, as we are not testing it for any version)


pepeiborra:
This is very nice, thank you @jneira  !!
To prevent the MIN_VERSION macros from coming back, I would also remove the CPP extension pragmas and remove the modules from the allowlist in the ghcide hlint file.

jneira:
> 
> 
> This is very nice, thank you @jneira !!
> To prevent the MIN_VERSION macros from coming back, I would also remove the CPP extension pragmas and remove the modules from the allowlist in the ghcide hlint file.

Hmm, afaics the unique module where i could remove CPP language pragma was `ghcide/session-loader/Development/IDE/Session.hs` and it was not listed in the modules where CPP is allowed: 

https://github.com/haskell/haskell-language-server/blob/ce15ff6eb1cd2e03c407536694b760ff8f8205a0/ghcide/.hlint.yaml#L88-L110

jneira:
I've marked circleci job and the ubuntu test job for 9.0.1 as required

jneira:
We have a succesful run with tests included for ghc-9.0.1 and macos: https://github.com/haskell/haskell-language-server/pull/2001/checks?check_run_id=2997169971
So i am gonna disable tests for it too :cry:

pepeiborra (approved):
I guess we still need CPP around imports and other non-expression sections

jneira:
The test job for windows and ghc-9.0.1 is getting stuck in the last builds :worried:

PR #2029, opened 2021-07-19
Tie plugins' pluginModifyDynflags to their enabled state

isovector:
This PR ensures that plugins' `pluginModifyDynflags` modifications run only when the plugin is enabled via `plcGlobalOn`. The prior behavior was that they'd always run; even if the plugin was disabled.


pepeiborra (approved):
Looks good, thanks!

PR #2030, opened 2021-07-20
Fix nix flake

Avi-D-coder:
The build was broken, and now it's not.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -71,10 +71,6 @@
           tweaks = hself: hsuper:
             with haskell.lib; {
 
-              hls-tactics-plugin = hsuper.hls-tactics-plugin.override {
-                refinery = hself.refinery;
-              };
-
               hiedb = hself.callCabal2nix "hiedb"
                 (builtins.fetchTarball {
                   url =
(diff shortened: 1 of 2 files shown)


Ailrun (approved):
LGTM

flake.nix
```diff
@@ -72,7 +72,7 @@
             with haskell.lib; {
 
               hls-tactics-plugin = hsuper.hls-tactics-plugin.override {
-                refinery = hself.refinery_0_3_0_0;
+                refinery = hself.refinery;
```
berberman, line 75:
```suggestion

```

flake.nix
```diff
@@ -83,6 +83,37 @@
...
+
+              ghc-source-gen = hself.callCabal2nix "ghc-source-gen"
+                (builtins.fetchTarball {
+                  url =
+                    "https://hackage.haskell.org/package/ghc-source-gen-0.4.1.0/ghc-source-gen-0.4.1.0.tar.gz";
+                  sha256 =
+                    "0kk599vk54ckikpxkzwrbx7z5x0xr20hr179rldmnlb34bf9mpnk";
+                }) { };
+
+              refinery = hself.callCabal2nix "refinery"
```
berberman, line 110: IIRC, we have refinery 0.4.0 in nixpkgs, so maybe we don't need this source override.

Avi-D-coder: Good catch, thanks

berberman (approved):
LGTM, thanks!

Issue #2005, opened 2021-07-08
Formatting with Brittany Fails When Warnings are Emitted


prikhi:
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
haskell-language-server version: 1.2.0.0 (GHC: 8.8.4) (PATH: /home/prikhi/.local/bin/haskell-language-server-wrapper) (GIT hash: b8bb06eb1b117943f4436a6fdafe5c09e76cac1c)
Tool versions found on the $PATH
cabal:		3.2.0.0
stack:		2.7.1
ghc:		8.10.5
```

Which OS do you use:
ArchLinux

Which lsp-client do you use:
Neovim + coc.nvim

### Steps to reproduce

Chang formatter to `brittany` & attempt to format a file that contains type operators in a multi-line expression:
```haskell
{-# LANGUAGE TypeOperators #-}
module HsOpTy where

-- brittany *should* remove the extra spaces between import & module name 
import     GHC.TypeLits

type Foo = 
  Int :
  '[]
```

See https://github.com/lspitzner/brittany/issues/271 for more repros.
As noted in that issue, if you remove the newlines, formatting proceeds correctly

### Expected behaviour

HLS ignores the warnings & uses the formatted text. 
This is the behavior brittany's CLI tool exhibits:
```
$ cat test.hs
{-# LANGUAGE TypeOperators #-}
module HsOpTy where

import     GHC.TypeLits

type Foo =
  Int :
  '[]
$ brittany --write-mode display test.hs >/dev/null
WARNING: encountered unknown syntactical constructs:
  HsOpTy{} at test.hs:(7,3)-(8,5)
  -> falling back on exactprint for this element of the module
$ brittany --write-mode display test.hs 2>/dev/null
{-# LANGUAGE TypeOperators #-}
module HsOpTy where

import GHC.TypeLits

type Foo =
  Int :
  '[]
```

### Actual behaviour

Nothing happens, following output appears in coc's HLS workspace output:
```
2021-07-08 00:11:10.475037968 [ThreadId 15168] INFO hls:	finish: brittany (took 0.00s)
[Error  - 12:11:10 AM] Request textDocument/formatting failed.
  Message: brittanyCmd: HsOpTy{}
HsOpTy{}
HsOpTy{}

  Code: -32602 
```

### Include debug information
Execute in the root of your project the command `haskell-language-server --debug .` and paste the logs here:

<details>
<summary>
[...]

PR #2036, opened 2021-07-23
[#2005] Fix Formatting When Brittany Returns Warnings

prikhi:
This is the same function I have written for my brittany PR(https://github.com/lspitzner/brittany/pull/351) but there hasn't been any movement on that yet. 

It'd be great if we could have this temporary fix merged until something similar is supported by upstream brittany.

### DIFF
--- a/plugins/hls-brittany-plugin/LICENSE
+++ b/plugins/hls-brittany-plugin/LICENSE
@@ -1,201 +1,661 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
-   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-   1. Definitions.
+                            Preamble
 
-      "License" shall mean the terms and conditions for use, reproduction,
-      and distribution as defined by Sections 1 through 9 of this document.
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
 
-      "Licensor" shall mean the copyright owner or entity authorized by
-      the copyright owner that is granting the License.
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+our General Public Licenses are intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.
 
-      "Legal Entity" shall mean the union of the acting entity and all
-      other entities that control, are controlled by, or are under common
-      control with that entity. For the purposes of this definition,
-      "control" means (i) the power, direct or indirect, to cause the
-      direction or management of such entity, whether by contract or
-      otherwise, or (ii) ownership of fifty percent (50%) or more of the
-      outstanding shares, or (iii) beneficial ownership of such entity.
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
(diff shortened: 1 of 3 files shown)


jneira:
Thanks for fixing this here too. We have to remember remove it when brittan includes it.
@Ailrun @pepeiborra and others are you fine with include this here?

pepeiborra:
No objections from me as long as licensing issues are taken into consideration, given that Brittany is LGPL and the plugin is Apache 2.0

prikhi:
> We have to remember remove it when brittany includes it.

I'm fine w/ being the one responsible for removing any relevant changes if/once my brittany PR gets merged + released.

> Brittany is LGPL and the plugin is Apache 2.0

Sorry, hadn't considered this aspect of it. Don't have much knowledge of these two licenses, but it seems like the only way to bring this change into HLS before brittany merges it would be for us to use my brittany fork instead of pulling the code directly into this repo?

I could attempt to write a new function that's not just a re-write of `parsePrintModule` w/ different return types. Have I spent too much time looking at the brittany implementations that _anything_ I write could be considered a "derivative work"? :shrug: I don't have enough exposure to these sort of licensing issues to know what to do in this situation...

pepeiborra:
> > We have to remember remove it when brittany includes it.
> 
> I'm fine w/ being the one responsible for removing any relevant changes if/once my brittany PR gets merged + released.
> 
> > Brittany is LGPL and the plugin is Apache 2.0
> 
> Sorry, hadn't considered this aspect of it. Don't have much knowledge of these two licenses, but it seems like the only way to bring this change into HLS before brittany merges it would be for us to use my brittany fork instead of pulling the code directly into this repo?
> 
> I could attempt to write a new function that's not just a re-write of `parsePrintModule` w/ different return types. Have I spent too much time looking at the brittany implementations that _anything_ I write could be considered a "derivative work"? 🤷 I don't have enough exposure to these sort of licensing issues to know what to do in this situation...

In my previous comment I mistakenly said brittany is LGPL, when I actually meant AGPL. This is what the AGPL license says about modified source code:

> 5. Conveying Modified Source Versions.

  You may convey a work based on the Program, or the modifications to
produce it from the Program, in the form of source code under the
terms of section 4, provided that you also meet all of these conditions:

    a) The work must carry prominent notices stating that you modified
    it, and giving a relevant date.

    b) The work must carry prominent notices stating that it is
    released under this License and any conditions added under section
    7.  This requirement modifies the requirement in section 4 to
    "keep intact all notices".

    c) You must license the entire work, as a whole, under this
    License to anyone who comes into possession of a copy.  This
    License will therefore apply, along with any applicable section 7
    additional terms, to the whole of the work, and all its parts,
    regardless of how they are packaged.  This License gives no
    permission to license the work in any other way, but it does not
    invalidate such permission if you have separately received it.

While you could plausibly write a new version of the function from scratch, I would not encourage this. 

So in order to go ahead with this PR, you would need to change the plugin license to be AGPL. This is fine by me since I have stopped using Brittany in all my projects due to the odd choice of license, but you would also need to obtain permission from all other contributors to this plugin (@Ailrun @berberman @jneira @wz1000 and any others who have made any changes).

jneira:
Well include Brittany itself introduces LGPL in hls so I dont see any problem in change the plugin license as well.
So if nobody disagree I would let the pr change it

prikhi:
I rebased & switched the cabal `license` field & `LICENSE` file to `AGPL-3.0-only`. `brittany` uses an invalid SPDX identifier `AGPL-3`, which I assumed meant v3.0 only - but could switch to `AGPL-3.0-or-later` if deemed acceptable.

prikhi:
Also, is it possible to switch back to Apache once the upstream PR has been merged and we can remove this function? I left a comment in the cabal file suggesting so, but can remove it if it's not accurate.

jneira:
> 
> 
> Also, is it possible to switch back to Apache once the upstream PR has been merged and we can remove this function? I left a comment in the cabal file suggesting so, but can remove it if it's not accurate.

i think it s not necessary, we cant include the plugin without brittany itself so if we have to exclude gpl software, we have to exclude both at same time. In fact i think is better to keep in sync both licenses to make it clear.

So i will merge this if no contributor manifest their opposition

prikhi:
> i think it s not necessary

:+1:, removed the comment.

jneira:
reruning tests, it failed with a transient error (i hope)

jneira (approved):
LGTM, thanks for fixing here the bug, please, remember to remove it once it is upstreamed

PR #2037, opened 2021-07-26
Disable refinement hole fits

pepeiborra:
Let's review the benchmark results to see how expensive **refinement** hole fits are in practice. Then we can decide whether to keep them or not


Tritlo:
Just a comment: typed-holes (and refinement holes) will become a lot faster soon https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5980

PR #1955, opened 2021-06-21
Call hierarchy support

July541:
Hello all,

This pr is about my GSoC 2021 project, call hierarchy support. I create a draft to show my progress to those who may be interested.

**Progress**

- [x] Prepare call hierarchy (working now)
- [x] Incoming calls
- [x] Outgoing calls

**Known issues**

- [x] Data family range error (see commented test)
- [x] Incoming/Outgoing test missing
- [x] Name display not optimized for `DuplicateRecordFields`

PR #2040, opened 2021-07-27
Completions from non-imported modules

### LINKED ISSUE #1327: Completions from non-imported modules
A feature I remember from Java IDEs that was very nice was that completion would offer options from modules that were not (yet) imported, but were available. Selecting that completion would also insert the appropriate import. When combined with auto-organizing imports this lets you *almost* ignore imports and just work from the completion prompt.

I think we could do this, although there's a risk that we might end up returning a really large number of completions! Completions from other modules in the package, would be nice, but doing it for dependencies too would be great. I would love to be able to just complete `Map` and have it imported!

(The LSP protocol supports this: `CompletionItem`s have a list of `additionalTextEdit`s which even calls out "adding an additional import at the top of the file" as a example usecase.)

pepeiborra:
Provide completions for non-imported modules and automatically inserts new imports when necessary. The identifiers are extracted from the exports map, which no longer includes entries from `*.Internal` modules. 

Things left for future work:

- Qualified imports
- Better display of completion suggestions by default

This partially closes #1327

https://user-images.githubusercontent.com/26626/127127619-d92ba213-65d8-409c-a73d-35bbe13cc759.mov


jneira (approved):
Thanks for this nice feature, tests looks great

pepeiborra:
I'll merge this after #1955

Issue #1958, opened 2021-06-22
Add language extension code action targets the nearest {-# pragma


isovector:
Given this:

```haskell
data Metaprogram = Metaprogram
  { mp_name             :: !Text
  , mp_known_by_auto    :: !Bool
  , mp_show_code_action :: !Bool
  , mp_program          :: !(TacticsM ())
  }
  deriving stock Generic
{-# ANN Metaprogram "hello" #-}

instance NFData Metaprogram where
  rnf (!(Metaprogram !_ !_ !_ !_)) = ()
```

without `-XBangPatterns` enabled, I get a code action on the last line, suggesting I add `BangPatterns`. Doing so produces this output:

```haskell
data Metaprogram = Metaprogram
  { mp_name             :: !Text
  , mp_known_by_auto    :: !Bool
  , mp_show_code_action :: !Bool
  , mp_program          :: !(TacticsM ())
  }
  deriving stock Generic
{-# ANN Metaprogram "hello" #-}
{-# LANGUAGE BangPatterns #-}

instance NFData Metaprogram where
  rnf (!(Metaprogram !_ !_ !_ !_)) = ()
```

which is super the wrong place for it.

isovector:
Also happens with `INLINE` pragmas

Ailrun:
https://github.com/haskell/haskell-language-server/blob/d4e2a6f6af9bc2536f6b3208f236546a74484004/plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs#L192
is the line caused the problem.

jneira:
pragmas plugin is having a hard time finding the correct place to add the extension :smile: 
we must add a regression test for this (and over whatever corner case we can think of!)

jneira:
This was closed by #2043 i suppose, closing optimistically

PR #2043, opened 2021-07-27
[#1958] Fix placement of language pragmas

nini-faroux:
HI everyone, 

I added a small fix for issue #1958, where the language pragma was being added in the wrong place. I tested it against the examples in the two issues, and checked it still works with the older tests too. Hope it's ok!


plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
```diff
@@ -187,7 +188,19 @@ completion _ide _ complParams = do
...
 endOfModuleHeader :: T.Text -> Range
 endOfModuleHeader contents = Range loc loc
-    where
-        loc = Position line 0
-        line = maybe 0 succ (lastLineWithPrefix "{-#" <|> lastLineWithPrefix "#!")
-        lastLineWithPrefix pre = listToMaybe $ reverse $ findIndices (T.isPrefixOf pre) $ T.lines contents
+  where
+    loc = Position line 0
+    line = maybe 0 succ (lastShebang contents <|> lastOptsGhc contents <|> lastLangPrag contents)
+    lastShebang = lastLineWithPrefix $ T.isPrefixOf "#!"
```
pepeiborra, line 194: Could you expand the doc comment (and fix the markup) to explain what the new behaviour is? It will make it easier to review and to understand in the future.

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
```diff
@@ -187,7 +188,19 @@ completion _ide _ complParams = do
...
+    lastOptsGhc = lastLineWithPrefix $ checkPragma "OPTIONS_GHC" 11
+
+lastLineWithPrefix :: (T.Text -> Bool) -> T.Text -> Maybe Int
+lastLineWithPrefix p contents = listToMaybe $ reverse $ findIndices p $ T.lines contents
+
+checkPragma :: T.Text -> Int -> T.Text -> Bool
+checkPragma name n = check
+  where
+    check l = isPragma l && getName l == name
+    getName l = T.take n $ T.dropWhile isSpace $ T.drop 3 l
```
pepeiborra, line 205: It looks like the second argument is always the length of the first. Why not derive it internally if that's the case?

pepeiborra (commented):
Thank you for this fix and for extending the test suite!

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
nini-faroux: hi, thanks for the quick feedback! 

Ah yes good point, I'll change that function now. I also realised after that there are a few cases where this won't work as expected - for example, it will add a new Language pragma before the other language pragmas if the there is a shebang at the start of the file -  like this:

#! /usr/bin/env nix-shell
{-# LANGUAGE OverloadedStrings #-} -- newly added, should be last

so I was thinking it would be better to calculate the final line number as a sum the shebang, options_ghc and language pragmas. I did this locally and he works for all the older cases and new ones I could think of, like the one above. I can update my pr with this change this evening and also with the expanded comment?

pepeiborra: I was wondering this too - which is why I asked for more comments. Maybe it would be a good idea to add a few more tests to exercise a few more scenarios ?

pepeiborra: I was wondering this too - which is why I asked for more comments. Maybe it would be a good idea to add a few more tests to exercise a few more scenarios ?

nini-faroux: Sure no problem, I can include a few more with the update.

plugins/hls-pragmas-plugin/test/testdata/AddLanguagePragma.expected
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections #-}
+
+module NeedsLanguagePragms where
+
+tupleSection = (1,) <$> Just 2
+
+{-# INLINE addOne #-}
+addOne :: Int -> Int
+addOne x = x + 1
```
Ailrun, line 10: Is this file generated? Why doesn't this name contain ".hs" extension?

nini-faroux: hi, no I tried the test manually first and then added it to the test suite but I must have left off the .hs, thanks for pointing it out.

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
nini-faroux: So I added a number of new tests, different combinations of shebangs, ghc_options, language pragmas and later inline pragmas. I can't really think of any other combinations right now, but do let me know if I missed something or if I could improve the solution. Thanks!

isovector:
Something I encountered today. When the `module` is the first line in the file, my recentish build of HLS puts the language pragma underneath:

```haskell
module Foo where
```

=>

```haskell
module Foo where
{-# LANGUAGE Blah #-}
```

I had a quick look through the tests, but didn't see anything that tackled this case. Does it work under this PR?

nini-faroux:
hi @isovector, yes it will add it above module on the first line now like - 

{-# LANGUAGE TupleSections #-]
module Main ...

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
nini-faroux: hi again, I just cleaned up the code a bit, added another test for when 'module ...' is on the first line, and also squashed it down to one commit.

pepeiborra (approved):
Awesome, and great set of tests. Thanks @nini-faroux !

PR #2055, opened 2021-07-31
run benchmarks on ghc 9.0.1

pepeiborra:


pepeiborra:
This will have to wait until we remove `cabal-ghc901.project`.

PR #2059, opened 2021-08-01
Build with lsp 1.2.0.1

pepeiborra:
Fix minbounds, remove the source repositories from the cabal and stack project descriptors, and deal with lsp-types-1.3.0.0 new breaking exports.


jneira (approved):
LGTM: thanks for updating it!

PR #2067, opened 2021-08-02
Document call hierarchy plugin

July541:
This document includes
- description
- demo
- requirement and configure
- acknowledge


jneira:
wow, that was really fast, many thanks for adding the docs, we miss them here a lot :sweat_smile:

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+![Call Hierarchy in Emacs](call-hierarchy-in-emacs.gif)
+
+![Call Hierarchy in VSCode](call-hierarchy-in-vscode.gif)
+
+## Prerequisite
+None. You can experience the whole feature without any setting.
+
+## Configuration
+None. Native support by lsp-based client, switch on and off is insignificant.
```
jneira, line 17: i would add the basic configuration to enable/disable the plugin (in json format copypasteable). With the value enabled a true of course (the default value)

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Acknowledgments
+Supoorted by
```
jneira, line 20:
```suggestion
Supported by
```
small typo

July541:
Yes, I got a lot of trouble when I first into this project 😅, hope it can bring convenience to others.

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+* [Google Summer of Code](https://summerofcode.withgoogle.com/)
+* Warm and timely help from mentors [@jneira](https://github.com/jneira) and [@pepeiborra](https://github.com/pepeiborra)
```
jneira, line 23: Thanks a lot for the mention :slightly_smiling_face: , i wish i could offer more help

jneira:
I would add a link to this readme in the main one, in the features section f.e.

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+![Call Hierarchy in Emacs](call-hierarchy-in-emacs.gif)
+
+![Call Hierarchy in VSCode](call-hierarchy-in-vscode.gif)
+
+## Prerequisite
+None. You can experience the whole feature without any setting.
+
+## Configuration
+Enable default. You can disable it in the setting file (JSON format) whenever you like.
```
July541, line 17: @jneira I am not sure how to point out where the setting is precisely, maybe we can link it to a unified place(lots of plugins need this)

jneira: mmm you can the entire hls server configuration with 2 commands:
* `haskell-language-server vscode-extension-schema`: it outputs a json usable in the vscode extension package.json config to add them auto to the extension config ui in vscode
* `haskell-language-server generate-default-config`: it outputs the default values of that config

So the configuration for hlint, for example, would be:

```json
{
  "haskell.plugin.hlint.diagnosticsOn": false
}
```
If you put a file with that content in a `settings.json` int the project root dir, you will disable hlint

... but i tried both commands with last version and it does not mention any hls-call-hierarchy plugin.. 🤔 

I think something has to be done to "register" the plugin and include its config in the output, but dont remember right now (i think @berberman did that nice addition to the cli)

berberman: > but i tried both commands with last version and it does not mention any hls-call-hierarchy plugin

Ah, that's because
https://github.com/haskell/haskell-language-server/blob/047ffc366461470a27ea47c2042c76bbf97e1d15/hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs#L87-L95

and

https://github.com/haskell/haskell-language-server/blob/047ffc366461470a27ea47c2042c76bbf97e1d15/hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs#L117-L124

need update for new call hierarchy handlers

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,23 @@
+# Call hierarchy plugin for the [Haskell Language Server](https://github.com/haskell/haskell-language-server#readme)
+
+The call hierarchy plugin can review the code to determine where functions are called and how they relate to other functions.
+
+This plugin is useful when debugging and refactoring code because it allows you to see how different parts of the code are related. And it is more conducive for users to quickly understand their macro architecture in the face of strange code.
```
Ailrun, line 5: I think it's helpful to mention somewhere in this doc that this works only on definitions, not type signatures. I recently saw some people said that this feature doesn't work, but it was actually that they used this on type signatures.

Ailrun:
> I would add a link to this readme in the main one, in the features section f.e.

Yeah, giving users the links to documented plugins will be great. Currently it's kinda hard to know which one is documented, and it's also not so user-friendly that asking users to check directories for all packages.

plugins/hls-call-hierarchy-plugin/README.md
jneira: out of curiosity anyone knows if other langs work differently?

plugins/hls-call-hierarchy-plugin/README.md
jneira: @July541 could you have the chance of update those definitions, to complete the hierarchy plugin integration?

July541: Sure, I'll do it.

July541:
Continue on #2072.

Issue #2066, opened 2021-08-02
Add documentation to plugins


jneira:
* We have no documentation about each plugin but [wingman ](https://github.com/haskell/haskell-language-server/tree/master/plugins/hls-tactics-plugin) (kudos to @isovector for taking care of)
* This is :disappointed: and makes their use, maintenance and contributor initial work harder
* The documentation could live in the readme of each one (like wingman)
* As part of such documentation we could include the relevant server configuration (so we will make progress on #1433)
* It should include:
  * demos on how it works and a general description
  * the configuration controlling it as commented above
  * requirements to make it work (see #2064), kown issues, etc
  * tech doc about its organization, tech decisions made and their rationale, etc

jneira:
> 
> 
> I added a document in the call hierarchy, maybe we can let this as an example.

Of course! many thanks, i think we should ask for something similar for the new plugins (and work hard to add it for the existing ones)

jneira:
We have to do a check list here with the plugins and the current status of their docs

July541:
A track list
- ghcide
  - [x] core (https://github.com/haskell/haskell-language-server/blob/master/ghcide/README.md)
  - [ ] hover and symbols
  - [ ] code actions imports exports
  - [ ] code actions type signatures
  - [ ] code actions bindings
  - [ ] code actions fill holes
  - [ ] completions
  - [ ] type lenses
- [ ] hls-brittany-plugin
- [x] hls-call-hierarchy-plugin
- [x] hls-class-plugin (https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/README.md)
- [ ] hls-eval-plugin
- [ ] hls-explicit-imports-plugin
- [ ] hls-floskell-plugin
- [ ] hls-fourmolu-plugin
- [ ] hls-haddock-comments-plugin
- [ ] hls-hlint-plugin
- [ ] hls-module-name-plugin
- [ ] hls-ormolu-plugin
- [ ] hls-pragmas-plugin
- [ ] hls-refine-imports-plugin
- [ ] hls-retrie-plugin
- [ ] hls-splice-plugin
- [ ] hls-stylish-haskell-plugin
- [x] hls-tactics-plugin (https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-tactics-plugin/README.md), [Command](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-tactics-plugin/COMMANDS.md)

[...]

PR #2072, opened 2021-08-03
Support call hierarchy on type signature & add plugin to generic config  & docs

July541:
- Call hierarchy on type signature has no response is wired indeed, it has the same result like definition. cc @Ailrun 
- A config has been added, users can switch to enable it or not.
- Call hierarchy plugin's document has a changelog field now.

Can we move the [previous](https://github.com/haskell/haskell-language-server/pull/2067) pr to this?


jneira:
> Can we move the previous pr to this?

If this one is written onto the old one and it became obsolete, we can simply close the old one

plugins/hls-call-hierarchy-plugin/test/Main.hs
```diff
@@ -164,6 +164,20 @@ prepareCallHierarchyTests =
           selRange = mkRange 1 13 1 14
           expected = mkCallHierarchyItemC "A" SkConstructor range selRange
       oneCaseWithCreate contents 1 13 expected
+  , testGroup "type singature"
```
jneira, line 167:
```suggestion
  , testGroup "type signature"
```

small typo

July541: 😅

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+![Call Hierarchy in Emacs](call-hierarchy-in-emacs.gif)
+
+![Call Hierarchy in VSCode](call-hierarchy-in-vscode.gif)
+
+## Prerequisite
+None. You can experience the whole feature without any setting.
+
+## Configuration
+Enable default. You can disable it in the setting whenever you like.
```
jneira, line 17:
```suggestion
Enabled by default. You can disable it in your editor settings whenever you like.

```json
{
  "haskell.plugin.callHierarchy.globalOn": true
}
```
```

Comment (jneira) [reply plugins/hls-call-hierarchy-plugin/README.md:17]: @July541 have you checked the hls commands related with the config include hierarchy (`haskell-language-server vscode-extension-schema` and `haskell-language-server generate-default-config` ? I think the generic config option to enable/disable globally it is `haskell.plugin.callHierarchy.globalOn` but not sure at 100%

**Comment by jneira** [comment]
It would be great add a link to the plugin readme in the main repo readme

File: plugins/hls-call-hierarchy-plugin/README.md [hunk 3]
```diff
@@ -0,0 +1,27 @@
+# Call hierarchy plugin for the [Haskell Language Server](https://github.com/haskell/haskell-language-server#readme)
+
+The call hierarchy plugin can review the code to determine where functions are called and how they relate to other functions.
+
+This plugin is useful when debugging and refactoring code because it allows you to see how different parts of the code are related. And it is more conducive for users to quickly understand their macro architecture in the face of strange code.
```
jneira, line 5: :heart: this intro

plugins/hls-call-hierarchy-plugin/README.md
July541: I tested the two commands, it's really like you said!

July541:
@jneira I added a table list of all plugins we have currently, is this your desire?

jneira:
> @jneira I added a table list of all plugins we have currently in the main readme, is this your desire?

The table looks great but maybe it is better to have done the plugin docs and then add the table. That table could live in the issue about (#2066) while it is unfinished and move to README later. Write the table in a comment in that issue would be fine. There are some missing plugins which live in ghcide but dont worry, we can add them after moving it in a comment of #2066.

For now i was thinking in adding a item in the list of features, somewhat like

```
* Function call hierarchy: [see demo here](/link/to/call-hierarchy/readme.md#demo)
* Many more (TBD)
```

hls-plugin-api/hls-plugin-api.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 2.4
 name:          hls-plugin-api
-version:       1.2.0.0
+version:       1.2.1.0
```
pepeiborra, line 3:
```suggestion
version:       1.2.0.1
```

plugins/hls-call-hierarchy-plugin/hls-call-hierarchy-plugin.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               hls-call-hierarchy-plugin
-version:            1.0.0.0
+version:            1.0.1.0
```
pepeiborra, line 3:
```suggestion
version:            1.0.0.1
```

plugins/hls-call-hierarchy-plugin/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Configuration
+Enabled by default. You can disable it in your editor settings whenever you like.
+
+```json
+{
+  "haskell.plugin.callHierarchy.globalOn": true
+}
+
+## Change log
+### 1.0.1.0
```
pepeiborra, line 25:
```suggestion
### 1.0.0.1
```

pepeiborra (approved):
Very nice! Just note that the changes to both packages are "other changes" according to the [PVP](https://pvp.haskell.org) and the version bumps can be therefore smaller.

July541:
@jneira I removed the edit of the main readme, you may insert the features when you have the final idea.

jneira (approved):
LGTM, thanks again!

Ailrun (approved):
Awesome, thank you!

PR #2077, opened 2021-08-03
Wingman: "Intro and destruct" code action

isovector:
The number of times I find myself refining a lambda and then immediately destruct it is staggering, so I thought --- why not put them together into one.


jneira (approved):
Really handy!

PR #2078, opened 2021-08-03
Amend fix for correct placement of file header pragmas (#1958)

nini-faroux:
hi again, I noticed two issues with the previous fix. 

The first was the case where the pragma types were interchanging, like: 
{-# OPTIONS_GHC ... #-}
{-# LANGUAGE ... #-}
{-# LANGUAGE ... #-} -- newly added, should be last
{-# OPTIONS_GHC ... #-}

Here the line number for the last LANGUAGE pragma was always calculated last and used even if it wasn't the largest of the three. This is fixed now, by taking the max.

The other issue was when cleaning things up last time I changed the 'afterShebang' function to avoid some duplication, the tests didn't have a case to check for this but it worked out that when there is no Language pragma at the top of the file this would basically match any type of pragma later in the file instead of none. This is fixed now just by calculating the shebang separately to LANGUAGE and OPTIONS_GHC.

I added a bunch of new tests too just for different types of interchanging and ignoring later pragmas like the BangPatterns example in the original issue.


pepeiborra (approved):
Thank you so much for the fixes and the amazing test suite. 

GHC has a parser for module headers (https://hackage.haskell.org/package/ghc-8.10.2/docs/Parser.html#v:parseHeader) that might be worth considering if/when our custom logic grows out of control. But we're not there yet, and I'm not sure the GHC parser would cover all the cases.

nini-faroux:
Oh cool, I'll have a look at the parser. I think this covers a number of corner cases hopefully but I'm sure there's others that will come up too. If I think of any more I'll add them.

Thanks again!

Issue #2033, opened 2021-07-22
HLS needs a website!


fendor:
Good morning everyone!

While I don't really think we need additional motivation, here are some reasons why we need a website:

**Structure**
HLS's README is quite overflowing, presenting itself, showing off features, installation guides, how to debug issues, configuration, how to get started with developing, etc...

All in *one* file! That is too much to stomach for most people, a website can bring structure into this mess, help with orientation and so on.

**News and blog posts**
If we have any news regarding IDE development, there is nothing to post that to! Someone has to write a text and post it to all relevant news sites (e.g. discourse, reddit, Hackernews (no idea whether anyone posts there actually)) and so on. It would be simpler if there was a news section on the website, where people can write news and blog posts that can be easily reviewed! Additionally, it serves as an archive of what we have announced already.

**Documentation**
Some plugins are non-trivial to use. Writing documentation in a readme is not the most discoverable way of advertising said features. On a website, we can have dedicated tutorials, not buried somewhere in the repository.

--- 

I think it is worth having a website dedicated to the Haskell Language Server!

jneira:
It is a nice idea, but a website... has to be maintained. It could be setup as github pages or we could open the project wiki.
Other alternative is create md docs in the documentation folder.

jneira:
As a POC i've published the docs folder in `https://haskell.github.io/haskell-language-server`, as it has not index.html it throws a 404 but you can download the md files, f.e. `https://haskell.github.io/haskell-language-server/plugin-tutorial.md`

googleson78:
Something in the style of readthedocs perhaps? (like ghc and agda currently do)

konn:
I strongly agree!
As a marginal note: I had even proposed it once as a Discussion: https://github.com/haskell/haskell-language-server/discussions/1319 but it got silently rejected at that time.
As HLS has been evolving since then, the presence of the dedicated documentation/frontpage makes much more sence than the past.

jneira:
Well if we can reorganize and polish docs folder and it will be automatically published in https://haskell.github.io/haskell-language-server the work to maintain maybe worths it, as we will have a good docs folder *and* a page with thos docs

michaelpj:
+1 for readthedocs, I have found it very good.

Ailrun:
one more +1 fro readthedocs, it's quite maintainable in my exprience.

jneira:
I like readthedocs too so prs welcome to add them, in this repo if possible

[...]

PR #2083, opened 2021-08-07
Add initial sphinx doc site for RTD

michaelpj:
This sets up an initial sphinx doc site that we can use with RTD. I would need some help from a repository admin to get it building on RTD itself, but you can build the site locally:
```
nix-shell
cd docs
make html
firefox _build/html/index.html
```

I moved all the content from the README to the docs, with the intention being to eventually just link to the RTD site from the README. Otherwise I haven't touched the content, apart from fixing a few links and header depths. There's a lot more organization etc. that we could do, but probably better to do that after we have an initial version.

I set up sphinx to use markdown, since I know people like that. The exception is the index pages, which I think need to be in RST for the toctree definitions. I do like RST, especially for better cross-reference, but the markdown version seems to work fairly well.

Fixes https://github.com/haskell/haskell-language-server/issues/2033.

(cc @berberman since I had to fiddle with the Nix stuff a bit)


pepeiborra:
You are an admin now

flake.nix
```diff
@@ -180,6 +180,19 @@
...
+          # For https://github.com/readthedocs/sphinx_rtd_theme/pull/1185, otherwise lists are broken locally
+          src = pkgs.fetchFromGitHub {
+            owner = "readthedocs";
+            repo = "sphinx_rtd_theme";
+            rev = "34f81daaf52466366c80003db293d50075c1b896";
+            sha256 = "0rkrsvvqr6g2p3v5vq88jhfp5sd0r1jqjh3vc5y26jn30z8s4fkz";
+          };
+        });
+        pythonWithPackages = pkgs.python3.withPackages (ps: [ps.sphinx myst-parser sphinx_rtd_theme]);
+
```
berberman, line 195: Can we wrap them into a derivation, which will essentially invoke `docs/Makefile` ?

michaelpj: We can definitely have a derivation that builds the doc site, but it's nice to have sphinx and the required packages in the shell so you can build them easily when you're working on them.

michaelpj:
Okay, I set up the build and it builds PRs too, so you can see a link to a preview from the PR status.

flake.nix
berberman: > but it's nice to have sphinx and the required packages in the shell so you can build them easily when you're working on them

Yes, but we can enter the dev shell of that derivation, not HLS's. So people don't need to download and build sphinx things while developing the Haskell parts.

michaelpj: Personally I find it nice if running `nix-shell` gives you what you need for everything in the project. Otherwise it can be quite obscure to figure out how to get the right shell. As it is, there is a "blessed" shell which is `devShell`/what you get if you just run `nix-shell`. I think the convenience of having that "just work" outweighs not having to download Sphinx.

michaelpj: I pushed a derivation for the docs.

jneira: i know I would ask for a chimera, dear fellow *nix hackers, but it would be nice to be able to build the website in windows native...
will investigate myself 😝

michaelpj:
I should add some instructions about how to build the doc site into the docs 

@jneira I'm afraid I don't know what to do about Windows. I think you should "just" need to install python and the required packages however you do that... There is this little requirements.txt I had to make for RTD, perhaps you can use that and pip or something?

jneira:
yeah don't worry, with a little bit of luck it only would need some python setup
A install guide would be great, will update it with the windows part

michaelpj:
Okay, I added a section about building the docs. I think this is good to go if we're happy with how it looks and we want to go with RTD.

jneira:
I've added the make.bat auto generated by sphinx-quickstart, but it is failing at a first blind attempt:

```
PS D:\hls\docs> .\make html
Running Sphinx v4.1.2

Extension error:
Could not import extension myst_parser (exception: No module named 'myst_parser')
```

jneira:
`pip install myst-parser && pip install sphinx_rtd_theme` did the trick

.readthedocs.yaml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
jneira, line 1: how could we use this config file (mentioning requirements.txt)? can we avoid the manual installation using pip?

michaelpj: No, this is for RTD.

michaelpj:
I think you just need to do `pip install -r requirements.txt`

michaelpj:
I'll add that to the doc.

michaelpj:
Done

jneira (approved):
docs looks fantastic, many thanks to spend time on this, documentation usually has not the attention it deserves so i value specially pr's like this, congrats!

jneira:
> I moved all the content from the README to the docs, with the intention being to eventually just link to the RTD site from the README

Mmm not sure about this, i wouldlike to see some brief content in the README, no a copy from the full docs but a resume or something for those who want to take a quick look

.readthedocs.yaml
jneira: Oh yeah i wanted to comment on that file, sorry

jneira:
The nix build is failing with

```
 error: builder for '/nix/store/31wvaghxbj402a1f1nl82mz1rrrr2az8-ormolu-0.1.4.1.drv' failed with exit code 1;
       last 10 log lines:
       >
       > src/Ormolu/Printer/SpanStream.hs:17:1: error:
       >     Could not find module ‘SrcLoc’
       >     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
       >    |
       > 17 | import SrcLoc
       >    | ^^^^^^^^^^^^^
       > [ 7 of 48] Compiling Ormolu.Processing.Common ( src/Ormolu/Processing/Common.hs, dist/build/Ormolu/Processing/Common.o, dist/build/Ormolu/Processing/Common.dyn_o )
       > [ 8 of 48] Compiling Ormolu.Processing.Cpp ( src/Ormolu/Processing/Cpp.hs, dist/build/Ormolu/Processing/Cpp.o, dist/build/Ormolu/Processing/Cpp.dyn_o )
       > [ 9 of 48] Compiling Ormolu.Processing.Postprocess ( src/Ormolu/Processing/Postprocess.hs, dist/build/Ormolu/Processing/Postprocess.o, dist/build/Ormolu/Processing/Postprocess.dyn_o )
```

michaelpj:
> Mmm not sure about this, i would like to see some brief content in the README, no a copy from the full docs but a resume or something for those who want to take a quick look

Yeah, not sure what to do about this. It would be nice not to repeat ourselves. Maybe we can write a section that we can include into both? There's not much introductory prose at all even in the current docs, just a sentence.

Maybe I can link to a few specific interesting sections like "features", "building the code", "contributing guidelines".

> The nix build (not required though) is failing with

It's broken on master in the same way :shrug:

PR #2084, opened 2021-08-07
Allow ormolu 0.2 and fix compatibility with GHC 9

felixonmars:
Ormolu 0.2 fixes compatibility with GHC 9. Turns out adding `ghc-api-compat` here is enough for GHC 9 compatibility.

Builds fine and all tests pass.


Ailrun (approved):
Awesome, thank you!

PR #2086, opened 2021-08-08
Enable tests for ormolu plugin

felixonmars:
It was never enabled in github workflow.

Also we should test it with GHC 9 now.


Ailrun (approved):
Oops, thank you for noticing and fixing this!

PR #2090, opened 2021-08-11
Add a bit more prose and some links to the README

michaelpj:
Just add links to the top-level sections to make it a bit less bare.


Ailrun (approved):
Awesome!

jneira (approved):
thanks, it looks much better now

PR #2094, opened 2021-08-13
Wingman: New AbstractLSP interface

isovector:
This PR introduces a new `Interaction` type, which bundles up all of the repeated machinery necessary to implement commands/handlers. It provides the same API to code lenses and actions, and abstracts away a lot of the pain of dealing with LSP directly. 

I've ported Wingman's code actions over to this new system. This is purely a refactoring, but cleans up a lot of the technical debt I'd been accumulating.


jneira (approved):
lgtm, do you think it (or parts of it) could be generalized to other plugins and move it to hls-plugin-api?

isovector:
> do you think it (or parts of it) could be generalized to other plugins and move it to hls-plugin-api? 

Yeah, probably, but I want to play with it for a while to make sure there are no glaring problems.

PR #2102, opened 2021-08-16
Wingman: streaming tactic solutions

isovector:
This PR updates to refinery v5, which supports streaming tactic results. This means that when Wingman times outs, it can still pick the best solution it found, rather than failing entirely.  

Needs to land after #2099


jneira (approved):
nice improvement!

Issue #1068, opened 2019-10-06
Reload required when adding new dependencies to package.yaml/.cabal file


ratherforky:
When adding new dependencies to `package.yaml` you need to reload VSCode to keep ghcide up to date.

To reproduce:

- Add a new dependency to the `package.yaml` of a project
- Import a module from that new dependency
- You get an error for the import
- Reload and the error should go away

Adding a new dependency happens relatively infrequently so this isn't a big deal, but I thought it might be worth noting as a limitation if it isn't easy to fix.

stevenfontanella:
Can we reopen this since the ghcide project has been archived? I've also seen this behavior when updating the .hlint.yaml; vscode needs to be reloaded for its changes to take effect.

jneira:
hi, thanks to note it, however, all the issues were translated to this repo, ghcide#50 is https://github.com/haskell/haskell-language-server/issues/1113 which is also closed, so reopening this one

stevenfontanella:
Thanks for reopening. I didn't know how to find the corresponding issue for ghcide#50. Now that you linked it I see that that issue was supposed to be solved. Do we know what's wrong? I'm still able to reproduce this using the steps from @ratherforky here. Of course it's possible that something is wrong with my setup too. I'm running `haskell-language-server-1.3.0-linux-8.10.4` and the haskell vscode extension 1.6.0.

ratherforky:
The other issue thread mentions a PR being worked on, but doesn't seem to have an actual PR linked to it, so it seems like it was just never implemented (certainly I still reload every time I add a dependency with HLS)

ratherforky:
I found the commit talked about in that other issue (the commit message linked to the issue with a url rather than a # to the issue number, so that might be why it didn't get linked back to the issue): https://github.com/haskell/haskell-language-server/commit/0782a6adb05909c992c3d207c4c4986eba220b16

Pull request: https://github.com/haskell/ghcide/pull/408

The test for it is `sessionDepsArePickedUp`: https://github.com/haskell/haskell-language-server/blob/5d83b6382abaecbee0c5cdf97be02fbd518611b7/ghcide/test/exe/Main.hs#L5174

It seems to only test the case where `hie.yaml` is changed directly rather than a `package.yaml` or `*.cabal` file being changed, and only for a new language extension being added to the cradle. I always use Stack, so I never have a `hie.yaml` file. 

[...]

PR #2060, opened 2021-08-01
Reimplement shake (continued)

pepeiborra:
Rebase of #1759 including bugfixes

- [x] pass all the tests
- [x] fix performance regressions (or mitigate via reverse deps)
- [x] implement reverse deps
- [x] test scalability
- [x] implement applicative parallelism (left for future work)

# Summary of changes

* hls-graph is now a full implementation of a reactive dependency graph (aka build system) and no longer wraps Shake. 
* "reactive" means that his-graph tracks reverse dependencies in order to minimize rebuilds when a list of external changes (e.g. which file has been edited) is provided. This in turn means that ghcide now scales much better - rebuilds are no longer linear in the number of transitive module imports (modulo the TH `NeedsCompilation` rule)
* A new flag `--conservative-change-tracking` is added to ghcide to disable reactive change tracking. The flag is intended for debugging only and not for users. A new benchmark example "cabal-conservative" uses it to measure the performance difference.
* Performance in the Cabal example is equal or better (up to 2X) than before
* Build profiling is supported and extends the Shake profiling to include information about dirty nodes and visited nodes.
* Fake `Binary` instances no longer needed and removed from the codebase

# What's missing

* Applicative parallelism - left for future work, hoping someone smarter than me can do it. Doesn't seem to have a major impact, probably will speed up startup times a bit?

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2060"></a>

PR #2106, opened 2021-08-17
Fix filepath identity in cradle dependencies when using reactive change tracking

pepeiborra:
When doing reactive change tracking we track the set of dirty keys as opposed to conservative change tracking where we check every key on every build.

Filepath identity becomes even more important in this setting, opening a new type of bug where the filepath used in the dirty key is equivalent but not structurally equal to the filepath used in the dependency graph.

This can happen e.g. with cradle dependencies where hie-bios gives us a set of relative paths and we create GetFileModification keys for them as dependencies of the GhcSession, but the LSP client raises FileWatched notifications using absolute paths that we use to create dirty keys.

To prevent this class of bugs, we would need to strengthen the notion of NormalizedFilePath to make it always absolute/relative. In practice, since 99% of the paths are introduced by the LSP client, we just need to make sure that the other 1% (introduced by e.g. hie-bios) follow the same convention.

This commit fixes the hie-bios paths to make them absolute before creating rule keys, following the LSP convention.

NOTE: reactive change tracking is not enabled or even implemented anywhere, so this change is a no-op for everyone who is not using #2060

### DIFF
--- a/ghcide/bench/lib/Experiments.hs
+++ b/ghcide/bench/lib/Experiments.hs
@@ -153,10 +153,10 @@ experiments =
                 forM_ identifierP $ \p -> changeDoc doc [charEdit p]
         )
         ( \docs -> do
-            Just hieYaml <- uriToFilePath <$> getDocUri "hie.yaml"
-            liftIO $ appendFile hieYaml "##\n"
+            hieYamlUri <- getDocUri "hie.yaml"
+            liftIO $ appendFile (fromJust $ uriToFilePath hieYamlUri) "##\n"
             sendNotification SWorkspaceDidChangeWatchedFiles $ DidChangeWatchedFilesParams $
-                List [ FileEvent (filePathToUri "hie.yaml") FcChanged ]
+                List [ FileEvent hieYamlUri FcChanged ]
             forM_ docs $ \DocumentPositions{..} ->
               changeDoc doc [charEdit stringLiteralP]
             waitForProgressDone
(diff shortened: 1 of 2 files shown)


jneira (approved):
lgtm, thanks for split it from the big pr
could be it related with the fact hls does not reloads when config files from hie-bios change? (see #1068)

EDIT: probably, as @fendor noted the solution could be make hie-bios return absolute paths

fendor:
One of the issues I encountered was that there is no reliable way with stack to find package.yaml (or .cabal) files, e.g. when a package is in a sub-directory, then the component directory is equal to the root directory, but package.yaml (or .cabal) is actually located in the sub-directory.

I don't think that has changed, for stack canonicalising will probably not always fix the issue.

pepeiborra:
I think it might be causing #1068 yes. Initially I assumed that this was a complete no-op because reactive builds are not enabled yet, but on second thoughts we already play [some tricks](https://github.com/haskell/haskell-language-server/blob/2ee29436fb3377f6c4bb7ece72595cea2dfd5f17/ghcide/src/Development/IDE/LSP/Notifications.hs#L87) with `GetFileExists` and `GetFileModification` relying on LSP `FileWatched` notifications to remove the `alwaysRerun` dependency, and file path identity will be relevant there too.

So, to summarise, I believe this could well fix #1068 which must have regressed in #1880

ratherforky:
I can confirm that making the hie-bios paths absolute should fix this problem (assuming the changes will also affect `ghcide/session-loader/Development/IDE/Session.hs`). I just discovered this independently and came to the repo to tell people, but it looks like I've been beaten to the punch!

I was suspicious of this line: https://github.com/haskell/haskell-language-server/blob/5d83b6382abaecbee0c5cdf97be02fbd518611b7/ghcide/session-loader/Development/IDE/Session.hs#L286 

...since it should include the hie.yaml file (which is correctly tracked) and the cradle dependencies (which are not correctly tracked), so I threw a trace on it and found for my test project: `deps = ["depsReloading.cabal","package.yaml","stack.yaml","~/Playground/depsReloading/hie.yaml"]`

So the tracked file is absolute and the non-tracked files are relative. I then hard-coded the cradle dependencies to be absolute as well, and sure enough that completely fixed the problem for me (using a default stack setup)

PR #2009, opened 2021-07-09
Draft: Add POC for HLS loading multiple components on startup

fendor:
Powered by hie-bios.

A prototype implementation for loading multiple components on startup using the WIP hie-bios interface: https://github.com/mpickering/hie-bios/pull/301

The main feature is the usage of `cabal show-build-info`, from here: https://github.com/haskell/cabal/pull/7478 
To test out this PR, you have to install the linked cabal executable from source.
NB: currently cabal only supports loading projects where the build-type is `Simple`.

For debugging, take a look at the result of `cabal build all` in your project, which  *currently must* succeed for `show-build-info` to succeed.

PR #2108, opened 2021-08-17
New rename plugin implementation

### LINKED ISSUE #282: Retrie: implement high-level refactorings including renaming
High-level refactorings should be doable by combining minimal code generation with retrie commands.

## Rename

To rename `foo` to `bar`, proceed as follows:

1. Rename `foo` to `bar` in place

2. Generate a new declaration:
```
foo = bar
```

3. Tell retrie to unfold `foo`

## Move

Similar to rename, but in step 1 we move `bar` to the target module, while in step 2 we generate `foo` in the source module.

## Extract definition

1. Ask the user for the desired name 
2. Generate a new declaration with the desired name, using the selected subexpression as the function body, and the unbound variables as formal parameters
3. Tell retrie to fold the new name

OliverMadine:
A new PR for my [GSoC](https://summerofcode.withgoogle.com/projects/#5971774971314176) rename plugin since I'm changing my approach to the solution. 

---

After spending some time implementing a solution with Retrie, I've realised it would be better just to modify the source directly (using SYB).

The advantages are:
- Compatibility with earlier ghc versions
- No issues with Retrie compatibility on windows
- We can rename non top-level names
- Easier solution to maintain

I am leaving the plugin disabled by default since we[ still need all relevant files to be indexed ](https://github.com/mpickering/hie-bios/issues/269)to ensure correctness. 
The work towards a full solution to this [has been started](https://github.com/mpickering/hie-bios/pull/301#issue-686541322), however, if we get #2009 merged, then I can look into making a temporary solution (by assuming the `hie.yaml` lists all components).

## Example
Renaming 231 occurrences of the `IdeState` data type/constructor in HLS.
HLS rebuilt successfully after this rename!

https://user-images.githubusercontent.com/30090176/129815071-fffff38e-57a7-4fd2-b797-6a97aa83c372.mov

Closes #282.


plugins/hls-rename-plugin/hls-rename-plugin.cabal
```diff
@@ -0,0 +1,51 @@
...
+    , containers
+    , ghc
+    , ghc-exactprint
+    , ghcide                >=1.4     && <1.5
+    , hiedb
+    , hls-plugin-api        ^>=1.2
+    , hls-retrie-plugin     >=1.0.1.1
+    , lsp
+    , lsp-types
+    , retrie                >=0.1.1.1
```
pepeiborra, line 33: remove the retrie depends

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+renameProvider :: PluginMethodHandler IdeState TextDocumentRename
+renameProvider state pluginId (RenameParams (TextDocumentIdentifier uri) pos _prog newNameText) =
+    response $ do
+        nfp <- safeGetNfp uri
+        oldName <- (handleMaybe "error: could not find name at pos" . listToMaybe) =<<
+            getNamesAtPos state pos nfp
+        refs <- refsAtName state nfp oldName
+        refFiles <- mapM safeGetNfp (nub [uri | Location uri _ <- refs])
```
pepeiborra, line 46: avoid `nub`, use `HashSet.toList . HashSet.fromList` instead

pepeiborra (changes requested):
Looks good to go with only minor comments!

plugins/hls-rename-plugin/hls-rename-plugin.cabal
OliverMadine: We use retrie for Retrie.ExactPrint. Should I try to avoid this?

mpickering:
What is the performance of this plugin like on a big project?

OliverMadine:
> What is the performance of this plugin like on a big project?

As a rough approximation, renaming 221 occurrences of the `IdeState` data type in HLS took around 15 seconds on my machine. I've attached a video example of this in the PR description.

plugins/hls-rename-plugin/hls-rename-plugin.cabal
pepeiborra: Are those mostly reexportarlo? If so yes, otherwise keep

pepeiborra:
> > What is the performance of this plugin like on a big project?
> 
> 
> 
> As a rough approximation, renaming 221 occurrences of the `IdeState` data type in HLS took around 15 seconds on my machine. I've attached a video example of this in the PR description. 

How does it compare to the retrie version of the plugin?

OliverMadine:
> How does it compare to the retrie version of the plugin?
This version actually seems much faster; The Retrie version was taking over 1 min for this rename. Maybe there was a bug or some overhead from calling Retrie individually for each file on that branch.

pepeiborra (approved):

pepeiborra:
This is merged. Congratulations @OliverMadine on landing the first version of the rename plugin!

jneira:
I am not able to see the demo in my firefox, @OliverMadine could you attach a demo in gif format to add to ChangeLog and other docs?

OliverMadine:
> I am not able to see the demo in my firefox, @OliverMadine could you attach a demo in gif format to add to ChangeLog and other docs? thanks!

Here you go.

rename-demo

Ciantic:
Is this still disabled as mentioned in the parent post? I can't seem to get it working on the brand new `stack new` project. It's not visible in quick fixes nor it's available from my normal shortcut:

image

OliverMadine:
> Is this still disabled as mentioned in the parent post?

Unfortunately it is still disabled (#2193) :(

You can enable it for use on a single component projects by building the HLS with `cabal build --flag="+rename"` and setting the `Server Executable Path` in vscode to the resulting executable.

Also note once it is enabled you can use it through the rename command, it is not a code action.

Karthik-Dulam:
Has this been merged into the release builds?
To me at least rename is not supported by the LSP.
(Using nvim-lsp.)

Issue #2100, opened 2021-08-16
Import Regression: Wrong Import Placement on first Import


emilypi:
Here's a weird one: first imports are inserted in wrong place. Unsure why. 

### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:
```sh
askell-language-server version: 1.3.0.0 (GHC: 8.10.4) (PATH: /Users/emilypi/.ghcup/bin/haskell-language-server-wrapper-1.3.0) (GIT hash: e7c5e90b6df5dff2760d76169eddaea3bdd6a831)
Tool versions found on the $PATH
cabal:		3.4.0.0
stack:		2.5.1
ghc:		8.10.4
```
Which OS do you use:
MacOS Big Sure

Which lsp-client do you use:

Emacs

Describe your project (alternative: link to the project):
*.cabal files, cabal.project

Contents of `hie.yaml`:
```yaml
cradle:
  cabal:
    - path: "src"
      component: "lib:order"

    - path: "test"
      component: "order:test:doctests"
```

### Steps to reproduce

1. Make a first import with functions defined (along with comments)
<img width="1219" alt="Screen Shot 2021-08-16 at 11 44 07 AM" src="https://user-images.githubusercontent.com/10459342/129606779-e6fe60dd-216b-4ffb-80cd-b51bc381cf75.png">

2. Import goes in wrong place.
<img width="1102" alt="Screen Shot 2021-08-16 at 11 44 15 AM" src="https://user-images.githubusercontent.com/10459342/129606797-db4a0c5e-c720-4498-a597-d5dc41abf703.png">

3. If importing another thing, imports are added below first import 

<img width="1205" alt="Screen Shot 2021-08-16 at 11 46 12 AM" src="https://user-images.githubusercontent.com/10459342/129606894-8aef89db-b051-46f3-95c3-9193ec77d104.png">

4. However, when physically moved, to the right spot, imports work as expected. 
<img width="1207" alt="Screen Shot 2021-08-16 at 11 44 40 AM" src="https://user-images.githubusercontent.com/10459342/129606815-7766e60f-d31b-4053-b900-5482f5022669.png">
<img width="1073" alt="Screen Shot 2021-08-16 at 11 44 46 AM" src="https://user-images.githubusercontent.com/10459342/129606823-2e57b6ab-e7aa-4a2b-88fb-7e1bdf158ad1.png">

### Expected behaviour

[...]

PR #2116, opened 2021-08-20
Add fix for correct placement of import (#2100)

nini-faroux:
hi, I added a fix for the import placement issue. (#2100)

So, the way it worked previously was to use the 'HsModule' type to find the next import position. Specifically it had two cases, when there's no existing imports in the file it would use the 'hsmodDecls' field, with type [LHsDecl name], and otherwise it would use 'hsmodImports', with type [LImportDecl name]'.

In the first case (hsmodDecls) the position would be one above the first element in that list, and this worked fine in general unless there was something above it which wasn't included in the list, like the comment in the example in the issue.

So for the fix, firstly it's changed to use the 'hsmodName' and 'hsmodExports' fields in 'HsModule' instead of 'hsmodDecls' to calculate the position in the case that the file included a 'module ... where' declaration, with no existing imports in the file. So in this case the new import will be added directly under the module declaration. 

But in the case where there is no module declaration and no existing imports I used the same logic as in the pragma plugin, basically the import will be one under the last file-header pragma or else at the very top. I needed to change a few functions to do this, just to pass down the Text version of the file contents, as there wasn't enough information in the 'HsModule' type to manage it otherwise.

I left the other case, where there are existing imports in the file the way it was, as I think inserting the next import directly underneath the last existing import is a sensible thing to do. This would only be a problem I guess if the user has placed the existing import in the wrong place themselves, or if there's another bug that does that.

I also added a good few tests, I hope it covers most of the cases!


pepeiborra:
Looks like the 8.6.5 build needs some compatibility fix:

```
src/Development/IDE/Plugin/CodeAction.hs:88:69: error:
1734
Error:     Module ‘SrcLoc’ does not export ‘HasSrcSpan(..)’
1735
   |
1736
88 | import           SrcLoc                                            (HasSrcSpan (..),
1737
   |                                                                     ^^^^^^^^^^^^^^^
1738
```

The usual recommendation is to `import Development.IDE.GHC.Compat` instead of direct GHC modules.

nini-faroux:
Ah ok, that extra import wasn't needed at all so, I'll update it tomorrow.

nini-faroux:
I removed that extra 'HasSrcSpan' import, but actually just checking now I noticed I'll have to amend the expected value of some of the older import tests too, as there's a few of them failing now because of the different behaviour - I can do this tomorrow.

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1282,26 +1284,70 @@ constructNewImportSuggestions exportsMap (qual, thingMissing) notTheseModules =
...
-newImportInsertRange :: ParsedSource -> Maybe (Range, Int)
-newImportInsertRange (L _ HsModule {..})
+-- | Finds the next valid position for inserting a new import declaration
+-- If the file already has existing imports it will be inserted under the last of these,
+-- it is assumed that the existing last import declaration is in a valid position
+-- If the file does not have existing imports, but has a (module ... where) declaration,
+-- the new import will be inserted directly under this declaration (accounting for explicit exports)
+-- If the file has neither existing imports nor a module declaration,
+-- the import will be inserted at line zero if there are no pragmas,
+-- otherwise inserted one line after the last file-header pragma
```
pepeiborra, line 1300:
```suggestion
-- | Finds the next valid position for inserting a new import declaration
-- * If the file already has existing imports it will be inserted under the last of these,
-- it is assumed that the existing last import declaration is in a valid position
-- * If the file does not have existing imports, but has a (module ... where) declaration,
-- the new import will be inserted directly under this declaration (accounting for explicit exports)
-- * If the file has neither existing imports nor a module declaration,
-- the import will be inserted at line zero if there are no pragmas,
-- * otherwise inserted one line after the last file-header pragma
```

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1282,26 +1284,70 @@ constructNewImportSuggestions exportsMap (qual, thingMissing) notTheseModules =
...
-            in Just ((srcLocLine $ realSrcSpanEnd s,col), col)
-        _            -> Nothing
+      [] -> findPositionNoImports hsmodName hsmodExports fileContents
+      _  -> findPositionFromImportsOrModuleDecl hsmodImports last True
     = Just (Range insertPos insertPos, col)
   | otherwise = Nothing
 
+-- | Insert the import under the Module declaration exports if they exist, otherwise just under the module declaration
+-- If no module declaration exists, then no exports will exist either, in that case
+-- insert the import after any file-header pragmas or at position zero if there are no pragmas
```
pepeiborra, line 1311:
```suggestion
-- | Insert the import under the Module declaration exports if they exist, otherwise just under the module declaration.
-- If no module declaration exists, then no exports will exist either, in that case
-- insert the import after any file-header pragmas or at position zero if there are no pragmas
```

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -209,7 +209,7 @@ extendImportHandler' ideState ExtendImport {..}
                 it = case thingParent of
                   Nothing -> newThing
                   Just p  -> p <> "(" <> newThing <> ")"
-            t <- liftMaybe $ snd <$> newImportToEdit n (astA ps)
+            t <- liftMaybe $ snd <$> newImportToEdit n (astA ps) ""
```
pepeiborra, line 212: Will this do the right thing if the module has a module name but no imports?

nini-faroux: I see what you mean, because of the empty string? I think it will be fine for that case as if the module has a name but no imports the new import will be added directly under the module declaration via the 'hsmodName' type, or via 'hsmodExports' if the declaration contains explicit exports.

But it would be an issue actually when the module has neither a module name nor imports - in that case it relies on the 'fileContents' argument when calling 'findNextPragmaPosition', but if it's just getting an empty string from there it won't always be accurate. 

I guess calling 'GetFileContents' earlier in the "extendImportHandler'" function, and then having (fromMaybe "" contents) in the line you highlighted should cover that case? I can check this evening and add a test too.

pepeiborra (approved):
Other than the failing tests, this looks good to me. Thanks again @nini-faroux !

nini-faroux:
I think it should be ok for the windows tests now too. 

No worries! let me know if there's any issues with it. Thanks for the help @pepeiborra and @jneira

jneira:
hlint is sad:

```
Run ./fmt.sh
Downloading and running hlint...
#=#=#                                                                         

######################################################################## 100.0%
                                                                           0.1%
####################################################################      94.5%
######################################################################## 100.0%
ghcide/src/Development/IDE/Core/FileStore.hs:292:9-33: Warning: Use void
Found:
  () <$ uses GetModIface rs
Perhaps:
  Control.Monad.void (uses GetModIface rs)

1 hint
```

jneira:
> hlint is sad

i think it is caused by some hlint change, i ve tried to fix it with 6e9d0b

ndmitchell:
I just released a new version of HLint that had this new hint added.

PR #2128, opened 2021-08-27
Drop ghc-api-compat from dependency closure

fendor:
Requires re-structuring GHC API Compatibility.

We try to go with the latest GHC API and provide backwards compatible
functions for older GHC versions. The intention is that ghcide code is
written with the latest GHC API in mind, older GHC versions may have
inefficiencies in order to satisfy the newer API.

This way dropping support for older GHC versions is trivial (only delete
a couple of '#if ... #endif' and maybe update some imports), and ghcide
does suffer less from bit-rot. Additionally, we want to have the latest
and best GHC API, not some semi-outdated frankenstein.

The implementation adds a new Compat sub-hierarchy that defines
backwards compatible functions for the GHC API, where possible
higher-level abstractions.

The module hierarchy re-organisation of GHC is taken care of by
'Development.IDE.GHC.Compat.Core', which takes care of almost every
definition used by ghcide. If this becomes to unwieldy, it makes sense
to split it up into smaller logical pieces.

Design Decisions:
* In ghcide, no GHC API import is allowed except for `GHC`.
* Restrict usage of CPP to modules in Compat hierarchy as much as possible.
* Plugins are allowed to provide their Backwards Compat methods.
  * E.g. they may use CPP instead of ghcide's Compat modules.

TODOS:
* [ ] ~~Proper GHC 9.2 support~~ (blocked on https://github.com/facebookincubator/retrie/issues/36)
* [x] Plugins
* [x] Stack build files
* [x] Nix Build Files

Status:
With all supported plugins, HLS compiles for GHC 8.6.5 ... 9.0.1.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2128"></a>


fendor:
@jneira this PR can be merged before GHC 9.2 support lands

cabal-ghc921.project
```diff
@@ -0,0 +1,139 @@
+packages:
```
fendor, line 1: This file is mostly temporary and incomplete

ghcide/.ghci
```diff
@@ -1,29 +0,0 @@
-:set -Wunused-binds -Wunused-imports -Worphans -Wunused-matches -Wincomplete-patterns
```
fendor, line 1: This made me mad as it took me 15 minutes to realize the issue.

It looks outdated to me, thus removed it

ghcide/bench/example/HLS
```diff
@@ -1 +0,0 @@
-../../..
```
fendor, line 1: This was an accident, undo!

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -919,13 +888,7 @@ parseFileContents env customPreprocessor filename ms = do
...
                srcs2 <- liftIO $ filterM doesFileExist srcs1
 
-               let pm =
-                     ParsedModule {
-                         pm_mod_summary = ms
-                       , pm_parsed_source = parsed'
-                       , pm_extra_src_files = srcs2
-                       , pm_annotations = hpm_annotations
-                      }
+               let pm = mkParsedModule  ms parsed' srcs2 hpm_annotations
```
fendor, line 891:
```suggestion
               let pm = mkParsedModule ms parsed' srcs2 hpm_annotations
```

ghcide/src/Development/IDE/GHC/Compat.hs
```diff
@@ -5,212 +5,164 @@
 {-# LANGUAGE ConstraintKinds   #-}
```
fendor, line 5: This module can moved entirely to `Core.hs`. Probably sensible

ghcide/src/Development/IDE/GHC/Compat/Core.hs
```diff
@@ -0,0 +1,964 @@
+{-# LANGUAGE CPP               #-}
```
fendor, line 1: While very convenient to have every Identifier listed, might be a maintenance nightmare.
Rather re-export whole modules? Or split up into more smaller modules?

cabal-ghc921.project
jneira: I ve updated it in #2131

ghcide/bench/example/HLS
fendor: Apparently, it should not have ever been committed, and CI is still happy, so keep deleted

ghcide/src/Development/IDE/GHC/Compat.hs
pepeiborra: Would it be worth to keep it in order to reexport everything

ghcide/src/Development/IDE/GHC/Compat/Core.hs
pepeiborra: Reexport all

fendor: Sorry, do you mean to re-export modules, or like this, every identifier?

jneira:
@fendor will try to help with stack files asap

jneira:
Solver error for ghc-9.0.1:

```
Resolving dependencies...
[__0] trying: hls-1.3.0.0 (user goal)
[__1] trying: hls:+refineimports
[__2] trying: hls-refine-imports-plugin-1.0.0.0 (dependency of hls
+refineimports)
[__3] next goal: hls-plugin-api (user goal)
[__3] rejecting: hls-plugin-api-1.2.0.1 (conflict: hls-refine-imports-plugin
=> hls-plugin-api^>=1.1)
[__3] rejecting: hls-plugin-api-1.1.0.2, hls-plugin-api-1.1.0.1,
hls-plugin-api-1.1.0.0, hls-plugin-api-1.0.0.0, hls-plugin-api-0.7.1.0,
hls-plugin-api-0.7.0.0, hls-plugin-api-0.6.0.0, hls-plugin-api-0.5.0.1,
hls-plugin-api-0.5.0.0, hls-plugin-api-0.4.1.0, hls-plugin-api-0.4.0.0
(constraint from user target requires ==1.2.0.1)
[__3] fail (backjumping, conflict set: hls-plugin-api,
hls-refine-imports-plugin)
goals I've had most trouble fulfilling: hls-plugin-api, hls,
hls-refine-imports-plugin, hls:refineimports
Try running with --minimize-conflict-set to improve the error message.
```

hls-test-utils/hls-test-utils.cabal
```diff
@@ -44,7 +44,7 @@ library
     , ghcide                  ^>=1.4
     , hls-graph
     , hls-plugin-api          ^>=1.2
-    , hspec                   <2.8
+    , hspec                   <2.9
```
jneira, line 47: this breaks hls-test-utils(https://github.com/haskell/haskell-language-server/runs/3494467098?check_suite_focus=true):

```
[1 of 2] Compiling Test.Hls.Util    ( src/Test/Hls/Util.hs, /home/runner/work/haskell-language-server/haskell-language-server/dist-newstyle/build/x86_64-linux/ghc-8.10.2/hls-test-utils-1.1.0.0/build/Test/Hls/Util.o, /home/runner/work/haskell-language-server/haskell-language-server/dist-newstyle/build/x86_64-linux/ghc-8.10.2/hls-test-utils-1.1.0.0/build/Test/Hls/Util.dyn_o )
Configuring executable 'hls-wrapper' for hls-1.3.0.0..

src/Test/Hls/Util.hs:176:32: error:
Error:     ‘configOutputFile’ is not a (visible) constructor field name
    |
176 |                              , configOutputFile = Right $ subdir </> "results.xml"
    |                                ^^^^^^^^^^^^^^^^
```

jneira: After reviweing last version of hspec i've not found how to set the ouput file in the new configuration :thinking: 

https://hackage.haskell.org/package/hspec-core-2.8.3/docs/Test-Hspec-Core-Runner.html#g:2

fendor: I would rather limit the upper bounds and say, fix later

fendor: Has nothing to do in principle with this pr

ghcide/bench/example/HLS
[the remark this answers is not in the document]
fendor: It looks like this file is ignored in `.gitignore`. Maybe it should be deleted? Otherwise, I would add an exemption to `.gitignore`

jneira:
stack.yamls don't use hiedb from hackage i think

jneira:
hls-splice-plugin tests have totally failed 😟

fendor:
Yeah and I can reproduce the issue locally. Trying to debug it

fendor:
Found the issue, will push a fix tomorrow, although it might require more changes

ghcide/src/Development/IDE/GHC/Compat.hs
fendor: I figured, the purpose is roughly this:
`Core.hs` handles simple changes in the GHC API (e.g. renames of functions). Other Compat modules may import `Core.hs`, to avoid the amount of CPP statements.
`Compat.hs` itself, depends on 'Core.hs' and on the most common modules from the `Compat/*` hierarchy. As such, it can provide backwards compatible functions that depend in some way on modules from the `Compat/*` hierarchy. Practical example:
```haskell
runPp :: Logger -> DynFlags -> [Option] -> IO ()
runPp =
#if MIN_VERSION_ghc(9,2,0)
    SysTools.runPp
#else
    const SysTools.runPp
#endif
```

Older GHC versions don't have a logger type. Thus, we can't have this function in 'Core.hs'. However, since it is basically a simple rename, it is not really necessary to add another module in the  'Compat/*' hierarchy. Thus, this function is defined in 'Compat', as it depends on 'Compat/Env.hs' which defines a backward compatible type alias for 'Logger'.

I think that makes relatively sense?

ghcide/src/Development/IDE/GHC/Compat/Core.hs
fendor: Reexporting all modules seems like a bad idea, since we are omitting certain definitions on purpose. My compromise would be to re-export simple modules, such as 'NameEnv', 'Predicate' and the like, but keep the explicit listing for more complicated functions and types, such as 'DynFlags'.

fendor:
@jneira @pepeiborra this is ready for merge once CI is green, which is basically only waiting for nix

jneira:
> 
> 
> @jneira @pepeiborra this is ready for merge once CI is green, which is basically only waiting for nix

Lets merge #2156 first, to backport it to 1.3.0-hackage as commented here: https://github.com/haskell/haskell-language-server/pull/2156#issuecomment-913846865
This will have to be rebased afterwards, to remove again ghc-api-compat

jneira:
will try a rebase, i hope i dont break things even more

jneira:
@fendor I've done a rebase here: https://github.com/jneira/haskell-language-server/tree/ghc92-rebased

cabal-ghc901.project
```diff
@@ -6,19 +6,19 @@ packages:
          ./ghcide
          ./hls-plugin-api
          ./hls-test-utils
-        --  ./plugins/hls-tactics-plugin
-        --  ./plugins/hls-brittany-plugin
-        --  ./plugins/hls-stylish-haskell-plugin
-        --  ./plugins/hls-fourmolu-plugin
+        -- ./plugins/hls-tactics-plugin
```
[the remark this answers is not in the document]
isovector: No idea. Using the old GHC module names, I guess. How do other plugins work around this?

ghcide/ghcide.cabal
```diff
@@ -156,6 +155,15 @@ library
...
         Development.IDE.GHC.Compat
+        Development.IDE.GHC.Compat.Core
+        Development.IDE.GHC.Compat.Env
+        Development.IDE.GHC.Compat.Iface
+        Development.IDE.GHC.Compat.Logger
+        Development.IDE.GHC.Compat.Outputable
+        Development.IDE.GHC.Compat.Parser
+        Development.IDE.GHC.Compat.Plugins
+        Development.IDE.GHC.Compat.Units
+        Development.IDE.GHC.Compat.Util
```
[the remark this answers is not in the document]
fendor: Probably indeed. But I figured, sometimes it might be nice to have more fine-grained imports.

cabal-ghc901.project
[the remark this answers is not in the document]
fendor: Can't quite remember, I think it was mainly a matter of correcting imports but it is already pretty complicated to get this merged as it touches so many places.

ghcide/src/Development/IDE/GHC/Compat.hs
fendor: Unfortunately, there are naming collisions between `Core.hs` and `Util.hs`, and I would like to refrain from renaming functions just to avoid the naming collision.
Can we have the compromise that every module *except* for `Util` is re-exported?

Naming conflict:
https://hackage.haskell.org/package/ghc-8.10.2/docs/BasicTypes.html#t:SuccessFlag
https://hackage.haskell.org/package/ghc-8.10.2/docs/Maybes.html#t:MaybeErr

ghcide/bench/example/HLS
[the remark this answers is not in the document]
fendor: Added

cabal-ghc901.project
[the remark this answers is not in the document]
fendor: It is mainly adding the appropriate CPP statements since the AST changed quite considerably.

ghcide/src/Development/IDE/GHC/Compat.hs
fendor: Compat re-exports Outputable now

cabal-ghc901.project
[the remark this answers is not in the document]
anka-213: It was blocked on upstream dependencies when I worked on 9.0 support for hls, but that is fixed now. So I'm guessing it's just that no-one has taken the time to do it.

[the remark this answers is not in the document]
fendor: Exactly, and as I am quite pressed for time at the moment, I would appreciate if supported is added in a different PR-

jneira (approved):
LGTM, thanks for this again

ghcide/ghcide.cabal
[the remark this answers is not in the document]
anka-213: Yeah, I think it's nice to expose them separately, since using `Development.IDE.GHC.Compat` is even more course-grained than using the old GHC api. It's basically re-exporting all the parts of the GHC api that we need compatibility code for (unless that has changed since last I touched it).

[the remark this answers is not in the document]
fendor: Exactly. However, it is really non-trivial to have a more fine-grained API.

anka-213:
Thanks for cleaning up the mess I caused/contributed to when adding GHC-9.0 support! ❤️

fendor:
@anka-213 Thank you for your GHC 9.0 support contribution, without this would have been harder!
You patched so many dependencies, it is awesome!

cabal-ghc901.project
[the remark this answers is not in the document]
anka-213: Oh, I just looked through the diff of this PR and noticed that you have already made a lot of progress on the tactics plugin compatibility. I'm currently looking at what else needs to be fixed for the tactics plugin to support 9.0. (I'm only looking at making it compile and run properly, I'll leave the actual support for 9.0 specific features like linear types to @isovector)

PR #2133, opened 2021-08-28
added completion for language to { -# LANGUAGE

alexnaspo:
fixes https://github.com/haskell/haskell-language-server/issues/681

### DIFF
--- a/plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
+++ b/plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
@@ -164,7 +164,7 @@ completion _ide _ complParams = do
                     = J.List $ map buildCompletion
                         (Fuzzy.simpleFilter (VFS.prefixText pfix) allPragmas)
                     | otherwise
-                    = J.List []
+                    = J.List [buildLanguageSnippetCompletion]
                 result Nothing = J.List []
                 buildCompletion p =
                     J.CompletionItem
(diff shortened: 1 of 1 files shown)


pepeiborra:
I'm not sure what's going on here, but we already have completions for pragmas and they seem to work for me: 

https://github.com/haskell/haskell-language-server/blob/868e4768fd7d91352672060709c88f65d3102644/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs#L658-L671

Perhaps you could look into refining the existing completion snippet so that it inserts the pragma close tokens `#-}` and/or characterise whether it works in all situations?


https://user-images.githubusercontent.com/26626/131214843-883082cf-af8e-4996-b54c-2b2bd523d24c.mov

alexnaspo:
@pepeiborra - sure I can take a look at that. Do I you think I should move the auto completions into the pragmas plugin?

The difference here is that it auto completes on "language" rather than "{-# LANGUAGE" as the ticket requests

ndmitchell:
Haskell is case-insensitive for its pragmas, so `lAnGuAgE` works just fine. Perhaps the completions module just needs to be educated to that fact?

pepeiborra:
> @pepeiborra - sure I can take a look at that. Do you think I should move the auto completions into the pragmas plugin?

That would be great

alexnaspo:
> > @pepeiborra - sure I can take a look at that. Do you think I should move the auto completions into the pragmas plugin?
> 
> That would be great

I opened up a separate PR because it no longer closes the above ticket @pepeiborra

jneira:
@alexnaspo so could we close this one, as it is included in #2134 iiuc?

PR #2134, opened 2021-08-28
Move pragmas completion to pragmas plugin

alexnaspo:
* Move the basic pragmas completion from GCHIDE to the pragmas module
* Add closing pragmas snippet  "-#}"

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -299,11 +299,6 @@ mkExtCompl label =
     Nothing Nothing Nothing Nothing Nothing Nothing Nothing
     Nothing Nothing Nothing Nothing Nothing Nothing
 
-mkPragmaCompl :: T.Text -> T.Text -> CompletionItem
-mkPragmaCompl label insertText =
-  CompletionItem label (Just CiKeyword) Nothing Nothing
-    Nothing Nothing Nothing Nothing Nothing (Just insertText) (Just Snippet)
-    Nothing Nothing Nothing Nothing Nothing Nothing
 
 fromIdentInfo :: Uri -> IdentInfo -> Maybe T.Text -> CompItem
 fromIdentInfo doc IdentInfo{..} q = CI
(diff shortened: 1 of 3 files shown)


jneira:
@alexnaspo thanks for your contribution, i think makes sense group all pragmasrelated code together. Are not tests about pragmas completion in the ghcide test suite? They should be moved too in that case

alexnaspo:
> @alexnaspo thanks for your contribution, i think makes sense group all pragmasrelated code together. Are not tests about pragmas completion in the ghcide test suite? They should be moved too in that case

I am new to this project and I am having trouble finding affected tests. When I run the ghcide test suite I get 67 errors on master branch. Is it expected for tests to fail locally on master?

Could you help point me in the right direction for tests that need to be moved?

alexnaspo:
> @alexnaspo thanks for your contribution, i think makes sense group all pragmasrelated code together. Are not tests about pragmas completion in the ghcide test suite? They should be moved too in that case

@jneira It looks like the functionality for pragmas completion is actually already tested in the pragmas plugin, rather than in the ghcdie test suite. (example: https://github.com/haskell/haskell-language-server/blob/6fa3e640b6ffac4ee9ffbd051b8d01fd220a7378/plugins/hls-pragmas-plugin/test/Main.hs#L198 ) There are no tests that i can see in `ghcide` that need to be moved. I have edited the tests in the plugin to reflect the new functionality of closing braces

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -628,8 +616,6 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
     -> return []
     | "{-# options_ghc" `T.isPrefixOf` T.toLower fullLine
     -> return $ filtOptsCompls (map (T.pack . stripLeading '-') $ flagsForCompletion False)
```
pepeiborra, line 618: shouldn't this case be moved as well?

alexnaspo: yes, I will work towards that

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
```diff
@@ -163,9 +166,19 @@ completion _ide _ complParams = do
                     | "{-# LANGUAGE" `T.isPrefixOf` VFS.fullLine pfix
                     = J.List $ map buildCompletion
                         (Fuzzy.simpleFilter (VFS.prefixText pfix) allPragmas)
+                    | "{-# options_ghc" `T.isPrefixOf` T.toLower (VFS.fullLine pfix)
+                    = J.List $ map mkExtCompl
+                        (Fuzzy.simpleFilter (VFS.prefixText pfix) flags)
```
pepeiborra, line 171: do we want to add the closing bracket (when needed) here as well?

alexnaspo: actually no, this is the case that fills in the values of the options, not the name. The completion for the string "{-# options_ghc ... #-} is handled in the other case and handles closing brackets.

pepeiborra (approved):

PR #2136, opened 2021-08-29
attempt to fix Nix build

peterbecich:
`nix-build` succeeds on this branch

SHA acquired with `nix-prefetch-url https://hackage.haskell.org/package/lsp-types-1.3.0.1/lsp-types-1.3.0.1.tar.gz --unpack`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -87,8 +87,8 @@
 
               lsp-types = hself.callCabal2nix "lsp-types"
                 (builtins.fetchTarball {
-                  url = "https://hackage.haskell.org/package/lsp-types-1.3.0.0/lsp-types-1.3.0.0.tar.gz";
-                  sha256 = "0qajyyj2d51daa4y0pqaa87n4nny0i920ivvzfnrk9gq9386iac7";
+                  url = "https://hackage.haskell.org/package/lsp-types-1.3.0.1/lsp-types-1.3.0.1.tar.gz";
+                  sha256 = "19k28zf1vw60wqfxllcs7zk9j6lnkx5kkvjnh22vkvn6m9zzflyw";
                 }) { };
 
               lsp-test = hself.callCabal2nix "lsp-test"


pepeiborra:
The Nix CI job is failing in a confusing way - it's trying to build the development shell with multiple versions of GHC in the same job. This should probably be done using a matrix instead. @berberman do you remember if this was a deliberate choice?

The failure is not related to this PR, but it's preventing us from finding out whether the PR works or not

berberman:
Sorry for being inactive these days, I was pretty busy.

What `Nix / build` jobs do is push the development shell and "nightly" HLS binary to cachix, and they are likely to be broken due to packages updates e.g. bumping the version of lsp.  The support of GHC 9.0.1 was introduced in #1995, which is very ad-hoc, and a matrix CI for `Nix / build` jobs would be a better choice.

Now I don't have nix environment around, but I guess to fix the jobs, at least we shoud:
* bump nixpkgs
* change https://github.com/haskell/haskell-language-server/blob/a800b9d71e4eac70d672d30b2488949ea26dcbc3/flake.nix#L82-L98

to something like

```nix
              lsp = hself.lsp_1_2_0_1;

              lsp-types = hself.lsp-types_1_3_0_1;

              lsp-test = hself.lsp-test_0_14_0_1;
```

because these packages we override the sources of are already in nixpkgs

* add `ormolu = hself.ormolu_0_2_0_0;` to https://github.com/haskell/haskell-language-server/blob/a800b9d71e4eac70d672d30b2488949ea26dcbc3/configuration-ghc-901.nix#L15

because we need ormolu 0.2.0.0 in GHC 9.0.1 (#2086)

jneira:
@peterbecich hi! thanks for trying to fix the nix build, have you got the chance to check the suggestion above?

lf-:
So I was poking at doing basically the same thing as this PR requests without noticing it already existed and did all the requested things from here. See #2159.

peterbecich:
Thank you @pepeiborra , @berberman , @jneira and @lf- ; sorry for my late response. Closing in favor of https://github.com/haskell/haskell-language-server/pull/2159
I have not tried the suggestion above, yet. I will do that now so that I can learn something.

Issue #2132, opened 2021-08-27
Add cli option to list what plugins are built in hls executable


jneira:
* Proposed in the matrix haskell channel: https://matrix.to/#/!hskonBonfjiIefqLUV:matrix.org/$r6XqaAxSOTy6_ezML5pr6qwQ0ZbFzJD-vRXi_HfrWyI?via=hpdeifel.de&via=matrix.org&via=rgb.us.eu.org by det0x
* With the motivation of having a explicit way to know what plugins are available in a concrete hls executable (f.e. it would be useful to see what plugins ar included in hls-1.3.0 for ghc-9.0.1)
* We are already listing them when starting the lsp server so it will not be too difficult to add it

jneira:
Closed by #2139, many thanks @July541

Issue #2115, opened 2021-08-20
Flaky GHCIDE test: `Timestamps do not have millisecond` (ignored for now)


isovector:
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:
CI 

Which OS do you use:
Ubuntu latest

### Steps to reproduce
CI is flakey. It reproduces itself.

### Expected behaviour
Tests should reliably pass.

### Actual behaviour

https://github.com/haskell/haskell-language-server/runs/3373456727

```
ghcide
  Unit
    timestamps have millisecond resolution: FAIL (0.02s)
      test/exe/Main.hs:5677:
      Timestamps do not have millisecond resolution: 10000us
```

This happens roughly 10% of the time

anka-213:
Maybe is just luck that makes it work as much as 90% of the time? Would increasing the tolerance to not require millisecond resolution be a solution?

We could modify `findResolution_us` to return the actual difference in timestamp (`t' - t`) in order to help diagnose the problem.

jneira:
https://github.com/haskell/haskell-language-server/runs/3504012776?check_suite_focus=true testing #2154

```
Test suite ghcide-tests: RUNNING...
ghcide
  Unit
    timestamps have millisecond resolution: FAIL (0.03s)
      test/exe/Main.hs:5717:
      Timestamps do not have millisecond resolution: 10000us
```

:worried: reopening just in case

isovector:
I just hit this too. Definitely still happening.

jneira:
Sorry but i am ignoring it with https://github.com/haskell/haskell-language-server/pull/2182/commits/692a26e2b358adf212f346b26eb9299103309593 to not block ci

isovector:
Thanks @jneira. I was just about to send a PR doing that

PR #2139, opened 2021-08-29
List all available plugins

July541:
This pr add an option to list all available plugins as #2132 described.

It looks like 
```
$ /home/treee/Code/haskell-language-server/dist-newstyle/build/x86_64-linux/ghc-8.10.4/haskell-language-server-1.3.0.0/x/haskell-language-server/build/haskell-language-server/haskell-language-server --list-plugins
brittany
callHierarchy
class
eval
floskell
fourmolu
ghcide-code-actions-bindings
ghcide-code-actions-fill-holes
ghcide-code-actions-imports-exports
ghcide-code-actions-type-signatures
ghcide-completions
ghcide-core
ghcide-hover-and-symbols
ghcide-type-lenses
haddockComments
hlint
importLens
moduleName
ormolu
pragmas
refineImports
retrie
splice
stylish-haskell
tactics
```

I think that simultaneously to list if this plugin is enabled default is better, but I don't find anything to capture that:(

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@
 [![Hackage][badge-hackage]][hackage]
 [![License Apache 2.0][badge-license]][license]
 [![CircleCI][badge-circleci]][circleci]
-![Github Testing Workflow](https://github.com/haskell/haskell-language-server/workflows/Testing/badge.svg)
-![Github Nix Workflow](https://github.com/haskell/haskell-language-server/workflows/Nix/badge.svg)
+[![GitHub Testing Workflow](https://github.com/haskell/haskell-language-server/actions/workflows/test.yml/badge.svg)](https://github.com/haskell/haskell-language-server/actions/workflows/test.yml)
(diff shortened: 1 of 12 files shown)


jneira (approved):
Wow many thanks, it looks great as is

> I think that simultaneously to list if this plugin is enabled default is better, but I don't find anything to capture that:(

Well i think for now all plugins are enabled in the config by default so it is not urgent

July541:
Test works well on my branch https://github.com/July541/haskell-language-server/pull/2, not sure why this was broken.

July541:
Looks like https://github.com/haskell/haskell-language-server/issues/2115

jneira:
> 
> 
> Looks like #2115

rerunning jobs

jneira:
> 
> 
> Wow many thanks, it looks great as is
> 
> > I think that simultaneously to list if this plugin is enabled default is better, but I don't find anything to capture that:(
> 
> Well i think for now all plugins are enabled in the config by default so it is not urgent

In any case you can get the default configuration, where you could find if a included plugin is enabled by default or does not

PR #2142, opened 2021-08-29
case sensitive language pragmas fix

alexnaspo:
Pragmas are not case sensitive - so the language pragma should auto complete with values accordingly


jneira:
Hi, many thanks (again 🙂 ) for the pr. To add a little bit of context:
* There was a pr that tries to accomplish this: #1609. It was stalled and needed tests and your already has it so dont worry

jneira (approved):
LGTM

PR #2146, opened 2021-08-30
Filter code actions based on prefix, not equality

michaelpj:
It's quite unclear in the spec, but in
https://github.com/microsoft/language-server-protocol/issues/970
it's suggested that the intention is that the kinds given in `only`
should be used as *prefix* filters of the generated code action kinds.

That is to say, if the client asks for `only = [ CodeActionRefactor ]`,
we should give them all kinds of refactoring code actions, including
those whose kind is `CodeActionRefactorInline` (because as "hierarchical
strings" they are represented as `"refactor"` and `"refactor.inline"`).

This is quite important for the client: e.g. I hit this because I wanted
to ask for all the import quickfixes so I could present them to the user
to pick one. But they use various subkinds of `"quickfix.import"`, so
currently you cannot ask for them all (asking for `"quickfix.import"`
currentl returns nothing!).

The ipmlemention is a little ugly: this needs some helper funcitons in
`lsp`, which I'll make a PR for separately, but I didn't want to block
this.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2146"></a>

### DIFF
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -192,7 +192,16 @@ instance PluginMethod TextDocumentCodeAction where
       wasRequested (InR ca)
         | Nothing <- _only context = True
         | Just (List allowed) <- _only context
-        , Just caKind <- ca ^. kind = caKind `elem` allowed
+        -- See https://github.com/microsoft/language-server-protocol/issues/970
+        -- This is somewhat vague, but due to the hierarchical nature of action kinds, we
+        -- should check whether the requested kind is a *prefix* of the action kind.
+        -- That means, for example, we will return actions with kinds `quickfix.import` and
+        -- `quickfix.somethingElse` if the requested kind is `quickfix`.
+        -- TODO: add helpers in `lsp` for handling code action hierarchies
+        -- For now we abuse the fact that the JSON representation gives us the hierarchical string.
+        , Just caKind <- ca ^. kind
+        , String caKindStr <- toJSON caKind =
+                any (\k -> k `T.isPrefixOf` caKindStr) [kstr | k <- allowed, let String kstr = toJSON k ]
         | otherwise = False
 
 instance PluginMethod TextDocumentCodeLens where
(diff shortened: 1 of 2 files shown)


michaelpj:
(My goal here is to be able to make a nice emacs function that pulls out all the import-fixing actions and gets the user to pick one, which requires both this and https://github.com/emacs-lsp/lsp-mode/pull/3071)

jneira:
build error:

```
 [4 of 6] Compiling Ide.Types        ( src/Ide/Types.hs, /home/runner/work/haskell-language-server/haskell-language-server/dist-newstyle/build/x86_64-linux/ghc-8.10.6/hls-plugin-api-1.2.0.1/build/Ide/Types.o, /home/runner/work/haskell-language-server/haskell-language-server/dist-newstyle/build/x86_64-linux/ghc-8.10.6/hls-plugin-api-1.2.0.1/build/Ide/Types.dyn_o )

src/Ide/Types.hs:75:12: error:
Error:     Illegal deriving strategy: newtype
    Use DerivingStrategies to enable this extension
   |
75 |   deriving newtype (Monoid, Semigroup)
   |            ^^^^^^^
cabal: Failed to build hls-plugin-api-1.2.0.1 (which is required by
test:wrapper-test from hls-1.3.0.0 and test:func-test from hls-1.3.0.0).
```

pepeiborra (approved):
This looks reasonable, it would be great if someone could try it with VSCode and verify that it works as expected

jneira:
> Seems unrelated? Hopefully merging master will fix it.

Are you sure? the pr is touching the file which does not compile, maybe the error comes from imports

```
 src/Ide/Types.hs:75:12: error:
Error:     Illegal deriving strategy: newtype
    Use DerivingStrategies to enable this extension
   |
75 |   deriving newtype (Monoid, Semigroup)
   |            ^^^^^^^
```

michaelpj:
> Are you sure? the pr is touching the file which does not compile, maybe the error comes from imports

Huh. For some reason the pre-commit hook strips a couple of language pragmas which are actually needed. Weird. Fixed.

> This looks reasonable, it would be great if someone could try it with VSCode and verify that it works as expected

No change in VSCode as far as I can see. It already seems to list all the quickfixes when you hit `Ctrl+.`, I think possibly it doesn't use the server-side filtering. I can't tell because I can't figure out how to turn on the actual message logging any more... anyway, it seems to still work fine.

(Rebased on master)

jneira (approved):
thanks!

Issue #2082, opened 2021-08-06
Autocompletion no longer counts qualified imports


Akhristenko:
### Steps to reproduce
I have 2 modules
```haskell
module MyModule (myid) where

myid :: a -> a
myid a = a
```
```haskell
module MyModuleB (myid') where

import qualified MyModule

myid' :: a -> a
myid' = MyModule.myid
```
I delete `myid` in second module and invoke autocompletion when cursor placed after `MyModule.`

### Expected behaviour
In autocompletion list i see only functions from MyModule

### Actual behaviour
In autocompletion list i see all function from all modules

pepeiborra:
Thanks for reporting, this is likely a regression after #2040

pepeiborra:
I'm not so sure this is a regression anymore. It's actually a new feature! Previously autocompletion only suggested imported identifiers, now it suggests also unimported identifiers and automatically inserts a new import if necessary. Qualifiers are no longer used only for filtering, but also to inform the format of the inserted import. However qualifiers are still used for filtering, and list of completions should put the functions from current imports first. Can you confirm whether this is the case for you?

Akhristenko:
> However qualifiers are still used for filtering, and list of completions should put the functions from current imports first.

It's not true, at least on version 1.3.0 it show all names in alphabetical order. But if hls will showing names imported in qualifiers ordered alphabetically before all other names it will solve this issue.

Akhristenko:
ideally i would like this behavior:
- if we invoke autocomplete on qualifier, which was not imported, we see all functions and autoimport perform after we select one of them.
- if we invoke autocomplete on qualifier, which was imported, we see only functions which was imported. But if we invoke autocomplete second time we see all functions and can autoimport one of them.

alexnaspo:
I would be interested in attempting this. Is it currently possible to get a list of all of the possible exported functions in a plugin for a given module?

alexnaspo:
I see a decent path forward on this one, is it worth picking up? @pepeiborra

pepeiborra:
I would start with a set of tests demonstrating the problem, and then we can discuss what needs to be done to enforce the behaviour that I outlined above.

Implementing the behaviour that @Akhristenko suggested is not something I would recommend. "If we invoke autocomplete a second time" is not a condition that the LSP server can detect.

Issue #797, opened 2021-01-01
Completions inside explicit imports should suggest module identifiers


ndmitchell:
When typing:

```
import Foo(xyz)
```

When at the end of `xyz` but before the bracket, it would be great to complete to identifiers that are in `Foo`. But instead I get completions of module names.

Issue #314, opened 2020-08-12
[Feature Request] Auto complete in imports


GuillaumeDesforges:
Feature request

Say I start writing:

```haskell
import Data.Map (
```

I'd like to have autocomplete that proposes what I can import from this module, with its doc/type signature.

This is a common feature in other module-based programming languages IDE/LSP.

jneira:
I would say this should be added in ghcide

jneira:
Afaiu this was fixed by #2152 (although it stil is  not released)

PR #2152, opened 2021-09-02
auto complete functions from imports

alexnaspo:
AutoComplete functions inside a module import

closes https://github.com/haskell/haskell-language-server/issues/2082

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -16,6 +16,7 @@ import           Data.Aeson
 import qualified Data.HashMap.Strict                          as Map
 import qualified Data.HashSet                                 as Set
 import           Data.List                                    (find)
+import qualified Data.Map                                     as DM (Map, fromListWith, empty)
 import           Data.Maybe
 import qualified Data.Text                                    as T
 import           Development.IDE.Core.PositionMapping
@@ -131,7 +132,7 @@ getCompletionsLSP ide plId
     fmap Right $ case (contents, uriToFilePath' uri) of
       (Just cnts, Just path) -> do
         let npath = toNormalizedFilePath' path
-        (ideOpts, compls) <- liftIO $ runIdeAction "Completion" (shakeExtras ide) $ do
+        (ideOpts, compls, moduleExports) <- liftIO $ runIdeAction "Completion" (shakeExtras ide) $ do
             opts <- liftIO $ getIdeOptionsIO $ shakeExtras ide
             localCompls <- useWithStaleFast LocalCompletions npath
             nonLocalCompls <- useWithStaleFast NonLocalCompletions npath
@@ -139,12 +140,13 @@ getCompletionsLSP ide plId
             binds <- fromMaybe (mempty, zeroMapping) <$> useWithStaleFast GetBindings npath
             exportsMapIO <- fmap(envPackageExports . fst) <$> useWithStaleFast GhcSession npath
             exportsMap <- mapM liftIO exportsMapIO
+            let moduleExports = buildModouleExportMap exportsMap
             let exportsCompItems = foldMap (map (fromIdentInfo uri) . Set.toList) . Map.elems . getExportsMap <$> exportsMap
                 exportsCompls = mempty{anyQualCompls = fromMaybe [] exportsCompItems}
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls
-            pure (opts, fmap (,pm,binds) compls)
-        case compls of
-          Just (cci', parsedMod, bindMap) -> do
+            pure (opts, fmap (,pm,binds) compls, moduleExports)
+        case (compls, moduleExports) of
+          (Just (cci', parsedMod, bindMap), mExports) -> do
             pfix <- VFS.getCompletionPrefix position cnts
             case (pfix, completionContext) of
               (Just (VFS.PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
@@ -152,13 +154,26 @@ getCompletionsLSP ide plId
               (Just pfix', _) -> do
                 let clientCaps = clientCapabilities $ shakeExtras ide
                 config <- getCompletionsConfig plId
-                allCompletions <- liftIO $ getCompletions plId ideOpts cci' parsedMod bindMap pfix' clientCaps config
+                allCompletions <- liftIO $ getCompletions plId ideOpts cci' parsedMod bindMap pfix' clientCaps config mExports
                 pure $ InL (List allCompletions)
               _ -> return (InL $ List [])
           _ -> return (InL $ List [])
       _ -> return (InL $ List [])
 
 ----------------------------------------------------------------------------------------------------
+
+identInfoToKeyVal :: IdentInfo -> (T.Text, T.Text)
+identInfoToKeyVal IdentInfo {rendered, moduleNameText} =
+  (moduleNameText, rendered)
+
+buildModouleExportMap:: Maybe (ExportsMap) -> DM.Map T.Text [T.Text]
+buildModouleExportMap (Just exportsMap) = do
+  sortAndGroup $ map identInfoToKeyVal $ 
+    concatMap (Set.toList . snd) $ toList $ getExportsMap exportsMap
+buildModouleExportMap (Nothing) = DM.empty 
+
+sortAndGroup :: [(T.Text, T.Text)] -> DM.Map T.Text [T.Text]
+sortAndGroup assocs = DM.fromListWith (++) [(k, [v]) | (k, v) <- assocs]  
 
 extendImportCommand :: PluginCommand IdeState
 extendImportCommand =
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -41,6 +41,7 @@ import           Control.Monad
 import           Data.Aeson                               (ToJSON (toJSON))
 import           Data.Either                              (fromRight)
 import           Data.Functor
+import qualified Data.Map                                 as DM (Map)
 import qualified Data.Set                                 as Set
 import           Development.IDE.Core.Compile
 import           Development.IDE.Core.PositionMapping
@@ -54,7 +55,7 @@ import           Development.IDE.Spans.LocalBindings
 import           Development.IDE.Types.Exports
 import           Development.IDE.Types.HscEnvEq
 import           Development.IDE.Types.Options
-import           GhcPlugins                               (flLabel, unpackFS)
+import           GhcPlugins                               (flLabel, unpackFS, lookupWithDefaultUFM)
 import           Ide.PluginUtils                          (mkLspCommand)
 import           Ide.Types                                (CommandId (..),
                                                            PluginId)
@@ -63,6 +64,7 @@ import           Language.LSP.Types.Capabilities
 import qualified Language.LSP.VFS                         as VFS
 import           Outputable                               (Outputable)
 import           TyCoRep
+
 
 -- From haskell-ide-engine/hie-plugin-api/Haskell/Ide/Engine/Context.hs
 
@@ -282,6 +284,13 @@ mkNameCompItem doc thingParent origName origMod thingType isInfix docs !imp = CI
 mkModCompl :: T.Text -> CompletionItem
 mkModCompl label =
   CompletionItem label (Just CiModule) Nothing Nothing
+    Nothing Nothing Nothing Nothing Nothing Nothing Nothing
+    Nothing Nothing Nothing Nothing Nothing Nothing
+
+
+mkModuleFunctionImport :: T.Text -> T.Text -> CompletionItem
+mkModuleFunctionImport moduleName label =
+  CompletionItem label (Just CiFunction) Nothing (Just moduleName)
     Nothing Nothing Nothing Nothing Nothing Nothing Nothing
     Nothing Nothing Nothing Nothing Nothing Nothing
 
@@ -530,9 +539,10 @@ getCompletions
     -> VFS.PosPrefixInfo
     -> ClientCapabilities
     -> CompletionsConfig
+    -> DM.Map T.Text [T.Text]
     -> IO [CompletionItem]
 getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qualCompls, importableModules}
-               maybe_parsed (localBindings, bmapping) prefixInfo caps config = do
+               maybe_parsed (localBindings, bmapping) prefixInfo caps config exportsMap = do
   let VFS.PosPrefixInfo { fullLine, prefixModule, prefixText } = prefixInfo
       enteredQual = if T.null prefixModule then "" else prefixModule <> "."
       fullPrefix  = enteredQual <> prefixText
@@ -619,9 +629,17 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         | s == c = ss
         | otherwise = s:ss
 
-  if
+  if 
+    | "import " `T.isPrefixOf` fullLine 
+      && (List.length (words (T.unpack fullLine)) >= 2)
+      && "(" `isInfixOf` T.unpack fullLine
+    -> do  
+      let moduleName = (words (T.unpack fullLine)) !! 1
+          funcs = Map.findWithDefault [] (T.pack moduleName) exportsMap
+      return (map (mkModuleFunctionImport (T.pack moduleName)) funcs)
     | "import " `T.isPrefixOf` fullLine
-    -> return filtImportCompls
+    -> do
+      return filtImportCompls
     -- we leave this condition here to avoid duplications and return empty list
     -- since HLS implements this completion (#haskell-language-server/pull/662)
     | "{-# language" `T.isPrefixOf` T.toLower fullLine
@@ -651,6 +669,7 @@ uniqueCompl x y =
         then EQ
         else compare (insertText x) (insertText y)
     other -> other
+
 -- ---------------------------------------------------------------------
 -- helper functions for pragmas
 -- ---------------------------------------------------------------------
--- a/test/functional/Completion.hs
+++ b/test/functional/Completion.hs
@@ -121,6 +121,31 @@ tests = testGroup "completions" [
          compls <- getCompletions doc (Position 5 7)
          liftIO $ length compls @?= maxCompletions def
 
+     , testCase "import function completions" $ runSession hlsCommand fullCaps "test/testdata/completion" $ do
+         doc <- openDoc "FunctionCompletions.hs" "haskell"
+
+         let te = TextEdit (Range (Position 0 30) (Position 0 41)) "A"
+         _ <- applyEdit doc te
+
+         compls <- getCompletions doc (Position 0 31)
+         let item = head $ filter ((== "Alternative") . (^. label)) compls
+         liftIO $ do
+             item ^. label @?= "Alternative"
+             item ^. kind @?= Just CiFunction
+             item ^. detail @?= Just "Control.Applicative"
+
+    , testCase "import second function completion" $ runSession hlsCommand fullCaps "test/testdata/completion" $ do
+         doc <- openDoc "FunctionCompletions.hs" "haskell"
+
+         let te = TextEdit (Range (Position 0 41) (Position 0 42)) ", l"
+         _ <- applyEdit doc te
+
+         compls <- getCompletions doc (Position 0 41)
+         let item = head $ filter ((== "liftA") . (^. label)) compls
+         liftIO $ do
+             item ^. label @?= "liftA"
+             item ^. kind @?= Just CiFunction
+             item ^. detail @?= Just "Control.Applicative"
      , contextTests
      , snippetTests
     ]
--- a/test/testdata/completion/FunctionCompletions.hs
+++ b/test/testdata/completion/FunctionCompletions.hs
@@ -1,0 +1,8 @@
+import Control.Applicative (Alternative)
+import qualified Data.List
+
+main :: IO ()
+main = putStrLn "hello"
+
+foo :: Either a b -> Either a b
+foo = id


jneira:
I am not sure if this fixes #2082. Afaics tests are demonstrating that completions works inside imports lists but the issue is about completions over qualified imports in the body of the module:

> I have 2 modules
> 
> module MyModule (myid) where
> 
> myid :: a -> a
> myid a = a
> 
> module MyModuleB (myid') where
> 
> import qualified MyModule
> 
> myid' :: a -> a
> myid' = MyModule.myid
> 
> I delete myid in second module and invoke autocompletion when cursor placed after MyModule.
> Expected behaviour
> 
> In autocompletion list i see only functions from MyModule
> Actual behaviour
> 
> In autocompletion list i see all function from all modules

alexnaspo:
Sorry, I misunderstood the ticket. Will remove from the description.

pepeiborra:
> I am not sure if this fixes #2082. Afaics tests are demonstrating that completions works inside imports lists but the issue is about completions over qualified imports in the body of the module:
> 
> 
> 
> > I have 2 modules
> 
> > 
> 
> > module MyModule (myid) where
> 
> > 
> 
> > myid :: a -> a
> 
> > myid a = a
> 
> > 
> 
> > module MyModuleB (myid') where
> 
> > 
> 
> > import qualified MyModule
> 
> > 
> 
> > myid' :: a -> a
> 
> > myid' = MyModule.myid
> 
> > 
> 
> > I delete myid in second module and invoke autocompletion when cursor placed after MyModule.
> 
> > Expected behaviour
> 
> > 
> 
> > In autocompletion list i see only functions from MyModule
> 
> > Actual behaviour
> 
> > 
> 
> > In autocompletion list i see all function from all modules

To clarify the expected behaviour is functions from MyModule followed by functions from all modules.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -131,35 +132,49 @@ getCompletionsLSP ide plId
...
+            let moduleExports = buildModouleExportMap exportsMap
             let exportsCompItems = foldMap (map (fromIdentInfo uri) . Set.toList) . Map.elems . getExportsMap <$> exportsMap
                 exportsCompls = mempty{anyQualCompls = fromMaybe [] exportsCompItems}
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls
-            pure (opts, fmap (,pm,binds) compls)
-        case compls of
-          Just (cci', parsedMod, bindMap) -> do
+            pure (opts, fmap (,pm,binds) compls, moduleExports)
+        case (compls, moduleExports) of
+          (Just (cci', parsedMod, bindMap), mExports) -> do
```
pepeiborra, line 149: You can revert these two lines, there is no need to pattern match on moduleExports

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -16,6 +16,7 @@ import           Data.Aeson
 import qualified Data.HashMap.Strict                          as Map
 import qualified Data.HashSet                                 as Set
 import           Data.List                                    (find)
+import qualified Data.Map                                     as DM (Map, fromListWith, empty)
```
pepeiborra, line 19: Please use Data.HashMap instead of Data.Map unless there is a very good reason to do so

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -131,35 +132,49 @@ getCompletionsLSP ide plId
...
       _ -> return (InL $ List [])
 
 ----------------------------------------------------------------------------------------------------
 
+identInfoToKeyVal :: IdentInfo -> (T.Text, T.Text)
+identInfoToKeyVal IdentInfo {rendered, moduleNameText} =
+  (moduleNameText, rendered)
+
+buildModouleExportMap:: Maybe (ExportsMap) -> DM.Map T.Text [T.Text]
+buildModouleExportMap (Just exportsMap) = do
```
pepeiborra, line 170: What does this function do?

jneira:
Afaics this fixes #797 (cc @ndmitchell)

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: I had a typo in the name, however, it builds a map of `"module.name" -> ["functions" .. ]` which I use to look up the available functions for that module

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: addressed

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: addressed

jneira:
stack builds are failing:

```
ghcide                     > /root/build/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs:636:19: error:
ghcide                     >     Not in scope: ‘HM.findWithDefault’
ghcide                     >     Perhaps you meant ‘Map.findWithDefault’ (imported from Data.Map)
ghcide                     >     Module ‘Data.HashMap.Strict’ does not export ‘findWithDefault’.
ghcide                     >     |
ghcide                     > 636 |           funcs = HM.findWithDefault [] (T.pack moduleName) exportsMap
ghcide                     >     |                   ^^^^^^^^^^^^^^^^^^
ghcide                     > 
```

ghcide/src/Development/IDE/Plugin/Completions.hs
pepeiborra: And why is it needed? Can't you use the ExportsMap as it is?

alexnaspo: I believe I need it because I have to group on `"module.name"` and use that as the key to look up. `ExportsMap` as is seems to be a map of `"function-name" -> [{moduleNameText: "module-name",  name: "function-name"}]`

pepeiborra: Ah yes, you are right. 

But why build a map to do only one lookup and then throw it away? 

* Cost of building the map: O(NlogN)
* Cost of doing the lookup by scanning the original exports map: O(N)

Neither is good enough, you want O(logN). How? Probably by changing the representation of the exports map to add a second index, the module name.

alexnaspoleap:
> stack builds are failing:
> 
> ```
> ghcide                     > /root/build/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs:636:19: error:
> ghcide                     >     Not in scope: ‘HM.findWithDefault’
> ghcide                     >     Perhaps you meant ‘Map.findWithDefault’ (imported from Data.Map)
> ghcide                     >     Module ‘Data.HashMap.Strict’ does not export ‘findWithDefault’.
> ghcide                     >     |
> ghcide                     > 636 |           funcs = HM.findWithDefault [] (T.pack moduleName) exportsMap
> ghcide                     >     |                   ^^^^^^^^^^^^^^^^^^
> ghcide                     > 
> ```

thank you - this is fixed

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspoleap: I don't think I follow your suggestion. Do you mean to add the moduleName as keys, or another layer of indexes?

pepeiborra: Add another field to `ExportsMap` containing a map from modules to exports. 

But a better solution would be to avoid using the exports map at all. Instead, retrieve the `ModIface` for the module and extract the list of exports. To do that, you first need to map the module name to a path using `GetLocatedImports`. At this point, I would consider moving the code to a new plugin.

alexnaspoleap: Do you think the first suggestion would be a good first pass? With potential follow up commits to pull out a plugin?

pepeiborra: The ExportsMap only indexes package modules. You will need to fetch the ModIface for local modules. So both changes are needed

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -620,6 +628,13 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         | otherwise = s:ss
 
   if
+    | "import " `T.isPrefixOf` fullLine 
+      && (List.length (words (T.unpack fullLine)) >= 2)
+      && "(" `isInfixOf` T.unpack fullLine
+    -> do  
+      let moduleName = words (T.unpack fullLine) !! 1
+          funcs = HM.findWithDefault [] (T.pack moduleName) exportsMap
+      return (map (mkModuleFunctionImport (T.pack moduleName)) funcs)
```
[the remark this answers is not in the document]

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: I took a first pass at this. There is still more work to do. As of this comment, I am just generating the map based on the already existing map. I also see that this does not work for local modules, as you suggested.

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -30,15 +30,20 @@ import           HieDb
 import           Name
 import           TcRnTypes                   (TcGblEnv (..))
 
-newtype ExportsMap = ExportsMap
-    {getExportsMap :: HashMap IdentifierText (HashSet IdentInfo)}
-    deriving newtype (Monoid, NFData, Show)
+data ExportsMap = ExportsMap
+    {getExportsMap :: HashMap IdentifierText (HashSet IdentInfo)
+    , getModuleExportsMap :: Map.HashMap IdentifierText [Text]
```
pepeiborra, line 35:
```suggestion
    , getModuleExportsMap :: Map.HashMap ModuleNameText [Text]
```

pepeiborra: I think the keys are module names right? Better change the type synonym to reflect that and even better, add Haddock comments.

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: Second pass - generate the map from the modIface rather than the existing map.. This still does not account for local modules

ghcide/src/Development/IDE/Types/Exports.hs
pepeiborra: Can we have a `HashSet IdentInfo` as the value type, for consistency with `getExportsMap`?

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -90,26 +97,45 @@ mkIdentInfos mod (AvailTC _ nn flds)
...
-        mn = moduleName $ mi_module mi
+    doOne mi = do
+      let getModDetails = unpackAvail $ moduleName $ mi_module mi
+      concatMap (fmap (second Set.fromList) . getModDetails) (mi_exports mi)
 
 createExportsMapMg :: [ModGuts] -> ExportsMap
-createExportsMapMg = ExportsMap . Map.fromListWith (<>) . concatMap doOne
+createExportsMapMg modIface = do
+  let exportsMap = (Map.fromListWith (<>) . concatMap doOne) modIface
+  ExportsMap exportsMap Map.empty
```
alexnaspo, line 125: Is it ok to have an empty map for the create methods that do not take modIface? @pepeiborra

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: (sorry for all the of the noise) - third pass - enable local module imports

ghcide/src/Development/IDE/Types/Exports.hs
alexnaspo: resolved

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -90,26 +97,44 @@ mkIdentInfos mod (AvailTC _ nn flds)
...
-      where
-        mn = moduleName $ mi_module mi
+    doOne mi = do
+      let getModDetails = unpackAvail $ moduleName $ mi_module mi
+      concatMap (fmap (second Set.fromList) . getModDetails) (mi_exports mi)
 
 createExportsMapMg :: [ModGuts] -> ExportsMap
-createExportsMapMg = ExportsMap . Map.fromListWith (<>) . concatMap doOne
+createExportsMapMg modIface = do
+  let exportsMap = (Map.fromListWith (<>) . concatMap doOne) modIface
```
pepeiborra, line 123:
```suggestion
createExportsMapMg modGuts = do
  let exportsMap = (Map.fromListWith (<>) . concatMap doOne) modGuts
```

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -90,26 +97,44 @@ mkIdentInfos mod (AvailTC _ nn flds)
...
+      moduleExportMap = buildModuleExportMap modIface
+  ExportsMap exportsMap moduleExportMap
   where
-    doOne mi = concatMap (fmap (second Set.fromList) . unpackAvail mn) (mi_exports mi)
-      where
-        mn = moduleName $ mi_module mi
+    doOne mi = do
+      let getModDetails = unpackAvail $ moduleName $ mi_module mi
+      concatMap (fmap (second Set.fromList) . getModDetails) (mi_exports mi)
```
pepeiborra, line 119: Can you find a way to rewrite this code so that the `IdentInfo` values in both hashmaps? The duplication is expensive both in time and space.

This comment applies to all the `createExportsMap` functions in this module.

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -121,7 +146,8 @@ createExportsMapHieDb hiedb = do
         let mn = modInfoName $ hieModInfo m
             mText = pack $ moduleNameString mn
         fmap (wrap . unwrap mText) <$> getExportsForModule hiedb mn
-    return $ ExportsMap $ Map.fromListWith (<>) (concat idents)
+    let exportsMap = Map.fromListWith (<>) (concat idents)
+    return $ ExportsMap exportsMap Map.empty
```
pepeiborra, line 150: Why is the module exports map empty in this case? It shouldn't be, we've got all the information needed to construct one

pepeiborra (changes requested):

ghcide/src/Development/IDE/Types/Exports.hs
pepeiborra: I meant to say "the values are reused in both hash maps". My keyboard ate a word 😅

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: OK, I now generate the map for local modules as well. I will take a look at your other comments.

ghcide/src/Development/IDE/Types/Exports.hs
alexnaspo: One simple way to do this is to generate it from the already existing map.. did you mean a different approach? This would also make the below comment simple

pepeiborra: Generating from the existing map is fine. Since exports are already organized by module, it might be more effective to generate the module exports map first, and derive the other one from it.

ghcide/src/Development/IDE/Types/Exports.hs
alexnaspo: addressed

ghcide/src/Development/IDE/Types/Exports.hs
alexnaspo: resolved - I now generate both from a list

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -152,13 +156,28 @@ getCompletionsLSP ide plId
...
+
+getModIface :: IdeState ->  (Located ModuleName, Maybe ArtifactsLocation) -> IO (Maybe ModIface)
+getModIface ide item = do
+  let nPath = artifactFilePath <$> snd item
+  case nPath of 
+    (Just path) -> do 
+      file <- runIdeAction "Completion" (shakeExtras ide) $ do
+        fmap fst <$> useWithStaleFast GetModIface path
+      pure $ fmap hirModIface file
+    Nothing -> pure Nothing
```
pepeiborra, line 180: For performance reasons, can you restructure this code so that it performs only one `runIdeAction` and one call to `usesWithStaleFast`?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -620,6 +628,13 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         | otherwise = s:ss
 
   if
+    | "import " `T.isPrefixOf` fullLine 
+      && (List.length (words (T.unpack fullLine)) >= 2)
+      && "(" `isInfixOf` T.unpack fullLine
```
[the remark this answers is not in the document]
pepeiborra: Can you add a TODO here?

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -91,25 +99,32 @@ mkIdentInfos mod (AvailTC _ nn flds)
...
+createExportsMap modIface = do
+  let exportList = concatMap doOne modIface
+  ExportsMap (Map.fromListWith (<>) exportList) (buildModuleExportMap exportList)
   where
-    doOne mi = concatMap (fmap (second Set.fromList) . unpackAvail mn) (mi_exports mi)
-      where
-        mn = moduleName $ mi_module mi
+    doOne modIFace = do
+      let getModDetails = unpackAvail $ moduleName $ mi_module modIFace
+      concatMap (fmap (second Set.fromList) . getModDetails) (mi_exports modIFace)
```
pepeiborra, line 108: The efficient thing to do here is to change `unpackAvail` and `doOne` to return a triple `(moduleName, identifierText, IdentInfo)`

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -121,7 +136,8 @@ createExportsMapHieDb hiedb = do
         let mn = modInfoName $ hieModInfo m
             mText = pack $ moduleNameString mn
         fmap (wrap . unwrap mText) <$> getExportsForModule hiedb mn
-    return $ ExportsMap $ Map.fromListWith (<>) (concat idents)
+    let exportsMap = Map.fromListWith (<>) (concat idents)
+    return $ ExportsMap exportsMap $ buildModuleExportMap (concat idents)
```
pepeiborra, line 140: Similarly, this code could easily return a triple `(moduleName, identifierText, IdentInfo)` instead of deriving the module export map from the identifiers one

pepeiborra (changes requested):
Almost there! Just some minor performance improvements requested

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: Thank you for all of your help on this PR! Your pointers have been extremely helpful.

I have a question here - Do I not need to call these for every path returned from  `GetLocatedImports`?

pepeiborra: If you knew which import is being autocompleted then you could query only that one.

alexnaspo: I was thinking as a followup to this PR, I could improve the completion of imports to also include local modules. In order to do this, I can use the keys from `moduleExportsMap` in `Logic.hs` rather than `importableModules`. Do you think this is a good direction? And if so, does it make sense to query for all located imports so we have this feature?

pepeiborra: > I was thinking as a followup to this PR, I could improve the completion of imports to also include local modules. In order to do this, I can use the keys from `moduleExportsMap` in `Logic.hs` rather than `importableModules`. Do you think this is a good direction? 

Or you could just use the list of imports, which can be extracted from the parsed module. 

> And if so, does it make sense to query for all `GetLocatedImports` so we have this feature?

Querying GetModIface for all located imports and then building a Map is O(nlogn) for the number of imports, but I'm not too worried since in most cases n will be small. So it does make sense, it's just a bit inefficient.

pepeiborra: But please, make sure to query all the GetModIface keys in a single call of usesWithStale so that they all happen in parallel

ghcide/src/Development/IDE/Types/Exports.hs
alexnaspo: addressed

ghcide/src/Development/IDE/Types/Exports.hs
alexnaspo: addressed

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: addressed

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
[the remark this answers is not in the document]
alexnaspo: addressed

pepeiborra:
This looks all good, but let's take a look at the Completion benchmark experiments before landing.

alexnaspo:
> This looks all good, but let's take a look at the Completion benchmark experiments before landing.

@pepeiborra - What is involved with this step? Is this part of CI or is it a manual step that I should run and report on? Happy to work through it

pepeiborra:
> > This looks all good, but let's take a look at the Completion benchmark experiments before landing.
> 
> @pepeiborra - What is involved with this step? Is this part of CI or is it a manual step that I should run and report on? Happy to work through it

There's a CI job that runs the benchmarks, renders the results summary, and produces downloadable artifacts. See https://github.com/haskell/haskell-language-server/actions/runs/1211084983 

However the CI results are not entirely reliable, so you can run locally with `cabal configure --enable-benchmarks && cabal bench ghcide`.

pepeiborra:
Running locally in gitpod, I get the following results :

```
gitpod /workspace/haskell-language-server $ column  -t -s, -c name ghcide/bench-results/unprofiled/cabal/results.csv 

version    name                             success   samples   startup              setup                  userTime             delayedTime             totalTime            maxResidency   allocatedBytes
upstream   edit                             True      50        15.106430794000001   0.0                    126.02090627599999   3.000104200000001e-2    126.06007647000001   209MB          42291MB
HEAD       edit                             True      50        16.701233245         0.0                    148.14813984300008   3.8794936999999995e-2   148.19721942         656MB          42067MB
upstream   hover                            True      50        16.690553153         0.0                    46.163053074000004   0.7427802169999999      47.006025115         189MB          7171MB
HEAD       hover                            True      50        15.991552218         0.0                    50.04384343000001    0.7385649630000003      50.801488767         188MB          7107MB
upstream   hover after edit                 True      50        14.994895467000001   0.0                    45.68902047200001    79.83457485499997       125.53222702500001   208MB          41356MB
HEAD       hover after edit                 True      50        15.504793956         0.0                    45.34694745099999    86.838486113            132.231911924        208MB          41399MB
upstream   getDefinition                    True      50        21.100101736000003   0.0                    48.847873189999994   0.759664964             49.705360482         187MB          7101MB
HEAD       getDefinition                    True      50        15.803872888         0.0                    43.204030774999985   0.49277879500000005     43.704813536         187MB          7025MB
upstream   getDefinition after edit         True      50        15.703868614000001   0.0                    47.07017039400001    61.172846302            108.25133311900001   210MB          39006MB
HEAD       getDefinition after edit         True      50        16.295290536         0.0                    48.24514458200001    78.08464355300003       126.338950823        208MB          38942MB
upstream   completions                      True      50        16.100244966         0.0                    67.26779727300001    0.8191487659999999      68.09844481200001    218MB          12642MB
HEAD       completions                      True      50        28.394669321000002   0.0                    77.16584082200002    1.2095481110000001      78.396245182         220MB          12963MB
upstream   completions after edit           True      50        15.904602636000002   0.0                    63.16818985          77.64433259400002       140.82452278         246MB          45151MB
HEAD       completions after edit           True      50        18.298170006         0.0                    161.2938064010001    1.4900570119999998      162.796853899        252MB          44954MB
...
```

There is a substantial regression in userTime for the "completions after edit" experiment, which takes almost 3 times longer. This is really bad considering that the experiment doesn't actually try to complete within imports - it's just completing on a random identifier inside the module body.  Bummer! 

1. Can we avoid doing all the work *unless* we are completing within an import? Yes, but it will require some major code changes.
2. Can we parallelise the work? Yes, and the most effective way to do so is by extracting a new plugin.

Therefore the way forward is to extract import completion out to a new plugin - we don't need a new package, it can stay within ghcide the same way that we split code actions into multiple plugins recently.

pepeiborra:
Ah, the other thing worth trying is switching from `usesWithStale` to `useWithStaleFast` which will have a major impact. 
Sadly there isn't a plural version of it - you could investigate writing one (probably in a separate PR)

alexnaspoleap:
> Ah, the other thing worth trying is switching from `usesWithStale` to `useWithStaleFast` which will have a major impact.
> Sadly there isn't a plural version of it - you could investigate writing one (probably in a separate PR)

Ahh ok, glad this was caught by the experiments! My original implementation made many calls to useWithStaleFast, we still don't want to use that approach, right? 

One other thought I had would be to pass in `moduleExports` as a lazily evaluated closure that only gets invoked when it is actually called.

I am also happy to separate it out into a separate plugin

pepeiborra:
> > Ah, the other thing worth trying is switching from `usesWithStale` to `useWithStaleFast` which will have a major impact.
> 
> > Sadly there isn't a plural version of it - you could investigate writing one (probably in a separate PR)
> 
> 
> 
> Ahh ok, glad this was caught by the experiments! My original implementation made many calls to useWithStaleFast, we still don't want to use that approach, right? 
> 
> 

It's not ideal, but would be good to benchmark. Overhead will be seen in the delayedTime column

alexnaspoleap:
> > > Ah, the other thing worth trying is switching from `usesWithStale` to `useWithStaleFast` which will have a major impact.
> > 
> > 
> > > Sadly there isn't a plural version of it - you could investigate writing one (probably in a separate PR)
> > 
> > 
> > Ahh ok, glad this was caught by the experiments! My original implementation made many calls to useWithStaleFast, we still don't want to use that approach, right?
> 
> It's not ideal, but would be good to benchmark. Overhead will be seen in the delayedTime column

Just pushed up a commit with this. I am still figuring out how to run the benchmark

alexnaspo:
`gitpod /workspace/haskell-language-server $ column  -t -s, -c name ghcide/bench-results/unprofiled/cabal/results.csv 
version    name                             success   samples   startup              setup          userTime             delayedTime             totalTime            maxResidency   allocatedBytes
upstream   edit                             True      50        15.909690368000001   0.0            59.66717681500001    44.027676416            103.70297152200001   209MB          42245MB
HEAD       edit                             True      50        15.39248249          0.0            96.09728630100003    2.6238283e-2            96.131176632         657MB          42117MB
upstream   hover                            True      50        16.106208916         0.0            43.79434917799999    0.40055335900000005     44.203569291         189MB          7150MB
HEAD       hover                            True      50        16.499268988         0.0            43.32357505200002    0.6733871419999997      44.003287005000004   188MB          7112MB
upstream   hover after edit                 True      50        20.009179772         0.0            42.55134783699999    52.87326434300001       95.43175849900001    209MB          41531MB
HEAD       hover after edit                 True      50        15.696507622         0.0            44.41968844100001    56.38936902699999       100.81651375300001   216MB          41432MB
upstream   getDefinition                    True      50        16.10848742          0.0            43.24465132600001    0.5626395780000001      43.902769687         187MB          7068MB
HEAD       getDefinition                    True      50        15.288892039         0.0            45.564271231999996   0.7326963720000003      46.304268461         187MB          7044MB
upstream   getDefinition after edit         True      50        16.187153753         0.0            45.95820615800001    51.153166584            97.119421434         209MB          39031MB
HEAD       getDefinition after edit         True      50        18.298126303         0.0            43.515627304999995   54.743886827000004      98.267455462         209MB          38993MB
upstream   completions                      True      50        15.405404156000001   0.0            63.33640568799997    0.8768328440000002      64.226908509         218MB          12625MB
HEAD       completions                      True      50        20.89437068          0.0            65.17306839999999    0.897136667             66.167063293         222MB          12848MB
upstream   completions after edit           True      50        16.798457944000003   0.0            62.34925103600001    58.41965749500001       120.779888704        246MB          45131MB
HEAD       completions after edit           True      50        16.407094916000002   0.0            58.350779114         44.505122895000014      102.86656093500001   254MB          44984MB
upstream   code actions                     True      50        15.889378276         39.599578759   5.5171933970000016   0.29360228499999996     5.90346518           181MB          7127MB
HEAD       code actions                     True      50        16.59566376          36.605918613   5.417252828          0.471805958             5.896062601000001    184MB          7113MB
upstream   code actions after edit          True      50        14.909439409         0.0            117.434062149        7.840787299999999e-2    117.519171399        198MB          33330MB
HEAD       code actions after edit          True      50        16.199739865         0.0            100.900419694        0.11985685599999996     101.026790855        201MB          33392MB
upstream   code actions after cradle edit   True      50        16.40297821          0.0            265.79091633400003   73.643542763            339.442235068        365MB          312883MB
HEAD       code actions after cradle edit   True      50        17.998559606         0.0            265.057900699        85.465094965            350.53132146300004   403MB          312544MB
upstream   documentSymbols after edit       True      50        17.000996001         0.0            20.71816345600001    41.770951530999994      62.497089678         187MB          12615MB
HEAD       documentSymbols after edit       True      50        16.399671549         0.0            23.27910427300001    45.531524925000014      68.90635907400001    183MB          12321MB
upstream   hole fit suggestions             True      50        20.501716008000002   0.0            125.378847756        3.0133202000000005e-2   125.41846698100001   219MB          62397MB
HEAD       hole fit suggestions             True      50        17.503889020000003   0.0            133.87482879200002   0.10469722899999999     133.988609357        733MB          62045MB`

alexnaspo:
The benchmarks look much better. `upstream   completions                      True      50        15.405404156000001   0.0            63.33640568799997    0.8768328440000002      64.226908509         218MB          12625MB
HEAD       completions                      True      50        20.89437068          0.0            65.17306839999999    0.897136667             66.167063293         222MB          12848MB
upstream   completions after edit           True      50        16.798457944000003   0.0            62.34925103600001    58.41965749500001       120.779888704        246MB          45131MB
HEAD       completions after edit           True      50        16.407094916000002   0.0            58.350779114         44.505122895000014      102.86656093500001   254MB          44984MB` - sorry for the format. I am not sure how you formatted it so nicely

alexnaspo:
Screen Shot 2021-09-08 at 9 37 53 PM
The benchmarks look much better

pepeiborra (approved):
And that's why benchmarks are important. 

Marking this as ready to merge now, follow ups are optional and can be done in future PRs.

Thanks @alexnaspo for your dedication, this was an awesome effort.

jneira (approved):
Amazing tour the force from you both @pepeiborra and @alexnaspo, congrats!

jneira:
This fixed (hopefully) #314

PR #2155, opened 2021-09-03
Create a citation

ndmitchell:
GitHub now supports citation files https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-citation-files. Might as well use that to link to the HLS paper many of us were involved in.


pepeiborra (approved):
That won't create a link to the paper. Ah, I see what you are doing now 😏

jneira (approved):
thanks!

PR #2156, opened 2021-09-04
Remove ghc-api source snapshot

pepeiborra:
We can replace the source snapshot with Cabal conditions to unblock Hackage uploads until https://github.com/haskell/haskell-language-server/pull/2128 lands

~~The source snapshot cannot be removed for ghc 8.10.7 nor 9.0.1. In the former, an allow-newer suffices, but in the later we need unreleased changes that can only be obtained via the source snapshot.~~

/cc @fendor


jneira:
this looks like can be backported to 1.3.0-hackage
I was thinking in give up trying to release 1.3.0 and go directly for 1.4.0 starting right now, thoughts?

pepeiborra:
> this looks like can be backported to 1.3.0-hackage
> I was thinking in give up trying to release 1.3.0 and go directly for 1.4.0 starting right now, thoughts?

When do you plan to release 1.4.0? We haven't had a Hackage release in a long time, and this has been blocking adoption for several user groups.

jneira:
btw ghc-api-compat does not have support for ghc-8.10.7 without allow newer 😕

pepeiborra:
This is ready to go! GHC up to 8.10.6 is the best I could do until Sylvain uploads [new versions to Hackage](https://gitlab.haskell.org/haskell/ghc-api-compat/-/issues/1) or until @fendor lands #2128

jneira (approved):
many thanks for the plumbing work, maybe the conditionals willbe removed soon (if fendor's pr lands soon)

pepeiborra:
Sylvain has uploaded new versions to Hackage, so I will add more conditionals and resubmit

jneira:
@pepeiborra it seems we could merge soon #2156 so we could:
* close this and go for it directly, i think it could be hard to backport, so we would go for 1.4.0 and skip 1.3.0 for hasckage
* merge this, backport to 1.3.0 and make the hackage release, and rebase and merge #2156 just after that
* backport this pr to 1.3.0-hackage without merge it in master

thoughts?

//cc @fendor

pepeiborra:
Merge, backport and upload to Hackage is my preference, assuming it's not holding #2128 back

jneira:
@MergefyIO backport 1.3.0-hackage

jneira:
@MergifyIO backport 1.3.0-hackage

pepeiborra:
Backported manually

PR #2157, opened 2021-09-04
Reduce duplication in pragma tests

nini-faroux:
Just removes a good bit of repetition from the pragma tests.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2157"></a>


jneira (approved):
nice catch!

PR #2160, opened 2021-09-05
Wingman: Let-bindings in metatactics

### LINKED ISSUE #2002: Wingman feature request: Support for tactic subgoals via let
Feature request: there should be a tactic that allows the user to introduce arbitrary subgoals. Currently, subgoals are only introduced via destruct_all or similar tactics. But in general some subgoals are desired as helpers that cannot be produced from destruct and it's variants, so we should have a `define` tactic (or something similar) that let's use make arbitrary subgoals. For example, given:
```
_ :: a
```
and using a tactic `define b :: b`, we should be able to produce:
```
let b = (_ :: b) in (_ :: a)
```
This then should give us `b` as a subgoal, which when completed gives us `b :: b` as a hypothesis for completing the `a` goal.

isovector:
This PR introduces the `let` tactic, which allows you to bind variables in tactic metaprograms. It takes a variadic number of binder names, and produces a hole for each, in order.

Fixes #2002


jneira (approved):
we were missing you 😉

isovector:
I've been on a nice vacation in Mexico :) Back in full force soon.

PR #2161, opened 2021-09-05
Add a Gitpod descriptor

pepeiborra:
Develop HLS from your browser in a prebuilt environment.  

Try it with https://gitpod.io/#https://github.com/haskell/haskell-language-server/tree/gitpod

The init script uses ghcup to install GHC and cabal to prebuild all the dependencies of the test suites. 

Using Nix is not an option since the root file system is read-only.

The Free tier allows for 50h monthly hours and the hardware is beefy (>80 cores, 64GB RAM). 

Super handy for testing compatibility with diverse GHC versions.


.gitpod.yml
```diff
@@ -0,0 +1,39 @@
...
+    # add a "Review in Gitpod" button to pull requests (defaults to false)
+    addBadge: false
+    # add a label once the prebuild is ready to pull requests (defaults to false)
+    addLabel: prebuilt-in-gitpod
+
+vscode:
+  extensions:
+    - "haskell.haskell"
+    - "justusadam.language-haskell"
+    - "usernamehw.errorlens"
```
ndmitchell, line 39: Do we recommend this for users? Should we?

ndmitchell (approved):

jneira (approved):
nice! i tried to add one of this but i was not able to make vscode see cabal nor ghc, maybe cause we did not update the extension in vscodium until recently, many thanks!

pepeiborra:
> nice! i tried to add one of this but i was not able to make vscode see cabal nor ghc, maybe cause we did not update the extension in vscodium until recently, many thanks!

It is a bit tricky, but not impossible hopefully

.gitpod.yml
pepeiborra: Yes, errorlens works great with HLS.

.gitpod.yml
```diff
@@ -0,0 +1,53 @@
...
+  - before: |
+      export GHCUP_INSTALL_BASE_PREFIX=/workspace
+      export BOOTSTRAP_HASKELL_NONINTERACTIVE=1
+      export BOOTSTRAP_HASKELL_MINIMAL=1
+      export CABAL_DIR=/workspace/.cabal
+      export STACK_ROOT=/workspace/.stack
+      curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+
+      # ensure that there is an HLS binary for the installed GHC version
+      /workspace/.ghcup/bin/ghcup install ghc 8.10.5 --set
```
Ailrun, line 12: Is there a specific reason for using 8.10.5 instead of any later version of GHC?

pepeiborra: Yes, is the comment not helpful?

Ailrun: Ah, I misunderstood the comment. I see it now, sorry for bothering you.

PR #2162, opened 2021-09-06
Improve incoming call for typeclass and type family instance

July541:
Now the incoming call about type class/family will jump to the corresponding instance.

```haskell
class F a where
  f :: a
instance F Bool where
  f = x
instance F Int where
  f = y

x = True
y = 3
```

Previously, If we request an incoming call for `x`, it will jump to line 2, but now it will go to line 4, the more direct result.


jneira (approved):
many thanks for the fix

July541:
In the latest commit, I replaced constant delay with `retry`.

jneira (approved):

PR #2163, opened 2021-09-06
Semiautomatic hackage releases

### LINKED ISSUE #1310: Automate hackage releases

jneira:
* This try to make hackage releases as smooth as possible, hopefully fixes #1310
* It is dumb about changes: if some file (module obvious exclusions like docs files) has changed, we have to bomp up the version or it will fail
* It builds all packages for three last ghc major versions so it will take some time (with cold cache)
* It upload package candidates, a human should confirm the definitive release (ahopefully fter some manual checks)
* It depends on #2156 to avoid errors about ghc-api-compat
* ~~It is failing for hls-call-hierarchy-plugin:~~

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2163"></a>


.github/workflows/hackage.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Cache Cabal
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-cabal
+        with:
+          path: |
+            ~/.cabal/packages
+            ~/.cabal/store
+          key: v2-${{ runner.os }}-${{ matrix.ghc }}-hackage-${{ hashFiles('**/*.cabal') }}
```
jneira, line 45: i think we could reuse the test/bench cache here

July541:
It is caused by `aeson`'s `Value`, `Value` support `Ord` instance after their [1.5.2.0](https://hackage.haskell.org/package/aeson-1.5.6.0/docs/src/Data.Aeson.Types.Internal.html#line-398).

I checked your test and found it uses [1.4.7.1](https://github.com/jneira/haskell-language-server/runs/3524694108?check_suite_focus=true#step:11:136)...

I think to add a lower bound for `aeson` in `hls-call-hierarchy-plugin` is enough (Maybe you can add it directly, I don't know how to participate in this thread..)

jneira:
> I think to add a lower bound for `aeson` in `hls-call-hierarchy-plugin` is enough (Maybe you can add it directly, I don't know how to participate in this thread..)

dont worry, the diagnostic is the "hard" part, will add the bound here, thanks!

.github/workflows/hackage.yml
```diff
@@ -0,0 +1,127 @@
...
+  push:
+    branches:
+      - '*-hackage'
+
+jobs:
+  check-and-upload-tarballs:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package: ["hie-compat", "hls-graph", "shake-bench",
```
isovector, line 13: Looks like this list needs to be kept up to date when new plugins are added? Do we have any documentation on how to add new plugins? If so, we should document this fact there too.

jneira: Yeah, fortunately plugins have an unique id and it is relatively easy search in the codebase using it. So search for a similar existing plugin is a required step in front of outdated documentation. But you are right, i guess the best place to list all the places (including .cabal and generic config) could be https://haskell-language-server.readthedocs.io/en/latest/contributing/plugin-tutorial.html#wrapping-up

jneira: @isovector #2166

.github/workflows/hackage.yml
pepeiborra: The total cache limit is still 5GB so I would say reusing the test cache would be much preferred to avoid over-running the limit:

https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

.github/workflows/hackage.yml
```diff
@@ -0,0 +1,127 @@
...
+          # This will throw an error if there is any difference casue we have to bump up the package version
+          diff -qr -x "*.md" -x "data" $(ls -d ./incoming/${{ matrix.package }}-*) $(ls -d ./current/${{ matrix.package }}-*)
+
+      - name: "Create appropiate cabal.project"
+        if: steps.get-hackage-version.exists != 'true'
+        run: |
+          cd $(ls -d ./incoming/${{ matrix.package }}-*)
+          echo "packages: . ../../* ../../plugins/*" > cabal.project
+          # TODO: remove when not needed
+          echo "allow-newer: Chart-diagrams:diagrams-core, SVGFonts:diagrams-core" >> cabal.project
```
pepeiborra, line 95: These are only needed to build the benchmark suite of ghcide. The benchmark suite is not a requirement for uploading to Hackage imho, and should be skipped

.github/workflows/hackage.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: "Build all package components in isolation"
+        if: steps.get-hackage-version.exists != 'true'
+        run: |
+          cd $(ls -d ./incoming/${{ matrix.package }}-*)
+          cabal build --enable-tests --enable-benchmarks
```
pepeiborra, line 101:
```suggestion
          cabal build --enable-tests
```

.github/workflows/hackage.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: "Generate haddock for hackage"
+        if: steps.get-hackage-version.exists != 'true'
+        run: |
+          cd $(ls -d ./incoming/${{ matrix.package }}-*)
+          cabal haddock --haddock-for-hackage
```
pepeiborra, line 107: given that we build them, can we upload the Haddocks too?

pepeiborra (approved):
Nice job, this is going to be super useful

pepeiborra:
Next step would be integrating @kowainik Policeman to validate the updated version numbers: https://kowainik.github.io/projects/policeman

.github/workflows/hackage.yml
jneira: ya, but to use a simple autogenerated cabal.project with `packages: . ../../* ../../plugins/*` cabal has to find a build plan for shake-bench. The alternative is a cabal project listing explicitly all packages but shake-bench.
Otoh in theory you should be able to run the benchmarks (like tests) with the hackage version too, and ideally those allow-newer should be removed to allow it

.github/workflows/hackage.yml
jneira: ideally benchmarks should be able to run using the hackage version

jneira:
> 
> 
> Next step would be integrating @kowainik Policeman to validate the updated version numbers: https://kowainik.github.io/projects/policeman

wow, i had forgotten it, it would be really nice

jneira:
I've used a hackage token generated from my account, it can be changed in the project secrets
However i have not perms to upload to all packages involved, missing are:
* [ ] https://hackage.haskell.org/package/shake-bench
* [ ] https://hackage.haskell.org/package/hie-compat
* [ ] https://hackage.haskell.org/package/hls-graph
* [ ] https://hackage.haskell.org/package/hls-ormolu-plugin
* [ ] https://hackage.haskell.org/package/hls-module-name-plugin
* [ ] https://hackage.haskell.org/package/hls-refine-imports-plugin
* [ ] *https://hackage.haskell.org/package/hls-call-hierarchy-plugin*, it is not uploaded

Most of them have @pepeiborra as unique maitainer so an option could be that @pepeiborra  generates a hackage api key (i would create a new, dedicated, one). But even doing that i would add some more maintainers to those plugin with only one parent :wink:

pepeiborra:
I have added you to the maintainers list for all of them

jneira:
Ok the workflow has failed in the last step, cause i did not put the hackage api key:

```
Run haskell-actions/hackage-publish@v1
  with:
    packagesPath: packages
    publish: false
    hackageServer: http://hackage.haskell.org
Uploading hls-module-name-plugin-1.0.0.1 to http://hackage.haskell.org/packages/candidates
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  7777    0     0  100  7777      0  42037 --:--:-- --:--:-- --:--:-- 42037
curl: (22) The requested URL returned error: 401 Unauthorized
Error: Process completed with exit code 22.
```

I think the deinitive test can be done with the next release

jneira:
@MergifyIO backport 1.3.0-hackage

pepeiborra:
It would be nice to back port this to the 1.3.0-hackage branch and use it for the Hackage upload

jneira:
I was waiting for the automatic backport, but CI is quite slow lately, will backport manually

jneira:
backported with 97003e2
ci running in https://github.com/haskell/haskell-language-server/actions/runs/1212661332

jneira:
As the workflow will take ages to start i've started it in my repo: https://github.com/jneira/haskell-language-server/runs/3545669772

PR #2164, opened 2021-09-07
Wingman: Add "New Unification Variable" helper

isovector:
Lots of things in Wingman need to create so-called "unification variables," which are type variable types that *are not skolems.* Unification variables will happily unify with anything, so they come up whenever Wingman needs to instantiate a polymorphic type, or let-generalize something.

When trying to write a test for #2160 I thought there was a bug in the unification variables created by the new let binding stuff. But it was hard to track down, because every call site that needed unification variables would spin them up by itself --- lots of room for error.

This PR makes a combinator for creating fresh unification variables.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2164"></a>


jneira (approved):

isovector:
Maybe one day CI will let this merge...

PR #2165, opened 2021-09-07
Complete contributing guide

jneira:
* with the contents of ./CONTRIBUTING.md (mainly the pre commit hook to format)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2165"></a>


docs/contributing/contributing.md
```diff
@@ -97,6 +131,59 @@ To do every time you changed code and want to test it:
...
+If you don't want to use [nix](https://nixos.org/guides/install-nix.html), you can instead use [pre-commit](https://pre-commit.com) with the following config.
+
+```json
+{
+  "repos": [
+    {
+      "hooks": [
+        {
+          "entry": "stylish-haskell --inplace",
+          "exclude": "(^Setup.hs$|test/testdata/.*$|test/data/.*$|^hie-compat/.*$|^plugins/hls-tactics-plugin/.*$)",
```
Ailrun, line 160: Note: we need to update this regex to include up-to-date "excludes" setting in nix config. The current Nix config ignores more files than these.

docs/contributing/contributing.md
```diff
@@ -72,11 +65,52 @@ To create binaries:
...
 GHC 8.6.5 is not supported here because `nixpkgs-unstable` no longer maintains the corresponding packages set.
 
-## Introduction tutorial
+## Testing
 
-See the [tutorial](./plugin-tutorial.md) on writing a plugin in HLS.
+The tests make use of the [Tasty](https://github.com/feuerbach/tasty) test framework.
+
+There are two test suites in the main haskell-language-server package, functional tests, and wrapper tests.
+Other project packages, like the core library or plugins, can have its own test suite.
```
jhrcek, line 73:
```suggestion
Other project packages, like the core library or plugins, can have their own test suite.
```

Ailrun (approved):
Thank you for the update!

docs/contributing/contributing.md
```diff
@@ -72,11 +65,52 @@ To create binaries:
...
+line arguments, to select the specific tests to run.
+
+```bash
+$ cabal test func-test --test-option "-p hlint"
+```
 
-## Test your hacked HLS in your editor
+The above recompiles everything every time you use a different test option though.
+
+An alternative is
```
jhrcek, line 107: Can you please give an explanation as to why the alternative is preferrable?
E.g. (not sure if that's true)

An alternative, which only recompiles when tests (or dependencies) change:

docs/contributing/contributing.md
jneira: @jhrcek thanks!

docs/contributing/contributing.md
jneira:
```suggestion
An alternative, which only recompiles when tests (or dependencies) change:
```
That would be the proposed change?

docs/contributing/contributing.md
jneira:
```suggestion
          "exclude": "(^Setup.hs$|test/testdata/.*$|test/data/.*$|test/manual/lhs/.*$|^hie-compat/.*$|^plugins/hls-tactics-plugin/.*$|^ghcide/src/Development/IDE/GHC/Compat.hs$|^plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs$|^ghcide/test/exe/Main.hs$|ghcide/src/Development/IDE/Core/Rules.hs|^hls-test-utils/src/Test/Hls/Util.hs$)",
```

that would be correct?

docs/contributing/contributing.md
jhrcek: Yes, if that's true :)

jneira: just checked it :slightly_smiling_face:

jhrcek (approved):
Thank you. LGTM

michaelpj (approved):

docs/contributing/contributing.md
```diff
@@ -72,11 +65,52 @@ To create binaries:
...
+$ cabal test func-test --test-option "-p hlint"
+```
 
-## Test your hacked HLS in your editor
+The above recompiles everything every time you use a different test option though.
+
+An alternative, which only recompiles when tests (or dependencies) change:
+
+```bash
+$ cabal run haskell-language-server:func-test -- -p "hlint enables"
```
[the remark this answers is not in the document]
pepeiborra: No, that won't help. `cabal run` does not change the environment, as far as I know. 
You can do `cabal exec cabal run h-l-s:func-test -- -- -p ...` and it will work sometimes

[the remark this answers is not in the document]
jneira: > 
> 
> No, that won't help. `cabal run` does not change the environment, as far as I know.

It seems so: https://github.com/haskell/cabal/issues/5411#issuecomment-402548725. It works for `cabal test`but no for `cabal run` iiuic that comment

> You can do `cabal exec cabal run h-l-s:func-test -- -- -p ...` and it will work sometimes

It works enough times to update the instructions? are you are referring to the fact `cabal exec` will not rebuild the executable and you could end using a older version?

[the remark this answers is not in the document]
jneira: Moreover with cabal-install-3.8, test arguments will not trigger a rebuild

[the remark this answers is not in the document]
pepeiborra: I believe `cabal exec` does trigger a rebuild, but it fails with weird errors in some cases.

docs/contributing/contributing.md
jneira: wow i did not apply this commit 🤦

PR #2169, opened 2021-09-07
Use maxBound of uinteger not Int.

pranaysashank:
This is a low effort fix for https://github.com/haskell/vscode-haskell/issues/445 . The lsp spec says 
```
/**
 * Defines an unsigned integer number in the range of 0 to 2^31 - 1.
 */
export type uinteger = number;

interface Position {
	line: uinteger;
	character: uinteger;
}
```

A proper fix would be to create a new type in `lsp-types` for uinteger. 

### Debug info
This check returns `false` https://github.com/microsoft/vscode-languageserver-node/blob/8ad6896d8910214b7a3c8fee8dd175077a743045/client/src/common/client.ts#L1940 and vscode tries to parse it as `SymbolInformation` instead of `DocumentSymbol` and you get the error reported in the above issue.

The check for uinteger is at https://github.com/microsoft/vscode-languageserver-node/blob/8ad6896d8910214b7a3c8fee8dd175077a743045/types/src/main.ts#L3756

### DIFF
--- a/ghcide/src/Development/IDE/LSP/Outline.hs
+++ b/ghcide/src/Development/IDE/LSP/Outline.hs
@@ -48,7 +48,7 @@ moduleOutline ideState DocumentSymbolParams{ _textDocument = TextDocumentIdentif
                    (defDocumentSymbol l :: DocumentSymbol)
                      { _name  = pprText m
                      , _kind  = SkFile
-                     , _range = Range (Position 0 0) (Position maxBound 0) -- _ltop is 0 0 0 0
+                     , _range = Range (Position 0 0) (Position 2147483647 0) -- _ltop is 0 0 0 0
                      }
                  _ -> Nothing
                importSymbols = maybe [] pure $


pranaysashank:
Looks like this is a recent change to the spec, from 3.16. I couldn't find a corresponding changelog in the specification.

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -48,7 +48,7 @@ moduleOutline ideState DocumentSymbolParams{ _textDocument = TextDocumentIdentif
                    (defDocumentSymbol l :: DocumentSymbol)
                      { _name  = pprText m
                      , _kind  = SkFile
-                     , _range = Range (Position 0 0) (Position maxBound 0) -- _ltop is 0 0 0 0
+                     , _range = Range (Position 0 0) (Position 2147483647 0) -- _ltop is 0 0 0 0
```
fendor, line 51: Magic constants are usually bad, a comment would be helpful, or maybe

```suggestion
                     , _range = Range (Position 0 0) (Position (2 ^31 - 1) 0) -- _ltop is 0 0 0 0
```

pranaysashank: I added a comment and hardcoded the constant (ghc warns and makes me add a type annotation to 31 otherwise). Also changed the usages of maxBound in test and bench which is probably not needed.

fendor (approved):
LGTM, thanks!

jneira (approved):
@pranaysashank many thanks for fix that disturbing bug

PR #2172, opened 2021-09-08
Wingman: Use awareness of skolems when filtering GADT data constructors

### LINKED ISSUE #2171: Wingman incorrectly ignores skolems when dealing with GADT evidence
Wingman uses `dataConCannotMatch` from GHC to check when it should filter out GADT data cons. But `dataConCannotMatch` only checks if two type indices certainly can't match. But if one of them is a skolem, GHC doesn't know this, and thinks it's merely a type variable. In which case, it _could_ match, but we know better.

isovector:
GHC provides `dataConCannotMatch`, which WIngman was using to filter out GADT data constructors which don't typecheck. But GHC doesn't do any reasoning about skolems in this function, so it assumes it can just unify any type variables that are present. This is entirely wrong --- skolems are definitely apart for all of Wingman's purposes.

Fixes #2171

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/EmptyCase.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/EmptyCase.hs
@@ -68,7 +68,9 @@ emptyCaseInteraction = Interaction $
             range = realSrcSpanToRange $ unTrack ss
         matches <-
           liftMaybe $
-            destructionFor
+          -- TODO(sandy): This needs a judgment in order to correctly compute
+          -- the skolems! But we don't have a judgement yet :(
+            destructionFor mempty
               (foldMap (hySingleton . occName . fst) bindings)
               ty
         edits <- liftMaybe $ hush $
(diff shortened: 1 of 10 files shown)


plugins/hls-tactics-plugin/src/Wingman/EmptyCase.hs
```diff
@@ -68,7 +68,9 @@ emptyCaseInteraction = Interaction $
             range = realSrcSpanToRange $ unTrack ss
         matches <-
           liftMaybe $
-            destructionFor
+          -- TODO(sandy): This needs a judgment in order to correctly compute
```
isovector, line 71: Filed at #2173

jneira: maybe it worths to add the link to the issue?

jneira (approved):
https://en.wikipedia.org/wiki/Thoralf_Skolem force be with you!

jneira:
I am afraid this needs a rebase

isovector:
The test failure in `MetaCataAST `:

should be:

```haskell
eval (IntLit n) = n
```

using GADT evidence, but instead we get:

```haskell
eval (IntLit n) = _w0
```

Looks like it didn't learn that the type variable is equal to `Int` in this case!

PR #2175, opened 2021-09-08
Set bounds for retrie by ghc

jneira:
* To be backported to 1.3.0-hackage (See #1990)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2175"></a>

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -71,7 +71,6 @@ library
         prettyprinter-ansi-terminal,
         prettyprinter,
         regex-tdfa >= 1.3.1.0,
-        retrie,
         rope-utf16-splay,
         safe,
         safe-exceptions,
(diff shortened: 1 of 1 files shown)


pepeiborra:
This shouldn't be necessary, ghcide doesn't care, it's the Cabal solver job to find a working set of versions right?

jneira:
> This shouldn't be necessary, ghcide doesn't care, it's the Cabal solver job to find a working set of versions right?

yep, but it is not able to do it in this case, will try increasing max-backjumps but the package should be built with the default ones :worried:

pepeiborra:
> > This shouldn't be necessary, ghcide doesn't care, it's the Cabal solver job to find a working set of versions right?
> 
> 
> 
> yep, but it is not able to do it in this case, will try increasing max-backjumps but the package should be built with the default one :worried: 

But why is it able to do it in the other workflows?

jneira:
> It is able to do it in CI just fine except in the new workflow. Why?

I've tried to reproduce locally, following the workflow steps verbatim and i am not able to reproduce neither, even with a fresh new cabal config and cache

jneira:
Aaaand https://github.com/haskell/haskell-language-server/actions/runs/1216111548 has built ghcide without the change of this pr but adding a upper bound to `extras`, wonderful

Will close eyes and the pr :stuck_out_tongue:

PR #2184, opened 2021-09-10
Wingman: Don't clobber where clauses

### LINKED ISSUE #2183: Wingman removes functions in nested where
```
foo = undefined
  where
    bar :: Maybe a -> b
    bar = _
      where
        pleaseDontHurtMe = undefined
```

Refine hole on the only hole (`bar = _`)

You get:

```
foo = undefined
  where
    bar :: Maybe a -> b
    bar m_a = _we
```

@isovector why did you hurt my nested where function :(

This might well be an issue with a Wingman dependency, I don't know Wingman's internals well enough to judge...

isovector:
Wingman generates lambdas and case expressions by default. When it thinks it's operating over a top-level hole, it will instead replace the entire match group, by desugaring the lambdas and cases into explicit matches. This works fine unless there's a `where` attached to the original match group --- in which case it gets clobbered.

This isn't the world's best fix, but it's a quick fix. The bug is pretty destructive so I'd rather get it fixed quickly than perfectly. And it's not clear what a perfect solution would look like here.

Fixes #2183

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/LanguageServer.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/LanguageServer.hs
@@ -273,7 +273,7 @@ mkJudgementAndContext cfg g (TrackedStale binds bmap) rss (TrackedStale tcg tcgm
         mkFirstJudgement
           ctx
           (local_hy <> cls_hy)
-          (isRhsHole tcg_rss tcs)
+          (isRhsHoleWithoutWhere tcg_rss tcs)
           g
     , ctx
     )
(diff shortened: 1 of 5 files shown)


plugins/hls-tactics-plugin/src/Wingman/LanguageServer.hs
```diff
@@ -479,12 +479,25 @@ mkIdHypothesis (splitId -> (name, ty)) prov =
...
--- | Is this hole immediately to the right of an equals sign?
-isRhsHole :: Tracked age RealSrcSpan -> Tracked age TypecheckedSource -> Bool
-isRhsHole (unTrack -> rss) (unTrack -> tcs) =
+-- | Is this hole immediately to the right of an equals sign --- and is there
+-- no where clause attached to it?
+--
+-- It's important that there is no where clause because otherwise it gets
+-- clobbered. See #2183 for an example.
+--
+-- This isn't a perfect check, and produces some ugly code. But it's much much
```
santiweight, line 488: In what way is this not a perfect check? I don't quite catch your meaning here.

santiweight: I also don't quite understand why this produces ugly code.

santiweight (commented):
Looks good to me :)

jneira (approved):

PR #2187, opened 2021-09-12
Completions for project identifiers

pepeiborra:
This fixes an oversight from #2040 where I forgot to weave in the project exports map. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2187"></a>


jneira (approved):
thanks for fixing it

PR #2190, opened 2021-09-13
enable completions of local imports

alexnaspo:
Enable completions of local project modules

`Import Development.IDE.<cursor-position>` will show completions for locally built modules

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2190"></a>


ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -604,12 +600,11 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         , enteredQual `T.isPrefixOf` label
         ]
 
-      filtImportCompls = filtListWith (mkImportCompl enteredQual) importableModules
+      filtImportCompls = filtListWith (mkImportCompl enteredQual) $ map fst $ HM.toList moduleExportsMap
```
jhrcek, line 603: You could use [HM.keys](https://hackage.haskell.org/package/unordered-containers-0.2.14.0/docs/Data-HashMap-Strict.html#v:keys) here to simplify

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -604,12 +600,11 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         , enteredQual `T.isPrefixOf` label
         ]
 
-      filtImportCompls = filtListWith (mkImportCompl enteredQual) importableModules
+      filtImportCompls = filtListWith (mkImportCompl enteredQual) $ HM.keys moduleExportsMap
```
pepeiborra, line 603: Strictly speaking this change is not correct in a multi-component project, because the keys of `moduleExportsMap` may include modules that are not importable in the current component. 

For example, in the usual lib and test project, it is common for the test component to have dependencies on test packages (hspec, tasty, quickCheck, etc) which are not available in the lib component.

This also applies to the completions that we generate for identifiers - they should be intersected with the importable modules from `envVisibleModules`

pepeiborra (changes requested):

jneira:
Cant believe there is no feature request for this (or i dont find it) cause it is a nice one, thanks!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
alexnaspo: I understand the problem, but I am a bit confused by the suggested solution. Won't intersecting with `envVisibleModuleNames` just leave us with what we had in the first place? (which did not include local modules)

pepeiborra: You are right, I did not look into the definition of `envVisibleModuleNames` and it only includes package modules. 

The `GetKnownTargets` key returns the list of local modules for all the components, you could use that. It will still not be 100% correct since a module from the lib component cannot import a module from the test component, but fixing that can be future work.

pepeiborra: - Either it's empty (what does the project look like?)
- Or it's returning a Nothing (it shouldn't, but check)
- Or it doesn't work with `useWithStaleFast` for some reason - try just `use`

alexnaspo: - I am working on the HLS project to test my changes
- It is returning `Nothing` when attempting to auto complete within `Completions.hs`
- `use` also returns `Nothing`

pepeiborra: I see the problem now. GetKnownTargets is a file-independent rule key so you should call useNoFile (or the equivalent StaleFast version)

alexnaspo:
@pepeiborra - I updated to use `GetKnownTargets`. I'm guessing there might be a faster way to achieve this. I did not see a `useStaleNoFile` but I may be missing it.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -151,7 +156,7 @@ getCompletionsLSP ide plId
                 exportsCompls = mempty{anyQualCompls = exportsCompItems}
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls
 
-            pure (opts, fmap (,pm,binds) compls, moduleExports)
+            pure (opts, fmap (,pm,binds) compls, moduleExports, lModules)
```
pepeiborra, line 159: Instead of returning `lModules`, I would weave them into the `compls` value reusing the Monoid instance. That way you don't need to add a return value here, nor an extra argument to `getCompletions

pepeiborra (changes requested):

ghcide/src/Development/IDE/Plugin/Completions.hs
alexnaspo: Much cleaner, thanks!

alexnaspo:
Addressed the rebase @jneira

pepeiborra (approved):
Awesome, thanks!

PR #2195, opened 2021-09-15
Consider all root paths when suggesting module name change.

### LINKED ISSUE #1903: Wrong module name suggestions when using a bios cradle
* Extracted from #1901, [reported by @chris-martin ](https://github.com/haskell/haskell-language-server/issues/1901#issuecomment-857064054)
* Cradle output:

```
-ihaskell/typeclasses/src
-XApplicativeDo
-XBlockArguments
-XDeriveFunctor
[...]
Typeclasses.Postgres.Querying
Typeclasses.Postgres.Connection.PoolSetup
Typeclasses.Postgres.Connection.Secrets
[...]
```

* Wrong recommendation:

Screenshot from 2021-06-08 13-48-00

/cc @tittoassini (plugin author)

cdsmith:
If there are source root dirs nested inside each other, a module might have
more than one possible name, depending on where it's intended to be imported
from.  In this case, a rename should not be suggested unless the module name
doesn't match any possible correct name.  When suggesting a name, the shortest
name should be suggested, since that's more likely to be the intended one.

Fixes #1903

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2195"></a>


jneira:
Many thanks for the fix, tests looks great

jneira (approved):

Issue #2168, opened 2021-09-07
ghc support deprecation schedule


jneira:
* As we have several issues about i think it is useful to unify all of them and inform users of our plans to deprecate ghc support.
* ghc support is the main source of ci consumption resources so we need to have the right balance between user needs and them
* otoh we have a huge amount of cpp'd code based on ghc version which can be removed (or even entire packages as `hie-compat`
* The proposed schedule could be:

| condition | removed ghc versions | supported ghc versions |
|-----------|--------------|-------------------------|
| *after* next release 1.4.0 with 8.10.7 | 8.10.4, 8.10.3, 8.10.2, 8.8.3, 8.6.4 | 9.0.1 (partial), 8.10.7, 8.10.6, 8.10.5, 8.8.4, 8.6.5 |
| with full support for ghc-9.0 | 8.6.5 | 9.2.* (partial)?, 9.0.*, 8.10.7, 8.10.6, 8.10.5, 8.8.4
| with full support for ghc-9.2 | 8.10.6, 8.10.5, 8.8.4 | 9.2.* , 9.1.*, 8.10.7 |

* We should announce this schedule in all comm channels and remind it in the previous to deprecation hls version
* It could change after any reasonable user feedback
* "full support" has to be defined but
  * it should include all plugins and features
  * the full supported ghc should be supported as well by a stack lts snapshot
* haskell-language-server-wrapper will continue to choose haskell-language-server-8.10.6 f.e. whatever would be the hls own version, so users will be able to continue using older versions of hls for deprecated ghc's

//cc @pepeiborra @Ailrun @ndmitchell @mpickering @wz1000 @berberman @isovector @hasufell

hasufell:
I'd suggest to support at most 2 GHC versions in a X.Y branch. So `8.10.7` and `8.10.6`, but not `8.10.5`.

jneira:
> 
> 
> I'd suggest to support at most 2 GHC versions in a X.Y branch. So `8.10.7` and `8.10.6`, but not `8.10.5`.

That would be the general criteria but i am afraid that 8.10.6 was a somewhat failed 8.10.7 so i think it worths to keep 8.10.5 a little bit but give users some more time to jump to 8.10.7

jneira:
[...]

PR #2197, opened 2021-09-16
Drop ghc support for 8.6.4, 8.10.2, 8.10.3, 8.10.4

jneira:
* Following the incoming ghc deprecation policy #2168  
* Include the bump of ghc-lib* versions to 8.10.7 and the use of 8.10.7 in tests and benchmarks
* Update hackage index to 2021-09-16T07:00:23Z

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2197"></a>


pepeiborra (approved):
Nice!

July541:
Should we change the default `stack.yaml`(https://github.com/haskell/haskell-language-server/blob/master/stack.yaml) to 8.10.7?

jneira:
> 
> 
> Should we change the default `stack.yaml`(https://github.com/haskell/haskell-language-server/blob/master/stack.yaml) to 8.10.7?

good catch! i forgot to update it, thanks!

Ailrun (approved):
LGTM! However, this PR cannot be automatically merged, as
1. This PR changes CI config
2. This PR adds/removes some CI entries, so branch protection rule should be updated

jneira:
* Changed branch protection rules, we have to update rest of prs

anka-213:
Why is 8.8.3 still here? Will it be removed at a later point or is it meant to stay for longer?

jneira:
> 
> 
> Is there any specific reason why 8.8.3 is still here? Will it be removed at a later point or is it meant to stay for longer or was it simply forgotten?

hi, we keep it following the deprecation policy discussed in #2168, it says we will keep the last 2 minor versions for relatively recent ghc versions

anka-213:
Ah, I see! I missed that part! Thanks!

anka-213:
I was looking at the table which said:

condition | removed ghc versions | supported ghc versions
-- | -- | --
after next release 1.4.0 with 8.10.7 | 8.10.4, 8.10.3, 8.10.2, 8.8.3, 8.6.4 | 9.0.1 (partial), 8.10.7, 8.10.6, 8.10.5, 8.8.4, 8.6.5

which is what confused me. (ghc-8.8.3 is under the removed section)

jneira:
oh, yeah, that was the initial proposal but it has changed due the discussion, we have to write down the definitive one with a new table

PR #2198, opened 2021-09-16
added tests for qualified completions

alexnaspo:
I have added some tests related to https://github.com/haskell/haskell-language-server/issues/2082

The first test passes, however the second test fails. 

Is this enough to prove that the functionality in issue 2082 is working fine?

However, I believe this also proves we have a more specific bug / feature request related to aliased qualified import completions

this is not intended to be merged as is, as the second test is failing

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2198"></a>


pepeiborra:
I agree that there is a bug, and I suspect it's related to local modules. Could you break the failing test down into two: one for qualified local module imports and one for qualified package module imports?

jneira:
It seems the test fail is legitimate

pepeiborra:
I included these tests in #2332

PR #2200, opened 2021-09-16
Add gitlab CI

hasufell:
## Motivation

Building FreeBSD, Darwin aarch64 (M1), Linux armv7, Linux aarch64

Currently running pipeline looks good: https://gitlab.haskell.org/maerwald/haskell-language-server/-/pipelines/40810

## Questions

This will require creating a repo at https://gitlab.haskell.org under the [haskell](https://gitlab.haskell.org/haskell) namespace (I can do that). The main question is whether it should be a proper import (with all issues) or just a mirror to build releases.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2200"></a>


jneira:
+1 for a mirror, as i think most of maintainers are used to github (but lets see their opinion)

hasufell:
I initialized https://gitlab.haskell.org/haskell/haskell-language-server

It's not an automatic mirror. I invited several people as maintainers.

jneira:
> 
> 
> I initialized https://gitlab.haskell.org/haskell/haskell-language-server
> 
> It's not an automatic mirror. I invited several people as maintainers. Issue tracker, PRs etc are disabled.

great

jneira (approved):
LGTM but i would like to know the opinion from other maintainers @Ailrun @pepeiborra @michaelpj (and everyone else)

.gitlab/ci.sh
```diff
@@ -0,0 +1,51 @@
...
+		;;
+esac
+
+mkdir -p "$CABAL_DIR"
+mkdir -p "$GHCUP_BINDIR"
+export PATH="$GHCUP_BINDIR:$PATH"
+
+export BOOTSTRAP_HASKELL_NONINTERACTIVE=1
+export BOOTSTRAP_HASKELL_GHC_VERSION=$GHC_VERSION
+export BOOTSTRAP_HASKELL_CABAL_VERSION=$CABAL_INSTALL_VERSION
```
Ailrun, line 27: One minor nitpicking here: Could you wrap these in double quotes?(Though I think these don't contain any spaces and even so this is safe in most modern shells.)

Ailrun (approved):
I'm not an expert of neither of GitLab CI or nix stuff, but LGTM. Thank you for the CI improvement!

cabal.project
```diff
@@ -25,6 +25,8 @@ packages:
          ./plugins/hls-ormolu-plugin
          ./plugins/hls-call-hierarchy-plugin
 
+optional-packages: vendored/*/*.cabal
+
```
pepeiborra, line 29: what's this for?

.gitlab/shell.nix
```diff
@@ -0,0 +1,90 @@
+{ system ? "aarch64-darwin"
```
pepeiborra, line 1: I note that this nix script is only used for the aarch64-darwin config, therefore I would change the file name to `shell-aarch64-darwin.nix` or similar

pepeiborra:
> ## Motivation
> Building FreeBSD, Darwin aarch64 (M1), Linux armv7, Linux aarch64
> 
> Currently running pipeline looks good: https://gitlab.haskell.org/maerwald/haskell-language-server/-/pipelines/40810
> 

What is the ultimate purpose? To ensure that HLS builds in those platforms or to produce binaries for distribution?

hasufell:
> > ## Motivation
> > Building FreeBSD, Darwin aarch64 (M1), Linux armv7, Linux aarch64
> > 
> > Currently running pipeline looks good: https://gitlab.haskell.org/maerwald/haskell-language-server/-/pipelines/40810
> > 
> 
> What is the ultimate purpose? To ensure that HLS builds in those platforms or to produce binaries for distribution? 
> 

Building binaries. I already did and added them to GHCup manually.

cabal.project
hasufell: Being able to use vendored packages. If they don't exist, nothing happens. If they exist, they get picked up.

This is required because the network package is broken on M1 and needs a autoreconf (but cabal is unable to do so for you). So we do it manually and put it into this directory.

.gitlab/ci.sh
hasufell: It doesn't matter if they contain spaces. This is a variable assignment.

```sh
$ foo="abc def"
$ export bar=$foo
$ echo $bar
abc def
```

hasufell: But I think you're right, some exotic shells exhibit different behavior when `export` is in front of the assignment: https://unix.stackexchange.com/a/97569

Without the `export`, no shell needs the quotes.

pepeiborra (approved):

cabal.project
pepeiborra: it probably deserves a comment "Used by the M1 build script"

jneira: hmm it is a somewhat standard way to include potential optional packages in any situation but a comment with an actual use case would be fine

hasufell:
@jneira what's with those reverse merges?

jneira:
> 
> 
> @jneira what's with those reverse merges?

they are the automatic merges made by github if you press the button to update branch with master (or made by mergifyio bot when you label it with `merge me`)

cabal.project
michaelpj: If it's (evidently) not obvious to the most active maintainers of this project, it definitely needs a comment for the rest of us poor mortals :)

michaelpj:
> It's not an automatic mirror. I invited several people as maintainers. Issue tracker, PRs etc are disabled.

Who will keep it up to date? Are there expectations on people other than @hasufell to do anything at all with this repo? (If there are, we should write them down in the contributing section) Do we expect that the CI will always pass if the other CI passes? Could it fail if the mirror gets out of date? If so, how would we fix that?

Basically, I think it needs a section in the contributing guide that explains this stuff.

cabal.project
michaelpj: The comment could even point to the script where it's used!

hasufell:
> Who will keep it up to date?

Whoever does the release at the time of release preparation.

> Are there expectations on people other than @hasufell to do anything at all with this repo?

Yes, whoever is involved in cutting the release.

> Do we expect that the CI will always pass if the other CI passes?

I think not

> Could it fail if the mirror gets out of date? If so, how would we fix that?

Not sure I understand. You'd basically push the release tag (manually or automatic, that's up to you) and the release build gets triggered. It may require some further work if you want automatic tar'ing etc.

michaelpj:
> Not sure I understand. You'd basically push the release tag (manually or automatic, that's up to you) and the release build gets triggered. It may require some further work if you want automatic tar'ing etc.

I have no idea what most of this stuff is for - but probably neither does anyone else, which is why some documentation would be great :)

.gitlab/ci.sh
Ailrun: Yes, some ancient shells require double quotes here, but it's safe for modern shells, so as I said, it's just a nitpicking thing.

cabal.project
hasufell: You know, github has a feature to add code suggestions ;)

jneira: @hasufell i hope you dont mind i've added a pair of comments directly in your branch

jneira (approved):
LGTM, thanks for adding those archs to ci

PR #2203, opened 2021-09-17
Add missing config options in documentation

jneira:
* Added diagnosticsDebounceDuration, checkProject, checkParents
* Also add link to the issue about liquid haskell
* Change the tone of uding hie.yaml suggestion: the default should be let hls detect it auto

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2203"></a>


docs/configuration.md
```diff
@@ -40,13 +40,16 @@ This option obviously would not make sense for language servers for other langua
...
-- Format on imports (`haskell.formatOnImportOn`, default true): whether to format after adding an import
-- Diagnostics on change (`haskell.diagnosticsOnChange`, default true): (currently unused)
-- Completion snippets (`haskell.completionSnippetsOn`, default true): whether to support completion snippets
-- Liquid Haskell (`haskell.liquidOn`, default false): whether to enable Liquid Haskell support (currently unused until the Liquid Haskell support is functional again)
-- Hlint (`haskell.hlintOn`, default true): whether to enable Hlint support
+- Formatting provider (`haskell.formattingProvider`, default `ormolu`): what formatter to use; one of `floskell`, `ormolu`, `fourmolu`, `stylish-haskell`, or `brittany` (if compiled with the brittany plugin).
+- Format on imports (`haskell.formatOnImportOn`, default true): whether to format after adding an import.
+- Diagnostics on change (`haskell.diagnosticsOnChange`, default true): (currently unused).
+- Diagnostics debounce duration (`haskell.diagnosticsDebounceDuration`, default 350000 milliseconds).
```
jneira, line 46: Would like to add more info about this but i am not sure about its semantics

docs/configuration.md
```diff
@@ -40,13 +40,16 @@ This option obviously would not make sense for language servers for other langua
...
+- Completion snippets (`haskell.completionSnippetsOn`, default true): whether to support completion snippets.
+- Liquid Haskell (`haskell.liquidOn`, default false): whether to enable Liquid Haskell support (currently unused until the Liquid Haskell support is functional again, see <https://github.com/haskell/haskell-language-server/issues/367>).
+- Hlint (`haskell.hlintOn`, default true): whether to enable Hlint support.
 - Max completions (`haskell.maxCompletions`, default 40): maximum number of completions sent to the LSP client.
+- Check project (`haskell.checkProject`, default true): whether to typecheck the entire project on load. AS it is activate by default could drive to bad perfomance in large projects.
+- Check parents (`haskell.checkParents`, default `CheckOnSaveAndClose`): when to typecheck reverse dependencies of a file; one of `NeverCheck`, `CheckOnClose`, `CheckOnSaveAndClose`, or `AlwaysCheck`.
```
jneira, line 52: If it can affect performance it should be noted how

docs/configuration.md
pepeiborra: This isn't used anywhere, rather than documenting it we should delete it

pepeiborra (changes requested):

docs/configuration.md
```diff
@@ -40,14 +40,46 @@ This option obviously would not make sense for language servers for other langua
...
-- Format on imports (`haskell.formatOnImportOn`, default true): whether to format after adding an import
-- Diagnostics on change (`haskell.diagnosticsOnChange`, default true): (currently unused)
-- Completion snippets (`haskell.completionSnippetsOn`, default true): whether to support completion snippets
-- Liquid Haskell (`haskell.liquidOn`, default false): whether to enable Liquid Haskell support (currently unused until the Liquid Haskell support is functional again)
-- Hlint (`haskell.hlintOn`, default true): whether to enable Hlint support
+- Formatting provider (`haskell.formattingProvider`, default `ormolu`): what formatter to use; one of `floskell`, `ormolu`, `fourmolu`, `stylish-haskell`, or `brittany` (if compiled with the brittany plugin).
+- Format on imports (`haskell.formatOnImportOn`, default true): whether to format after adding an import.
+- Diagnostics on change (`haskell.diagnosticsOnChange`, default true): (currently unused).
+- Diagnostics debounce duration (`haskell.diagnosticsDebounceDuration`, default 350000 milliseconds).
```
jneira, line 46:
```suggestion
```

docs/configuration.md
jneira: deleting here for now

docs/configuration.md
pepeiborra: but can we delete it from the source code as well?

jneira: yeah, i was thinking in do it in another pr but here is fine

docs/configuration.md
```diff
@@ -40,14 +40,45 @@ This option obviously would not make sense for language servers for other langua
 
 Here is a list of the additional settings currently supported by `haskell-language-server`, along with their setting key (you may not need to know this) and default:
 
-- Formatting provider (`haskell.formattingProvider`, default `ormolu`): what formatter to use; one of `floskell`, `ormolu`, `fourmolu`, `stylish-haskell`, or `brittany` (if compiled with the brittany plugin)
-- Format on imports (`haskell.formatOnImportOn`, default true): whether to format after adding an import
-- Diagnostics on change (`haskell.diagnosticsOnChange`, default true): (currently unused)
-- Completion snippets (`haskell.completionSnippetsOn`, default true): whether to support completion snippets
-- Liquid Haskell (`haskell.liquidOn`, default false): whether to enable Liquid Haskell support (currently unused until the Liquid Haskell support is functional again)
-- Hlint (`haskell.hlintOn`, default true): whether to enable Hlint support
+- Formatting provider (`haskell.formattingProvider`, default `ormolu`): what formatter to use; one of `floskell`, `ormolu`, `fourmolu`, `stylish-haskell`, or `brittany` (if compiled with the brittany plugin).
```
michaelpj, line 43: I think these might be better as tables?

docs/configuration.md
```diff
@@ -40,14 +40,45 @@ This option obviously would not make sense for language servers for other langua
...
+
+Plugins have a generic config to control their behaviour. The schema of such config is:
+
+- `haskell.plugin.${pluginName}.globalOn`: usually with default true. Wheter the plugin is enabled at runtime or it is not. That is the option you might use if you want to disable completely a plugin.
+  - Actual plugin names are: `ghcide-code-actions-fill-holes`, `ghcide-completions`, `ghcide-hover-and-symbols`, `ghcide-type-lenses`, `ghcide-code-actions-type-signatures`, `ghcide-code-actions-bindings`, `ghcide-code-actions-imports-exports`, `eval`, `moduleName`, `pragmas`, `refineImports`, `importLens`, `class`, `tactics` (aka wingman), `hlint`, `haddockComments`, `retrie`, `splice`.
+  - So to disable the import lens with an explicit list of module definitions you could set `haskell.plugin.importLens.globalOn: false`
+- `haskell.plugin.${pluginName}.${lspCapability}On`: usually with default true. Whether a concrete plugin capability is enabled. 
+  - Capabilities are the different ways a lsp server can interact with the editor. The current available capabilities of the server are: `callHierarchy`, `codeActions`, `codeLens`, `diagnostics`, `hover`, `symbols`, `completion`, `rename`.
```
michaelpj, line 61: :exploding_head:

docs/configuration.md
```diff
@@ -40,14 +40,45 @@ This option obviously would not make sense for language servers for other langua
...
+    - `haskell.plugin.tactics.config.max_use_ctor_actions`, default 5: Maximum number of `Use constructor <x>` code actions that can appear.
+    - `haskell.plugin.tactics.config.proofstate_styling`, default true: Should Wingman emit styling markup when showing metaprogram proof states?
+  - `ghcide-completions`:
+    - `haskell.plugin.ghcide-completions.config.snippetsOn`, default true: Inserts snippets when using code completions.
+    - `haskell.plugin.ghcide-completions.config.autoExtendOn`, default true: Extends the import list automatically when completing a out-of-scope identifier.
+  - `ghcide-type-lenses`:
+    - `haskell.plugin.ghcide-type-lenses.config.mode`, default `always`: Control how type lenses are shown. One of `always`, `exported`, `diganostics`.
+  - `hlint`:
+    - `haskell.plugin.hlint.config.flags`, default empty: List of flags used by hlint.
+This reference of configuration can be outdated at any time but we can query the `haskell-server-executable` about what configuration is effectively used:
```
michaelpj, line 78: It would be nice if we had some way to auto-generate this stuff into the docs. Probably not worth the effort to do fully, but perhaps we could have a halfway house, like:

- Write a tool like `vscode-extension-schema` that prints out a markdown or RST table of all the options. 
- Commit the output to the repo. Regenerate it occasionally if we notice it's out of date.
- Include the generated table into the docs.

michaelpj (approved):

docs/configuration.md
pepeiborra: I insist on deleting it from code as well, otherwise future maintainers will be puzzled as well:

https://github.com/haskell/haskell-language-server/blob/d815d0474d226e00ba2e3f93f9878b109e1ea2e0/hls-plugin-api/src/Ide/Plugin/Config.hs#L55

docs/configuration.md
jneira: yeah, it will look cleaner

docs/configuration.md
jneira: a table here too?

docs/configuration.md
jneira: yeah, but I would do it manually for now in this pr, we could open an issue to track it

jneira: I was thinking in include the default config/JSON schema directly in docs for reference and link it here, it is easy to regenerate just with actual hls cli output and later we can process it to make it nicer.

docs/configuration.md
michaelpj: I like tables!

Also this is a good conceptual explanation, but for our users I don't think it's enough that they can know which options they can actually set - they'd have to know which plugins have which capabilities.

I think the best thing is just a big, dumb, table. Ideally auto-generated!

docs/configuration.md
michaelpj: Sure, but I don't think we should make our users read a JSON file. Either we'll be making a table by hand or we write something to do it :shrug:

docs/configuration.md
jneira: @pepeiborra removed!

pepeiborra (approved):

docs/configuration.md
jneira: agree, what do you think about merge this as is and improve it in other pr's?

michaelpj: Oh yes, definitely!

docs/configuration.md
```diff
@@ -40,14 +40,45 @@ This option obviously would not make sense for language servers for other langua
...
 - Max completions (`haskell.maxCompletions`, default 40): maximum number of completions sent to the LSP client.
-
+- Check project (`haskell.checkProject`, default true): whether to typecheck the entire project on load. As it is activated by default could drive to bad perfomance in large projects.
+- Check parents (`haskell.checkParents`, default `CheckOnSaveAndClose`): when to typecheck reverse dependencies of a file; one of `NeverCheck`, `CheckOnClose`, `CheckOnSaveAndClose`, or `AlwaysCheck`. 
+
+#### Generic plugin configuration
+
+Plugins have a generic config to control their behaviour. The schema of such config is:
+
+- `haskell.plugin.${pluginName}.globalOn`: usually with default true. Wheter the plugin is enabled at runtime or it is not. That is the option you might use if you want to disable completely a plugin.
```
jneira, line 57:
```suggestion
- `haskell.plugin.${pluginName}.globalOn`: usually with default true. Whether the plugin is enabled at runtime or it is not. That is the option you might use if you want to disable completely a plugin.
```

PR #2212, opened 2021-09-19
ghcide: Add flags to toggle building each executable

hololeap:
This package is a transient dependency of `haskell-language-server`.
However, only the library is needed. These three flags give the option for
those who do not need the executables to disable them.

`executable`:
* Toggles the `ghcide` executable
* enabled by default.

`test-exe`:
* Toggles the `ghcide-test-preprocessor` executable
* disabled by default

`bench-exe`:
* Toggles the `ghcide-bench` executable
* disabled by default

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2212"></a>

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -217,6 +217,10 @@ library
       exposed-modules:
         Development.IDE.GHC.Compat.CPP
 
+flag test-exe
+    description: Build the ghcide-test-preprocessor executable
+    default: True
+
 executable ghcide-test-preprocessor
     default-language: Haskell2010
     hs-source-dirs: test/preprocessor
@@ -224,6 +228,9 @@ executable ghcide-test-preprocessor
     main-is: Main.hs
     build-depends:
         base == 4.*
+
+    if !flag(test-exe)
+        buildable: False
 
 benchmark benchHist
     type: exitcode-stdio-1.0
@@ -262,6 +269,10 @@ benchmark benchHist
         shake,
         text,
         yaml
+
+flag executable
+    description: Build the ghcide executable
+    default: True
 
 executable ghcide
     default-language:   Haskell2010
@@ -321,6 +332,9 @@ executable ghcide
         TupleSections
         TypeApplications
         ViewPatterns
+
+    if !flag(executable)
+        buildable: False
 
 test-suite ghcide-tests
     type: exitcode-stdio-1.0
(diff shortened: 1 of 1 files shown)


pepeiborra:
I'm not supportive of this change as it is:
* The `test-exe` flag breaks the ghcide test suite
* The `bench-exe` flag breaks the ghcide benchmark suite
* The `executable` flag will break both the test and benchmark suites

I understand that you are coming at this from "why do I have to build all these executables if all I want is HLS". That's understandable, but I would only be supportive of a change that solves that problem if it has no negative impact on the ghcide development workflow.

hololeap:
I changed it so that all three flags have `default: True` in hopes this will not break the workflow. This seems like a good compromise to me since building the exe's can still be disabled from the command line.

pepeiborra (approved):
Thanks!

michaelpj:
I'm slightly missing the motivation. HLS doesn't depend on these, so if you're just building HLS then cabal won't build them. Is it so that you can run `cabal build all` and not have to build them?

jneira:
> 
> 
> I'm slightly missing the motivation. HLS doesn't depend on these, so if you're just building HLS then cabal won't build them. Is it so that you can run `cabal build all` and not have to build them?

hmm cabal builds executables component by default i think

hololeap:
@michaelpj @jneira Yes, cabal builds all executable components unless you create a mechanism like this to turn them off. I run into this when packaging for Gentoo, which strives to give the end user as much control over their packages as possible (and currently uses `cabal` under the hood). Without this patch I would have to either:

1. Give the user no control over whether the executables are installed, even if they don't want them
2. Maintain a patch for the `ghcide.cabal` file for every new release.

(Just so you know what my personal motivations are.)

michaelpj:
> Yes, cabal builds all executable components unless you create a mechanism like this to turn them off.

Well, perhaps the question becomes: how are you invoking cabal?

If you build the executable component for `haskell-language-server` it doesn't build the ghcide executables.
```
> cabal build exe:haskell-language-server
...
> vim ghcide/exe/Main.hs # make trivial edit
> cabal build exe:haskell-language-server
Up to date
```

hololeap:
@michaelpj The way we have been doing it is to convert each `.cabal` package into a Gentoo package, where the dependency graph is handled solely by our package manager. We then compile `Setup.hs` and invoke `setup` for the configure, build, test, etc. phases on each package individually, and install them to a global location. So each package is built in relative isolation and there is no current integration of the v2 cabal features.

PR #2215, opened 2021-09-19
Inline Text.Fuzzy to add INLINABLE pragmas

pepeiborra:
I found a noticeable perf regression in completions performance when trying to upgrade our custom Sigma IDE from HLS 1.2 to 1.4. The reason is that 1.3 added support for local completions from hiedb and that brings an additional >500k identifiers into scope, so the performance of fuzzy matching becomes a bottleneck.

The fuzzy package is parametric on the text type via the `TextualMonoid` type class. Specializing the `Fuzzy.filter` function to `Text` cuts the allocations by 66% and the time by ~33%.

I sent [a PR](https://github.com/joom/fuzzy/pull/3) upstream, but it might take a while to get merged. Since the license (MIT) allows for it, let's inline the code here and drop the dependency until then.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2215"></a>

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -50,7 +50,6 @@ library
         dlist,
         -- we can't use >= 1.7.10 while we have to use hlint == 3.2.*
         extra >= 1.7.4 && < 1.7.10,
-        fuzzy,
         filepath,
         fingertree,
         ghc-exactprint,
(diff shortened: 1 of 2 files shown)


jneira (approved):
I am fine with the temporary inclusion, with the appropiate comments

PR #2217, opened 2021-09-20
Return completions lazily for massive savings

pepeiborra:
This is the second completions performance issue (after #2215) uncovered by upgrading Sigma IDE from HLS 1.2 to 1.4. 

The problem is that `getCompletions` returns the completions strictly, only to have them cropped to `maxCompletions` in hls-plugin-api. This is bad because there is an O(nlogn) `nubOrdBy` in the result thunk, and it can take a second or more to nub >100k results.

Thankfully `nubOrdBy` is lazy in the result so the fix is simple. By restoring result laziness in `getCompletions`, only up to `maxCompletions` results are processed by `nubOrdBy`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2217"></a>


jneira (approved):
good catch

Issue #1408, opened 2021-02-20
Failed to run pre-commit hook


berberman:
I couldn't commit without specifying  `--no-verify`...

```
$ nix-shell
$ git commit -m "..."
stylish-haskell..........................................................Failed
- hook id: stylish-haskell
- exit code: 1

stylish-haskell: 0:0:Language.Haskell.Stylish.Config.loadConfig: fromJSON: expected Bool, but encountered String

CallStack (from HasCallStack):
  error, called at lib/Language/Haskell/Stylish/Config.hs:107:32 in stylish-haskell-0.11.0.3-F2KP2B6JiKw5YfjeWoLVgI:Language.Haskell.Stylish.Config
```

Ailrun:
Do you know what part of the config caused the problem? I cannot reproduce it at least only with `stylish-haskell` command line.

berberman:
Hmm `stylish-haskell` does run successfully from command line, but executing the hook manually would fail, even if I removed all `.stylish-haskell.yaml` files.

```
$ .git/hooks/pre-commit
stylish-haskell..........................................................Failed
- hook id: stylish-haskell
- exit code: 1

stylish-haskell: 0:0:Language.Haskell.Stylish.Config.loadConfig: fromJSON: expected Bool, but encountered String

CallStack (from HasCallStack):
  error, called at lib/Language/Haskell/Stylish/Config.hs:107:32 in stylish-haskell-0.11.0.3-F2KP2B6JiKw5YfjeWoLVgI:Language.Haskell.Stylish.Config
```

Ailrun:
That's weird, when I call `stylish-haskell` directly in the shell, I got `stylish-haskell 0.12.2.0`, but the hook executes `stylish-haskell-0.11.0.3`...

Ailrun:
I see, it's because `nix-pre-commit-hooks` uses its own source, and we need to somehow replace it with our source...

PR #2218, opened 2021-09-20
flake: remove the 'follows' directive

teto:
nixFlake suffers from a bug when using "follows" keyword:
https://github.com/NixOS/nix/issues/3602

I think the (possibly) extra-downloads is worth having the peace of mind that it always works.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2218"></a>


berberman (changes requested):
We have to make sure that formatters in pre-commit hooks are consistent (#1408), so I think the follow of nixpkgs is required. For flake-utils, I agree with removing that follow.

teto:
the "follow" is broken in nix which means I can't use the flake without a fork. 

Alternatively, could we :
- fork pre-commit-hook so that we can manually make sure  they use the same nixpkgs version as hls ?
- override stylish-haskell in the pre-commit flake with the one from hls' overlay ?

berberman:
> fork pre-commit-hook so that we can manually make sure they use the same nixpkgs version as hls

I don't think that's a good idea, since it causes extra maintenance

> override stylish-haskell in the pre-commit flake with the one from hls' overlay

Actually we already do this: https://github.com/haskell/haskell-language-server/blob/48221c7ed1db4dd992391a0b4ba1e304e05da5d9/flake.nix#L137

So maybe we don't need "follow" nixpkgs at all, and this PR may work. Could you try executing the pre-commit hook?

teto:
`stylish-haskell --version`  gives me `stylish-haskell 0.12.2.0`.
If I trigger the hook. It works correctly but not sure how to check it uses the same version.

berberman (approved):
Thanks for confirming, then we are good to go

PR #2221, opened 2021-09-20
feat(flake): expose hie-bios

teto:
since it is helpful to help troubleshoot hie.yaml issues and since the binaries are built anyway.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2221"></a>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -246,6 +246,11 @@
           haskell-language-server-8107 = mkExe ghc8107;
           haskell-language-server-901 = mkExe ghc901;
 
+          # hie-bios packages
+          hie-bios-884 = ghc884.hie-bios;
+          hie-bios-8107 = ghc8107.hie-bios;
+          hie-bios-901 = ghc901.hie-bios;
+
           # docs
           docs = docs;
         };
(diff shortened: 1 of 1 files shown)


jneira (approved):
lgtm!

berberman:
Thanks, that's a good point. Can we add hie-bios to the buildInputs of dev shell (I assume it should be used in command line), so that we don't need maintain the attributes like hie-bios-901 for each GHC version?

teto:
@berberman like this ?

flake.nix
```diff
@@ -246,6 +247,11 @@
           haskell-language-server-8107 = mkExe ghc8107;
           haskell-language-server-901 = mkExe ghc901;
 
+          # hie-bios packages
+          hie-bios-884 = ghc884.hie-bios;
+          hie-bios-8107 = ghc8107.hie-bios;
+          hie-bios-901 = ghc901.hie-bios;
+
```
berberman, line 254:
```suggestion
```

jneira:
@berberman thanks for reviewing it, good to merge now?

PR #2225, opened 2021-09-21
Parallel fuzzy filtering

pepeiborra:
Following #2215  and #2217 this is another performance improvement to completions driven by the large Sigma codebase.

On every completion attempt we need to match the prefix against all the identifiers in the project, with a cost of O(Identifiers*C) where C is the average number of characters per identifier. This PR parallelises the matching and uses vectors to reduce the amount of allocations and enable in-place sorting of the results.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2225"></a>

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -2,7 +2,7 @@ cabal-version:      2.4
 build-type:         Simple
 category:           Development
 name:               ghcide
-version:            1.4.2.0
+version:            1.4.2.1
 license:            Apache-2.0
 license-file:       LICENSE
 author:             Digital Asset and Ghcide contributors
@@ -64,6 +64,7 @@ library
         hiedb == 0.4.1.*,
         lsp-types >= 1.3.0.1 && < 1.4,
         lsp == 1.2.*,
+        monoid-subclasses,
         mtl,
         network-uri,
         optparse-applicative,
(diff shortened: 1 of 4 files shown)


ghcide/ghcide.cabal
```diff
@@ -2,7 +2,7 @@ cabal-version:      2.4
 build-type:         Simple
 category:           Development
 name:               ghcide
-version:            1.4.2.1
+version:            1.4.2.0
```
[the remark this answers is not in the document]
pepeiborra: No, must be a merge issue, good spot!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -53,6 +53,10 @@ import           Language.LSP.Types
 import           Language.LSP.Types.Capabilities
 import qualified Language.LSP.VFS                         as VFS
 
+-- Chunk size used for parallelizing fuzzy matching
+chunkSize :: Int
+chunkSize = 1000
```
jneira, line 58: does not need this be configured?

jneira (approved):
Not sure if there are existing tests which might be changed to cover changes, i let it to your consideration

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
pepeiborra: No, this is an implementation detail.

pepeiborra:
> Not sure if there are existing tests which might be changed to cover changes, i let it to your consideration

All the completions test suite cover these changes

Issue #2216, opened 2021-09-19
Create proper source tarballs in CI/on releases


hasufell:
Implementing https://gitlab.haskell.org/haskell/ghcup-hs/-/merge_requests/180 I need proper source tarballs.

The default ones are:

1. autogenerated and would depend on github not messing with the tarball layout (they did so at least twice) for the hash to be stable
2. depend on the tag (yeah, if you force push, the hash changes too)
3. have bad tarball names like `1.4.0.tar.gz`, causing trouble for ghcup

Something like:

```sh
git archive --prefix=haskell-language-server-${version}/ --format=tar.gz -o haskell-language-server-${version}.tar.gz HEAD
```

in CI should do the trick.

jneira:
@hasufell the mr is merged and released, how could improve it creating source tarballs here? are you using them from gitlab in the actual version?

hasufell:
ghcup is using the source tarballs

jneira:
> 
> 
> ghcup is using the source tarballs

the default ones from github? so improving them would allow clean ghcup code

hasufell:
> the default ones from github? so improving them would allow clean ghcup code

I don't understand what you mean. ghcup needs source tarballs to compile HLS from source. The automatically generated aren't good enough as I described in the first comment.

jneira:
> ghcup is using the source tarballs

hmm so if default ones are not valid and we are not generating in github the good ones yet, Where are the tarballs it is using now from?

hasufell:
> Where are the tarballs it is using now from?

Created on my own.

jneira:
thanks, will add a job in the build workflow to generate and add them to sha256 sums

PR #2230, opened 2021-09-22
Add ghcup compile option

### LINKED ISSUE #2204: Update freebsd installation instructions
* After having prebuilt binaries with #2200
* From #1769
//cc @arrowd @hasufell

jneira:
* closes #2204
//cc @hasufell @arrowd

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2230"></a>


docs/installation.md
```diff
@@ -9,14 +9,21 @@
 ## ghcup
 
 If you are using [`ghcup`](https://www.haskell.org/ghcup/) to manage your installations, you can install `haskell-language-server` with
-```
+```bash
 ghcup install hls
 ```
 
+Since ghcup-1.17.0 you can also install HLS from source without checking out the code manually:
```
hasufell, line 16:
```suggestion
Since ghcup-0.1.17 you can also install HLS from source without checking out the code manually:
```

docs/installation.md
```diff
@@ -9,14 +9,21 @@
...
+```bash
+ghcup compile hls -v 1.4.0.0 8.10.7
```
hasufell, line 18:
```suggestion
ghcup compile hls -v 1.4.0 8.10.7
```

docs/installation.md
```diff
@@ -50,13 +57,16 @@ See [ArchWiki](https://wiki.archlinux.org/index.php/Haskell) for the details of
...
 
-```
+```bash
 pkg install hs-haskell-language-server
 ```
 
 to install it. At the moment, HLS installed this way only supports the same GHC
 version as the ports one.
 
+Since 1.4.0 version you also can install HLS via [ghcup](#ghcup).
```
michaelpj, line 67: Nit, but I don't see much point in giving these kind of historical details here. Users probably care about the current state. If we want to provide docs for old versions we can do that with RTD's versioned documentation!

Also, does this add anything over the ghcup section? Nothing in that section says you *can't* use it on FreeBSD, and none of the other linux sections need to say "btw ghcup works", so... maybe it's fine? (or maybe ghcup *doesn't* work on FreeBSD?) Maybe we should just link to something in the ghcup docs about OS support if there is something like that?

jneira: hmm yeah, it sounds more like an announcement than a pemanent install instruction. But i wanted to stress you can use ghcup although it is not a linux flavour but a unix one. But same could be applicable for windows cause you can use ghcup there nowadays.

jneira: > Maybe we should just link to something in the ghcup docs about OS support if there is something like that?

That could be a good one

//cc @hasufell

docs/installation.md
```diff
@@ -9,14 +9,21 @@
 ## ghcup
 
 If you are using [`ghcup`](https://www.haskell.org/ghcup/) to manage your installations, you can install `haskell-language-server` with
-```
+```bash
 ghcup install hls
 ```
 
+Since ghcup-0.1.17 you can also install HLS from source without checking out the code manually:
```
jneira, line 16:
```suggestion
You can also install HLS from source without checking out the code manually:
```

Avoid hisotrical remarks

docs/installation.md
jneira:
```suggestion
```

michaelpj:
Aha, I found a platform support table for `gchup`! Maybe we should link to that? Or I guess we could just let users find it themselves if they wonder.

https://gitlab.haskell.org/haskell/ghcup-hs#supported-platforms

jneira:
> 
> 
> Aha, I found a platform support table for `gchup`! Maybe we should link to that? Or I guess we could just let users find it themselves if they wonder.
> 
> https://gitlab.haskell.org/haskell/ghcup-hs#supported-platforms

looks good, many thanks! i think it worths to link it here

docs/installation.md
```diff
@@ -9,14 +9,21 @@
 ## ghcup
 
 If you are using [`ghcup`](https://www.haskell.org/ghcup/) to manage your installations, you can install `haskell-language-server` with
-```
+```bash
 ghcup install hls
 ```
 
+You can also install HLS from source without checking out the code manually:
```
jneira, line 16:
```suggestion
You can check if HLS is available for your platorm via ghcup here: <https://gitlab.haskell.org/haskell/ghcup-hs#supported-platforms>.

You can also install HLS from source without checking out the code manually:
```

pepeiborra (approved):

jneira:
@michaelpj do you think this could be nerged?

michaelpj:
LGTM!

hasufell:
Sidenote, https://github.com/haskell/haskell-language-server/issues/2216 is relevant to this

jneira:
I think i ve tried 5 or so reruns and the test for windows still fails over and over (last time tactics for win and ghc-810.7)

pepeiborra:
> I think i ve tried 5 or so reruns and the test for windows still fails over and over (last time tactics for win and ghc-810.7)

The tactics suite is flaky because it runs tests in parallel and the tactics plugin has an internal timeout. We should probably run all the test suites sequentially since the CI runners only have 1 core anyway.

jneira:
hmm, but tests are already being run with `j1`, what do you mean with run sequentally? pass `j1` to cabal itself?

PR #2231, opened 2021-09-22
Add ghc deprecation policy to documentation

### LINKED ISSUE #2168: ghc support deprecation schedule

jneira:
* To make it definitive and public
* Closes #2168 and hopefully capture the outcome of the discussion established in

//cc @cdsmith @juhp @hasufell and everyone else!

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2231"></a>


docs/ghc-deprecation.md
```diff
@@ -0,0 +1,39 @@
...
+  - for the latest major GHC version we will support at least 2 minor versions
+  - for the rest of major supported GHC versions we will support at least the latest GHC minor version in stackage LTS (so 1 minor version)
+- We may extend the existing discovery mechanisms (hls-wrapper, automatic download in vscode extension) to find and download older HLS binaries when it detects a legacy GHC version
+- We will warn users about the deprecated ghc version in the notes of the release *prior* to the deprecation itself.
+  
+## Deprecation prevision
+
+| newest LTS ghc version |            supported ghc versions               |
+|------------------------|-------------------------------------------------|
+| 8.10.7                 | 8.10.7, 8.10.6, 8.10.5(\*), 8.8.4, 8.8.3, 8.6.5 |
```
hasufell, line 33: Why is `8.8.3` in there?

docs/ghc-deprecation.md
```diff
@@ -0,0 +1,39 @@
...
+## Deprecation policy
+
+- A GHC version is legacy if it is 3 or more major versions away from the newest stackage LTS ghc version
+- HLS will build on all non-legacy (major) versions of GHC,
+  - for the latest major GHC version we will support at least 2 minor versions
+  - for the rest of major supported GHC versions we will support at least the latest GHC minor version in stackage LTS (so 1 minor version)
+- We may extend the existing discovery mechanisms (hls-wrapper, automatic download in vscode extension) to find and download older HLS binaries when it detects a legacy GHC version
+- We will warn users about the deprecated ghc version in the notes of the release *prior* to the deprecation itself.
+  
+## Deprecation prevision
```
michaelpj, line 29: I think this is not the table that our users actually care about. Our users probably want to know which GHC versions are supported *now*, and how will that change in future.

Maybe that should be a separate table. I suggest something like this (content is random since I don't know off the top of my head!)


| GHC version | Supported | Future status | 
| -- | -- | -- |
| 9.2.* | Partial |  Full support in 1.4.0.0 |  
| 8.10.{7,6} | Y | Supported unitil some LTS thing happens | 
| 8.6.5 | Y | Will be deprecated in 1.4.0.0 |
| 7.10 | N | |


Then we can link to this policy to explain when support gets dropped.

docs/ghc-deprecation.md
```diff
@@ -0,0 +1,39 @@
+# GHC support deprecation policy
+
+- `haskell-language-server`(HLS) is highly tied to ghc api so much so that it needs a specific flavour for each ghc minor version to ensure it will work in a reliable way
+- It supposes the codebase is riddled with code conditioned to each supported ghc versions.  
+  - It even needs entire packages to fully support older versions of ghc.
```
michaelpj, line 5: I think this is a prose masquerading as a bullet list :)

docs/ghc-deprecation.md
jneira: Well it is the actual situation and the fact is we are supporting it and we didnt warn about its deprecation in the release notes for 1.4.0. Will add a `(*)`

michaelpj:
You need to reference this in `index.rst` or similar otherwise it won't appear in the documentation.

docs/ghc-deprecation.md
jneira:
```suggestion
| 8.10.7                 | 8.10.7, 8.10.6, 8.10.5(\*), 8.8.4, 8.8.3(\*), 8.6.5 |
```

jneira:
> You need to reference this in index.rst or similar otherwise it won't appear in the documentation.

Sure, thanks for noting it, done

docs/ghc-deprecation.md
jneira: hmm i am afraid that it will not depend on the hls version itself, only on stackage lts but i got the idea
do you think that separate table should live in that document or in other place more visible?

docs/ghc-deprecation.md
```diff
@@ -0,0 +1,39 @@
+# GHC support deprecation policy
+
+- `haskell-language-server`(HLS) is highly tied to ghc api so much so that it needs a specific flavour for each ghc minor version to ensure it will work in a reliable way
+- It supposes the codebase is riddled with code conditioned to each supported ghc versions.
```
jneira, line 4:
```suggestion
- It supposes the codebase is riddled with code conditioned to each supported ghc versions. It even needs entire packages to fully support older versions of ghc.
```

docs/ghc-deprecation.md
jneira:
```suggestion
```

docs/ghc-deprecation.md
```diff
@@ -0,0 +1,39 @@
+# GHC support deprecation policy
+
+- `haskell-language-server`(HLS) is highly tied to ghc api so much so that it needs a specific flavour for each ghc minor version to ensure it will work in a reliable way
+- It supposes the codebase is riddled with code conditioned to each supported ghc versions.  
+  - It even needs entire packages to fully support older versions of ghc.
+- Our continouos integracion setup has to cover all those cases so it have to use lot of resources to test and build the executable.
+- So we need to limit the ghc support to save maintainers and contributors time and reduce ci resources consumption.
```
jneira, line 7:
```suggestion
- So we need to limit the ghc support to save maintainers and contributors time and reduce ci resources consumption.  It becomes vital to make HLS development manageable.
```

docs/ghc-deprecation.md
```diff
@@ -0,0 +1,39 @@
+# GHC support deprecation policy
+
+- `haskell-language-server`(HLS) is highly tied to ghc api so much so that it needs a specific flavour for each ghc minor version to ensure it will work in a reliable way
+- It supposes the codebase is riddled with code conditioned to each supported ghc versions.  
+  - It even needs entire packages to fully support older versions of ghc.
+- Our continouos integracion setup has to cover all those cases so it have to use lot of resources to test and build the executable.
+- So we need to limit the ghc support to save maintainers and contributors time and reduce ci resources consumption.  
+  - It becomes vital to make HLS development manageable.
```
jneira, line 8:
```suggestion
```

docs/ghc-deprecation.md
hasufell: I thought we agreed that policy exceptions have no place in the policy. As such, we should remove `8.10.5` and `8.8.3` here and simply leave a note that this policy simply specifies the MINIMUM amount of GHC versions supported. Let's keep this clean.

hasufell:
I thought we agreed that policy exceptions have no place in the policy. As such, we should remove `8.10.5` and `8.8.3` here and simply leave a note that this policy simply specifies the MINIMUM amount of GHC versions supported. Let's keep this clean.

jneira:
> 
> 
> I thought we agreed that policy exceptions have no place in the policy. As such, we should remove `8.10.5` and `8.8.3` here and simply leave a note that this policy simply specifies the MINIMUM amount of GHC versions supported. Let's keep this clean.

Yeah but the policy itself is clean (`at least`...).
Whoever the prevision for the near future should be informative and reflect what is gonna to happen for real imo. Otoh it has the note:

` this prevision is not definitive and may vary depending on changes in the criteria listed above (but always honouring the policy)`

So we could decide to drop ghc-8.6.5 and "simply" update the prevision (although i would not change it)

Maybe we can move the table out of there as @michaelpj suggested and let the policy itself clean of facts.

docs/ghc-deprecation.md
```diff
@@ -0,0 +1,37 @@
...
+- To establish the policy and the possible exceptions we aim to take in account:
+   - completeness of support: the GHC flavour should include all plugins and features
+   - ghc versions supported by newer [stackage](https://www.stackage.org/) LTS's
+   - ghc versions supported by default in the most popular [linux distributions](https://repology.org/project/ghc/versions)
+   - the specific history of ghc releases and their realibility on the major operating systems (linux, windows, macos)
+- It worths to note that users of deprecated ghc versions always will have the option of:
+  - Continue using the last HLS with support for their ghc version. So they will miss new bug fixes and features.
+  - For deprecated minor versions of a still supported major version, try to build HLS from source.  
+    - We will not guarentee it will work but will do quite likely.
+
```
pepeiborra, line 17:
```suggestion
`haskell-language-server`(HLS) is highly tied to the ghc api.This imposes a high maintenance cost:
- The codebase is littered with conditional logic,
- We own auxiliary packages to support older versions of ghc.
- CI has to cover all the supported versions.

So we need to limit the ghc support to save maintainers and contributors time and reduce CI resources.

At same time we aim to support the right balance of ghc versions to minimize impact to final users.
To establish the policy we take into account:
   - completeness: support includes all plugins and features
   - The most recent [stackage](https://www.stackage.org/) LTS snapshot
   - The GHC versions used in the most popular [linux distributions](https://repology.org/project/ghc/versions)

It is worth noting that users always will have the option of:
  - using the last HLS with support for their ghc version.
  - build HLS from source.  
```

hasufell:
> Maybe we can move the table out of there as @michaelpj suggested and let the policy itself clean of facts.

Yes. Policy table is to help maintainers, not users. For users we can write a **support plan** somewhere else.

docs/ghc-deprecation.md
jneira: It looks cleaner, thanks, but why does it remove `the specific history of ghc releases and their realibility...`?
I think this is the real reason behind keep 8.6.5 and 8.10.5 and it is previsible that we should take that in account, even being an ambiguous consideration

jneira:
> 
> 
> > Maybe we can move the table out of there as @michaelpj suggested and let the policy itself clean of facts.
> 
> Yes. Policy table is to help maintainers, not users. For users we can write a **support plan** somewhere else.

Great, where could be a better place? A dedicated document?

jneira:
ghc release plan/history moved to installation, maybe not the better place

michaelpj:
I started having a look and I ended up with https://github.com/haskell/haskell-language-server/pull/2232, feel free to take any of it you like.

docs/ghc-deprecation.md
pepeiborra: Is it? I thought we are not making any exceptions and therefore not keeping 8.6.5, and for 8.10.5 the reason was a Linux distro.

I removed because it's poorly defined and likely to lead to disagreements, and I don't think it's needed.

jneira:
> 
> 
> I started having a look and I ended up with #2232, feel free to take any of it you like.

@michaelpj i prefer the structure of your version, so will close this directly. 
@pepeiborra do you mind translate your changes to the new pr?, i think the content is similar

pepeiborra:
> > 
> 
> > 
> 
> > I started having a look and I ended up with #2232, feel free to take any of it you like.
> 
> 
> 
> @pepeiborra do you mind translate your changes to the new pr?, i think the content is similar 

I'd prefer if you or Michael do it.

jneira:
Ops, michaelpj pr was against this

docs/ghc-deprecation.md
jneira: Well , we disagree here i think users will like to know we aim to be flexible and will listen their opinion about their use cases about ghc versioning. It is more a statement of principles than a definition.
Maintainers will have always the last word.

docs/supported-versions.md
```diff
@@ -0,0 +1,69 @@
...
+| ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------- |
+| 9.2.0       | not supported yet                                                                                                                                        |                                          |
+| 9.0.1       | [current](https://github.com/haskell/haskell-language-server/releases/latest) ([partial](https://github.com/haskell/haskell-language-server/issues/297)) |                                          |
+| 8.10.7      | [current](https://github.com/haskell/haskell-language-server/releases/latest)                                                                            |                                          |
+| 8.10.6      | [current](https://github.com/haskell/haskell-language-server/releases/latest)                                                                            | will be deprecated after LTS for ghc-9.0 |
```
Ailrun, line 12: Which one will come first, full 9.0.1 support or stackage LTS for 9.0? If it's the former, I think it's maybe better to mention that one instead of this, or both.

(In "both" case, we may say
> will be deprecated after complete support of GHC 9.0 as mentioned in this section

with link to the last section)

Ailrun (approved):
I love how version deprecating factors are described. Thank you for the awesome work!

docs/supported-versions.md
jneira: Finally i used `will be deprecated after LTS and HLS full support for..`, hope it is complete enough

jneira:
> 
> 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > I started having a look and I ended up with #2232, feel free to take any of it you like.
> > 
> > 
> > @pepeiborra do you mind translate your changes to the new pr?, i think the content is similar
> 
> I'd prefer if you or Michael do it.

Ok, i've included your suggestion, introducing again a rephrased mention to reliability of ghc versions on oss.
I think the intent is similar to `The GHC versions used in the most popular linux distributions` but make it more inclusive.

jneira:
this needs clearly a consensus between maintainers so i would ask you to approve explicitly it or expose the cocerns about the actual version, thanks all!

docs/supported-versions.md
michaelpj: Hmmmm, sounds like we need a term for "a fully supported major version of GHC with a LTS snapshot". That's also important for defining the major version support window. Perhaps the "mainline" GHC version? Then we could say "will be deprecated when 9.0 becomes mainline"?

michaelpj (approved):

pepeiborra (approved):

PR #2234, opened 2021-09-22
Fix build with fbghc

pepeiborra:
Just need to use the pattern synonym for `ExposePackage` defined in `Compute.Core`

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2234"></a>


fendor (approved):
LGTM, just need to fix the imports

PR #2235, opened 2021-09-23
Enable 3 tests for highlighting references and going to definitions.

peterwicksstringfield:
The test for ghcide.Development.IDE.Spans.AtPoint.referencesAtPoint doesn't support the "includeDeclaration" parameter, so that test is marked expect fail.

And the tests for goto definition that go from one module to another seem kinda flaky ... presumably because the other module isn't necessarily getting loaded in time for the test to succeed? I just deleted those tests. I don't think having them is worth making the CI flakey again.


jneira:
> I just deleted those tests. I don't think having them is worth making the CI flakey again.

Hmm i think it worths to keep them if they work sometimes and mark them as ignored by that reason. I still hope someone someday will have time to fix them :slightly_smiling_face:

test/functional/Definition.hs
```diff
@@ -2,63 +2,19 @@ module Definition (tests) where
...
 import           Test.Hls
 import           Test.Hls.Command
 
 tests :: TestTree
-tests = testGroup "definitions" [
-
-    ignoreTestBecause "Broken: file:///Users/jwindsor/src/haskell-language-server/test/testdata/References.hs" $
-       testCase "goto's symbols" $ runSession hlsCommand fullCaps "test/testdata" $ do
-        doc <- openDoc "References.hs" "haskell"
-        defs <- getDefinitions doc (Position 7 8)
```
pepeiborra, line 15: you probably need to insert a `waitForAllProgressDone` call after opening the first document

PR #2240, opened 2021-09-24
Partial sort of fuzzy filtering results

pepeiborra:
This improves the performance of completions by around 25% on a large project (>500k identifiers) and also removes the use of unsafe primitives in the previous in-place solution

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2240"></a>

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -2,7 +2,7 @@ cabal-version:      2.4
 build-type:         Simple
 category:           Development
 name:               ghcide
-version:            1.4.2.2
+version:            1.4.2.3
 license:            Apache-2.0
 license-file:       LICENSE
 author:             Digital Asset and Ghcide contributors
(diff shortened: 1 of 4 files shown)


jhrcek:
I don't know the sizes these vectors can reach, but if this is such performance critical piece of code,
would it make make sense using Vector [slice](https://hackage.haskell.org/package/vector-0.12.3.1/docs/Data-Vector.html#v:slice) instead of list comprehensions in the code you wrote previously?

https://github.com/haskell/haskell-language-server/pull/2225/files#diff-2930c3bdb9fee287ddd29a4f2db79ec551b5641b99b5de8e6b2e1ba4faac9779R87-R88

pepeiborra:
> I don't know the sizes these vectors can reach, but if this is such performance critical piece of code,
> would it make make sense using Vector [slice](https://hackage.haskell.org/package/vector-0.12.3.1/docs/Data-Vector.html#v:slice) instead of list comprehensions in the code you wrote previously?

Good point, `slice` instead of `fromListN` saves another 2%. Win!

jneira (approved):
LGTM, thanks for your work in improve perf

PR #2241, opened 2021-09-24
Let cabal-ghc901.project be cabal.project.local

jneira:
* The intent of the change is let users do `cp cabal-ghc901.project cabal.project.local` to make the workflow smoother
  * but the file can be used in its own as before
* To do so the packages should be uncommented (as they are so in main cabal.project) and the packages without support for ghc-9 should be make buildable with `allow-newer` (thanks to @gbaz for the suggestion)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2241"></a>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
-# Used for ci setup in the gitlab mirror of the project: 
+# Used for ci setup in the gitlab mirror of the project:
 # https://gitlab.haskell.org/haskell/haskell-language-server/-/pipelines
 variables:
   # Commit of ghc/ci-images repository from which to pull Docker images
@@ -9,21 +9,17 @@ variables:
 .default_matrix: &default_matrix
   matrix:
     - GHC_VERSION: 8.8.4
-      CABAL_PROJECT: cabal.project
     - GHC_VERSION: 8.10.7
-      CABAL_PROJECT: cabal.project
     - GHC_VERSION: 9.0.1
-      CABAL_PROJECT: cabal-ghc901.project
+      LOCAL_CABAL_PROJECT: cabal-ghc901.project.local
 
 .m1_matrix: &m1_matrix
   matrix:
     - GHC_VERSION: 8.10.7
-      CABAL_PROJECT: cabal.project
 
 .arm_matrix: &arm_matrix
   matrix:
     - GHC_VERSION: 8.10.7
-      CABAL_PROJECT: cabal.project
 
 
 workflow:
@@ -93,7 +89,6 @@ build-aarch64-darwin:
         --keep CI_PROJECT_DIR \
         --keep MACOSX_DEPLOYMENT_TARGET \
         --keep GHC_VERSION \
-        --keep CABAL_PROJECT \
         --keep CABAL_INSTALL_VERSION \
         --run "$1" 2>&1
     }
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -89,6 +89,7 @@ build-aarch64-darwin:
         --keep CI_PROJECT_DIR \
         --keep MACOSX_DEPLOYMENT_TARGET \
         --keep GHC_VERSION \
+        --keep LOCAL_CABAL_PROJECT \
         --keep CABAL_INSTALL_VERSION \
         --run "$1" 2>&1
     }
--- a/.gitlab/ci.sh
+++ b/.gitlab/ci.sh
@@ -30,6 +30,10 @@ export BOOTSTRAP_HASKELL_ADJUST_CABAL_CONFIG=yes
 
 curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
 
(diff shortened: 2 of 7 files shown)


pepeiborra:
Can we just replace cabal.project with cabal-ghc901.project ?

jneira:
> 
> 
> Can we just replace cabal.project with cabal-ghc901.project ?

🤔 hmm i had not thought in this possibility, but it would be right one if it is possible, will try, thanks!

jneira:
Possible caveats:
- Source repository packages not compatible with older versions of ghc. I think it is not the case
- Allow newers causing build errors
- Allow newer shadowing errors wich will arise when the version will be released in hackage

But it worths to investigate

jneira:
> 
> 
> Can we just replace cabal.project with cabal-ghc901.project ?

But maybe we could merge this as is? merge both cabal.project is not trivial

jneira:
Other alternative is convert cabal-ghc901.project in cabal-ghc901.local and remove duplication

jneira:
> 
> 
> Other alternative is convert cabal-ghc901.project in cabal-ghc901.local and remove duplication

I think it will be the best option, implementing

pepeiborra:
> > 
> 
> > 
> 
> > Other alternative is convert cabal-ghc901.project in cabal-ghc901.local and remove duplication
> 
> 
> 
> I think it will be the best option, implementing

This might break ghcup /cc @maerwald

.gitlab-ci.yml
```diff
@@ -93,7 +89,6 @@ build-aarch64-darwin:
         --keep CI_PROJECT_DIR \
         --keep MACOSX_DEPLOYMENT_TARGET \
         --keep GHC_VERSION \
-        --keep CABAL_PROJECT \
```
jneira, line 96: @hasufell not sure what has to be done here to take in account LOCAL_CABAL_PROJECT

hasufell: `--keep LOCAL_CABAL_PROJECT`

.gitlab/ci.sh
```diff
@@ -30,6 +30,10 @@ export BOOTSTRAP_HASKELL_ADJUST_CABAL_CONFIG=yes
 
 curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
 
+if [[ ! -z "$LOCAL_CABAL_PROJECT" ]]; then
```
hasufell, line 33:
```suggestion
if [[ -n "$LOCAL_CABAL_PROJECT" ]]; then
```

.gitlab/ci.sh
```diff
@@ -30,6 +30,10 @@ export BOOTSTRAP_HASKELL_ADJUST_CABAL_CONFIG=yes
...
+    cp $LOCAL_CABAL_PROJECT cabal.project.local
```
hasufell, line 34:
```suggestion
    run cp "$LOCAL_CABAL_PROJECT" cabal.project.local
```

.gitlab-ci.yml
hasufell: Also, this will break the ci.sh script which still uses this variable.

hasufell (changes requested):
`ci.sh` needs to be fixed

jneira:
Testing gitlab here: https://gitlab.haskell.org/jneira/haskell-language-server/-/jobs/796348

jneira:
@hasufell finally i've kept the CABAL_PROJECT in the script, maybe it is neededin the future

jneira:
I ha

> 
> 
> Testing gitlab here: https://gitlab.haskell.org/jneira/haskell-language-server/-/jobs/796348

```
.gitlab/ci.sh: line 33: CABAL_PROJECT: unbound variable
Cleaning up file based variables 00:01
ERROR: Job failed: exit code 
```

ugh

.gitlab/ci.sh
```diff
@@ -30,6 +30,14 @@ export BOOTSTRAP_HASKELL_ADJUST_CABAL_CONFIG=yes
 
 curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
 
+if [[ -n "$CABAL_PROJECT" ]]; then
+    CABAL_PROJECT="cabal.project"
+fi
```
jneira, line 35:
```suggestion
```

jneira:
> 
> 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > Other alternative is convert cabal-ghc901.project in cabal-ghc901.local and remove duplication
> > 
> > 
> > I think it will be the best option, implementing
> 
> This might break ghcup /cc @maerwald

do you refer to `ghcup compile hls`?

jneira:
it seems ci is ignoring the copied `cabal.project.local` and it is trying to build the disabled plugins anyways

jneira:
Not sure if the deduplication worths change the status quo: the gitlab script assumes this is the way we give support for new ghc's nd `ghcup compile hls` has `--cabal-project` to use it too (you would have to download manually the cabal-ghcx.project.local)
So i am gonna close it, sorry for the time spend on this

PR #2243, opened 2021-09-26
Enable the ghcide test plugin in HLS test suites

pepeiborra:
The ghcide test plugin installs a `CustomMethod` handler that allows to interact with the build queue and do things like waiting for the queue to be empty. It's installed by the ghcide main driver but not by the HLS test driver. 

This PR refactors things so that it will be installed automatically by the `defaultMain` driver.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2243"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Test.hs
+++ b/ghcide/src/Development/IDE/Plugin/Test.hs
@@ -50,11 +50,9 @@ newtype WaitForIdeRuleResult = WaitForIdeRuleResult { ideResultSuccess::Bool}
 
 plugin :: PluginDescriptor IdeState
 plugin = (defaultPluginDescriptor "test") {
-    pluginHandlers = mkPluginHandler (SCustomMethod "test") (\st pId ->
-        if pId == "test"
-            then testRequestHandler' st
-            else const $ return $ Left $ ResponseError MethodNotFound "" Nothing)
-}
+    pluginHandlers = mkPluginHandler (SCustomMethod "test") $ \st _ ->
+        testRequestHandler' st
+    }
   where
       testRequestHandler' ide req
         | Just customReq <- parseMaybe parseJSON req
--- a/ghcide/src/Development/IDE/Plugin/Test.hs
+++ b/ghcide/src/Development/IDE/Plugin/Test.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DeriveAnyClass     #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GADTs              #-}
+{-# LANGUAGE PolyKinds          #-}
 -- | A plugin that adds custom messages for use in tests
 module Development.IDE.Plugin.Test
   ( TestRequest(..)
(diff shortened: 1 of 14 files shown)


ghcide/src/Development/IDE/Plugin/Test.hs
```diff
@@ -50,10 +48,12 @@ data TestRequest
...
     deriving newtype (FromJSON, ToJSON)
 
-plugin :: P.Plugin c
-plugin = def {
-    P.pluginRules = return (),
-    P.pluginHandlers = requestHandler (SCustomMethod "test") testRequestHandler'
+plugin :: PluginDescriptor IdeState
+plugin = (defaultPluginDescriptor "test") {
+    pluginHandlers = mkPluginHandler (SCustomMethod "test") (\st pId ->
+        if pId == "test"
```
berberman, line 54: Why do we need check `pId` here? I think it should always be `"test"` which we pass to `defaultPluginDescriptor`.

pepeiborra: That's a good point, hls-plugin-api already takes care of this

berberman (approved):

jneira:
Not sure what dark spell made #2060 pass ci (:stuck_out_tongue_closed_eyes:) but it is failing since then more frequently, for tactics and rename plugin.
I tested manually in my machine renaming works fine with master after #2060 but i've not run tests yet.

jneira:
> The tests run all in parallel on the same workspace, and most of the modules have the same module name Main. Result: parallel writes to the same .hie files in the local tmp folder that lead to IO errors in Windows

thanks, i was about to make optional the rename test suite for windows and ghc-9.0.1

jneira:
eval plugin tests failed for linux int two consecutive runs

pepeiborra:
My recent commit broke it - still working on it

.github/workflows/test.yml
```diff
@@ -30,6 +30,8 @@ jobs:
   test:
     needs: pre_job
     runs-on: ${{ matrix.os }}
+    env:
+      TASTY_NUM_THREADS: 1
```
jneira, line 34: this has a different effect from pass `-1` to the test executable?

jneira: it seems the env var is equivalent to the arg, if docs are correct: https://github.com/UnkindPartition/tasty#runtime

pepeiborra: It is equivalent, but I don't think it worked

jneira: but we are already passing `-j1` to tasty via `--test-options` in the workflow, do you suspect the cli arg could have a bug?

pepeiborra: Oh you are right, I missed that `-j1` is already there

jneira: it is a shot in the dark but i am trying with `-j1` directly for cabal

pepeiborra:
The problem with the tactics test suite was also Main modules. Fixed it in my last commit, this PR should be getting merged very soon.

jneira:
> The problem with the tactics test suite was also Main modules. Fixed it in my last commit, this PR should be getting merged very soon. 

many many thanks, in my last try almost all tactic tests failed with session timeouts

Issue #2253, opened 2021-10-04
hls-eval-plugin test error in ubuntu related with interactive linking


jneira:
* there had been two builds failing in the eval plugin with:

```
Evaluation of expressions:                                                                          FAIL (1.55s)
    Test output was different from 'test/testdata/T1.expected.hs'. Output of ["git","-c","core.fileMode=false","diff","--no-index","--text","--exit-code","test/testdata/T1.expected.hs","/tmp/T1.expected84426-112.actual"]:
    diff --git a/test/testdata/T1.expected.hs b/tmp/T1.expected84426-112.actual
    index 622b9c1..8053d04 100644
    --- a/test/testdata/T1.expected.hs
    +++ b/tmp/T1.expected84426-112.actual
    @@ -3,6 +3,16 @@ module T1 where
     import Data.List (unwords)
     
     -- >>> unwords example
    --- "This is an example of evaluation"
    +-- <BLANKLINE>
    +-- ByteCodeLink.lookupCE
    +-- During interactive linking, GHCi couldn't find the following symbol:
    +--   interactive_Ghci1_evalPrint_closure
    +-- This may be due to you not asking GHCi to load extra object files,
    +-- archives or DLLs needed by your current session.  Restart GHCi, specifying
    +-- the missing library using the -L/path/to/object/dir and -lmissinglibname
    +-- flags, or simply by naming the relevant files on the GHCi command line.
    +-- Alternatively, this link failure might indicate a bug in GHCi.
    +-- If you suspect the latter, please send a bug report to:
    +--   glasgow-haskell-bugs@haskell.org
     example :: [String]
     example = ["This","is","an","example","of","eva<truncated>
    Use --accept or increase --size-cutoff to see full output.
    Use -p '$0=="eval.Evaluation of expressions"' to rerun this test only.
```

for [ghc-8.8.4](https://github.com/haskell/haskell-language-server/runs/3787034417?check_suite_focus=true) 

Maybe related with the recent update of github vms: https://github.com/actions/virtual-environments/releases/tag/ubuntu18%2F20210929.1

```
Haskell
    Cabal 3.6.0.0
    GHCup 0.1.17.1
```

Worths try updating the cache?

jneira:
One more using ghc-9.0.1: https://github.com/haskell/haskell-language-server/runs/3787949834?check_suite_focus=true

jneira:
[...]

PR #2254, opened 2021-10-04
Bump Fourmolu to 0.4

georgefst:
Addresses the relevant checklist item in #297.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2254"></a>

### DIFF
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               hls-fourmolu-plugin
-version:            1.0.0.2
+version:            1.0.1.0
 synopsis:           Integration with the Fourmolu code formatter
 description:
   Please see the README on GitHub at <https://github.com/haskell/haskell-language-server#readme>
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -19,6 +19,7 @@ extra-source-files:
 library
   exposed-modules:  Ide.Plugin.Fourmolu
   hs-source-dirs:   src
+  ghc-options:      -Wall
   build-depends:
     , base            >=4.12 && <5
     , filepath
@@ -38,7 +39,7 @@ test-suite tests
   default-language: Haskell2010
   hs-source-dirs:   test
   main-is:          Main.hs
-  ghc-options:      -threaded -rtsopts -with-rtsopts=-N
+  ghc-options:      -Wall -threaded -rtsopts -with-rtsopts=-N
   build-depends:
     , base
     , filepath
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -22,7 +22,7 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3
+    , fourmolu        ^>=0.3 || ^>=0.4
     , ghc
     , ghc-boot-th
     , ghcide          >=1.2  && <1.5
--- a/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
+++ b/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
@@ -8,19 +8,19 @@ module Ide.Plugin.Fourmolu (
     provider,
 ) where
 
-import           Control.Exception           (try)
-import           Control.Lens                ((^.))
+import           Control.Exception               (try)
+import           Control.Lens                    ((^.))
 import           Control.Monad.IO.Class
-import           Data.Bifunctor              (first)
-import qualified Data.Text                   as T
-import           Development.IDE             hiding (pluginHandlers)
-import           Development.IDE.GHC.Compat  as Compat hiding (Cpp)
+import           Data.Bifunctor                  (first)
+import qualified Data.Text                       as T
+import           Development.IDE                 hiding (pluginHandlers)
+import           Development.IDE.GHC.Compat      as Compat hiding (Cpp)
 import qualified Development.IDE.GHC.Compat.Util as S
-import           GHC.LanguageExtensions.Type (Extension (Cpp))
-import           GhcPlugins                  (HscEnv (hsc_dflags))
-import           Ide.PluginUtils             (makeDiffTextEdit)
+import           GHC.LanguageExtensions.Type     (Extension (Cpp))
+import           GhcPlugins                      (HscEnv (hsc_dflags))
+import           Ide.PluginUtils                 (makeDiffTextEdit)
 import           Ide.Types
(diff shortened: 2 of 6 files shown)


plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -8,19 +8,19 @@ module Ide.Plugin.Fourmolu (
...
 import           Control.Monad.IO.Class
-import           Data.Bifunctor              (first)
-import qualified Data.Text                   as T
-import           Development.IDE             hiding (pluginHandlers)
-import           Development.IDE.GHC.Compat  as Compat hiding (Cpp)
+import           Data.Bifunctor                  (first)
+import qualified Data.Text                       as T
+import           Development.IDE                 hiding (pluginHandlers)
+import           Development.IDE.GHC.Compat      as Compat hiding (Cpp)
 import qualified Development.IDE.GHC.Compat.Util as S
```
georgefst, line 18: I don't know how this file was passing CI before? Seeing as this line causes `stylish-haskell` to reformat all the other imports (I personally detest this behaviour, but that's an argument for another day).

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -22,7 +22,7 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3
+    , fourmolu        ^>=0.3 || ^>=0.4
```
georgefst, line 25: Ideally, I'd remove 0.3 entirely, as 0.4 includes some important bug fixes. But this isn't possible while `stylish-haskell` is tied to older versions of GHC and Cabal.

(Same applies to Ormolu here - right now, the latest version will never be chosen by the dependency solver)

jneira:
@georgefst many thanks to spend time on https://github.com/fourmolu/fourmolu/issues/104 and this, getting support for ghc-9

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
georgefst: Actually, looks like this only applies to GHC < 9. The `stylish-haskell` plugin is disabled for 9.0.

jneira:
I hope you dont mind i enabled fourmolu for ghc-9 to give it a shot inci

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
jneira: Maybe adding a comment about, to remove 0.3 some day?

georgefst:
> @georgefst many thanks to spend time on [fourmolu/fourmolu#104](https://github.com/fourmolu/fourmolu/issues/104) and this, getting support for ghc-9

Thanks, but @brandonchinn178 has actually done most of the hard work recently.

jneira (approved):
lgtm i hope fourmolu will continue rocking in the formatter side

jneira:
There is a build error for ghc-9:

```
 src/Ide/Plugin/Fourmolu.hs:20:1: error:
Error:     Could not find module ‘GhcPlugins’
    Perhaps you meant GHC.Plugins (from ghc-9.0.1)
   |
20 | import           GhcPlugins                      (HscEnv (hsc_dflags))
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

georgefst:
> There is a build error for ghc-9:
> 
> ```
>  src/Ide/Plugin/Fourmolu.hs:20:1: error:
> Error:     Could not find module ‘GhcPlugins’
>     Perhaps you meant GHC.Plugins (from ghc-9.0.1)
>     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
>    |
> 20 | import           GhcPlugins                      (HscEnv (hsc_dflags))
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

That import should have been removed in ec53fcb7b54434d62901a73bbf3e3085a579c487. Fixed in 46d3264.

This would have been caught by `-Wall`, but for some reason that isn't enabled for many of the plugins?

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
georgefst: > Maybe adding a comment about, to remove 0.3 some day?

Actually, I think there's no need. I'll probably end up at least trying to remove it every time I bump the upper bound, and eventually it'll be possible.

jneira:
> This would have been caught by -Wall, but for some reason that isn't enabled for many of the plugins?

Jumm they should afaiu, maybe is cargo culting and it is enabled for hls itself, maybe it is time to give a try at least here

jneira (commented):
what about `ghc-options:      -Wall -Wredundant-constraints`, like hls itself?

(without `-Wno-name-shadowing -Wno-unticked-promoted-constructors` if this does not need it)

georgefst:
> `-Wredundant-constraints`

Meh, I think there are legitimate use cases for redundant constraints, though it's very unlikely to come up in this plugin anyway.
 
> `-Wno-name-shadowing`

I prefer to be warned

> `-Wno-unticked-promoted-constructors`

No strong opinion, but again unlikely to come up.

On the whole, I'd go for keeping things simple with just `-Wall`. But I don't feel strongly about any of it.

jneira:
> 
> 
> > `-Wredundant-constraints`
> 
> Meh, I think there are legitimate use cases for redundant constraints, though it's very unlikely to come up in this plugin anyway.
> 
> > `-Wno-name-shadowing`
> 
> I prefer to be warned.
> 
> > `-Wno-unticked-promoted-constructors`
> 
> No strong opinion, but again unlikely to come up.
> 
> On the whole, I'd go for keeping things simple with just `-Wall`. But I don't feel that strongly about any of it.

Agree with you, i forgot to mention `-Wincomplete-uni-patterns` (it is in ghcide but no hls, ew *have* to unify all falgs :sweat:)

georgefst:
> it is in ghcide but no hls, we _have_ to unify all flags for sure 😓

Yeah, the fact that I know of no good way to do this in a multi-cabal-package project (`dhall-to-cabal` with flags etc. defined in a single file? probably a totally uncharted path and almost certainly more effort than it's worth right now) is why I suggest keeping it simple.

georgefst:
CI test failure looks unrelated to Fourmolu...

jneira:
> 
> 
> CI test failure looks unrelated to Fourmolu...

Yeah, all pr's are affected by #2253 :cry:

georgefst:
> > CI test failure looks unrelated to Fourmolu...
> 
> Yeah, all pr's are affected by #2253 😢

Ah, ok.

jneira:
> 
> 
> > > CI test failure looks unrelated to Fourmolu...
> > 
> > 
> > Yeah, all pr's are affected by #2253 cry
> 
> Ah, ok.

I was wrong, not all pr's are affected only those which are updating the cabal.project index state so maybe a new dep version or the cache invalidation is triggering it

PR #2257, opened 2021-10-05
Add more features and demos in docs

jneira:
* Copied from https://github.com/haskell/vscode-haskell#features

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -1,36 +1,64 @@
 # Features
 
-- Code evaluation codelens ([Tutorial](https://github.com/haskell/haskell-language-server/tree/master/plugins/hls-eval-plugin/README.md)):
+- Warning and error diagnostics from GHC
+- Type information and documentation on hover, [including your own comments](#how-to-show-local-documentation-on-hover).
+- Jump to definition: for now only between local code definitions
+- Document symbols
+- Highlight references in document
+- Code completion
+- Formatting via Brittany, Floskell, Fourmolu, Ormolu or Stylish Haskell
+- Code evaluation (Haskell Language Server), see ([Tutorial](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/README.md))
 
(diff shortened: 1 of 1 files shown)


docs/features.md
```diff
@@ -1,36 +1,64 @@
...
+  ![Module Name Demo](https://user-images.githubusercontent.com/54035/110860755-78ad8680-82bd-11eb-9845-9ea4b1cc1f76.gif)
 
-  This will cause compilation errors if a dependency contains invalid Haddock markup, though from GHC version 9.0, [these will be demoted to warnings](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2377).
+- Call hierarchy support
 
- - Integration with [retrie](https://hackage.haskell.org/package/retrie)
+  ![Call Hierarchy in VSCode](https://github.com/haskell/haskell-language-server/raw/2857eeece0398e1cd4b2ffb6069b05c4d2308b39/plugins/hls-call-hierarchy-plugin/call-hierarchy-in-vscode.gif)
+  
+  
+## How to show local documentation on hover
```
michaelpj, line 39: This doesn't seem like it belongs in a "features" list per se? It's more like "setup instructions to get this feature to work". I'm not sure where that belongs, perhaps in `Configuration`?

michaelpj (approved):
Nice! In the long run I think it would be cool to have a page per plugin which explains their features, has demos etc, and then this could link out to all those? Otherwise in the long run I think it's going to be hard for us to keep track of all our features!

jneira:
> 
> 
> Nice! In the long run I think it would be cool to have a page per plugin which explains their features, has demos etc, and then this could link out to all those? Otherwise in the long run I think it's going to be hard for us to keep track of all our features!

Yeah that is the plan outlined here: https://github.com/haskell/haskell-language-server/issues/2066, will update it

docs/features.md
jneira: done!

PR #2263, opened 2021-10-08
Garbage collection of dirty keys

pepeiborra:
This runs a garbage collection process when a file is closed, releasing all the keys that remain dirty after a build of all the open files. Essential keys are not GC'ed to avoid information loss and inefficient rebuilds. 

For example, if we have A, B and C open in the IDE and close A, the garbage collection will release all the keys that were specific to A, and preserve all the keys that are still needed by B and C. 

I have been able to validate heap usage reductions after closing all the files in the editor and triggering a keys GC followed by a heap GC. So it does seem to do what we want. 

TODO

- [x] Test
- [x] Measure improvement
- [x] Restore heap tracing code
- [x] Fix flaky tests
- [x] Think a bit more about GC strategy
- [x] GC seems to break diagnostics
- [x] Check plugins with custom rules
- [x] Break down in smaller PRs

EDIT: moved EKG to another PR since it introduces new deps that break the GHC 9 build

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2263"></a>


pepeiborra:
This is now ready for review and dogfooding

pepeiborra:
> Maybe we should wait to merge this after the next release?

Is this still the plan or should we go ahead and merge ?

jneira:
> 
> 
> > Maybe we should wait to merge this after the next release?
> 
> Is this still the plan or should we go ahead and merge ?

i think we have to postpone the release so i would go ahead and merge

PR #2267, opened 2021-10-09
Expose runtime metrics via EKG

pepeiborra:
Install a metrics server at port 8000 via EKG with the following metrics:

- `ghcide.values_count`- count of build results in the store
- `ghcide.database_count` - count of build keys in the store (these two would be the same in the absence of GC)
- `ghcide.build_count` - build count. A key is GC'ed if it is dirty and older than 100 builds
- `ghcide.dirty_keys_count` - non transitive count of dirty build keys
- `ghcide.indexing_pending_count` - count of items in the indexing queue
- `ghcide.exports_map_count` - count of identifiers in the exports map.

Unfortunately the EKG packages need allow-newer entries for GHC 9. This is also the case for the snap-server packages and dependencies. Since it doesn't look like these are getting resolved in Hackage any time soon, let's hold back on merging this PR for now

Upstream issues:
- https://github.com/tibbe/ekg/issues/85
- https://github.com/tvh/ekg-wai/issues/6

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2267"></a>

### DIFF
--- a/.github/workflows/flags.yml
+++ b/.github/workflows/flags.yml
@@ -65,11 +65,7 @@ jobs:
         run: cabal v2-build hls-graph --flags="embed-files stm-stats"
 
       - name: Build `ghcide` with flags
-        run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe"
-
-      - if: matrix.ghc == '8.10.7'
-        name: Build `ghcide` with ekg
-        run: cabal v2-build ghcide --flags="ekg"
+        run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe ekg"
 
       # we have to clean up warnings for 9.0 and 9.2 before enable -WAll
       - if: matrix.ghc != '9.0.2' && matrix.ghc != '9.2.2'
--- a/.github/workflows/flags.yml
+++ b/.github/workflows/flags.yml
@@ -67,6 +67,10 @@ jobs:
       - name: Build `ghcide` with flags
         run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe"
 
+      - if: matrix.ghc == '8.10.7'
+        name: Build `ghcide` with ekg
+        run: cabal v2-build ghcide --flags="ekg"
+
       # we have to clean up warnings for 9.0 and 9.2 before enable -WAll
       - if: matrix.ghc != '9.0.2' && matrix.ghc != '9.2.2'
         name: Build with pedantic (-WError)
--- a/cabal.project
+++ b/cabal.project
@@ -53,7 +53,10 @@ constraints:
   ghc-lib-parser-ex -auto,
   stylish-haskell +ghc-lib
 
- source-repository-package
+-- This is benign and won't affect our ability to release to Hackage,
+-- because we only depend on `ekg-json` when a non-default flag
+-- is turned on.
+source-repository-package
   type:git
   location: https://github.com/vshabanov/ekg-json
   tag: 00ebe7211c981686e65730b7144fbf5350462608
--- a/cabal.project
+++ b/cabal.project
@@ -53,6 +53,12 @@ constraints:
   ghc-lib-parser-ex -auto,
   stylish-haskell +ghc-lib
 
+ source-repository-package
+  type:git
+  location: https://github.com/vshabanov/ekg-json
+  tag: 00ebe7211c981686e65730b7144fbf5350462608
+  -- https://github.com/tibbe/ekg-json/pull/12
+
 allow-newer:
   -- ghc-9.2
   ----------
@@ -70,3 +76,11 @@ allow-newer:
   -- for shake-bench
   Chart:lens,
   Chart-diagrams:lens,
+
+  -- for ekg
+  ekg-core:base,
+  ekg-wai:base,
+
+  -- for shake-bench
+  Chart-diagrams:diagrams-core,
+  SVGFonts:diagrams-core
--- a/cabal.project
+++ b/cabal.project
@@ -82,7 +82,9 @@ allow-newer:
 
   -- for ekg
   ekg-core:base,
+  ekg-core:ghc-prim,
   ekg-wai:base,
+  ekg-wai:time,
 
   -- for shake-bench
   Chart-diagrams:diagrams-core,
--- a/cabal.project
+++ b/cabal.project
@@ -58,8 +58,8 @@ constraints:
 -- is turned on.
 source-repository-package
   type:git
-  location: https://github.com/vshabanov/ekg-json
-  tag: 00ebe7211c981686e65730b7144fbf5350462608
+  location: https://github.com/pepeiborra/ekg-json
+  tag: 7a0af7a8fd38045fd15fb13445bdcc7085325460
   -- https://github.com/tibbe/ekg-json/pull/12
 
 allow-newer:
--- a/docs/contributing/contributing.md
+++ b/docs/contributing/contributing.md
@@ -225,6 +225,35 @@ If you don't want to use [nix](https://nixos.org/guides/install-nix.html), you c
 
 See the [tutorial](./plugin-tutorial.md) on writing a plugin in HLS.
 
+## Measuring, benchmarking and tracing
+
+### Metrics
+
+HLS opens a metrics server on port 8000 exposing GC and ghcide metrics. The ghcide metrics currently exposed are:
+
+- `ghcide.values_count`- count of build results in the store
+- `ghcide.database_count` - count of build keys in the store (these two would be the same in the absence of GC)
+- `ghcide.build_count` - build count. A key is GC'ed if it is dirty and older than 100 builds
+- `ghcide.dirty_keys_count` - non transitive count of dirty build keys
+- `ghcide.indexing_pending_count` - count of items in the indexing queue
+- `ghcide.exports_map_count` - count of identifiers in the exports map.
+
+### Benchmarks
+
+If you are touching performance sensitive code, take the time to run a differential
+benchmark between HEAD and master using the benchHist script. This assumes that
+"master" points to the upstream master.
+
+Run the benchmarks with `cabal bench ghcide`.
+
+It should take around 25 minutes and the results will be stored in the `ghcide/bench-results` folder. To interpret the results, see the comments in the `ghcide/bench/hist/Main.hs` module.
+
+More details in [bench/README](../../ghcide/bench/README.md)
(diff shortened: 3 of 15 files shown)


pepeiborra:
I have reworked this to abstract monitoring and put it behind a Cabal flag, to avoid acquiring a hard dependency on the `ekg` ecosystem

cabal.project
```diff
@@ -47,7 +47,14 @@ index-state: 2022-03-08T10:53:01Z
 constraints:
     hyphenation +embed
 
+source-repository-package
```
michaelpj, line 50:
```suggestion
-- This is benign and won't affect our ability to release to Hackage, 
-- because we only depend on `ekg-wai` when a non-default flag
-- is turned on.
source-repository-package
```

ghcide/exe/Arguments.hs
```diff
@@ -19,6 +20,9 @@ data Arguments = Arguments
     ,argsVerbose                    :: Bool
     ,argsCommand                    :: Command
     ,argsConservativeChangeTracking :: Bool
+#ifdef MONITORING_EKG
```
michaelpj, line 23: I wonder if it would be simpler to define this argument unconditionally and simply have it do nothing if we haven't built with EKG support. This way is also fine, I just hate CPP :sweat_smile:

ghcide/ghcide.cabal
```diff
@@ -30,6 +30,11 @@ flag ghc-patched-unboxed-bytecode
     default:     False
     manual:      True
 
+flag ekg
+    description: Enable EKG monitoring of the build graph and other metrics on port 8999
```
michaelpj, line 34: The port is configurable, right? Just seems like an odd detail to have here.

ghcide/exe/Main.hs
```diff
@@ -142,4 +147,10 @@ main = withTelemetryLogger $ \telemetryLogger -> do
                 , optCheckProject = pure $ checkProject config
                 , optRunSubset = not argsConservativeChangeTracking
                 }
+#ifdef MONITORING_EKG
+        , IDEMain.argsMonitoring = OpenTelemetry.monitoring <> EKG.monitoring logger argsMonitoringPort
+#ifdef MONITORING_EKG
```
michaelpj, line 152: double ifdef

ghcide/exe/Main.hs
```diff
@@ -142,4 +147,10 @@ main = withTelemetryLogger $ \telemetryLogger -> do
                 , optCheckProject = pure $ checkProject config
                 , optRunSubset = not argsConservativeChangeTracking
                 }
+#ifdef MONITORING_EKG
+        , IDEMain.argsMonitoring = OpenTelemetry.monitoring <> EKG.monitoring logger argsMonitoringPort
```
michaelpj, line 151: This is inconsistent with `ghcide/src/Development/IDE/Main.hs`. I think they should probably be the same, and that one looks nicer?

ghcide/src/Development/IDE/Main.hs
```diff
@@ -268,6 +275,10 @@ defaultArguments recorder logger = Arguments
                 -- the language server tests without the redirection.
                 putStr " " >> hFlush stdout
                 return newStdout
+        , argsMonitoring = OpenTelemetry.monitoring
+#ifdef MONITORING_EKG
+                        <> EKG.monitoring logger 8999
```
michaelpj, line 280: What about confining the CPP to `Development.IDE.Monitoring.EKG`. If ekg support is disabled, it could just export a no-op `Monitoring`.

ghcide/src/Development/IDE/Monitoring/EKG.hs
```diff
@@ -0,0 +1,34 @@
...
+                    Right s -> do
+                        logInfo logger $ pack $
+                            "Started monitoring server on port " <> show port
+                        return $ Just s
+                    Left e -> do
+                        logInfo logger $ pack $
+                            "Unable to bind monitoring server on port "
+                            <> show port <> ":" <> show e
+                        return Nothing
+            return $ mapM_ (killThread . Monitoring.serverThreadId) server
```
michaelpj, line 33: do we want to log when stopping the thread?

ghcide/src/Development/IDE/Types/Monitoring.hs
```diff
@@ -0,0 +1,30 @@
+module Development.IDE.Types.Monitoring
+(Monitoring(..)
+) where
+
+import           Data.Int
+import           Data.Text                (Text)
+
+-- | An abstraction for runtime monitoring.
```
michaelpj, line 8: could you say a little about what a "gauge" and a "counter" are? I think I know but it would be nice to write it down.

ghcide/src/Development/IDE/Types/Monitoring.hs
```diff
@@ -0,0 +1,30 @@
...
+data Monitoring = Monitoring {
+    registerGauge   :: Text -> IO Int64 -> IO (),
+    registerCounter :: Text -> IO Int64 -> IO (),
+    start :: IO (IO ())
```
michaelpj, line 12:
```suggestion
    start :: IO (IO ()) -- ^ Start the monitoring system, returning an action which will stop the system.
```

ghcide/src/Development/IDE/Monitoring/EKG.hs
```diff
@@ -0,0 +1,34 @@
...
+import           Development.IDE.Types.Monitoring (Monitoring (..))
+import qualified System.Metrics                   as Monitoring
+import qualified System.Remote.Monitoring.Wai     as Monitoring
+
+-- | Monitoring using EKG
+monitoring :: Logger -> Int -> IO Monitoring
+monitoring logger port = do
+    store <- Monitoring.newStore
+    Monitoring.registerGcMetrics store
+    let registerCounter name read = Monitoring.registerGauge name read store
```
michaelpj, line 15:
```suggestion
    let registerCounter name read = Monitoring.registerCounter name read store
```

ghcide/src/Development/IDE/Monitoring/OpenTelemetry.hs
```diff
@@ -0,0 +1,31 @@
...
+-- | Dump monitoring to the eventlog using the Opentelemetry package
+monitoring :: IO Monitoring
+monitoring
+  | userTracingEnabled = do
+    actions <- newIORef []
+    let registerCounter name read = do
+            observer <- mkValueObserver (encodeUtf8 name)
+            let update = observe observer . fromIntegral =<< read
+            atomicModifyIORef'_ actions (update :)
+        registerGauge = registerCounter
```
michaelpj, line 22: I guess it's fine that we use the same implementation? I'm not sure whether anything cares about this distinction.

ghcide/src/Development/IDE/Monitoring/OpenTelemetry.hs
```diff
@@ -0,0 +1,31 @@
...
+    let start = do
+            a <- regularly 1 $ sequence_ =<< readIORef actions
```
michaelpj, line 24: Do we want to make the delay a parameter? it probably doesn't matter.

docs/contributing/contributing.md
```diff
@@ -225,6 +225,35 @@ If you don't want to use [nix](https://nixos.org/guides/install-nix.html), you c
...
+
+Run the benchmarks with `cabal bench ghcide`.
+
+It should take around 25 minutes and the results will be stored in the `ghcide/bench-results` folder. To interpret the results, see the comments in the `ghcide/bench/hist/Main.hs` module.
+
+More details in [bench/README](../../ghcide/bench/README.md)
+
+### Tracing
+
+HLS records opentelemetry eventlog traces via [opentelemetry](https://hackage.haskell.org/package/opentelemetry). To generate the traces, build with `-eventlog` and run with `+RTS -l`. To visualize the traces, install [Tracy](https://github.com/wolfpld/tracy) and use [eventlog-to-tracy](https://hackage.haskell.org/package/opentelemetry-extra) to open the generated eventlog.
```
michaelpj, line 255: Where do the results end up?

.github/workflows/flags.yml
```diff
@@ -71,6 +71,10 @@ jobs:
       - name: Build `ghcide` with flags
         run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe"
 
+      - if: matrix.ghc == '8.10.7'
```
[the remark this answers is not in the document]
pepeiborra: Because `ekg` is not compatible with other GHC versions

ghcide/exe/Arguments.hs
pepeiborra: It's too late now, CPP is everywhere :(

ghcide/ghcide.cabal
pepeiborra: Putting it here makes it easy to find :)

docs/contributing/contributing.md
pepeiborra: GHC contains extensive docs around eventlogs. I'll add a link.

ghcide/exe/Main.hs
pepeiborra: In what way is it inconsistent? The one in `ghcide/src/Development/IDE/Main.hs` is setting the default value, and this one is overriding it to customise the port number

michaelpj: Oh I see. I thought they were both setting it.

ghcide/src/Development/IDE/Monitoring/OpenTelemetry.hs
pepeiborra: This is just packaging the previously existing open telemetry monitoring, which didn't make this distinction,  to fit in the new abstraction

ghcide/src/Development/IDE/Monitoring/OpenTelemetry.hs
pepeiborra: Also part of the previously existing open telemetry monitoring code

ghcide/src/Development/IDE/Types/Monitoring.hs
pepeiborra: These are EKG names, I'll add a link

ghcide/src/Development/IDE/Main.hs
pepeiborra: Good idea, thanks

pepeiborra:
@michaelpj thanks for the review, I think I've addressed all your feedback now and this is ready to go

michaelpj (approved):
Nice!

PR #2275, opened 2021-10-15
Point to GitHub from Contributing.md

georgefst:
Follow-up to #2083.

Now that all this documentation is hosted on an external site, there really ought to be a link back here.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2275"></a>


georgefst:
Is CI not caching anything? There's an insane amount of unnecessary work here for a doc change.

PR #2296, opened 2021-10-23
[skip circleci] Enable func-test suite for windows

jneira:


jneira:
Ok, i've isolated the problematic tests to progress onesso the rest can be enabled. I think we can merge this and try to discern what concrete progress test(s) are triggering the hang

PR #1230, opened 2021-01-19
Delete global hie.yaml config

jneira:
Closes #741

PR #2311, opened 2021-10-29
Sir4ur0n/doc/cleanup hie

sir4ur0n:
It seems that https://github.com/haskell/haskell-language-server/pull/1230 removed the top level `hie-xxx.yaml` files but documentation was not updated accordingly, thus confusing for new contributors.

Also, it seems that implicit configuration detects as Stack project, so I clarified how to configure to force Cabal usage.

I also took the liberty to slightly improve visibility of some documentation links, when using `inline code` with links, it's really hard to tell it is a link, e.g.:
image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2311"></a>


jneira (approved):
Many thanks for the cleaning up and improve docs, looks really great

PR #2313, opened 2021-10-29
Nix: Add HLS as buildInput (flake, nix-shell)

sir4ur0n:
So I don't know what is the status quo/strategy regarding Nix shell providing, but just in case, I'm opening this PR.

I was surprised that neither `nix-shell` nor `nix develop` provide HLS itself (a prepackaged built, not the one being hacked). It means even after `nix-shell`ing or `nix develop`ing, one must also install HLS by some other means.

This PR thus adds it.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2313"></a>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -206,6 +206,7 @@
                 # ormolu
                 # stylish-haskell
                 opentelemetry-extra
+                haskell-language-server
               ]);
 
             src = null;


berberman:
The purpose of current nix shell is to provide an environment for developers to build HLS project via cabal build. Yes, the nixpkgs version of HLS executable is not included in the shell, because I feel like most developers would use/test HLS being hacked.

flake.nix
```diff
@@ -206,6 +206,7 @@
                 # ormolu
                 # stylish-haskell
                 opentelemetry-extra
+                haskell-language-server
```
berberman, line 209: I think `haskell-language-server` here is not the one in nixpkgs, but the one being hacked. So after each modification, entering the shell would rebuild HLS.

sir4ur0n: Oh you're right, my bad 🤦‍♂️

PR #2316, opened 2021-10-30
Improve recompilation avoidance in the presence of TH

wz1000:
The old recompilation avoidance scheme performs quite poorly when code generation is needed. We end up needing to recompile modules basically any time anything in their transitive dependency closure changes.

Most versions of GHC we currently support don't have a working implementation of code unloading for object code, and no version of GHC supports this on certain platforms like Windows. This makes it completely infeasible for interactive use, as symbols from previous compiles will shadow over all future compiles.

This means that we need to use bytecode when generating code for Template Haskell. Unfortunately, we can't serialize bytecode, so we will always need to recompile when the IDE starts. However, we can put in place a much tighter recompilation avoidance scheme for subsequent compiles:

1. If the source file changes, then we always need to recompile
   a. For files of interest, we will get explicit `textDocument/change` events that will let us invalidate our build products
   b. For files we read from disk, we can detect source file changes by comparing the `mtime` of the source file with the     build product (.hi/.o) file on disk.
2. If GHC's recompilation avoidance scheme based on interface file hashes says that we need to recompile, the we need to recompile.
3. If the file in question requires code generation then, we need to recompile if we don't have the appropriate kind of build products.
   a. If we already have the build products in memory, and the conditions 1 and 2 above hold, then we don't need to recompile
   b. If we are generating object code, then we can also search for it on disk and ensure it is up to date.
   Notably, we did _not_ previously re-use old bytecode from memory when `hls-graph`/`shake` decided to rebuild the `HiFileResult` for some reason

4. If the file in question used Template Haskell on the previous compile, then we need to recompile if any `Linkable` in its transitive closure changed. This sounds bad, but it is possible to make some improvements.
In particular, we only need to recompile if any of the `Linkable`s actually used during the previous compile change.

   How can we tell if a `Linkable` was actually used while running some TH?

   GHC provides a `hscCompileCoreExprHook` which lets us intercept bytecode as it is being compiled and linked. We can inspect the bytecode to see  which `Linkable` dependencies it requires, and record this for use in
   recompilation checking.
   We record all the home package modules of the free names that occur in the bytecode. The `Linkable`s required are then the transitive closure of these modules in the home-package environment. This is the same scheme as used by GHC to find the correct things to link in before running bytecode.

   This works fine if we already have previous build products in memory, but what if we are reading an interface from disk? Well, we can smuggle in the necessary information (linkable `Module`s required as well as the time they
   were generated) using `Annotation`s, which provide a somewhat general purpose way to serialise arbitrary information along with interface files.

   Then when deciding whether to recompile, we need to check that the versions of the linkables used during a previous compile match whatever is currently in the HPT.

The changes that were made to `ghcide` in order to implement this scheme include:

1. Add `RuleWithOldValue` to define Rules which have access to the previous value.
   This is the magic bit that lets us re-use bytecode from previous compiles
2. `IsHiFileStable` rule was removed as we don't need it with this scheme in place.
3. Everything in the store is properly versioned with a `FileVersion`, not just FOIs.
4. The `VFSHandle` type was removed. Instead we now take a VFS snapshot on every restart, and use this snapshot for all the `Rules` in that build. This ensures that Rules see a consistent version of the VFS for each run and also makes 3 possible.
   The `setVirtualFileContents` function was removed since it was not being used anywhere.
   If needed in the future, we can easily just modify the VFS using functions from `lsp`.
5. Fix a bug with the `DependencyInformation` calculation, were modules at the top of the hierarchy (no incoming edges) weren't being recorded properly

A possible future improvement is to use object-code on the first load (so we have a warm cache) and use bytecode for subsequent compiles.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2316"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -30,11 +30,52 @@ module Development.IDE.Core.Compile
   , loadModulesHome
   , getDocsBatch
   , lookupName
-  ,mergeEnvs) where
+  , mergeEnvs
+  ) where
 
+import           Control.Concurrent.Extra
+import           Control.Concurrent.STM.Stats      hiding (orElse)
+import           Control.DeepSeq                   (force, liftRnf, rnf, rwhnf)
+import           Control.Exception                 (evaluate)
+import           Control.Exception.Safe
+import           Control.Lens                      hiding (List)
+import           Control.Monad.Except
+import           Control.Monad.Extra
+import           Control.Monad.Trans.Except
+import           Data.Aeson                        (toJSON)
+import           Data.Bifunctor                    (first, second)
+import           Data.Binary
+import qualified Data.Binary                       as B
+import qualified Data.ByteString                   as BS
+import qualified Data.ByteString.Lazy              as LBS
+import           Data.Coerce
+import qualified Data.DList                        as DL
+import           Data.Functor
+import qualified Data.HashMap.Strict               as HashMap
+import           Data.IORef
+import           Data.IntMap                       (IntMap)
+import qualified Data.IntMap.Strict                as IntMap
+import           Data.List.Extra
+import           Data.Map                          (Map)
+import qualified Data.Map.Strict                   as Map
+import           Data.Maybe
+import qualified Data.Text                         as T
+import           Data.Time                         (UTCTime (..),
+                                                    getCurrentTime)
+import           Data.Time.Clock.POSIX             (posixSecondsToUTCTime)
+import           Data.Tuple.Extra                  (dupe)
+import           Data.Unique                       as Unique
+import           Debug.Trace
 import           Development.IDE.Core.Preprocessor
 import           Development.IDE.Core.RuleTypes
 import           Development.IDE.Core.Shake
+import           Development.IDE.Core.Tracing      (withTrace)
+import           Development.IDE.GHC.Compat        hiding (loadInterface,
+                                                    parseHeader, parseModule,
+                                                    tcRnModule, writeHieFile)
+import qualified Development.IDE.GHC.Compat        as Compat
+import qualified Development.IDE.GHC.Compat        as GHC
+import qualified Development.IDE.GHC.Compat.Util   as Util
 import           Development.IDE.GHC.Error
 import           Development.IDE.GHC.Orphans       ()
 import           Development.IDE.GHC.Util
@@ -43,24 +84,24 @@ import           Development.IDE.Spans.Common
 import           Development.IDE.Types.Diagnostics
 import           Development.IDE.Types.Location
 import           Development.IDE.Types.Options
-
-import           Development.IDE.GHC.Compat        hiding (loadInterface,
-                                                    parseHeader, parseModule,
-                                                    tcRnModule, writeHieFile)
-import qualified Development.IDE.GHC.Compat        as Compat
-import qualified Development.IDE.GHC.Compat        as GHC
-import qualified Development.IDE.GHC.Compat.Util   as Util
-
+import           GHC                               (ForeignHValue,
+                                                    GetDocsFailure (..),
+                                                    mgModSummaries,
+                                                    parsedSource)
+import qualified GHC.LanguageExtensions            as LangExt
+import           GHC.Serialized
 import           HieDb
-
+import qualified Language.LSP.Server               as LSP
 import           Language.LSP.Types                (DiagnosticTag (..))
-
-import           Control.DeepSeq                   (force, liftRnf, rnf, rwhnf)
+import qualified Language.LSP.Types                as LSP
+import           System.Directory
+import           System.FilePath
+import           System.IO.Extra                   (fixIO, newTempFileWithin)
+import           Unsafe.Coerce
 
 #if !MIN_VERSION_ghc(8,10,0)
 import           ErrUtils
 #endif
-
 
 #if MIN_VERSION_ghc(9,0,1)
 import           GHC.Tc.Gen.Splice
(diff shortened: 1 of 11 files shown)


pepeiborra:
This is too big to review in one go. Any chance that you could break it in smaller PRs?

wz1000:
I guess I could separate out the VFS and versioning changes out, but they aren't that big. The bulk of it is synergystic changes to the recompilation checking logic, it is not immediately clear to me how it could be separated out

ghcide/ghcide.cabal
```diff
@@ -295,6 +297,7 @@ executable ghcide
                 -Wno-name-shadowing
                 -- allow user RTS overrides
                 -rtsopts
+                -eventlog
```
pepeiborra, line 300: is this required?

ghcide/ghcide.cabal
```diff
@@ -219,6 +220,7 @@ library
                 -Wincomplete-uni-patterns
                 -Wno-unticked-promoted-constructors
                 -fno-ignore-asserts
+                -j12
```
pepeiborra, line 223: this is bad for GHC <9.2

wz1000: Yes, these were commited by mistake, I will remove them.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -934,24 +1028,29 @@ defineEarlyCutoff' doDiagnostics cmp key file old mode action = do
...
-                modTime <- liftIO $ (currentValue . fst =<<) <$> getValues state GetModificationTime file
+
+                modTime <- case eqT @k @GetModificationTime of
+                  Just Refl -> pure res
+                  Nothing
+                    | file == emptyFilePath -> pure Nothing
+                    | otherwise -> liftIO $ do
+                        m1 <- (currentValue . fst =<<) <$> getValues state (GetModificationTime_ False) file
+                        m2 <- (currentValue . fst =<<) <$> getValues state (GetModificationTime_ True ) file
+                        pure (m1 <|> m2)
```
pepeiborra, line 1049: You don't need to do this, the `Eq` and `Hashable` instance ignore the boolean

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -934,24 +1028,29 @@ defineEarlyCutoff' doDiagnostics cmp key file old mode action = do
...
+
                 (bs, (diags, res)) <- actionCatch
-                    (do v <- action; liftIO $ evaluate $ force v) $
+                    (do v <- action staleV; liftIO $ evaluate $ force v) $
                     \(e :: SomeException) -> do
                         pure (Nothing, ([ideErrorText file $ T.pack $ show e | not $ isBadDependency e],Nothing))
-                modTime <- liftIO $ (currentValue . fst =<<) <$> getValues state GetModificationTime file
+
+                modTime <- case eqT @k @GetModificationTime of
+                  Just Refl -> pure res
```
pepeiborra, line 1043: This change seems an unrelated bug fix, right?

wz1000:
@shapr you might want to try this out.

GavinRay97:
This is fantastic, thank you = D

Anton-Latukha:
Mentioned dependency #2263 - is merged.

This PR status seems to be WIP.

Anton-Latukha:
I wish web UI of rebase was allowed to do it in many commits.

Many of these files seem pretty rebasable.

myShoggoth:
@wz1000 What is the status of this PR from your point of view? @pepeiborra explained that it was a huge perf gain for HLS use cases, and would work for previous GHCs as well, so I'd love to see it pushed through. Do you need help? Thank you for doing this!

mpickering:
Looks like it needs a rebase and CI is failing.

wz1000:
My last commit tries to change the lsp-types benchmark to better showcase the improvements to recompilation avoidance because of this patch, but unfortunately `lsp-types` has a relatively "flat" module dependency graph.

This patch shows the most improvement when you have a "deep" module graph and you ask for recompilation of modules both at the bottom and the top of the graph - and the modules in between are a mixture of requiring and not requiring TH or linking.

Here are the benchmarks for `lsp-types` run locally on my machine: 

```
version    name                       success   samples   startup              setup   userTime             delayedTime     totalTime            buildRulesBuilt   buildRulesChanged   buildRulesVisited   buildRulesTotal   buildEdges   maxResidency   allocatedBytes
upstream   edit                             True      50        1.798771925          0.0                 191.91978414899998      2.3124167000000008e-2   191.947932916        155               130                 286                 1178              9138         494MB          284073MB
HEAD       edit                             True      50        1.766769395          0.0                 48.072982939            2.1007206000000004e-2   48.099120334000006   73                43                  235                 1155              8577         368MB          59120MB
upstream   hover                            True      50        1.9401033040000002   0.0                 6.7993e-5               4.8781324029999995      4.882002377          473               472                 481                 1159              9126         258MB          8776MB
HEAD       hover                            True      50        1.584940787          0.0                 6.4439e-5               4.162078842             4.16556116           757               756                 759                 1107              8651         258MB          8652MB
upstream   hover after edit                 True      50        2.23405389           0.0                 3.1066998000000002e-2   7.622193109000003       7.669040757          13                5                   202                 1202              9126         258MB          9246MB
HEAD       hover after edit                 True      50        1.757586773          0.0                 3.0687872e-2            5.010605698000001       5.057394084          23                7                   174                 1149              8624         263MB          9269MB
upstream   getDefinition                    True      50        1.699694912          0.0                 6.879799999999997e-5    6.944418013             6.948739521          376               375                 442                 1159              9126         257MB          8630MB
HEAD       getDefinition                    True      50        2.0626965260000003   0.0                 7.986000000000002e-5    6.418572896000001       6.424054067          369               368                 405                 1107              8651         258MB          8595MB
upstream   getDefinition after edit         True      50        2.26279817           0.0                 4.507241e-2             8.454230090000001       8.515164201000001    29                8                   205                 1202              9126         258MB          9886MB
HEAD       getDefinition after edit         True      50        2.069157503          0.0                 3.2946489e-2            4.430497608000002       4.47954948           22                7                   174                 1150              8623         259MB          9308MB
upstream   completions                      True      50        2.1221533420000003   0.0                 1.0265099999999998e-4   8.334979066999995       8.342324000000001    577               576                 583                 1159              9126         259MB          8818MB
HEAD       completions                      True      50        1.600802762          0.0                 6.1072e-5               4.494912567000004       4.498352855          595               594                 599                 1107              8651         258MB          8760MB
upstream   completions after edit           True      50        2.219558311          0.0                 3.1378321e-2            7.379409227000004       7.425725392          17                7                   204                 1202              9126         258MB          9134MB
HEAD       completions after edit           True      50        1.027144498          0.0                 2.8329477999999995e-2   4.463608328000001       4.502147832          20                7                   174                 1150              8623         260MB          9303MB
upstream   code actions                     False     50        0.0                  0.0                 0.0                     0.0                     0.0                  0                 0                   0                   0                 0            118MB          1459MB
HEAD       code actions                     False     50        0.0                  0.0                 0.0                     0.0                     0.0                  0                 0                   0                   0                 0            116MB          2001MB
upstream   code actions after edit          False     50        0.0                  0.0                 0.0                     0.0                     0.0                  0                 0                   0                   0                 0            114MB          1600MB
HEAD       code actions after edit          False     50        0.0                  0.0                 0.0                     0.0                     0.0                  0                 0                   0                   0                 0            118MB          1743MB
upstream   code actions after cradle edit   False     50        2.5541577180000004   8.389741869         0.0                     0.0                     6.160159156000001    527               378                 581                 1164              9138         456MB          17603MB
HEAD       code actions after cradle edit   False     50        1.801486665          5.651785252000001   0.0                     0.0                     3.228125078          393               249                 533                 1155              8574         415MB          11408MB
upstream   documentSymbols after edit       True      50        1.613184221          0.0                 3.1915035579999995      6.702542298000001       9.904618432000001    41                9                   207                 1202              9126         272MB          13751MB
HEAD       documentSymbols after edit       True      50        1.810905519          0.0                 2.1611687269999997      7.287025813999999       9.457577542000001    11                9                   176                 1150              8623         269MB          13153MB
upstream   hole fit suggestions             False     50        0.0                  0.0                 0.0                     0.0                     0.0                  0                 0                   0                   0                 0            284MB          9693MB
HEAD       hole fit suggestions             True      50        2.0343340100000002   0.0                 95.41021875600002       2.1052176000000002e-2   95.436770108         77                46                  301                 1154              8578         388MB          99495MB
```

The "after edit" benchmarks show some reduction in total time and rules built.

Unfortunately some of the code action benchmarks fail (on both this patch and upstream), so I will revert the changes from the last commit in this patch.

Here is another set of benchmarks for "get definition after edit" run on https://github.com/hasura/graphql-engine which is a large project satisfying the constraints mentioned above:

```
version    name                       success   samples   startup              setup   userTime             delayedTime     totalTime            buildRulesBuilt   buildRulesChanged   buildRulesVisited   buildRulesTotal   buildEdges   maxResidency   allocatedBytes
HEAD       getDefinition after edit   True      20        2.4624424570000003   0.0     135.65716052999997   18.553615088    154.21270514100001   77                45                  1751                9523              390324       1115MB         363910MB
upstream   getDefinition after edit   True      20        2.21295533           0.0     115.71000826900003   287.580292801   403.291925787        198               165                 1812                9887              395673       1185MB         747853MB
```

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -179,16 +187,102 @@ typecheckModule (IdeDefer defer) hsc keep_lbls pm = do
...
+
+    -- We want to record exactly which linkables/modules the typechecker needed at runtime
+    -- This is useful for recompilation checking.
+    -- From hscCompileCoreExpr'
+    -- Only compute direct dependencies instead of transitive dependencies.
+    -- It is much cheaper to store the direct dependencies, we can compute
+    -- the transitive ones when required.
+    -- Also only record dependencies from the home package
+    compile_bco_hook :: IORef (UniqSet ModuleName) -> HscEnv -> SrcSpan -> CoreExpr -> IO ForeignHValue
+    compile_bco_hook var hsc_env srcspan ds_expr
```
pepeiborra, line 211: I imagine that the body of this function is lifted from the default implementation of this hook in GHC.

Could you add a reference together with instructions on how to update it on subsequent GHC versions?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -216,23 +310,44 @@ captureSplices env k = do
...
 
-tcRnModule :: HscEnv -> [Linkable] -> ParsedModule -> IO TcModuleResult
+tcRnModule :: HscEnv -> ModuleEnv UTCTime -> ParsedModule -> IO TcModuleResult
 tcRnModule hsc_env keep_lbls pmod = do
   let ms = pm_mod_summary pmod
       hsc_env_tmp = hscSetFlags (ms_hspp_opts ms) hsc_env
+      hpt = hsc_HPT hsc_env
 
-  unload hsc_env_tmp keep_lbls
+  unload hsc_env_tmp $ map (\(mod, time) -> LM time mod []) $ moduleEnvToList keep_lbls
```
pepeiborra, line 319: This seems to be unloading too much, as `keep_lbls` omits the transitive dependencies. Why is this ok?

michaelpj:
There's a lot of careful reasoning about how this scheme works in the PR description. Can we make sure that this is captured somewhere for posterity, maybe a note in the code?

wz1000:
> There's a lot of careful reasoning about how this scheme works in the PR description. Can we make sure that this is captured somewhere for posterity, maybe a note in the code?

I've added a note to this effect

Issue #2114, opened 2021-08-20
Suggesting "Hiding other imports"-like action when ambiguity includes local definition


Ailrun:
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
haskell-language-server version: 1.3.0.0 (GHC: 9.0.1) (PATH: $PATH/haskell-language-server) (GIT hash: 2857eeece0398e1cd4b2ffb6069b05c4d2308b39)
Tool versions found on the $PATH
cabal:          3.4.0.0
stack:          2.7.1
ghc:            8.10.5
```
Which OS do you use: Manjaro

Which lsp-client do you use: Emacs

Describe your project (alternative: link to the project): 

Contents of `hie.yaml`: Implicit

### Steps to reproduce
With this file,
```Haskell
module Test where

useLex :: String
useLex = lex "" -- 'lex' can be either the local definition or 'Prelude.lex'.

lex :: String -> String
lex = id
```
HLS does not give "Use local, hiding imports" or similar, and just shows "Replace with qualified: Prelude.lex". This works well when ambiguity is caused by two imported definitions.

### Expected behaviour
Described as above.

### Actual behaviour
Described as above.

### Related information
https://github.com/haskell/haskell-language-server/blob/2fef04193cfe2ac08372a008a5b5cfb6275d5bea/ghcide/src/Development/IDE/Plugin/CodeAction.hs#L886-L900
checks only "imported from ...", whereas the actual error message with local definition is "defined at ...".

yoshitsugu:
hi. i'm new here. i'll try to fix this if no one else is already working on.

jneira:
@yoshitsugu welcome and many,bmany thanks to work on this nice feature, looking forward to your pr

PR #2320, opened 2021-11-01
Suggest hiding imports when local definition exists

yoshitsugu:
pull request for #2114
I'm new here so please point out if this fix has any problems

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2320"></a>


jneira:
@yoshitsugu thanks for the pr, i am rerunning ci as it is failing with transient errors lately

jneira (approved):
It looks great and it has the required tests, thanks!

Issue #2290, opened 2021-10-21
Cannot apply hlint hints when a trailing comment is on the same line.


shapr:
### Your environment
haskell-language-server 1.4
Which OS do you use: NixOS
Which lsp-client do you use: haskell-lsp in emacs
Describe your project (alternative: link to the project): https://github.com/shapr/hlsexamples/blob/main/src/Examples.hs#L14

### Steps to reproduce
Insert the line `biggest items = foldr1 max items -- the line above will show two hlint hints, "eta reduce" and "use maximum"
` into a file, and attempt to apply any of the hlint hints.
If you move the comment to a different line, the hlint hints apply with no problems.
### Expected behaviour
The hints should be applied.
### Actual behaviour
The hints are not applied.

jneira:
Hi, thanks for reporting the issue i've just tried to reproduce and i was not able to do it with several ghc versions (9.0.1, 8.10.7, 8.8.4). I am gonna add a test anyways to check it in ci

jneira:
@shapr I ve added a repro test here: https://github.com/haskell/haskell-language-server/blob/9818ab2d5271c6b534d36b0a5968b2f85c4d611c/plugins/hls-hlint-plugin/test/testdata/TwoHintsAndComment.hs

and it was succesfull for all CI configurations, let me know how could be changed to reproduce it

Issue #2280, opened 2021-10-17
HLint reports incorrect redundant irrefutable pattern


noughtmare:
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
haskell-language-server version: 1.4.0.0 (GHC: 8.10.7) (PATH: /nix/store/mxpy0nnbirprj791gsbncmkrvi67g8x8-haskell-language-server-1.4.0.0/bin/haskell-language-server)
Tool versions found on the $PATH
cabal:		3.4.0.0
stack:		Not found
ghc:		8.10.4
```
Which OS do you use: NixOS 21.05

Which lsp-client do you use: Coc on Neovim (same issue in VS Codium)

Describe your project (alternative: link to the project):
```haskell
{-# LANGUAGE Strict #-}
f ~x = x
```

### Steps to reproduce

Load the file.

### Expected behaviour

No HLint warnings.

### Actual behaviour

A HLint warning about a redundant irrefutable pattern:

```
[hlint refact:Redundant irrefutable pattern] [I] Redundant irrefutable pattern
Found:
  ~x
Why not:
  x
```

### Other comment

Running `hlint` (version 3.3.4) manually does work correctly.

### Include debug information
Execute in the root of your project the command `haskell-language-server --debug .` and paste the logs here:

<details>
<summary>
Debug output:
</summary>

```
No 'hie.yaml' found. Try to discover the project type!
Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 1.4.0.0 x86_64 ghc-8.10.7
Current directory: /tmp
Operating system: linux
Arguments: ["--debug","B.hs"]
Cradle directory: /tmp
Cradle type: Default

Tool versions found on the $PATH
cabal:		3.4.0.0
stack:		Not found
ghc:		8.10.4


Consulting the cradle to get project GHC version...
Project GHC version: 8.10.4
haskell-language-server exe candidates: ["haskell-language-server-8.10.4","haskell-language-server"]
Launching haskell-language-server exe at:/run/current-system/sw/bin/haskell-language-server-8.10.4
haskell-language-server version: 1.1.0.0 (GHC: 8.10.4) (PATH: /nix/store/xx32gx7bl1w2wk588h1bm02r8xcvflyc-haskell-language-server-1.1.0.0/bin/haskell-language-server)
 ghcide setup tester in /tmp.
[...]

Issue #2042, opened 2021-07-27
HLint doesn't report "Unused LANGUAGE pragma"


georgefst:
File contains just:
```hs
{-# LANGUAGE BangPatterns #-}
```
HLS reports no warnings. HLint 3.2.7 on the command line says:

```
haskell/Main.hs:1:1-29: Warning: Unused LANGUAGE pragma
Found:
  {-# LANGUAGE BangPatterns #-}
Perhaps you should remove it.
```

Odd. Is HLS intentionally disabling these hints for some reason?

jneira:
Hmm i would swear i've seen warning about unused pragmas at some point, will try to reproduce

georgefst:
Thanks. Let me know if you can't and I'll try to narrow it down.

jneira:
Can reproduce at first, not sure when it has started to fail, maybe since the pr which added support for ghc-9
We need a regression test here and investigate

PR #2321, opened 2021-11-01
[skip circleci] Rename hlint test data files and add regression tests

jneira:
* Change test module cases from ApplyRefactN to more descriptive names to help in its discovery
* Add regression tests for #2042 #2280 and a rperoduction test for #2290
* Correct `waitForProgressDone` hls-test-utils implementation: it was a exact duplicate of `waitForAllProgressDone` so it is replaced with the definition from the same function in the ghcide test suite

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2321"></a>


jneira:
Tests are green after the usual reruns, so I think it could be reviewed although most changes are naming related

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -285,3 +316,9 @@ hlintConfigWithFlags flags =
   where
     unObject (Object obj) = obj
     unObject _            = undefined
+
+knownBrokenForHlintOnGhcLib :: String -> TestTree -> TestTree
```
michaelpj, line 320: It's not immediately obvious to me what these mean or why it's these versions. Haddock?

michaelpj (approved):
> Correct waitForProgressDone hls-test-utils implementation: it was a exact duplicate of waitForAllProgressDone so it is replaced with the definition from the same function in the ghcide test suite

Can we share this? Perhaps we need to upstream some more test helpers into `lsp-test`?

plugins/hls-hlint-plugin/test/Main.hs
jneira: Hmm it is a somewhat long history, will try to add a concise comment although the entire thing should be explained in the plugin README or something

jneira: Not sure if i explained it clearly :thinking:

jneira:
Err why mergify merged the pr without waiting for required checks?, argh

jneira:
I think the post_job's are not shown until they are finished, so if test postjob has not finished but benchmark or nix one does, you can merge the pr, will try to fix giving them unique names

Issue #2317, opened 2021-10-31
Reconsider cabal flags manual setting in haskell-language-server.cabal


pepeiborra:
The flags are currently set at `manual: True`, preventing the Cabal solver from switching them in order to find a build plan.
This has two clear disadvantages:

1. HLS doesn't `cabal install` in Hackage with GHC 9.x because not all the plugins are available
2. We need to keep separate `cabal-ghcXXX.project` files in order to build locally, which works fine in the command line but doesn't work with the IDE because there is no way to change the `cabal.project` file from the cradle

On the other hand, setting the flags to `manual: False` has the following disadvantages:
1.  Cabal can reach an unexpected/incomplete build plan if users have an older index snapshot and forget to do `cabal update`
2. are there any others?

To me, 1 is not something we should care about, since users doing `cabal install haskell-language-server` are very few these days and probably know what they are doing.

jneira:
well myain concern is just about the automatic part: not only you can get a build error but it can find a plan excluding plugins you want to be available. Terrible  thing imho, even for whom "know what are they doing"

pepeiborra:
> well my main concern is just about the automatic part: not only you can get a build error but it can find a plan excluding plugins you want to be available. Terrible thing imho, even for whom "know what are they doing"

If that's the case, wouldn't the right outcome be a bug report in the Cabal issue tracker, rather than preventively disabling the feature? 

Anyhow, I don't think that we should be making decisions on a hunch. Do you have any evidence of this happening including metrics of how often it does?

jneira:
> If that's the case, wouldn't the right outcome be a bug report in the Cabal issue tracker, rather than preventively disabling the feature?

Dont think so cause that unwanted behaviour would be caused for opt-in in a known feature with known caveats. The legitime response will be: dont use automatic flags if you dont want cabal choose them automatically.

> Anyhow, I don't think that we should be making decisions on a hunch. Do you have any evidence of this happening including metrics of how often it does?

It is something that can happen due to how works the feature and how cabal produces a build plan. 

We could release hls for ghc-9 in hackage adding explicitly conditionals to hls.cabal. We cant drop cabal-ghc*.project cause we need `allow-newer`'s and `source-repository-package`'s. That is the main cause of needing it and no the flags.

[...]

Issue #2179, opened 2021-09-10
Support for ghc-9.2.*


jneira:
* Similar issue to #297
* There is a work in progress by @fendor started with #2128 although it still does not add initial support, only preparing it (maybe @fendor could tell us more about)

--------------------------------

## State of support

* The core functionality of haskell-language-server is supported via ghcide. 
* Plugins and formatters needs to be updated to make it work with ghc-9.0.1 so they are not included until then.
* So you should consider use other ghc version only if you are specially interested in those plugins o formatters 
* The plan is include a prebuilt executable for ghc-9.2.1 in the next hls release (1.6.0)

## Has ghc-9.2.1 support
- [x] `ghcide`
- [x] `haskell-language-server`
- [ ] `hls-haddock-comments-plugin`
- [x] `hls-hlint-plugin`
- [x] `hls-pragmas-plugin`
- [ ] `hls-retrie-plugin` without test, not included via flags
- [x] `hls-eval-plugin`
- [x] `hls-module-name-plugin`: it was not being tested in ci: see #2587. Doesnt compile for ghc-9.2.1. Fixed with #2594
- [x] `hls-class-plugin`
- [x] `hls-refine-imports-plugin`: #2601
- [ ] `hls-splice-plugin`
- [ ] `hls-tactic-plugin` 
- [x] `hls-explicit-imports-plugin`
- [x] `hls-call-hierarchy-plugin` it is being tested so it works but it is not included in hls via flags: see #2587
- [x] `hls-alternate-number-format-plugin`
- [x] `hls-qualify-imported-names-plugin`: #2600 
 
### Formatters
- [x] `hls-floskell-plugin`
- [x] `hls-ormolu-plugin`
- [x] `hls-fourmolu-plugin`
- [ ] `hls-brittany-plugin`
- [x] `hls-stylish-haskell-plugin`

### Build system and CI
- [x] included in github release
  - [x] `.github/workflows/build.yml` 
- [ ] hackage release with support for ghc-9.2.1: #2606 
  - [ ] `.github/workflows/hackage.yml` 
- [x] `stack-9.2.1.yaml`, added but not working
  - [x] `.circleci/config.yml`
- [x] `cabal-ghc921.project` - Used for disabling the not yet working plugins and fetching upstream patches
- [x]  nix support: with #2621 
  - [x] ./`configuration-ghc-921.nix`
  - [x] update `.github/workflows/nix.yml` 
[...]

PR #2322, opened 2021-11-02
Add support for install hls from hackage using ghc 9.0.1

jneira:
* As commented in https://github.com/haskell/haskell-language-server/issues/2317#issuecomment-956178685 it is an alternative to make plugin flags automatic to get hls buildable from hackage and ghc-9.0.1
* This pr:
  * Encode explicitly what plugins are buildable for each ghc versions in `haskell-language-server.cabal`, the only config file seen by hackage builds
  * Add a new flag `ignore-plugins-ghc-bounds` to allow custom builds out of hackage force the inclusion of plugins, supposelly adding `allow-newer` or `remote-source-packages` in a `cabal.project`
  * Remove `all-plugins` and `all-formatters` as afaics they are almost useless with actual default values for them and individual plugin flags and being used with `||` in conditions
  * Remove unused plugins conditions in `test-func` as almost all plugins were moved to their own packages. Deprecate the remaining ones as hopefully they will be moved sooner or later
  * Change hackage github workflow to include ghc-9.0.1 and the option to only run the check step, without upload effectively packages 
  * I've added checks for ghc-9.2.1 but they are an aproximation based in the actual `cabal-ghc921-project` which afaiu does not work

This has not been done earlier cause i thought we should "publish" a complete version of hls in hackage. Assuming that it would not take so long to be achieved as it did. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2322"></a>

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -86,7 +86,7 @@ library
 -- - Bulk flags should be default:False
 -- - Individual flags should be default:True
 
-flag force-plugins
+flag ignore-plugins-ghc-bounds
   description: Force the inclusion of plugins even if they are not buildable by default with a specific ghc version
   default:     False
   manual:      True
@@ -189,12 +189,12 @@ common example-plugins
                  Ide.Plugin.Example2
 
 common class
-  if flag(class) && (impl(ghc < 9.0.1) || flag(force-plugins))
+  if flag(class) && (impl(ghc < 9.0.1) || flag(ignore-plugins-ghc-bounds))
     build-depends: hls-class-plugin ^>=1.0.0.1
     cpp-options: -Dclass
 
 common callHierarchy
-  if flag(callHierarchy) && (impl(ghc < 9.2.1) || flag(force-plugins))
+  if flag(callHierarchy) && (impl(ghc < 9.2.1) || flag(ignore-plugins-ghc-bounds))
     build-depends: hls-call-hierarchy-plugin ^>=1.0.0.0
     cpp-options: -DcallHierarchy
 
@@ -214,7 +214,7 @@ common importLens
     cpp-options: -DimportLens
 
 common refineImports
-  if flag(refineImports) && (impl(ghc < 9.2.1) || flag(force-plugins))
+  if flag(refineImports) && (impl(ghc < 9.2.1) || flag(ignore-plugins-ghc-bounds))
     build-depends: hls-refine-imports-plugin ^>=1.0.0.0
     cpp-options: -DrefineImports
 
@@ -224,12 +224,12 @@ common rename
     cpp-options: -Drename
 
 common retrie
-  if flag(retrie) && (impl(ghc < 9.2.1) || flag(force-plugins))
+  if flag(retrie) && (impl(ghc < 9.2.1) || flag(ignore-plugins-ghc-bounds))
     build-depends: hls-retrie-plugin ^>=1.0.0.1
     cpp-options: -Dretrie
 
 common tactic
-  if flag(tactic) && (impl(ghc < 9.0.1) || flag(force-plugins))
+  if flag(tactic) && (impl(ghc < 9.0.1) || flag(ignore-plugins-ghc-bounds))
     build-depends: hls-tactics-plugin >=1.2.0.0 && <1.5
     cpp-options: -Dtactic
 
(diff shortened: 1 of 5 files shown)


haskell-language-server.cabal
```diff
@@ -83,18 +83,11 @@ library
...
 
-flag all-plugins
-  description: Enable all non formatter plugins
-  default:     False
-  manual:      True
-
-flag all-formatters
-  description: Enable all fomatters
+flag force-plugins
+  description: Force the inclusion of plugins even if they are not buildable by default with a specific ghc version
```
michaelpj, line 90: Hmm, from the name/description I would also expect this to enable all the plugins. In fact it just removes a constraint. So maybe we could call it something like `ignore-ghc-bounds`?

michaelpj (approved):
LGTM! I guess we still need the multiple `cabal.project`s? I guess the ghc-9 ones should switch to setting `force-plugins` or whatever we call it.

Slightly off-topic, but looking at the cabal.projects, could we make the ghc-9 parts into something that could be used as `cabal.project.local`? That would avoid the need to repeat themselves so much, then they would basically just set `force-plugins` and add `source-repository-package`s or whatever?

haskell-language-server.cabal
```diff
@@ -196,94 +189,94 @@ common example-plugins
                  Ide.Plugin.Example2
 
 common class
-  if flag(class) || flag(all-plugins)
+  if flag(class) && (impl(ghc < 9.0.1) || flag(force-plugins))
```
michaelpj, line 192: Also, it might be neat to put a comment on each of these linking to the issue that describes the blocker, so if keen people end up here they know what to look for!

haskell-language-server.cabal
jneira: yeah you are right: we have the rename plugin disabled in the flag and i would expect the flag enable it

what about `ignore-plugins-ghc-bounds`?

haskell-language-server.cabal
jneira: ya, i also miss that in the actual `cabal-ghc<version>.project` so it will improve things

jneira:
> 
> Slightly off-topic, but looking at the cabal.projects, could we make the ghc-9 parts into something that could be used as `cabal.project.local`? That would avoid the need to repeat themselves so much, then they would basically just set `force-plugins` and add `source-repository-package`s or whatever?

I tried (and got it somewhat work) here: https://github.com/haskell/haskell-language-server/pull/2241
Maybe it would be matter of reopening it

haskell-language-server.cabal
michaelpj: Sounds good, verbose names are fine IMO

haskell-language-server.cabal
jneira: Although all of them could point to the general issues about ghc support: https://github.com/haskell/haskell-language-server/issues/297 and https://github.com/haskell/haskell-language-server/issues/2179?

michaelpj:
> I tried (and got it somewhat work) here

Hmm I see there were various reasons not to do it. Maybe Pepe's idea of trying to make a ghc9-compatible main `cabal.project` that also works for the others is sensible...

jneira:
> 
> 
> > I tried (and got it somewhat work) here
> 
> Hmm I see there were various reasons not to do it. Maybe Pepe's idea of trying to make a ghc9-compatible main `cabal.project` that also works for the others is sensible...

It is an option but it has its own caveats: we will use for all ghc versions package versions not in hackage. And make actual bounds work with not needed `allow-newer`'s. They can be retrocompatible or doesnt and project buildability could differ between the project and hackage without being aware until the release is about to be done.

I am for trying hard to remove all `source-repository-package` and `allow-newer`. Isolating them in specific project config files is a way to keep the main one clean from them. We had bad experiences in the past about that.

michaelpj:
> I am for trying hard to remove all source-repository-package and allow-newer. Isolating them in specific project config files is a way to keep the main one clean from them. We had bad experiences in the past about that.

Right: if the main cabal.project is clean from those, then we should be able to release to Hackage at any time, with some plugins disabled as per the cabal files. Whereas otherwise we could accidentally make that impossible again.

jneira:
@pepeiborra thoughts about this approach to make possible the hackage release for ghc-9.0.1?

PR #2323, opened 2021-11-02
Improve the performance of GetModIfaceFromDisk in large repos and delete GetDependencies

### LINKED ISSUE #2304: GetModIfaceFromDisk has bad first call complexity
The current implementation of `GetModIface` is O(# of transitive import statements), i.e. it does an amount of work proportional to the number of imports in the project. To see why, notice that to evaluate `GetModIface M`, that is to load the interface for a module M, we first need to load the interfaces of all its DD direct imports, which involves DD recursive calls to `GetModIface`. Transitively, this makes one `GetModIface M` call for every import in the module graph of `M`. This leads to bad startup performance in projects with a dense module graph since:
```
O(# of import statements) = O(edges in module graph) ==> O(modules^2).
```
Once `GetModIface M` has been called at least once the ongoing cost is O(direct imports), thanks to early cutoff, which is very good.

How can we improve the bad performance of the first call while preserving the good performance of the second call?

pepeiborra:
This changes `GhcSessionDeps` to reuse the `GhcSessionDeps` of the dependencies, and then `GetModIfaceFromDisk` to use `GhcSessionDeps` instead of calling `ghcSessionDepsDefinition`. 

There are three benefits:
1. `GetModIfaceFromDisk` and `GhcSessionDeps` no longer depend on the transitive module summaries. This means fewer edges in the build graph = smaller build graph = faster builds
2. Avoid duplicate computations in setting up the GHC session with the dependencies of the module. Previously the total work done was O(NlogN) in the number of transitive dependencies, now it is O(N).
3. Increased sharing of HPT and FinderCache. 

Ideally we should also share the module graphs, but the datatype is abstract, doesn't have a Monoid instance, and cannot be coerced to something that has. We will need to add the Monoid instance in GHC first.

On the Sigma repo:
- the startup metric goes down by ~34%.
- The edit metric also goes down by 15%.
- Max residency is down by 30% in the edit benchmark.

Interestingly, the memory usage goes down despite caching more!

Fixes #2304 (in that it improves the complexity as much as I can think of)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2323"></a>


pepeiborra:
One surprising and **unintended** consequence of this is that `GetDependencies` is not getting pulled by anything anymore, and therefore neither is `ReportImportCycles`. This is unfortunate for two reasons: 

1. Import cycles no longer get diagnostics
2. Import cycles will lead to build graph cycles, and the IDE will stop producing diagnostics or even blow up

I suppose we'll need to introduce an artificial dependency on `GetDependencies`, but doing that will reintroduce O(imports^2) edges in the build graph which is damn expensive. 

A better option would be to handle cycles in the build system (his-graph), and make it possible to handle them in build rules somehow. Any thoughts @ndmitchell  ?

ndmitchell:
> You have cycle detection in Shake, would it work for his-graph?

There are two cycle detections in Shake. One uses a stack of basically where you have come from, and produces moderately good error messages. If you enhanced that with the stack in terms of import definitions you would get nice user actionable error messages.

The second cycle detector is the worse one. When A and B are mutually depended upon, and both start simultaneously, they both notice each other is in progress, and then everything halts. Shake gives a really poor error here, but does spot the cycle because it has things to be done, and can't make progress. To get a good error message would require fairly complex graph algorithms. There are some, but they are probably a lot more work than doing a dependency scan in the rules.

pepeiborra:
@jneira any idea of why the funs-test testsuite is timing out on Windows w/ ghc 9.0 after >4h? It normally takes 3m

pepeiborra:
@wz1000 are you able to review this core change?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1055,20 +1056,28 @@ writeHiFileAction hsc hiFile = do
         resetInterfaceStore extras $ toNormalizedFilePath' targetPath
         writeHiFile hsc hiFile
 
+data RulesConfig = RulesConfig
+    { -- | Disable import cycle checking for improved performance in large codebases
+      checkForImportCycles :: Bool
+    -- | Disable TH for improved performance in large codebases
+    , enableTemplateHaskell :: Bool
```
wz1000, line 1063: The Eval plugin will also not work if this is `False`. Is that correct?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -686,6 +691,30 @@ loadModulesHome mod_infos e =
...
+        hsc_mod_graph = mkModuleGraph $ extraModSummaries ++ nubOrdOn ms_mod (concatMap (mgModSummaries . hsc_mod_graph) envs)
+    }
+    where
+    -- required because 'FinderCache':
+    --  1) doesn't have a 'Monoid' instance,
+    --  2) is abstract and doesn't export constructors
+    -- To work around this, we coerce to the underlying type
+    -- To remove this, I plan to upstream the missing Monoid instance
+        concatFC :: [FinderCache] -> FinderCache
+        concatFC = unsafeCoerce (mconcat @(Map InstalledModule InstalledFindResult))
```
wz1000, line 716: This can easily break with new GHC releases. Can we guard this using CPP for GHC versions where this is known to be ok? It can be a compile error for as of yet unreleased GHC versions.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -691,49 +679,62 @@ loadGhcSession = do
...
+
+            depSessions <- map hscEnv <$> uses_ GhcSessionDeps deps
+            let uses_th_qq =
+                    xopt LangExt.TemplateHaskell dflags || xopt LangExt.QuasiQuotes dflags
+                dflags = ms_hspp_opts ms
+            ifaces <- if uses_th_qq || forceLinkables
+                        then uses_ GetModIface deps
+                        else uses_ GetModIfaceWithoutLinkable deps
+
+            let inLoadOrder = map hirHomeMod ifaces
```
wz1000, line 716: So we don't need the `reverse` after all? The `inLoadOrder` name is a bit misleading in that case.

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
```diff
@@ -533,30 +529,19 @@ prettyWarn Warn{..} =
...
     -- to avoid interfering with ghcide
+    -- UPDATE: I suspect that this doesn't really work, we always get the same Session
+    --         we probably cache hscEnvs in the Session state
     IdeGhcSession{loadSessionFun} <- useNoFile_ GhcSessionIO
     let fp = fromNormalizedFilePath nfp
     ((_, res),_) <- liftIO $ loadSessionFun fp
-    let hscEnv = fromMaybe (error $ "Unknown file: " <> fp) res
-    ghcSessionDepsDefinition hscEnv nfp
+    let env = fromMaybe (error $ "Unknown file: " <> fp) res
+        ghcSessionDepsConfig = def{forceLinkables = True, checkForImportCycles = False}
```
wz1000, line 542: I don't quite understand this interaction with the eval plugin. Why do we need `forceLinkables = True` here?

wz1000 (changes requested):
Looks like a good improvement.

ghcide/src/Development/IDE/Core/Rules.hs
pepeiborra: Quite the opposite, this setting will not work if the Eval plugin is enabled, since it will redefine the `NeedsCompilation` rule and override this setting.

ghcide/src/Development/IDE/Core/Compile.hs
pepeiborra: I think we will find out if this breaks, there's no need to add special CPP for that

ghcide/src/Development/IDE/Core/Rules.hs
pepeiborra: The `reverse` stopped being necessary but I don't remember if it's because it was superfluous or because something else changed

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
pepeiborra: `ghcSessionDepsImplementation` requests `GetModIfaceWithoutLinkables` by default, unless the module itself uses TH.

ghcide/src/Development/IDE/Core/Compile.hs
pepeiborra: Moreover, the missing instances will be added hopefully soon - https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6922

jneira:
> 
> 
> @jneira any idea of why the func-test testsuite is timing out on Windows w/ ghc 9.0 after >4h? It normally takes 3m

I got to enable all tests but progress ones in #2296 but by pure brute force, after that it has not timed out. I have to investigate why progress tests continue causing hangs although i suspect about eval one.

pepeiborra:
> > 
> 
> > 
> 
> > @jneira any idea of why the func-test testsuite is timing out on Windows w/ ghc 9.0 after >4h? It normally takes 3m
> 
> 
> 
> I got to enable all tests but progress ones in #2296 but by pure brute force, after that it has not timed out. I have to investigate why progress tests continue causing hangs although i suspect about eval one.
> 
> 
> 
> https://github.com/haskell/haskell-language-server/blob/6fb0dc00a46f711f4c2462fd5f91fcdf549c38b2/test/functional/Main.hs#L37

I see a timeout as well in the CI results for #2296

jneira:
Yeah, i tried all combinations of ignore specific test groups several times and several attempts ended with hangs for windows and 9.0.1 until https://github.com/haskell/haskell-language-server/runs/4079556307?check_suite_focus=true ignoring only progress tests.

After that i've not seen another run get stuck in windows and 9.0.1 in any branch (i've just did a search in subsequent runs to find it and i did not see anyone but maybe i missed it).

jneira:
Oh you are referring to https://github.com/haskell/haskell-language-server/runs/4088340871?check_suite_focus=true? 
Maybe the hang is less probable now but it continue happening... :thinking: 
In my repo the same commits are not causing the hangs and of course they are not reproduced locally

jneira:
Lets see how is going the actual workflow run, will try to diable other test groups if the hang is reproduced (or afterwards in another branch)

jneira:
The func-test suite did not hang this time: https://github.com/haskell/haskell-language-server/runs/4150189056?check_suite_focus=true
With a little bit of luck the rest of jobs will pass :crossed_fingers:

pepeiborra:
I'm not going to merge this yet since the benchmarks show a perf regression after introducing the synthetic dependency on `ReportImportCycles`.

pepeiborra:
> The func-test suite did not hang this time: https://github.com/haskell/haskell-language-server/runs/4150189056?check_suite_focus=true
> With a little bit of luck the rest of jobs will pass 🤞

While it didn't time out this time around, it timed out for 4/5 previous attempts, so I strongly think that we should disable it (or the offending tests)

pepeiborra:
With the new build graph statistics outputs in the benchmark suite (#2343) it's easier to see what's causing the regression:

```
upstream   edit               True      50        2.195506             0.0     10.804728000000003   1.7481000000000003e-2   10.829789            17                13                  1499                6054              138396       403MB          29825MB
nocycles   edit               True      50        2.439721             0.0     10.207598            9.900999999999997e-2    10.314124000000001   18                14                  1463                5584              36562        366MB          28741MB
HEAD       edit               True      50        1.9325120000000002   0.0     13.451666999999997   1.7155999999999998e-2   13.475761            183               97                  1629                6040              117117       390MB          38599MB
```

Compared to upstream, this PR HEAD increases the number of rules built on edit by a factor of 10. 

It also shows the cost (3X build graph edges) of the import cycle tracking rules.

pepeiborra:
Perf regression fixed, benchmarks look fine: 
```
version    name                             success   samples   startup              setup                   userTime             delayedTime             totalTime            buildRulesBuilt   buildRulesChanged   buildRulesVisited   buildRulesTotal   buildEdges   maxResidency   allocatedBytes
5
upstream   edit                             True      50        2.740945656          0.0                     23.840065159999998   5.403511798             29.251490501000003   18                14                  1500                6054              138396       347MB          28401MB
6
HEAD       edit                             True      50        2.678469225          0.0                     29.585207137000012   2.8267619e-2            29.620908206000003   17                13                  1628                6040              117117       349MB          28736MB
```

wz1000:
What was the fix?

Issue #2309, opened 2021-10-29
Clarify `install/` sub-directory role, or integrate it into HLS project


sir4ur0n:
[Contributing guidelines](https://haskell-language-server.readthedocs.io/en/latest/contributing/contributing.html#building-haskell-language-server) explain:

> Note: the ./install/ folder is not directly tied to the project so it has dedicated ./install/hie.yaml.[cbl|stack] templates.

But this leads to several questions:
* Why isn't the `install/` folder treated as another package of the root directory HLS project? It would make everything (build, HIE/HLS integration, etc.) simpler, wouldn't it?
* What is the purpose of this sub-directory? There is no `README.md` in that directory, and I couldn't find any documentation about it. Based on the description in `install/hls-install.cabal` I guess it's about installing HLS (for users? for something else?) but this limited information :sweat_smile:

PR #2324, opened 2021-11-02
Add README.md in install/ subproject

sir4ur0n:
Closes https://github.com/haskell/haskell-language-server/issues/2309


jneira (approved):
Looks really great (modulo one minor change), thanks for complete the info

sir4ur0n:
What change? :sweat_smile:

install/README.md
```diff
@@ -0,0 +1,8 @@
+# hls-install
+
+This directory contains the `hls-install` project, an Haskell Language Server installer.
+It is used when [installing from sources](../docs/installation.md#building).
```
jneira, line 4: We are using absolute urls to readthedocs for links, see the main README as example

jneira:
[skip circleci]

> 
> 
> What change? sweat_smile

sorry i was looking why ci failed :sweat_smile:

install/README.md
sir4ur0n: Done (and I took the liberty to fix a few other links in the documentation)

jneira: oh, but links inside docs themselves can be relative, we need absolute ones in external resources like README's

sir4ur0n: Oh I see, my bad :facepalm: I have force-pushed the fix

jneira: dont worry, i could be more specific, will update to head to get the ci fix

PR #2325, opened 2021-11-02
Remove automatic comment to [skip circleci]

jneira:
It does not work for prs from external repos like #2324 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2325"></a>


jneira:
Merging using admin rights to unblock other pr like #2324

PR #2329, opened 2021-11-05
Not suggest exported imports

### LINKED ISSUE #2079: Don't suggest making imports explicit when the module is exported
image

This is almost certainly not what a user wants to do. i.e. being selective about what is reexported when reexporting a whole module.

yoshitsugu:
I attempted to fix #2079
I'm not familiar with GHC AST kind things. I'd be happy to get some advice if anyone has more smart way to implement this.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2329"></a>

### DIFF
--- a/plugins/hls-explicit-imports-plugin/hls-explicit-imports-plugin.cabal
+++ b/plugins/hls-explicit-imports-plugin/hls-explicit-imports-plugin.cabal
@@ -32,3 +32,16 @@ library
   default-extensions:
     DataKinds
     TypeOperators
+
+test-suite tests
+  type:             exitcode-stdio-1.0
+  default-language: Haskell2010
+  hs-source-dirs:   test
+  main-is:          Main.hs
+  ghc-options:      -threaded -rtsopts -with-rtsopts=-N
+  build-depends:
+    , base
+    , filepath
+    , hls-explicit-imports-plugin
+    , hls-test-utils
+    , text
--- a/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
+++ b/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
@@ -23,7 +23,8 @@ import           Data.Aeson.Types                     (FromJSON)
 import qualified Data.HashMap.Strict                  as HashMap
 import           Data.IORef                           (readIORef)
 import qualified Data.Map.Strict                      as Map
-import           Data.Maybe                           (catMaybes, fromMaybe)
+import           Data.Maybe                           (catMaybes, fromMaybe,
+                                                       isJust)
 import qualified Data.Text                            as T
 import           Development.IDE                      hiding (pluginHandlers,
                                                        pluginRules)
@@ -175,6 +176,13 @@ instance Show MinimalImportsResult where show _ = "<minimalImportsResult>"
 
 instance NFData MinimalImportsResult where rnf = rwhnf
 
+exportedModuleStrings :: ParsedModule -> [String]
+exportedModuleStrings ParsedModule{pm_parsed_source = L _ HsModule{..}}
+  | Just export <- hsmodExports,
+    exports <- unLoc export
+    = map show exports
+exportedModuleStrings _ = []
+
 minimalImportsRule :: Rules ()
 minimalImportsRule = define $ \MinimalImports nfp -> do
   -- Get the typechecking artifacts from the module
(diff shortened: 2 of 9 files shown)


plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -175,6 +176,13 @@ instance Show MinimalImportsResult where show _ = "<minimalImportsResult>"
 
 instance NFData MinimalImportsResult where rnf = rwhnf
 
+exportedModuleStrings :: ParsedModule -> [String]
+exportedModuleStrings ParsedModule{pm_parsed_source = L _ HsModule{..}}
+  | Just export <- hsmodExports,
+    exports <- unLoc export
+    = map show exports
```
yoshitsugu, line 183: I'm not sure how to extract export module names

plugins/hls-explicit-imports-plugin/hls-explicit-imports-plugin.cabal
```diff
@@ -32,3 +32,16 @@ library
   default-extensions:
     DataKinds
     TypeOperators
+
+test-suite tests
```
jneira, line 36: Wow the plugin had no test suite??? Many thanks for adding it, a great contribution in its own

jneira: It would be great to add it to the github workflow to run them in ci, adding an entry like

https://github.com/haskell/haskell-language-server/blob/a0cd84b4f6c6e48eb56e135faf5f9b1e3139886c/.github/workflows/test.yml#L207-L209

jneira:
Many thanks for your new contribution!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jneira: Does it works as is and you would like to improve it?

yoshitsugu: @jneira Yes. It passed the test I wrote and I can verify it not suggest "explicit import" for exported modules. But i'm not sure it's a good way to implement this by just using `show`. maybe  there is more appropriate function to extract module name I think. :thinking:

plugins/hls-explicit-imports-plugin/hls-explicit-imports-plugin.cabal
yoshitsugu: OK. I'll fix.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
pepeiborra: You could use pattern matching to precisely extract module reexports, but I'm not sure that that's an improvement overall since it also means a stronger coupling to the ghc api, which is bad since we need to support multiple GHC versions. So using `show` seems fine to me.

pepeiborra (approved):
Thanks for improving this plugin and adding tests!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
yoshitsugu: @takoeight0821 I see. thank you for your comment 😄

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -207,19 +215,27 @@ extractMinimalImports (Just hsc) (Just TcModuleResult {..}) = do
   let tcEnv = tmrTypechecked
       (_, imports, _, _) = tmrRenamed
       ParsedModule {pm_parsed_source = L loc _} = tmrParsed
+      emss = exportedModuleStrings tmrParsed
       span = fromMaybe (error "expected real") $ realSpan loc
+  let notExportedImports = filter (notExported emss) imports
```
michaelpj, line 220:
```suggestion
  -- Don't make suggestions for modules which are also exported, the user probably doesn't want this!
  -- See https://github.com/haskell/haskell-language-server/issues/2079
  let notExportedImports = filter (notExported emss) imports
```

yoshitsugu: @michaelpj thanks!

jneira (approved):
many thanks!

PR #2332, opened 2021-11-06
Include sortText in completions and improve suggestions

### LINKED ISSUE #2291: feature request: provide sortText with completionItems
We received this issue on the neovim client: https://github.com/neovim/neovim/issues/15414

One thing that was reported by @tomsmeding (regarding HLS) is that the results are pre-sorted by the server to include 

* case-insensitive matching
* typo-fixing
* intelligent result ordering so that most likely matches come first

According to the specification, if sortText is not supplied, the client should sort on the label. See [here](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionItem)

>
	 * A string that should be used when comparing this item
	 * with other items. When `falsy` the label is used
	 * as the sort text for this item.
	sortText?: string;

If the results should not be sorted, sortText should be supplied to indicate the results should not be sorted as such, otherwise according to the spec we should sort on the label as we currently do which perturbs the HLS returned ordering.

I can look into providing a PR if welcome :) I don't any experience with the HLS codebase though

pepeiborra:
Changes: 

* Merge sort (by fuzzy score) all the sources of completions (module names, keywords, and identifiers). Previously they were individually sorted but never merged, leading to some poor completion choices
* Sort qualified completions up - previously some keywords could appear before
* Sort prefix completions up - e.g. `head` for `hea` over `NoHeapProfiling`
* Sort suggestions alphabetically
* Show the provenance in the detail field when no type info available
 
image

* Match the first character case-sensitively - e.g. don't suggest `truncate` for `Tru`, only `True` 
* Fill in the sortText field to include our ordering (#2291). Previously the LSP client was sorting by label which was disrupting our ordering and breaking qualified completions in some cases (#2082) 
* Extend the documentation attached to every identifier completion to state whether the identifier is already imported

Closes #2291 and #2082

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2332"></a>


jneira (approved):
LGTM thanks for improving this

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -156,17 +156,37 @@ getCompletionsLSP ide plId
...
+   Ordering is important because local/nonlocal are import aware, whereas
+   global are not and will always insert import statements, potentially redundant.
+
+   On the other hand the fuzzy sort algorithm doesn't always sort in the optimal way,
+   there is room for the LSP client to improve.
+
+   According to the LSP specification, if no sortText is provided, the label is used.
+   To allow the LSP client to reorder identifiers while preserving the relative ordering
+   of repeated occurrences we generate sortText values that include both the label and
+   an index denoting the relative order
```
michaelpj, line 175: If I understand this right, given a list of completions like:
```
-- local
x
y
-- global
x
y
```
this will produce
```
x -- local
x -- global
y -- local
y -- global
```

Is that what we want? Maybe it would be better to put *all* the local ones first, i.e. have the sort text be just the index. The client can still filter on the label (or we can set `filterText` explicitly). Either way, would be good to have an example in this comment.

pepeiborra: This is what we want if the LSP client thinks that `x` is more relevant than `y`. The reason is that the fuzzy sort package that we use doesn't do an optimal job imho, so we want to allow the LSP client to improve on that as long as it keeps local before global

pepeiborra: I'll extend the comment later

michaelpj: Okay, so we're basically adding a second measure of completion "goodness", which is lexicographically less relevant than the text? That makes sense to me.

(Tangent: this makes me wonder whether we can skip name-based sorting on the server at all? I know you did a bunch of work making that faster... the fastest thing is not to do it! Perhaps some clients don't sort and so it's helpful to have some sort order from our end?)

michaelpj: Or, if it's just the local/nonlocal/global thing, we could just add 1/2/3 priority digits to the sort text when we construct them. The ordering here is a perhaps unnecessarily fine-grained. Not that it makes much difference.

michaelpj (approved):

ghcide/src/Development/IDE/Plugin/Completions.hs
pepeiborra: > (Tangent: this makes me wonder whether we can skip name-based sorting on the server at all? I know you did a bunch of work making that faster... the fastest thing is not to do it! Perhaps some clients don't sort and so it's helpful to have some sort order from our end?)

That was sorting the fuzzy matches by score - definitely still needed. We don't sort alphabetically otherwise

pepeiborra: > Or, if it's just the local/nonlocal/global thing, we could just add 1/2/3 priority digits to the sort text when we construct them. The ordering here is a perhaps unnecessarily fine-grained. Not that it makes much difference.

This would result in all the local completions first, regardless of how fuzzy, then the global. It would not be what we want. 

But we ought to make the fuzzy score visible to the LSP client in the sortText field, so that the sorting respects it.

pepeiborra:
I made some more changes this morning after realising that alphabetical ordering is incompatible with qualified completions, so I have ditched the multi-dimensional sort key and gone back to a flat numbered sequence.

michaelpj:
> I made some more changes this morning after realising that alphabetical ordering is incompatible with qualified completions, so I have ditched the multi-dimensional sort key and gone back to a flat numbered sequence.

Is it incompatible? It seems to me like this means that there are *three* lexicographically ordered sorting factors:
1. local/nonlocal/global
2. qualified/unqualified (or maybe this is first? not sure)
3. name

We can encode all this into the sort text if we want: an integer for 1, and integer for 2, and then the text for 3. 

(But I think I must be missing something, since it looks like you tried something like that already?)

pepeiborra:
> > I made some more changes this morning after realising that alphabetical ordering is incompatible with qualified completions, so I have ditched the multi-dimensional sort key and gone back to a flat numbered sequence.
> 
> Is it incompatible? It seems to me like this means that there are _three_ lexicographically ordered sorting factors:
> 
> 1. local/nonlocal/global
> 2. qualified/unqualified (or maybe this is first? not sure)
> 3. name
> 
> We can encode all this into the sort text if we want: an integer for 1, and integer for 2, and then the text for 3.
> 
> (But I think I must be missing something, since it looks like you tried something like that already?)

Yes, this is true, we could extend the product with another factor to track qualified suggestions. It just gets a bit tedious to encode the triple in a string for the alphabetical ordering used by the client. 

Basically, there are two ways that we can go about it: 

a. Extend our sorting to cover all 3 (currently it's missing the last one, name) and then do a flat sequential ordering.
b. Encode all 3 factors in a string for  the LSP client to sort alphabetically

It probably makes more sense to go with a, since we crop the list of suggestions by `maxFactor` before sending it, and we want to sort it optimally before cropping to ensure that we send the best suggestions.

michaelpj:
> It probably makes more sense to go with a, since we crop the list of suggestions by maxFactor before sending it, and we want to sort it optimally before cropping to ensure that we send the best suggestions.

Ooh, this is a great point, worth including in the comment. I think this alone basically means that the client _can't_ realistically sort the results, since it doesn't have access to the full list. So we might as well just force our ordering.

pepeiborra:
> > It probably makes more sense to go with a, since we crop the list of suggestions by maxFactor before sending it, and we want to sort it optimally before cropping to ensure that we send the best suggestions.
> 
> Ooh, this is a great point, worth including in the comment. I think this alone basically means that the client _can't_ realistically sort the results, since it doesn't have access to the full list. So we might as well just force our ordering.

The client can sort, but if we don't sort then the quality of suggestions will be worse. I have implemented a few more improvements (see PR description) but have not enabled alphabetical sorting since it breaks a lot of tests and I'm lazy

tomsmeding:
This is awesome! Thanks for the work!

> Match the first character case-sensitively - e.g. don't suggest `truncate` for `Tru`, only `True`

Is it reasonable to have this be configurable? I'd like to reduce my usage of the shift key, and I was very happy with the past behaviour where the whole match was case insensitive. If you consider this too much of a maintenance burden, that's fine too. :)

PR #2333, opened 2021-11-06
Trace diagnostics

pepeiborra:
Include rule diagnostics in traces for diagnosing diagnostic issues

Also disable checkProject in HLS test suite.
We already disable this in ghcide tests, since it introduces noise (traces are harder to read), and can potentially break tests too (e.g. eval plugin)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2333"></a>


jneira:
could we keep some tests with the option enabled? it is the default for now and maybe would be good to keep it covered in some way

pepeiborra:
> could we keep some tests with the option enabled? it is the default for now and maybe would be good to keep it covered in some way

There are tests that exercise this option in the ghcide test suite already. 

I'll have a look at the splice plugin test failures, maybe that test suite depends on this option being enabled for some reason

pepeiborra:
@jneira this is ready to merge now

jneira:
> 
> 
> @jneira this is ready to merge now

I am fixing required checks with #2337 so mergifyio will merge correctly this one as soon as the latter is merged

jneira (approved):
LGTM

PR #2335, opened 2021-11-09
Cancel prev runs for bench and nix

jneira:
* It is working really well to the test workflow

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2335"></a>

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,5 +1,10 @@
 name: Benchmark
 
+# See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency.
+concurrency:
+  group: ${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
 on:
   pull_request:
     branches:
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -19,7 +19,7 @@ jobs:
       - id: skip_check
         uses: fkirc/skip-duplicate-actions@v3.4.0
         with:
-          cancel_others: true
+          cancel_others: false
           paths_ignore: '["**/docs/**", "**.md", "**/LICENSE", ".circleci/**", "install/**", "**.nix", "**/test/**", "flake.lock", "**/README.md", "FUNDING.yml"]'
 
   bench_init:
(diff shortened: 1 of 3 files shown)


pepeiborra:
If I understand correctly this is a new feature of Github CI providing built-in support to deal with duplicate actions. 
So far we have been using a third-party action to do this (`fkirc/skip-duplicate-actions`).

> It is working really well to the test workflow

It looks like the test workflow now has both mechanisms enabled, is that what we want?

https://github.com/haskell/haskell-language-server/blob/9fa6abe2cb910bd07a3eeb8a751974d96dadea5e/.github/workflows/test.yml#L24-L28

https://github.com/haskell/haskell-language-server/blob/9fa6abe2cb910bd07a3eeb8a751974d96dadea5e/.github/workflows/test.yml#L8-L10

jneira:
> It looks like the test workflow now has both mechanisms enabled, is that what we want?

Oh i forgot to set it to false as i did for the test workflow, thanks. However we are using other features of skip-duplicate-actions so i would keep it around (like in test workflow)

pepeiborra:
> > It looks like the test workflow now has both mechanisms enabled, is that what we want?
> 
> 
> 
> Oh i forgot to set it to false as i did for the test workflow, thanks. However we are using other features of skip-duplicate-actions so i would keep it around (like in test workflow)

Can you elaborate on why we need both?

jneira:
> Can you elaborate on why we need both?

Cause we are using skip-duplicate-actions for skipping tests in the current workflow by file paths changed by the pr. It is interesting enough to keep it until we find an alternative.

pepeiborra:
So if we are keeping it, why not keep using it to cancel duplicate runs as it already does? In other words, what's the motivation for this PR?

jneira:
> 
> 
> So if we are keeping it, why not keep using it to cancel duplicate runs as it already does? In other words, what's the motivation for this PR?

Oh sorry, i should had linked the fisrt pr adding it to the test workflow: https://github.com/haskell/haskell-language-server/pull/2310

jneira:
uh oh funct-tests are hanging in windows 9.0.1 here :facepalm:

will disable it again

pepeiborra (approved):

michaelpj (approved):

Issue #1892, opened 2021-06-05
[GHC 9.0] No need to parse without Haddocks


pepeiborra:
Haddock markup parse errors are flagged as warnings in 9.0 (no longer errors) so there's no longer a need to parse twice

yoshitsugu:
@pepeiborra 
I'm now taking a look at this :eyes:  but what is the expected behavior? 
In the current test case, it looks like no output is expected above GHC90
https://github.com/haskell/haskell-language-server/blob/master/ghcide/test/exe/Main.hs#L562-L564

jneira:
> I'm now taking a look at this but what is the expected behavior?

I think the expected behaviour is the same but we can remove the double parsing *for ghc >= 9.0.1*, as haddock will not throw error for those ghc versions

here: 

https://github.com/haskell/haskell-language-server/blob/229faac11a517efdd75f7d23afce9daa6345259e/ghcide/src/Development/IDE/Core/Rules.hs#L893-L900

and here

https://github.com/haskell/haskell-language-server/blob/229faac11a517efdd75f7d23afce9daa6345259e/ghcide/src/Development/IDE/Core/Rules.hs#L197-L250

jneira:
```haskell
liftIO $ (fmap.fmap.fmap.fmap)
```` 

wonderful :sweat_smile:

konn:
> I think the expected behaviour is the same but we can remove the double parsing _for ghc >= 9.0.1_, as haddock will not throw error for those ghc versions

Then, perhaps also adding a regression test with erroneous Haddock comment(s) would make sense to ensure that HLS doesn't suffer from such malformed Haddock comments.

jneira:
> Then, perhaps also adding a regression test with erroneous Haddock comment(s) would make sense to ensure that HLS doesn't suffer from such malformed Haddock comments.

Hmm i think the linked test already does it: the haddock error is converted in a warning. So if it is reported incorrectly as an error the test would fail afaiu.

yoshitsugu:
@jneira @konn thanks for quick reply 😄 i'll try this.

jneira:
Thinking in this twice it is unfortunate that we are losing the haddock diagnostics for ghc >= 9.0as maybe some users find them useful, before they hit the error when building docs.
[...]

PR #2338, opened 2021-11-10
Skip parsing without haddock for above GHC9.0

yoshitsugu:
For #1892
Just skip by `#if`, is it enough for this issue? :thinking: 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2338"></a>


jneira:
> 
> 
> For #1892 Just skip by `#if`, is it enough for this issue? thinking
> 

Thanks for the pr, i think we could use runtime checks instead cpp as both branches compile with all ghcs

Runtime checks are here:

https://github.com/haskell/haskell-language-server/blob/7011d5e363bdfae0598d309bbb5d2386cd57e4b7/ghcide/src/Development/IDE/GHC/Compat.hs#L277-L299

Example here:

https://github.com/haskell/haskell-language-server/blob/613ec40efaeccdd8d3d8bd3ddae1b4d920e1189d/ghcide/session-loader/Development/IDE/Session.hs#L529

yoshitsugu:
@jneira oh i see. i will fix to use runtime check

jneira:
Looks good to me but i would like to know @pepeiborra thoughts

pepeiborra:
This looks reasonable to me, but does it actually work? 
1. Do we still collect haddocks in GHC 9? There seems to be some problem here, according to the testsuite. 
2. Do we no longer parse twice in GHC 9 when there is a Haddock error? The testsuite cannot verify this, but maybe you can use a trace to verify empirically.

yoshitsugu:
Sorry but I don't have enough knowledge to answer @pepeiborra' s questions. Could anyone give me any advice where to start to investigate about these questions? ( @jneira )

jneira:
> 
> 
> Sorry but I don't have enough knowledge to answer @pepeiborra' s questions. Could anyone give me any advice where to start to investigate about these questions? ( @jneira )

Hi, well there is a failing test which seems to be related with the pr:

https://github.com/haskell/haskell-language-server/runs/4165727646?check_suite_focus=true

```
ghcide
  get
    hover
      documentation                   #1129: FAIL (1.19s)
        test/exe/Main.hs:3705:
        failed to find: `Recognizable docs: kpqz` in hover message:
        
        '''haskell
        documented :: forall (m :: * -> *) a. Monad m => Either Int (m a)
        '''
        
        *Defined at /tmp/extra-dir-14222270143316/GotoHover.hs:46:1*
        
        Use -p '/documentation                   #1129/' to rerun this test only.
```

As mentioned by pepe maybe we were relying in the double parsing to effectively get docs so removing it seems to be not enough, i am afraid. The test fail has to be investigated and fixed.

Related with that is the second question. Add temporary some traces (using `Debug.Trace` for example) could help to:
* know why the test is failing (so to observe the behaviour when there are docs without errors, which should be collected and send in the hover information to the client)
* confirm we are not double parsing in front of docs with errors so we should get the test case with haddock errors and check the traces added inside the else branch are not printed.

yoshitsugu:
@jneira i see. thank you for your kind comment. i'll try.

yoshitsugu (commented):
@pepeiborra @jneira
I've fixed code and confirmed ghcide test code passed in my local machine. (It seems to fail testing in Windows environment on GitHub Actions, but I don't know whether it's required to pass. Please notify me if I need to confirm the Windows environemnt.)

For question 2, I confirmed not to parse twice in GHC 9.0.1 with Debug.Trace by inserting trace to `getParsedModuleDefinition` function.

Please point out if I misunderstood something or need to do something else.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -213,38 +213,43 @@ getParsedModuleRule =
...
-                -> pure (diagsM, resh)
-              -- If we fail to parse haddocks, report the haddock diagnostics as well and
-              -- return the non-haddock parse.
-              -- This seems to be the correct behaviour because the Haddock flag is added
-              -- by us and not the user, so our IDE shouldn't stop working because of it.
-              _ -> pure (diagsM, res)
+    -- We still parse with Haddocks whether Opt_Haddock is True or False to collect information
+    -- but we no longer need to parse with and without Haddocks separately for above GHC90.
+    res@(_,pmod) <- if Compat.ghcVersion >= Compat.GHC90 then
+      liftIO $ (fmap.fmap.fmap) reset_ms $ getParsedModuleDefinition hsc opt file (withOptHaddock ms)
```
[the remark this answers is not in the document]
pepeiborra: That's right, with GHC >9.0 we want to parse with `Opt_Haddock` on always

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -213,38 +213,43 @@ getParsedModuleRule =
...
+                liftIO $ (fmap.fmap.fmap) reset_ms mainParse
+            else do
+                let haddockParse = getParsedModuleDefinition hsc opt file (withOptHaddock ms)
+
+                -- parse twice, with and without Haddocks, concurrently
+                -- we cannot ignore Haddock parse errors because files of
+                -- non-interest are always parsed with Haddocks
+                -- If we can parse Haddocks, might as well use them
+                --
+                -- HLINT INTEGRATION: might need to save the other parsed module too
```
pepeiborra, line 238: this comment is confusing , I think it can be deleted since `GetParsedModuleWithComments` was introduced. What do you think @jneira ?

pepeiborra (approved):
The test failure looks unrelated,  testsuite is flaky in Windows. I don't see any problems with the change, looks good. Thanks for testing @yoshitsugu !

ghcide/src/Development/IDE/Core/Rules.hs
jneira: yeah, it was even not hlint specific and can be removed

yoshitsugu: I've removed this comment :+1:

jneira:
circleci build is stuck, it says there is no circleci config, which is not true, let's see what is doing in the next branch update

jneira:
I've opened a ticket with circleci support but the issue is not resolved so i am gonna merging it manually, as gh workflow is green

Ailrun:
@jneira Sometimes it does, and you can reactivate it by close and reopen PR.

yoshitsugu:
@pepeiborra @jneira thanks for advising me!

jneira:
> 
> 
> @jneira Sometimes it does, and you can reactivate it by close and reopen PR.

yeah, but this time create a manual workflow run in all pr's fails due to a false positive: 
```
We have detected an anomaly that violates the CircleCI Terms of Service.
We did not test this push because the project has been suspended. Please contact us if you think this is a mistake.
```

https://app.circleci.com/pipelines/github/haskell/haskell-language-server/4422/workflows/5aa08e6e-359e-4da8-8b8a-ed7a810935fe/jobs/38455

PR #2346, opened 2021-11-11
Use hackage version of czipwith

jneira:
* One less to finish 😅 
* I updated the living stack.yaml's and cabal-ghc912.project. cabal finds a plan with it but dont know if it compiles

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2346"></a>


cabal-ghc921.project
```diff
@@ -78,191 +61,3 @@ allow-newer:
...
-
-constraints:
-  Agda ==2.6.1.3,
-  Diff ==0.4.0,
-  EdisonAPI ==1.3.1,
-  EdisonCore ==1.3.2.1,
-  FPretty ==1.1,
-  HTTP ==4000.3.16,
-  HUnit ==1.6.2.0,
-  QuickCheck ==2.14.2,
```
pepeiborra, line 90: did you intend to delete all the constraints?

jneira: Yep, afaiu we dont need them having pinned the hackage index. They usually live in a cabal.project.freeze and no in the main one,. But maybe @fendor could expand the rationale behind

stack-9.0.1.yaml
```diff
@@ -36,12 +36,8 @@ extra-deps:
 - bytestring-encoding-0.1.0.0@sha256:460b49779fbf0112e8e2f1753c1ed9131eb18827600c298f4d6bb51c4e8c1c0d,1727
 - dependent-map-0.4.0.0@sha256:ca2b131046f4340a1c35d138c5a003fe4a5be96b14efc26291ed35fd08c62221,1657
 - dependent-sum-0.7.1.0@sha256:0e419237f5b86da3659772afff9cab355c0f8d5b3fdb15a5b30e673d8dc83941,2147
-- extra-1.7.9
 - floskell-0.10.5
```
jneira, line 39: extra-1.7.9 has a breaking change so no way
```suggestion
- extra-1.7.9
- floskell-0.10.5
```

cabal-ghc921.project
fendor: IIRC, these are constraints from head.hackage to make sure the patched versions are used. I *think* removal of them is fine.

fendor (approved):
LGTM!

cabal-ghc921.project
jneira: Oh, i see, maybe it would be nice to add another cabal-9.2.1.project.local with all the info needed to use head.hackage, the head.hackage repository info itself (it is advised in the head.hackage repo to set it per project/cabal.project.local) + constraints
So any developer could use such cabal-9.2.1.project.local out of the box

fendor: If you prefer an additional file containing it, fine by me. BTW, these constraints might be outdated, unsure how to best proceed with updating these.

jneira: `cabal freeze` having head.hackage as a overlay repo could help? maybe we could add the head.hackage repo in cabal-921.project if it is needed to make it work for sure (and remove it when patches are upstreamed)

And use directly the constraints it offer: https://ghc.gitlab.haskell.org/head.hackage/cabal.constraints 
Or use `head.hackage.sh init-local`

We also could check if `index-state` affects head.hackage so it could help to avoid the need of constraints

jneira:
As for November 16, circleci has set a timeout of 1 hour for jobs in free plans
Trying to increase resources to make the build faster as suggested by its support center

jneira:
> 
> 
> As for November 16, circleci has set a timeout of 1 hour for jobs in free plans Trying to increase resources to make the build faster as suggested by its support center

Thanks to using the large docker machine and being able to set 4 threads to build times are significantly better: 30 m without cache and 12 with it :tada:

Issue #2348, opened 2021-11-12
Want to kindly ask - is there a signature comments API somewhere


Anton-Latukha:
I'd loved to see those function "signature comments" displayed. HLS already helps to write them. I found them often being a true way of documenting what function does, as people often in function docs just white what functions accepts & does with args - people white that quite intuitively & that fits greatly to be a signature comments. To see those sometimes would be an immense simplification to read code.

Several years ago - I remember the dialog that Haddock-like signature comments were not yet in the LSP protocol specification.

What is the current status there (and here 8) )?

![Screenshot-2021-11-12-01:10:32](https://user-images.githubusercontent.com/20933385/141381982-3ea9b347-0ad4-437b-ad2d-ef37c67966a9.png)

As I arrived though reading `lsp-haskell` -> `lsp-ui` -> `lsp-mode` -> searching though `HLS` & looking at `Microsoft LSP specs` and seeing there as good as :bat: :dark_sunglasses: :sunglasses: :night_with_stars: :wavy_dash: :crescent_moon: :wavy_dash: & in that way I arrived here. & decided you should know something about what are the current affairs there 8)

Anton-Latukha:
Thanks for the direction & transfer. & Thanks for showing the place to look at in the spec.

Anton-Latukha:
Actually I happen to touch-related code yesterday. I already have some minor changes in the local branch.
It is `foundHover` -> `hover` -> `Spans/AtPoint.hs` `hoverInfo` - somewhere around there, trying looking it up ... <loading> ...

Anton-Latukha:
Probably GHC stores annotation info in https://hackage.haskell.org/package/ghc-9.2.1/docs/src/GHC.Iface.Ext.Types.html#NodeInfo

& `nodeInfo` retrieves it. So it is a question of where & how to show `nodeAnnotations` besides `nodeType`.

Anton-Latukha:
Seems like I almost pieced everything together. Only searching where the signature comments are hidden. Since https://hackage.haskell.org/package/ghc-9.2.1/docs/src/GHC.Iface.Ext.Types.html does not have anything on `Doc`(umenttaion) or `Com`(ment).

![Screenshot-2021-11-12-20:56:37](https://user-images.githubusercontent.com/20933385/141520320-b027f948-c79a-4259-ac72-435a92e57884.png)

I wonder, does GHC exposes signature comments or not, we could lived on Haddock parsing all this time.

[...]

PR #2349, opened 2021-11-13
Streamline code for getDocumentationTryGhc

Anton-Latukha:
<sub>at least trying to</sub>

Work towards #2348 & cleaning-up worked-with code on the go.

---

PR solves:
https://github.com/haskell/haskell-language-server/blob/bd0046b7089e32e309f7b29ee6ec1a808ca1db86/ghcide/src/Development/IDE/Spans/Documentation.hs#L64-L67

As `getDocumentationTryGhc` is the main code that retrieves the docs from GHC - the PR provides the element retrieval code.

---

This PR is a stage toward provisioning the argument docs into the LSP interface. PR addressed some underlying code.

---

Initial `master` code conditions:

* `getDocsBatch` had a doc:
https://github.com/haskell/haskell-language-server/blob/854d2c5a1aa3095c93e3f38ef3731be89b4ab09a/ghcide/src/Development/IDE/Core/Compile.hs#L994-L996

* `getDocsBatch` not optimized for batch retrieval, it does 1 element retrieval for every element. (Locates & loads a module interface, retrieves 1 element, locates & loads another module interface, retrieves 1 element ...).

* `getDocumentationsTryGhc` (plural `*s*`) - in the project *had only one use* - in `getDocumentationTryGhc` (single) & so was its implementation.

* So, over the project only (single) (meaning `getDocumentationTryGhc`) was used - & all doc requests to GHC - were 1 element requests, which got wrapped into singleton & processed as a list & by list function down the codepath.
* So `getDocsBatch` both was externally invoked for 1 elem & internally was processing lists per 1 elem. Batch processing optimization is addressed in #2371.

---

The results of the PR:
1. The non-interactive analog of GHCs `getDocs` - `getDocsNonInteractive` is created (for purposes mentioned in the initial doc above) it is designed for (all current requests) case of looking-up an entity.
2. Creation of `getDocsNonInteractive` allows to stop using `getDocumentationsTryGhcs` (plural) for `getDocumentationTryGhc` (single), which makes possible to remove unsafe `head` from the main code path.
3. Optimize the use case currently used everywhere.
3. Code shared between `getDocsNonInteractive` & `getDocsBatch` formed `getDocsNonInteractive'` & `initTypecheckEnv` which self-explained the code.

---

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2349"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -998,19 +998,19 @@ getDocsBatch
   :: HscEnv
   -> Module  -- ^ a moudle where the names are in scope
   -> [Name]
-  -> IO (Map.Map Name (Either String (Maybe HsDocString, Map.Map Int HsDocString)))
+  -> IO (Map.Map Name (Either T.Text (Maybe HsDocString, Map.Map Int HsDocString)))
 getDocsBatch hsc_env _mod _names = do
     ((_warns,errs), res) <- initTc hsc_env HsSrcFile False _mod fakeSpan $ Map.fromList <$> traverse findNameInfo _names
     case res of
         Just x  -> return $ fun x
         Nothing -> throwErrors errs
   where
-    fun :: Map.Map Name (Either GetDocsFailure c) -> Map.Map Name (Either String c)
+    fun :: Map.Map Name (Either GetDocsFailure c) -> Map.Map Name (Either T.Text c)
     fun =
       Map.map fun1
      where
-      fun1 :: Either GetDocsFailure c -> Either String c
-      fun1 = first $ T.unpack . showGhc
+      fun1 :: Either GetDocsFailure c -> Either T.Text c
+      fun1 = first showGhc
 
     throwErrors = liftIO . throwIO . mkSrcErr
 
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -991,11 +991,12 @@ mkDetailsFromIface session iface linkable = do
     initIfaceLoad hsc' (typecheckIface iface)
   return (HomeModInfo iface details linkable)
 
-fakeSpan :: RealSrcSpan
-fakeSpan = realSrcLocSpan $ mkRealSrcLoc (Util.fsLit "<ghcide>") 1 1
-
 initTypecheckEnv :: HscEnv -> Module -> TcRn r -> IO (Messages, Maybe r)
 initTypecheckEnv hsc_env mod = initTc hsc_env HsSrcFile False mod fakeSpan
+  where
+    fakeSpan :: RealSrcSpan
+    fakeSpan = realSrcLocSpan $ mkRealSrcLoc (Util.fsLit "<ghcide>") 1 1
+
 
 getDocsNonInteractive'
     :: Name
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -998,12 +998,12 @@ getDocsBatch
   :: HscEnv
   -> Module  -- ^ a moudle where the names are in scope
   -> [Name]
-  -> IO (Either ErrorMessages (Map.Map Name (Either T.Text (Maybe HsDocString, Map.Map Int HsDocString))))
+  -> IO (Either ErrorMessages (Map.Map Name (Either T.Text (Maybe HsDocString, Maybe (Map.Map Int HsDocString)))))
 getDocsBatch hsc_env _mod _names = do
     ((_warns,errs), res) <- initTc hsc_env HsSrcFile False _mod fakeSpan $ Map.fromList <$> traverse findNameInfo _names
-    case res of
-        Just x  -> pure $ pure $ fun x
-        Nothing -> pure $ Left errs
+    pure $ case res of
+        Just x  -> pure $ fun x
+        Nothing -> Left errs
   where
     fun :: Map.Map Name (Either GetDocsFailure c) -> Map.Map Name (Either T.Text c)
     fun =
@@ -1012,7 +1012,7 @@ getDocsBatch hsc_env _mod _names = do
       fun1 :: Either GetDocsFailure c -> Either T.Text c
       fun1 = first showGhc
 
-    findNameInfo :: Name -> IOEnv (Env TcGblEnv TcLclEnv) (Name, Either GetDocsFailure (Maybe HsDocString, Map.Map Int HsDocString))
+    findNameInfo :: Name -> IOEnv (Env TcGblEnv TcLclEnv) (Name, Either GetDocsFailure (Maybe HsDocString, Maybe (Map.Map Int HsDocString)))
     findNameInfo name =
         case nameModule_maybe name of
             Nothing -> return (name, Left $ NameHasNoModule name)
@@ -1026,7 +1026,7 @@ getDocsBatch hsc_env _mod _names = do
               pure . (name,) $
                 if isNothing mb_doc_hdr && Map.null dmap && Map.null amap
                 then Left $ NoDocsInIface mod $ compiled name
-                else Right (Map.lookup name dmap, Map.findWithDefault mempty name amap)
+                else Right (Map.lookup name dmap, Map.lookup name amap)
     compiled n =
       -- TODO: Find a more direct indicator.
       case nameSrcLoc n of
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -28,6 +28,7 @@ module Development.IDE.Core.Compile
   , loadInterface
   , loadModulesHome
   , setupFinderCache
+  , getDocsNonInteractive
   , getDocsBatch
   , lookupName
   ,mergeEnvs) where
@@ -990,12 +991,20 @@ mkDetailsFromIface session iface linkable = do
     initIfaceLoad hsc' (typecheckIface iface)
   return (HomeModInfo iface details linkable)
 
+fakeSpan :: RealSrcSpan
+fakeSpan = realSrcLocSpan $ mkRealSrcLoc (Util.fsLit "<ghcide>") 1 1
 
--- | Non-interactive modification of 'GHC.Runtime.Eval.getDocs'.
---   The interactive paths create problems in ghc-lib builds
----  and lead to fun errors like "Cannot continue after interface file error".
-getDocsNonInteractive :: Name -> IOEnv (Env TcGblEnv TcLclEnv) (Name, Either GetDocsFailure (Maybe HsDocString, Maybe (Map.Map Int HsDocString)))
-getDocsNonInteractive name = do
+initTypecheckEnv :: HscEnv -> Module -> TcRn r -> IO (Messages, Maybe r)
+initTypecheckEnv hsc_env mod = initTc hsc_env HsSrcFile False mod fakeSpan
+
+getDocsNonInteractive'
+    :: Name
+    -> IOEnv
+        (Env TcGblEnv TcLclEnv)
+        (Name,
+        Either
+            GetDocsFailure (Maybe HsDocString, Maybe (Map.Map Int HsDocString)))
+getDocsNonInteractive' name =
     case nameModule_maybe name of
         Nothing -> return (name, Left $ NameHasNoModule name)
         Just mod -> do
@@ -1007,7 +1016,7 @@ getDocsNonInteractive name = do
                 <- loadModuleInterface "getModuleInterface" mod
             let
                 isNameCompiled =
-                    -- TODO: Find a more direct indicator.
+                    -- comment from GHC: Find a more direct indicator.
                     case nameSrcLoc name of
                         RealSrcLoc {}   -> False
                         UnhelpfulLoc {} -> True
@@ -1015,6 +1024,15 @@ getDocsNonInteractive name = do
                 if isNothing mb_doc_hdr && Map.null dmap && Map.null amap
                     then Left $ NoDocsInIface mod isNameCompiled
                     else Right (Map.lookup name dmap, Map.lookup name amap)
+
+-- | Non-interactive modification of 'GHC.Runtime.Eval.getDocs'.
+--   The interactive paths create problems in ghc-lib builds
+---  and lead to fun errors like "Cannot continue after interface file error".
+getDocsNonInteractive :: HscEnv -> Module -> Name -> IO (Either ErrorMessages (Name, Either GetDocsFailure (Maybe HsDocString, Maybe (Map.Map Int HsDocString))))
+getDocsNonInteractive hsc_env mod name = do
+    ((_warns,errs), res) <- initTypecheckEnv hsc_env mod $ getDocsNonInteractive' name
+    pure $ maybeToEither errs res
+
 
 -- | Non-interactive, batch version of 'GHC.Runtime.Eval.getDocs'.
 getDocsBatch
@@ -1024,12 +1042,9 @@ getDocsBatch
   --  2021-11-18: NOTE: Map Int would become IntMap if next GHCs.
   -> IO (Either ErrorMessages (Map.Map Name (Either GetDocsFailure (Maybe HsDocString, Maybe (Map.Map Int HsDocString)))))
   -- ^ Return a 'Map' of 'Name's to 'Either' (no docs messages) (general doc body & arg docs)
-getDocsBatch hsc_env _mod _names = do
-    ((_warns,errs), res) <- initTc hsc_env HsSrcFile False _mod fakeSpan $ Map.fromList <$> traverse getDocsNonInteractive _names
+getDocsBatch hsc_env mod names = do
+    ((_warns,errs), res) <- initTypecheckEnv hsc_env mod $ Map.fromList <$> traverse getDocsNonInteractive' names
     pure $ maybeToEither errs res
-
-fakeSpan :: RealSrcSpan
-fakeSpan = realSrcLocSpan $ mkRealSrcLoc (Util.fsLit "<ghcide>") 1 1
 
 -- | Non-interactive, batch version of 'InteractiveEval.lookupNames'.
 --   The interactive paths create problems in ghc-lib builds
@@ -1039,7 +1054,7 @@ lookupName :: HscEnv
            -> Name
            -> IO (Maybe TyThing)
 lookupName hsc_env mod name = do
-    (_messages, res) <- initTc hsc_env HsSrcFile False mod fakeSpan $ do
+    (_messages, res) <- initTypecheckEnv hsc_env mod $ do
         tcthing <- tcLookup name
         case tcthing of
             AGlobal thing    -> return thing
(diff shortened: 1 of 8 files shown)


Anton-Latukha:
https://github.com/Anton-Latukha/haskell-language-server/blob/19f65cb391e781ffba3dbecaedead01c3e967317/ghcide/src/Development/IDE/Spans/Documentation.hs#L71-L79

That IO gathering/handling smells fishy to me, it catches IO exceptions from `getDocsBatch`.

While:
https://github.com/Anton-Latukha/haskell-language-server/blob/19f65cb391e781ffba3dbecaedead01c3e967317/ghcide/src/Development/IDE/Core/Compile.hs#L964-L982

There is a lot of `Maybe` & `Either` already around it & `IO` exceptions handling. Code throws exception just to catch it immediately. The amount of exception handling seems excessive & probably can be reduced. & the use of the IO monad to throw & catch exceptions ...

Anton-Latukha:
`SpanDoc` needs to be changed, as:
```haskell
getDocumentationsTryGhc :: HscEnv -> Module -> [Name] -> IO (Map.Map Name SpanDoc)
```
While `SpanDoc` does not support any form of the arg docs:
https://github.com/haskell/haskell-language-server/blob/07042d404340017b473833c1fb470e3c25fdf6d2/ghcide/src/Development/IDE/Spans/Common.hs#L62-L67

I also do not understand what is such `Span*` about it - it literally does not holds any `Span` information. It is for on `Uri`s.

Anton-Latukha:
`SpanDoc` seems like a premature type. It does nothing except wrapping things into 2 types of constructors.

It is there to be come:
```
type DocMap = NameEnv SpanDoc
```

Where `NameEnv` is just: https://hackage.haskell.org/package/ghc-9.2.1/docs/GHC-Types-Name-Env.html#g:1

Which builds a map of unique names to their docs. Sounds familiar?
As we get `Name` from GHC - it is already guaranteed to be unique, so it does not need a uniqueness guard in `NameEnv`.

While:
```
data DocAndKindMap = DKMap {getDocMap :: !DocMap, getKindMap :: !KindMap}
```

Also seems strange. Main docs & arg docs - are one things & would be realistically always requested together.

While Kind map - is a complete thing of itself - it is called when Kind needs to be retrieved/shown. It can be used in the hover docs info, but not really, what gets used is type signature & concrete inferred type signature. From that, I infer that `DocAndKindMap` is a bad abstraction, moreover that it is not `*Map`, it is a sum type of 2 separate maps, while every Name is an index in itself (so logical would've been then to `Map Name (Doc, Kind)`.

`SpanDoc -> DocMap -> DocAndKindMap`

Moreover - those `Doc` types does not account for `ArgDocs`, so to implement functionality properly - they would need to be changed anyway. & so far I do not see the reasoning to disassemble maps, assemble them again while implicitly hiding that are maps & bundle unrelated maps together. Would make it waay simpler. Would just pass map until the proper way to do it & in later GHCs those maps become GHC documentation types.

Anton-Latukha:
Opened #2371 - with a bulking request code.

I currently think singleton process should be kept, but element retrieval should be written as it is used - for one element: `getDocumentationTryGhc` & its `getDocsNonInteractive`. - that is probably nice for `hover`-like docs retrieval (main agenda)

& separately have proper bulk-optimized retrieval: `getDocumentationsTryGhc` & its `getDocsBatch`.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -474,13 +473,13 @@ generateHieAsts hscEnv tcm =
...
-        , DL.fromList $ map fst declSplices
-        , DL.fromList $ map fst awSplices
-        ]
+  fold
+    [ map fst exprSplices
+    , map fst patSplices
+    , map fst typeSplices
+    , map fst declSplices
+    , map fst awSplices
+    ]
```
Anton-Latukha, line 482: Thinking that marshaling the: `[[xs]] -> n fromList -> DList.concat -> toList -> [xs]` - is controversial. I noted superfluous back & forth conversion & looked into official docs.

The commit message point to the official docs which mention that conversion to/fromList are also `n` constly, in https://hackage.haskell.org/package/dlist-1.0/docs/Data-DList.html

> `fromList`: This function is implemented with '++'. Repeated uses of @fromList@ are just as inefficient as repeated uses of '++'.

Because DList Haddocks do not have tags to point into `to/fromList` - in the commit message I could only point to the library source.
 
So the code was: every `fromList` is `++` (means `<>` of `[]` ) already & then with additionally added `DList.concat` on top for them. So, more optimal is just fold with `++`, which is `fold` on `[]`.

Anton-Latukha:
Since HLS uses per-1-element retrieval from GHC:

The implementations for according mode are now provided.

And switched the code into it.

HLS runs loads docs from module interfaces.

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -152,9 +154,9 @@ typeRowToLoc (row:.info) = do
 
 documentHighlight
   :: Monad m
-  => HieASTs a
-  -> RefMap a
-  -> Position
+  => HieASTs a -- ^ get AST
+  -> RefMap a -- ^ get map of identifiers to spans
+  -> Position -- ^ get position
```
Ailrun, line 159: What do you mean by "get" here?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -386,8 +394,11 @@ defRowToSymbolInfo (DefRow{..}:.(modInfoSrcFile -> Just srcFile))
     end   = Position (defELine - 1) (defECol - 1)
 defRowToSymbolInfo _ = Nothing
 
-pointCommand :: HieASTs t -> Position -> (HieAST t -> a) -> [a]
-pointCommand hf pos k =
+pointCommand :: HieASTs t -- ^ Get AST
+  -> Position -- ^ point at position
+  -> (HieAST t -> a) -- ^ extact data with a getter
+  -> [a] -- ^ all data recieved
```
Ailrun, line 400: I don't think these comments are very informative.

Ailrun (commented):
Thank you for the update; However, some changes seem not so useful and somewhat inconsistent, especially some comments and type signatures.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: That the argument gets applied to a function. Or function takes an argument.

---

Simple `get` helps mentally read docs as text, the function gets this this and this, returns that. Terminology of course can be disputed, aka function applies to argument or the other way around, & do Haskell function return a value or other terms can be used.

Of course to say is: "bind the ... argument to the ... parameter of a function"
But for simple dirt-cheap basic explanation - a simple explanation of how arguments are treated semantically, how combination produces a result - documenting it arguments works great.

Since programming is not about objects, but about processes - I try to turn object naming documentation (which type signatures "as is" are) into process describing.

---

The main thing to explain in the function is how to think about `HieASTs` plural here. For the purpose of the function it is not important are there many ASTs or we pretend we have 1 giant complex AST of everything. So just `AST`.

Next question is what `RefMap` is & how to think about it in the scope of the function. Function reminds & at the same time simplifies to the reader what obscure `RefMap` name is about. That it is important to see that is it simply a map from some identifiers to some spans.

In that, we connected the reader that some AST is taken & then identifiers of it get mapped to spans.

Then the position is taken. Probably here it needs to be additionally cleared-out that that position determines an identifier that gets highlighted.

The return type of the function is self-explanatory, `DocumentHighlight` - is a great self-explanatory name whose semantic meaning directly corresponds with the agenda of the function.

Ailrun: I mean, if `get` is nothing but for readability, then these comments are saying `AST` for `HieASTs`, `map of identifiers to spans` for `RefMap`, and `position` for `Position`. These are mostly information-less. Why do we even want to give a doc if it does not say more than the type?

1. If you want to emphasize such point, you should emphasize that. Otherwise, just saying `AST` is redundant.
2. `RefMap` type is possibly worth some explanations, but your explanation applies to the type, not this argument. If you want to comment something on the type, it's better to write it on the type.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: But for me they are.

The lack of information (or superficiality of information) in the documentation for the reader - means that the next reader would find new information & would add it into documentation.

I am just trying to doc functions so I understand them & later can work with them.

I needed to go look into the sources of all these arguments. Sometimes I can do just jump to definition, sometimes I need to go Hoogle & frequently Hoogle returns a GHC 9.2.1 interface & I develop in 8.10.7, so I need to go find the definitions/sources manually to look into them.

I am a newb in the project.

And the docs are for newbs. When person implicitly knows what the function does, or they understand all they need to know from the type signature - on that level of understanding docs are not needed & so the person does not look into docs.

So if left untouched the docs tend to go from nothing there to something there, converging on the medium level, since above medium level understanding no longer needs to look-up medium level info or add to it.

For me - it is useful information. `point` helps to indicate that `Position` is a main topic of the function, that `(HieAST t -> a)` is a getter that extracts info from `AST` based on `Position` & returns plural.

& noting that `point` is `Position` also helps to understand what function is about.

Because signature can have different implementations. It can take AST & `(HieAST t -> a)` is a pointer to something & `Position` is a filter identifier & then the body of function is a `Command`.

Doc explains that `pointCommand` is a command when user points at some position - information is retried through getter.

The question generally is not to remove documentation, but how to modify it so it optimally explains the function. Before this, I did not understand how `pointCommand` & `(HieAST t -> a)` combine.

Ailrun: > The question generally is not to remove documentation, but how to modify it so it optimally explains the function.

If they are not informative, I'm more inclined to *remove* documentations. They give readers extra clutters to read. Especially that `getter` part. It can compute anything from `HieAST`. It is in a sense "getter", but it explains nothing more than the actual type. The type already says that the argument is the function which accepts AST and computes something.

One issue of these comments, IMO, is that the target of explanation is unclear. For example, `point at position` comment. This one does not describe `Position` argument, but some `point` (which you probably meant to refer to `point` of `pointCommand`).
    - If the description is for the position argument, please describe it. For example, `Position of the target point`, or `Position where the command will be executed`, etc. Although I don't think even these are good comments, they at least explain the position argument, not "point".
    - If the description is for the entire function, why do you put those comments on the arguments? You can put a document for the function itself.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: Easy understandability & readability for docs are important. Otherwise, if it to transmit information by Shannon boundary (a mathematically maximally dense form of information) was optimal for agenda - we'd would not needed to learn protocols to correspond, languages & then we can just send people to read the source code or the binary directly.

---

In the doc for function `RefMap` purposefully gets oversimplified.
Real type is:
```haskell
type RefMap a = Map.Map Identifier [(Span, IdentifierDetails a)]
```

But doc purposefully says it is "a map identifiers to spans", not mentioning anything about `IdentifierDetails`. Docs are to streamline understanding & so because highlighting happens not based on `IdentifierDetails` but based on spans - docs forget to mention details unimportant to understanding & agenda. As a result - docs forget to mention unimportant details, but a person looking-up the meaning of a function hopefully would understand/recap it through reading the tiny doc.

Any statement is an emphasis in itself. Then generalized type names from a remote location, function doc is closer in information & scope to the purpose of the function putting emphasis, people tend to read the docs of a function they have interest first & think in the scope of a function. Additionally, if doc successfully conveys the process/meaning - the successful system of mental meaning/understanding is the emphasis how to read/understand the type names in the scope of the function.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Ailrun: Actually I'm not sure what you mean by
> Because signature can have different implementations. It can take AST & (HieAST t -> a) is a pointer to something & Position is a filter identifier & then the body of function is a Command, or other readings of the signature.

Your comments are not any better than the types. If `HieAST t -> a` can be a pointer, than that "getter" can also be a pointer. If `Position` is a filter identifier, then what's the difference of `position of point` (I swapped the words so that it actually explains the argument)?

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: > This does not justify commenting types on a function.
It does not comment types on the functions. Those are docs to arguments of a function, even Haddock internally calls them comments for arguments: https://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields, https://haskell-haddock.readthedocs.io/en/latest/markup.html#function-arguments.

It comments what argument are for & about, arguments happen to have/be related to types closely & so the type semantics are present in the documentation for arguments, but bears the function argument scope.

---

> If you think that this function is hard to understand without any comments, then it is more helpful to put a comment directly relevant to this function

The meaning of the function documentation is to be readable/looked-up from the external scope. If to hide explanations inside the code - it defeats the purpose of documentation.

---

By the way, because no suggestions were given - but weak points of documentation were found - documentation get new edition.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: That is not a final form of documentation, documentation can be easily modified.

With documentation, both sides need to propose improvements. So far received only "remove" suggestions, which is not helpful to improve documentation, if to have an inverse Socratic reflection.

Because I get not suggestions of how to improve the situation - I need to improve it myself, which results in new versions of the documentation, while old thread addressed the previous version.

Anton-Latukha:
All I want here - is to be able point to function with complex type signature & argument types - & pop-up doc in as few words as possible to explain me what function does.

Since the function is `\ a b c -> process a b c` - a main part of the function is how arguments get used by the process.

People frequently write documentation as: "Function takes ASTs, takes a getter of data, takes a Position, returns the data", or "Returns data of a position, retrieving it by getter from AST" - notice how that documentation also restates the arguments retrieved & the type signature.
That kind of documentation is prevalent across the docs, see `concat` or `lookup` or `filter` - a lot of docs pretend type signature is not there & restate what is passed as args & mentioned in the type signature.
That kind of documentation people place in the wrong spot - it is a documentation for arguments. & the main doc body should convey the general meaning of a function, or details of implementation/use.

Ailrun:
1. All those examples you listed there explains what function does as a function document, not as documents on arguments. Moreover, none of them explains, for example, what `Foldable` is. `Foldable`, or any other types, should be documented as their own, not as a part of a function document.
2. Likewise, none of them uses `[(a, b)] -- key association we will lookup` or so. This is what you write in this PR, and this is what I'm not inclined to.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Ailrun: > It comments what argument are for & about

`map of identifiers to spans`, the part of your comment, is exactly the explanation of the type. If you want to explain what the argument is for/about, you should at the very least mention why this function needs such an identifiers-to-span map.

> If to hide explanations inside the code

What do you mean? I haven't mentioned anything to do with "hide explanations". I'm literally suggesting you to *expose* the explanation by putting it as a function doc comment.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Ailrun: > which is not helpful to improve documentation

As I said, putting uninformative documentation goes against *code quality*. Surely, it does not improve *documentation*, but it at least protects *code quality*.

> Because I get not suggestions of how to improve the situation

I cannot suggest you a comment that helps you to understand, as I am not you. What I can do is just arguing against bad and possibly unnecessary comments.

Let me summarize my comments.

1. If you want to explain a function, please use a function doc comment.
2. If you want to explain an argument, please explain the argument, not its type.
3. If you want to explain an argument type, please put that on the type definition, not on the usage.
4. Even in any of the above cases, please do not just repeat what already being said in a type or a function name. For example, in the case of `Position`, one possibility is that mentioning that `Position` is actually where we find a smallest containing node to which we apply those "getter"s.

Ailrun:
> The contributor works on the code to show the argument documentation. That people before had no problem not having the argument documentation & this contributor shown-up to care about it - shows that people have different understanding of the importance of argument comments & that context transmits that the person cares about argumentation comments enough to go study the project & contribute code.

I'm not against this. I'm just asking you to provide comments that makes sense.

> Especially if to threaten contributor to reject the code work basing on the little comments he did to explain to oneself the functions. That is counterproductive.

At this stage, I'm not sure that you just want to debate against me or want to improve the code quality. Are those comments only to explain it to yourself? Then will you delete those once this PR reaches a non-draft stage? If not, then that's why you need to convince others (at least me) that every developer *should* show your *personal* understanding.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: It is already exposed. It is a Haddock argument documentation. It is just a part of documentation that is not supported by HLS so far & showing those docs is what I am working on & what this PR is an initial stage of.

Ailrun: Again, what do you mean by "to hide"? I don't follow the context. It feels like you randomly mentioned that hiddeness.

Anton-Latukha: >  then it is more helpful to put a comment directly relevant to this function

I've read it as supplying them as comments in the function.

Anton-Latukha:
Oh, I'm sorry.

> I will definitely reject the change as it goes directly against "optimal" information.

I understood it wrong, so overreacted to it.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Ailrun: Oh I see, sorry for confusing words. I meant as a doc comment of this function.

Anton-Latukha: Ok. Deleted the docs. Done.

ghcide/src/Development/IDE/Spans/AtPoint.hs
Anton-Latukha: Ok. Deleted the docs. Done.

Anton-Latukha:
After the merge it can be thought - if the case can be explained/constructed - maybe to request GHC include `getDocsNonInteractive` into the [`GHC.Runtime.Eval`](https://hackage.haskell.org/package/ghc-9.2.1/docs/src/GHC.Runtime.Eval.html#getDocs). So local implementation would serve for old GHCs & then phased-out.

Anton-Latukha:
Since this function was mentioned in https://github.com/haskell/haskell-language-server/pull/2503#issuecomment-998126155 as preferable,

asking @pepeiborra to review, would ask to account that `getDocsBatch` changes would be shipped in the #2371.

pepeiborra (changes requested):
Please break this PR down further for easier review: 

1. Extract all the pure refactorings into a standalone PR 
2. Break down the documentation changes in one or more PRs.

Anton-Latukha:
[skip circleci]

Issue #1560, opened 2021-03-12
Feature idea: alternate format for numeric literals


alanz:
The rust language server has a code action that provides alternative formats for numeric literals, in different bases.

<img width="528" alt="Screenshot 2021-03-12 at 10 01 06" src="https://user-images.githubusercontent.com/409607/110924593-22226580-831a-11eb-85b0-edb893c76386.png">

This might be something to consider for haskell

jneira:
@drsooch great, thanks for willing to contribute. The docs i've found about numeric literals are:
* numeric literals from the haskell 2010 language report: https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-190002.5
* the ghc extensions about literals: http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/literals.html
  * extensions should be enabled to allow the different literal syntaxes

PR #2350, opened 2021-11-13
HLS Plugin to provide Alternate Literal Formats.

drsooch:
The following PR is for issue https://github.com/haskell/haskell-language-server/issues/1560. Currently all functionality is "complete", however there is a lot of dead code (code that exists but the plugin doesn't need to use). I am looking to get some feedback on the plugin itself, and hopefully through the feedback most of the dead code can be culled. 

**General Mechanics:**

A new rule `CollectLiterals` traverses the GHC source tree to look for literals (`Literals.hs`). The rule returns the exhaustive list of literals and also a list of `FormatType`. `FormatType`'s are derived from currently active extensions as well as a base set that are valid no matter the extensions active. 

When a user invokes a `codeAction` keybind, the `alternateNumberFormat` handler will filter out only the literals that are in the current range of the invoked `codeAction`. In the handler, the logic for generating possible alternatives is done, and each possibility is transformed into a `codeAction` sent to the user.

Most of the alternate formats are provided by the `Numeric` package (show* functions).

HLSInt

HLSFrac

**Potential Issues:**

- `Literals.hs` traverses nearly the entire source tree just to make sure all of the Literals are captured. Certain types of literals end up being discarded (Literals with no source text attached or source location). When `codeAction`s are generated we just immediately throw those out. As a result the `Literal` constructors of `IntPrimLiteral` and `FracPrimLiteral` aren't actually used.
- Similarly, in `Conversion.hs` we transform the Extensions `NumericUnderscores` and `NegativeLiterals` (`LexicalNegation` in GHC > 9), into `NoFormat` which essentially means we don't care about them right now. I'm not sure if generating cases for this would be useful or needed. (`NegativeLiterals` may not actually be useful to this plugin)
- The `NumDecimals` extension allows for literals such as `1.23e5` to be used places where Integer's are expected. I could've gone ahead and defaulted to using the Numeric package, but I tried to encode some logic to display different possible exponents. Ex. `12.3e4`, `123.0e3`, and `1.23e5` for `123000`.
- I believe all of the necessary code changes for files such as `cabal.project` and `haskell-language-server.cabal` but please let me know if there is more to do.
- I didn't attach any bounds to the plugins `cabal` file as I was using Stack to build and I am not quite clear on how to generate these bounds effectively.
- There is (presumably) no compatibility with other GHC versions other than the one I used (8.10.7). I'm not sure how to go about fixing those issues (other than just building with each possible GHC). 
- I have no idea why my Test setup is not working. I modeled the testing functionality after explicit imports plugin, but for whatever reason, the test server does not find ANY codeActions let alone just alternateNumberFormat plugins.

Let me know what you think. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2350"></a>


stack.yaml
```diff
@@ -25,6 +25,7 @@ packages:
   - ./plugins/hls-pragmas-plugin
   - ./plugins/hls-module-name-plugin
   - ./plugins/hls-ormolu-plugin
+  - ./plugins/hls-alternate-number-format-plugin
```
Ailrun, line 28: You need to update other `stack-*.yaml` files too.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
```diff
@@ -0,0 +1,401 @@
+{-# LANGUAGE DeriveGeneric     #-}
```
fendor, line 1: I think this whole module can be replaced by a fitting call to [`syb`](https://hackage.haskell.org/package/syb).

plugins/hls-alternate-number-format-plugin/test/Main.hs
```diff
@@ -0,0 +1,88 @@
...
+
+main :: IO ()
+main = defaultTestRunner test
+
+alternateNumberFormatPlugin :: PluginDescriptor IdeState
+alternateNumberFormatPlugin = AlternateNumberFormat.descriptor "alternateNumberFormat"
+
+test :: TestTree
+test = testGroup "alternateNumberFormat" [
+    codeActionHex "TIntDtoH" 3 13]
```
fendor, line 23: you should run `stylish-haskell` on your codebase.

If this is already `stylish-haskell` formatted, ignore this comment.

plugins/hls-alternate-number-format-plugin/hls-alternate-number-format-plugin.cabal
```diff
@@ -0,0 +1,61 @@
...
+license-file:       LICENSE
+author:             Nick Suchecki
+maintainer:         nicksuchecki@gmail.com
+category:           Development
+build-type:         Simple
+extra-source-files: LICENSE
+
+library
+  exposed-modules:  Ide.Plugin.AlternateNumberFormat
+  other-modules:    Ide.Plugin.Literals, Ide.Plugin.Conversion
```
fendor, line 18: You use `Ide.Plugin.Conversion` in your tests, so it needs to be exposed.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
```diff
@@ -0,0 +1,170 @@
...
+
+instance NFData AnyFormatType
+
+alternateFormat :: [FormatType] -> Literal -> [Text]
+alternateFormat fmts lit = case lit of
+  IntLiteral _ (Just _) val  -> concatMap (alternateIntFormat val) (removeCurrentFormat lit fmts)
+  FracLiteral _ (Just _) val -> if denominator val == 1 -- floats that can be integers we can represent as ints
+      then concatMap (alternateIntFormat (numerator val)) (removeCurrentFormat lit fmts)
+      else concatMap (alternateFracFormat val) (removeCurrentFormat lit fmts)
+  _                          -> ["Uh Oh"] -- This means there is no Source Text so we just ignore it
```
fendor, line 51: While I actually chuckled while reading this, we should not use such magic strings to indicate errors.
If this is shown to a user, it is a bad error message.
If this is never shown to a user, it should not be an actual value and rather use `Maybe [Text]`, imo.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
```diff
@@ -0,0 +1,170 @@
...
+intFormats = map snd intPairs
+
+fracPairs :: [(Extension, FormatType)]
+fracPairs = [(HexFloatLiterals, FracFormat HexFloatFormat)]
+
+fracFormats :: [FormatType]
+fracFormats = map snd fracPairs
+
+-- Generate up to 3 possible choices where:
+-- dropWhile (\d -> val `div` d) > 1000) implies we want at MOST 3 digits to left of decimal
```
fendor, line 126:
```suggestion
-- | Generate up to 3 possible choices where:
-- dropWhile (\d -> val `div` d) > 1000) implies we want at MOST 3 digits to left of decimal
```

fendor: This way it is haddock compliant and on hover you actually see the comment.

Additionally, check for valid Haddock syntax so it renders properly.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
```diff
@@ -0,0 +1,170 @@
...
+-- takeWhile (val >) implies we want to stop once we start to get numbers like: 0.1e[N]
+-- take 3 implies we want at most three choices which will center around the format:
+--    - 500.123e4
+--    - 50.0123e5
+--    - 5e.00123e6
+-- NOTE: showEFloat would also work, but results in only one option
+toNumDecimal :: Integer -> [Text]
+toNumDecimal val = map (showNumDecimal val) $ take 3 $ takeWhile (val >= ) $ dropWhile (\d -> (val `div` d) > 1000) divisors
```
fendor, line 134: Given the comment above, this feels like a slightly inaccurate name for the function (sorry, I know, I am nitpicky)

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
```diff
@@ -0,0 +1,170 @@
...
+hexFloatRegex = "0[xX][a-fA-F0-9]+\\.[a-fA-F0-9]+(p[+-]?[0-9]+)?"
+binaryRegex = "0[bB][0|1]+"
+octalRegex = "0[oO][0-8]+"
+decimalRegex = "\\d+"
+numDecimalRegex = "[0-9]+\\.[0-9]+[eE][+-]?[0-9]+"
+
+-- we export the regex for tests, but we want to be explicit in our matches
+-- so we need to match the beginning/end of the source text
+wrap :: Text -> Text
+wrap regex = "^" <> regex <> "$"
```
fendor, line 86: If this is exported, you should probably document it via a haddock comment so other users have an easier time figuring out what this does.

Also, considering this function is going to be exported, `wrap` does not really help with understanding what this does.
Maybe `matchLineRegex` or something like that expresses the intent better?

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,141 @@
...
+                                                       toFormatTypes)
+import           Ide.Plugin.Literals                  (Literal (..),
+                                                       collectLiterals,
+                                                       getSrcSpan, getSrcText)
+import           Ide.Plugin.Retrie                    (handleMaybe,
+                                                       handleMaybeM, response)
+import           Ide.Types
+import           Language.LSP.Types
+import           Language.LSP.Types.Lens              (uri)
+import           Prelude
```
fendor, line 39:
```suggestion

```

fendor (changes requested):

fendor (changes requested):
Awesome, this is a great plugin! I have some general comments and ideas but feel free to disagree.

I've looked into your testsuite issue, but can't tell right now what the issue is.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
fendor: Shout if you require assistance here

jneira:
Many thanks for working on this! It looks great overall

> * I didn't attach any bounds to the plugins `cabal` file as I was using Stack to build and I am not quite clear on how to generate these bounds effectively.

A usual way is use `cabal gen-bounds` but we are not being too strict with bounds in other plugins so i would follow what other plugins are doing to get consistenncy between them, take a look to:

https://github.com/haskell/haskell-language-server/blob/ab46fe8d06d8d9ee745592c57d92fc764e9d8995/plugins/hls-splice-plugin/hls-splice-plugin.cabal#L32-L50

It has bounds on base, ghcide and hls-plugin-api and nothing more
 
> * There is (presumably) no compatibility with other GHC versions other than the one I used (8.10.7). I'm not sure how to go about fixing those issues (other than just building with each possible GHC).

If you have the chance of building your version for other ghc versions (f.e 8.6.5, 8.8.4 and 9.0.1) it would be great. But We have the ci covering all supported versions so you can leverage it. I would add a line like

https://github.com/haskell/haskell-language-server/blob/ab46fe8d06d8d9ee745592c57d92fc764e9d8995/.github/workflows/test.yml#L187-L189

> * I have no idea why my Test setup is not working. I modeled the testing functionality after explicit imports plugin, but for whatever reason, the test server does not find ANY codeActions let alone just alternateNumberFormat plugins.

Well, a good way to trace test problems is enable full log output of lsp messages, take a look to the test suite linked just above and use the same invocation locally (the relevant part is `LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true`)

stack.yaml
drsooch: Will do that.

plugins/hls-alternate-number-format-plugin/test/Main.hs
drsooch: Yeah, I normally use Brittany, and for whatever reason `stylish-haskell` seems to just accept this?

plugins/hls-alternate-number-format-plugin/hls-alternate-number-format-plugin.cabal
drsooch: Yep, not sure how I missed that originally.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
drsooch: That should be an empty list, I wanted to make sure I wasn't hitting the rest of the Patterns and forgot about it...

drsooch:
> > * I have no idea why my Test setup is not working. I modeled the testing functionality after explicit imports plugin, but for whatever reason, the test server does not find ANY codeActions let alone just alternateNumberFormat plugins.
> 
> Well, a good way to trace test problems is enable full log output of lsp messages, take a look to the test suite linked just above and use the same invocation locally (the relevant part is `LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true`)

I ran the tests with logs enabled. Looks like an exception occurs (which I think is relevant?) and then a shutdown message is sent to the server.
AsyncCancelled

It looks like it's attempting to do some actual work but for whatever reason a shutdown message is sent?

ShutdownMessage

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
drsooch: So just taking a look briefly at syb and the (very slim documentation, I found one paper related to it). The idea is to generify (is that word) each `traverseLBlah` into just a few functions that unwrap data constructors into it's components and recurses until `LHsExpr` is encountered?

fendor: Basically yes, in your particular issue, I think you can get away with only matching on the leaf constructors (e.g. your literal constructors).  I *think* `everywhereM` is useful for you.

jneira:
@drsooch i have taken the liberty of add the plugin to the rest of cabal.project's (for ghc 9.0.1 and 9.2.1) and to the test workflow, to see how is going in ci

plugins/hls-alternate-number-format-plugin/test/Main.hs
```diff
@@ -0,0 +1,88 @@
...
+
+goldenAlternateFormat :: FilePath -> (TextDocumentIdentifier -> Session ()) -> TestTree
+goldenAlternateFormat fp = goldenWithHaskellDoc alternateNumberFormatPlugin (fp <> " (golden)") testDataDir fp "expected" "hs"
+
+codeActionTest :: (Maybe Text -> Bool) -> FilePath -> Int -> Int -> TestTree
+codeActionTest filter' fp line col = goldenAlternateFormat fp $ \doc -> do
+  actions <- getAllCodeActions doc
+  -- can't generate code actions?
+  liftIO $ assertBool "Actions is Not Null" (not $ null actions)
+--   case find (filter' . caTitle) actions of
```
jneira, line 38: Could we uncomment here the core of the test to see what has to say the ci testing workflow?

jneira:
> I ran the tests with logs enabled. Looks like an exception occurs (which I think is relevant?) and then a shutdown message is sent to the server.

That message about finishing the build is normal. I guess the test is not finding the code action, not sure why

Take a look to this test in the haddock plugin:

https://github.com/haskell/haskell-language-server/blob/1eb133f09838c473860ee516a9304cd8c1145e7b/plugins/hls-haddock-comments-plugin/test/Main.hs#L41

It is looking for a code action in a specific code range, maybe you need something similar to trigger it?

jneira:
Hi, it seems the plugin is not buildable with ghc-8.6.5, a version we still support. I hope only changing plugin base constraints could make the build work:

```
Resolving dependencies...
[__0] trying: ghcide-1.4.2.4 (user goal)
[__1] trying: base-4.12.0.0/installed-4.12.0.0 (dependency of ghcide)
[__2] next goal: hls-alternate-number-format-plugin (user goal)
[__2] rejecting: hls-alternate-number-format-plugin-0.1.0.0 (conflict:
base==4.12.0.0/installed-4.12.0.0, hls-alternate-number-format-plugin =>
base>=4.13 && <5)
[__2] fail (backjumping, conflict set: base,
hls-alternate-number-format-plugin)
goals I've had most trouble fulfilling: base, ghcide,
hls-alternate-number-format-plugin
```

https://github.com/haskell/haskell-language-server/runs/4209830202?check_suite_focus=true

drsooch:
> Hi, it seems the plugin is not buildable with ghc-8.6.5, a version we still support. I hope only changing plugin base constraints could make the build work:
> 
> ```
> Resolving dependencies...
> cabal: Could not resolve dependencies:
> [__0] trying: ghcide-1.4.2.4 (user goal)
> [__1] trying: base-4.12.0.0/installed-4.12.0.0 (dependency of ghcide)
> [__2] next goal: hls-alternate-number-format-plugin (user goal)
> [__2] rejecting: hls-alternate-number-format-plugin-0.1.0.0 (conflict:
> base==4.12.0.0/installed-4.12.0.0, hls-alternate-number-format-plugin =>
> base>=4.13 && <5)
> [__2] fail (backjumping, conflict set: base,
> hls-alternate-number-format-plugin)
> After searching the rest of the dependency tree exhaustively, these were the
> goals I've had most trouble fulfilling: base, ghcide,
> hls-alternate-number-format-plugin
> ```
> 
> https://github.com/haskell/haskell-language-server/runs/4209830202?check_suite_focus=true

Yeah, I started testing out 9.0.1 as well, and there are some changes needed However, those are pending @fendor's comments about using syb. I will aim to clean up the builds for each GHC version tonight.

jneira:
> It is looking for a code action in a specific code range, maybe you need something similar to trigger it?

Or maybe it is matter of waiting for the initial work of load the file, see:

https://github.com/haskell/haskell-language-server/blob/9415e55672a0c60a353f0bff3ae8546bdf0e7903/plugins/hls-splice-plugin/test/Main.hs#L88

drsooch:
Let me know if you want the Test Suite added in a different PR to reduce the clutter.

jneira:
Adding it here is fine with me

drsooch:
Also this is ready for PR review :) I've made decent amount of changes since the first iteration.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+data CollectLiterals = CollectLiterals
+                     deriving (Show, Eq, Generic)
+
+instance Hashable CollectLiterals
+instance NFData CollectLiterals
+
+type instance RuleResult CollectLiterals = CollectLiteralsResult
+
+data CollectLiteralsResult = CLR {
+    literals      :: [Literal]
```
fendor, line 55: Since we are at the nitpicking stage, I would suggest that the formatting here matches the one from: https://github.com/haskell/haskell-language-server/pull/2350/files#diff-9239cafe7a929f7c038ea0f8e32efc75af52e0f736e501259a3f7848c508e041R42

E.g.
```suggestion
data CollectLiteralsResult = CLR
    { literals      :: [Literal]
```

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
```diff
@@ -0,0 +1,146 @@
...
+import           Data.Text                     (Text)
+import qualified Data.Text                     as T
+import           Development.IDE.GHC.Compat
+import           Development.IDE.GHC.Util      (unsafePrintSDoc)
+import           Development.IDE.Graph.Classes (NFData (rnf))
+import qualified GHC.Generics                  as GHC
+import           Generics.SYB                  (Data, Typeable, cast,
+                                                everything)
+
+-- data type to capture what type of literal we are dealing with
```
fendor, line 18:
```suggestion
-- | data type to capture what type of literal we are dealing with
```

Make it haddock :)

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
```diff
@@ -0,0 +1,146 @@
...
+
+getLiteralAsList' :: HsLit GhcPs -> RealSrcSpan -> Set Literal
+getLiteralAsList' lit = maybe S.empty S.singleton . flip getLiteral lit
+
+-- Translate from Hs Type to our Literal type
+getLiteral :: RealSrcSpan -> HsLit GhcPs -> Maybe Literal
+getLiteral sSpan = \case
+  HsInt _ val   -> fromIntegralLit sSpan val
+-- Ignore this case for now
+--   HsInteger _ val _           -> Just $ IntLiteral sSpan Nothing val
```
fendor, line 91: What is this case, and why is it ignored?

plugins/hls-alternate-number-format-plugin/test/Main.hs
```diff
@@ -0,0 +1,178 @@
...
+    , codeActionProperties "TExpectNumDecimalFormat" [(4, 14)] $ \actions -> do
+        liftIO $ length actions @?= 5
+        liftIO $ actions `contains` numDecimalRegex @? "Contains numDecimal codeAction"
+    , conversions
+    ]
+
+codeActionProperties :: TestName -> [(Int, Int)] -> ([CodeAction] -> Session ()) -> TestTree
+codeActionProperties fp locs assertions = testCase fp $ do
+    runSessionWithServer alternateNumberFormatPlugin testDataDir $ do
+        openDoc (fp <> ".hs") "haskell" >>= codeActionsFromLocs >>= findAlternateNumberActions >>= assertions
```
fendor, line 76:
```suggestion
        openDoc (fp <.> ".hs") "haskell" >>= codeActionsFromLocs >>= findAlternateNumberActions >>= assertions
```

fendor (approved):
Pretty good! Great work!

I have a couple of comments, but I think the PR is basically ready!

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+    , formatTypes :: [FormatType]
+    } deriving (Generic)
+
+instance Show CollectLiteralsResult where
+    show _ = "<CollectLiteralResult>"
+
+instance NFData CollectLiteralsResult
+
+collectLiteralsRule :: Rules ()
+collectLiteralsRule = define $ \CollectLiterals nfp -> do
```
pepeiborra, line 65: What's the motivation for defining this rule over a plain function?

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+import           Development.IDE.Types.Logger         as Logger
+import           GHC.Generics                         (Generic)
+import           Ide.Plugin.Conversion                (FormatType,
+                                                       alternateFormat,
+                                                       toFormatTypes)
+import           Ide.Plugin.Literals                  (Literal (..),
+                                                       collectLiterals,
+                                                       getSrcSpan, getSrcText)
+import           Ide.Plugin.Retrie                    (handleMaybe,
+                                                       handleMaybeM, response)
```
pepeiborra, line 35: Acquiring a dependency on the retrie plugin just for these convenience functions seems like a bad idea. I would extract those functions to a shared place instead (or even duplicate them)

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+
+module Ide.Plugin.AlternateNumberFormat where
+
+import           Control.Lens                         ((^.))
+import           Control.Monad.Except                 (ExceptT, MonadIO, liftIO)
+import qualified Data.HashMap.Strict                  as HashMap
+import           Data.Maybe                           (fromMaybe, isJust)
+import           Data.Text                            (Text)
+import qualified Data.Text                            as T
+import           Debug.Trace                          (traceM)
```
pepeiborra, line 15: hopefully not used anywhere?

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+
+requestLiterals :: MonadIO m => IdeState -> NormalizedFilePath -> ExceptT String m CollectLiteralsResult
+requestLiterals state = handleMaybeM "Error: Could not get ParsedModule"
+                . liftIO
+                . runAction "AlternateNumberFormat.CollectLiterals" state
+                . use CollectLiterals
+
+
+logIO :: (MonadIO m, Show a) => IdeState -> a -> m ()
+logIO state = liftIO . Logger.logInfo (ideLogger state) . T.pack . show
```
pepeiborra, line 133:
```suggestion
logIO state = liftIO . Logger.logDebug (ideLogger state) . T.pack . show
```

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+        inCurrentRange lit = let srcSpan = getSrcSpan lit
+                              in currRange `contains` srcSpan
+
+        mkCodeAction :: NormalizedFilePath -> Literal -> Text -> Command |? CodeAction
+        mkCodeAction nfp lit alt = InR CodeAction {
+            _title = "Convert " <> getSrcText lit <> " into " <> alt
+            -- what should this actually be?
+            , _kind = Just $ CodeActionUnknown "alternate.style"
+            , _diagnostics = Nothing
+            , _isPreferred = Just True
```
pepeiborra, line 100: I would change `isPreferred` to `Nothing` or `Just False`

pepeiborra (changes requested):
Just a few more nitpicks

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
jneira: A good place for them (without look at its implementaions) could be https://github.com/haskell/haskell-language-server/blob/master/hls-plugin-api/src/Ide/PluginUtils.hs

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -0,0 +1,134 @@
...
+    where
+        inCurrentRange :: Literal -> Bool
+        inCurrentRange lit = let srcSpan = getSrcSpan lit
+                              in currRange `contains` srcSpan
+
+        mkCodeAction :: NormalizedFilePath -> Literal -> Text -> Command |? CodeAction
+        mkCodeAction nfp lit alt = InR CodeAction {
+            _title = "Convert " <> getSrcText lit <> " into " <> alt
+            -- what should this actually be?
+            , _kind = Just $ CodeActionUnknown "alternate.style"
```
michaelpj, line 98: Great question. I'd love to be a bit more systematic about our code action kinds...

For this one, I think it should arguably go under `quickfix`. We use that not only for things that are "wrong" but also for simple alternative versions of things, e.g. all the import rewriting code actions. So I think something like `quickfix.literals.style` or something would be plausible.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
michaelpj: Agree, I'm pretty sure this shouldn't be preferred. From the spec:

> A quick fix should be marked preferred if it properly addresses the underlying error. A refactoring should be marked preferred if it is the most reasonable choice of actions to take.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: To be quite honest, prior to my knowledge about HLS internals I thought I needed a Diagnostic for each Code Action. The only way I found was via creating the Rule. I'm happy to translate back to a plain function.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: Agreed. Would it be worth coming in with a follow up PR that replaces all usages? I pulled it from rename plugin. Looks like eval has it's own version in a util.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: Sigh, I had to turn off -Wall for a period and I forgot to reenable it.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
drsooch: I never found `HsInteger` in any of the test files I used. I didn't want to lose note of it just in case I ran into it.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: Updated.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
drsooch: Updated (changed the wording to be less clunky). That was mostly an internal comment.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
drsooch: Removed.

plugins/hls-alternate-number-format-plugin/test/Main.hs
drsooch: Updated.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: Cleaned up a bunch of imports...

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: So just to clarify it would be `Just $ CodeActionUnknown "quickfix.literals.style"` or would it be `Just CodeActionQuickFix`?

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
pepeiborra: I think the motivation is to avoid running the expensive AST traversal more often than strictly necessary. Rule results get cached, so the `CollectLiterals` result will be reused if the parsed module hasn't changed. On the other hand, without the rule, the AST traversal will be evaluated every time that the code action handler is invoked, right? So you probably do want the rule.

drsooch: Ahh! Sorry I misinterpreted your first comment. I can revert the changes I made. Would you prefer if I used any other `use` combinator? I currently just used `use` but I can see maybe `useWithStaleFast` being more appropriate?

pepeiborra: You should validate my conjecture as I am not quite sure of whether the client calls the code action handler more than once. 

Using stale data is optional, I think use is probably fine since a parse tree is almost always available.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
michaelpj: The former.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: Based on some rudimentary testing, it looks like the actual traversal portion is called much less when placed in the rule. So unless there are objections, it seems Rule-based is the better approach.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
drsooch: So just to reconfirm. I will be making a separate PR to pull out the `handleMaybe` code for `eval`, `rename`, and `retrie` into a common area. So as of now this code will go. (unless that's not how you guys prefer things)

Ailrun (approved):
Minor nitpicking here

pepeiborra (approved):

plugins/hls-alternate-number-format-plugin/hls-alternate-number-format-plugin.cabal
```diff
@@ -0,0 +1,68 @@
...
+description:
+  Please see the README on GitHub at <https://github.com/haskell/haskell-language-server#readme>
+
+license:            Apache-2.0
+license-file:       LICENSE
+author:             Nick Suchecki
+maintainer:         nicksuchecki@gmail.com
+category:           Development
+build-type:         Simple
+extra-source-files: LICENSE
```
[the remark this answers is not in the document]
drsooch: Nope, good catch.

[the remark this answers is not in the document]
jneira: sorry but testdata files should also be included or tests will not be able to work using the hackage distribution
take a look to other plugin .cabal file to see the glib syntax to do it, thanks!

jneira (changes requested):
test data files in .cabal are needed to merge this
also, could you add the plugin to the features docs page, with a demo if possible? that will make it more relevant and discoverable
thanks!

drsooch:
> test data files in .cabal are needed to merge this also, could you add the plugin to the features docs page, with a demo if possible? that will make it more relevant and discoverable thanks!

Should be all set now (had to make a small logic change to drop trailing zeroes from numbers with exponents.)

jneira:
> > test data files in .cabal are needed to merge this also, could you add the plugin to the features docs page, with a demo if possible? that will make it more relevant and discoverable thanks!
> 
> What about `hackage.yml`? Does that need updating?

oh, you are damn right, thanks for noting it

jneira:
There is a failing test plugin related for ubuntu and 8.10.7:

```
   Conversions
    Match NumDecimal:        FAIL
      *** Failed! Falsified (after 21 tests):
      10
      Use --quickcheck-replay=414296 to reproduce.
      Use -p '/Match NumDecimal/' to rerun this test only.
```

jneira:
> > > test data files in .cabal are needed to merge this also, could you add the plugin to the features docs page, with a demo if possible? that will make it more relevant and discoverable thanks!
> > 
> > 
> > What about `hackage.yml`? Does that need updating?
> 
> oh, you are damn right, thanks for noting it could you add it yourself, please?

done with eb05482

jneira (approved):
the plugin looks great, many thanks for working on this
only left fix the failing test

Issue #2270, opened 2021-10-10
Completions can insert redundant imports


pepeiborra:
In 1.4.0 we introduced completions for all local identifiers in the project. These completions automatically insert missing imports. It turns out that we don't verify whether the imports are actually missing, leading to some problems:

1. Redundant imports, if the import already exists
2. Conflicting import, if it's a redundant import that conflicts with something that already exists
3. Circular imports, if the completed identifier is defined in the module being edited

All the above are painful, but 2 and 3 are specially annoying. 

The solution for this would be:
- for 1, to perform an isNeeded check in the command that inserts the new import.
- for 3, to verify that the inserted module and the current module are different 
- 2 is a bug in the ordering of the completion suggestions. For example the user could legitimally want to use `ByteString` from `Data.ByteString.Lazy` even if `Data.ByteString` is already imported. The bug is in suggesting the non-imported completion before the imported one.

michaelpj:
It looks like this got partially fixed, but it seems like point 3 at least is still a problem.

PR #2354, opened 2021-11-14
More completions fixes

### LINKED ISSUE #2271: Completion suggestions contain prefixes of identifiers
In 1.4.0 we introduced completions for all the identifiers defined in the project. The database of possible identifiers is refreshed on every `kick`, as you type. This means new entries being added literally as you type, even if you hadn't finished typing yet :)

In the screenshot below, `GarbageCollectDirtyKeys` is a new constructor I have just defined a minute ago. I get completions for several prefixes of it. Ideally I only want to get only one completion for the full identifier:

image

How to prevent this? When refreshing the database of identifiers for a module M, wipe out all the previous entries of M in the database. This might require additional indexes.

pepeiborra:
* Clean up previous entries in the exports map when updating it (fixes #2271)
* Add typeText for all local completions to ensure they override global completions (fixes most instances of #2270)

We rely on local completions "shadowing" global ones and preventing redundant imports from being inserted (#2270). However, local completions require typechecking to compute, and because we use stale data to produce completions, there can be still some corner cases in which redundant import statements can be inserted if local completions stale data was not available. This should be rare

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2354"></a>


jneira (approved):
lgtm, thanks for continue improving completions

PR #2355, opened 2021-11-15
Qualify imported names plugin

eddiemundo:
qualify-imported-names-demo

## Purpose
If you have unqualified names imported by a unqualified import and you want to qualify them, this plugin can do it.

## Usage
1. Put cursor over the import declaration you want to qualify names from.
2. Initiate a Code Action.
3. Select `Qualify imported names`.

## Features
- Names are qualified on a per-import declaration basis.
- Names are qualified by the imported module's alias if it has one, otherwise by the imported module's name.
- If the import declaration has an explicit import list then the plugin will qualify only names on the list.
- If the import declaration has an explicit hiding list then the plugin will qualify names from the imported module that are not on the list.

I think this plugin is complete but I'm uncertain about GHC 9+.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2355"></a>


jneira:
First of all many thanks for working on this useful feature. Will do a review asap.
I wonder if all functionality about imports should live in one unique plugin, we will have 3 plugins about and seems to me that would be time to unify them.
It is a general consideration and it should be in another pr if we consider it makes sense.

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
```diff
@@ -0,0 +1,259 @@
...
+        _documentChanges = Nothing
+        _diagnostics = Nothing
+        _isPreferred = Nothing
+        _disabled = Nothing
+        _xdata = Nothing
+        _changeAnnotations = Nothing
+
+getTypeCheckedModule :: IdeState -> NormalizedFilePath -> IO (Maybe TcModuleResult)
+getTypeCheckedModule ideState normalizedFilePath =
+  runAction "QualifyImportedNames.TypeCheck" ideState (use TypeCheck normalizedFilePath)
```
pepeiborra, line 115: You may want to use `useWithStale` here, so that your plugin is able to work even when the file doesn't typecheck.

pepeiborra:
You forgot to add your test suite to CI in  .github/workflows/test.yml

```
      - if: matrix.test
        name: Test hls-qualify-imported-names-plugin test suite
        run: cabal test hls-qualify-imported-names-plugin --test-options="-j1 --rerun-update" || cabal test hls-qualify-imported-names-plugin --test-options="-j1 --rerun" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-qualify-imported-names-plugin --test-options="-j1 --rerun"

```

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
```diff
@@ -0,0 +1,259 @@
...
+                   , let lineStartingAtIdentifier = Text.drop (startCol - colOffset) line
+                   , Just (c, _) <- Text.uncons lineStartingAtIdentifier
+                   , let isParenthesized = c == '('
+                   , let isBackticked = c == '`'
+                   , let replacementText =
+                           if | isParenthesized -> "(" <> qualifiedIdentifierText <> ")"
+                              | isBackticked -> "`" <> qualifiedIdentifierText <> "`"
+                              | otherwise -> qualifiedIdentifierText ->
+                       (replacementText, lineStartingAtIdentifier : remainingLines)
+                   | otherwise -> (qualifiedIdentifierText, lines)
```
pepeiborra, line 225: Do you check whether the identifier is already qualified?

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
```diff
@@ -0,0 +1,259 @@
...
+    folder :: [TextEdit] -> UsedIdentifier -> State ([Text], Int, Int) [TextEdit]
+    folder prevTextEdits (UsedIdentifier identifierName identifierSpan)
+      | Just importedBys <- lookupNameEnv nameToImportedByMap identifierName
+      , Just (ImportedBy alias _) <- find (isRangeWithinImportedBy range) importedBys
+      , let IdentifierSpan row startCol endCol = identifierSpan
+      , let identifierRange = identifierSpanToRange identifierSpan
+      , let aliasText = Text.pack $ moduleNameString alias
+      , let identifierText = Text.pack $ occNameString $ nameOccName identifierName
+      , let qualifiedIdentifierText = aliasText <> "." <> identifierText = do
+          (sourceTextLines, lineOffset, updateColOffset row lineOffset -> colOffset) <- State.get
```
pepeiborra, line 212: I suppose that this is to account for multiple identifiers in the same line. 
Is there a test for that?

pepeiborra (changes requested):
Amazing work! The only thing missing is adding the testsuite to CI

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
eddiemundo: The `qualifiedIdentifierText` comes from a `Name` which contains only a `OccName` that is always unqualified (to my knowledge), so qualifying it should always be safe.

I've added a test to show qualified names are not double qualified.

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
eddiemundo: Yes. I didn't know a way to see whether a name is parenthesized, or backticked in the parsing pass, and saw an example in the "suggest import disambiguation" code action where they simply checked the span in the source. So I did the same thing except more involved because I wanted one pass through the file for all names.

I added a test to show parenthesized and backticked names on the same line get qualified. For just regular names on the same line there are other tests that show this.

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
eddiemundo: I'm not sure that's good because if someone is typing names in a file that doesn't typecheck and expects those names to be qualified by the plugin then they might be confused as to why older used names are qualified, but the ones they've been typing aren't. I feel like the plugin is used like a tidy up or refactoring tool after the program generally works, and so it's okay to have the file typecheck before using the plugin.

jneira (approved):
lgtm the plugin look really great
could you update the [features documentation](https://github.com/haskell/haskell-language-server/blob/master/docs/features.md) adding it as well, thanks!

pepeiborra (approved):

jneira:
Will merge after releasing 1.5.1, thanks for your patience :slightly_smiling_face:

jneira (changes requested):
sorry, I ve just observed the .cabal file does not include the readme and test data files in source-extra-files, take a look to other plugins .cabal to see the glob syntax available
thanks

eddiemundo:
> sorry, I ve just observed the .cabal file does not include the readme and test data files in source-extra-files, take a look to other plugins .cabal to see the glob syntax available thanks

I looked at some of the other plugins and they don't seem to include the README.md or gif, but I included them anyway.

jneira:
> I looked at some of the other plugins and they don't seem to include the README.md or gif, but I included them anyway.

yeah you are right, we have to eventually add it in the rest of plugins, thanks for doing it for this one

jneira (approved):
LGTM, thanks a lot for this nice plugin

PR #2359, opened 2021-11-16
Avoid unnecessary Target canonicalisation in Session setup

fendor:
Closes #2358  

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2359"></a>

PR #2361, opened 2021-11-16
Prepare 1.5.0

jneira:
This one includes #2346, will rebase when it is merged in master 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2361"></a>


stack-9.0.1.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: nightly-2021-09-16
+resolver: nightly-2021-11-06
```
Ailrun, line 1: I think this `package` should also be fixed.

Ailrun (approved):
Awesome, I'm eagerly waiting for the full support of GHC 9.0!

jneira:
> 
> 
> Awesome, I'm eagerly waiting for the full support of GHC 9.0!

I am afraid although i uncommented packages we still miss brittany, floskell, tactics and class plugins

stack-9.0.1.yaml
jneira: I am afraid we cant uncomment packages here cause stack tries to build them even if hls has those plugins disabled.
cabal-install takes them in account for the build plan (so i had to add some allo-newer) but only builds them if they are enabled in hls

jneira:
Run checking the hackage release here: https://github.com/haskell/haskell-language-server/actions/runs/1471669261
As you can observe, it has 9.0.1 disabled :worried:

fendor:
Can we get #2359 also merged?

jneira:
> 
> 
> Can we get #2359 also merged and into this release, or rather not?

I would prefer dogfooding it until the next release, if you are not in a hurry to have the fix

fendor:
Ok, good point!

PR #2371, opened 2021-11-19
getDocsBatch for bulk processing

Anton-Latukha:
A branch from: #2349

Holds a relevant `getDocsBatch` code to make bulk requests.

TODO:
  - [ ] `error` -> exception handling code
  - [ ] `Map`s be Strict
  - [ ] Recompose (functor law & `first`s can be flattened)
  - [ ] Wire it in

---

Currently in `master` primary use of `getDocsBatch` is: `getDocumentationTryGhc` that runs singleton element doc on: `getDocumentationsTryGhc` that runs `getDocsBatch` on singleton.

`getDocsBatch` internally has a modified (with removed interactive check) copy of GHCs `getDocs`. If that internal 1-elem non-interactive function is needed - it probably needs to be `getDocsNonInteractive` & so be module scope.

And then `getDocsBatch` is designed for what it is intended - a bulk `[Name]` requests.

Then I went further: if `getDocsBatch` would be used really in batch mode - why is to do `getDocsBatch` on `getDocsNonInteractive` - that searches & loads module interface on each element:

So, currently I do in `getDocsBatch`:

Load `[Name]`.

For each `Name` get module it says being part of.

Then group all `Names` by according `Module`: `[(Module, [Name])]`.

Then load module interface once & process through it in bulk [Name]s that are in that module.

For current way function is used (singleton retrieval) that would introduce `+ C` steps for retrieval  (that may be specialized-away by the compiler).

If HLS starts bulking retrieval from GHC interfaces based on [Name] - that would be optimal way.


jneira:
I have only a suprficial knowledge about the ghcide part of the codebase (well and about the other ones 😆), but your changes looks very interesting. I think maybe @pepeiborra would be interested in taking a look, even only to design, as it is still a WIP

pepeiborra:
This is unreasonably hard to review as a first contribution. As general advice, please try to keep PRs small, change only one thing, make it easy for reviewers to understand. 

For this particular PR:
1. Extract all cosmetic/style changes, like replacing `concatMap` for `traverse`, into one or more separate PRs
2. Extract all controversial changes, like removing `List`, into individual PRs for discussion
3. This PR should contain only the minimal changes to do bulk docs processing

Make sure to review the benchmark results of each PR individually to prevent performance regressions

Anton-Latukha:
Yes. It is WIP decomposed into a lot of atomic functions, so it is easy to modify stages.

The main thing there - is the question is bulk interface retrieval is needed at all.

The main souse there is & starts in `groupSort`, which I love to use for such stuff.

Anton-Latukha:
Woops - messed-up. Made a branch, stored the code & then not switched the branch, so overwritten the bulking code stored here with the other code.

The batching I talk about is drafted in 1d658151d4e1f55a149a43aedd6669d99a2bd292.

But it is also not a PR for review. I wanted to ask "Is it needed" first.

Anton-Latukha:
@pepeiborra

Sorry, I was not expecting review here, and during the review process the branch was borked - so the draft opened about one thing & code was about the other.

Anton-Latukha:
@jneira @pepeiborra

This `DList` talk produced a: https://github.com/haskell-perf/sequences/pull/23, now https://github.com/haskell-perf/sequences in the comparison shows `DList` & where it applies. To additionally check that it is proper results in https://github.com/spl/dlist/issues/102 I notified the `DList` author, `spl` & he responded - everything seems proper.

If to make `DList` strict - it becomes terrible (it becomes `[]` with additional complexity). `DList` main power is Haskell laziness.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -987,33 +987,88 @@ mkDetailsFromIface session iface linkable = do
...
-               then pure (Left (NoDocsInIface mod $ compiled name))
-               else pure (Right ( Map.lookup name dmap
-                                , Map.findWithDefault Map.empty name amap))
-    case res of
-        Just x  -> return $ map (first $ T.unpack . showGhc) x
-        Nothing -> throwErrors errs
-  where
-    throwErrors = liftIO . throwIO . mkSrcErr
-    compiled n =
+    ((_warns,errs), res) <- initTc hsc_env HsSrcFile False _mod fakeSpan $ Map.fromList <$> traverse undefined undefined
```
Anton-Latukha, line 1003: From https://gitlab.haskell.org/ghc/ghc/-/issues/20831#note_398227:

> 
> I looked at the HLS code and you should probably rather use:
> `initIfaceLoad hsc_env (loadSysInterface "docs" module)`
> to load the interface rather than calling `initTc`.
> Then that will cache the loaded interface so you don't load it multiple times from disk.

Anton-Latukha: (code currently is created so that it loads & opens interfaces once, but mentioned code may cache interfaces between function runs).

PR #2373, opened 2021-11-20
hls-explicit-imports-plugin: Add golden files to hackage tarball

maralorn:


jneira (approved):
ugh good catch, wondering why the hackage workflow didn't fail, will take a look

jneira:
> ugh good catch, wondering why the hackage workflow didn't fail, will take a look

simply the reason is the workflow does not run tests so we dont catch packaging issues 🤦

PR #2374, opened 2021-11-20
hls-module-name-plugin: Add missing golden file to hackage tarball

maralorn:


plugins/hls-module-name-plugin/hls-module-name-plugin.cabal
```diff
@@ -16,6 +16,7 @@ extra-source-files:
   LICENSE
   test/testdata/*.hs
   test/testdata/*.yaml
+  test/testdata/subdir/*.hs
```
jneira, line 19: what about leverage globs available in cabal spec 2.4?
```suggestion
  test/testdata/**/*.yaml
  test/testdata/**/*.hs
```

jneira (approved):
lgtm, thanks again

Issue #2365, opened 2021-11-18
HLS 1.5 appears to not handle hs-boot files correctly


sheaf:
With HLS 1.5 (using GHC 9.0.1), using the VS Code extension on Windows 10, I'm getting errors when running HLS on the GHC codebase. These errors seem to be caused by a problem loading an `hi-boot` interface file.

Running HLS on [GHC commit f8c1c549bbb02cb0a943a9ae49dee922c96d85c4](https://gitlab.haskell.org/ghc/ghc/-/commit/f8c1c549bbb02cb0a943a9ae49dee922c96d85c4), I'm getting the following errors:

<details>
<summary>Show/hide interface file errors</summary>

```
GHC.Types.ForeignCall:

• No instance for (Eq (GenUnit UnitId))
    arising from the third field of ‘StaticTarget’ (type ‘Maybe Unit’)
  Possible fix:
    use a standalone 'deriving instance' declaration,
      so you can specify the instance context yourself
• When deriving the instance for (Eq CCallTarget)

• No instance for (Data (GenUnit UnitId))
    arising from the third field of ‘StaticTarget’ (type ‘Maybe Unit’)
  Possible fix:
    use a standalone 'deriving instance' declaration,
      so you can specify the instance context yourself
• When deriving the instance for (Data CCallTarget)

[...]

GHC.Types.Name:

• Can't find interface-file declaration for variable isInteractiveModule
    Probable cause: bug in .hi-boot file, or inconsistent .hi file
    Use -ddump-if-trace to get an idea of which file caused the error
• In the second argument of ‘(||)’, namely
    ‘isInteractiveModule mod’
  In the expression: from == mod || isInteractiveModule mod
  In an equation for ‘nameIsLocalOrFrom’:
      nameIsLocalOrFrom from name
        | Just mod <- nameModule_maybe name
        = from == mod || isInteractiveModule mod
        | otherwise = True

[...]

GHC.Unit.Info:

• Can't find interface-file declaration for type constructor or class UnitKey
    Probable cause: bug in .hi-boot file, or inconsistent .hi file
    Use -ddump-if-trace to get an idea of which file caused the error
• In the first argument of ‘GenUnitInfo’, namely ‘UnitKey’
  In the type ‘GenUnitInfo UnitKey’
  In the type declaration for ‘UnitKeyInfo’
```
</details>

These all seem to do with failing to load the `hi-boot` file for `GHC.Unit.Types`.

[...]

PR #2377, opened 2021-11-20
Boot files

### LINKED ISSUE #2362: hls does not work in the ghcup-hs codebase
* As reported by @hasufell here: https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/286hls does not work in `lib/GHCup.hs`(f.e.) from our beloved cousin project
* I tried in windows and i got the ide features in some utility modules, i have to check it which one
* We should recollect the usual logs as described https://haskell-language-server.readthedocs.io/en/latest/troubleshooting.html#troubleshooting-the-server

pepeiborra:
Fixes the regression in #2365 addressing 2 problems identified by @mpickering and @wz1000 
Tested with the repros described in that ticket (using the ghc codebase).

Does not_fix #2362 but that appears to be a slightly different problem (and verified it's not a regression).

I don't know if I'll have time to add a new test before next weekend

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2377"></a>


wz1000 (commented):
Seems fine to me, I will try to come up with a test case.

guibou:
I tested that this also closes #2385.

jneira:
maybe a regression test based on #2385 would be great, sad it is not reproducible with the simple case described in the issue

guibou:
@jneira I discovered at the end of writting the issue that boot file may be involved. So my repro without boot file was not succesful.

I was at work, so a bit busy by "we need to fix HLS right now", but I'll try to write a repro when I'll get a bit of time.

pepeiborra:
I basically gave up on constructing a test case for this, so any ideas will be appreciated @guibou @wz1000 
In the meantime we should land this and start thinking about an HLS 1.5.1 release @jneira

mpickering:
Perhaps the boot5688 test in the GHC tree triggers this case?

pepeiborra:
> Perhaps the boot5688 test in the GHC tree triggers this case?

That worked, it fails without this patch and succeeds with it. Test added now, thanks Matt.

guibou:
Damned, I did not saw @mpickering answer before starting to work on my repro.

Here is my repro: https://github.com/guibou/hls_repro_2385, in case it may help.

In any case, thank you for providing a solution for this problem.

jneira (approved):
Perfect, regression test included, thanks @pepeiborra for the fix and everyone for the help

Will cut a 5.1 release asap after merge this

jneira:
> Damned, I did not saw @mpickering answer before starting to work on my repro.
> 
> Here is my repro: https://github.com/guibou/hls_repro_2385, in case it may help.
> 
> In any case, thank you for providing a solution for this problem.

Thanks for working in the repro anyways 👍

jneira:
there are almost as flaky tests in ubuntu than windows lately :worried:

PR #2380, opened 2021-11-21
Fix regression in GhcSessionDeps

### LINKED ISSUE #2379: HLS 1.5 does not handle TH correctly
This is another regression from #2323 and can generate errors like below in TH codebases:

```
  tried to look at linkable for GetModIfaceWithoutLinkable for NormalizedFilePath
  "/home/adam/src/ghc/ghc-8.10/hls/th-pq-segfault/pqexample/src/PQ.hs"
  error, called at src/Development/IDE/Core/Rules.hs:888:79 in
  ghcide-1.5.0-inplace:Development.IDE.Core.Rules
```

pepeiborra:
We cannot use GetModIfaceWithoutLinkable since the session might be reused later to load a module that needs linkables

Note that this does not have any effects on performance, since GetModIfaceWithoutLinkable is just a synonym for GetModIface that removes the linkable

Fixes #2379

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2380"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/RuleTypes.hs
+++ b/ghcide/src/Development/IDE/Core/RuleTypes.hs
@@ -252,10 +252,6 @@ type instance RuleResult GetModIfaceFromDiskAndIndex = HiFileResult
 -- | Get a module interface details, either from an interface file or a typechecked module
 type instance RuleResult GetModIface = HiFileResult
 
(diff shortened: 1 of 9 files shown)


jneira (approved):
many thanks for the quick fix, including regression tests

PR #2386, opened 2021-11-22
docs: Fix typo

nh2:


jneira (approved):
Thanks for finding and fixing it

[skip circleci]

PR #2402, opened 2021-11-25
Add fix for import placement

nini-faroux:
This is related to #2401 and #2392 but with imports in ghcide. 

So it's only an issue when the file doesn't have any existing imports or a module declaration already - in those cases it relies on the ParsedSource type from the GHC parsing.

But in the case where neither of those are in the file it would fall back on the same logic as in the pragma plugin, and so had the same bug when there was multiline LANGUAGE or OPTS_GHC pragmas, like:

Before:
{-# LANGUAGE RecordWildCards,
   OverloadedStrings #-}

After:
{-# LANGUAGE RecordWildCards,
    Import Data.Monoid
    OverloadedStrings #-}

This fixes that issue. It works for all the cases except for the issue raised in #2392, if there's an OPTIONS_GHC pragma later in the file the import will be incorrectly added under that.

I think it's probably better to wait for the parser to deal with that case instead of trying to do the same thing in two places, and I can adapt this solution to use that then. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2402"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -21,10 +21,12 @@ module Development.IDE.Plugin.CodeAction
 import           Control.Applicative                               ((<|>))
 import           Control.Arrow                                     (second,
                                                                     (>>>))
-import           Control.Monad                                     (guard, join)
+import           Control.Monad                                     (foldM,
+                                                                    guard, join)
 import           Control.Monad.IO.Class
 import           Data.Char
 import qualified Data.DList                                        as DL
+import           Data.Either.Extra                                 (fromEither)
 import           Data.Function
 import           Data.Functor
 import qualified Data.HashMap.Strict                               as Map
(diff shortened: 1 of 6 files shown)


jneira:
thanks for the pr, I am gonna mark it as WIP,  to wait to the pr using the new parser
alternatively regression tests could be extracted, marked as known broken if needed and be merged before any try to fix it

nini-faroux:
thanks that makes sense, I'll update it when the parser is ready and add the new tests then too

jneira:
#2392 is already merged

nini-faroux:
hi @jneira, I'm not too sure about the best way to integrate the parser in ghcide to be honest, I don't think it makes sense to copy the whole thing into ghcide and have it in two places, or to include a different solution to it in ghcide, so I guess it might be a case of putting it in it's own module and importing it into the pragmas plugin and ghcide - but I think it would be better to leave that to you to decide on. So I might just extract the tests like you mentioned and add 'known broken'. 

The place it's needed is where the 'findNextPragmaPosition' function is now in 'ghcide/src/Development/IDE/Plugin/CodeAction.hs' - which arises when there's no module declaration or imports in the file.

I also noticed #2402 - I can include a fix for that if you want - it's just a small change and unrelated to the case that needs the parser.

nini-faroux:
I'll close this and add the known broken tests in a separate pr now.

PR #2393, opened 2021-11-23
Prepare 1.5.1

jneira:
* The hackage check is [here](https://github.com/jneira/haskell-language-server/runs/4305202883?check_suite_focus=true). It is failing for ghc-8.6.5 with

```
cabal: Could not resolve dependencies:
[__0] trying: ghcide-1.5.0.1 (user goal)
[__1] rejecting: ghcide:!test (constraint from config file, command line flag,
or user target requires opposite flag selection)
[__1] trying: ghcide:*test
[__2] trying: lsp-test-0.14.0.1 (dependency of ghcide *test)
[__3] trying: conduit-parse-0.2.1.0 (dependency of lsp-test)
[__4] trying: parsers-0.12.10 (dependency of conduit-parse)
[__5] trying: semigroups-0.20 (dependency of parsers)
[__6] trying: text-1.2.3.1/installed-1.2.3.1 (dependency of ghcide)
[__7] next goal: sqlite-simple (dependency of ghcide)
[__7] rejecting: sqlite-simple-0.4.18.0 (conflict: semigroups==0.20,
sqlite-simple => semigroups>=0.18 && <0.20)
[__7] skipping: sqlite-simple-0.4.16.0, sqlite-simple-0.4.15.0 (has the same
characteristics that caused the previous version to fail: excludes
'semigroups' version 0.20)
[__7] rejecting: sqlite-simple-0.4.14.0, sqlite-simple-0.4.13.0,
sqlite-simple-0.4.12.1, sqlite-simple-0.4.12.0, sqlite-simple-0.4.11.0,
sqlite-simple-0.4.10.0, sqlite-simple-0.4.9.0, sqlite-simple-0.4.8.0,
sqlite-simple-0.4.7.0, sqlite-simple-0.4.6.1, sqlite-simple-0.4.6.0,
sqlite-simple-0.4.5.2 (conflict: text => base==4.12.0.0/installed-4.12.0.0,
sqlite-simple => base>=4.5 && <4.11)
[__7] rejecting: sqlite-simple-0.4.5.1, sqlite-simple-0.4.5.0 (conflict: text
=> base==4.12.0.0/installed-4.12.0.0, sqlite-simple => base>=4.6 && <4.11)
[__7] rejecting: sqlite-simple-0.4.4.0, sqlite-simple-0.4.3.1,
sqlite-simple-0.4.3.0, sqlite-simple-0.4.2.0, sqlite-simple-0.4.1.0,
[...]

PR #2408, opened 2021-11-27
[skip circleci] Rerun tests between workflow run attempts

jneira:
* To try to alliviate flakiness. Even if we get to reduce the number of flaky tests i think we could keep this as tests always can fail due to external factors
* The trick is cache `.tasty-rerun-log` per commit and workflow matrix. So f.e. linux+ghc-8.6.5+commit is succesful and windows+ghc-9.0.1 fails the previous one will be succesfull without actually running any test (which might fail this time)
* Test options are `--rerun-update --rerun-filter failures,exceptions,new`, if there is no `.tasty-rerun-log`, the situation in the first attempt of the run, it will execute all tests. All succesives test executions will filter out the succesful ones, even between different run attempts

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2408"></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,7 +144,7 @@ jobs:
 
       - name: Set test options
         run: |
-          echo "TEST_OPTS=\"-j1 --rerun-update --rerun-filter failures,exceptions,new\"" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


jneira:
Included in #2393 to help get it merged. This changes are only ci config ones, so it cant introduce itself any bug

PR #2410, opened 2021-11-28
Enable tactic tests for linux+8.6.5

jneira:
* Disabled in #2393

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2410"></a>


jneira:
* The first run when tactics failed was: https://github.com/haskell/haskell-language-server/runs/4308191268?check_suite_focus=true (commit 68eb37c511e20cb9045fdcf597ce2555dd45c9df)
* That belongs the first workflow of the branch https://github.com/haskell/haskell-language-server/actions/runs/1496923126
* As usual #2393 did not any code changes, only build config files, so the unique reason i can think of it could affect tactics test would be via dependencies changes

To investigate

jneira:
to ensure changes in test options was not related (I think i saw the error before that change but I was not sure 100%) I ve restore the original ones, and the suite continue failing

michaelpj:
The tests are already enabled on master, it seems. The diff no longer turns them on!

PR #2357, opened 2021-11-16
Decrease contention in Progress reporting

pepeiborra:
Out of curiosity I put together a lock-less version of hls-graph using the sim-containers package as suggested by @wz1000 and others in IRC a few months ago. 

So far I haven't seen any perf. improvement, in fact the amount of work done (allocations) is much higher. Perhaps with more tuning it could be competitive? Leaving it here in case anyone wants to have a play

EDIT: The original PR was broken in chunks, and now that there is only one major change remaining, it's worth reviewing and merging this PR to preserve the discussion instead of extracting it into yet another PR.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2357"></a>

PR #2411, opened 2021-11-28
Lock-less hls-graph 

pepeiborra:
First PR extracted from #2357. While I don't have performance measurements for this change alone, the final performance results are in https://github.com/haskell/haskell-language-server/pull/2357#issuecomment-981146589

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2411"></a>


jneira (approved):
lgtm by pure trust 😉
hope the performance will not be worse in machines with few cores or with j2

pepeiborra:
> lgtm by pure trust 😉, thanks!
> 
> hope the performance will not be worse in machines with few cores or with j2

Check the comment on the mother PR, performance does regress with j2 but is competitive with j4. This is expected and I think it's a reasonable trade off.

jneira:
> > lgtm by pure trust wink, thanks!
> > hope the performance will not be worse in machines with few cores or with j2
> 
> Check the comment on the mother PR, performance does regress with j2 but is competitive with j4. This is expected and I think it's a reasonable trade off.

thanks for the reference,it seems the loss is reasonable, will complain if it is not as i have to live with 2 cores in my main dev laptop :stuck_out_tongue_closed_eyes:

pepeiborra:
Let's not merge this just yet

PR #2425, opened 2021-12-01
Add known broken tests for import placement

nini-faroux:
Tests for known issues with import placement.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2425"></a>


ghcide/test/exe/Main.hs
```diff
@@ -855,7 +855,18 @@ watchedFilesTests = testGroup "watched files"
 
 insertImportTests :: TestTree
 insertImportTests = testGroup "insert import"
-  [ checkImport "above comment at top of module" "CommentAtTop.hs" "CommentAtTop.expected.hs" "import Data.Monoid"
+  [ (`xfail` "known broken") (checkImport "module where keyword lower in file no exports" "WhereKeywordLowerInFileNoExports.hs" "WhereKeywordLowerInFileNoExports.expected.hs" "import Data.Int")
```
jneira, line 858: please could you use the `expectFailBecause` existing function, if possible with a short insight about why do you think it fail ("known broken" would be fine though)

jneira:
Many thanks for adding tests, they will help a lot in an eventual resolution

ghcide/test/exe/Main.hs
nini-faroux: Sure, I'll do that later today.

ghcide/test/data/import-placement/PragmaNotAtTopMultipleComments.hs
```diff
@@ -0,0 +1,22 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+{-# OPTIONS_GHC -Wall,
+   OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
eddiemundo, line 4: The expected version of this test doesn't have the `OPTIONS_GHC -Wno-unrecognized-pragmas` line. I think the expected version of the test should have this line unless the test does more than add an import.

eddiemundo:
I see what @nini-faroux was talking about on where to extract the "pre-decl" parser functionality since it's needed in both ghcide CodeAction.hs, and the pragmas plugin. I also need it in the hlint plugin.

Of the existing places I think putting it in its own module under ghcide/src/Development/IDE/Spans would be ok since Spans is vaguely related, but stuffing it into CodeAction.hs or in its own module in that folder seems ok too since it has random stuff in there.

I think I might extract that functionality into one of those places later today.

ghcide/test/data/import-placement/PragmaNotAtTopMultipleComments.hs
nini-faroux: no you're right, that's updated now.

ghcide/test/exe/Main.hs
```diff
@@ -855,7 +855,29 @@ watchedFilesTests = testGroup "watched files"
 
 insertImportTests :: TestTree
 insertImportTests = testGroup "insert import"
-  [ checkImport "above comment at top of module" "CommentAtTop.hs" "CommentAtTop.expected.hs" "import Data.Monoid"
+  [ expectFailBecause "'findPositionFromImportsOrModuleDecl' function adds import directly under line with module declaration, not accounting for case when 'where' keyword is placed on lower line"
+ (checkImport "module where keyword lower in file no exports" "WhereKeywordLowerInFileNoExports.hs" "WhereKeywordLowerInFileNoExports.expected.hs" "import Data.Int")
```
jneira, line 859: minor formatting thing, maybe the tests could be more legible if  they are splitted in several lines, something like
```
  [ expectFailBecause 
      $  "'findPositionFromImportsOrModuleDecl' function adds import directly under line with module declaration, " 
      ++ "not accounting for case when 'where' keyword is placed on lower line"
    (checkImport 
      "module where keyword lower in file no exports" 
      "WhereKeywordLowerInFileNoExports.hs" 
      "WhereKeywordLowerInFileNoExports.expected.hs" 
      "import Data.Int")
```

nini-faroux: that's true, they're updated now

jneira (approved):
many thanks again, I hope tests would help in an eventual resolution

PR #2419, opened 2021-11-30
Centralized caching workflow

Anton-Latukha:
Discussed in: https://matrix.to/#/!oOjZFsoNYPAbTEgSOA:libera.chat/$ckv7h8SSMbimsa_5pCaCFn2uMm_nEVIy6gy0QyFgc8M?via=libera.chat&via=matrix.org&via=monoid.al

Short: Cabal builds deps from the source. Currently, cache gets saved only on the successful passing of the CI workflow pipeline (includes dependencies built, then all HLS projects built & then successful passing of all tests, core word is "successful" passing of all of that), while successful Cabal dependency build stage depends only on does Hackage snapshot state (success of Cabal dependency build stage).

Basic idea - is to run `cabal v2-build --only-dependencies`, which would ensure saving & sharing of the snapshot of successfully built dependencies into pull requests. To run on schedule & on `push` to `master`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2419"></a>

PR #2426, opened 2021-12-01
CI: cache-deps: rm pull request hook

Anton-Latukha:
I needed to remove https://github.com/haskell/haskell-language-server/pull/2419/commits/b0621d74904ee8bfb27e1452bcb64b0e8c620736 before merge.

That is my fault, I should be more conservative with signaling draft<->mergable PR.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2426"></a>


Anton-Latukha:
To be clear: this `master` cache building workflow is made to run on `master` only, hook was to test&show inside the PR that & how it works.

Additional thought: Currently `test.yml` already does almost the exact thing unstably inside the PRs & that is the next thing I wanted to look into, first - how to keep `master` caches intact in the cache pool. The easiest effective solution there is to prevent PRs from saving cache, so they would be required to follow `master` to have the cache. But I also wanted to look is it possible to at the same time give PRs some ability to additionally cache their build artifacts during active work.

jneira (approved):
dont worry, i saw the comment when the pr was a draft and i did not a proper review once it was marked as ready to review
no final user was harmed in the process so 🤷

PR #2427, opened 2021-12-01
Fixing the cache situation

Anton-Latukha:
Because the number of small things are addressed here - it is easier to make documentation in commits & address them in review mode.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2427"></a>

### DIFF
--- a/.github/workflows/cache-deps.yml
+++ b/.github/workflows/cache-deps.yml
@@ -102,11 +102,11 @@ jobs:
           cache-name: compiled-deps
         with:
           path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
-          key:          ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.HACKAGE_TIMESTAMP }}-${{ hashFiles('cabal.project') }}
+          key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.HACKAGE_TIMESTAMP }}-${{ hashFiles('cabal.project') }}
           restore-keys: |
-                        ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.HACKAGE_TIMESTAMP }}-
-                        ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
-                        ${{ env.cache-name }}-${{ runner.os }}-
+                ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.HACKAGE_TIMESTAMP }}-
+                ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
+                ${{ env.cache-name }}-${{ runner.os }}-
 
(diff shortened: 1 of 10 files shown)


.github/workflows/cache-deps.yml
```diff
@@ -76,13 +76,20 @@ jobs:
...
-          cache-name: cache-cabal
+          cache-name: hackage-sources
         with:
-          path: |
-            ${{ steps.HaskEnvSetup.outputs.cabal-store }}
-            ${{ env.CABAL_PKGS_DIR }}
-          key: v2-${{ runner.os }}-${{ matrix.ghc }}-build-${{ hashFiles('cabal.project') }}
-          restore-keys: v2-${{ runner.os }}-${{ matrix.ghc }}-build-
+          path: ${{ env.CABAL_PKGS_DIR }}
+          key:          ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-${{ hashFiles('cabal.project') }}
```
jneira: what is the env.cache_name?

Anton-Latukha:
Current source caching works great:
```yaml
      - name: Hackage sources cache
        uses: actions/cache@v2
        env:
          cache-name: hackage-sources
        with:
          path: ${{ env.CABAL_PKGS_DIR }}
          #  2021-12-01: NOTE: Cabal Hackage source tree storage does not depend on OS or GHC.
          key:          ${{ env.cache-name }}-${{ env.HACKAGE_TIMESTAMP }}
          restore-keys: ${{ env.cache-name }}-
```

All GHCs use one `193Mb` cache instance & none of them needed to download additional versions.

But depending on the library versions supporting particular `base` - some package downloads can happen for particular GHCs here & there down the line in the future.

The most ideal way to cache is to differentiate on GHC versions:
```yaml
      - name: Hackage sources cache
        uses: actions/cache@v2
        env:
          cache-name: hackage-sources
        with:
          path: ${{ env.CABAL_PKGS_DIR }}
          #  2021-12-01: NOTE: Cabal Hackage source tree storage does not depend on OS or GHC really, but can depend on `base`.
          key:          ${{ env.cache-name }}-${{ matrix.ghc }}-${{ env.HACKAGE_TIMESTAMP }}
          restore-keys: |
            ${{ env.cache-name }}-${{ matrix.ghc }}-
            ${{ env.cache-name }}-
```

But that means storing (6 GHCs) - 6 times more sources.

Source caching is really to be done inside `master` branch for caching & that gets shared into PRs.
But PRs really would not change or even use dependency sources, since they would have & use (should) dependencies compiled.

And if the person updates the Hackage pin - it anyway needs full redownload & rebuild, of caches also.

So source caching is really only happens in `master` & only for `master` having a faster compilation of deps. In that way it is revealed that occasional partial redownload from Hackage is a redundant theme to drive attention to & discuss & that source caching at all can be removed without notacible implications.

I would keep it as it is & would put a statement to "think hard - that is a `master` for `master` purposes".

Anton-Latukha:
Closing in favour of #2446  continuation.

PR #2429, opened 2021-12-02
lock-less Values state

pepeiborra:
Third extract from #2357. This change removes the locking around the build cache by replacing it with an STM Map

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2429"></a>


jneira (approved):
lgtm

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -420,14 +423,14 @@ shakeDatabaseProfileIO mbProfileDir = do
...
-          => Var Values
+          => Values
           -> k
           -> NormalizedFilePath
           -> Value v
           -> Vector FileDiagnostic
           -> IO ()
 setValues state key file val diags =
-    void $ modifyVar' state $ HMap.insert (toKey key file) (ValueWithDiagnostics (fmap toDyn val) diags)
+    atomically $ STM.insert (ValueWithDiagnostics (fmap toDyn val) diags) (toKey key file) state
```
michaelpj, line 433: I wonder whether for things that are in `IO` and end with a top-level `atomically` it would make sense to change the type to stay in STM so the caller has the option of batching it if they want to?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -438,7 +441,7 @@ deleteValue
   -> NormalizedFilePath
   -> IO ()
 deleteValue ShakeExtras{dirtyKeys, state} key file = do
-    void $ modifyVar' state $ HMap.delete (toKey key file)
+    atomically $ STM.delete (toKey key file) state
     atomicModifyIORef_ dirtyKeys $ HSet.insert (toKey key file)
```
michaelpj, line 445: This pairing seems a bit suspicious! Maybe these two modifications *should* be in a transaction together? I guess we can't fix that without changing `dirtyKeys`.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -456,13 +459,12 @@ recordDirtyKeys ShakeExtras{dirtyKeys} key file = withEventTrace "recordDirtyKey
...
   IdeRule k v =>
-  Var Values ->
+  Values ->
   k ->
   NormalizedFilePath ->
   IO (Maybe (Value v, Vector FileDiagnostic))
 getValues state key file = do
-    vs <- readVar state
-    case HMap.lookup (toKey key file) vs of
+    atomically (STM.lookup (toKey key file) state) >>= \case
```
michaelpj, line 467: ditto about returning STM

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -786,8 +788,9 @@ garbageCollectKeys :: String -> Int -> CheckParents -> [(Key, Int)] -> Action [K
 garbageCollectKeys label maxAge checkParents agedKeys = do
     start <- liftIO offsetTime
     extras <- getShakeExtras
-    (n::Int, garbage) <- liftIO $ modifyVar (state extras) $ \vmap ->
-        evaluate $ foldl' removeDirtyKey (vmap, (0,[])) agedKeys
+    let values = state extras
+    (n::Int, garbage) <- liftIO $ atomically $
```
michaelpj, line 792: given your previous changes involved pushing `atomically` inside lops, should this one also be inside the loop?

michaelpj (approved):

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: Yeah, these should be mutated atomically. It's not hard to change `dirtyKeys`, I'll add it to this PR

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: I agree, it makes sense to move the `atomically` to the call site in these cases

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: Yeah, but it also needs to be done atomically with `removeDirtyKey`.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -437,9 +438,9 @@ deleteValue
   -> k
   -> NormalizedFilePath
   -> IO ()
-deleteValue ShakeExtras{dirtyKeys, state} key file = do
-    void $ modifyVar' state $ HMap.delete (toKey key file)
-    atomicModifyIORef_ dirtyKeys $ HSet.insert (toKey key file)
+deleteValue ShakeExtras{dirtyKeys, state} key file = atomically $ do
```
michaelpj, line 441: this one should maybe have got pushed to the call sites too? particularly paralleling `setValues`

PR #2434, opened 2021-12-03
Lockless diagnostics

pepeiborra:
Fourth instalment of #2357 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2434"></a>

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -247,7 +247,7 @@ loadSessionWithOptions SessionLoadingOptions{..} dir = do
                       } <- getShakeExtras
     let invalidateShakeCache = do
             void $ modifyVar' version succ
-            recordDirtyKeys extras GhcSessionIO [emptyFilePath]
+            atomically $ recordDirtyKeys extras GhcSessionIO [emptyFilePath]
 
     IdeOptions{ optTesting = IdeTesting optTesting
               , optCheckProject = getCheckProject
@@ -264,7 +264,7 @@ loadSessionWithOptions SessionLoadingOptions{..} dir = do
               TargetModule _ -> do
                 found <- filterM (IO.doesFileExist . fromNormalizedFilePath) targetLocations
                 return (targetTarget, found)
-          recordDirtyKeys extras GetKnownTargets  [emptyFilePath]
+          atomically $ recordDirtyKeys extras GetKnownTargets  [emptyFilePath]
           modifyVarIO' knownTargetsVar $ traverseHashed $ \known -> do
             let known' = HM.unionWith (<>) known $ HM.fromList $ map (second Set.fromList) knownTargets
             when (known /= known') $
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -193,7 +193,7 @@ data ShakeExtras = ShakeExtras
     ,state :: Values
     ,diagnostics :: Var DiagnosticStore
     ,hiddenDiagnostics :: Var DiagnosticStore
-    ,publishedDiagnostics :: Var (HMap.HashMap NormalizedUri [Diagnostic])
+    ,publishedDiagnostics :: STM.Map NormalizedUri [Diagnostic]
     -- ^ This represents the set of diagnostics that we have published.
     -- Due to debouncing not every change might get published.
     ,positionMapping :: Var (HMap.HashMap NormalizedUri (Map TextDocumentVersion (PositionDelta, PositionMapping)))
@@ -511,7 +511,7 @@ shakeOpen lspEnv defaultConfig logger debouncer
         state <- STM.newIO
         diagnostics <- newVar mempty
         hiddenDiagnostics <- newVar mempty
-        publishedDiagnostics <- newVar mempty
+        publishedDiagnostics <- STM.newIO
         positionMapping <- newVar HMap.empty
         knownTargetsVar <- newVar $ hashed HMap.empty
         let restartShakeSession = shakeRestart ideState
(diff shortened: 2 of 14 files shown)


ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -758,13 +758,11 @@ instantiateDelayedAction (DelayedAction _ s p a) = do
 
 getDiagnostics :: IdeState -> IO [FileDiagnostic]
 getDiagnostics IdeState{shakeExtras = ShakeExtras{diagnostics}} = do
-    val <- readVar diagnostics
-    return $ getAllDiagnostics val
+    atomically $ getAllDiagnostics diagnostics
```
michaelpj, line 761: push to call-sites?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -758,13 +758,11 @@ instantiateDelayedAction (DelayedAction _ s p a) = do
...
 
 getHiddenDiagnostics :: IdeState -> IO [FileDiagnostic]
 getHiddenDiagnostics IdeState{shakeExtras = ShakeExtras{hiddenDiagnostics}} = do
-    val <- readVar hiddenDiagnostics
-    return $ getAllDiagnostics val
+    atomically $ getAllDiagnostics hiddenDiagnostics
```
michaelpj, line 765: push to call sites?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1154,30 +1152,26 @@ updateFileDiagnostics fp k ShakeExtras{logger, diagnostics, hiddenDiagnostics, p
...
+        update new store = setStageDiagnostics uri ver (T.pack $ show k) new store
     mask_ $ do
         -- Mask async exceptions to ensure that updated diagnostics are always
         -- published. Otherwise, we might never publish certain diagnostics if
         -- an exception strikes between modifyVar but before
         -- publishDiagnosticsNotification.
-        newDiags <- modifyVar diagnostics $ pure . update (map snd currentShown)
-        _ <- modifyVar hiddenDiagnostics $  pure . update (map snd currentHidden)
+        newDiags <- liftIO $ atomically $ update (map snd currentShown) diagnostics
+        _ <- liftIO $ atomically $ update (map snd currentHidden) hiddenDiagnostics
```
michaelpj, line 1162: should these be in one transaction?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -191,9 +191,9 @@ data ShakeExtras = ShakeExtras
     ,logger :: Logger
     ,globals :: Var (HMap.HashMap TypeRep Dynamic)
     ,state :: Values
-    ,diagnostics :: Var DiagnosticStore
-    ,hiddenDiagnostics :: Var DiagnosticStore
-    ,publishedDiagnostics :: Var (HMap.HashMap NormalizedUri [Diagnostic])
+    ,diagnostics :: STMDiagnosticStore
```
michaelpj, line 194: Is the `DiagnosticStore` type now dead?

michaelpj (approved):

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: I think it comes from the `lsp` package

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: I don't believe so (a single transaction will introduce more contention in exchange for no more correctness)

pepeiborra: notice that this block is already `mask`ed

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -447,10 +447,11 @@ recordDirtyKeys
...
   -> k
   -> [NormalizedFilePath]
-  -> IO ()
-recordDirtyKeys ShakeExtras{dirtyKeys} key file = withEventTrace "recordDirtyKeys" $ \addEvent -> do
-    atomically $ modifyTVar' dirtyKeys $ \x -> foldl' (flip HSet.insert) x (toKey key <$> file)
-    addEvent (fromString $ "dirty " <> show key) (fromString $ unlines $ map fromNormalizedFilePath file)
+  -> STM (IO ())
+recordDirtyKeys ShakeExtras{dirtyKeys} key file = do
+    modifyTVar' dirtyKeys $ \x -> foldl' (flip HSet.insert) x (toKey key <$> file)
+    return $ withEventTrace "recordDirtyKeys" $ \addEvent -> do
```
michaelpj, line 453: Hmm interesting. I guess we may end up with this happening a bit if we push STM out further and we hit more places where we want to do tracing. I wonder if this is a pattern that appears elsewhere: this is something like a "final" IO action that should be run only after the STM action succeeds.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -247,7 +247,7 @@ loadSessionWithOptions SessionLoadingOptions{..} dir = do
                       } <- getShakeExtras
     let invalidateShakeCache = do
             void $ modifyVar' version succ
-            recordDirtyKeys extras GhcSessionIO [emptyFilePath]
+            atomically $ recordDirtyKeys extras GhcSessionIO [emptyFilePath]
```
fendor, line 250: I just noticed, this changes the type of `invalidateShakeCache` to `:: IO (IO ())`. I suppose that is unintentional?

fendor: I think, either here we need a `join` or have to consume the action produced at: https://github.com/haskell/haskell-language-server/pull/2434/files#diff-d7140bbfca1b7aaafd8c82d1d6aa21a3f941eb27ba8f08e4d9fb13766b21b418L393

Issue #767, opened 2020-12-31
Hover on inner signature gives confusing information


jacg:
Hovering on the *signature* of `inner` in the following sample, gives the signature of `outer`. Hovering on the *definition* of `inner` gives the correct signature of `inner`.

````haskell
outer :: Bool
outer = undefined where

  inner :: Char
  inner = undefined
````

Tests in haskell/ghcide#311.

**Comment by jneira** [comment]
Tests were moved to hls and there are here: 

https://github.com/haskell/haskell-language-server/blob/2e59c60d9f0b2ecaf87fb4551ec6e478e0219330/ghcide/test/exe/Main.hs#L4011-L4012

With hls-1.5.2 i got

```
outer :: Bool
Defined at D:\ws\haskell\issues\empty\Main.hs:6:1
```

```
inner
Defined at D:\ws\haskell\issues\empty\Main.hs:9:3
```

So it seems to me that outer is right and no broken, as stated by the test

jneira:
Checking both tests in #2435, i would expect the outer one should be ok and the inner fail

hasufell:
I get no signature:
![Screenshot_2022-07-13_18-52-53](https://user-images.githubusercontent.com/1241845/178788682-31efc931-f7c2-41e4-b19a-d9a820f1b7eb.png)

PR #2435, opened 2021-12-03
Enable top-level hover signature test

jneira:
* To check the status of #767

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2435"></a>


jneira:
Results: 

```
      top-level signature              #767:                                                              FAIL (1.02s)
        test/exe/Main.hs:3890:
        expected: Position {_line = 47, _character = 1}
         but got: Position {_line = 51, _character = 1}
        Use -p '/top-level signature              #767/' to rerun this test only.
      inner     signature              #767:                                                              FAIL (1.06s)
        test/exe/Main.hs:3896:
        failed to find: `Char` in hover message:
```

So top-level signature works but in another position and the inner fails as expected

ghcide/test/exe/Main.hs
```diff
@@ -3954,7 +3954,7 @@ findDefinitionAndHoverTests = let
   chrL36 = Position 41 24  ;  litC   = [ExpectHoverText ["'f'"]]
   txtL8  = Position 12 14  ;  litT   = [ExpectHoverText ["\"dfgy\""]]
   lstL43 = Position 47 12  ;  litL   = [ExpectHoverText ["[8391 :: Int, 6268]"]]
-  outL45 = Position 49  3  ;  outSig = [ExpectHoverText ["outer", "Bool"], mkR 46 0 46 5]
+  outL45 = Position 49  3  ;  outSig = [ExpectHoverText ["outer", "Bool"], mkR 50 0 50 5]
```
fendor, line 3957: Sorry, I don't know this test, do you change it because the test did not check what we thought it did, or would it make sense to have both test lines?

jneira: hi, I revised the linked issue manually and I checked hover on top level type signatures works, but no inner ones

But the test still was failing cause the position was different. Correcting the position it passes

fendor: Awesome, nice!

fendor (approved):
LGTM!

PR #2439, opened 2021-12-04
Makes local record field completion respects the fields sharing one single type signature

konn:
Currently, HLS generates wrong completion snippets for a record with some fields sharing the same type signature.
For example, consider the following:

```hs
data Foo = Foo { arg1, arg2 :: Int, arg3 :: Int, arg4 :: Bool }
```

Above, `arg1, arg2` shares the same `Int` signature and this is a completely legitimate definition.
If this is defined in a local module, the current HLS generates the following record snippet for `Foo`, which lacks `arg2`:

```hs
Foo { arg1 = _arg1, arg3 = _arg3, arg4 = _arg4 }
```

The base cause of this seems as follows.
HLS handles record snippets for datatypes in a local module via parsed modules (`findRecordCompl` logic), especially from the `RecCon` constructor.
In `RecCon`, field information is stored as a list of `ConDeclField`s, which stores field label information in `cd_fld_names` field.
This `cd_fld_names` is *a list of* `LFieldOcc`s, and it seems that this nested structure reflects the signature-sharing relation.
The current implementation in HLS, however, picks only the first element of `cd_fld_names`, resulting in the incomplete completion snippet.

This PR aims at fixing this issue by just returning lists as-is.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2439"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -20,7 +20,6 @@ import           Data.List.Extra                          as List hiding
 import qualified Data.Map                                 as Map
 
 import           Data.Maybe                               (fromMaybe, isJust,
-                                                           listToMaybe,
                                                            mapMaybe)
 import qualified Data.Text                                as T
 import qualified Text.Fuzzy.Parallel                      as Fuzzy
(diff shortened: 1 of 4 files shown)


konn:
Oops, I somehow miss-clicked "remove request" link. I've just re-sent review request again.

Ailrun (approved):
LGTM

PR #2444, opened 2021-12-06
Remove support for ghc 8.8.3/8.10.5

jneira:
* Following the [ghc deprecation policy and schedule](https://haskell-language-server.readthedocs.io/en/latest/supported-versions.html) and as announced in the [1.5.0 release](https://github.com/haskell/haskell-language-server/releases/tag/1.5.0)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2444"></a>

PR #2446, opened 2021-12-07
CI: organizing bootstraping

Anton-Latukha:
Be free at merging #2444, I would rebase this to whatever needed.

---

Using the same bootstrapping process so workflows have the same setup & so can share the caches between them.

<sub>(A work derived from #2427)</sub>

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2446"></a>

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,5 +1,9 @@
 name: Benchmark
 
+defaults:
+  run:
+    shell: bash
+
 # See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency.
 concurrency:
   group: ${{ github.head_ref }}-${{ github.workflow }}
@@ -72,17 +76,14 @@ jobs:
     - run: cabal configure --enable-benchmarks --max-backjumps 12000
 
     - name: Build
-      shell: bash
       run: cabal build ghcide:benchHist
 
     - name: Bench init
-      shell: bash
       run: cabal bench ghcide:benchHist  -j --benchmark-options="all-binaries"
 
     # tar is required to preserve file permissions
     # compression speeds up upload/download nicely
     - name: tar workspace
-      shell: bash
       run: tar -czf workspace.tar.gz * .git
 
     - name: tar cabal
@@ -142,11 +143,9 @@ jobs:
         tar xzf cabal.tar.gz --directory ~/.cabal
 
     - name: Bench
-      shell: bash
       run: cabal bench ghcide:benchHist  -j --benchmark-options="${{ matrix.example }}"
 
     - name: Display results
-      shell: bash
       run: |
         column -s, -t < ghcide/bench-results/unprofiled/${{ matrix.example }}/results.csv | tee ghcide/bench-results/unprofiled/${{ matrix.example }}/results.txt
 
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -36,8 +36,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-      with:
-        submodules: true
 
     - run: git fetch origin master # check the master branch for benchmarking
 
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -52,11 +52,6 @@ jobs:
         cabal-version: ${{ matrix.cabal }}
         enable-stack: false
 
-    - name: Platform config
-      run: |
-        echo "CABAL_STORE_DIR=~/.cabal/store" >> $GITHUB_ENV
-        echo "CABAL_PKGS_DIR=~/.cabal/packages" >> $GITHUB_ENV
-
     - name: Cache Cabal
       uses: actions/cache@v2
       with:
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -34,6 +34,9 @@ jobs:
         os: [ubuntu-latest]
         cabal: ['3.6']
 
+    # This code is fitted to the strategy: assumes Linux is used ... etc,
+    # change of the strategy may require changing the bootstrapping/run code
+
     steps:
     - uses: actions/checkout@v2
 
@@ -45,38 +48,10 @@ jobs:
         cabal-version: ${{ matrix.cabal }}
         enable-stack: false
 
-    - if: runner.os == 'Windows'
-      name: (Windows) Platform config
-      run: |
-        echo "CABAL_STORE_DIR=$SYSTEMDRIVE\\SR" >> $GITHUB_ENV
-        echo "CABAL_PKGS_DIR=~\\AppData\\cabal\\packages" >> $GITHUB_ENV
-    - if: ( runner.os == 'Linux' ) || ( runner.os == 'macOS' )
-      name: (Linux,macOS) Platform config
+    - name: Platform config
       run: |
         echo "CABAL_STORE_DIR=~/.cabal/store" >> $GITHUB_ENV
         echo "CABAL_PKGS_DIR=~/.cabal/packages" >> $GITHUB_ENV
-
-    # Needs to be before Cache Cabal so the cache can detect changes to the modified cabal.project file
-    - if: matrix.ghc == '9.0.1'
-      name: (GHC 9.0.1) Use modified `cabal.project`
-      run: |
-        cp cabal-ghc901.project cabal.project
-    - if: runner.os == 'Windows' && matrix.ghc == '8.8.4'
-      name: (Windows,GHC 8.8) Modify `cabal.project` to workaround segfaults
-      run: |
-        echo "package floskell" >> cabal.project
-        echo "  ghc-options: -O0" >> cabal.project
-
-    # Shorten binary names as a workaround for filepath length limits in Windows,
-    # but since tests are hardcoded on this workaround -
-    # all platforms (in 2021-12-07) need it.
-    - name: Workaround shorten binary names
-      run: |
-        sed -i.bak -e 's/haskell-language-server/hls/g' \
-                   -e 's/haskell_language_server/hls/g' \
-                   haskell-language-server.cabal cabal.project
-        sed -i.bak -e 's/Paths_haskell_language_server/Paths_hls/g' \
-                   src/**/*.hs exe/*.hs
 
     - name: Cache Cabal
       uses: actions/cache@v2
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -36,6 +36,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        submodules: true
 
     - run: git fetch origin master # check the master branch for benchmarking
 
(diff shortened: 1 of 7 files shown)


Anton-Latukha:
On `caching` workflow: added the PR trigger, ran the successful build/test on `caching` workflow & removed the trigger.

.github/workflows/bench.yml
```diff
@@ -36,6 +36,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
```
jneira, line 39: We have no submodules nor plans to have hem so i think we could remove it from all workflows

.github/workflows/caching.yml
```diff
@@ -1,8 +1,8 @@
...
+# 1. GitHub cache scoping goes: main branch -> PR main branch -> PR. (essentially in a tree fashion). Building & keeping caches on `master` allows to the main project state cache be shared to the whole tree.
+# 2. GitHub has a 10G as default limit of cache pool per repo. HLS is a big community project & monorepo of many projects, so cache pool can get exhausted & invaidated easilly. Because of that, to keep cache useful - the main branch state caches should be preserved & kept well below the limit & that preferred to the PR caches, since PRs from internal branches - would count into the repo 10G pool, while PR cache is available only inside of the PR, which would bork cache fore the whole community. That is a short story why `dist-newstyle` (especially full) - is not includded into `master` cache.
```
jneira, line 5: Those comments are gold, any chance to have them in multiple lines to help in their reading?

.github/workflows/test.yml
```diff
@@ -123,14 +131,9 @@ jobs:
 
       - run: cabal update
 
-      # Need this to work around filepath length limits in Windows
-      - name: Shorten binary names
+      - name: "HLint via ./fmt.sh"
```
jneira, line 134: why moving hlint here? just curious

jneira (commented):
Thanks for reorganizing ci scripts, they will look great, only some minor comments

jneira:
[circleci skip]

.github/workflows/test.yml
Anton-Latukha: 1. To have the same sequence of actions between workflows.

2. To do bootstrap first & then do everything else. Bootstrap steps are ~ instant anyway, except for fast `cabal v2-update`. So there is no need to not look/track what is happening in/during the `hlint` oneliner script, does it install something additional & does that depend on the Hackage state. & install can be made faster if what it wants to install is already in the cache. So there goes the quick bootstrapping phase & then quick `hlint` install & goes code that analyzes & tests the project.

Overall 1 run of `hlint` is mostly enough. It can run just on the latest GHC. & so if it needs to look between CPP - in time - all the code would get looked through. Would ping upstream on that action, there was almost a month after submission of the action.

.github/workflows/bench.yml
```diff
@@ -45,6 +47,39 @@ jobs:
         cabal-version: ${{ matrix.cabal }}
         enable-stack: false
 
+    - if: runner.os == 'Windows'
```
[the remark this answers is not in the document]
Anton-Latukha: Yes. Tried to keep the same code between workflows.

It is an interesting question :thinking:. From the side of code sharing, it is seamless, generalized & not "overfitted" to local code, but from the workflow point of view it can be said to have superfluous features, it sounds like something covered in some book.

.github/workflows/caching.yml
Anton-Latukha: Yes. It is useful to state something like this at the top of the workflow (or any design that is restricted by external things), to save people time, especially if they had some obvious ideas to improve caching while not knowing details of restrictions.

Ok. Would also add links there.

.github/workflows/bench.yml
Anton-Latukha: Well, I picked the submodule part from some workflow checkout & since trying to make everything seamless - included the submodule directive. Ok removing it.

.github/workflows/bench.yml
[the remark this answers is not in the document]
jneira: hmm if we got to reuse steps between workflows it would be needed but i would prefer to not have unused steps even losing the overall consistency.

Anton-Latukha:
Would try to remove `bash` shell entries where *afaik* that entry is not needed.

I remember `bash` entry to being a workaround for something in some situations, but *idk* if it is needed or still applies.

Anton-Latukha:
[circleci skip]

jneira:
ha I forgot the key is [skip circleci] 🤦‍♂️
it works if it is in one comment though

.github/workflows/bench.yml
```diff
@@ -45,6 +52,28 @@ jobs:
         cabal-version: ${{ matrix.cabal }}
         enable-stack: false
 
+    - name: Platform config
+      run: |
+        echo "CABAL_STORE_DIR=~/.cabal/store" >> $GITHUB_ENV
+        echo "CABAL_PKGS_DIR=~/.cabal/packages" >> $GITHUB_ENV
+
+    # Needs to be before Cache Cabal so the cache can detect changes to the modified cabal.project file
+    - if: matrix.ghc == '9.0.1'
```
jneira, line 61: this section could be removed

.github/workflows/bench.yml
```diff
@@ -45,6 +52,28 @@ jobs:
...
+      name: (GHC 9.0.1) Use modified `cabal.project`
+      run: |
+        cp cabal-ghc901.project cabal.project
+
+    # Shorten binary names as a workaround for filepath length limits in Windows,
+    # but since tests are hardcoded on this workaround -
+    # all platforms (in 2021-12-07) need it.
+    - name: Workaround shorten binary names
```
jneira, line 69: and this one, sorry to not be clearer in my first comment

.github/workflows/bench.yml
Anton-Latukha: Ok. Wanted to keep it so people would not forget it on the next update.

.github/workflows/bench.yml
Anton-Latukha: Well, ok.

It is strange to have this be literally required for `test`s, but not present in `bench`.

For me it is was surprising that running tests on Linux require this Windows workaround.

Anton-Latukha:
> [skip circleci]

Where is to find that information.

I found, https://circleci.com/docs/2.0/skip-build/ but that talks about `[skip ci]` & mentions to not work for the fork PRs.

jneira:
> > [skip circleci]
> 
> Seems I encountered that in several forms. Where is to find that information?
> 
> I found, https://circleci.com/docs/2.0/skip-build/ but that talks about `[skip ci]` & mentions to not work for the fork PRs.

yeah and it only works in commit messages, I made up a hack to skip the build 
see 

https://github.com/haskell/haskell-language-server/blob/ddacbced0f983a0b41dbb8a3b5340a54052095a1/.circleci/config.yml#L21

.github/workflows/bench.yml
jneira: in fact it is not needed, it could be under a condition

jneira (approved):
Looks good, thanks for addressing comments

Anton-Latukha:
Interesting. I re-ran the CI checks. & Circle-CI still reports green in https://github.com/haskell/haskell-language-server/pull/2446/checks & here it is still "Expected — Waiting for status to be reported". Well.

jneira:
it happens sometimes, have you tried a forced push?

Anton-Latukha:
After stopping skipping with `[skip circeci]` & after several force-pushes & after the addition of `circle-ci` config change.

Still the same. I do not know what API bug or quota is going on here, but I hope this would happen only to this PR.

Anton-Latukha:
Oh, in the build section it updates from separate build reports to a workflow now.

.github/workflows/caching.yml
```diff
@@ -55,42 +67,45 @@ jobs:
...
+        name: (GHC 9.0.1) Use modified `cabal.project`
+        run: |
+          cp cabal-ghc901.project cabal.project
+      - if: runner.os == 'Windows' && matrix.ghc == '8.8.4'
+        name: (Windows,GHC 8.8) Modify `cabal.project` to workaround segfaults
         run: |
           echo "package floskell" >> cabal.project
           echo "  ghc-options: -O0" >> cabal.project
 
+      # Shorten binary names as a workaround for filepath length limits in Windows,
```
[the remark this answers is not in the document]
Anton-Latukha: Too long to explain.

It is connected with that builds & runs that use the cache & the cache building - should share the same project setup. And with that that further down the line the full build would be run in caching, because Cabal simply does not allow otherwise (so far does not support that command, to build dependencies of all targets).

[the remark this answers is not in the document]
Anton-Latukha: & since Cabal provides an interface to get configuration information - I am not sure that executable names are not involved in use somewhere in the code, so I just preserved the same bootstrapping.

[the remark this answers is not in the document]
Anton-Latukha: Removed it.

jneira (approved):
many thanks again, i hope my succesive cmments were not too burdensome

Anton-Latukha:
Well, besides we could established minimization a bit earlier. I do not know how to properly establish that. I wanted to copy-paste the bootstrap code, I did not wanted to fit it to the workflows, so it would be the same & essentially sacked or cleaned-up afterward. & I also had an eyeball to preserve the same code to submit the next steps, & so some of them would be used & then look at & the clean-up.  But that is probably too much of a stretch to multiple PRs.

I basically wanted to get 1 PR & split its code into 3 stages & merge them & then do clean-up. But as I said that is probably a stretch & submitting PRs which code has unused features & code that may be used in the future, that is: ["4. Code with added features"](https://youtu.be/lKXe3HUG2l4?t=538).

A kind & strong request to fit the code to the purpose so maintainer does not need to go over it, would've made me go over the whole codebase more thoroughly & submit all additional code that is not related to current code into next PRs where it would relate. & the law of triviality & the understanding of that looms over the process. Several small details some put & some removed & the [last](https://github.com/haskell/haskell-language-server/pull/2446#discussion_r766582777) note probably still would've happened nevertheless. It all also was heavily multiplied by numerous CI blockages & waiting times in the process, rebases.

What I want to take from this is to on splitting the PR I should treat each small PR as a one PR & code "in itself".

.github/workflows/caching.yml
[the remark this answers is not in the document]
Anton-Latukha: In fact, this code should been there all along.

Because the caching key defined through: `-${{ hashFiles('cabal.project') }}`
& that live patching actively:
```yaml
          sed -i.bak -e 's/haskell-language-server/hls/g' \
                     -e 's/haskell_language_server/hls/g' \
                     haskell-language-server.cabal cabal.project
          sed -i.bak -e 's/Paths_haskell_language_server/Paths_hls/g' \
                     src/**/*.hs exe/*.hs
```

Changes the `cabal.project`, which means that cache does not get hit directly & so workflows that have different `cabal.project` store the same hash as many times.

PR #2450, opened 2021-12-08
Fix rerun log cache handling

jneira:
* The cache of tasty logs to not repeat succesful tests between attempts of the same run was not working correctly
* `new` makes rerun all tests again, as succesful one in one attempt are "new" for the next one
* Logs are saved in each package root dir, so the cache should be `**./.tasty-rerun-log*`
* `func-tests` and `wrapper-tests` are from the same hls top level package so at least one of them needs a specific name (did it for wrapper tests)
* I've checked a second rerun does not run any effective test here: https://github.com/jneira/haskell-language-server/actions/runs/1550353254

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2450"></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -138,14 +138,14 @@ jobs:
 
       - name: Set test options
         run: |
-          echo "TEST_OPTS=-j1 --rerun-update --rerun-filter failures,exceptions,new" >> $GITHUB_ENV
+          echo "TEST_OPTS=-j1 --rerun-update --rerun-filter failures,exceptions" >> $GITHUB_ENV
 
       - name: Cache test log bewteen attempts of the same run
         uses: actions/cache@v2
         env:
           cache-name: cache-test-log
         with:
-          path: .tasty-rerun-log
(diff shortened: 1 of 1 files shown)


pepeiborra:
The approach of preserving rerun logs across test runs doesn't seem entirely safe to me:

1. I send the empty PR and wait for it to pass all the tests (you would hope)
2. I send a commit that breaks X tests, but since X had already passed it doesn't rerun.
3. The commit gets reviewed, approved and merged
4. HEAD is now broken

jneira:
> Let me know if I am misunderstanding how it works

Sorry i missed to stress the more important thing: the cache is `v1-${{ runner.os }}-${{ matrix.ghc }}-test-log-${{ github.sha }}` so it is per os, ghc version and *commit*. If the test suite passes for a concrete commit i think it is quite safe to skip it in the next attempt. It only can be affected by external things like dependencies (and we have pinned the hackage index) or other things which might be transient.

pepeiborra (approved):

PR #2453, opened 2021-12-08
lock-less progress-reporting

pepeiborra:
6th? instalment of #2357


ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -143,7 +157,8 @@ delayedProgressReporting before after lspEnv optProgressStyle = do
                 loop _ _ | optProgressStyle == NoProgress =
                     forever $ liftIO $ threadDelay maxBound
                 loop id prev = do
-                    InProgress{..} <- liftIO $ readVar inProgress
+                    done <- liftIO $ readTVarIO doneVar
+                    todo <- liftIO $ readTVarIO todoVar
```
michaelpj, line 161: Should these be in an atomic transaction? what happens if someone updates `todo` between the two lines?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -69,26 +73,37 @@ updateState _     StopProgress          (Running a) = cancel a $> Stopped
 updateState _     StopProgress          st          = pure st
 
 -- | Data structure to track progress across the project
-data InProgress = InProgress
-    { todo    :: !Int  -- ^ Number of files to do
-    , done    :: !Int  -- ^ Number of files done
-    , current :: !(HMap.HashMap NormalizedFilePath Int)
+data InProgressState = InProgressState
```
michaelpj, line 76: What do we gain from having this be a structure containing TVars rather than itself just contained within a TVar? It seems like most accesses of this structure work with all the parts? i.e. `recordProgress`

If you just had a `TVar InProgressState` I guess you wouldn't need the funny stuff in the tests either.

michaelpj: The best argument i can think of is that `recordProgress` could conflict with the reads down on line 160. But having multiple TVars doesn't help with that, since `recordProgress always modifies all the `TVars`, no?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: Atomicity is not crucial here as the values are only used to display a progress report and it's ok to be off by one.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
pepeiborra: The gain is in using an `STM.Map` (a HAMT with TVars in the branch nodes)  which has much lower contention than a `TVar InProgressState`.

michaelpj: But how could that happen? Maybe I'm missing something about how STM works, but doesn't `recordProgress` always write to `todo`, `done` (and part of `current`), atomically? And so all progress-updating transactions are guaranteed to contend with each other? So how could we update progress for multiple files?

pepeiborra: At the moment, yes. But in a follow-up PR, this will be relaxed: https://github.com/haskell/haskell-language-server/pull/2357/commits/b47ab52acbfd494e7187c98d845d4c36ec2d3d0a

pepeiborra: And it can be relaxed even more by extracting the `todo` and `done` writes out into a separate transaction

michaelpj: Makes sense. Glad I'm not going crazy then :)

pepeiborra:
@michaelpj any other comments?

michaelpj (approved):
Oops, meant to approve to begin with!

PR #2458, opened 2021-12-09
#600 Code action to ignore hlint hints module wide

eddiemundo:
ignore-hlint-hint-module-wide-demo

It will also insert a `-Wno-unrecognized-pragmas` options pragma only if that dynflag isn't already set.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2458"></a>


jneira (approved):
Amazing work, including sensible new test cases and nice refactorings, many many thanks

jneira:
There is some problem with the pragmas placement in ubuntu and ghc-8.8.4:

```
 FAIL (1.88s)
      Test output was different from 'test/testdata/UnrecognizedPragmasOn.expected.hs'. Output of ["git","-c","core.fileMode=false","diff","--no-index","--text","--exit-code","test/testdata/UnrecognizedPragmasOn.expected.hs","/tmp/UnrecognizedPragmasOn.expected34086-3.actual"]:
      diff --git a/test/testdata/UnrecognizedPragmasOn.expected.hs b/tmp/UnrecognizedPragmasOn.expected34086-3.actual
      index 564503c..e65d316 100644
      --- a/test/testdata/UnrecognizedPragmasOn.expected.hs
      +++ b/tmp/UnrecognizedPragmasOn.expected34086-3.actual
      @@ -1,5 +1,5 @@
      -{-# OPTIONS_GHC -Wunrecognised-pragmas #-}
       {-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
       {-# HLINT ignore "Eta reduce" #-}
      +{-# OPTIONS_GHC -Wunrecognised-pragmas #-}
       module UnrecognizedPragmasOn where
       foo x = id x
```

eddiemundo:
I think I've fixed the tests, but it's weird.

The issue was me not using the right dynflags in some cpp for the pragmas parser initialization, but that should (and does) fail the Pragmas plugin tests for 8.8.4. Yet I never saw any 8.8.4 test failures from the CI in the PR where I extracted the pragma parsing stuff into its own module. This means that the CI didn't run the Pragmas plugin tests for that PR?

Either that or somehow default dynflags have changed since then, and the Pragma plugin tests were passing back then but not now.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -105,6 +107,15 @@ import qualified Language.LSP.Types.Lens                            as LSP
 import           GHC.Generics                                       (Generic)
 import           Text.Regex.TDFA.Text                               ()
 
+import           Development.IDE.GHC.Compat.Core                    (WarningFlag (Opt_WarnUnrecognisedPragmas),
+                                                                     wopt)
+import           Development.IDE.Spans.Pragmas                      (LineSplitTextEdits (LineSplitTextEdits),
```
jneira, line 112: nice to see the new module in action 👍

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -9,6 +12,7 @@ import           Data.List               (find)
 import qualified Data.Map                as Map
 import           Data.Maybe              (fromJust, isJust)
 import qualified Data.Text               as T
+import qualified Debug.Trace             as Debug
```
jneira, line 15: i guess this should be removed before merging?

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
eddiemundo: yup and at least 1 more use of it to come

jneira (approved):
Really a great work, happy to see a contibution to my beloved hlint plugin
looking forward to your next contributions

PR #2459, opened 2021-12-09
Join nested IO actions of the form `IO (IO ())`

fendor:
I think these are some accidental remnants of #2434. 

Nothing broke for me, I just reviewed code and noticed the type of `invalidateShakeCache` was inferred to `IO (IO ())`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2459"></a>


jneira (approved):
thanks for the clean up

pepeiborra:
Are there any warnings that could have prevented this?

michaelpj:
Isn't there a warning saying a do-block expression is discarding a result? I'd have expected that to fire here.

fendor:
> Isn't there a warning saying a do-block expression is discarding a result? I'd have expected that to fire here.

Yeah, same, but it didn't. No idea why not.

fendor:
We can use `-Wunused-do-bind`, will try it out.

jneira:
> I thought we are already using `-Wunused-do-bind`, but maybe we aren't, I'll give it a try.

In this pr? or could we merge it?

fendor:
Really weird, no usage of `-Wunused-do-bind` works on ghcide, but it works in my example project... 

image

This is interesting. I don't know what is going on.

fendor:
Let's look into some more details before merging, imo.

fendor:
Very interesting: 
image

fendor:
I don't understand, copying that code to another project highlights correctly both occurrences. Even deleting everything except these two functions doesnt work for me, i.e. still shows only one warning.

pepeiborra:
I found it: it's `ApplicativeDo`, which is enabled in `ghcide.cabal`. Please disable it @fendor since it doesn't make any difference with hls-graph anymore.

pepeiborra (approved):
Thanks for catching this!

Issue #2461, opened 2021-12-09
Distribute Alpine binaries


pepeiborra:
As agreed in #2431 we want to be distributing Alpine binaries. Unfortunately Github doesn't offer any Alpine workers, so we cannot reuse the existing build Github action to generate them.

However, the Gitlab CI contributed by @hasufell is already generating these binaries, e.g.:

https://gitlab.haskell.org/maerwald/haskell-language-server/-/pipelines/44731

Things to be done: 
1. Record releases in the Gitlab project in order to generate release metadata
2. Modify the downloader to fetch the Alpine Linux binaries from Gitlab: https://gitlab.haskell.org/maerwald/haskell-language-server/-/pipelines/44731

I don't know if we want to grab only the Linux binaries from Gitlab, or also the Windows and Mac OS binaries. 
This would also be a good opportunity to distribute Apple M1 binaries for the supported GHC versions (#1896)

hasufell:
> Unfortunately Github doesn't offer any Alpine workers, so we cannot reuse the existing build Github action to generate them.

You can use Github actions for that too: https://github.com/hasufell/stack2cabal/blob/master/.github/workflows/release.yaml

But IMO, it would make more sense to focus all release CI on GitLab.

jneira:
I am not against the use of gitlab releases but we have to consider some points:
* To have only the release build and no the test or bench ones could make the release build less reliable. We are exercising the gh setup for every pr (and recently also in master), executing the test suite and bench for the workflow artifact. So if we want to keep the gitlab release up to date and reliable we should move or copy the bench and test workflows to gitlab and sync automatically both repos.
  * If we move the test and bench workflows we should find a way to keep the branch protection rules in githubagainst the gitlab workflows. Ultimately it would ask to make gitlab the main repo. 
* We should complete the gitlab scripts to make it generate artifacts matching the gh scheme: one tar per suppported ghc and os+arch and one tar per os+arch  with all ghc versions for ghcup. It lacks a supported minor ghc versions (8.10.6) and other minor things like generating the sha256sums. Not a big deal but has to be done.
* hls workflows will have to compete for ci resources with other gitlab projects, specially with ghc itself, which consumes lot of ci resources. We will lose the resources boost github gives us recently.
* As noted we should have to change the vscode extension to fetch releases from gitlab. We should keep compatibility with github now we are downloading old reelases, or copy all the missed releases to gitlab.

hasufell:
[...]

PR #2463, opened 2021-12-10
Build linux binaries in alpine container

pepeiborra:
Shortest path towards #2461
Moving to Gitlab CI binaries is still very much desirable but also much more work.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2463"></a>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,12 @@ jobs:
       with:
         ghc-version  : ${{ matrix.ghc   }}
         cabal-version: ${{ matrix.cabal }}
-
+# some alpines need workaround
+    - name: Force integer-simple
+      run: |
+        if ghc --info | grep -q integer-simple ; then
+          echo -e 'package blaze-textual\n    flags: +integer-simple' >> cabal.project.local
+        fi
     - name: Use modified cabal.project for ghc9
       if: ${{ matrix.ghc == '9.0.1' }}
       run: cp cabal-ghc901.project cabal.project
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,16 +12,22 @@ on:
       - '*-check-build'
 
 jobs:
-  build-mac-win:
+  build:
     runs-on: ${{ matrix.os }}
+    container: ${{ (matrix.os == 'ubuntu-18.04' && 'alpine:3.12') || '' }}
     strategy:
       fail-fast: false
       matrix:
         ghc: ['9.0.1', '8.10.7', '8.10.6', '8.8.4', '8.6.5']
-        os: [macOS-latest, windows-latest]
+        os: [ubuntu-18.04, macOS-latest, windows-latest]
         cabal: ['3.6']
 
     steps:
+    - name: Install system dependencies
+      if: matrix.os == 'ubuntu-18.04'
+      run: |
+        apk add --no-cache curl gcc g++ gmp-dev ncurses-dev libffi-dev make xz gzip tar perl git bash sudo binutils-gold
+        apk add --no-cache zlib zlib-dev zlib-static gmp gmp-dev ncurses-static
     - uses: actions/checkout@v2
 
     - uses: haskell/actions/setup@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -154,9 +152,134 @@ jobs:
         name: haskell-language-server-wrapper-${{ runner.OS }}${{env.EXE_EXT}}.${{ steps.compress_wrapper_binary.outputs.extension }}
         path: ${{ steps.compress_wrapper_binary.outputs.path }}
 
+  build-linux:
+    runs-on: ubuntu-latest
+    container: alpine:3.12
```
jneira, line 157: could not this be set conditionally to not have to duplicate the entire job?

pepeiborra: Do you know how?

jneira:
```suggestion
    container: ${{ (matrix.os == 'ubuntu-18.04' && 'alpine:3.12') || '' }}
```
https://github.com/haskell/haskell-language-server/compare/alpine-check-build...jneira:alpine-check-build?expand=1

this trick seems to work, i.e. it starts ubuntu in a container and the rest without a container, but the build within the container is failing due to other errors: 
https://github.com/jneira/haskell-language-server/actions/runs/1564975136

jneira: And the fail seems to be related with the install of ghc using ghcup and apt-get:

``` [ Info  ] downloading: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.6.yaml as file /github/home/.ghcup/cache/ghcup-0.0.6.yaml
  ghcup: curl: executeFile: does not exist (No such file or directory)
  [ Warn  ] Could not get download info, trying cached version (this may not be recent!)
  [ ...   ] If this problem persists, consider switching downloader via: 
  [ ...   ]     ghcup config set downloader Wget
  [ Error ] JSON decoding failed with: YAML exception:
  [ ...   ] Yaml file not found: /github/home/.ghcup/cache/ghcup-0.0.6.yaml
  [ ...   ] Consider removing /github/home/.ghcup/cache/ghcup-0.0.6.yaml manually.
  [ ...   ] 
```

and there is no sudo in alpine it seems:

```
Error: Unable to locate executable file: sudo. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

jneira: https://stackoverflow.com/questions/49225976/use-sudo-inside-dockerfile-alpine

jneira: i think we should try to use the same image than gitlab:

https://github.com/haskell/haskell-language-server/blob/779439876855872c523088e400526c6f1807bca2/.gitlab-ci.yml#L127-L129

jneira:
```suggestion
    container: ${{ (matrix.os == 'ubuntu-18.04' && 'docker://registry.gitlab.haskell.org/ghc/ci-images/x86_64-linux-alpine3_12:4ed1a4f27828ba96a34662dc954335e29b470cd2') || '' }}
```

this get the gitlab image

https://github.com/jneira/haskell-language-server/runs/4488270256?check_suite_focus=true

but it has to be tuned to make the rest of the job work: install curl and whatever else, alpine seems to be a minimalistic image

jneira: > but it has to be tuned to make the rest of the job work: install curl and whatever else, alpine seems to be a minimalistic image

oh you already was doing it here: https://github.com/haskell/haskell-language-server/pull/2463/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R168

jneira: The gitlab image throws a permission error trying to do the git checkout. It has cabal/ghc preinstalled, not sure what would be other changes compared with the base alpine:3.12

jneira:
@pepeiborra sorry for the stream of comments, i think this version works like this one (hitting the integer-gmp error) without deduplicating the jobs, feel free to use the bits you find interesting:

https://github.com/haskell/haskell-language-server/compare/alpine-check-build...jneira:alpine-check-build?expand=1

jneira (approved):
lgtm, thanks for the change, let's hope  we can close some issues related with static linking thanks to

PR #2464, opened 2021-12-10
#2418 Also use .hlint.yaml fixity rules when HLINT_ON_LIB_GHC not defined

### LINKED ISSUE #2418: hlint fixity issues
### Your environment

Which OS do you use: Red Hat 7 VM
Which lsp-client do you use: Emacs
Describe your project (alternative: link to the project): Shake + Stack; internal project so can't be shared

### Steps to reproduce

```yaml
# .hlint.yaml
- fixity: "infixl 3 <!>"
```

Haskell source file:
```haskell
-- https://hackage.haskell.org/package/semigroupoids-5.3.6/docs/Data-Functor-Alt.html
import Data.Functor.Alt ((<!>))

-- Simplified function matching behaviour of actual functions
foo :: Char -> Maybe Int -> Maybe String
foo c mInt =
      show <$> mInt
  <!> pure <$> Just c

-- Note the relative fixities here: this is equivalent to: (...) <!> (...)
```

Load this in $EDITOR, list all suggestions from hls.

### Expected behaviour
No warnings related to `foo`.

### Actual behaviour

```
Functor law
 Found:
   show <$> mInt <!> pure <$> Just c
 Why not:
   show . (mInt <!> pure) <$> Just c
```

### Include debug information

Running hlint explicitly does not have any problems.

The `.hlint.yaml` file is indeed being picked up by the hls hlint plugin as other suggestions are appearing based upon the configuration (custom linting rules, etc.).  As such, this might be due to ndmitchell/hlint#1232

`haskell-language-server-wrapper --debug .` output only partially shown as it's quite large:
```
haskell-language-server version: 1.5.1.0 (GHC: 8.10.7) (PATH: /home/<elided>/haskell-language-server) (GIT hash: 1d82c8ec31628fc84f0ffe55ff572602212e07a2)
 ghcide setup tester in /home/<elided>
Report bugs at https://github.com/haskell/haskell-language-server/issues

Step 1/4: Finding files to test in /home/<elided>
Found 16145 files
```

Logs from LSP client:

```
Command "~/<elided>/haskell-language-server --lsp" is present on the path.
Command "~/<elided>/haskell-language-server --lsp" is present on the path.
Found the following clients for /path/to/Module.hs: (server-id lsp-haskell, priority 0)
The following clients were selected based on priority: (server-id lsp-haskell, priority 0)
```

eddiemundo:
Fixes #2418

If [#1325](https://github.com/ndmitchell/hlint/pull/1325#issue-1077062712) is merged then we can modify this fix to remove the `ghc-lib-parser-ex` dependency.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2464"></a>


plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -68,6 +68,7 @@ library
     , text
     , transformers
     , unordered-containers
+    , ghc-lib-parser-ex
```
jneira, line 71: please could you add a comment about when we could remove the dep

jneira (approved):
Fantastic including its appropiate regression test, thanks!

PR #2465, opened 2021-12-11
Adding name to not implemented placeholder errors

sordina:
It's nice to see what isn't implemented when you get an error.

When you use the `define` action on undefined names, you get something like this:

```haskell
step :: Generation -> Generation
step = error "not implemented"
```

But wouldn't it be nicer to see this?

```haskell
step :: Generation -> Generation
step = error "step not implemented"
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2465"></a>


pepeiborra:
Actually, I would prefer if the body of the unimplemented function was just a hole:

```
step :: Generation -> Generation
step = _
```

Could you change it to that instead?

sordina:
@pepeiborra Sure! My workflow actually depends on hitting the errors, so I'm guessing there's a 'defer-holes' kind of option to allow running.

pepeiborra (approved):
Thank you! 
Couple of [tests](https://github.com/haskell/haskell-language-server/runs/4494143978?check_suite_focus=true) need to be updated, but otherwise it looks good to me.

sordina:
@pepeiborra having some issues running the tests right now. I'll come back to this soon.

Anton-Latukha:
Yep, consistently getting the test error.

pepeiborra:
This is already fixed in #2740

PR #2466, opened 2021-12-11
STM debouncer

pepeiborra:
8th? instalment of #2357. This one not only makes the debouncer lock-less but also potentially more efficient.

The debouncer associates delayed actions with project files. It works by waiting for a short _delay_ before executing the action. If a new action is registered for this file before the _delay_ is over, the old action is throw away and the delay restarts.

The old implementation does this using `async`, and a new thread is spawned on every reset. 

This PR replaces threads by STM transactions. Associating an action starts an STM transaction that reads the `TVar` and then sleeps for the `delay`. Reassociating a new action writes to the `TVar` causing any long-lived transactions to `retry`.

STM transactions should be more lightweight than threads but I don't expect this to make a huge difference in practice and wouldn't oppose to reverting the change.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2466"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Debouncer.hs
+++ b/ghcide/src/Development/IDE/Core/Debouncer.hs
@@ -9,12 +9,12 @@ module Development.IDE.Core.Debouncer
     ) where
 
 import           Control.Concurrent.Async
-import           Control.Concurrent.STM.Stats (atomically, atomicallyNamed)
+import           Control.Concurrent.STM
+import           Control.Concurrent.STM.Stats (atomicallyNamed)
 import           Control.Exception
 import           Control.Monad                (join)
-import           Data.Foldable                (traverse_)
 import           Data.Hashable
-import qualified Focus
+import           GHC.Conc                     (unsafeIOToSTM)
 import qualified StmContainers.Map            as STM
 import           System.Time.Extra
 
@@ -39,20 +39,21 @@ newAsyncDebouncer = Debouncer . asyncRegisterEvent <$> STM.newIO
 -- If there is a pending event for the same key, the pending event will be killed.
 -- Events are run unmasked so it is up to the user of `registerEvent`
 -- to mask if required.
-asyncRegisterEvent :: (Eq k, Hashable k) => STM.Map k (Async ()) -> Seconds -> k -> IO () -> IO ()
-asyncRegisterEvent d 0 k fire = do
-    join $ atomically $ do
-        prev <- STM.focus Focus.lookupAndDelete k d
-        return $ traverse_ cancel prev
-    fire
+asyncRegisterEvent :: (Eq k, Hashable k) => STM.Map k (TVar (Seconds, IO())) -> Seconds -> k -> IO () -> IO ()
 asyncRegisterEvent d delay k fire = mask_ $ do
-    a <- asyncWithUnmask $ \unmask -> unmask $ do
-        sleep delay
-        fire
-        atomically $ STM.delete k d
-    do
-        prev <- atomicallyNamed "debouncer" $ STM.focus (Focus.lookup <* Focus.insert a) k d
-        traverse_ cancel prev
+    prev <- atomically $ STM.lookup k d
+    case prev of
+        Just v -> do
+            atomicallyNamed "debouncer - reset" $ writeTVar v (delay, fire)
+        Nothing -> do
+            var <- newTVarIO (delay, fire)
+            _ <- asyncWithUnmask $ \unmask -> unmask $ do
+                join $ atomicallyNamed "debouncer - sleep" $ do
+                    (s,act) <- readTVar var
+                    unsafeIOToSTM $ sleep s
+                    STM.delete k d
+                    return act
+            atomicallyNamed "debouncer2" $ STM.insert var k d
 
 -- | Debouncer used in the DAML CLI compiler that emits events immediately.
 noopDebouncer :: Debouncer k
--- a/ghcide/src/Development/IDE/Core/Debouncer.hs
+++ b/ghcide/src/Development/IDE/Core/Debouncer.hs
@@ -11,7 +11,6 @@ module Development.IDE.Core.Debouncer
 import           Control.Concurrent.Async
 import           Control.Concurrent.STM
 import           Control.Concurrent.STM.Stats (atomicallyNamed)
-import           Control.Exception
 import           Control.Monad                (join, void)
 import           Data.Hashable
 import           GHC.Conc                     (unsafeIOToSTM)
@@ -49,8 +48,9 @@ asyncRegisterEvent d delay k fire = join $ atomicallyNamed "debouncer - register
                 join $ atomicallyNamed "debouncer - sleep" $ do
                     (s,act) <- readTVar var
                     unsafeIOToSTM $ sleep s
-                    STM.delete k d
-                    return act
+                    return $ do
+                        atomically (STM.delete k d)
+                        act
 
 -- | Debouncer used in the DAML CLI compiler that emits events immediately.
 noopDebouncer :: Debouncer k
--- a/ghcide/src/Development/IDE/Core/Debouncer.hs
+++ b/ghcide/src/Development/IDE/Core/Debouncer.hs
@@ -34,23 +34,50 @@ newAsyncDebouncer = Debouncer . asyncRegisterEvent <$> STM.newIO
 
 -- | Register an event that will fire after the given delay if no other event
 -- for the same key gets registered until then.
-asyncRegisterEvent :: (Eq k, Hashable k) => STM.Map k (TVar (Seconds, IO())) -> Seconds -> k -> IO () -> IO ()
+asyncRegisterEvent
+    :: (Eq k, Hashable k)
+    => STM.Map k (TVar (Maybe (Seconds, IO())))
+    -> Seconds
+    -> k
+    -> IO ()
+    -> IO ()
 asyncRegisterEvent d delay k fire = join $ atomicallyNamed "debouncer - register" $ do
+    -- The previous TVar for this key, if any
     prev <- STM.lookup k d
     case prev of
-        Just v -> writeTVar v (delay, fire) >> return (pure ())
+        Just v -> do
+            current <- readTVar v
+            case current of
+                -- Not empty, means that there is a thread running the actions
+                Just _  -> writeTVar v (Just (delay, fire)) >> return (pure ())
+                -- Empty = no thread. We need to start one for running the action
+                Nothing -> writeTVar v (Just (delay, fire)) >> return (restart v)
+
+        -- No previous TVar, we need to insert one and restart a thread for running the action
         Nothing
           | delay == 0 -> return fire
           | otherwise  -> do
-            var <- newTVar (delay, fire)
+            var <- newTVar (Just (delay, fire))
             STM.insert var k d
-            return $ void $ async $
+            return (restart var)
+    where
+        -- | Restart a thread to run the action stored in the given TVar
+        --   Once the action is done, the thread dies.
+        --   Assumes the Tvar is not empty
+        restart var =
+            void $ async $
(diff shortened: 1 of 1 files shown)


ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,29 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
-        sleep delay
-        fire
-        modifyVar_ d (evaluate . Map.delete k)
-    join $ modifyVar d $ \m -> do
-        (cancel, !m') <- evaluate $ Map.alterF (\prev -> (traverse_ cancel prev, Just a)) k m
-        return (m', cancel)
+    prev <- atomically $ STM.lookup k d
+    case prev of
+        Just v -> do
+            atomicallyNamed "debouncer - reset" $ writeTVar v (delay, fire)
```
michaelpj, line 47: Should this be atomic with the transaction on line 44?

ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,29 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
+        Nothing -> do
+            var <- newTVarIO (delay, fire)
+            _ <- asyncWithUnmask $ \unmask -> unmask $ do
+                join $ atomicallyNamed "debouncer - sleep" $ do
+                    (s,act) <- readTVar var
+                    unsafeIOToSTM $ sleep s
+                    STM.delete k d
+                    return act
+            atomicallyNamed "debouncer2" $ STM.insert var k d
```
michaelpj, line 56: Should this be atomic with the transaction on line 51?

michaelpj (commented):
I have failed to understand what the code here is dong, so suggestions may be nonsense :D

ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,29 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
-    join $ modifyVar d $ \m -> do
-        (cancel, !m') <- evaluate $ Map.alterF (\prev -> (traverse_ cancel prev, Just a)) k m
-        return (m', cancel)
+    prev <- atomically $ STM.lookup k d
+    case prev of
+        Just v -> do
+            atomicallyNamed "debouncer - reset" $ writeTVar v (delay, fire)
+        Nothing -> do
+            var <- newTVarIO (delay, fire)
+            _ <- asyncWithUnmask $ \unmask -> unmask $ do
```
michaelpj, line 50: From the PR description I thought you'd got rid of the `async` part, but it seems to still be here?

pepeiborra: The debouncer still runs in a background thread. We got rid of the `cancel`

ghcide/src/Development/IDE/Core/Debouncer.hs
pepeiborra: No, it cannot be, since the transaction in line 51 is running in another thread

ghcide/src/Development/IDE/Core/Debouncer.hs
pepeiborra: Yes, we can join these two transactions and get rid of the `mask_`

ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,24 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
-    join $ modifyVar d $ \m -> do
-        (cancel, !m') <- evaluate $ Map.alterF (\prev -> (traverse_ cancel prev, Just a)) k m
-        return (m', cancel)
+asyncRegisterEvent :: (Eq k, Hashable k) => STM.Map k (TVar (Seconds, IO())) -> Seconds -> k -> IO () -> IO ()
+asyncRegisterEvent d delay k fire = join $ atomically $ do
+    prev <- STM.lookup k d
+    case prev of
+        Just v -> writeTVar v (delay, fire) >> return (pure ())
+        Nothing -> return $ do
+            var <- newTVarIO (delay, fire)
```
michaelpj, line 44: Shouldn't the creation and insertion of this TVar be in the STM transaction, rather than the IO action? Because otherwise we could have two threads hitting the path in line 43 and installing two delayed action threads. I'd have thought we want to either update the TVar or insert it, atomically.

michaelpj: Something like
```
do
   var <- newTVar (delay, fire)
   STM.insert var k d
   return $ async $ do ...
```

pepeiborra: Yes, you are right. Thanks!

ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,26 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
+    case prev of
+        Just v -> writeTVar v (delay, fire) >> return (pure ())
+        Nothing
+          | delay == 0 -> return fire
+          | otherwise  -> do
+            var <- newTVar (delay, fire)
+            STM.insert var k d
+            return $ void $ async $
+                join $ atomicallyNamed "debouncer - sleep" $ do
+                    (s,act) <- readTVar var
```
michaelpj, line 50: Do we actually need the inner TVar here? What if the async thread just did an STM map lookup to get the value? Wouldn't that have the right properties already?

ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,26 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
-    a <- asyncWithUnmask $ \unmask -> unmask $ do
-        sleep delay
-        fire
-        modifyVar_ d (evaluate . Map.delete k)
-    join $ modifyVar d $ \m -> do
-        (cancel, !m') <- evaluate $ Map.alterF (\prev -> (traverse_ cancel prev, Just a)) k m
-        return (m', cancel)
+asyncRegisterEvent :: (Eq k, Hashable k) => STM.Map k (TVar (Seconds, IO())) -> Seconds -> k -> IO () -> IO ()
+asyncRegisterEvent d delay k fire = join $ atomicallyNamed "debouncer - register" $ do
+    prev <- STM.lookup k d
```
michaelpj, line 40: I wonder if this could now be simplified with some use of the `Focus` interface.

michaelpj:
(Sorry for the repeated rounds of comments, you keep making it clearer and then I see more things :sweat_smile: )

ghcide/src/Development/IDE/Core/Debouncer.hs
pepeiborra: Not exactly, as the transaction would range over all the contents of the STM Map (or rather, all the contents of the branch that holds this key)

ghcide/src/Development/IDE/Core/Debouncer.hs
pepeiborra: I think that the result would be far more complex

ghcide/src/Development/IDE/Core/Debouncer.hs
```diff
@@ -31,28 +31,26 @@ newtype Debouncer k = Debouncer { registerEvent :: Seconds -> k -> IO () -> IO (
...
+    case prev of
+        Just v -> writeTVar v (delay, fire) >> return (pure ())
+        Nothing
+          | delay == 0 -> return fire
```
michaelpj, line 44: This is a change in semantics, right? Since this would run the action in the calling thread, whereas before it was always run in a background thread.

pepeiborra: No, the previous code has this special case as well

michaelpj: Hmm, I guess it seems a bit suspicious that `asyncRegisterEvent` sometimes runs things synchronously. Seems like it might result in us blocking the main thread when we didn't mean to.

pepeiborra: It's there to reduce nondeterminism in tests

ghcide/src/Development/IDE/Core/Debouncer.hs
michaelpj: Wait, but it already does that because it has to delete the key at the end!

ghcide/src/Development/IDE/Core/Debouncer.hs
michaelpj: Isn't that what the noop debouncer is for?

pepeiborra: My last comment was wrong. The delay is set to 0 when reporting on non null diagnostics:

https://github.com/haskell/haskell-language-server/blob/89c44bfaf75326b908548e74ec85eada6c9d3e19/ghcide/src/Development/IDE/Core/Shake.hs#L1168-L1169

And the tests do not run with delay 0 at all, so changing the test suite to use the noop debouncer is likely to result in test failures

michaelpj: IDK, it's all a bit indirect. There's a secret behaviour that if you pass 0 delay the event runs not only with no delay but synchronously. I don't really have a better suggestion, though.

And it's not actually for determinism, it's just to... get reports on diagnostics ASAP?

ghcide/src/Development/IDE/Core/Debouncer.hs
pepeiborra: That's no good. The delete needs to happen in a separate transaction. But then there is a tiny race condition window. Hmm

michaelpj: Okay, so one option would be to just accept that we're locking a bit of the map, and write some simpler code that just looks up the value in the map.

Another perhaps ugly idea: store a `TMVar` instead of a `TVar` in the map. Then the async task can:
1. In a transaction: try to look at (and empty) the `TMVar`, and then do the action
2. In a transaction: look at the `TMVar`, if it's empty, then remove it from the map (otherwise someone has put something else in there, so we want to leave it anyway).

That is, we use a `TMVar` as a one element queue, and the async thread first draws from the queue atomically and executes an action, and then optionally tries to garbage-collect the queue so as to keep the map small.

michaelpj: So then there's a race window between 1 and 2, but it's benign, because all that can happen is that someone has put something in the queue again, in which case we can just not garbage-collect it.

pepeiborra: Ok, I have gone with a `TVar (Maybe ..)` and simplified your solution a bit by never deleting from the map

michaelpj: That makes me a little nervous - isn't that a space leak?

pepeiborra: Technically yes, but a bounded one. Only in projects with infinite files would this be a worry.

michaelpj: Worth a comment.

pepeiborra:
Let's do this in steps and land the lock-less change first, extracted to #2469

pepeiborra:
Closing as I cannot measure any improvement over the current async debouncer

michaelpj:
Yeah, I guess you could argue that this version is easier to understand, though?

Issue #2451, opened 2021-12-08
Adding (\\) operator to import list adds it without surrounding it in parenthesis


jhrcek:
### Your environment

Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 1.5.1.0, Git revision dcd2163d81f4bb3228203598a761b005c54b1796 (2504 commits) x86_64 ghc-8.10.7
Current directory: /home/jhrcek/Devel/github.com/Holmusk/PI/backend
Operating system: linux
Arguments: ["--lsp"]
Cradle directory: /home/jhrcek/Devel/github.com/Holmusk/PI/backend
Cradle type: Stack

Tool versions found on the $PATH
cabal:		3.6.2.0
stack:		2.7.3
ghc:		8.10.7

### Steps to reproduce
See this gif

![ImportBug](https://user-images.githubusercontent.com/2716069/145226677-e6c65a46-9f50-4e49-956d-d5fab12dbbc1.gif)

### Expected behaviour
The `\\` operator should be surrounded in `()` after it's added to import list

### Actual behaviour
The `\\` operator is added without `()`

jneira:
Thanks for reporting the issue, i've tried to reproduce it in windows using ghc-9.0.1 and i was not able:

![imagen](https://user-images.githubusercontent.com/54035/145249525-40aeb58b-c5ae-4917-8160-5a9d47348fc9.png)

My test module looks like yours 🤔

jhrcek:
Interesting.
Seeing that the thing to add to import list is extracted from GHC error via regex,
I tried to compile the demo module with various versions of GHC and each time I got the same output, each time without `()`s
```bash
~/Tmp> ghcup set ghc 8.8.4
[ Info  ] GHC 8.8.4 successfully set as default version
~/Tmp> ghc Demo.hs 
[1 of 1] Compiling Demo             ( Demo.hs, Demo.o )

Demo.hs:3:12: error:
    • Variable not in scope: (\\) :: [Bool] -> [a0] -> t
    • Perhaps you want to add ‘\\’ to the import list in the import of
      ‘Data.List’ (Demo.hs:2:1-19).
  |
3 | x = [True] \\ []
  |            ^^
~/Tmp> ghcup set ghc 8.10.7
[ Info  ] GHC 8.10.7 successfully set as default version
~/Tmp> ghc Demo.hs 
[1 of 1] Compiling Demo             ( Demo.hs, Demo.o )

[...]

PR #2468, opened 2021-12-11
Handle re-exported modules when constructing ExportsMap

jhrcek:
Attempt at fixing https://github.com/haskell/haskell-language-server/issues/2451 based on @pepeiborra 's suggestion: https://github.com/haskell/haskell-language-server/issues/2451#issuecomment-991637640

This fixes the issue for me with GHC 8.10.7, let's see how it fares in CI..

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2468"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Types/HscEnvEq.hs
+++ b/ghcide/src/Development/IDE/Types/HscEnvEq.hs
@@ -16,7 +16,6 @@ import           Control.Concurrent.Strict       (modifyVar, newVar)
 import           Control.DeepSeq                 (force)
 import           Control.Exception               (evaluate, mask, throwIO)
 import           Control.Monad.Extra             (eitherM, join, mapMaybeM)
-import           Control.Monad.IO.Class
 import           Data.Either                     (fromRight)
 import           Data.Set                        (Set)
 import qualified Data.Set                        as Set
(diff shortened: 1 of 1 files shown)


ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -76,22 +76,25 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
...
             depends = explicitUnits pkgst
-            targets =
-                [ (pkg, mn)
+            modules =
+                [ m
                 | d        <- depends
                 , Just pkg <- [lookupPackageConfig d hscEnv]
-                , (mn, _)  <- unitExposedModules pkg
+                , (modName, maybeOtherPkgMod) <- unitExposedModules pkg
+                , m <- pure $ case maybeOtherPkgMod of
```
pepeiborra, line 84: Why not use `let` here? Something like
```suggestion
                , let m =  case maybeOtherPkgMod of
```

jhrcek: I'm not used to using list comprehensions so much, so it didn't occur to me

pepeiborra (approved):
Thanks for reporting, diagnosing and fixing this!

jhrcek:
I don't understand what the failure in the windows job it about.
What do you usually do / what's the right thing to do next?

PR #2470, opened 2021-12-11
Warn if TH and Mac and static binary

### LINKED ISSUE #2391: Display a warning when using Template Haskell and static binaries
We know that HLS is likely to crash in this situation, specially in Mac OS, so perhaps HLS should use a `WindowShowMessage` LSP notification to warn the user and link to the documentation of the recommended solution, which is to build locally with dynamic linking.

pepeiborra:
Closes #2391 

https://user-images.githubusercontent.com/26626/145691163-11d0c2b5-2dd6-4aef-9299-1d86a16caf23.mov

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2470"></a>


jneira:
This is really a good idea and the suggestion is helpful. However is not the first statement a little bit strong? I mean, i thought template haskell was working at least for *some* cases. The warning could be surprising for those who th works.

pepeiborra:
> This is really a good idea and the suggestion is helpful. However is not the first statement a little bit strong? I mean, i thought template haskell was working at least for _some_ cases. The warning could be surprising for those who th works.

For me Haskell should never segfault, it's a horrible user experience. Therefore, it doesn't work.

Secondly, the statement is strictly true: we don't support TH in Mac OS with a statically linked binary: no one is working on fixing the problems, which involve bugs in the Mach linker in GHC.

Do you have any concrete suggestions?

jneira:
I just saw it is exclusive for macos and it is not emitted for Linux. I would add a weak version for Linux as th cause problems in that os but we can wait to see what is the effect of alpine binaries

jneira (approved):

docs/troubleshooting.md
```diff
@@ -36,15 +36,11 @@ An usual symptom is the presence of errors containing `unknown symbol` and it is
...
 
-Due to how Template Haskell code is evaluated at compile time and some limitations in the interaction between HLS and GHC, the loading of modules using TH can be problematic.
+Template Haskell should work fine in Linux and Windows with the distributed binaries. In Mac Os a dynamically linked binary of HLS is required to avoid segmentation faults. The easiest way to obtain a dynamically linked HLS binary is to build it locally. With cabal install this can be done as follows:
 
-The errors thrown are usually related to linking and usually make HLS crash: `Segmentation fault`, `GHC runtime linker: fatal error`, etc
-
-A workaround which has helped in some cases is to compile HLS from source with the ghc option `-dynamic` enabled, as in the previous issue.
-
-We have a [dedicated label](https://github.com/haskell/haskell-language-server/issues?q=is%3Aissue+is%3Aopen+label%3A%22type%3A+template+haskell+related%22) in the issue tracker and an [general issue](https://github.com/haskell/haskell-language-server/issues/1431) tracking support for TH.
+    cabal update && cabal install haskell-language-server --enable-executable-dynamic
```
pepeiborra, line 43: @jneira One issue is that the command line above doesn't work for GHC 9.0 as certain plugins (class, tactics, etc.) require flags to be disabled. This goes back to #2317 I'm afraid

pepeiborra: Nevermind, I see that #2322 should fix the issue

jneira:
It seems hie-bios should be updated to 0.8.0 in stack.yaml's to get `hostIsDynamic`

PR #2474, opened 2021-12-12
Fix some pragma completion cases

Ailrun:
In some editors, parenthesis completion can handle a language specific cases. For example, in my Emacs, when I type
```Haskell
{-#
```
it automatically gives
```Haskell
{-##-}
```
However, the current pragma completion plugin does not recognize this case, and completes
```Haskell
{-# LANGUAGE #- #-}
```
which is invalid. This PR cares the editors which fill `#-}` or `-}` for Haskell.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2474"></a>


jneira (approved):
Nice to see you had the chance to work in hls, it was missing you 😄 
Thanks for the fix

Ailrun:
Yeah, sadly I'm still under tons of work for some papers, but this one and class plugin bothers me too much even for the papers :) I will also soon send a class plugin patch for GHC 9.0.

PR #2475, opened 2021-12-12
Class plugin bump up

Ailrun:
Now it supports GHC 9.0

Ref: #297

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2475"></a>


jneira (commented):
great, we are getting closer to ghc-9 full support
I would take a look to hls .cabal file, now plugins have ghc bounds in that file to allow hackage installations and ghc-9

Ailrun:
Oops, you are right, I missed the hls .cabal file. I will fix that soon.

jneira (approved):
many thanks! to merge when ci is happy

Anton-Latukha:
Looks like 1 test build was not terminating. Restarted the workflow.

Anton-Latukha:
Thank you, @Ailrun, & everybody,

Ailrun:
@Anton-Latukha
> simultaneously aborted with error return code

This part is actually because I cancelled it. It's somehow blocked indefinitely. Thank you for the restart.

PR #2476, opened 2021-12-13
add emacs configuration example

freckletonj:
I love HLS, I love wingman, and struggled to figure out how to configure it. Here's an example that I think will help emacsers along their way.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2476"></a>


jneira:
thanks for improving docs, I asked a revision from @michaelpj and @Ailrun, maintainers and Emacs users iirc

docs/configuration.md
```diff
@@ -378,6 +378,18 @@ various parts of the Emacs integration.
 In particular, `lsp-haskell` provides customization options for the `haskell-language-server`-specific parts,
 such as the path to the server executable.
 
+#### Vanilla Emacs
```
Ailrun, line 381: Is this specific for Vanilla Emacs? I think it's applicable Emacs in general

docs/configuration.md
```diff
@@ -378,6 +378,18 @@ various parts of the Emacs integration.
...
+
+If you're using `lsp-mode` generic plugin configuration can be added using `lsp--set-configuration`:
+
+```emacs-lisp
+(add-hook 'lsp-after-initialize-hook
+          '(lambda ()
+             (lsp--set-configuration
+              `(:haskell (:plugin (:tactics (:config (:timeout_duration 5)))))
```
Ailrun, line 389: Out of curiosity, why do you use backquote/quasiquote here instead of normal quote?

Ailrun (commented):
Thanks for the update! Here's my two cents.

docs/configuration.md
```diff
@@ -378,6 +378,18 @@ various parts of the Emacs integration.
 In particular, `lsp-haskell` provides customization options for the `haskell-language-server`-specific parts,
 such as the path to the server executable.
 
+#### Vanilla Emacs
+
+If you're using `lsp-mode` generic plugin configuration can be added using `lsp--set-configuration`:
```
michaelpj, line 383: I'm not really sure who the audience for this is: I don't think anyone is likely to be using `lsp-mode` without `lsp-haskell`. So this looks like it belongs in the `lsp-haskell` readme... and I think what you _really_ want is a bug report there pointing out that the `defcustom` options exposed by `lsp-haskell` are woefully out-of-date and incomplete... which I know about and just haven't fixed yet. Regardless, this is a useful workaround and worth documenting there.

docs/configuration.md
freckletonj: TBH it took me a while to find the right concoction, and, this was just some cruft that was left, it may not need it, but it's where my configuration stood when it started working!

docs/configuration.md
freckletonj: I saw documentation for doom and spacemacs, so I assumed that would be a better method for non-vanilla users, and us vanilla users would benefit from setting the config directly.

docs/configuration.md
freckletonj: I mention `lsp-mode` because HIE used to mention that you had an option between it and eglot and I wasn't sure if that was still relevant here.

So as far as audience, this pertains to anyone using vanilla emacs, and there are code snippets in the docs for how to configure HLS given different environments, just, not vanilla emacs which I feel like is a common env. Also, given that there's a stack of dependencies (HLS, lsp-mode, lsp-haskell, lsp-ui, elisp), it's difficult to know how to configure HLS. I don't use the lsp for anything but haskell, and rarely touch my elisp config, so, I haven't ever needed that knowledge, and therefore configuring HLS took me weeks of avoiding, and then 1-2 hours of solving. I agree that lsp-haskell could use similar updates, and probably lsp-mode too. 

Could this be a stopgap until lsp-haskell's defcustom + documentation gets updated?

docs/configuration.md
Ailrun: Other sections explain the installation method, which actually corresponds to the explanation above this subsection.

docs/configuration.md
Ailrun: I think `'(:haskell (:plugin (:tactics (:config (:timeout_duration 5)))))` should also work. Could you try that, and if it work, use it instead? I think, for document, it would be better to be consistent (your previous lines use `'` quote).

freckletonj: done!

docs/configuration.md
freckletonj: I think I see what you're saying now. How about this tweak? (https://github.com/haskell/haskell-language-server/blob/c6065259e294e397b285dc6bed8c7b38d3a9aefd/docs/configuration.md#emacs)

Anton-Latukha (approved):

Anton-Latukha:
Ok, 2 maintainers made reviews, all get addressed.

I'm also a user of Emacs, so with a clear heart can merge it.

docs/configuration.md
Ailrun: Looks good to me :)

docs/configuration.md
```diff
@@ -376,7 +376,15 @@ are included below.
 Make sure to check the READMEs of each of these packages, which explain how to configure the
 various parts of the Emacs integration.
 In particular, `lsp-haskell` provides customization options for the `haskell-language-server`-specific parts,
-such as the path to the server executable.
+such as the path to the server executable, or configuring the tactics plugin. For example:
```
michaelpj, line 379: This is in conflict with the previous clause, which says that `lsp-haskell` provides customization options... and then you give an example of how to bypass them! I still really think this should be in the `lsp-haskell` readme if anywhere. Better to keep the Emacs docs with the Emacs package.

Ailrun: FYI, michael means this repository: https://github.com/emacs-lsp/lsp-haskell/

freckletonj: @michaelpj I see what you're saying and personally agree that it should be in `lsp-haskell` (though, in addition to being in HLS's docs). If there's a way to configure `lsp-haskell` better than `lsp--set-configuration`, I think that would be relevant to have in HLS's docs instead of what I suggested, but, `lsp-haskell` doesn't have any documentation I could find suggesting just _what_ incantation it wants to configure variables, like, tactics configuration.

michaelpj: > I see now that lsp-haskell does mention (customize-group 'lsp-haskell), but that doesn't have the tactics config in it

Alright, there we go. That's a bug. Which I have been procrastinating on fixing for a while, but you pushed me over the edge ;) https://github.com/emacs-lsp/lsp-haskell/pull/144 .

> Could you suggest a better alternative (IE the proper way to configure lsp-haskell), or point me to documentation?

Could you rather help me figure out what route you took in looking for information so we can make the right route clearer? Evidently you didn't find the bit in the `lsp-haskell` readme that you needed and I'd like to understand why!

The route I was hoping people would take is:
- HLS configuration doc
- `lsp-haskell` readme (since the configuration doc tells you to go there)
- "Emacs configuration" section of `lsp-haskell` readme, which talks about the customization options

> I think that would be relevant to have in HLS's docs instead of (or in addition to) what I suggested.

Perhaps I'm being overly dogmatic, but I would really prefer to keep the Emacs documentation next to the Emacs package, we're much more likely to keep it up-to-date that way!

-----

Also, I would definitely appreciate a PR to the `lsp-haskell` readme with your workaround for configuring settings that aren't exposed. That's a very useful trick for when we haven't got an option for something yet!

freckletonj: understood, I'll close this PR then, and offer this one in its place: https://github.com/emacs-lsp/lsp-haskell/pull/146

> what route you took in looking for information 

My expectations were to find these answers first in HLS's README, second in HLS's docs, third in... probably lsp-mode was third, fourth was lsp-haskell, I missed the `customize-group` comment in lsp-haskell's readme (it _was_ in the configuration heading, but a lot of that heading was just installation pointers so I just missed it toward the end, and it wouldn't have helped since it had fallen out of parity with HLS (thanks for your new pr btw!)), so, I probably poked at lsp-ui next, then I hunted through github issues of various projects via search engine until I stumbled across someone configuring a python project using the same technique I show here, and extrapolated that the proper configuration key paths would align with those in the docs, which, still took some fiddling to get right.

I'm super grateful for all you guys have done though, so, massive thanks and sorry I submitted this PR in the wrong place.

jneira: @freckeltonj Don't worry and thanks again for the patch wherever is finally merged
And for explain your path looking for info, that pov is invaluable for helping improve docs

Anton-Latukha: @freckletonj, I know how that is. Docs are indeed frequently hard to merge. Besides Wadler's Law, docs workflow essentially should be built on the Cunningham Wiki principles, but projects frequently use code review workflow for them. (I in no way appeal "to authority" there, its mine conclusions in concise public-available terms).

freckletonj: I love the wiki idea. Actually, my favorite all-time documentation is kinda similar, have you ever seen [ClojureDocs](https://clojuredocs.org/)? I seriously think haskell would benefit from this format.

Anton-Latukha: Heard good stuff about ClosureDocs.

Essentially made alike Haskell docs for myself to learn the language.

Something like ClosureDocs may start/be tried in near time. Especially after the new base. But HaskellWiki is already there & so as GHC wiki. Back in the day there was an idea/try to migrate Hackage package docs into HaskellWiki, so every package had an article there. As put "a wiki that needs to invent its own HTML for markup is a failed design", which is one of the things why people are not enthusiastic about writing in MediaWikis & HaskellWiki also. & Hoogle is there. So basically seems like if a couple of people would not spearhead the agenda for several years & manage to connect enough pieces together - likely the situation would stay a cluster ***.

Haskell does not have good basic tutorial open learning materials for newcomers.

But overall I consider the situation not ideal, but not that bad either, it is bearable.

---

It is easier to give users the ability to change docs. & sit & review changes done by users in the last week/month (even team-sporadically/chaotically) & improve changes even more, than for 4 people to review 1 paragraph of documentation.

Devs anyway never (to be safe - almost never) change manual docs in the same PR they change code, even because code review changes would change the doc updates.

---

Just to be clear, the power of wikis I mention, is in these:

Screenshot-2021-12-15-00:36:17

That is what I mean by "cornerstone wiki design principles".

All ideas basically converge on having an easily editable community Zettelkasten. MediaWiki (Wikipedia-related wikis) are a bad example due to formatting. All wikis also should be interlinked & preview articles of one another. MediaWikis have those integration abilities.

I've seen several contemporary innovative approaches to docs - they all converge to have community Zettelkastens. Even back in the day, Oxford English dictionary was created so.

PR #2482, opened 2021-12-13
Update flake to use fourmolu plugin in GHC 9

Ailrun:
Ref: #297

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2482"></a>


Ailrun:
I also have tried to partially enable GHC9.2 for nix but `ptr-poker` block the build, unfortunately

jneira (approved):
good ctach, thanks!

PR #2483, opened 2021-12-13
Fix redundant import actions for names starting with _

Ailrun:
Names starting with `_` is often used, for example, with `lens` among many others. However, the current redundant import action does not remove those even when they are redundant.

For instance,
```haskell
import Control.Lens ((^.), _1, _2)

myVal = (1, 2) ^. _1
```
Here, `_2` is redundant, but GHCIDE does not give any actions. Adding one more redundant import like
```haskell
import Control.Lens ((^.), _1, _2, use)
```
gives the action, but the action removes only `use`, not `_2`.

This PR resolves this issue.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2483"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -24,6 +24,9 @@ import           Control.Arrow                                     (second,
 import           Control.Concurrent.STM.Stats                      (atomically)
 import           Control.Monad                                     (guard, join)
 import           Control.Monad.IO.Class
+#if !MIN_VERSION_ghc(9,0,0)
+import           Ctype                                             (is_ident)
+#endif
 import           Data.Char
 import qualified Data.DList                                        as DL
 import           Data.Function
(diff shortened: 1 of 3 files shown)


ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -24,6 +24,9 @@ import           Control.Arrow                                     (second,
 import           Control.Concurrent.STM.Stats                      (atomically)
 import           Control.Monad                                     (guard, join)
 import           Control.Monad.IO.Class
+#if !MIN_VERSION_ghc(9,0,0)
+import           Ctype                                             (is_ident)
+#endif
```
pepeiborra, line 29: Please move all the CPP to the compatibility layer at `Development.IDE.Compat.GHC`

Ailrun: Thank you for the pointing that out. Fixed.

pepeiborra (approved):
Awesome, thanks for the fix!

PR #2484, opened 2021-12-13
Send unhandled exceptions to the user

### LINKED ISSUE #2390: Report error conditions via LSP message dialogs
Fatal errors that do not show up in diagnostics should be handled an LSP `WindowShowMessage` including a short description of the error together with a link to the HLS docs and issue tracker. 

The poster example would be GHC<->HLS version mismatches.

pepeiborra:
Closes #2390

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2484"></a>

### DIFF
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -142,18 +142,23 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             let handleServerException (Left e) = do
                     logError logger $
                         T.pack $ "Fatal error in server thread: " <> show e
+                    sendErrorMessage e
                     exitClientMsg
                 handleServerException _ = pure ()
 
-                uncaughtError (e :: SomeException) = do
-                    logError logger $ T.pack $
-                        "Unexpected exception on notification, please report!\n" ++
-                        "Exception: " ++ show e
+                sendErrorMessage (e :: SomeException) = do
                     LSP.runLspT env $ LSP.sendNotification SWindowShowMessage $
                         ShowMessageParams MtError $ T.unlines
-                        [ "Unhandled error, please [report](" <> issueTrackerUrl <> "): "
+                        [ "Unhandled exception, please [report](" <> issueTrackerUrl <> "): "
                         , T.pack(show e)
                         ]
+
+                exceptionInHandler e = do
+                    logError logger $ T.pack $
+                        "Unexpected exception, please report!\n" ++
+                        "Exception: " ++ show e
+                    sendErrorMessage e
+
                 logger = ideLogger ide
 
                 checkCancelled _id act k =
@@ -170,7 +175,7 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
                                     k $ ResponseError RequestCancelled "" Nothing
                                 Right res -> pure res
                         ) $ \(e :: SomeException) -> do
-                            uncaughtError e
+                            exceptionInHandler e
                             k $ ResponseError InternalError (T.pack $ show e) Nothing
             _ <- flip forkFinally handleServerException $ runWithDb logger dbLoc $ \hiedb hieChan -> do
               putMVar dbMVar (hiedb,hieChan)
@@ -179,7 +184,7 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
                 -- We dispatch notifications synchronously and requests asynchronously
                 -- This is to ensure that all file edits and config changes are applied before a request is handled
                 case msg of
-                    ReactorNotification act -> handle uncaughtError act
+                    ReactorNotification act -> handle exceptionInHandler act
                     ReactorRequest _id act k -> void $ async $ checkCancelled _id act k
             pure $ Right (env,ide)
 
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -40,6 +40,8 @@ import           Development.IDE.Types.Logger
 
 import           System.IO.Unsafe                      (unsafeInterleaveIO)
 
+issueTrackerUrl = "https://github.com/haskell/haskell-language-server/issues"
+
 runLanguageServer
     :: forall config. (Show config)
     => LSP.Options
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -137,48 +139,49 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
...
                     exitClientMsg
                 handleServerException _ = pure ()
+
+                uncaughtError (e :: SomeException) = do
+                    logError logger $ T.pack $
+                        "Unexpected exception on notification, please report!\n" ++
+                        "Exception: " ++ show e
+                    LSP.runLspT env $ LSP.sendNotification SWindowShowMessage $
+                        ShowMessageParams MtError $ T.unlines
+                        [ "Unhandled error, please [report](" <> issueTrackerUrl <> "): "
```
michaelpj, line 153: I don't believe this is supposed to support markup, it's not a `MarkedString`.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -137,48 +139,49 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
...
             let handleServerException (Left e) = do
-                    logError (ideLogger ide) $
+                    logError logger $
                         T.pack $ "Fatal error in server thread: " <> show e
                     exitClientMsg
                 handleServerException _ = pure ()
+
+                uncaughtError (e :: SomeException) = do
+                    logError logger $ T.pack $
+                        "Unexpected exception on notification, please report!\n" ++
```
michaelpj, line 149:
```suggestion
                        "Unexpected exception or notification, please report!\n" ++
```

michaelpj: Generally, is there a reason why the log message and the window message are different here?

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -137,48 +139,49 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             registerIdeConfiguration (shakeExtras ide) initConfig
 
             let handleServerException (Left e) = do
-                    logError (ideLogger ide) $
+                    logError logger $
```
michaelpj, line 142: Should we be sending a message here too?

michaelpj (approved):

ghcide/src/Development/IDE/LSP/LanguageServer.hs
pepeiborra: VSCode supports some basic markup, see https://github.com/microsoft/vscode/issues/20595#issuecomment-281099832
I expect other clients to support it, and if not, everyone can read markdown :)

ghcide/src/Development/IDE/LSP/LanguageServer.hs
pepeiborra: This function is called from multiple places, not just the notification handler.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
pepeiborra: No, since that will result in double notifications being sent, notice the `finally` in `checkCancelled`. 

I'll take the chance to clean it up

pepeiborra: Nevermind, `checkCancelled` doesn't propagate exceptions, so we should be sending a message here too indeed

ghcide/src/Development/IDE/LSP/LanguageServer.hs
michaelpj: Ugh, this stupid poorly-specified protocol :roll_eyes:

michaelpj (approved):

jneira:
The test suite for tactics has failed twice for windows 8.10.7, with different test error cases, not sur if it will be transient

Anton-Latukha:
Recent failed test builds were:
`pepeiborra` - https://github.com/haskell/haskell-language-server/runs/4513326877?check_suite_focus=true
`AntonLatukha` - https://github.com/haskell/haskell-language-server/runs/4527885363?check_suite_focus=true
`jneira` - https://github.com/haskell/haskell-language-server/runs/4530467808?check_suite_focus=true

pepeiborra:
The last failure is due to the old sqlite3 hiedb problem, not something specific to this change. We have a ticket for it somewhere, but I am on my phone right now and can't look. 

We have a couple of sources of flakiness, and I would appreciate if someone would find the time to look into the hiedb one. Why is it getting triggered? Is the test suite   running tests in parallel?

```
SQLite3 returned ErrorBusy while attempting to perform step: database is locked\n",

```

pepeiborra:
Added code to stop the hiedb on shutdown. Might or might not help

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -57,11 +58,16 @@ runLanguageServer
...
-    -- LSP loop will be canceled when it's full.
+    -- LSP server will be canceled when it's full.
     clientMsgVar <- newEmptyMVar
     -- Forcefully exit
     let exit = void $ tryPutMVar clientMsgVar ()
 
+    -- An MVar to control the lifetime of the reactor loop.
+    -- The loop will be stopped and resources freed when it's full
+    reactorLifetime <- newEmptyMVar
+    let stopReactorLoop = void $ tryPutMVar reactorLifetime ()
```
michaelpj, line 69: Why `tryPutMVar`? So it doesn't block? Shouldn't we at least return the bool and log something in the shutdown handler if it fails to shutdown the reactor?

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -57,11 +58,16 @@ runLanguageServer
 runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChange userHandlers getIdeState = do
 
     -- This MVar becomes full when the server thread exits or we receive exit message from client.
-    -- LSP loop will be canceled when it's full.
+    -- LSP server will be canceled when it's full.
     clientMsgVar <- newEmptyMVar
     -- Forcefully exit
     let exit = void $ tryPutMVar clientMsgVar ()
```
michaelpj, line 64: and tbh ditto here

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -190,15 +194,22 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
             pure $ Right (env,ide)
 
 
+-- | Runs the action until it ends or until the given MVar is put.
+--   Rethrows any exceptions.
+untilMVar :: MonadUnliftIO m => MVar () -> m () -> m ()
```
michaelpj, line 199: Kind of surprised these sorts of utilities aren't available elsewhere. I couldn't find them, anyway.

michaelpj: Especially the pattern of forking a thread with an MVar to control terminating it :thinking:

michaelpj (approved):

ghcide/src/Development/IDE/LSP/LanguageServer.hs
pepeiborra: Why? So that it can be called twice (not that we depend on that)
Not sure which bool you have in mind.
The reactor is guaranteed to be killed, and it has an exception handler attached (which hopefully should only capture non-async exceptions)

ghcide/src/Development/IDE/LSP/LanguageServer.hs
pepeiborra: They may exist, I'm not sure either.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
pepeiborra: We already log a "Received shutdown message", that's good enough for me

michaelpj:
The windows jobs seemed to be hanging (stuck for 4h). I attempted to cancel them but it's unclear that it's worked.

michaelpj:
... how did that work?

jneira:
> ... how did that work?

it should not 😕
the test post-jobs was supposed to be cancelled but it was not
https://github.com/haskell/haskell-language-server/runs/4549724494?check_suite_focus=true

michaelpj:
I _think_ this might be making the ghcide tests hang on windows. I think we should revert it until we can confirm.

jneira:
> I _think_ this might be making the ghcide tests hang on windows. I think we should revert it until we can confirm.

well i have a pr fixing the cancelling thing and including this pr, if its tests are not stuck it would be fine

jneira:
pr in my fork with this reverted https://github.com/jneira/haskell-language-server/pull/56

jneira:
Ok the pr in my fork with this reverted has passed the ghcide tests: https://github.com/jneira/haskell-language-server/runs/4559275575?check_suite_focus=true
I am gonna open here

PR #2486, opened 2021-12-13
Correctly handle LSP shutdown/exit

pepeiborra:
The LSP spec says:

> The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit. Clients must not send any notifications other than exit or requests to a server to which they have sent a shutdown request. Clients should also wait with sending the exit notification until they have received a response from the shutdown request.

We are doing the shut down in the `Exit` notification handler, which is too late. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2486"></a>

### DIFF
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -93,6 +93,7 @@ runLanguageServer options inH outH getHieDbLoc defaultConfig onConfigurationChan
           [ ideHandlers
           , cancelHandler cancelRequest
           , exitHandler exit
+          , shutdownHandler
           ]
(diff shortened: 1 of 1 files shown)


michaelpj (approved):
I have no idea if this is visible, presumably it's not easy to test that we do something sensible here.

pepeiborra:
Before this change, the Shake profile for the last build when a test times out (using `GHCIDE_BUILD_PROFILING`) was a 0 bytes file because `lsp-test` kills the process shortly after sending the Exit notification. With this change, the last build profile is written correctly.

Issue #2480, opened 2021-12-13
Nix build is unstable; Should we keep it?


Anton-Latukha:
Recently `master` & so all PRs started to:
https://github.com/haskell/haskell-language-server/runs/4509561080?check_suite_focus=true

```
> nix build --print-build-logs
...
ghcide> Setup: Encountered missing or private dependencies:
ghcide> hie-bios >=0.8 && <0.9.0
...
error: 1 dependencies of derivation '/nix/store/ckcngbdahcmykc05rljj0cpkqvjicslb-haskell-language-server-1.5.1.0.drv' failed to build
Error: Process completed with exit code 1.
```

jneira:
ouch, the nix build seems to be fragile lately, do we know if some pr broke it? maybe we should trigger the nix build in all prs and not only in master?

//cc @maralorn @michaelpj

Anton-Latukha:
Yep, it is: https://github.com/haskell/haskell-language-server/pull/2470/files it raised the `hie-bios` bound to `0.8` & Nixpkgs current & that is now pinned use `0.7.1`.

That is a nice example why Nix builds should run on PRs.

Nixpkgs (especially sure of `haskellPackages`) has a different versioning solving system (no version solving system) than Cabal & Stack, that is why these can't be avoided.

The downside is - putting checks to the PRs would now suddenly put a requirement of Nix code maintenance on the unaware Haskell code writing devs. & Nix loving Haskell devs frequently can use Nix dev env - so their PRs in frequently are Nix-aware already. & people which do not know Nix - would not be able to solve Nix things without maintainer support.

& then the solution starts from that position. As I see it - both variants under evaluation are basically the same value but with different side effects. Which is decided - is not for me to decide.

michaelpj:
Well, or it could run on changes that touch cabal files only.

Honestly, I'm not convinced that the nix build as it stands is fit for purpose. HLS is a fast-moving project that frequently uses bleeding-edge versions of libraries that have just been released to Hackage. The nixpkgs Haskell setup is based on infrequently updated Hackage snapshots, and is more suited for building stable software.

As it is, to fix things like this would require some tedious patching in of the new versions of libraries every time someone tries to use a newly released library. Not a good UX for contributors.

(I don't even use the fancy Nix setup to work on HLS! I just work in a `nix-shell` with `ghc` and `cabal` and it's fine.)

Anton-Latukha:
[...]

PR #2488, opened 2021-12-14
CI: Nix: make Nix dev env passing optional

Anton-Latukha:
This makes Nix dev env configuration in HLS development a non-blocking process.

Now CI would simply report the status of the build, but wouldn't block the workflow/merges.

This allows both HLS to be developed & for Nix dev env to be maintained at its own pase.

---

*derived from #2480*

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2488"></a>

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -35,6 +35,7 @@ jobs:
   develop:
     if: needs.pre_job.outputs.should_skip_develop != 'true'
     needs: pre_job
+    continue-on-error: true
     runs-on: ${{ matrix.os }}
 
     strategy:
@@ -68,6 +69,7 @@ jobs:
   build:
     needs: pre_job
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
     env:
       HAS_TOKEN: ${{ secrets.HLS_CACHIX_AUTH_TOKEN != '' }}
     if: (needs.pre_job.outputs.should_skip_build != 'true' && needs.pre_job.outputs.should_skip_pr != 'true') || (github.repository_owner == 'haskell' && github.ref == 'refs/heads/master')


jneira (commented):
well, I think the problematic nix builds are only triggered for master (and it does not block the pr) and if the pr is touching nix config files, thanks to skip-action rules about file globs triggering the job

And I think failing in the changing nix files case should be kept.

Anton-Latukha:
Well, then yesterday the skip rule seemed to skip the build in the https://github.com/haskell/haskell-language-server/pull/2470.

& then master branch failed to build.

I looked through reviews & the doc in #2476, everything was addressed & change was minor. Decided that it is good enough to be done with, so marked green & set `merge me`. Then CI failed. I thought "what the heck, build failure on minor doc update" & so I noticed that red status of build on `master` caused the same build failure in #2476, so opened #2480 with "& all PRs".

So from what I've seen, skipping a workflow means that PR does not need to check it. But when after that `master` runs the workflows & fails - the fail makes PRs run the same check, or something like that.

I'm solid that was the case. Because CI failed with that error & when I was doing a report & stuff only after that PR got blocked in review 8)

Ailrun:
@Anton-Latukha I think it's more like a (arguable) bug of `skip-duplicate-actions`. The "non-skip" behaviour happened probably because of merge commit, which "modifies" other files than the document files. I'm not sure whether skipping the check in that case always makes sense or not.

Anton-Latukha:
Well, it is also a skipping action bug. It is both at the same time.

I just generally found allowing the build to report error & PRs still be mergeable - a good soft management approach.

But also essentially containing Nix checks to PRs that do changes to Nix configuration & to the main branch - is almost the same, but better :laughing:

Anton-Latukha:
Ok then.

jneira:
Yeah take a look to that pr jobs:

https://github.com/haskell/haskell-language-server/actions/workflows/nix.yml?query=branch%3Aadd-emacs-config-example

The three first commits skipped correctly the nix build, but the merge commit with master (42a9d59288ae180c6a02fe59ede8958fc78092c7) included commits with changes to .nix files:

From the pre job of https://github.com/haskell/haskell-language-server/runs/4510223222?check_suite_focus=true:

```
Run fkirc/skip-duplicate-actions@v3.4.1
Stop backtracking at commit https://github.com/haskell/haskell-language-server/commit/42a9d59288ae180c6a02fe59ede8958fc78092c7 because '.github/workflows/test.yml,cabal-ghc901.project,cabal-ghc921.project,cabal.project,configuration-ghc-901.nix,docs/troubleshooting.md,flake.nix,ghcide/ghcide.cabal,ghcide/src/Development/IDE/Core/Compile.hs,ghcide/src/Development/IDE/Core/ProgressReporting.hs,ghcide/src/Development/IDE/Core/Rules.hs,ghcide/src/Development/IDE/Core/Shake.hs,ghcide/test/exe/Progress.hs,haskell-language-server.cabal,plugins/hls-class-plugin/src/Ide/Plugin/Class.hs,stack-8.10.6.yaml,stack-8.10.7.yaml,stack-8.6.5.yaml,stack-8.8.4.yaml,stack-9.0.1.yaml,stack.yaml' are not skippable against paths '**.nix' or paths_ignore ''
Do not skip execution because we did not find a transferable run
```

And it triggered the nix build incorrectly cause the pr itself did not change any of those files. So i think doing a rebase instead of merge would not triggered it.

I would prefer continue being as strict as possible without bother contributors. If the nix build is broken on master we can rebase master onto the pr branch before merging the pr itslef, or even merge the pr forcily if the unique failing job is the nix build and it is already failing in master.
If doing those exceptions becomes too frequent and nix build fixes are excesively delayed we could add this config changes at that point.

PR #2493, opened 2021-12-16
Fail if main or pre jobs are cancelled

jneira:
* To hopefully ensure the job does no terminate succesfully before the cancellation is effective
* This workflow was marked as succesful although the cancel action was succesful: https://github.com/haskell/haskell-language-server/runs/4549724494?check_suite_focus=true
* It seems the cancel request is asynchronous and here https://github.community/t/is-there-a-way-to-cancel-a-running-workflow-within-one-of-its-jobs/17493/4 uses this trick to ensure the cancel action is effective

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2493"></a>

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -206,3 +206,5 @@ jobs:
       - if: contains(needs.*.result, 'cancelled')
         name: cancelling
         uses: andymckay/cancel-action@0.2
+      - name: Waiting for cancellation
+        run: sleep 60
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -203,8 +203,5 @@ jobs:
           echo "jobs info: ${{ toJSON(needs) }}"
       - if: contains(needs.*.result, 'failure')
         run: exit 1
-      - if: contains(needs.*.result, 'cancelled')
-        name: cancelling
-        uses: andymckay/cancel-action@0.2
-      - name: Waiting for cancellation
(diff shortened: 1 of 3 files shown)


jneira:
I am not convinced with the sleep, we dont know how time could take the completion of the cancellation so that could happen again.
Otoh there had been several cancellations and this is the first failing one so it seems the delay is very short

jneira:
sleep did not work:
* https://github.com/haskell/haskell-language-server/actions/runs/1591128009
* https://github.com/haskell/haskell-language-server/actions/runs/1591128008
 :worried:

jneira:
logic changed to fail if the main or pre job is cancelled and it was not due to skip-duplicate-actions

jneira:
I've checked in my fork a pr changing only docs is cancelling the main job but marking post job as succesful: https://github.com/jneira/haskell-language-server/actions/runs/1591406645

michaelpj (approved):

michaelpj:
I'm concerned that the windows jobs seem to be stuck at over an hour testing again. Could Pepe's changes have actually made them hang?

jneira:
> I'm concerned that the windows jobs seem to be stuck at over an hour testing again. Could Pepe's changes have actually made them hang?

ops, i did not read this, yeah might be, i am gonna try it twice to be sure

jneira:
Abusing admin perms to merge this asap

Issue #2490, opened 2021-12-15
100% CPU usage with AlternateNumberFormat plugin


Philonous:
After upgrading haskell-language-server to b7949aa5c54d8319029cc2bfba164ce5f6eab924 I noticed that it used 100% CPU while staying otherwise responsive. I found this in the logs

```
2021-12-14 13:33:11.542240523 [ThreadId 8807] INFO hls:	finish: AlternateNumberFormat.CollectLiterals (took 1m32s)
2021-12-14 13:33:11.542255151 [ThreadId 975] DEBUG hls:	"Literals: [IntLiteral SrcSpanOneLine \"/home/phil/projects/nejla/app/backend/src/Handlers/Job.hs\" 173 27 34 \"1000000\" 1000000,IntLiteral SrcSpanOneLine \"/home/phil/projects/nejla/app/backend/src/Handlers/Job.hs\" 259 13 14 \"0\" 0,IntLiteral SrcSpanOneLine \"/home/phil/projects/nejla/app/backend/src/Handlers/Job.hs\" 260 55 56 \"1\" 1]"
```

* The mentioned line is just a `threadDelay 1000000` and haskell-language-server seems to produce one of these log lines every few minutes for a different literal in my code base.
* It's a 350 line module in a medium-sized project (14kloc) 
* Light use of Template Haskell and QuasiQuotes (some `$logDebug`s and quotes for string interpolation)
* No NumericalUnderscores

Unfortunately I can't link the code since it's proprietary.

### Your environment

Which OS do you use:
NixOS

Which lsp-client do you use:
emacs with lsp-mode
Describe your project (alternative: link to the project):
stack.yaml

### Steps to reproduce
Run haskell-language-server on my project as normal

### Expected behaviour
0% CPU usage after initial setup

### Actual behaviour
100% CPU usage permanently (or at least multiple hours)

### Include debug information

I don't have access to the machine right now but I'll reproduce them if required.

Execute in the root of your project the command `haskell-language-server-wrapper --debug .` and paste the logs here:
(if you are using the vscode extension check the executable location [here](https://github.com/haskell/vscode-haskell#downloaded-binaries))
<details>
<summary>
Debug output:
</summary>

```
<paste your logs here>
```
</details>

Paste the logs from the lsp-client, you can check instructions about for VS Code [here](https://github.com/haskell/vscode-haskell#troubleshooting)

<details>
<summary>
LSP logs:
</summary>

```
<paste your logs here>
```
</details>

[...]

PR #2498, opened 2021-12-17
Disable alternate numbers format plugin temporary

jneira:
* Due to the performance issue reported here: https://github.com/haskell/haskell-language-server/issues/2490
* The plugin could be included in the build forcing the manual flag: `cabal install -falternateNumberFormat`

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2498"></a>

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -17,7 +17,7 @@ You can watch demos for some of these features [below](#demos).
 - [Module name suggestions](#module-names) for insertion or correction
 - [Call hierarchy support](#call-hierarchy)
 - [Qualify names from an import declaration](#qualify-imported-names) in your code
-- [Suggest alternate numeric formats](#alternate-number-formatting)
+- [Suggest alternate numeric formats](#alternate-number-formatting). This plugin is not included by default yet.
 
 ## Demos
 
--- a/docs/features.md
+++ b/docs/features.md
@@ -17,7 +17,7 @@ You can watch demos for some of these features [below](#demos).
 - [Module name suggestions](#module-names) for insertion or correction
 - [Call hierarchy support](#call-hierarchy)
 - [Qualify names from an import declaration](#qualify-imported-names) in your code
-- [Suggest alternate numeric formats](#alternate-number-formatting). This plugin is not included by default yet.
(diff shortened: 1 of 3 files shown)


.github/workflows/hackage.yml
```diff
@@ -25,7 +25,8 @@ jobs:
                   "hls-module-name-plugin", "hls-pragmas-plugin",
                   "hls-refine-imports-plugin", "hls-retrie-plugin",
                   "hls-splice-plugin", "hls-tactics-plugin",
-                  "hls-call-hierarchy-plugin", "hls-alternate-number-format-plugin",
+                  "hls-call-hierarchy-plugin",
+                  # "hls-alternate-number-format-plugin",
```
michaelpj, line 29: Turning it off as a build dep shouldn't stop us uploading to hackage, though?

jneira: hmm not for the default build but it would not allow to do a pure hackage build forcing the flag to include the plugin

not sure if it worths upload the package to hackage though

docs/features.md
```diff
@@ -17,7 +17,7 @@ You can watch demos for some of these features [below](#demos).
 - [Module name suggestions](#module-names) for insertion or correction
 - [Call hierarchy support](#call-hierarchy)
 - [Qualify names from an import declaration](#qualify-imported-names) in your code
-- [Suggest alternate numeric formats](#alternate-number-formatting)
+- [Suggest alternate numeric formats](#alternate-number-formatting). This plugin is not included by default yet.
```
Ailrun, line 20: Maybe it's worth to mention the issue as a reason of non-defaulting?

Ailrun (approved):
LGTM other than the hackage issue mentioned above. Here's some minor suggestion:

drsooch (approved):
Not sure if I actually need to approve. But I will anyways.... lgtm

docs/features.md
jneira: Good idea, done

jneira:
> Not sure if I actually need to approve. But I will anyways.... lgtm

no needed technically but needed anyways, thanks!

.github/workflows/hackage.yml
jneira: finally i included it in the hackage build, we have to be strict and avoid possible build errors

PR #2500, opened 2021-12-17
Automatically read in the doc version from the cabal file

michaelpj:
It was out of date. This way it will always be read in from the file, so
we can forget about it.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2500"></a>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -179,10 +179,10 @@
 
         docs = pkgs.stdenv.mkDerivation {
           name = "hls-docs";
-          src = pkgs.lib.sourceFilesBySuffices ./docs [ ".py" ".rst" ".md" ".png" ".gif" ".svg" ];
+          src = pkgs.lib.sourceFilesBySuffices ./. [ ".py" ".rst" ".md" ".png" ".gif" ".svg" ".cabal" ];
           buildInputs = [ pythonWithPackages ];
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -179,10 +179,10 @@
 
         docs = pkgs.stdenv.mkDerivation {
           name = "hls-docs";
-          src = pkgs.lib.sourceFilesBySuffices ./docs [ ".py" ".rst" ".md" ".png" ".gif" ".svg" ];
+          src = pkgs.lib.sourceFilesBySuffices ./. [ ".py" ".rst" ".md" ".png" ".gif" ".svg" ".cabal" ];
```
Ailrun, line 182: I think it might be safer to use regex so that it does not include markdown files and image files in each plugin/ghcide/...? But I haven't been generated the docs so don't know how long it takes in general.

Ailrun (approved):
Awesome, thanks!

jneira (approved):
lgtm

flake.nix
michaelpj: It doesn't matter in practice, this just affects what's hanging around in the nix build, it doesn't get processed. And we're not using this for production usage anyway, it's just for convenience.

PR #2501, opened 2021-12-17
small stack yaml updates to ease maintenance

simonmichael:
As discussed in chat.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2501"></a>

### DIFF
--- a/install/stack.yaml
+++ b/install/stack.yaml
@@ -1,13 +1,14 @@
-resolver: lts-14.27 # Last 8.6.5 
+# keep synced with ../stack.yaml to minimise rebuilding
+resolver: lts-18.18
 
 packages:
   - .
 
-extra-deps:
-  - cabal-install-parsers-0.3.0.1
-  - Cabal-3.2.0.0
-  - binary-instances-1.0.0.1
-  - lukko-0.1.1.2
+# extra-deps:
+#   - cabal-install-parsers-0.3.0.1
+#   - Cabal-3.2.0.0
+#   - binary-instances-1.0.0.1
+#   - lukko-0.1.1.2
 
 flags:
   hls-install:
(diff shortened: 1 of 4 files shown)


install/stack.yaml
```diff
@@ -1,13 +1,14 @@
...
 
 packages:
   - .
 
-extra-deps:
-  - cabal-install-parsers-0.3.0.1
-  - Cabal-3.2.0.0
-  - binary-instances-1.0.0.1
-  - lukko-0.1.1.2
+# extra-deps:
```
jneira, line 7: that should be uncommented i guess

jneira (commented):
minor but useful changes, specially comments, thanks!

install/stack.yaml
simonmichael: Replying again: 

- `extra-deps:` is commented because it can't be empty, and left in place because it's often needed in future
- I left the individual extra deps there also commented, for now, in case they are needed after all in the near future. Can remove them if you prefer.

jneira: but lukko f.e. was not in stackage when we did that stack.yaml, does it work without extra-deps?

Anton-Latukha:
Great work. Thank you @simonmichael.

Anton-Latukha (approved):

install/stack.yaml
Anton-Latukha: In the past I had to go to Stackage team & requested a `hnix-store-core` package removal. It managed to be submitted by someone into Nightly & from there straight into LTS. In one merge window.

I would thunk the continuation, because knowledge is obtained - it is possible to leap from Hackage to Stackage LTS in 1 Stackage LTS merge window.

So: https://www.stackage.org/package/lukko

Also CircleCi is happy with this PR.

---

Seems resolved.

jneira: I am afraid that circleci is not testing the install script so I am not sure at 100% it is working, will try myself if the author does not confirm it before

PR #2502, opened 2021-12-17
Ignore stack.yamls in test cabal workflow

jneira:


Anton-Latukha (approved):
Reasonable to do.

Ailrun (approved):
Awesome, thanks :)

PR #2497, opened 2021-12-17
Revert "Send unhandled exceptions to the user (#2484)"

jneira:
* This reverts commit 0211f75c96402dda53be69473714134244763d95.
* That commit causes ghcide tests in windows hangs for all ghcs
* The test suite was marked incorrectly as ok due a bug in the workflows (fixed with #2493)
* See #2484 for more details

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2497"></a>

PR #2504, opened 2021-12-18
Unhandled exceptions fixed

pepeiborra:
Revert of #2497 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2504"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -150,9 +150,6 @@ import           Ide.Types                                    (DynFlagsModificat
                                                                PluginId)
 import Control.Concurrent.STM.Stats (atomically)
 import Language.LSP.Server (LspT)
-import System.Environment (getExecutablePath)
-import System.Process.Extra (readProcessWithExitCode)
-import Text.Read (readMaybe)
 import System.Info.Extra (isMac)
 import HIE.Bios.Ghc.Gap (hostIsDynamic)
 
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -154,7 +154,7 @@ import           Data.Aeson                             (toJSON)
 import qualified Data.ByteString.Char8                  as BS8
 import           Data.Coerce                            (coerce)
 import           Data.Default
-import           Data.Foldable                          (toList)
+import           Data.Foldable                          (for_, toList)
 import           Data.HashSet                           (HashSet)
 import qualified Data.HashSet                           as HSet
 import           Data.String                            (fromString)
@@ -589,10 +589,11 @@ shakeSessionInit ide@IdeState{..} = do
     logDebug (ideLogger ide) "Shake session initialized"
 
 shakeShut :: IdeState -> IO ()
-shakeShut IdeState{..} = withMVar shakeSession $ \runner -> do
+shakeShut IdeState{..} = do
+    runner <- tryReadMVar shakeSession
     -- Shake gets unhappy if you try to close when there is a running
     -- request so we first abort that.
-    void $ cancelShakeSession runner
+    for_ runner cancelShakeSession
     void $ shakeDatabaseProfile shakeDb
     shakeClose
     progressStop $ progress shakeExtras
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -583,9 +583,10 @@ startTelemetry db extras@ShakeExtras{..}
 
 -- | Must be called in the 'Initialized' handler and only once
 shakeSessionInit :: IdeState -> IO ()
-shakeSessionInit IdeState{..} = do
+shakeSessionInit ide@IdeState{..} = do
     initSession <- newSession shakeExtras shakeDb [] "shakeSessionInit"
     putMVar shakeSession initSession
+    logDebug (ideLogger ide) "Shake session initialized"
 
 shakeShut :: IdeState -> IO ()
 shakeShut IdeState{..} = withMVar shakeSession $ \runner -> do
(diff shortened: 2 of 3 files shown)


jneira:
I hope 65fa86c could be the good one.
I've reproduced the hang locally, i guess you already have took a look to the logs, mine locally are:

<details> <summary>local logs</summary>

```
OK (17.49s)
  initialize response capabilities
       text doc sync:                                                                                     --> {
    "method": "initialize",
    "params": {
        "processId": 8660,
        "trace": "off",
        "capabilities": {
 
        },
        "clientInfo": {
            "name": "lsp-test",
            "version": "0.14.0.1"
        },
        "rootUri": "file:///C:/TEMP/extra-dir-701178746171",
        "rootPath": "C:\\TEMP\\extra-dir-701178746171"
    },
    "jsonrpc": "2.0",
    "id": 0
}
ghcide version: 1.5.0.1 (GHC: 8.10) (PATH: D:\hls\dist-newstyle\build\x86_64-windows\ghc-8.10.7\ghcide-1.5.0.1\x\ghcide\build\ghcide\ghcide.exe) (GIT hash: 5cc51de6e13ad8b60cb4eb238951de5b44893696)
 [Info] Starting LSP server...
[Info] If you are seeing this in a terminal, you probably should have run WITHOUT the --lsp option!
[Info] Started LSP server in 0.01s
[Debug] setInitialDynFlags cradle: Cradle {cradleRootDir = "C:\\TEMP\\extra-dir-701178746171", cradleOptsProg = CradleAction: Default}
[Debug] runSubset: True
[Debug] Initializing exports map from hiedb
[Info] Registering ide configuration: IdeConfiguration {workspaceFolders = fromList [NormalizedUri (-6299900041106719098) "file:///C:/TEMP/extra-dir-701178746171"], clientSettings = hashed Nothing}
<-- {
    "result": {
        "capabilities": { ...
        }
    },
    "jsonrpc": "2.0",
    "id": 0
}
--> {
    "method": "initialized",
    "params": {},
    "jsonrpc": "2.0"
}
--> {
    "method": "workspace/didChangeConfiguration",
    "params": {
        "settings": {
            "haskell": {
                "checkParents": "CheckOnSave",
                "checkProject": false,
                "maxCompletions": 40,
                "formattingProvider": "ormolu",
                "plugin": {}
            }
        }
    },
    "jsonrpc": "2.0"
}
--> {
    "method": "shutdown",
    "params": null,
    "jsonrpc": "2.0",
    "id": 0
}
[Debug] Received shutdown message
```

</details>

I've generated a dump of the ghcide process in case it could help to trace the issue.

pepeiborra:
> Thanks for working on this, let me know if i can help in any way

Do you know why build caching is not working in CI? All the Windows builds are rebuilding the whole cabal database and taking >1h.

jneira:
hmm, now the cache is populated in the caching workflow for the master branch. That cache is propagated to all prs.
However last caching workflow is failing for all oss:

F.e. https://github.com/haskell/haskell-language-server/runs/4571064676

```
src\Language\Haskell\Brittany\Internal\Prelude.hs:17:1: error:

Error:     Could not find module ‘RdrName’
   |
17 | import RdrName                        as E ( RdrName )
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

That workflow is configured to not fail in any case, to make available the cache even partially. This pr is using that caching workflow with key (for win+ghc-9.0.1): `compiled-deps-Windows-9.0.1-2021-11-29T12-30-07Z-0c0e51e7952f2bb6f6a68b93790dcd1c9321c830a6ec4745f9abf8bda0b5994c`

Those error could be related with the fact he cache is not working . //cc @Anton-Latukha 
I guess we should have a way to know caching is failing, to fix it asap.

As the rest of os's but windows gets built at least once in the pr, the cache is being populated for them. But windows are not stored the cache precisely cause they are failing.
We should separate the build of the executable from the tests in two jobs, storing the cache if the build is succesful even if tests are not.

I was planning such separation which also could allow us use the alpine executable for benchs and tests easily. It will make bench, test and build have a common job which could be extracted: build the executables.

jneira:
> As the rest of os's but windows are getting built at least once in the pr, the cache is being populated for them. But windows ones are not stored the cache precisely cause they are failing.

hmm in the first run of this pr the cache for linux already worked 🤔

jneira:
hmm i think i find out a possible cause, i have to investigate a little bit more though

Anton-Latukha:
In response to:
https://github.com/haskell/haskell-language-server/pull/2504#issuecomment-997289195

tl; dr: It works as intended. `caching` builds all which is possible (& some GHC 9.0 do fail & failed at the time of creation of workflow), while agenda of caching is to cache & share as much cache of the current state & that means being fault-tolerant to `hls` build result.

The caching is made to cache all deps that were able to get built & it is specifically made project build result tolerant. Because caching workflows is not to check the builds, but to create & share as much relevant cache as possible, so for example a PR that fixes some build errors in the project - builds & merges faster because cache is present. Fault tolerance also has another good property, not all HLS GHC targets (plugins, their tests & bench deps included) are able to build with GHC 9.0, but fault tolerance to that allows to not manage targets for the workflow, while all the same time all that can get cached - gets cached. So yes, GHC 9.0 you linked to - would fail until all GHC 9.0 targets compile successfully (& that is not what caching should check or worry about). And notice that all other GHC versions & according to platform builds - build all targets successfully.

jneira:
> In fact i cant find a recent run where the cache was working fow windows. hmm i think i find out a possible cause, i have to investigate a little bit more though

https://github.com/haskell/haskell-language-server/pull/2507

//cc @Anton-Latukha

jneira:
> tl; dr: It works as intended. caching builds all which is possible (& some targets in GHC 9.0 do fail & failed at the time of the creation of workflow). The agenda of caching is to cache & share as much cache of the current state & that means being fault-tolerant to hls build result.

Yeah you are right and the windows problem is not related at all with
However i would prefer to keep the caching workflow not failing, without losing the fault tolerance.
Cause its cache will not be as good: if the caching run fails in half of the build it only will cache half of the dependencies, and not failing pr's using it will have to build the other half.
So i would like to have a way to know the caching workflow is failing. And i would prefer to lose the cache, be aware of the fail and fix it asap than having a incomplete cache for several days.

Anton-Latukha:
It is not directly related to the PR topic.

Lets talk about it separately. I would fold these.

PR #2505, opened 2021-12-18
Fix tracing of recordDirtyKeys

pepeiborra:
I misunderstood how [`OpenTelemetry.Eventlog.addEvent`](https://hackage.haskell.org/package/opentelemetry-0.7.0/docs/OpenTelemetry-Eventlog.html#v:addEvent) works, which is not totally surprising since there are no docs.

But it looks like, judging by other call sites, that the event goes in the second argument and the first argument is always "":

https://github.com/haskell/haskell-language-server/blob/779439876855872c523088e400526c6f1807bca2/ghcide/src/Development/IDE/Core/Tracing.hs#L123

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2505"></a>


Anton-Latukha (approved):
`""` argument leads to: https://hackage.haskell.org/package/opentelemetry-0.7.0/docs/src/OpenTelemetry.Eventlog_Internal.html#local-6989586621679057309
It is `k` there.

So it is a prefix/part of the message monoid.

There are is a minor cleaned-up detail (`withEventTrace :: ` no longer needs CPP), but that is a refactoring detail.

Seems trivial enough to be approved by me.

PR #2506, opened 2021-12-18
CI: caching: add early termination & run check on schedule

Anton-Latukha:
This PR allows to forget about cache pool exhaustion through internal branches.

Despite `caching` workflow runs on `master` only, GitHub caching is designed such that every workflow run that uses the cache `(test, bench)` can also save cache into the pool of own git repo. Which means the changes to the project configuration files inside PRs would produce caches for PR, which when there is enough internal branches PRs activity would push `master` cache out of the 10G FIFO pool because GitHub removes from the pool by the creation time. While PR caches that would stay - have a sharing & access scope limited to the PRs they are built for.

Screenshot-2021-12-18-22:36:49

Removal of the `master` cache out of the pool means that cache that was shared for all PRs - no longer there, which `⇒ ∀` PRs to build & save their own cache - which creates the pool exhaustion & congestion problem.

For example: in #2502 https://github.com/haskell/haskell-language-server/runs/4571293708?check_suite_focus=true we see that master cache got pushed-out & see this situation.

*afaik* there is no way (documented one or more KISS/elegant one) to prioritize main branch cache better.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2506"></a>

### DIFF
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -30,6 +30,18 @@ on:
   push:
     branches:
       - master
+  schedule:
+    # Refresh snapshot every (02+8*x):25 UTC
+    # When cache is present it is a light check workflow with early termination.
+    # When primary cache is not hit - runs the cache generation.
+    # Why: GitHub repo has 10G pool & on overflow GitHub removes caches in FIFO manner.
+    # When internal branche PRs save into the same pool -
+    #   their cache is accessible only inside of the scope of the PR.
+    # If main cache is forced out - there are no cache shared between PRs,
+    #   which implies all PRs would start to create & save their cache.
+    # Reinstitution of the main chache puts it back into FIFO
+    #   & so it gets shared across all PRs.
+    - cron: "25 2/8 * * *"
 
 env:
   cabalBuild: "v2-build all --enable-tests --enable-benchmarks"
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -130,6 +130,7 @@ jobs:
           restore-keys: ${{ env.cache-name }}-
 
       - name: Compiled deps cache
+        id: compiled-deps
         uses: actions/cache@v2
         env:
           cache-name: compiled-deps
@@ -141,9 +142,12 @@ jobs:
                 ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
                 ${{ env.cache-name }}-${{ runner.os }}-
 
-      - run: cabal update
+      - if: (! steps.compiled-deps.outputs.cache-hit)
+        run: |
+          cabal update
 
-      - name: Download all sources
+      - if: (! steps.compiled-deps.outputs.cache-hit)
+        name: Download all sources
         run: |
           cabal $cabalBuild --only-download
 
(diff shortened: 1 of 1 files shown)


jneira:
And what about if we use a double key: one for master and other for pr's and make pr's use their own and master as recovery key?
That way master cache will not be pushed out by prs. Taking in account pr's cache from forks are not taking our limit...

Anton-Latukha:
No, that is not how GitHub caching works.

What got suggested is to revert into old solution, which not accounts for how GitHub CI caching works.

Despite of having any number of name spaces - all of them are counted into a 1 repo pool.

Briefly: 10G pool is given for every repo. HLS repo has 10G pool, every separate fork has 10G pool. Pools are FIFO. The newest cache saves push out the oldest ones. Regardless how caches or keys, or any configuration of them is made or named. GitHub does not provide other way of how caches get trimmed - it is just FIFO done by GitHub.

Every forks has 10G pool. But also is able to use `master` cache.

jneira:
Ok that makes sense (again). 
The run you have linked as example is a windows one, but in that workflow the ubuntu builds are very fast (f.e. https://github.com/haskell/haskell-language-server/runs/4571293606?check_suite_focus=true). 
So it seems it is affecting only windows. And windows seems to not use the cache since long time ago.
Not sure if that fits the hypothesis of the lose of master cache as the cause.

Anton-Latukha:
Yep. I became aware of the Windows caching problem & its cache somehow does not work properly. I think it is better to devote attention to it in https://github.com/haskell/haskell-language-server/pull/2507

Anton-Latukha:
Early termination is seen in: https://github.com/haskell/haskell-language-server/actions/runs/1598559261

Now removing the WIP PR hook.

Anton-Latukha:
Added skipping of `{caching, test, bench}` to `{caching, test, bench, nix}`. To give a fast CI development cycle. Ironically, because providing the skip also changed all those configurations - they all need a rebuild check. It would work for further commits/CI development of those workflows.

Anton-Latukha:
The result now was: https://github.com/haskell/haskell-language-server/runs/4574676469?check_suite_focus=true

Well, there are doc post-fixes. Not we are able to check how workflow skipping works. Because I upload changes to docs in `caching` - CI should skip `{test, bench, nix}` builds.

Anton-Latukha:
Hmmm: I do not understand why skipping does not work. https://github.com/haskell/haskell-language-server/runs/4575267862?check_suite_focus=true

Anywaay... nice idea. Since I do not know enough how it works & it is not the scope of the PR - opening a report #2510 & reverting that feature here.

jneira (approved):
great analysis and a pragmatic solution, thanks!

Anton-Latukha:
Yep. 3 checks a day may seem excessive. Until some hackathon 8), or community growth.

Thank you 8)

Anton-Latukha:
Restarted the checks to see the caching at work.

We shortened the CI cycle by 50 minutes: https://github.com/haskell/haskell-language-server/runs/4591646270?check_suite_focus=true :cyclone: :partying_face:

jneira:
really great news, love to see again a windows build of 6 min

Anton-Latukha:
Yes, >2 hours were very painful pace.
1:15 is much tolerable (this time feel very tolerable to me), now the person knows that basically in an hour testing would be done.

Would've been great to sneak favors from GitHub :wink:.

jneira:
> Would've been great to sneak favors from GitHub wink.

I am on it :smile:

PR #2511, opened 2021-12-19
Sort out some compatibility issues

alanz:
`mkPrintUnqualified` needs a `UnitEnv`, I have no idea where to get one.
I suspect we could use a different print function, for the only place this is called.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2511"></a>


jhrcek:
> mkPrintUnqualified needs a UnitEnv, I have no idea where to get one.

UnitEnv is part of https://hackage.haskell.org/package/ghc-9.2.1/docs/GHC-Driver-Env.html#t:HscEnv and can be retrieved using `hsc_unit_env :: HscEnv -> UnitEnv`

pepeiborra:
Merging Alan's branch to the `haskell` one, further changes can be done there

PR #2515, opened 2021-12-20
cabal-*.project: index-state +1s

Anton-Latukha:
It is made to fix the current situation, for example, if `Caching` workflow in `master` got canceled for some reason, on a successful run of the *action* in the *workflows* - workflows still runs the post-hooks that *actions* have, which would cache & share the `cabal` `store` of the canceled workflow.

On merge, this change would refresh caches, which should fix `Windows` build waits.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2515"></a>


Anton-Latukha:
Ok. It currently fixes the cache situation for the time being.

In https://github.com/haskell/haskell-language-server/runs/4584037502?check_suite_focus=true we see that the input cache was 110Mb, output & the cache that got saved is 203Mb.

Anton-Latukha:
As we agreed on in the chat room https://matrix.to/#/!oOjZFsoNYPAbTEgSOA:libera.chat/$-Wv69VWdRDLGm-KNPvn5HkCGdRAwd1Ygk1601yE8db0?via=libera.chat&via=matrix.org&via=monoid.al.

I would merge this.

jneira:
Maybe it is a minor thing but with non existent hackage index timestamps all builds logs:

```
Warning: Requested index-state 2021-11-29T12:30:08Z is newer than
'hackage.haskell.org'! Falling back to older state (2021-11-29T12:28:17Z).
```

which is a little bit annoying.

I see the advantage of using artifical index state to invalidate cache but maybe we could update to the last index state given by `cabal update`
I mean we will have to update it sooner or later and ci will tell us if there is something wrong so we can try to find an older one.

Anton-Latukha:
1. The timestamp was reverting to `2021-11-29T12:28:17Z` before also. In fact that is why I do +1s - it changes literally nothing, as `cabal` still uses the last Hackage snapshot, it is being `2021-11-29T12:28:17Z`.

2. CI already does `cabal v2-update` always anyway. But because of `cabal.project` pinning - it uses `2021-11-29T12:28:17Z` Hackages snapshot.

3. If I to update Hackage index not +`1`s, but +`2629800`s (1 month) - it has all sort of bigger consequences - I would need to address the update to all kinds of development updates - which expands the scope from fixing the cache - to doing the project dependency update without discussion is that needed in the first place - & so 1) that PR would be harder to do & 2) harder to pass 3) agenda of PR was to fix caching to continue work on the situation.

jneira (approved):
ok, let it be +1, we will end to have a real one anyways and the actual one emits the warning, thanks for fixing cache

Anton-Latukha:
Thank you.

I just was focused on the other thing in this PR & was not prepared to change the scope.

Anton-Latukha:
Superb, workflow finished: https://github.com/haskell/haskell-language-server/runs/4591215047?check_suite_focus=true

Issue #2514, opened 2021-12-20
Reduce use of partial functions from HLS


michaelpj:
HLS has quite a lot of use of partial functions. I'm sure many of these are benign, but we do occasionally get users reporting crashes with such (usually totally useless) errors.

Gradually reducing these is an easy, helpful thing for someone to do. Good new contributor issue!

pepeiborra:
Could you share a few examples?

michaelpj:
30-odd usages of `head`, 15 or so `foldl1`s, 10 or so `!!`s.

I think many of these are fine, but therefore usually trivial to avoid. It would be nice to have a blanket "no partial functions" rule.

michaelpj:
30-ish `fromJust`...

jhrcek:
I volunteer to work on this, as I wanted to contribute to hls in more substantial way.
Having looked at some of the partial functions usages, many of them require some domain-specific knowledge, e.g. about the Shake build stuff etc. For example I'm not at all sure about these: https://github.com/haskell/haskell-language-server/blob/070f16fbc21be5c9c889e09fe78ec2f2e3d90ed9/ghcide/src/Development/IDE/Core/Shake.hs#L854-L867 
Can we prove these will never fail? What should the function do if they fail?

My preliminary plan is this:
1. Come up with a list of partial functions that we'd like to eventually replace with something better.
Apart from the above mentioned  `head`, `foldl1`, `(!!)`, `fromJust` i noticed quite a few `Map.!`, `IntMap.!`, `Array.!`, `last`
Can you think of others? Let's keep expanding the list in the description of this issue!
2. replace the obvious ones
3. when it's impossible to "prove" that given partial function will always succeeds, replace it with some variant that gives more informative exception (e.g.  [headNote](https://hackage.haskell.org/package/safe-0.3.19/docs/Safe.html#v:headNote) from safe). This way we at least get more informative errors when one of these fails and we'll know from user reports which ones actually occur and should be fixed. Wdyt?

jneira:
The plan sounds really great to me, given at minimum we will have more info about the error.
Many thanks for willing to contribute

pepeiborra:
> Re: `usesWithStale`, I believe it could be simplified to take a `NonEmpty NormalizedFilePath` and return a `NonEmpty` list of results, which would allow avoiding the `head`. Or it could even work over an arbitrary `Traversable` if you wanted to get fancy, and then you could use `Identity` for the singleton cases you see there.

[...]

PR #2519, opened 2021-12-21
Reduce usages of Prelude/Data.Text.head

jhrcek:
First batch of fixes to https://github.com/haskell/haskell-language-server/issues/2514

This is going to be harder than I thought :P
There are more `head`s to come, but opening this just to get initial feedback on the general approach.

Also (due to my ocd) I had to fix bunch of hlint warnings which lead to meaningful simplifications.

Vague questions: how are you guys dealing with running tests locally?
I'm getting bunch of `fd:28: hPutBuf: resource vanished (Broken pipe)` when running `cabal test` locally..

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2519"></a>

### DIFF
--- a/ghcide/bench/lib/Experiments.hs
+++ b/ghcide/bench/lib/Experiments.hs
@@ -49,6 +49,7 @@ import           Language.LSP.Types              hiding
 import           Language.LSP.Types.Capabilities
 import           Numeric.Natural
 import           Options.Applicative
+import           Safe                            (headNote)
 import           System.Directory
 import           System.Environment.Blank        (getEnv)
 import           System.FilePath                 ((<.>), (</>))
@@ -474,7 +475,7 @@ runBench runSess b = handleAny (\e -> print e >> return badRun)
         output $ "Setting up document contents took " <> showDuration d
         -- wait again, as the progress is restarted once while loading the cradle
         -- make an edit, to ensure this doesn't block
-        let DocumentPositions{..} = head docs
+        let DocumentPositions{..} = headNote "Experiments.runBench" docs
         changeDoc doc [charEdit stringLiteralP]
         waitForProgressDone
         return docs
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -35,6 +35,8 @@ import qualified Data.HashMap.Strict                  as HM
 import           Data.Hashable
 import           Data.IORef
 import           Data.List
+import           Data.List.NonEmpty                   (NonEmpty ((:|)))
+import qualified Data.List.NonEmpty                   as NE
 import qualified Data.Map.Strict                      as Map
 import           Data.Maybe
 import qualified Data.Text                            as T
@@ -303,9 +305,9 @@ loadSessionWithOptions SessionLoadingOptions{..} dir = do
                   -- compilation but these are the true source of
                   -- information.
                   new_deps = RawComponentInfo (homeUnitId_ df) df targets cfp opts dep_info
-                                : maybe [] snd oldDeps
+                                :| maybe [] snd oldDeps
                   -- Get all the unit-ids for things in this component
-                  inplace = map rawComponentUnitId new_deps
+                  inplace = fmap rawComponentUnitId new_deps
 
               new_deps' <- forM new_deps $ \RawComponentInfo{..} -> do
                   -- Remove all inplace dependencies from package flags for
@@ -347,7 +349,9 @@ loadSessionWithOptions SessionLoadingOptions{..} dir = do
               -- . The information for the new component which caused this cache miss
               -- . The modified information (without -inplace flags) for
               --   existing packages
-              pure (Map.insert hieYaml (newHscEnv, new_deps) m, (newHscEnv, head new_deps', tail new_deps'))
+              pure ( Map.insert hieYaml (newHscEnv, NE.toList new_deps) m
+                   , (newHscEnv, NE.head new_deps', NE.tail new_deps')
+                   )
 
 
     let session :: (Maybe FilePath, NormalizedFilePath, ComponentOptions, FilePath)
@@ -495,13 +499,12 @@ loadSessionWithOptions SessionLoadingOptions{..} dir = do
             sessionOpts (join cachedHieYamlLocation <|> hieYaml, file) `Safe.catch` \e ->
                 return (([renderPackageSetupException file e], Nothing), maybe [] pure hieYaml)
 
-    returnWithVersion $ \file -> do
-      opts <- liftIO $ join $ mask_ $ modifyVar runningCradle $ \as -> do
+    returnWithVersion $ \file ->
+      liftIO $ join $ mask_ $ modifyVar runningCradle $ \as -> do
         -- If the cradle is not finished, then wait for it to finish.
         void $ wait as
         as <- async $ getOptions file
         return (as, wait as)
-      pure opts
 
 -- | Run the specific cradle on a specific FilePath via hie-bios.
 -- This then builds dependencies or whatever based on the cradle, gets the
@@ -758,7 +761,7 @@ getDependencyInfo fs = Map.fromList <$> mapM do_one fs
 -- ID. Therefore we create a fake one and give them all the same unit id.
 removeInplacePackages
     :: UnitId     -- ^ fake uid to use for our internal component
-    -> [UnitId]
+    -> NonEmpty UnitId
     -> DynFlags
     -> (DynFlags, [UnitId])
 removeInplacePackages fake_uid us df = (setHomeUnitId_ fake_uid $
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -168,6 +168,7 @@ import qualified Ide.PluginUtils                        as HLS
 import           Ide.Types                              (PluginId)
 import qualified "list-t" ListT
 import qualified StmContainers.Map                      as STM
+import           Safe                                   (headNote)
 
 -- | We need to serialize writes to the database, so we send any function that
 -- needs to write to the database over the channel, where it will be picked up by
@@ -853,18 +854,18 @@ defineNoDiagnostics op = defineEarlyCutoff $ RuleNoDiagnostics $ \k v -> (Nothin
 -- | Request a Rule result if available
 use :: IdeRule k v
     => k -> NormalizedFilePath -> Action (Maybe v)
-use key file = head <$> uses key [file]
+use key file = headNote "Development.IDE.Core.Shake.use" <$> uses key [file]
 
 -- | Request a Rule result, it not available return the last computed result, if any, which may be stale
 useWithStale :: IdeRule k v
     => k -> NormalizedFilePath -> Action (Maybe (v, PositionMapping))
-useWithStale key file = head <$> usesWithStale key [file]
+useWithStale key file = headNote "Development.IDE.Core.Shake.useWithStale" <$> usesWithStale key [file]
 
 -- | Request a Rule result, it not available return the last computed result which may be stale.
 --   Errors out if none available.
 useWithStale_ :: IdeRule k v
     => k -> NormalizedFilePath -> Action (v, PositionMapping)
-useWithStale_ key file = head <$> usesWithStale_ key [file]
+useWithStale_ key file = headNote "Development.IDE.Core.Shake.useWithStale_" <$> usesWithStale_ key [file]
 
 -- | Plural version of 'useWithStale_'
 usesWithStale_ :: IdeRule k v => k -> [NormalizedFilePath] -> Action [(v, PositionMapping)]
@@ -932,7 +933,7 @@ useNoFile :: IdeRule k v => k -> Action (Maybe v)
 useNoFile key = use key emptyFilePath
 
 use_ :: IdeRule k v => k -> NormalizedFilePath -> Action v
-use_ key file = head <$> uses_ key [file]
+use_ key file = headNote "Development.IDE.Core.Shake.use_" <$> uses_ key [file]
 
 useNoFile_ :: IdeRule k v => k -> Action v
 useNoFile_ key = use_ key emptyFilePath
--- a/ghcide/src/Development/IDE/GHC/Compat/Units.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat/Units.hs
@@ -89,6 +89,7 @@ import           Data.Map                        (Map)
 #endif
 import           Data.Either
 import           Data.Version
+import qualified Data.List.NonEmpty              as NE
 
 #if MIN_VERSION_ghc(9,0,0)
 type PreloadUnitClosure = UniqSet UnitId
@@ -324,7 +325,7 @@ moduleUnit =
     Module.moduleUnitId
 #endif
 
-filterInplaceUnits :: [UnitId] -> [PackageFlag] -> ([UnitId], [PackageFlag])
+filterInplaceUnits :: NE.NonEmpty UnitId -> [PackageFlag] -> ([UnitId], [PackageFlag])
 filterInplaceUnits us packageFlags =
   partitionEithers (map isInplace packageFlags)
   where
@@ -335,7 +336,7 @@ filterInplaceUnits us packageFlags =
         then Left $ toUnitId  u
         else Right p
 #else
-      if u `elem` us
+      if u `elem` NE.toList us
         then Left u
         else Right p
 #endif
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -169,7 +169,7 @@ findSigOfBind range bind =
 
     findSigOfGRHSs :: GRHSs p (LHsExpr p) -> Maybe (Sig p)
     findSigOfGRHSs grhs = do
-        if _start range `isInsideSrcSpan` (getLoc $ grhssLocalBinds grhs)
+        if _start range `isInsideSrcSpan` getLoc (grhssLocalBinds grhs)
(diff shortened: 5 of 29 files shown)


plugins/hls-tactics-plugin/src/Wingman/Judgements.hs
```diff
@@ -34,6 +34,9 @@ buildHypothesis
   where
     go (occName -> occ, t)
       | Just ty <- t
+      -- TODO what's the intention behind this check?
+      -- can it be replace with something like `not . isSymOcc`?
+      -- https://hackage.haskell.org/package/ghc-9.2.1/docs/GHC-Plugins.html#v:isSymOcc
```
jhrcek, line 39: Probably question for @isovector 
My intention is to get rid of this usage of `head`

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -706,9 +706,10 @@ suggestModuleTypo Diagnostic{_range=_range,..}
 --     Could not find module ‘Data.Cha’
 --     Perhaps you meant Data.Char (from base-4.12.0.0)
     | "Could not find module" `T.isInfixOf` _message
-    , "Perhaps you meant"     `T.isInfixOf` _message = let
-      findSuggestedModules = map (head . T.words) . drop 2 . T.lines
-      proposeModule mod = ("replace with " <> mod, TextEdit _range mod)
+    , "Perhaps you meant"     `T.isInfixOf` _message =
+      -- TODO findSuggestedModules never extracts module name - fix the parsing logic
```
jhrcek, line 710: This parsing logic is definitely broken. If I'm not wrong, this will always extract the word "Perhaps" ?!

pepeiborra: Did you check the tests?

ghcide/src/Development/IDE/GHC/Compat/Units.hs
```diff
@@ -324,7 +325,7 @@ moduleUnit =
     Module.moduleUnitId
 #endif
 
-filterInplaceUnits :: [UnitId] -> [PackageFlag] -> ([UnitId], [PackageFlag])
+filterInplaceUnits :: NE.NonEmpty UnitId -> [PackageFlag] -> ([UnitId], [PackageFlag])
```
jneira, line 328: I would use types unqualified, hope it is the pattern for other data types

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
```diff
@@ -321,7 +322,7 @@ extendImportViaParent df parent child (L l it@ImportDecl{..})
       let childRdr = L srcChild $ mkRdrUnqual $ mkVarOcc child
           isParentOperator = hasParen parent
       when hasSibling $
-        addTrailingCommaT (head pre)
+        addTrailingCommaT (headNote "Development.IDE.Plugin.CodeAction.ExactPrint.extendImportViaParent" pre)
```
pepeiborra, line 325: Remove

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
```diff
@@ -396,9 +397,9 @@ extendHiding symbol (L l idecls) mlies df = do
   if hasSibling
     then when hasSibling $ do
       addTrailingCommaT x
-      addSimpleAnnT (head lies) (DP (0, 1)) []
+      addSimpleAnnT (headNote "Development.IDE.Plugin.CodeAction.ExactPrint.extendHiding1" lies) (DP (0, 1)) []
```
pepeiborra, line 400: Remove

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
```diff
@@ -396,9 +397,9 @@ extendHiding symbol (L l idecls) mlies df = do
...
       unless (null $ tail lies) $
-        addTrailingCommaT (head lies) -- Why we need this?
+        addTrailingCommaT (headNote "Development.IDE.Plugin.CodeAction.ExactPrint.extendHiding2" lies) -- Why we need this?
```
pepeiborra, line 402: Remove

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -210,7 +211,7 @@ extendImportHandler :: CommandFunction IdeState ExtendImport
 extendImportHandler ideState edit@ExtendImport {..} = do
   res <- liftIO $ runMaybeT $ extendImportHandler' ideState edit
   whenJust res $ \(nfp, wedit@WorkspaceEdit {_changes}) -> do
-    let (_, List (head -> TextEdit {_range})) = fromJust $ _changes >>= listToMaybe . toList
+    let (_, List (headNote "Development.IDE.Plugin.Completions.extendImportHandler" -> TextEdit {_range})) = fromJust $ _changes >>= listToMaybe . toList
```
pepeiborra, line 214: Remove

ghcide/src/Development/IDE/Spans/Documentation.hs
```diff
@@ -63,7 +64,9 @@ lookupKind env mod =
     fmap (fromRight Nothing) . catchSrcErrors (hsc_dflags env) "span" . lookupName env mod
 
 getDocumentationTryGhc :: HscEnv -> Module -> Name -> IO SpanDoc
-getDocumentationTryGhc env mod n = head <$> getDocumentationsTryGhc env mod [n]
+getDocumentationTryGhc env mod n =
+    headNote "Development.IDE.Spans.Documentation.getDocumentationTryGhc"
```
pepeiborra, line 68: Remove

jneira:
> I'm getting bunch of fd:28: hPutBuf: resource vanished (Broken pipe) when running cabal test locally..

I got them in windows long time ago, what is your setup?

ghcide/test/exe/Main.hs
```diff
@@ -4886,7 +4886,9 @@ projectCompletionTests =
               "import ALocal"
             ]
         compls <- getCompletions doc (Position 1 13)
-        let item = head $ filter ((== "ALocalModule") . (^. Lens.label)) compls
+        item <- case find (\c -> c ^. Lens.label == "ALocalModule") compls of
+            Nothing -> liftIO . assertFailure $ "No completion with label ALocalModule found in : " <> show compls
```
pepeiborra, line 4890: Do we enable -fno-ignore-assert ? I don't remember

plugins/hls-tactics-plugin/src/Wingman/CaseSplit.hs
```diff
@@ -105,5 +106,5 @@ splitToDecl fixity name ams = do
 iterateSplit :: AgdaMatch -> [AgdaMatch]
 iterateSplit am =
   let iterated = iterate (agdaSplit =<<) $ pure am
-   in fmap wildify . head . drop 5 $ iterated
+   in fmap wildify . (\xs -> atNote "Wingman.CaseSplit.iterateSplit" xs 5) $ iterated
```
pepeiborra, line 109: I would just use !!

pepeiborra (changes requested):
Most test suites need TASTY_NUM_THREADS=1 to override the default behaviour that runs tests concurrently. The ghcide testsuite sets this implicitly I believe, but others don't. 

If that's not enough, just copy the command line from CI

plugins/hls-tactics-plugin/src/Wingman/Judgements.hs
isovector: This is checking that the bindings in scope were created in source Haskell. GHC makes lots of bindings whose name start with `$` or other odd things.

You could replace it with `maybe False isAlpha $ headMay $ occNameString occ`

ghcide/bench/lib/Experiments.hs
```diff
@@ -474,7 +475,7 @@ runBench runSess b = handleAny (\e -> print e >> return badRun)
         output $ "Setting up document contents took " <> showDuration d
         -- wait again, as the progress is restarted once while loading the cradle
         -- make an edit, to ensure this doesn't block
-        let DocumentPositions{..} = head docs
+        let DocumentPositions{..} = headNote "Experiments.runBench" docs
```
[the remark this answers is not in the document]
michaelpj: Yes, and I think if we're going to do this we should replace it with a meaningful error, not just the position where it occurred (although that's somewhat useful).

We also could be more lenient about partial functions in things like tests and benchmarks.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -853,18 +854,18 @@ defineNoDiagnostics op = defineEarlyCutoff $ RuleNoDiagnostics $ \k v -> (Nothin
 -- | Request a Rule result if available
 use :: IdeRule k v
     => k -> NormalizedFilePath -> Action (Maybe v)
-use key file = head <$> uses key [file]
+use key file = headNote "Development.IDE.Core.Shake.use" <$> uses key [file]
```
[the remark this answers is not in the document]
michaelpj: As I commented on the issue, I think you could solve all of these at a stroke by making the underlying `uses` functions work on `NonEmpty`, which I think would be fine.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -758,7 +761,7 @@ getDependencyInfo fs = Map.fromList <$> mapM do_one fs
 -- ID. Therefore we create a fake one and give them all the same unit id.
 removeInplacePackages
     :: UnitId     -- ^ fake uid to use for our internal component
-    -> [UnitId]
+    -> NonEmpty UnitId
```
michaelpj, line 764: Perhaps simpler to just convert to a normal list before calling this - I don't think _this_ function needs to require its argument to be non-empty!

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1553,7 +1551,7 @@ rangesForBinding' b (L l (IEThingWith _ thing _  inners labels))
 rangesForBinding' _ _ = []
 
 -- | 'matchRegex' combined with 'unifySpaces'
-matchRegexUnifySpaces :: T.Text -> T.Text -> Maybe [T.Text]
+matchRegexUnifySpaces :: T.Text -> T.Text -> Maybe (NonEmpty T.Text)
```
michaelpj, line 1554: I get why this is the return type, but it's funny to see `Maybe (NonEmpty a)` instead of `[a]`!

ghcide/src/Development/IDE/GHC/Compat/Units.hs
michaelpj: Ditto, not sure this one should change.

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
michaelpj: Suggest instead changing the `not null` guard on line 316 into a pattern guard that matches on `pre` and extracts the head there.

ghcide/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -945,9 +949,11 @@ suggestImportDisambiguation df (Just txt) ps@(L _ HsModule {hsmodImports}) fileC
...
                 ++ [HideOthers restImports | not (null restImports)]
             ] ++ [ ( renderUniquify mode T.empty symbol True
-              , disambiguateSymbol ps fileContents diag symbol mode
-              ) | local, not (null targetsWithRestImports)
-                , let mode = HideOthers (uncurry (:) (head targetsWithRestImports))
+                   , disambiguateSymbol ps fileContents diag symbol mode
+                   )
+                 | local
+                 , not (null targetsWithRestImports)
+                 , let mode = HideOthers (uncurry (:) (head targetsWithRestImports))
```
michaelpj, line 956: still got a head here

ghcide/src/Development/IDE/Plugin/Completions.hs
michaelpj: I think this should just do nothing if the match fails. It's already inside a `whenJust` for the response coming in.

ghcide/src/Development/IDE/Spans/Documentation.hs
michaelpj: This seems potentially kind of bad! I think we should do something here. Just return a `Maybe`?

plugins/hls-floskell-plugin/src/Ide/Plugin/Floskell.hs
```diff
@@ -49,7 +50,9 @@ findConfigOrDefault file = do
   case mbConf of
     Just confFile -> readAppConfig confFile
     Nothing ->
-      let gibiansky = head (filter (\s -> styleName s == "gibiansky") styles)
+      let gibiansky = case find (\s -> styleName s == "gibiansky") styles of
+            Just gibStyle -> gibStyle
+            Nothing -> error "Ide.Plugin.Floskell.findConfigOrDefault: Style with name 'gibiansky' not found"
```
michaelpj, line 55: Can we just not set the `appStyle` if this fails?

plugins/hls-tactics-plugin/src/Wingman/CaseSplit.hs
michaelpj: Right, this one is a bit funny because we know the list is infinite. Hmm.

plugins/hls-tactics-plugin/src/Wingman/Machinery.hs
```diff
@@ -342,7 +342,7 @@ lookupNameInContext name = do
 getDefiningType
     :: TacticsM CType
 getDefiningType = do
-  calling_fun_name <- fst . head <$> asks ctxDefiningFuncs
+  calling_fun_name <- asks (fst . headNote "Wingman.Machinery.getDefiningType" . ctxDefiningFuncs)
```
michaelpj, line 345: Just below we do something with `failure`, can we also use that branch if we fail here?

plugins/hls-tactics-plugin/src/Wingman/Simplify.hs
```diff
@@ -35,9 +36,9 @@ pattern Lambda pats body <-
 -- | Simlify an expression.
 simplify :: LHsExpr GhcPs -> LHsExpr GhcPs
 simplify
-  = head
-  . drop 3   -- Do three passes; this should be good enough for the limited
-             -- amount of gas we give to auto
+  = -- Do three passes; this should be good enough for the limited
+    -- amount of gas we give to auto
+    (\xs -> atNote "Wingman.Simplify.simplify" xs 3)
```
michaelpj, line 41: Another annoying infinite list.

Anton-Latukha:
*Nota bene*: I addressed:
https://github.com/haskell/haskell-language-server/blob/bd0046b7089e32e309f7b29ee6ec1a808ca1db86/ghcide/src/Development/IDE/Spans/Documentation.hs#L64-L67
In https://github.com/haskell/haskell-language-server/pull/2349, - would soon merge it.

michaelpj (approved):
Generally looks good!

In future it would make this sort of thing easier to review if you don't also apply hlint suggestions. It's tempting, but reviewer bandwith is a bottleneck for us. You can always do a trivial "fix hlint suggestions PR" :)

Pepe, I'm interpreting your "Remove" comments as "go back to `head` here, because due to the implicit `HasCallStack` it is equally informative to `headNote`", is that right?

pepeiborra:
> Pepe, I'm interpreting your "Remove" comments as "go back to `head` here, because due to the implicit `HasCallStack` it is equally informative to `headNote`", is that right?

Correct

ghcide/src/Development/IDE/Spans/Documentation.hs
pepeiborra: I think this is similar to `uses` in that we are local equational reasoning suffices to prove the absence of errors. You could generalise `getDocumentationsTryGhc` to take  `Traversable` and use a `NonEmpty` list in `getDocumentationTryGhc`

ghcide/src/Development/IDE/Core/Shake.hs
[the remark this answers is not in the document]
pepeiborra: Yes, I support the generalization to any `Traversable` to use `NonEmpty` here

ghcide/bench/lib/Experiments.hs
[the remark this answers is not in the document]
pepeiborra: We don't want to have any errors escaping out in this way in ghcide. 

But this particular call site is in the benchmark suite which is allowed to error out, and a position is more helpful than any description in that case.

jhrcek:
Thanks for reviews, I'll address your comments tomorrow.

ghcide/src/Development/IDE/Plugin/CodeAction.hs
jhrcek: This doesn't seem to be tested at all.
Also no action is generated for these kinds of errors, probably because the "Perhaps you meant" in on the 2nd line of the message, so it gets dropped by the drop 2 part of the original impl.
Let me add test and fix this in separate PR.

jhrcek:
I decided to close this PR as it contains too many unrelated changes.
I'll open multiple smaller, easier to review and hopefully less controversial PRs, addressing most of your comments and suggestions.

ghcide/src/Development/IDE/Plugin/CodeAction.hs
googleson78: It *is* funny, but in my experience it's actually the more convenient/"right" choice, whenever you want to do something special for the empty list case.

PR #2520, opened 2021-12-21
CI: caching: fix early termination expression check & cabal.project replacement

Anton-Latukha:
Details are in the commit messages.

1. We not noticed the first specific of the CI, because the early cutoff behaved as it should. It was respected to return `'false'` to terminate early, but that check always returned `'false'`. When working in GHC 9.2 PR could not force the caching build to start, because check that I fix - always skipped the steps (always returned `False`). New check is done in the way how it is done in other `if: ... != 'true'` statements in the workflow.

2. The second specific is both Ubuntu-specific & maybe  GitHub specific. Some OSes run `alias cp='cp -i'` & when I tried to `rm` manually - `cabal.project` appeared to be protected from simple `rm` by setinning (most probably) of a "sticky bit" on `$HOME` directory. I found that-out by working in GHC 9.2 PR (where `index-state` is actually different between `.project` files & noticed that CI cache key was with the same inferred `index-state` in all jobs). Now `rm -f` forced & that allows removing file with a sticky bit. log reports the situation of files changes there: https://github.com/haskell/haskell-language-server/runs/4597911492?check_suite_focus=true#step:14:460

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2520"></a>

### DIFF
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -100,7 +100,9 @@ jobs:
       - if: matrix.ghc == '9.0.1'
         name: (GHC 9.0.1) Use modified `cabal.project`
         run: |
-          # This seems insane, but it is needed.
+          # File has some protections preventing regular `rm`.
+          # `&&` insures `rm -f` returns positive.
+          # Some platforms have `alias cp='cp -i'`.
           rm -f -v cabal.project && cp -v cabal-ghc901.project cabal.project
       - if: runner.os == 'Windows' && matrix.ghc == '8.8.4'
         name: (Windows,GHC 8.8) Modify `cabal.project` to workaround segfaults
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -103,7 +103,7 @@ jobs:
           # File has some protections preventing regular `rm`.
           # (most probably sticky bit is set on $HOME)
           # `&&` insures `rm -f` return is positive.
-          # Many platforms aslo have `alias cp='cp -i'`.
+          # Many platforms also have `alias cp='cp -i'`.
           rm -f -v cabal.project && cp -v cabal-ghc901.project cabal.project
       - if: runner.os == 'Windows' && matrix.ghc == '8.8.4'
         name: (Windows,GHC 8.8) Modify `cabal.project` to workaround segfaults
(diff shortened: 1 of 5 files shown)


Anton-Latukha:
Ok. Now lets actually check the early termination & the from that to new key (should be full run).

Anton-Latukha:
Early termination: https://github.com/haskell/haskell-language-server/runs/4599033379?check_suite_focus=true

Anton-Latukha:
I advanced the `index-state`.

Which changes the key. So the old key of cache & new id for cache of the workflow does not align & so the caching starts.

Resulting into the proper long caching run: https://github.com/haskell/haskell-language-server/runs/4599160496?check_suite_focus=true

Anton-Latukha:
Ok, now I need to wait until this run completes & then drop the modifications. & have a 1 clean run on `{test,bench}`, it is honest to do at least because I changed `test` workflow.

Anton-Latukha:
Ok, now run is complete https://github.com/haskell/haskell-language-server/pull/2520/checks.

Disabling `caching` & enabling `{test,bench}`.

Anton-Latukha:
CI run completed successfully: https://github.com/haskell/haskell-language-server/runs/4599610364?check_suite_focus=true.

Anton-Latukha:
Seems I checked everything about early termination this time.

& seem that `cabal.project` switcharoo now works.

I hope we solved the main problems there & hope that soon we would exhaust hidden bugs in the CI bugs would start to approach zero.

.github/workflows/caching.yml
```diff
@@ -100,7 +100,11 @@ jobs:
       - if: matrix.ghc == '9.0.1'
         name: (GHC 9.0.1) Use modified `cabal.project`
         run: |
-          cp cabal-ghc901.project cabal.project
+          # File has some protections preventing regular `rm`.
+          # (most probably sticky bit is set on $HOME)
+          # `&&` insures `rm -f` return is positive.
+          # Many platforms aslo have `alias cp='cp -i'`.
```
jneira, line 106: small typo: also

jneira (approved):
lgtm, good catches, thanks

.github/workflows/caching.yml
jneira, line 106:
```suggestion
          # Many platforms also have `alias cp='cp -i'`.
```

Anton-Latukha:
Would note, that now because according `.project` are loaded - CI pipeline may ship according context & so effects/bugs into pull requests.

PR #2523, opened 2021-12-22
Fix some hlint warnings

jhrcek:
Splitting of hlint fixes from https://github.com/haskell/haskell-language-server/pull/2519

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2523"></a>


hie-compat/src-ghc86/Compat/HieUtils.hs
```diff
@@ -126,14 +126,13 @@ recoverFullType i m = go i
...
-        Nothing -> do
-          ht <- go t
-          extendHTS t ht
+getTypeIndex t = do
+    tm <- gets tyMap
+    case lookupTypeMap tm t of
+      Just i -> return i
+      Nothing -> do
+        ht <- go t
+        extendHTS t ht
```
Ailrun, line 135: I think the point of hie-compat library is using the original code as-is. The original `ghc` library code itself contains this redundant `otherwise`, I think it's better to keep it as-is.

hie-compat/src-ghc86/Compat/HieTypes.hs
```diff
@@ -5,7 +5,6 @@ For more information see https://gitlab.haskell.org/ghc/ghc/wikis/hie-files
 -}
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE DeriveDataTypeable #-}
-{-# LANGUAGE TypeSynonymInstances #-}
```
Ailrun, line 8: Same here

Ailrun (commented):
Thank you for the update!

hie-compat/src-ghc86/Compat/HieUtils.hs
jhrcek: Alright, I reverted all changes within hie-compat module.

Anton-Latukha (approved):
Looked through the changes.

There are 2 places where `maybeToMonoid` can be used, but that is more a possibility than even a note.

Simple enough. Even mostly trivial.

Looks like you really stuck to HLint suggestions. Thanks. This allows for me to advance the structural `hlint` agenda (https://github.com/haskell/actions/pull/85) in the project. 

I love to contribute rules there & overall arrived at the same style that HLint suggests.

That together with preview reviews in #2519 & the successful `Ailrun` review, seems good to merge.

Anton-Latukha:
Thank you very much :gift: 

& Happy New Year's & Christmas  holidays :christmas_tree:

PR #2503, opened 2021-12-18
Build with GHC 9.2

pepeiborra:
Build a GHC 9.2 binary via head.hackage:

    cabal update --project-file=cabal-ghc921.project
    cabal build --project-file cabal-ghc921.project

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2503"></a>

PR #2526, opened 2021-12-23
cabal-ghc921.project: updated index-state

Anton-Latukha:


jneira:
hard to make the cache works without some of changes in .cabal config files in the branch
an alternative could be enable manual runs of the caching workflow in any branch, see #2528

Anton-Latukha:
Yes.

But as we discussed earlier updates to Hackage are still wanted.

Closing since: https://github.com/haskell/haskell-language-server/pull/2503#issuecomment-1000497645

Anton-Latukha:
Nope.

Would keep it to monitor the package dependency & configuration troubles on consistent code (if that would be possible), until GHC 9.2 support would solve into a particular index & be near merge.

pepeiborra:
> Nope.
> 
> Would keep it to monitor the package dependency & configuration troubles on consistent code (if that would be possible), until GHC 9.2 support would solve into a particular index & be near merge.

Thank you for your help @Anton-Latukha. Your last commit seems to have broken Cabal. I am going to revert it in order to unblock this PR.

What is it that you are trying to accomplish? Can it wait until we have finished with this PR? I don't think that the cache performance is a priority right now.

michaelpj:
This got bumped recently, I think this is superseded?

Anton-Latukha:
Yep.

PR #2532, opened 2021-12-23
CI: {caching,test,bench}: mk cache aware of package dep versions

Anton-Latukha:
1. CI caches 3rd patry dependency binaries.

2. That is why we do not want to trigger on all `.cabal` changes. We want to
cache only when changes in `.cabal` descriptions lead to changes in versions of
deps used, or in their compilation flags. `freeze` files:

For example the bits of the resulting `.freeze` file:

```
active-repositories: hackage.haskell.org:merge
constraints: any.Boolean ==0.2.4,
             ...
             any.HsYAML ==0.2.1.0,
             HsYAML -exe,
             any.HsYAML-aeson ==0.2.0.1,
             HsYAML-aeson -exe,
             any.JuicyPixels ==3.3.6,
             JuicyPixels -mmap,
             any.MemoTrie ==0.6.10,
             MemoTrie -examples,
             ...
             any.QuickCheck ==2.14.2,
             QuickCheck -old-random +templatehaskell,
             aeson -bytestring-builder -cffi -developer -fast,
             ...
             any.aeson-pretty ==0.8.9,
             aeson-pretty -lib-only,
             bifunctors +semigroups +tagged,
             any.blaze-textual ==0.2.2.1,
             blaze-textual -developer -integer-simple +native,
             ...
             any.brittany ==0.13.1.2,
             brittany -brittany-dev-lib -brittany-test-perf,
             ...
             any.clock ==0.8.2,
             clock -llvm,
             any.cmdargs ==0.10.21,
             cmdargs +quotation -testprog,
             ...
             any.comonad ==5.0.8,
             comonad +containers +distributive +indexed-traversable,
             ...
             any.hashtables ==1.2.4.2,
             hashtables -bounds-checking -debug -detailed-profiling -portable -sse42 +unsafe-tricks,
             haskell-language-server -alternatenumberformat +brittany +callhierarchy +class +eval +floskell +fourmolu +haddockcomments +hlint -ignore-plugins-ghc-bounds +importlens +modulename +ormolu -pedantic +pragmas +qualifyimportednames +refineimports -rename +retrie +splice +stylishhaskell +tactic,
             ...
             any.hlint ==3.2.7,
             hlint -ghc-lib +gpl -hsyaml +threaded,
             hls-eval-plugin -pedantic,
             hls-graph -embed-files -pedantic -stm-stats,
             hls-hlint-plugin -ghc-lib -hlint33 -pedantic,
             ...
             any.integer-logarithms ==1.0.3.1,
             integer-logarithms -check-bounds +integer-gmp,
             ...
             any.lens ==5.0.1,
             lens -benchmark-uniplate -dump-splices +inlining -j +test-hunit +test-properties +test-templates +trustworthy,
             ...
             any.scientific ==0.3.7.0,
             scientific -bytestring-builder -integer-simple,
             any.semigroupoids ==5.3.6,
             semigroupoids +comonad +containers +contravariant +distributive +tagged +unordered-containers,
             any.semigroups ==0.19.2,
             semigroups +binary +bytestring -bytestring-builder +containers +deepseq +hashable +tagged +template-haskell +text +transformers +unordered-containers,
             ...
             any.transformers-compat ==0.6.6,
             transformers-compat -five +five-three -four +generic-deriving +mtl -three -two,
             ...
             any.vector ==0.12.3.1,
             vector +boundschecks -internalchecks -unsafechecks -wall,
             any.vector-algorithms ==0.8.0.4,
             vector-algorithms +bench +boundschecks -internalchecks -llvm +properties -unsafechecks,
             ...
index-state: hackage.haskell.org 2021-11-29T12:28:17Z
```

You got the idea. `freeze` collects stuff for package compilation & all dependencies.

& this file satisfies the caching for dependencies agenda.

I also checked - it solves & collects all package configurations across all
subprojects in this one file.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2532"></a>

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -75,6 +75,7 @@ jobs:
         echo "INDEX_STATE=$INDEX_STATE1" >> $GITHUB_ENV
 
     - name: Form the package list ('cabal.project.freeze')
+      continue-on-error: true
       run: |
         cabal v2-freeze
 
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -97,7 +97,7 @@ jobs:
         path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
         key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}-${{ hashFiles('cabal.project.freeze') }}
         restore-keys: |
-              ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}
+              ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}-
               ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-
               ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
               ${{ env.cache-name }}-${{ runner.os }}-
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -78,6 +78,10 @@ jobs:
       continue-on-error: true
       run: |
         cabal v2-freeze
+        echo ''
+        echo 'Output:'
+        echo ''
+        cat 'cabal.project.freeze'
 
     - name: Hackage sources cache
       uses: actions/cache@v2
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -99,9 +99,8 @@ jobs:
         cache-name: compiled-deps
       with:
         path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
-        key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}-${{ hashFiles('cabal.project.freeze') }}
+        key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project.freeze') }}
         restore-keys: |
-              ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}-
               ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-
               ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
               ${{ env.cache-name }}-${{ runner.os }}-
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -74,6 +74,10 @@ jobs:
         INDEX_STATE1=$(echo "$INDEX_STATE_ENTRY" | cut -d' ' -f2 | tr ':' '-')
         echo "INDEX_STATE=$INDEX_STATE1" >> $GITHUB_ENV
 
+    - name: Form the package list ('cabal.project.freeze')
+      run: |
+        cabal v2-freeze
+
     - name: Hackage sources cache
       uses: actions/cache@v2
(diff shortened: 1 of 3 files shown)


Anton-Latukha:
Would note that caching `.cabal` - caches only 1 of (including some test samples) 53 `.cabal` files in monorepo (& has other implications).

Anton-Latukha:
So.

When freeze is not possible - it essentially does what it was doing before this PR. But if `freeze` is possible - it caches what `Cabal` chooses.

Anton-Latukha:
I've go another idea how to make caching even more fault tolerant. (So `caching` would be able to run & cache even inside #2503 environment, where dependency versions for all targets don't add up).

.github/workflows/bench.yml
```diff
@@ -74,6 +74,15 @@ jobs:
         INDEX_STATE1=$(echo "$INDEX_STATE_ENTRY" | cut -d' ' -f2 | tr ':' '-')
         echo "INDEX_STATE=$INDEX_STATE1" >> $GITHUB_ENV
 
+    - name: Form the package list ('cabal.project.freeze')
+      continue-on-error: true
+      run: |
+        cabal v2-freeze
+        echo ''
+        echo 'Output:'
```
jneira, line 82: We are not uploading artifacts for the workflow, but uploading those files could help to reproduce the ci build locally.
To do in another pr.

Anton-Latukha: So far I did not work with artifacts really.

I prefer to learn it separately & would ship it in the next PRs.

Anton-Latukha: On the other thought. They are saved nearby the logs. Logs are kept for ~7 days. It is really easy to copy paste it into the file.

& the `freeze` file really is just a representation of the `index-state`, `.cabal` & `.project` files - so the source state.

jneira: yeah, maybe a little bit more convenient or useful for scripting via the github api

.github/workflows/bench.yml
```diff
@@ -84,13 +93,15 @@ jobs:
...
 
     - name: Compiled deps cache
+      id: compiled-deps
       uses: actions/cache@v2
       env:
         cache-name: compiled-deps
       with:
         path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
-        key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}
+        key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project') }}-${{ hashFiles('cabal.project.freeze') }}
```
jneira, line 102: could we not only use the freeze file instead cabal.project + freeze file?

Anton-Latukha: We use `cabal.project` & `freeze` file. Look at the line 8)

I went & looked how people do it. & do 2 caches like this is more effective organization of cache namespaces & inheritance between them.

Since `project` file changes rarely.

.github/workflows/bench.yml
Anton-Latukha: Ok. Understood.

.github/workflows/bench.yml
jneira: I see, but all the build info which controls the cache is (or should be) in the freeze file, including the hackage index.
Also the freeze file is autogenerated, so it is stable.
If you do cosmetic changes or changes not affecting the cache at all in the cabal.project file you will invalidate the cache for no real reason.
I think the hackage index will give us a good enough base cache and all things that are in cabal.project and no in the freeze file only could throw away good caches. Otoh cabal files are not changed so frequently. And usually you need to update the hackage index to get the new dep version, so the cache is invalidated.

Anton-Latukha: Ok. Understood.

I read the https://github.com/haskell/haskell-language-server/pull/2532#discussion_r774771451 quite the other way first 8)

Anton-Latukha:
Added a workaround to GitHub CI.

.github/workflows/bench.yml
Anton-Latukha: Done.

Anton-Latukha: Oh. I remembered why I kept `project` - so there some symbols can be added to refresh into new keyspace. Otherwise, it is back to `index-state` changes. But after cache workflow stabilizes - it would be needed rarely.

jneira (approved):
LGTM, thanks!

PR #2535, opened 2021-12-23
CI: caching: keep-going

Anton-Latukha:
This solves the `build all` problem when all projects can't infer the consitent
package version.

Also would allow caching workflow on manual start - to work & cache in a broken
or complex PRs.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2535"></a>

### DIFF
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -45,7 +45,7 @@ on:
     - cron: "25 2/8 * * *"
 
 env:
-  cabalBuild: "v2-build all --enable-tests --enable-benchmarks"
+  cabalBuild: "v2-build all --enable-tests --enable-benchmarks --keep-going"
 
 jobs:
 


jneira (approved):
yeah we have to build as many deps as possible, thanks!

jneira:
Still a draft, merge it when you consider

Anton-Latukha:
Yes. @jneira, the main idea & strength is that PR checks (`test` & `bench`) are the checkgate of configuration consistency. & caching runs in kept consistent `master` after that. That is why we can weaken the build rules in caching to increase caching & consistency was already insured before that.

But inside the PRs this same property allows to build through failing parts of the project configuration like through butter & cache what is possible.

Anton-Latukha:
Well. Wanted to investigate how to manually run caching workflow. Long story short - it is only possible from the internal branch (wanted to do a PR to myself in the fork, but found that an additional requirement is that the main branch already has that workflow).

PR #2536, opened 2021-12-24
CI: caching: closer match work/CI guarantees

Anton-Latukha:
* Would note we have a `cabal v2-build all --enable-tests --enable-benchmark` problem - that command requires the level of dependency consistency guarantees that PR CI does not checks.

`test` workflow does essentially `cabal v2-build --enable-tests` on matrix by target.
`bench` workflow does `cabal v2-build --enable-benchmarks` by target.

`caching` used `all --enable-tests --enable-benchmarks` - requiring total consistency.

These changes are to make `caching` able to work when `test` & `bench` deps
versions are inconsistent (which is how CI & people use them - separately).

* "`Workaround to CI platform`" needs to be removed either way, since I found out post hooks does not run if any of (even marked `continue-on-error: true`) steps fail.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2536"></a>

### DIFF
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -201,10 +201,5 @@ jobs:
         continue-on-error: true
         # Done separately, matching the tested work/PR workflow guarantees
         run: |
-
-      # Despite the `continue-on-error: true` directive - CI does not ignore the return code of the last step
-      - name: Workaround to CI platform
-        run: |
-          true
           cabal $cabalBuild --enable-benchmarks || cabal $cabalBuild --enable-benchmarks || cabal $cabalBuild --enable-benchmarks
           cabal $cabalBuild --enable-test || cabal $cabalBuild --enable-test || cabal $cabalBuild --enable-test
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -206,6 +206,7 @@ jobs:
 
       - if: steps.compiled-deps.outputs.cache-hit != 'true' && runner.os == 'Linux' && matrix.ghc == '8.10.7'
         name: (For Bench workflow) Build benchmark targets
+        continue-on-error: true
         #  Downloaded separately, to match the tested work/PR workflow guarantees
         run: |
           cabal $cabalBuild --enable-benchmarks || cabal $cabalBuild --enable-benchmarks || cabal $cabalBuild --enable-benchmarks
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -45,7 +45,7 @@ on:
     - cron: "25 2/8 * * *"
 
 env:
-  cabalBuild: "v2-build all --enable-tests --enable-benchmarks --keep-going"
+  cabalBuild: "v2-build all --keep-going"
 
 jobs:
 
@@ -186,8 +186,10 @@ jobs:
 
       - if: steps.compiled-deps.outputs.cache-hit != 'true'
         name: Download all sources
+        #  Downloaded separately, to match the tested work/PR workflow guarantees
         run: |
-          cabal $cabalBuild --only-download
+          cabal $cabalBuild --enable-benchmarks --only-download
+          cabal $cabalBuild --enable-tests      --only-download
 
       # repeating builds to workaround segfaults in windows and ghc-8.8.4
       # This build agenda in not to have successful code,
@@ -197,10 +199,12 @@ jobs:
       - if: steps.compiled-deps.outputs.cache-hit != 'true'
         name: Build all targets; try 3 times
         continue-on-error: true
+        # Done separately, matching the tested work/PR workflow guarantees
         run: |
-          cabal $cabalBuild || cabal $cabalBuild || cabal $cabalBuild
 
       # Despite the `continue-on-error: true` directive - CI does not ignore the return code of the last step
       - name: Workaround to CI platform
         run: |
           true
+          cabal $cabalBuild --enable-benchmarks || cabal $cabalBuild --enable-benchmarks || cabal $cabalBuild --enable-benchmarks
+          cabal $cabalBuild --enable-test || cabal $cabalBuild --enable-test || cabal $cabalBuild --enable-test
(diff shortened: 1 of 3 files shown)


.github/workflows/caching.yml
```diff
@@ -165,25 +162,24 @@ jobs:
...
         name: Build all targets; try 3 times
         continue-on-error: true
         run: |
-          cabal $cabalBuild || cabal $cabalBuild || cabal $cabalBuild
-
-      # Despite the `continue-on-error: true` directive - CI does not ignore the return code of the last step
-      - name: Workaround to CI platform
-        run: |
-          true
+          cabal $cabalBuild || cabal $cabalBuild || cabal $cabalBuild || cabal v2-build all --enable-tests --keep-going || cabal v2-build all --enable-benchmarks --keep-going || cabal v2-build all --keep-going || cabal v2-build --enable-tests --enable-benchmarks --keep-going || cabal v2-build --enable-tests --keep-going || cabal v2-build --keep-going
```
jneira, line 185: Not sure about this. I think we should optimize workflows for the "happy" path, the stable master branch and pr's not tweaking too much the project dependencies. 
`ghc-9.2.1` branch is somewhat a extraordinary one and as pepe commented there, optimize caching for it is not the main goal, but make it work. Authors of commits are maintainers testing changes locally or in their forks and fast ci is desirable but no essential imo.

Anton-Latukha:
Note: "`Workaround to CI platform`" needs to be removed either way, because post hooks does not run if any of (even `continue-on-error: true`) steps fail.

.github/workflows/caching.yml
Anton-Latukha: Well, yes this is also true.

This would've been a nice thing to have simply because, since we have manual runs inside PRs - this would be logical to have to cache through difficulties in the PR.

While `||` also ensures that after success commands on the right do not run, CI in `master` would terminate on first 4-5 of them - which would create a good cache. & that is why (because it is free to be added and skipped over) I was free to make that long line of commands "just in case" it runs inside PRs.

But at this point I also have no interest, that is a minor thing, this PR has what to attend to: *(updated the head post)*

Anton-Latukha:
I reformulated the PR to address the most important stuff in it.

To match the process completely - we would need to run the `v2-build --enable-tests` command for every plugin separately, so it matches how `test` does it.

.github/workflows/caching.yml
Anton-Latukha: But whould note, this was also the way to follow the provided by `test` & `bench` guarantees, while not requiring maintenance of https://github.com/haskell/haskell-language-server/pull/2536#issuecomment-1001551417

Anton-Latukha:
[skip circleci]

.github/workflows/caching.yml
```diff
@@ -182,25 +184,35 @@ jobs:
...
       # This build agenda in not to have successful code,
       # but to cache what can be cached, so step is fault tolerant & would always succseed.
       # 2021-12-11: NOTE: Building all targets, since
       # current Cabal does not allow `all --enable-tests --enable-benchmarks --only-dependencies`
-      - if: steps.compiled-deps.outputs.cache-hit != 'true'
-        name: Build all targets; try 3 times
-        continue-on-error: true
+
+      - if: steps.compiled-deps.outputs.cache-hit != 'true' && runner.os == 'Linux' && matrix.ghc == '8.10.7'
+        name: (For Bench workflow) Build benchmark targets
```
jneira, line 208: Should not have this
```suggestion
        name: (For Bench workflow) Build benchmark targets
        continue-on-error: true
```
?

.github/workflows/caching.yml
```diff
@@ -182,25 +184,35 @@ jobs:
...
+        run: |
+          cabal $cabalBuild --enable-benchmarks --only-download
 
       - if: steps.compiled-deps.outputs.cache-hit != 'true'
-        name: Download all sources
+        name: Download the rest of the sources
+        #  Downloaded separately, to match the tested work/PR workflow guarantees
         run: |
-          cabal $cabalBuild --only-download
+          cabal $cabalBuild --enable-tests      --only-download
```
jneira, line 199: minor whitespace thing
```suggestion
          cabal $cabalBuild --enable-tests --only-download
```

.github/workflows/test.yml
```diff
@@ -177,7 +179,8 @@ jobs:
                 ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
                 ${{ env.cache-name }}-${{ runner.os }}-
 
-      - run: cabal v2-update
+      - run: |
+          cabal v2-update
```
jneira, line 183: not related with this pr but i've just was thinking if `cabal update` should not be before the freeze step, we have the hackage index fixed but freeze would change if the index change (only an observation, no asking for changes)

Anton-Latukha: `freeze` seems to save the real Hackage index used. So if `index-state` is pinned to something, the `freeze` infers the real snapshot time before that moment.

That's why it seems that the placement of `update` does not even matter really. It even probably may be removed.

But `update` would play an important role, if `haskell/actions/setup` for some reason has the older Hackage update than the `index-state` package uses. Currently *idk* if `haskell/actions/setup` does `cabal update`:

Screenshot-2021-12-27-17:06:32

so action reports nothing.

Currently, it does does the update: https://github.com/haskell/actions/blob/a64ac8cca0d4aec4e3fd4a5fbc9ae3c4953e20aa/setup/src/setup-haskell.ts#L75 (well it should update at every run).

jneira (approved):
lgtm, thanks!

Anton-Latukha:
Nice review.

Would belabor additionally: now CI would start to cache `benchmarks` only to the same configuration that is used in `bench`. That, on update of GHC in `bench` to update `caching. But in that way we both deduplicate that build-cache over several workflows & cache only used `benchmarks` dependencies.

Anton-Latukha:
After the recent merge `HLint` check works.

PR #2537, opened 2021-12-24
CI: add hlint workflow

Anton-Latukha:
As I understand, CI installs executable & runs it.

HLint gives the same output regardless of the GHC versions, so there is no reason to run it inside GHC matrix. In fact HLint executable installed does not need GHC at all.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2537"></a>

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -20,4 +20,5 @@ jobs:
     - name: 'Checking code'
       uses: rwe/actions-hlint-run@v1
       with:
-        path: '[ "--with-group=extra", "--hint=ghcide/.hlint.yaml", "ghcide/src", "ghcide/exe", "ghcide/bench/lib", "ghcide/bench/exe", "ghcide/bench/hist", "shake-bench/src", "ghcide/test/exe" ]'
+        hlint-bin: "hlint --with-group=extra --hint=ghcide/.hlint.yaml"
+        path: '[ "ghcide/src", "ghcide/exe", "ghcide/bench/lib", "ghcide/bench/exe", "ghcide/bench/hist", "shake-bench/src", "ghcide/test/exe" ]'
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -15,7 +15,7 @@ jobs:
     - name: 'Installing'
       uses: rwe/actions-hlint-setup@v1
       with:
-        version: '3.3.5'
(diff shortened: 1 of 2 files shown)


Anton-Latukha:
[skip circleci]

Anton-Latukha:
There is no hints currently, because all of them were cleared in https://github.com/haskell/haskell-language-server/pull/2523 just merged.

For example, after syncing GHC 9.2 PR to `master` - inside PR there is also: https://github.com/haskell/haskell-language-server/runs/4629194854?check_suite_focus=true no hints under the old way of running workflow.

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -205,7 +205,7 @@ atPoint
 atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ pointCommand hf pos hoverInfo
   where
     -- Hover info for values/data
-    hoverInfo ast = (Just range, prettyNames ++ pTypes)
+    hoverInfo ast = (Just range, prettyNames ++ [] ++ pTypes)
       where
```
Anton-Latukha, line 209: Here is a Lint warning:
Screenshot-2021-12-25-01:54:48

Anton-Latukha: Kept this open to show workflow works.

pepeiborra:
I like this. Will wait for @jneira to share his thoughts and approve. 

Also need to add the new workflow to the GitHub PR requirements.

Anton-Latukha:
Of course with availability to use different levels of messaging, the hints can be tuned.

Anton-Latukha:
Also, since I had a tiny thing to - added a tiny addition to `caching` workflow in b6dc520.

jneira (approved):
great, I like see this in another workflow, as it can be setup and run independently,
thanks!

jneira:
hlint workflow added to required checks to merge

Anton-Latukha:
We not see `warning` suggestions because #2523 covered all `warning` suggestions which current HLint setup monitors.

Workflow blocks on `error` severity just as before it was blocking inside `test`, but now it also presents `warning` level messages.

It is the same setup, but now information messages are not hidden inside build, but shown in review. Which motivates people to follow even soft linter suggestions.

PR #2538, opened 2021-12-25
CI: linting

Anton-Latukha:


Anton-Latukha:
[skip circleci]

jneira (approved):
Looks much cleaner this way, thanks!

Anton-Latukha:
I mainly do this for paths blocks, as they have advanced logic & so being able to read & parse them is useful.

Anton-Latukha:
Oh. CI can not merge. It is me.

I pushed & then dropped the last commits, & seems CI got puzzled, some workflows appeared twice.

Anton-Latukha:
Hmm...

Anton-Latukha:
@jneira maybe you know what this means:
Screenshot-2021-12-26-00:39:45

HLint workflow passes, but there is a branch protection check, maybe it does not catch the HLint (or HLint workflow needs something) or something?

Anton-Latukha:
I had to add hlint_post_job or something?

Anton-Latukha:
The same is in: https://github.com/haskell/haskell-language-server/pull/2527

Ailrun:
The branch protection rule is global so the same rule applies to all PRs. It should work without any extra "post" jobs, so it seems strange that CI doesn't check the passed job... Let's disable the requirement for now.

jneira:
I think it is working now, will enable again

jneira:
> I think it is working now, will enable again

nope there are 2 checks, and GH does not know they are the same

Anton-Latukha:
@Ailrun 

Thanks.

I did workflows the default way & not familiar with what's happening there & those processes.

Issue #708, opened 2020-12-29
Can not go to definition of  non local libraries (i.e. dependencies)


i-am-the-slime:
Is this because it is downloaded in some binary form?

jneira:
The new pull request that will add hiedb support in hls is #704

wz1000:
This will not be addressed in #704 but in an upcoming PR.

jneira:
@wz1000 hi! had you the chance to prepare that pr finally?

jneira:
There is a closely related ghc proposal: https://discourse.haskell.org/t/extended-dependency-generation-support/2811
Not sure if it is necessary to implement the feature (as there is a wip since quite time ago)

jneira:
@pepeiborra @wz1000 it would be great to have some idea about the plans for this, as there are lot of people eager to see this land, thanks!

wz1000:
I had a somewhat working implementation of this, but there were some issues. In principle, most of what needs to be done is provide an implementation for `lookupMod` [here](https://github.com/haskell/ghcide/pull/898/commits/d17cf3117c944abc7ddcffaf738bf3bc6cf624e8#diff-58f4f7994c1bd5f37a68b33388ee77fcc49d32bf1dba7dc75d35ff9e3d0c4f9dL134) - possibly taking inspiration from [this commit](https://github.com/haskell/ghcide/pull/898/commits/d17cf3117c944abc7ddcffaf738bf3bc6cf624e8) which reverted support for this feature from my WIP branch (there were some unresolved problems with the feature and I decided to focus on getting the rest of the branch merged first).

The feature as before it was reverted by that commit worked in the following fashion

- The user to build their dependencies with `-fwrite-ide-info` and then index the resulting `.hie` files with `ghcide`/`HLS`. This can be done by `cd`ing to the project root directory and calling `ghcide index dir/with/hie/files`.
- Once the dependencies are indexed into the database, when we get a `go to definition` request for something in an indexed module, we check if we have a source file registered for the module in the database. If so, we can just return that location. (this is still the code path as implemented today)
  
  Otherwise `lookupMod` is called. Currently `lookupMod` always fails. To implement the non-local definition feature, `lookupMod` needs to get the contents of the module from the indexed `.hie` files and then write out a file somewhere onto the filesystem with those contents. It then needs to register this file in the database as the source file for this module (so that future lookups would directly jump to the file we wrote). Finally, it returns the location of this file so the client can jump to it.

[...]

PR #2543, opened 2021-12-26
Fix ghcide handling project root

drsooch:
~NOTE: The two commits in this PR are distinct options for handling this issue.~

This PR provides an alternate way to grab the project root/current working directory.

Prior to this PR the relative filepath "." was hard-coded for both
Db and Custom Commands. This results in inconsistent behavior with how
HLS derives it's hiedb location. 

This new argument to the internal Ghcide Arguments, maps from the
executable Arguments `argsCwd` or by grabbing the current working
directory. If the user provides an option to `--cwd` we need to make
sure we make that filepath absolute.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2543"></a>


pepeiborra:
> This PR provides an alternate way to grab the project root/current working directory.
> Prior to this PR the relative filepath "." was hard-coded for both Db and Custom Commands. This results in inconsistent behavior with how HLS derives it's hiedb location.

For my benefit, could you explain the inconsistency or share an example of what could go wrong?

ghcide/src/Development/IDE/Main.hs
```diff
@@ -122,12 +122,12 @@ import           Text.Printf                           (printf)
...
 data Command
     = Check [FilePath]  -- ^ Typecheck some paths and print diagnostics. Exit code is the number of failures
-    | Db {projectRoot :: FilePath, hieOptions ::  HieDb.Options, hieCommand :: HieDb.Command}
+    | Db {hieOptions ::  HieDb.Options, hieCommand :: HieDb.Command}
      -- ^ Run a command in the hiedb
     | LSP   -- ^ Run the LSP server
     | PrintExtensionSchema
     | PrintDefaultConfig
-    | Custom {projectRoot :: FilePath, ideCommand :: IdeCommand IdeState} -- ^ User defined
+    | Custom {ideCommand :: IdeCommand IdeState} -- ^ User defined
```
pepeiborra, line 130: Thanks for removing the superfluous `projectRoot` field here

ghcide/src/Development/IDE/Main.hs
```diff
@@ -161,13 +161,14 @@ commandP plugins =
...
     pluginCommands = mconcat
-        [ command (T.unpack pId) (Custom "." <$> p)
+        [ command (T.unpack pId) (Custom <$> p)
         | (PluginId pId, PluginDescriptor{pluginCli = Just p}) <- ipMap plugins
         ]
 
 
 data Arguments = Arguments
-    { argsOTMemoryProfiling     :: Bool
+    { argsProjectRoot           :: IO FilePath
```
pepeiborra, line 170: I prefer the first version with the `Maybe FilePath` type here. The second version breaks `lsp --cwd` since `argsProjectRoot` is not invoked in the `LSP` command.

drsooch:
> > This PR provides an alternate way to grab the project root/current working directory.
> > Prior to this PR the relative filepath "." was hard-coded for both Db and Custom Commands. This results in inconsistent behavior with how HLS derives it's hiedb location.
> 
> For my benefit, could you explain the inconsistency or share an example of what could go wrong?

Yes, I was just taking a look at the state of go-to documentation for external dependencies (https://github.com/haskell/haskell-language-server/issues/708). I was attempting to follow what @wz1000 had done, and noticed that when `ghcide hiedb index ...` is run it was loading the wrong project hiedb. 

This effectively nullifies the point of indexing, as HLS will never pick up the indexed files. The actual call to `getHieDbLoc` is always done with `"."` which will hash to the same location each time (not the current project root). So it doesn't really effect anything that I'm aware but if anyone uses ghcide outlined by @wz1000 it should work as expected (sort of).

ghcide/src/Development/IDE/Main.hs
drsooch: Lol... good point. I preferred the first way, but since we should always reach getHieDbLoc with a FilePath I tried to alleviate the problem. I'm good with reverting back.

wz1000 (approved):
Thanks!

pepeiborra (changes requested):
Please fix `lsp --cwd` before merging. It would be great to get tests as well in a later PR

ghcide/exe/Main.hs
```diff
@@ -50,13 +50,18 @@ main = withTelemetryLogger $ \telemetryLogger -> do
     if argsVersion then ghcideVersion >>= putStrLn >> exitSuccess
     else hPutStrLn stderr {- see WARNING above -} =<< ghcideVersion
 
-    whenJust argsCwd IO.setCurrentDirectory
+    -- getHieDbLoc takes a directory path (the project root) and hashes it to find the location of the hiedb
+    -- when running commands directly from GHCIDE we need to provide the ABSOLUTE path to the project root (that's what HLS uses)
+    argsCwd <-case argsCwd of
+      Nothing   -> IO.getCurrentDirectory
+      Just root -> IO.setCurrentDirectory root >> IO.getCurrentDirectory
```
michaelpj: This is a little obscure: why not just call `makeAbsolute` on `root`? This relies on you knowing that `getCurrentDirectory` returns an absolute path.

drsooch: You're right and I originally did it that way, but I wanted to match the semantics found throughout out the codebase. For example, in `Development.IDE.Main.defaultMain` all of the `argCommands` use the same `maybe IO.getCurrentDirectory return rootPath` expression to get the project root.

drsooch:
@pepeiborra just curious where does `LSP.resRootPath` get set? If I'm following the code correctly we should have access to the new `argsProjectRoot` argument in `runLanguageServer`, and it should (?) get mapped to the `LSP.LanguageContextEnv` in some way.

drsooch:
Also just as an aside would it be worth exploring using `Typed Paths`? I know there are a few packages out there to use, this way it's possible to remove all confusion as to what state we are in?

pepeiborra:
> @pepeiborra just curious where does `LSP.resRootPath` get set? If I'm following the code correctly we should have access to the new `argsProjectRoot` argument in `runLanguageServer`, and it should (?) get mapped to the `LSP.LanguageContextEnv` in some way.

I think it's (optionally) set by the LSP [`initialize`](https://microsoft.github.io/language-server-protocol/specification#initialize) message.

pepeiborra (approved):

Anton-Latukha:
Is this ready to be merged?

Anton-Latukha:
Thanks, @drsooch for the work.

PR #2544, opened 2021-12-27
Set an unique name for Hlint job

jneira:
As we have to add the *job* to branch pr merge checks and `Run` is too generic

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2544"></a>

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build10:
-    name: "Run"
+    name: "Hlint check run"
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


jneira:
I ve added the `Run` job just in case, will change when this will be merged

.github/workflows/hlint.yml
```diff
@@ -21,4 +21,5 @@ jobs:
       uses: rwe/actions-hlint-run@v1
       with:
         hlint-bin: "hlint --with-group=extra --hint=ghcide/.hlint.yaml"
-        path: '[ "ghcide/src", "ghcide/exe", "ghcide/bench/lib", "ghcide/bench/exe", "ghcide/bench/hist", "shake-bench/src", "ghcide/test/exe" ]'
+        path: '[ "ghcide/src", "ghcide/exe", "ghcide/bench/lib", "ghcide/bench/exe", "ghcide/bench/hist", "shake-bench/src", "ghcide/test/exe" pop]'
```
Anton-Latukha, line 24: `pop` seems to be pu like a typo?

Anton-Latukha, line 24: `"ghcide/test/exe" pop` seems to be put as a typo?

Anton-Latukha (changes requested):

.github/workflows/hlint.yml
jneira: oh yeah, good catch
```suggestion
        path: '[ "ghcide/src", "ghcide/exe", "ghcide/bench/lib", "ghcide/bench/exe", "ghcide/bench/hist", "shake-bench/src", "ghcide/test/exe"]'
```

Anton-Latukha (approved):

Issue #2541, opened 2021-12-26
Please, support GHC 9.0.2


Anton-Latukha:
This report is a notification & tracking report for https://www.haskell.org/ghc/blog/20211225-ghc-9.0.2-released.html support

Released on `2021-12-25`:

Complete release notes: https://downloads.haskell.org/ghc/9.0.2/docs/html/users_guide/9.0.2-notes.html

---

To have a glance for HLS: quoting

<details><summary>GHC 9.0.2 summary release notes</summary>

</p>

GHC 9.0.2 adds first-class AArch64/Darwin support using the LLVM backend, as well as fixing a number of critical correctness bugs with the 9.0.1 release, along with numerous improvements to compiler performance and memory usage:
  * [ ] First-class support for Apple M1 hardware using GHC’s LLVM ARM backend.
  * [ ] Fix a number of incorrect rewrite rules for `fromInteger` ([GHC#19345](https://gitlab.haskell.org/ghc/ghc/-/issues/19345), [GHC#20066](https://gitlab.haskell.org/ghc/ghc/-/issues/20066)).
  * [ ] Fix a number of runtime system and GC bugs which typically manifested in crashes ([GHC#19417](https://gitlab.haskell.org/ghc/ghc/-/issues/19417), [GHC#19297](https://gitlab.haskell.org/ghc/ghc/-/issues/19297), [GHC#19645](https://gitlab.haskell.org/ghc/ghc/-/issues/19645), [GHC#19667](https://gitlab.haskell.org/ghc/ghc/-/issues/19667), [GHC#18033](https://gitlab.haskell.org/ghc/ghc/-/issues/18033), [GHC#20132](https://gitlab.haskell.org/ghc/ghc/-/issues/20132), [GHC#19715](https://gitlab.haskell.org/ghc/ghc/-/issues/19715), [GHC#20399](https://gitlab.haskell.org/ghc/ghc/-/issues/20399), [GHC#20414](https://gitlab.haskell.org/ghc/ghc/-/issues/20414), [GHC#20649](https://gitlab.haskell.org/ghc/ghc/-/issues/20649), [GHC#19147](https://gitlab.haskell.org/ghc/ghc/-/issues/19147), [GHC#20051](https://gitlab.haskell.org/ghc/ghc/-/issues/20051), [GHC#20093](https://gitlab.haskell.org/ghc/ghc/-/issues/20093))
[...]

PR #2548, opened 2021-12-27
doc: contributing: add notes on adding new GHCs

### LINKED ISSUE #2545: Please, document a workflow on the support of new GHCs
Support of new versions is a constant periodic task.

Denoting main steps/phases would allow for reports as #2541 & for maintainers & users to be more context-aware of the process. So the tracking reports created would have a more stable form & the process would be easier to track.

Anton-Latukha:
* Less people would bother with questions on new GHCs support
* More people would know the workflow of HLS to provide support for new GHCs, so would be able themselves to find the state of progress & jump-in to contribute.
 * Reports #2541 would be able to address the workflow better.

Closes #2545.


Anton-Latukha:
[skip circleci]

docs/supported-versions.md
```diff
@@ -25,6 +26,26 @@ GHC versions not in the list have never been supported by HLS, or are not planne
 
 The policy for when we deprecate support for versions of GHC is given below. The table reflects that, but we may decide to deviate from it for good reasons.
 
+### Support for the new GHC version
```
michaelpj, line 29: I think this belongs in Contributing. The intended audience of this page is users, and this isn't of interest to users, but rather to contributors who would go through this process.

Anton-Latukha: *idk*, the deprecation process described in this file, & the versioning information is here, so both support versions & deprecation of versions in this file, I thought it only logical to add an introduction of the new version here also.

I agree that I formulate it & present it as being useful for contributors. But that is one of agendas. (I imagine hypothetically) the majority of people who would search "why new version is not yet supported" - are users & hopefully would be directed to doc & would read the doc & majority of them would be satisfied with the answer that "it is complex, we working on it" & factually would not contribute. And that is also why doc is formulated to encourage taking part in the activity.

michaelpj: The deprecation policy is in here because we expect users to want to know "why is version X of GHC (which I care about) not supported".

> the majority of people who would search "why new version is not yet supported" - are users & hopefully would be directed to doc & would read the doc & majority of them would be satisfied with the answer that "it is complex, we working on it"

That's not what this piece of documentation conveys, though! I think it *would* make sense to have a section here describing the basic policy for when new versions are included (e.g. "upstream tools and packages must support it, and some compatibility work must be done, here are links to the tracking tickets"), but a detailed checklist of steps belongs in the contributing documentation IMO.

jneira: maybe the main content could go to contributing and link it here?

Issue #2422, opened 2021-11-30
GitHub CI: Deduplicate workflow configuration


Anton-Latukha:
After adding https://github.com/haskell/haskell-language-server/pull/2419, workflow configs, platform crutches & matrix'es became shotgun-surgery enough to think about doing the https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#overview work to duplicate in workflows as much as possible.

Anton-Latukha:
At the same time it would be useful to look into: does those `$CABAL*` env variables even need to be set, or the:
```yml
      - name: "Haskell env setup"
        id: HaskEnvSetup
        uses: haskell/actions/setup@v1
        with:
          ghc-version: ${{ matrix.ghc }}
          
### ...
      - name: "Configuring GitHub cache"
        uses: actions/cache@v2
        with:
          path: |
            ${{ steps.HaskEnvSetup.outputs.cabal-store }}
```
is enough (seems enough).

jneira:
Thanks for opening the issue, reusing workflows looks promising

Anton-Latukha:
First thing to do with reusable workflows - is to have a workflow for `cabal.project` processing. The file basically can be passed even as a value. But it is useful to do it through artifact passing.

Anton-Latukha:
The first thing to do with reusable workflows - is to start with having a workflow for `cabal.project` processing. The file basically can be passed even as a value. But it is useful to do it through artifact passing.

Anton-Latukha:
I looked into what GitHub has & it is ...

I basically wanted/imagined simple things:
  * as loading & sharing the steps or jobs (for example loading them from `json`).
  * macros are impossible.

Also GitHub CI allows only 1 level of inheritance (does not allow to inherit workflow that inherited anything).

So ... deduplication is limited. Best I can do initially is still make files seamless & then look at how to do https://github.com/haskell/haskell-language-server/issues/2422#issuecomment-984174843

jneira:
What about custom actions? could we extract some reusable bits and make it actions (like our use of skip actions)?

Anton-Latukha:
Custom actions are also partial solutions & not flawless.

There are JS, `sh` & `docker` actions, each has its pros & cons. I know that now GitHub allows caching inside actions, but do not know all limitations of that (which are frequent in the CIs designs & get learned during working on them).

Anton-Latukha:
I've looked into DHall & its alternatives.

[...]

PR #2549, opened 2021-12-27
Restore TemplateHaskell pragma in hls-graph

pepeiborra:
The pragma is required for the `embed-files` flag  and was removed recently in #2523.

@jneira @Anton-Latukha the PR is open for any ideas that extend CI to check the build with this flag.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2549"></a>


Anton-Latukha:
I obviously have limited knowledge of the project.

Since I'm already sitting - would contribute to the patch.

The patch may uncover a bit of work.

Anton-Latukha:
Oh. I thought we in CI do not test pre `ghcide` sub-project & so though to add a test-suites into the process.

But:
```
🅸 ~/s/h/c/haskell-language-server:fix-hls-graph-build⨯>cabal test shake-bench
No tests to run for the package shake-bench-0.1.0.2
🅸 ~/s/h/c/haskell-language-server:fix-hls-graph-build⨯>cabal test hls-graph
No tests to run for the package hls-graph-1.5.1.1
🅸 ~/s/h/c/haskell-language-server:fix-hls-graph-build⨯>cabal test hie-compat
No tests to run for the package hie-compat-0.2.1.0
🅸 ~/s/h/c/haskell-language-server:fix-hls-graph-build⨯>cabal test hls-plugin-api
No tests to run for the package hls-plugin-api-1.2.0.2
🅸 ~/s/h/c/haskell-language-server:fix-hls-graph-build⨯>cabal test hls-test-utils
No tests to run for the package hls-test-utils-1.1.0.2
🅸 ~/s/h/c/haskell-language-server:fix-hls-graph-build⨯>
```

There are no test-suites, no benchmarks for the.
So CI there is not of any help.

pepeiborra:
Sorry I wasn't clear - we want to test that hls-graph builds with the  embed-files flag

jneira:
well the unique thing I can think off without too much changes is add a step in the testing workflow building the package with the non default value for the flag
the package is not big and will not take too much time I hope

michaelpj:
Possibly dumb question: how bad would it be to just use `file-embed` always?

Anton-Latukha:
[skip circleci]

Anton-Latukha:
`michaelpj` raised a sound question.

Anton-Latukha:
The request/PR shows that there is a broad structurally important question. There is a number of subprojects & they have a number of flags. & CI does not tests any of them.

1. We can not test all flag combinations.
2. We can not test flag combinations * platform * GHC matrix.
3. That is why decided to create a workflow: 1) which tests for "default" case, under default currently took Linux & recommended GHC version 2) which builds one combination where all important flags are maximally enabled & so have some cross-testing.

pepeiborra:
> Possibly dumb question: how bad would it be to just use `file-embed` always?

It introduces a dependency on TH, which makes HLS-graph less portable. Which probably doesn't matter much.

.github/workflows/caching.yml
```diff
@@ -151,7 +151,7 @@ jobs:
           echo '' && \
           cat 'cabal.project.freeze' && \
           echo '' || \
-          echo 'WARNING: Could not produce the `freeze`.
```
Anton-Latukha, line 154: whoopsie daisy

.github/workflows/caching.yml
```diff
@@ -151,7 +151,7 @@ jobs:
...
+          echo 'WARNING: Could not produce the `freeze`.'
```
Anton-Latukha, line 154: Whoopsie daisy

Anton-Latukha:
The details (early termination rules) we can tweak in other PRs.

I consider this ready to be reviewed & merged.

Anton-Latukha:
I would start CI configuration consolidation effort at https://github.com/haskell/haskell-language-server/issues/2422#issuecomment-1002279380.

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+          cache-name: compiled-deps
+        with:
+          path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
+          key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project.freeze') }}
+          restore-keys: |
+                ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-
+                ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
+                ${{ env.cache-name }}-${{ runner.os }}-
+
+      - if: needs.pre_job.outputs.should_skip != 'true'
```
jneira, line 134: please add a `cabal update` here too, with the same comment as the other workflows if possible

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+        name: Build `hls-plugin-api` with flags
+        run: cabal v2-build hls-plugin-api --flags="pedantic"
+
+      - if: needs.pre_job.outputs.should_skip != 'true'
+        name: Build `hls-test-utils` with flags
+        run: cabal v2-build hls-test-utils --flags="pedantic"
+
+      # repeating builds to workaround segfaults in windows and ghc-8.8.4
+      - if: needs.pre_job.outputs.should_skip_ghcide != 'true'
```
jneira, line 151: we are only in ubuntu here, so not need for three builds

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+                         , "install/**"
+                         , "**.nix"
+                         , "flake.lock"
+                         , "**/README.md"
+                         , "FUNDING.yml"
+                         , ".circleci/**"
+                         , "**/stack*.yaml"
+                         ]'
+      # If we only change ghcide downstream packages we have not test ghcide itself
+      - id: skip_ghcide_check
```
jneira, line 40: afaics we are not using this condition in the workflow

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+        with:
+          path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
+          key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project.freeze') }}
+          restore-keys: |
+                ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-
+                ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-
+                ${{ env.cache-name }}-${{ runner.os }}-
+
+      - if: needs.pre_job.outputs.should_skip != 'true'
+        name: Build `hls-graph` with flags
```
jneira, line 135:
```suggestion
      - name: Build `hls-graph` with flags
```

the job itself is skipped if the condition holds

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: cabal v2-build hls-graph --flags="pedantic embed-files stm-stats"
+
+      - if: needs.pre_job.outputs.should_skip != 'true'
+        name: Build `hie-compat` with flags
```
jneira, line 139:
```suggestion
      - name: Build `hie-compat` with flags
```

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: cabal v2-build hie-compat --flags="ghc-lib"
+
+      - if: needs.pre_job.outputs.should_skip != 'true'
+        name: Build `hls-plugin-api` with flags
```
jneira, line 143:
```suggestion
      - name: Build `hls-plugin-api` with flags
```

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: cabal v2-build hls-plugin-api --flags="pedantic"
+
+      - if: needs.pre_job.outputs.should_skip != 'true'
+        name: Build `hls-test-utils` with flags
```
jneira, line 147:
```suggestion
      - name: Build `hls-test-utils` with flags
```

.github/workflows/flags.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: cabal v2-build hls-test-utils --flags="pedantic"
+
+      # repeating builds to workaround segfaults in windows and ghc-8.8.4
+      - if: needs.pre_job.outputs.should_skip_ghcide != 'true'
+        name: Build
```
jneira, line 152:
```suggestion
      - name: Build
```

jneira (changes requested):

.github/workflows/flags.yml
Anton-Latukha: Yes.

I still do this with deduplication in mind (doing deduplication locally).

Anton-Latukha:
This state confuses me.

Seems like all requested changes were addressed & the PR is still in the `change requested` mode.

I rechecked a couple of times & can not tell what is the cause.

jneira:
if a revision requests changes the revision author (in this case me) has to approve manually the pr to make it mergeable

jneira (approved):
To have a new workflow is nice for correctness and completeness but we have too much workflow filed so I would try to deduplicate workflows before adding more

Anton-Latukha:
The difficulty with computation would start where there would no be new build workers for PRs, so far it looks like CI have more then enough build workers, so far I haven't seen waits on resources.

We had `caching` workflow added, which runs async and has short-circuiting. `caching` saves computational time overall.

HLint workflow runs one worker & takes ~1m to run.

Flags runs one worker ~5m.

Testing takes 15 workers, takes 3-4 of them ~90m & Benchmark takes 3 workers, 2 workers for ~95m.

So we only have a code complexity & shotgun surgery because of CI design (but most CIs have alike yaml-based design). And looking at DHall results (but I would still work on it a bit before reporting) - the current shotgun surgery situation in CI starts to look quite good.

Issue #2542, opened 2021-12-26
Operator Documentation Links and All Source Links don't work


rvLMdE7:
Apologies if this issue has already been filed and/or fixed; I did a quick search through all issues/PRs and couldn't find anything relating to this.

### Your environment

I suspect that this issue would be independent of the setup, but just in case, I experienced this issue on Manjaro (Linux) using VSCode (OSS version) with the *Haskell* extension. I can give more details if required.

I tested this on a fresh project (meaning, I created a new dummy project just for testing) with

- GHC 8.10.7 and `haskell-language-server-8.10.7~1.5.1`, git hash 745ef26f406dbdd5e4a538585f8519af9f1ccb09 to be specific
- I also tested using GHC-9.0.1 with `haskell-language-server-9.0.1~1.5.1`, same git hash

In both cases I go the same problem. I'll continue describing the problem as I experienced it with ghc 9.0.1.

### Steps to reproduce

Open any project with HLS enabled. Find an operator, let's say `=<< :: Monad m => (a -> m b) -> m a -> m b` for example's sake.[^1] Hover over it, to show the information about it including a 'Documentation' and 'Source' link, and then try to use those links to examine the documentation / source code for `=<<`.

### Expected behaviour

The *Documentation* and *Source* links should work.

### Actual behaviour

They do not work. The links seem to be in a different format then haddock expects.

- The documentation link takes me to the right module, but it doesn't take to the part of the page where the operator is actually defined.
- The source link doesn't seem to work at all.

### Include debug information

I'm not familiar with the specifics on how haddock expects links to be formatted, but playing around with the links seems to suggest what the problem is (and what the solution would be):

#### Documentation link

- what HLS produced: https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Base.html#v:=%3C%3C
- what the correct link seems to be: https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Base.html#v:-61--60--60-

The links are the same up until the part at the end, `-Base.html#v:...`. Instead of ending the link with `...#v:=<<` and doing URL encoding, we need to escape the name `=<<` according to a different scheme: escape each character in its' name according to the unicode code point value of each character, and surround it on the left and the right with a `-` character. That seems to be what haddock expects.

[...]

Issue #2508, opened 2021-12-19
Clicking "Source" to open haddock source on hackage opens invalid URL


jhrcek:
### Your environment

Which OS do you use:
Linux, fedora 34

Which lsp-client do you use:
Vscode
Describe your project (alternative: link to the project):
stack based project

### Steps to reproduce
See steps gif:
![open-source-incorrect](https://user-images.githubusercontent.com/2716069/146671879-daf9d789-38da-4fff-8024-8022ff36e74f.gif)

### Expected behaviour
Correct page is opened on hackage
https://hackage.haskell.org/package/base-4.14.3.0/docs/src/Data-Foldable.html#concatMap

### Actual behaviour
Invalid URL is opened on hackage
https://hackage.haskell.org/package/base-4.14.3.0/docs/src/Data.Foldable.html#concatMap
`                                                                                      ^ It should be Data-Foldable, not Data.Foldable

### Include debug information
Execute in the root of your project the command `haskell-language-server-wrapper --debug .` and paste the logs here:
(if you are using the vscode extension check the executable location [here](https://github.com/haskell/vscode-haskell#downloaded-binaries))
<details>
<summary>
Debug output:
</summary>

```
<paste your logs here>
```
</details>

Paste the logs from the lsp-client, you can check instructions about for VS Code [here](https://github.com/haskell/vscode-haskell#troubleshooting)

<details>
<summary>
LSP logs:
</summary>

```
<paste your logs here>
```
</details>

jhrcek:
This seems to be problem in vscode-haskell, which is replacing `-`s with `.`s here:
https://github.com/haskell/vscode-haskell/blob/7eb6a485e1e443c2eea195f0afe37e6b9963c22c/src/docsBrowser.ts#L160-L163

jhrcek:
Looking around, I found that the **dots in haddock source documentation** are most likely result of this change in haddock: https://github.com/haskell/haddock/pull/977/files#diff-1ba3997884082b72ff50a289d8b53fa9df6948198e8494ffe7ffdf0044c19223R38

It seems to me that those dots started to appear as unintended consequence of changes in that PR. I'd like to check this with the author of that PR, but I can't tag him here. Anyone knows of a good person to ask?

jneira:
Yeah, thanks for the report and the further investigation, moving to the extension repo

PR #2550, opened 2021-12-27
fix haddock documentation links (#2542)

lawbel:
This doesn't completely resolve the linked issue (#2542), but it does seem to solve the documentation link problems. I've looked through the haddock codebase and tried to mirror the way that it creates links/anchors, and it looks to be working. In the process of doing this I realized that there is another issue with the current links - they don't work for function names like `foldl'` that have apostrophes in them. This commit also fixes that problem.

Getting the source links to behave is proving more difficult. I can't figure out how the module names are supposed to be formatted. I had thought, when writing the issue, that they should always be separated with dashes (like `GHC-Base.html`). But it seems that sometimes dots are needed (like `GHC.Base.html`). I'm at a loss trying to understand it. Any help would be welcome here.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2550"></a>


jneira:
Many thanks for the pr and the fixes

> Getting the source links to behave is proving more difficult. I can't figure out how the module names are supposed to be formatted. I had thought, when writing the issue, that they should always be separated with dashes (like GHC-Base.html). But it seems that sometimes dots are needed (like GHC.Base.html). I'm at a loss trying to understand it. Any help would be welcome here.

There is an issue in haddock about: https://github.com/haskell/haddock/issues/1447, opened from the hls issue #2508

jneira:
It would be great to add a regression test to check haddock doc links are not going to be broken. The test suite for hover is here:

https://github.com/haskell/haskell-language-server/blob/3de244e25ac6f93ea960805a8efb0bc3b8709fc6/ghcide/test/exe/Main.hs#L3867

feel free to ask any question about

rvLMdE7:
Thanks for the pointer, I will do that and try to write some suitable tests. Should I do that in a separate PR?

From reading those links about the source links, it seems that the issue I was experiencing with them is not the fault of HLS, but rather some combination of haddock changes and the vscode HLS extension?

jneira:
> Thanks for the pointer, I will do that and try to write some suitable tests. Should I do that in a separate PR?

AS you consider, both options are fine

> From reading those links about the source links, it seems that the issue I was experiencing with them is not the fault of HLS, but rather some combination of haddock changes and the vscode HLS extension?

yeah, a combination of, i think haddock is inconsistent in its url generation, hls follow one schema and the vscode tried to fix it standalone 🤦

rvLMdE7:
Okay, I have added a couple tests to check that names in the documentation links are getting encoded properly. If that looks good to you, then I think we could close the original issue, as the business with source links already has open issues in the appropriate places?

michaelpj:
Looks like one of the ghcide tests fails now?

rvLMdE7:
Yeah, at first glance it looks like some (older?) versions of GHC and/or HLS don't generate the Documentation links, which is making the tests fail. I'll see if I can add some conditionals to just run the tests for some versions.

rvLMdE7:
Sorry for the delay, it was a bit of a slog to look through the logs. It seems like many of the builds got timed out, and others skipped the `ghcide` tests? Looking at the builds that got far enough to run the two tests I added (which only happened for `ubuntu-latest`), the results are as follows:

ghc    | def operator | def function | hover operator | hover function
-------|--------------|--------------|----------------|---------------
9.0.1  | ok           | ok           | fail           | fail
8.10.7 | ok           | ok           | ok             | ok
8.8.4  | ok           | ok           | ok             | ok
8.6.5  | ok           | ok           | fail           | ok

I admit to being completely lost as to why the `def` cases are all passing, but a selection of the `hover` cases are failing.

Looking at the results, it would seem sensible to keep the def tests as-is, and just skip over all the hover tests. I've pushed a commit that does exactly that, so we'll see whether the tests pass this time or not.

michaelpj:
Possibly dumb question: could we actually depend on `haddock` and pull in the functions that do this stuff?

rvLMdE7:
Perhaps that would be a better approach. I'm not very familiar with the internals of haddock, or HLS for that matter, so I just tried to hack something together that works. I suppose I had been implicitly assuming that the way haddock does this is relatively stable, i.e. it's not something that's likely to change often if at all.

The function that I added called `makeDocAnchor` is a minor adaptation of `Haddock.Utils.makeAnchorId` from `haddock-api`, link [here](https://github.com/haskell/haddock/blob/ghc-9.2/haddock-api/src/Haddock/Utils.hs#L194) for the source code on github. I think it is the *exact* function that creates the documentation links in the first place. (Though I might be mistaken, there's a lot of code in haddock, I may have missed something.) It is *not* exported as part of the public API of `haddock-api` though... I'm not sure whether equivalent functionality is exposed through some of the other parts that are exported.

michaelpj:
I'm not so much worried about Haddock _changing_ how it does it, as there just being more awkward corners we hadn't thought about. If they don't expose it, though, nothing we can do. Maybe open an issue in `haddock` recording that it would be useful to expose functions for creating links?

michaelpj:
Also, the tests didn't time out this time. Are you planning to investigate what's up with the disabled tests or just leave it for now? At least we should include a comment as to why they're disabled if we do that.

pepeiborra:
Sad that this PR was closed without merging. Reopening in case someone can resubmit it, or at least salvage the tests

fendor:
this was fixed by #4746

PR #2551, opened 2021-12-27
Add support for brittany and floskell with ghc-9.0.1

jneira:
* It requires support for aeson-2 for ghc >= 9.0.1 but we have to keep aeson < 2.0 for older ghcs. It is done through a Compat module in hls-plugin-api
* It requires #2494 as the new version of lsp libs supports aeson-2.0


ghcide/src/Development/IDE/GHC/Orphans.hs
```diff
@@ -132,24 +133,24 @@ srcSpanEndColTag = "srcSpanEndCol"
 instance ToJSON RealSrcSpan where
   toJSON spn =
       object
-        [ srcSpanFileTag .= unpackFS (srcSpanFile spn)
-        , srcSpanStartLineTag .= srcSpanStartLine spn
-        , srcSpanStartColTag .= srcSpanStartCol spn
-        , srcSpanEndLineTag .= srcSpanEndLine spn
-        , srcSpanEndColTag .= srcSpanEndCol spn
+        [ toJsonKey srcSpanFileTag .= unpackFS (srcSpanFile spn)
```
michaelpj, line 136: In these cases I think you could exploit the fact that `Key` has a `IsString` instance (and so does `Text`):
1. Define `srcSpanFileTag` and friends as type `String`
2. Use `fromString srcSpanFileTag` here, which will work to produce both a `Key` and a `Text`.

hls-plugin-api/src/Ide/Plugin/Properties.hs
```diff
@@ -352,7 +354,7 @@ toDefaultJSON :: Properties r -> [A.Pair]
 toDefaultJSON (Properties p) = [toEntry s v | (s, v) <- Map.toList p]
   where
     toEntry :: String -> SomePropertyKeyWithMetaData -> A.Pair
-    toEntry (T.pack -> s) = \case
+    toEntry (toJsonKey . T.pack -> s) = \case
```
michaelpj, line 357: `fromString` would work here.

hls-plugin-api/src/Ide/Plugin/Properties.hs
```diff
@@ -247,9 +249,9 @@ parseProperty kn k x = case k of
       )
       x
   where
-    keyName = T.pack $ symbolVal kn
+    key = toJsonKey . T.pack $ symbolVal kn
```
michaelpj, line 252: `fromString`

hls-plugin-api/src/Ide/Plugin/Properties.hs
```diff
@@ -371,7 +373,7 @@ toDefaultJSON (Properties p) = [toEntry s v | (s, v) <- Map.toList p]
 -- | Converts a properties definition into kv pairs as vscode schema
 toVSCodeExtensionSchema :: T.Text -> Properties r -> [A.Pair]
 toVSCodeExtensionSchema prefix (Properties p) =
-  [(prefix <> T.pack k) A..= toEntry v | (k, v) <- Map.toList p]
+  [toJsonKey (prefix <> T.pack k) A..= toEntry v | (k, v) <- Map.toList p]
```
michaelpj, line 376: `fromString` could work if you make `prefix` a `String` too.

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
```diff
@@ -26,9 +26,9 @@ import           Language.LSP.Types
 pluginsToDefaultConfig :: IdePlugins a -> A.Value
 pluginsToDefaultConfig IdePlugins {..} =
   A.Object $
-    HMap.adjust
+    adjustJson
       ( \(unsafeValueToObject -> o) ->
-          A.Object $ HMap.insert "plugin" elems o -- inplace the "plugin" section with our 'elems', leaving others unchanged
+        A.Object $ insertJson "plugin" elems o -- inplace the "plugin" section with our 'elems', leaving others unchanged
```
michaelpj, line 31: I think you could use the monoid instances here, i.e.
```
A.Object $ o <> object ["plugin" .= elems]
```

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
```diff
@@ -26,9 +26,9 @@ import           Language.LSP.Types
 pluginsToDefaultConfig :: IdePlugins a -> A.Value
 pluginsToDefaultConfig IdePlugins {..} =
   A.Object $
-    HMap.adjust
+    adjustJson
```
michaelpj, line 29: This is the only one I can't think of a nice way around :(

michaelpj (commented):
A few ideas: I think you could *mostly* get away without the CPP. The only one I'm stuck on is the `adjustJson` call...

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
jneira: damn, i was hoping we can take ride of the Compat module. Would you keep it only for adjust or do you think is better to copy here the cpp?

ghcide/src/Development/IDE/GHC/Orphans.hs
jneira: Thanks for the suggestion, not a big fan of `IsString`, but i dislike cpp even more :-)

hls-plugin-api/src/Ide/Plugin/Properties.hs
```diff
@@ -352,26 +353,26 @@ toDefaultJSON :: Properties r -> [A.Pair]
 toDefaultJSON (Properties p) = [toEntry s v | (s, v) <- Map.toList p]
   where
     toEntry :: String -> SomePropertyKeyWithMetaData -> A.Pair
-    toEntry (T.pack -> s) = \case
+    toEntry s = \case
       (SomePropertyKeyWithMetaData SNumber MetaData {..}) ->
-        s A..= defaultValue
+        fromString s A..= defaultValue
```
jneira, line 358: I had to inline `fromString` cause ghc was not able to infer the type (tips on how to convince it appreciated)

cabal-ghc901.project
```diff
@@ -41,22 +41,25 @@ index-state: 2021-12-29T12:30:08Z
 
 constraints:
   -- These plugins don't work on GHC9 yet
-  haskell-language-server +ignore-plugins-ghc-bounds -brittany -stylishhaskell -tactic,
+  -- Add a plugin needs remove the -flag but also update ghc bounds in hls.cabal
+  haskell-language-server +ignore-plugins-ghc-bounds -class -stylishhaskell -tactic,
```
Ailrun, line 45: Is this re-disabling of class intended?

Ailrun (commented):
Thank you for the update!

cabal-ghc901.project
jneira: hmm, not sure if i understand it, but just in case, `-flag` sets the flag value to false, so removing `-brittany` make the build use the default value (true)

jneira: no, no, good catch, thanks

jneira, line 45:
```suggestion
  haskell-language-server +ignore-plugins-ghc-bounds -stylishhaskell -tactic,
```

hls-plugin-api/src/Ide/Compat.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE CPP #-}
+module Ide.Compat where
+
+#if MIN_VERSION_aeson(2,0,0)
+import qualified Data.Aeson.KeyMap     as A
```
jneira, line 5:
```suggestion
import qualified Data.Aeson.Key             as A
import qualified Data.Aeson.KeyMap     as A
```

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
michaelpj: Okay, so what we'd probably want to avoid this is a aeson-lens library that handled both versions and gave us lenses for adjusting inside objects. 

Indeed, it looks like https://hackage.haskell.org/package/lens-aeson-1.1.3/docs/Data-Aeson-Lens.html#v:_Object would do it, and is set up to work with both pre and post 2.0 aeson. We could pull that in if we wanted, it's not got any deps we don't have already, I think?

michaelpj: There's also an `Ixed Value` instance.

jneira: Afaics we are already depending on all dependencies of lens-aeson

jneira: > There's also an `Ixed Value` instance.

Sorry but how can we leverage that instance?

michaelpj: I'd need to go and poke at all the lens stuff to figure it out, but I *think* we could use something like https://hackage.haskell.org/package/lens-4.17/docs/Control-Lens-At.html#v:iix. I can maybe try and take a look later.

jneira: Many thanks intermmediate/advanced lenses usage is still hard to me 😅

michaelpj: Or maybe just using the prism to focus on the object case and then using `at`? I always have to work this stuff out from scratch every time :sweat_smile:

michaelpj:
Boom https://github.com/jneira/haskell-language-server/pull/62 @jneira

jneira:
@michaelpj wow, many thanks for the patch, it makes me eager to learn more advanced lenses

michaelpj:
Now it looks like just the brittany plugin itself and some stack fixes?

jneira:
yeah the plugin code should use the compat module from ghcide instead directly ghc
I ve started to work on it but little progress, no problem if anyone tackles it

michaelpj:
Oh god, lots more aeson in the tests. I wonder if we should pull out the aeson 2 changes into a separate PR? A bit easier to iterate on, and there are a bunch of them.

jneira:
> Oh god, lots more aeson in the tests. I wonder if we should pull out the aeson 2 changes into a separate PR? A bit easier to iterate on, and there are a bunch of them.

I am afraid the unique thing we could separate is the tiny changes made to include floskell, brittany for ghc-9.0.1 needs aeson-2

jneira:
I was trying to replace HashMap calls in FunctionalCodeAction with lens-aeson accesors but my lens knowledge is very basic

jneira:
> I was trying to replace HashMap calls in FunctionalCodeAction with lens-aeson accesors but my lens knowledge is very basic

i think i got it 😅 🔪

jneira:
`shake-bench` also needs aeson-2 compatibility 🤦

michaelpj:
> i think i got it

yeah, looks great!

michaelpj:
I'll take a look at `shake-bench`.

jneira:
hmm dont know why but the error only is shown in the stack build:

```

shake-bench                       > /root/build/shake-bench/src/Development/Benchmark/Rules.hs:501:33: error:
shake-bench                       >     • Couldn't match type ‘Text’ with ‘Data.Aeson.Key.Key’
shake-bench                       >       Expected: GHC.Exts.Item
shake-bench                       >                   aeson-2.0.2.0:Data.Aeson.Types.Internal.Object
shake-bench                       >         Actual: (Text, Value)
shake-bench                       >     • In the pattern: (name, String gitName)
shake-bench                       >       In the pattern: [(name, String gitName)]
shake-bench                       >       In the pattern: toList -> [(name, String gitName)]
shake-bench                       >     |
shake-bench                       > 501 |   parseJSON (Object (toList -> [(name, String gitName)])) =
shake-bench                       >     |                                 ^^^^^^^^^^^^^^^^^^^^^^
shake-bench                       > 
shake-bench                       > /root/build/shake-bench/src/Development/Benchmark/Rules.hs:503:33: error:
shake-bench                       >     • Couldn't match type ‘Text’ with ‘Data.Aeson.Key.Key’
shake-bench                       >       Expected: GHC.Exts.Item
shake-bench                       >                   aeson-2.0.2.0:Data.Aeson.Types.Internal.Object
shake-bench                       >         Actual: (Text, Value)
shake-bench                       >     • In the pattern: (name, Object props)
shake-bench                       >       In the pattern: [(name, Object props)]
shake-bench                       >       In the pattern: toList -> [(name, Object props)]
shake-bench                       >     |
shake-bench                       > 503 |   parseJSON (Object (toList -> [(name, Object props)])) =
shake-bench                       >     |                                 ^^^^^^^^^^^^^^^^^^^^
shake-bench                       > 
shake-bench                       > /root/build/shake-bench/src/Development/Benchmark/Rules.hs:515:37: error:
shake-bench                       >     • Couldn't match type ‘Text’ with ‘Data.Aeson.Key.Key’
shake-bench                       >       Expected: GHC.Exts.Item
shake-bench                       >                   aeson-2.0.2.0:Data.Aeson.Types.Internal.Object
shake-bench                       >         Actual: (Text, Value)
shake-bench                       >     • In the expression: (n, String gitName)
shake-bench                       >       In the first argument of ‘fromList’, namely ‘[(n, String gitName)]’
shake-bench                       >       In the second argument of ‘($)’, namely
shake-bench                       >         ‘fromList [(n, String gitName)]’
shake-bench                       >     |
shake-bench                       > 515 |       Just n  -> Object $ fromList [(n, String gitName)]
shake-bench                       >     | 
```

michaelpj:
Probably because we don't have `benchmarks: true` in `cabal.project`, so `shake-bench` gets built only if you ask for it.

jneira:
> Probably because we don't have `benchmarks: true` in `cabal.project`, so `shake-bench` gets built only if you ask for it.

i just did a `cabal build --enable-benchmarks --project-file .\cabal-ghc901.project` and it did not build `shake-bench` but a `cabal build shake-bench  --project-file .\cabal-ghc901.project` worked. 🤷

jneira:
> > Probably because we don't have `benchmarks: true` in `cabal.project`, so `shake-bench` gets built only if you ask for it.
> 
> i just did a `cabal build --enable-benchmarks --project-file .\cabal-ghc901.project` and it did not build `shake-bench` but a `cabal build shake-bench --project-file .\cabal-ghc901.project` worked. 🤷

due to https://github.com/haskell/cabal/issues/6259 🤦

michaelpj:
https://github.com/jneira/haskell-language-server/pull/63

jneira:
Wow that was fast, many thanks

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
```diff
@@ -28,7 +28,7 @@ import           Language.LSP.Types
 -- | Generates a default 'Config', but remains only effective items
 pluginsToDefaultConfig :: IdePlugins a -> A.Value
 pluginsToDefaultConfig IdePlugins {..} =
-  -- Use 'ix' to look at all the "haskell" keys in the outer value (since we're not
+  -- Use 'key' to look at all the "haskell" keys in the outer value (since we're not
   -- setting it if missing), then we use '_Object' and 'at' to get at the "plugin" key
   -- and actually set it.
   A.toJSON defaultConfig & key "haskell" . _Object . at "plugin" ?~ elems
```
jneira, line 34: @michaelpj i changed `ix` for `key` here for readability but maybe you wanted to use `ix` on porpuse

michaelpj: nah, it's fine. `ix` is maybe more familiar to people since it's from `lens` but NBD

shake-bench/src/Development/Benchmark/Rules.hs
```diff
@@ -68,16 +68,19 @@ module Development.Benchmark.Rules
 
 import           Control.Applicative
 import           Control.Monad
+import           Control.Lens ((^.))
```
Ailrun, line 71: This one is a fairly minor thing but the formatting is not applied

haskell-language-server.cabal
```diff
@@ -472,7 +473,7 @@ test-suite func-test
   if flag(eval)
     cpp-options: -Deval
 -- formatters
-  if flag(floskell) && (impl(ghc < 9.0.1) || flag(ignore-plugins-ghc-bounds))
+  if flag(floskell) && (impl(ghc < 9.2.1) || flag(ignore-plugins-ghc-bounds))
```
Ailrun, line 476: It looks like Brittany flags are not reflected here

test/functional/FunctionalCodeAction.hs
```diff
@@ -56,11 +56,11 @@ renameTests = testGroup "rename suggestions" [
...
             cmd <- liftIO $ inspectCommand cars ["Replace with", "putStrLn"]
-            let Just (List [Object args]) = cmd ^. L.arguments
-                Object editParams = args HM.! "fallbackWorkspaceEdit"
+            let Just (List [args]) = cmd ^. L.arguments
+                editParams = args ^. key "fallbackWorkspaceEdit" . _Object
             liftIO $ do
-                "changes" `HM.member` editParams @? "Contains changes"
-                not ("documentChanges" `HM.member` editParams) @? "Doesn't contain documentChanges"
+                (editParams & has (ix "changes"))  @? "Contains changes"
+                not (editParams & has (ix "documentChanges")) @? "Doesn't contain documentChanges"
```
Ailrun, line 63: I guess they are more `key` cases?

Ailrun (approved):

jneira:
the new parsing in benchmarks is slighty different from the old one:

```
Error: Artifact path is not valid: /unprofiled/lsp-types/"upstream"/code_actions.csv. Contains the following character: Double quote "
```

haskell-language-server.cabal
jneira: yeah, sadly some of its dependencies still needs `allow-newer`, so it can not be included in the hackage build (there is a comment in cabal-ghc901.project remind us we have to change the condition when is possible)

test/functional/FunctionalCodeAction.hs
jneira: yeah, we should be consistent, `key` failed for me in the FunctionalCodeAction module so maybe we should switch to `ix` everywhere

haskell-language-server.cabal
Ailrun: Oh, I missed that one. Thank you for the answer!

michaelpj:
> the new parsing in benchmarks is slighty different from the old one

Ugh, I screwed up my string conversions trying to make them generic. I'll try again.

michaelpj:
https://github.com/jneira/haskell-language-server/pull/64

jneira:
> [jneira#64](https://github.com/jneira/haskell-language-server/pull/64)

many thanks, lets see ig ci is green here too ;-)

jneira:
Ok it seems is ok to be merged, @michaelpj @Ailrun many thanks!

michaelpj:
What a marathon lol

jneira:
> What a marathon lol

Yeah lib releases with security fixes which breaks everyone code are really fun

michaelpj:
Mergify doesn't want to merge for some reason, but the checks are all green, so I'll hit the button.

PR #2552, opened 2021-12-29
Remove cabal.project.freeze files in workflows after computing the cache key

jneira:
* Cause they break builds using `allow-newer` (like the ghc-9.2.1 branch)
* It does not need `continue-on-error` as it has a final `||`?
* And test floskell plugin 🤦‍♂️

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2552"></a>


.github/workflows/bench.yml
```diff
@@ -86,24 +86,29 @@ jobs:
         echo "INDEX_STATE=$INDEX_STATE1" >> $GITHUB_ENV
 
     - name: Form the package list ('cabal.project.freeze')
-      continue-on-error: true
```
Anton-Latukha, line 89: Yes, this no longer is needed.

.github/workflows/bench.yml
```diff
@@ -86,24 +86,29 @@ jobs:
...
         cache-name: hackage-sources
       with:
-        path: ${{ env.CABAL_PKGS_DIR }}
-        key:          ${{ env.cache-name }}-${{ env.INDEX_STATE }}
-        restore-keys: ${{ env.cache-name }}-
+        path:  ${{ env.CABAL_PKGS_DIR }}
+        key:   ${{ env.cache-name }}-${{ env.INDEX_STATE }}-${{ steps.compute-cache-key.outputs.value }}
+        restore-keys: |
+               ${{ env.cache-name }}-${{ env.INDEX_STATE }}-
+               ${{ env.cache-name }}-
```
Anton-Latukha, line 111: Please, keep the old version.

Skimming from full 100% caching of sources - is where we can save a lot of space.

The size of the sources ~ equals the size of the build results.

If we include `freeze` distinction on sources - that would produce x15 times more copies of source cache without real benefit.

Sources are fast to download. CI can live even without caching of sources.
Cachin of sources is done only for the purpose of the `cashing` workflow & reused only inside of it. The 30%-0% of sources that need additional download between versions in the matrix - downloading them takes seconds.

And `caching` build is an asynchronous process to the PR cycle, so downloading sources for a couple of seconds does not really affect PR loop speed.

That is why `key` is kept such - `index-state` is considered enough of a metric to be roughly in the area all builds need  (or how much `index-state` they have).

jneira: sure, I assumed the difference between caching and the others was not wanted

Anton-Latukha: When CI works properly - other `cabal` workflows should not need to download source code ever. Except when PR diff actively changes package configuration.

This is source code caching. It is used only inside the `caching` workflow. The workflow builds those sources & so what gets reused between workflows is compiled results.

We keep only 1 storage space (that depends on `index-state`) of it & in that way share most of the sources between `caching` jobs while at the same time not a lot of caching pool space.

Anton-Latukha: Oh, yes. Sorry. `hackage-sources` is still provided to other workflows, but main cause/use of it is for `caching` & to have minimal yet satisfactory source sharing.

.github/workflows/caching.yml
```diff
@@ -164,7 +167,7 @@ jobs:
           cache-name: hackage-sources
         with:
           path:  ${{ env.CABAL_PKGS_DIR }}
-          key:   ${{ env.cache-name }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project.freeze') }}
+          key:   ${{ env.cache-name }}-${{ env.INDEX_STATE }}-${{ steps.compute-cache-key.outputs.value }}
```
Anton-Latukha, line 170: Still, please `hackage-sources` should not depend of `freeze`, depending on `index-state` is enough for its cause.

jneira: ah ok, so I wanted to have all consistent using one path and finally we can do it the other way around, nice anyway

Anton-Latukha: Oh - that started from my mistake. I put `-${{ hashFiles('cabal.project.freeze') }}` there in the first place :facepalm:

jneira: fixed 😀

Anton-Latukha (approved):

Anton-Latukha:
Happy New Year 8) :christmas_tree:

jneira:
> Happy New Year 8) :christmas_tree: 

happy new year everyone! (4 hours left here in Spain)

PR #2556, opened 2022-01-01
Apply missing update for stack-9.0.1.yaml

Ailrun:
I somehow forgot to commit this change when I updated the settings related to the class plugin for GHC 9.0.1

(That's why I need Magit, haha)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2556"></a>


jneira (approved):
Thanks!

PR #2557, opened 2022-01-01
Fix nix flake by explicit version for `lsp-xxx` packages

guibou:
The nix flake was failing:

```
nix build .#haskell-language-server
```

Was failing because of bounds issue related to hackage `lsp`,
`lsp-types` and `lsp-test` projects.

I've bump theses version, introducing new flake input (so they can be
changed easily in the future) and now the build works.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2557"></a>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,9 +21,22 @@
       url = "github:hercules-ci/gitignore.nix";
       flake = false;
     };
+
+    lsp-source = {
+      url = "https://hackage.haskell.org/package/lsp-1.4.0.0/lsp-1.4.0.0.tar.gz";
+      flake = false;
+    };
+    lsp-types-source = {
+      url = "https://hackage.haskell.org/package/lsp-types-1.4.0.0/lsp-types-1.4.0.0.tar.gz";
+      flake = false;
+    };
+    lsp-test-source = {
+      url = "https://hackage.haskell.org/package/lsp-test-0.14.0.2/lsp-test-0.14.0.2.tar.gz";
+      flake = false;
+    };
   };
   outputs =
-    { self, nixpkgs, flake-compat, flake-utils, pre-commit-hooks, gitignore }:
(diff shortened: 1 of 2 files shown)


jneira (commented):
Thanks for fixing the nix build, will wait for a review from an actual nix user to merge

flake.nix
```diff
@@ -21,9 +21,22 @@
...
+      flake = false;
+    };
+    lsp-test-source = {
+      url = "https://hackage.haskell.org/package/lsp-test-0.14.0.2/lsp-test-0.14.0.2.tar.gz";
+      flake = false;
+    };
   };
   outputs =
-    { self, nixpkgs, flake-compat, flake-utils, pre-commit-hooks, gitignore }:
+    { self, nixpkgs, flake-compat, flake-utils, pre-commit-hooks, gitignore, lsp-source, lsp-types-source, lsp-test-source }:
```
michaelpj, line 39: Cute. In the past we've just used an explicit fetchurl for this sort of thing, but this is nicer, I think. That said, it might be simpler to update if the flake input pointed to the corresponding tag in the github repository instead of hackage?

michaelpj (approved):

flake.nix
michaelpj: No obligation, I'll merge this in a day or so regardless.

guibou: Initially I tried to load it from github directly.. However I had problem finding the tags in the git repository. For example, there is no `1.4.0.0` tag (or any other variation, unless I'm wrong) in the https://github.com/haskell/lsp repository.

Then I realized that the source of truth is hackage. The `cabal` file for Haskell language server is actually using bounds based on hackage version and hence fetching the revision as it is on hackage seems more robust from my point of view.

michaelpj: The tags are confusing indeed, they're prefixed by the package. And it's no wonder you weren't able to find them... I forgot to push the tags (done now) :sweat_smile: anyway, not a big deal.

Issue #2492, opened 2021-12-16
Fix and improve server logging


jneira:
* hls architecture is complex, with many moving pieces (hie-bios, hiedb, ghcide, plugins, etc) and many points of failure so we need a good log setup allowing users and maintainers trace bug causes as fast as possible
* Now, afaics the logging system has some flaws:
  * It uses different logging frameworks, config options and criteria about what and where is the output
  * Logging levels are not consistent across components
  * And the worse thing, not always errors are reported in the log and users ends with a no working hls session with no clue about why is the cause

As @alanz suggested in [this issue](https://github.com/haskell/haskell-language-server/issues/39 about harmonising logging :  

> * how and what are we going to log? Tapping into the LSP expectations makes sense, which seems to be to use stderr, which can then be managed per process and per client.
> * Which logging framework will we use? In my mind this should be common across haskell-lsp, ghcide, and hls 

I would add:
* ensure all errors even catastrophic ones are reported with the max info available
* use proper error levels setup with at least error, info and debug
* move all actual traces about rules to debug, as afaics the main information they give is what rule was fired and how many time it took, so useful for performance analysis (but only partially)
* expose all log level selection in all components and make it configurable up down from hls via cli arg and lsp config

pepeiborra:
Sigma IDE uses a contra variant logger with support for tracing. It has the usual backends (console, event log) as well as an LSP backend for checking invariants in tests.

```
data LogMessage
  = LogRule LogRuleMessage
  | LogCore LogCoreMessage
  | LogCompile LogCompileMessage
  | LogOther LogOtherMessage

data LogRuleMessage
  = LogReindexingFile FilePath
  | LogRegeneratingInterface FilePath
  | ...

-- | A contravariant tracing abstraction
data Tracer msg = Tracer
  { logMsg_ :: forall m . (HasCallStack, MonadCatch m) => msg -> m ()
    -- ^ Log a message
  , traceMsg_ :: forall a m . (HasCallStack, MonadCatch m) => msg -> m a -> m a
    -- ^ Trace the begin and end of a computation
  }
```

jneira:
Wow that sounds cool, any chance to use it in hls? i mean the source code is available to be copied or used via a public library

pepeiborra:
This idea is implemented by logging libraries like `co-log` but Sigma IDE just rolls its own solution with the abstractions pasted above. There really isn't that much code.

```
-- Explicit record accessors to preserve call stacks

[...]

PR #2558, opened 2022-01-01
Improve logging

eddiemundo:
This PR tries to implement the contravariant logging/tracing library described by https://github.com/haskell/haskell-language-server/issues/2492#issuecomment-995575301

Currently the most straightforward way of contravariant logging that this PR does:
1. If module has no `Log` datatype then add one
2. Invent a constructor (name) for the `Log` datatype taking the values you want to log
3. Log using it, e.g. `logWith recorder Warning $ LogConstructorName value1 value2`
4. If a function needs a logger you pass a logger to it, but if this function is in another module you contramap to the `Log` type of that module e.g.
```haskell
-- Shake.hs
data Log = LogStuff

function :: Recorder (WithPriority Log) -> IO Thing

-- MyModule.hs
import qualified Shake

data Log = LogShake Shake.Log

myFunc :: Recorder (WithPriority Log) -> IO Thing
myFunc recorder = Shake.function (cmapWithPrio LogShake recorder)
```
4. _cont._ This means each logged value when fully constructed is a sequence of constructor wrappers of different modules 
`Log` datatypes.
e.g.
```haskell
WithPriority Info (MainModule.LogAnotherModule (AnotherModule.LogMyModule (MyModule.LogShake (Shake.LogStuff))))
```
5. Currently each module is able to convert its `Log` value to `Doc`s by a `Pretty Log` instance defined in it. The final log value is converted to a `Doc` by that module's `Pretty Log` instance which delegates to other `Pretty Log` instances if necessary.
6. We don't convert `Log` values to priorities. Instead they are specified at the log call site (like how every other log framework does it), and stored alongside the log value using `WithPriority a`. The "first/final" logger is a `Recorder (WithPriority (Doc ann))`. This allows us to more easily alter priorities when contramapping the logger which is more flexible than converting a log value to a priority.

### Cons
- Logging is not very casual. If you want to log random messages you can add a `LogOther Text` constructor to the `Log` type of the current module, but if you are logging "for real" then you invent a constructor that takes the values you want to log, and then add to the `Pretty Log` instance (add a pattern matching branch) to convert the `Log` to a `Doc`
- Logging in a plugin means also you need to pass a logger to the plugin descriptor function which means in `exe/Plugins.hs`, a `Log` constructor should be added to the existing `Log` datatype for the plugin if it doesn't exist, and a contramapped recorder should be passed to the plugin's descriptor function. Additionally, you need to add entries for the `pretty` pattern match. This pattern matching needs `CPP` like everything else in that module. This should be simple to do by following the existing example, but still an extra step.
- Contravariant logging composes functions to build new loggers, so there isn't an easy way of taking a logger and decomposing it to get the base version of the logger (to say, remove a priority filter).
- Requires understanding `contramap`

### Pros
- The logger can do pretty much anything depending on how its logging function is defined. So this means we can inspect log values in tests, convert log values to LSP messages, text, structured logging, etc.
- Conversion to backend values is more cohesive because it's done in one place per module.
- The source is less cluttered with formatting details, value conversions, checking of test flags, but sometimes logging text can serve as comments
- Can do more fine-grained filtering, like filtering specific log messages based on their values, but that is often not ergonomic. You can also do filtering on the caller side by contramapping loggers before passing them on.

### Questions (for the future):
- To implement other backends like LSP messages, there seem to be multiple ways of doing it:
  1. The Duncan Coutts video way which is to bundle another logger that sends LSP messages with the usual console/file logger. Then at the logging call site you can decide to use which logger depending on which backend you want to send to. This seems like the simplest way.
  2. Encode whether the message should go to the LSP backend or console/file backend in the `Log` data type. In that case the single logger decides how to interpret the message later.
  3. Have a separate log datatype per backend.
- There is boilerplate when it comes to converting log values to backend values, and when constructing `Log` values. Maybe generic programming or TH can be useful here.

Todo:
- [x] make test utils compile
- [x] allow both previous and contravariant style to write same log file
- [x] convert log values to match previous text
- [x] convert log values to match previous priorities
- [x] the final log type should probably be `WithPriority (Doc a)` instead of `WithPriority Text` since we're using prettyprinter
- [x] fix test loggers to use pretty printer and `logToPriority` function
- [x] delete log comments that helped me remember what the original logging did
- [x] ghcide exe default logger doesn't use hslogger and prints to stderr, so match that behaviour (noticeable in CI test logs)
- [x] provide log priorities at log call sites instead of separate function

Todo in other PRs:
- convert all instances of `logInfo`, `logDebug`, etc, to `logWith`
- some kind of generics to reduce boilerplate?
- lsp message backend, convert all notifyTesting calls
- implement/use tracing functionality where appropriate
- eliminate all places where exceptions can be swallowed
- add more logging for everything we care about and all the extra stuff @jneira was saying in the original issue.
- write docs explaining how to use contravariant logger / what it can do

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2558"></a>


ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -27,33 +54,164 @@ data Priority
...
+messageToText = \case
+  LogOther LogOtherMessage{ stack, priority, payload } -> do
+    threadId <- myThreadId
+    utcTime <- getCurrentTime
+    pure $ Text.intercalate " | "
+      [ utcTimeToText utcTime
+      , callStackToLocationText stack
+      , threadIdToText threadId
+      , priorityToText priority
+      , payload ]
```
pepeiborra, line 200: I'm not keen on some of these details as I spend a lot of time reading test logs from CI:

- timestamps can be recorded by the CI system (or your terminal)\
- thread ids are not particularly helpful most of the time

pepeiborra:
This is an example of a custom logger used in an in-process test:
```
 where
    -- a logger that intercepts and delays Buck targets fetching for 20s
    -- to ensure that the test is not sensitive to timing issues
    delayBuckInit =
      mempty
        { traceMsg_ = \msg act -> case msg of
            LogKnownTargets (KnownTargets.LogBuck Buck.Log {}) -> do
              liftIO $ sleep 20
              act
            _ -> act
        }
```

pepeiborra:
This is an example of a test that parses (strongly typed) log messages to prove invariants: 
```
   let eitherFormatOrInit =
          asum
            [ Left
                <$> satisfyLog
                  ( \case
                      KnownTargets.LogGetSigmaTargetsResult {} -> pure ()
                      _ -> empty
                  )
            , Right <$> responseForId STextDocumentFormatting rid
            ]
    res <- skipManyTill anyMessage eitherFormatOrInit
   ...
```

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -27,33 +54,164 @@ data Priority
...
     deriving (Eq, Show, Ord, Enum, Bounded)
 
+data LogMessage =
+  LogOther LogOtherMessage
+
+data LogOtherMessage = LogOtherMessage
+  { stack    :: !CallStack
+  , priority :: Priority
+  , payload  :: Text
+  }
```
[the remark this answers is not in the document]
eddiemundo: I remember reading this and it wouldn't go into my brain.

How would it look to log something? For instance would it be more verbose like
```haskell
data GhcideLog = GhcideLog Text

logMsg logger $ LogGhcide $ GhcideLog "something is happening"
```
and then how do we get from `LogMessage` to (for instance) the text that is output to `stderr` including the priority? Is it something like this?
```haskell
logMessageToText :: Priority -> LogMessage -> Text

infoLogger = cmap (logMessageToText Info)  textStderrLogger
debugLogger = cmap (logMessageToText Debug) textStderrLogger
```
and then I guess we can recover our `logInfo`, `logDebug` api by doing
```haskell
logInfo text = logMsg_ infoLogger (LogGhcide $ GhcideLog text)
logDebug text = logMsg_ debugLogger (LogGhcide $ GhcideLog text)
-- or more generally
logInfo :: GhcideLog -> IO ()
logInfo msg = logMsg_ infoLogger (LogGhcide msg)
```
but this means that we need to define `logInfo`, `logDebug`, etc per different kind of `LogMessage` if we want the convenience. 

So if `LogMessage` is in the `Main` module then so would the `logMessageToText priority` function, as well as the `infoLogger`, `debugLogger`, ... and so that is what is meant by `Main` deciding things? I guess if we want minimum log levels we'd have conditionals in the functions `infoLogger`, `debugLogger`, etc, to be the empty logger if below the log level.

Or maybe instead each plugin or message type defines its own logger like `Logger GhcideLog` and then Main can convert it into a `Logger LogMessage` and then `Main` eventually passes this logger around to everybody else?

I feel like I could be totally off base.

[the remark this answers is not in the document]
pepeiborra: This is a good introduction: https://www.youtube.com/watch?v=qzOQOmmkKEM

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: I think we can config it so when testing no timestamps or thread id, because I think timestamps can be useful for users reporting issues.

ghcide/src/Development/IDE/Types/Logger.hs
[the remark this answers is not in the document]
pepeiborra: > how do we get from LogMessage to (for instance) the text that is output to stderr including the priority?

Using `contramap` you transform your `Text` logger to accept more specific log types. 

The priority is subjective and decided by the application, not by the `logMsg` call site. 

To support this in HLS, a few architectural changes will be needed:

1. The `PluginDescriptor` type will need to be parameterised by the type of the plugin log messages
2. The `PluginDescriptor` type will need an additional field to provide a priority function: `logtype -> Priority`
3. The plugins will no longer use the `logger` field from `ShakeExtras`. Instead, they will expect a logger of the type they need as an argument to construct the `PluginDescriptor`

For 2 there are alternative encodings, like wrapping all log types in a `WithDefaultPriority` tuple type.

3 does have some implications - it will make dynamic loading of plugins harder to support if we ever decide to go down that route, which seems very unlikely anyway.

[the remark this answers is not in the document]
pepeiborra: > Or maybe instead each plugin or message type defines its own logger like Logger GhcideLog and then Main can convert it into a Logger LogMessage and then Main eventually passes this logger around to everybody else?

This is the way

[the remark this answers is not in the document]
eddiemundo: I watched the video and I'm pretty sure I get it now. I'll sleep and then try and do the do.

pepeiborra:
Just wanted to say that this is going in the right direction and very impressively so, in case you still had any doubts. 

And a suggestion - instead of a Big Bang, try to land it incrementally. Don't make any more changes, polish what you have and land it before converting any more modules and plugins.

michaelpj:
> Converting these log values to text/priorities/other is done separately and maybe in a different location. I think the idea is that each entry point will have a table converting Logs to priorities, and/or text, and or other things. Or some other centralized place.

I think this may become unwieldy.
- Perhaps we could provide `Pretty` instances for `Log` types? That at least gives a straightforward route to "normal" logging.
- I'm not sure what to do with priorities. It does seem that within a module we usually have information about the *relative* priorities of logs at least. Perhaps we could always log with `WithPriority`, and then allow callers to rearrange the priorities if they care (e.g. "demote all info messages from plugins to debug" would be fairly doable, I think)

eddiemundo:
Yeah I was thinking with the various `Main` modules, concentrating logging stuff in each them it would be messy. If instead we have some concentration point before them that they import it would still bit weird because this concentration point would have to also contain any `Log` types for the `Main` modules themselves used to have or there would be circular imports, another strategy is we can also just not log things in `Main` modules but that is a weird restriction, and hard to realize I think.

So like you said the easiest thing would to have each module know how to convert its `Log` into a final log type which currently is `WithPriority Text`. So `Pretty` instances would work because currently I'm using `Show` and that's pretty convenient, or we could just have a function per module that converts log values to `Doc`.
For the priorities we could also have a `logToPriority` function in each module to handle priorities. I don't think that is too bad.

However, in the Duncan Coutts video, the impression I got was they already have a log value data type that is independent of the module structure. If that's the case then centralization is a lot easier. That's technically possible... I just look at every module that logs stuff and gather all the logging types in one place. It would still mimic the module structure but not literally needing to import each module aliased. It's also more flexible because well a module doesn't actually have to use a `Log` type specific to the module its in. The disadvantage is rearranging things to avoid circular imports, maybe a lot.

exe/Plugins.hs
```diff
@@ -2,107 +2,122 @@
...
+  | LogExample Example.Log
+  | LogExample2 Example2.Log
+#if tactic
+  | LogTactic Tactic.Log
+#endif
+  | LogEval Eval.Log
+  | LogExplicitImports ExplicitImports.Log
+  | LogRefineImports RefineImports.Log
+  | LogHlint Hlint.Log
+  | LogAlternateNumberFormat AlternateNumberFormat.Log
```
pepeiborra, line 109: `Eval.Log` is not available unless `eval` is CPP defined. Similarly for all the other plugin log types. 

Bear in mind that CI won't catch this, since it doesn't test non-default flag combinations

pepeiborra (changes requested):

exe/Plugins.hs
jneira: we have a job which is testing non default flags, it would be great to add this one

michaelpj (commented):
I did a very partial look at this, will try and look later. I think it looks very promising, I'm keen to build on top of it :)

michaelpj:
FYI I have a PR open to use a similar approach in `lsp`, which I think will work well with this PR. Comments welcome! https://github.com/haskell/lsp/pull/401

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1186,13 +1185,13 @@ newtype Priority = Priority Double
 setPriority :: Priority -> Action ()
 setPriority (Priority p) = reschedule p
 
-ideLogger :: IdeState -> Logger
-ideLogger IdeState{shakeExtras=ShakeExtras{logger}} = logger
+ideLogger :: IdeState -> Recorder LogMessage
+ideLogger IdeState{shakeExtras=ShakeExtras{recorder}} = recorder
```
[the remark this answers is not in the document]
eddiemundo: Somehow this diff no longer exists... so currently the `ideLogger` still exists for the sake of not having the PR change everything.

exe/Main.hs
```diff
@@ -1,21 +1,61 @@
...
+  LogIdeMain log -> IdeMain.logToPriority log
+  LogPlugins log -> Plugins.logToPriority log
+
+logToDocWithPriority :: Log -> WithPriority (Doc a)
+logToDocWithPriority log = WithPriority (logToPriority log) (pretty log)
 
 main :: IO ()
 main = do
-    args <- getArguments "haskell-language-server" (idePlugins False)
+    -- passing mempty for recorder to idePlugins means that any custom cli
```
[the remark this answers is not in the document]
eddiemundo: I think the problem I found was to have a logger to pass into `idePlugins` for `getArguments` I need the result of `getArguments` but `getArguments` needs `idePlugins` which needs a logger.

At the time I felt like the most straightforward way is to have getArguments parse any logging arguments first, then create a logger, then parse the remaining arguments but I thought that might be too crazy to change up the argument parsing in the PR, and currently there are no plugins that use the CLI functionality except for the example plugin.

Another way is to have some default logger that doesn't care about command line arguments to pass in for the plugin CLI functionality, but it's kind of weird.

exe/Main.hs
```diff
@@ -1,21 +1,61 @@
...
+    -- command provided by a plugin will not have logging powers
+    args <- getArguments "haskell-language-server" (Plugins.idePlugins mempty False)
+
+    let (hsLoggerMinPriority, minPriority, logFilePath, includeExamplePlugins) =
+          case args of
+            Ghcide GhcideArguments{ argsTesting, argsDebugOn, argsLogFile, argsExamplePlugin } ->
+              let (minHsLoggerPriority, minPriority) =
+                    if argsDebugOn || argsTesting then (HsLogger.DEBUG, Debug) else (HsLogger.INFO, Info)
```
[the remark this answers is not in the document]
eddiemundo: Yes that's a good idea.

ghcide/exe/Main.hs
```diff
@@ -55,26 +81,36 @@ main = withTelemetryLogger $ \telemetryLogger -> do
       Nothing   -> IO.getCurrentDirectory
       Just root -> IO.setCurrentDirectory root >> IO.getCurrentDirectory
 
-    let logPriority = if argsVerbose then Debug else Info
-        arguments = if argsTesting then Main.testing else Main.defaultArguments logPriority
+    let (hsLoggerMinPriority, minPriority) = if argsVerbose then (HsLogger.DEBUG, Debug) else (HsLogger.INFO, Info)
+
+    defaultRecorder <- makeDefaultStderrRecorder hsLoggerMinPriority
```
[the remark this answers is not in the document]
eddiemundo: Yeah, ghcide doesn't have a logfile option. Additionally there is a testing version of the `Arguments` that is used by the ghcide/test/Main.hs. It adds plugins that I don't know what they are for, and it sets a testing flag that is used to send log notifications.

ghcide/exe/Main.hs
```diff
@@ -55,26 +81,36 @@ main = withTelemetryLogger $ \telemetryLogger -> do
...
+    IDEMain.defaultMain (cmap LogIDEMain recorder) arguments
+        { IDEMain.argsProjectRoot = Just argsCwd
+        , IDEMain.argCommand = argsCommand
+        , IDEMain.argsLogger = IDEMain.argsLogger arguments <> pure telemetryLogger
 
-        ,Main.argsRules = do
+        , IDEMain.argsRules = do
             -- install the main and ghcide-plugin rules
-            mainRule def
+            mainRule (cmap LogRules recorder) def
```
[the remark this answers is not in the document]
eddiemundo: Not sure what you mean. Should `LogRules` be `LogIDEMain`? `mainRule` is in the `Rules` module and currently doing one `Log` type per module.

ghcide/src/Development/IDE/Core/OfInterest.hs
```diff
@@ -32,24 +34,40 @@ import qualified Data.ByteString                          as BS
...
 import           Development.IDE.Types.Exports
 import           Development.IDE.Types.Location
 import           Development.IDE.Types.Logger
+import qualified Development.IDE.Types.Logger             as Logger
 import           Development.IDE.Types.Options            (IdeTesting (..))
 import qualified Language.LSP.Server                      as LSP
 import qualified Language.LSP.Types                       as LSP
+import           Prettyprinter                            (Pretty (pretty))
+
+data Log = LogShake Shake.Log
```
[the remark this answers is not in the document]
eddiemundo: I think that would be OK but I also feel like if we needed to add more logging it would be less confusing for there to be a template already existing that could be added to.

exe/Main.hs
```diff
@@ -1,21 +1,61 @@
...
+
+data Log
+  = LogIdeMain IdeMain.Log
+  | LogPlugins Plugins.Log
+  deriving Show
+
+instance Pretty Log where
+  pretty log = case log of
+    LogIdeMain ideMainLog -> pretty ideMainLog
+    LogPlugins pluginsLog -> pretty pluginsLog
```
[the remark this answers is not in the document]
eddiemundo: I think my main reservation about this entire PR is really the boilerplate. Each function that wants logging needs a logger passed, and yeah the delegation that each Pretty instance and `logToPriority` function does is a bit annoying. This annoyance might extend to future `logToX`things. I'm just worried that for a contributor wants logging that this might be too annoying.

The above particular example is extra bad because I didn't do what I started doing later which was not to name the log variable too specifically.

Using TH or generics to reduce boilerplate I don't mind, if there's a clear explanation for it probably in the the contributor guide, and the generic function itself. I was planning on writing an explanation for this stuff anyway so maybe it's worth it.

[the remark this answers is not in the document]
michaelpj: I've been thinking about dealing with the priority issue like this:
```
data WithSeverity a = WithSeverity { getMsg :: a, getSeverity :: Severity }

data Log1 = ...
data Log2 = ALog1 Log1 | ...

adapt :: LogAction m (WithSeverity Log2) -> LogAction m (WithSeverity Log1)
adapt = cmap (fmap ALog1)

logWithSeverity :: LogAction m (WithSeverity msg) -> Severity -> msg -> m ()
```

Then you'd end up stating the severities dynamically at the log site, but TBH I think that's okay, given that most messages are logged in only one place. And people are used to seeing `logDebug` etc. at the log site.

exe/Main.hs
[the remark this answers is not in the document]
michaelpj: I think it would be sane to just pass a dumb stderr logger here? Or comment that we can't pick a logger because we don't know the user's logging preferences yet.

Or, indeed, we could use a dummy log that just accumulates the logs in a `Writer` and then re-log them when we get the real logger!

ghcide/src/Development/IDE/Core/OfInterest.hs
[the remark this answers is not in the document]
michaelpj: Fair enough.

ghcide/src/Development/IDE/Core/Shake.hs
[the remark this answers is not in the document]
michaelpj: :+1:

exe/Main.hs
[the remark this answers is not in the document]
michaelpj: See also https://github.com/co-log/co-log-core/issues/13

ghcide/exe/Main.hs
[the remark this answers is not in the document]
michaelpj: I was just getting confused. `LogRules` is a constructor of `LogIDEMain`.

michaelpj:
I'm keen to help out with this. Let me know what would be useful. I can do more reviewing, or if there's a more minimal version of this we can merge then I can also do some boilerplate wrangling.

exe/Main.hs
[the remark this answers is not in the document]
eddiemundo: I was considering that before but it directly opposed part of https://github.com/haskell/haskell-language-server/pull/2558#discussion_r777182269
> The priority is subjective and decided by the application, not by the `logMsg` call site.

Priorities being separate pros:
- more centralized so glancing over all priorities is easier
- some things don't have natural priorities (yet they would still have to add a pattern matching branch or use catchall branch)
- every log constructor is prioritized the same way (could also be a Con)
- any priority logic involving the logged values can be written here
- less noise for code reader (usually priorities are noise?)

Priorities being separate cons:
- more boilerplate
  - function(s) has to exist somewhere to do the conversion
  - pattern matching syntax requires writing the constructor (except for catch-all)
  - a significant amount of pattern matching branches delegate to some other `logToPriority` function

I think there was exactly 1 time so far where there was some priority logic that depended on the logged values and it felt cool to have that logic separate from the main code.

My reservations are totally about whether other people think it's tolerable to have these extra steps to log stuff. The boilerplate doesn't bother me much personally...

exe/Main.hs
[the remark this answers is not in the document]
eddiemundo: Yeah I think the dumb stderr logger might be OK since for the accumulator version the logs wouldn't appear in real time. Another strategy would be some way of having that plugin cli command receive a logger later using laziness or mutable variable, but I think keeping it simple for now is better.

eddiemundo:
> I'm keen to help out with this. Let me know what would be useful. I can do more reviewing, or if there's a more minimal version of this we can merge then I can also do some boilerplate wrangling.

I think I'll try and fix conflicts, and do the suggestions. Reviewing any remainder in the mean time or after would be good and then can merge.

exe/Main.hs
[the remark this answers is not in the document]
michaelpj: Hmmm. I can see why you're doing that, but I'm not 100% sure it's desirable. I guess I have a sort of status-quo argument for not doing it: the old system set priorities at the log site; most other logging systems in other languages (that I've used) set priorities at the log site; it can't be *that* bad.

I'm also not sure that it makes it easier to manipulate priorities. It depends where that happens. How would I do something like the following?

```
suppressErrors  :: LogAction m (WithSeverity msg) -> LogAction m (WithSeverity m)
suppressErrors = cmap (\(WithSeverity sev msg) -> WithSeverity (if sev == Error then Warning else sev) msg)

-- Suppress errors from this subsystem, they're not fatal for us
doSomething (suppressErrors log)
```

It might also make the work simpler. Step 1 could leave priorities at the log sites as now, and then step 2 could pull them out. It might make this change easier to land?

eddiemundo:
is there a way of having `prettyprinter` be version 1.7+ for circleci ghc 8.6.5 or do I have to just replace all
```
import Prettyprinter
```
with
```
import Data.Text.Prettyprint.Doc
```
and ignore the deprecation warnings?

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: For anything that uses the ghcide executable it should be very similar to how it is now.

exe/Main.hs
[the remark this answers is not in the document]
eddiemundo: Yeah I've been thinking that logs and filtering seem pretty intrinsically tied. I guess it doesn't matter what your backend is, you always want a way to be able to filter messages. You might be able to filter them based on the their constructor and values, but we have no way of passing that information to down stream, and so having a value that represents that information is good.

I don't think there is a way of doing your example with `logToPriority` functions since all of the priorities get finally calculated at the end.

Ironically I'd have to change a lot of things in the PR to put priorities back at the log sites, but I don't mind doing that. Actually, I'll just do that since if anybody objects I can just revert to the current state which I think is ready for review.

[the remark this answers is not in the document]
michaelpj: Sorry for suggesting such an annoying change! :sweat_smile:

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -82,16 +86,105 @@ import           Control.Concurrent.STM.Stats         (atomically, modifyTVar',
...
+  | LogHieDbWriterThreadSQLiteError !SQLError
+  | LogHieDbWriterThreadException !SomeException
+  | LogInterfaceFilesCacheDir !FilePath
+  | LogKnownFilesUpdated !(HashMap Target (HashSet NormalizedFilePath))
+  | LogMakingNewHscEnv ![UnitId]
+  | LogDLLLoadError !String
+  | LogCradlePath !FilePath
+  | LogCradleNotFound !FilePath
+  | LogSessionLoadingResult !(Either [CradleError] (ComponentOptions, FilePath))
+  | forall a. Show a => LogCradle !(Cradle a)
```
fendor, line 118: I think it should be fine to choose here the concrete cradle

```suggestion
  | LogCradle !(Cradle Void)
```

fendor (approved):
I only looked at the Session loading code, but that looks awesome!

fendor:
This is really impressive! I especially love that we can convert log messages to lsp messages, since log messages from Session.hs should be reported to users more often, but it was so far quite tricky to do that.

ghcide/session-loader/Development/IDE/Session.hs
eddiemundo: Done, but that also required changing `cradleToOptAndLibDir` to take a `Cradle Void` instead of a `Show a => Cradle a` as it did previously.

berberman:
Can we re-export `prettyprinter` stuffs in `Development.IDE.Types.Logger`, so that downstream packages don't have to explicitly depend on this library?

eddiemundo:
> Can we re-export `prettyprinter` stuffs in `Development.IDE.Types.Logger`, so that downstream packages don't have to explicitly depend on this library?

Yeah that sounds like a good idea. Gonna do that.

pepeiborra (approved):
Reading the PR description, this is pretty much what Sigma IDE does as described in #2492. Approving early as I don't want to hold back the PR leading to inevitable merge conflicts.

For future work, we should have a logging backend that forwards some LSP messages (encoded as JSON values) to the LSP stream for testing invariants.

ghcide/src/Development/IDE/Main.hs
```diff
@@ -187,18 +235,16 @@ data Arguments = Arguments
...
-defaultArguments :: Priority -> Arguments
-defaultArguments priority = Arguments
+defaultArguments :: Recorder (WithPriority Log) -> Logger -> Arguments
+defaultArguments recorder logger = Arguments
         { argsProjectRoot = Nothing
         , argsOTMemoryProfiling = False
         , argCommand = LSP
-        , argsLogger = stderrLogger priority
-        , argsRules = mainRule def >> action kick
+        , argsLogger = pure logger
```
pepeiborra, line 244: What is the old `Logger` type used for now? If it is still being used, can it be subsumed with a `LogLegacy Priority` recorder?

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -8,12 +8,48 @@
...
+                                             defaultLayoutOptions, layoutPretty,
+                                             vcat, (<+>))
+import           Prettyprinter.Render.Text  (renderStrict)
+import           System.IO                  (Handle, IOMode (AppendMode),
+                                             hClose, hFlush, hSetEncoding,
+                                             openFile, stderr, utf8)
+import qualified System.Log.Formatter       as HSL
+import qualified System.Log.Handler         as HSL
+import qualified System.Log.Handler.Simple  as HSL
+import qualified System.Log.Logger          as HsLogger
```
pepeiborra, line 50: This is a significant change in behaviour. Switching to HsLogger seems unrelated to the contravariant logging changes, and better left for another PR

ghcide/src/Development/IDE/Main.hs
eddiemundo: The old logger type is still being used, it's just being initialized at various entry points and passed in.

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: The only reason for the hslogger stuff is so we can still see them in the logs because hie-bios and a few other places in the code base haven't been converted to the contravariant logging system.

pepeiborra: I don't follow this. Are you saying that the old stderr logger implementation didn't work with the contravariant system?

ghcide/src/Development/IDE/Main.hs
eddiemundo: Not exactly sure what the `LogLegacy Priority` means, but the original plan was to remove `argsLogger` completely eventually.

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: Although now that I look, it seems the only place in hls that uses hslogger is the Hlint plugin, but I'm pretty sure I need the hslogger stuff to get the hiebios output for log files. The reason to copypaste the hiebios stuff was so the outputs wouldn't interleave improperly when we write to the handle.

pepeiborra: But the old stderr logger was already using locking to prevent undesired interleaving. I'm not keen on adding a new dependency in ghcide, specially as @michaelpj  is planning to drop the hslogger dependency in `lsp`

eddiemundo: Oh OK didn't see the newer message.

Previously HLS used both hslogger, and a plain stderr logger depending on the entry point. If `ghcide` process was run it would use stderr logger, and otherwise I think it would overwrite the default stderr logger with the hslogger logger. 
Additionally `hie-bios` uses hslogger to write compilation logs to output. The hslogger was initialized by calling `setupLogger` from `lsp`. If I kept using the `setupLogger` from `lsp` then the logs would get interleaved weirdly because they don't have access to the `lock`, and so I decided to copy-paste that code with some modifications so we could still get any logs written to some handle using hslogger, but also not have them displayed weirdly.

eddiemundo: Maybe the confusion is thinking that we're using hslogger as the console/file backend. We're not... we're using just hslogger to setup a hslogger so we can pick up `hie-bios` logs to the right handle, and give it a lock so it doesn't do funny things with messages we write to the actual handle backend.

eddiemundo:
There is a commit on my branch 2 hours ago that this PR is based on and I don't see it here :shrug:

isovector (approved):
idk what's going on but lgtm!

exe/Plugins.hs
```diff
@@ -1,97 +1,173 @@
...
 #if brittany
-import           Ide.Plugin.Brittany               as Brittany
+import qualified Ide.Plugin.Brittany               as Brittany
+#endif
+
+data Log
+  = LogGhcIde GhcIde.Log
+  | LogExample Example.Log
+  | LogExample2 Example2.Log
+#if tactic
```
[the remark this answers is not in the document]
eddiemundo: This technique turns out to be a perfect fit for `Plugins.hs` since it's a hub module and probably will always be a hub module.

Taking it one step further most of the delegation boilerplate is gone but the leaf log value boilerplate goes up:
```haskell
-- Module1.hs
data Log = forall a. Pretty a => Log a
instance Pretty Log where
  pretty (Log a) = pretty a

func :: Recorder Log -> IO a
func = _

-- Module2.hs
import qualified Module1

data Log = forall a. Pretty a => Log a
instance Pretty Log where
  pretty (Log a) = pretty a

data LogSomething1 = LogSomething1 Int
instance Pretty LogSomething1 where
  pretty = _

data LogSomething2 = LogSomething2 Int
instance Pretty LogSomething2 where
  pretty = _

func = do
  Module1.func (cmap Log recorder)
  logWith recorder $ Log (LogSomething1 99)
```

eddiemundo:
I'm starting to think that the fact that the log values are of the form
```
LogModule1Constr (LogModule2Constr (LogModule3Constr (LogModule4Constr value)))
```
is a problem.

There may be a number of modules that don't need logging and hence only have `Log` datatype to wrap `Log` values for loggers to get contramapped and passed on to another module's function. This forces any conversion of the final `Log` value to backend value (`Doc`, or some lsp message, or something else) to torturously delegate to other modules conversion functions.

If there are `n` backends then we have `n` delegations chains where each chain is as long as however many modules traversed to get to the leaf function call that does the logging.

The advantage of building `Log` values such that they have a module constructor chain is to be able to consider the module chain information when doing the logging action, but I can't see that as being really too useful.

It makes more sense to me to instead pass `Recorder (WithPriority Doc)` and/or `Recorder (WithPriority LspMsg)` (and/or any other backend logger) to each module directly in a product type (maybe via `ShakeExtras`, or manually through function parameters). That way each module can decide on a local `Log -> Doc` or `Log -> LspMsg` (or `Log -> Whatever`) and pick one of the backend loggers and contramap it, then use it to log. There is no `logToBackendValue` delegation except for possibly special cases.

So... unless having module chain information in log values is valuable gonna simplify everything to not contramap between modules.

ghcide/src/Development/IDE/Types/Logger.hs
michaelpj: So, if I understand your analysis right, the `hie-bios` logs were actually broken for `ghcide`, since `hie-bios` unconditionally uses `hslogger`, and the `ghcide` entry point wasn't setting it up? That would also be true for the logs from `lsp`, which currently use `hslogger` (although I'm getting rid of it).

I'm okay with putting this in there as a temporary measure until we can get rid of `hslogger` in our dependencies. We should just clearly explain why we're doing it.

michaelpj:
> This forces any conversion of the final Log value to backend value (Doc, or some lsp message, or something else) to torturously delegate to other modules conversion functions.

This isn't *so* bad, right? Typically this means writing a `Pretty` instance and then calling `pretty` on the sub-logs. So in particular it's not like you need to remember lots of conversion functions.

> The advantage of building Log values such that they have a module constructor chain is to be able to consider the module chain information when doing the logging action, but I can't see that as being really too useful.

I think the advantage is rather locality of reasoning: you can see in that module what logs it emits and they're just there. We could imagine instead having a `GlobalLog` type that had one alternative for every log in the entire codebase. Then we would only have to do one conversion, but everything would depend on this type, and all the logs for each module would be mixed up together. So I think it's a maintainability thing rather than that it provides useful runtime information.

This makes me think that there are 3 (?) reasonable points on the spectrum:
1. Per-module `Log` types as now, embedded into each other.
2. Per-package `Log` types, with everything in that package. Still need to be embedded into each other e.g. to get plugin logs embedded into the HLS logs.
3. Truly global `Log` type, by picking a final representation type and just using that, e.g. `Doc`.

I think 3 would actually be quite annoying if we ever do want to emit a different type. The most practical example I can think of is providing a `--json-log` option to HLS that would emit all logs as JSON objects to make them easier to analyze. There it might actually be nice to go via the `ToJSON` instance of the log objects rather than having to post-process a `Doc`.

Maybe we should consider 2 instead? It's a little ugly, but given the number of packages we have, perhaps things still get scoped somewhat reasonably?

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -57,3 +91,180 @@ logTelemetry x = logPriority x Telemetry
...
+    Recorder
+      { logger_ = logger_ . f }
+
+cmap :: (a -> b) -> Recorder b -> Recorder a
+cmap = contramap
+
+cmapWithPrio :: (a -> b) -> Recorder (WithPriority b) -> Recorder (WithPriority a)
+cmapWithPrio f = cmap (fmap f)
+
+cmapIO :: (a -> IO b) -> Recorder b -> Recorder a
```
michaelpj, line 127: I'd probably call this `cmapM` (to match `co-log`), with the implicit constraint that we fix `m ~ IO` in our recorder. This is also fine though.

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -57,3 +91,180 @@ logTelemetry x = logPriority x Telemetry
...
+data WithPriority a = WithPriority { priority :: Priority, payload :: a } deriving Functor
+
+-- | Note that this is logging actions _of the program_, not of the user.
+--   You shouldn't call warning/error if the user has caused an error, only
+--   if our code has gone wrong and is itself erroneous (e.g. we threw an exception).
+data Recorder msg = Recorder
+  { logger_ :: forall m. (HasCallStack, MonadIO m) => msg -> m () }
+
+logWith :: (HasCallStack, MonadIO m) => Recorder (WithPriority msg) -> Priority -> msg -> m ()
+logWith recorder priority msg = withFrozenCallStack $ logger_ recorder (WithPriority priority msg)
```
michaelpj, line 104: I feel like I'm missing something: where does the call stack get used?

michaelpj, line 104: If you switch the order of the constructors of `WithPriority` then you could just make this
```
logWith :: ... => Recorder msg -> msg -> m ()
```
and have the call sites do this
```
logWith logger $ "hello" `WithPriority` Info
```
which is a bit more verbose but kind of cute :) Probably this is better, given this gets called a lot.

ghcide/test/exe/HieDbRetry.hs
```diff
@@ -5,25 +5,34 @@ import           Control.Concurrent.Extra     (Var, modifyVar, newVar, readVar,
...
 import           Test.Tasty                   (TestTree, testGroup)
 import           Test.Tasty.HUnit             (assertFailure, testCase, (@?=))
 
-makeLogger :: Var [(Priority, Text)] -> Logger
-makeLogger msgsVar = Logger $ \priority msg -> modifyVar msgsVar (\msgs -> pure ((priority, msg) : msgs, ()))
+data Log
+  = LogSession Session.Log
+  deriving Show
+
+makeLogger :: Var [Log] -> Recorder (WithPriority Log)
```
michaelpj, line 25: Probably not for this PR to keep it minimal, but there's a nice generic version of this backed by a `TQueue` or something, so you could have
```
mkTQueueLogger :: IO (Recorder msg, TQueue msg)
```
or something.

ghcide/test/exe/Main.hs
```diff
@@ -119,6 +118,21 @@ import           Test.Tasty.Ingredients.Rerun
...
+data Log 
+  = LogGhcIde Ghcide.Log 
+  | LogIDEMain IDE.Log
+
+instance Pretty Log where
+  pretty = \case
+    LogGhcIde log -> pretty log
+    LogIDEMain log -> pretty log
+
+logToDoc :: Log -> Doc a
```
michaelpj, line 134: these functions seem redundant now?

michaelpj (approved):
I think this is great, and we should merge it soon and iterate from there.

michaelpj:
A note from working on the `lsp` logging.

To get a logger that sends messages to the client, we need a `LogAction (LspM config) msg` rather than a `LogAction IO msg`. You can't get the former from the latter - not surprising, since we need the language server setup information to do it!

That means we'll inevitably be somewhat limited at what point we can start doing that kind of logging. But it would be nice for plugins to get this kind of logger. In particular, that means we probably can't do what you're doing in this PR and pass the log handler when we create the plugin descriptor - it's too early! Rather we'll probably need to pass it as another argument to a `PluginHandler`.

Anyway, I don't think we should do this as part of this PR, just something to think about.

michaelpj:
Ignore my previous message, it was in fact just wrong! :see_no_evil:

eddiemundo:
> I think 3 would actually be quite annoying if we ever do want to emit a different type. The most practical example I can think of is providing a --json-log option to HLS that would emit all logs as JSON objects to make them easier to analyze. There it might actually be nice to go via the ToJSON instance of the log objects rather than having to post-process a Doc.

The thing I had in mind is not necessarily post-processing the final type, but having multiple loggers of different final types bundled together.

So for the `--json-log` option we'd read that argument then create a `jsonLogger` logger  and a `regularLogger`, then we'd pass both loggers to the log sites so for instance
```
data Loggers = Loggers {
  logger :: Recorder (WithPriority Doc),
  jsonLogger :: Recorder (WithPriority Object)
}
```
if there is no `--json-log` we'd pass `mempty` for the logger.

Then this `Loggers` datatype value gets passed around. Each module still has its own `Log` datatype and a way to convert to these final representation types if necessary. So something like
```
data Log = LogThing Int

logToDoc = ...
logToJson = ...

logSomething :: Loggers -> m ()
logSomething Loggers{ jsonLogger } = do
  logWith logger $ logToDoc (LogThing Int)
  logWith jsonLogger $ logToJson (LogThing Int)
```
but I do see that we need to be considering which backend to use (or use all of them) at the call site instead of just use a single logger everywhere.

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: It never gets used, but it was in the original Pepe example, and at one point I used it to print the source location of the log call in the logs, but since things are different now not sure if that's still possible. I'll remove it.

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -57,3 +91,180 @@ logTelemetry x = logPriority x Telemetry
...
 noLogging :: Logger
 noLogging = Logger $ \_ _ -> return ()
+
+data WithPriority a = WithPriority { priority :: Priority, payload :: a } deriving Functor
+
+-- | Note that this is logging actions _of the program_, not of the user.
+--   You shouldn't call warning/error if the user has caused an error, only
+--   if our code has gone wrong and is itself erroneous (e.g. we threw an exception).
+data Recorder msg = Recorder
+  { logger_ :: forall m. (HasCallStack, MonadIO m) => msg -> m () }
```
pepeiborra, line 101: `Recorder` is a `Semigroup` where `<>` is used to compose multiple backends. It's also a `Monoid`.

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: I think in order to get `cmapM` working with the rank N type of `logger_` in `Recorder` I'd need an `UnliftIO` or `MonadBaseControl` constraint, and I remember having some issue there too.

I'm not sure if the rank N type of the `logger_` field is worth, but it does save carrying around a type parameter. Without the `logger_` having rank N type then using `co-log` directly would probably be better, although Pepe's initial example had a `tracer_` field bundled in which would be different from `co-log`.

ghcide/src/Development/IDE/Types/Logger.hs
pepeiborra: Oh no, I find it really useful to print call sites in log lines. What makes it not possible anymore?

eddiemundo: It is cooler to have `logWith` not know anything about `Priority`. I think in the end going back to
```
logDebug logger log = logWith logger $ WithPriority Debug log
```
will be better since I expect everything to use priorities/severities

ghcide/test/exe/HieDbRetry.hs
eddiemundo: That's more convenient. It never occurred to me that the recorder is kind of like the in pipe so you can create them both at same time.

ghcide/test/exe/Main.hs
eddiemundo: True will remove.

pepeiborra:
> This makes me think that there are 3 (?) reasonable points on the spectrum:

> 1. Per-module Log types as now, embedded into each other.
> 2. Per-package Log types, with everything in that package. Still need to be embedded into each other e.g. to get plugin logs embedded into the HLS logs.
> 3. Truly global Log type, by picking a final representation type and just using that, e.g. Doc.

Option 3 defeats the option of contra variant logging, doesn't it? You can no longer use the trick to test invariants as in https://github.com/haskell/haskell-language-server/pull/2558#issuecomment-1003634026

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: I see, so another argument to have a mega `Log` type is so we can compose backends easily.

pepeiborra: Not at all. What makes you think that?

Top level logger from Sigma IDE in case it helps:
```
  logger =
    VLog.logger (priorityFunction opts)
      <> scubaTracer
      <> (toTelemetry >$< eventlogTracer)
      <> if testInterfaceLoading opts
        then enforceInterfaceLoadingLogger myTid
        else mempty
```

eddiemundo: I think what made me think that is if we have
```
(<>) :: Logger msg -> Logger msg -> Logger msg
```
then `msg` need to be the same for both arguments.

If we have a logger that sends text to console, and a logger that sends json to client then maybe we start with
`Logger Text` and `Logger Json`, and so in order to make `Text` and `Json` the same we need to contramap them to the same type. To do this for all possible log values in the program we can embed the log values in each other like an exception hierarchy so that at some point every possible log value is contained somewhere under a `Log` type. Then we can `(<>)` the loggers that have been contramapped to this `Log` type.

Without a final `Log` type we could still combine backends but more locally, and not in a single place.

michaelpj: I think you might be talking about different things. I think Pepe maybe means "backends" to just be "recorders", in which case it already *has* those `Semigroup`/`Monoid` instances, I think?

pepeiborra: Oh, right, the instances are already there below a long comment. Sorry, should have scrolled down further

pepeiborra: @eddiemundo exactly, I guess this doesn't mean a mega `Log` type to me. It's a composable `Log` type.

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: OK back in it goes. I added a `callStack_ :: CallStack` field to `WithPriority` to do it easily.

I think I didn't remember correctly. Wasn't using `WithPriority` in the past, so was trying to create a chain of `HasCallStacks` from call site to output without storing a `CallStack`, but there was missing link somewhere, and I didn't spend any time to figure it out. So after `WithPriority` was added I got confused about why I removed it.

ghcide/src/Development/IDE/Types/Logger.hs
eddiemundo: Yes, I'll add a better comment.

michaelpj:
:wave: I'm returning to the `lsp` part of this work this weekend. I can wire it up to the existing logging, I think, but I'm still excited to get this in!

eddiemundo:
Fixed conflicts and if tests are good then I think this can be merged then can iterate

michaelpj:
Failure is just some warnings from building with `-Werror`.

michaelpj:
Okay, let's do this.

Reading back, I see this from @pepeiborra 

> For future work, we should have a logging backend that forwards some LSP messages (encoded as JSON values) to the LSP stream for testing invariants.

I'm not sure I quite understood this, could you make an issue so we don't forget?

PR #2560, opened 2022-01-02
Fix freeze cache key correctly

jneira:
* The problem with the freeze file in 9.2.1 was not the use of allow-newer: `cabal freeze` use the same solver as `cabal build` and it takes in account the same config. The problem was `cabal freeze` was being done before `cabal update` so it cant take in account the head.hackage updates
* So remove the freeze file as i did in #2552 is not needed but move the `cabal update` step

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2560"></a>

### DIFF
--- a/.github/workflows/flags.yml
+++ b/.github/workflows/flags.yml
@@ -99,6 +99,8 @@ jobs:
           echo "" || \
           echo 'WARNING: Could not produce the `freeze`.'
           echo ::set-output name=value::${{ hashFiles('cabal.project.freeze') }}
+          # Removing freeze file as it breaks builds with alternative flags
+          rm -rf cabal.project.freeze
 
       - name: Hackage sources cache
         uses: actions/cache@v2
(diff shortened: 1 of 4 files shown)


michaelpj (approved):

Anton-Latukha (approved):

.github/workflows/flags.yml
```diff
@@ -112,16 +118,12 @@ jobs:
           cache-name: compiled-deps
         with:
           path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
-          key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project.freeze') }}
+          key:  ${{ env.cache-name }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.INDEX_STATE }}-${{ steps.compute-cache-key.outputs.value  }}
```
Anton-Latukha, line 121: Thank you for catching this.

Anton-Latukha (approved):
The case is understandable.

PR #2562, opened 2022-01-03
Fix ci update hackage index

jneira:
* #2560 fixed the `cabal freeze` step putting `cabal update` before it
* But it had a unintended effect detected in the ghc-9.2 branch: the hackage index bump done by cabal update was being overwritten by the package sources cache restoring done immediately after
* Also reviewing ci logs i also checked the freeze file hash was not being computed. It seems the `${{ hasFiles('...') }}` is resolved before the script is run 🤦 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2562"></a>

### DIFF
--- a/.github/workflows/hackage.yml
+++ b/.github/workflows/hackage.yml
@@ -86,8 +86,7 @@ jobs:
       - name: "Ensure we will use hackage head"
         run: cabal update
 
-      - name: Compute the cache key
-        id: compute-cache-key
+      - name: Form the package list ('cabal.project.freeze')
         run: |
           cabal v2-freeze && \
           echo "" && \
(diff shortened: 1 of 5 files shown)


.github/workflows/hackage.yml
```diff
@@ -78,16 +78,32 @@ jobs:
           cache-name: hackage-sources
         with:
           path: ${{ env.CABAL_PKGS_DIR }}
-          key:          ${{ env.cache-name }}-${{ env.INDEX_STATE }}
+          key:  ${{ env.cache-name }}-${{ env.INDEX_STATE }}
           restore-keys: ${{ env.cache-name }}-
 
+      # To ensure we get the lastest hackage index and not relying on haskell action logic
```
Anton-Latukha, line 84: Note that specifically so far was not including `hackage.yml` into the dev loop. As wanted to keep it as stable & untouched as possible. To first iron-out setup in PR CI loop & then spread the results in release workflow.

.github/workflows/test.yml
```diff
@@ -159,16 +170,6 @@ jobs:
           cat 'cabal.project.freeze' && \
           echo '' || \
           echo 'WARNING: Could not produce the `freeze`.'
-          echo ::set-output name=value::${{ hashFiles('cabal.project.freeze') }}
-
```
Anton-Latukha, line 163: Can you recap on this detail, by "resolving `${{ hashFiles( ... ) }}` before the step", you likely meant this use to produce output inside the step.

Anton-Latukha:
[skip circleci]

Anton-Latukha (approved):

Anton-Latukha:
Please, consider [`hackage.yml`](https://github.com/haskell/haskell-language-server/pull/2562#discussion_r777954263) comment before merging.

Anton-Latukha:
It is also good that `hackage` store cache does not depend on the `freeze` file, so we can do this (load the source cache before determining the general cache key).

Anton-Latukha:
Yes. It is seen in https://github.com/haskell/haskell-language-server/runs/4701985268?check_suite_focus=true that `echo ::set-output name=value::${{ hashFiles('cabal.project.freeze') }}` was not producing the hash

.github/workflows/hackage.yml
jneira: I see, well, i think we have to make all uniform sooner or later.
I tried the workflow in my branch and it looks fine: https://github.com/jneira/haskell-language-server/runs/4695347296?check_suite_focus=true

.github/workflows/test.yml
Anton-Latukha: https://github.com/haskell/haskell-language-server/pull/2562#issuecomment-1004776728 pretty much added additional context to understand the message.

Anton-Latukha:
Nice.

Anton-Latukha:
Seems to work in `ghc-9.2`: https://github.com/haskell/haskell-language-server/runs/4703678203?check_suite_focus=true.

Issue #86, opened 2020-05-04
Install fails in windows for ghc >= 8.8.3 on floskell:  Access violation in generated code


tuomohopia:
I'm seeing the install fail on my system for some reason with just
```
--  While building package floskell-0.10.2 using:
      C:\sr\setup-exe-cache\x86_64-windows\Cabal-simple_Z6RU0evB_3.0.1.0_ghc-8.8.3.exe --builddir=.stack-work\dist\29cc6475 build --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 11
Progress 3/5

********************************************************************************
Building failed, Try running `stack clean` and restart the build
If this does not work, open an issue at
        https://github.com/haskell/haskell-language-engine
********************************************************************************

Error when running Shake build system:
  at want, called at src\Development\Shake\Internal\Args.hs:83:69 in shake-0.18.5-JIltN70Z6uA8zMbpBJImj0:Development.Shake.Internal.Args
* Depends on: hls-8.8.3
  at command, called at src\\Stack.hs:83:3 in hls-install-0.8.0.0-G5s3uly41eMHHYnugIYx2Q:Stack
* Raised the exception:
Development.Shake.command, system command failed
Command line: stack --stack-yaml=stack-8.8.3.yaml install --verbosity=info
Exit code: 1
```

I installed `haskell-language-server` by:
* First changing resolver in `shake.yaml` in `install` folder to: `lts-15.11` (GHC 8.8.3)
* Then running install via `stack ./install.hs hls-8.8.3`

I'm running on Windows 10 and `stack` version `2.3.1`.

jneira:
I am able to reproduce and the log building floskell says:
```
floskell                         > [10 of 12] Compiling Floskell.Pretty
floskell                         >
floskell                         > Access violation in generated code when executing data at 0xffffffff80e86b40
floskell                         >
floskell                         >  Attempting to reconstruct a stack trace...
floskell                         >
floskell                         >    Frame     Code address
floskell                         >  * 0x461dab0 0xffffffff80e86b40
floskell                         >  * 0x461dab8 0x2d006300680067
floskell                         >  * 0x461dac0 0x2e0038002e0038
floskell                         >  * 0x461dac8 0x690000005c0033
floskell                         >  * 0x461dad0 0x5c006e D:\bin\stack\x86_64-windows\ghc-8.8.3\bin\ghc.exe+0x1c006e
```
[...]

PR #2563, opened 2022-01-04
Reuse build setup using a dedicated github action

jneira:
* The cache handling and several workarounds were duplicated in 5 workflows, making changes cumbersome and error prone
* This pr leverages composite actions to extract out the common steps of setting up the build:
  * https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
* I've tried this in my repo and the workflows looks good for now:  https://github.com/jneira/haskell-language-server/actions?query=branch%3Areuse-ci+

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2563"></a>


Anton-Latukha:
Is this was allowed 8)?

I didn't knew yaml composite exists.

Would review in near time.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+inputs:
+  ghc:
+    description: "Ghc version"
+    required: true
+  cabal:
+    description: "Cabal version"
+    required: false
+    default: "3.6"
+  os:
+    description: "Operating system: Linux, Windows or macOS"
```
Anton-Latukha, line 12: Interesting. Seems like you discovered we fully abstracted from `ubuntu-latest`-kind of values into `runner.os`.

It would be such until installation & run would depend on OS versions, lets hope it would be invariant for a long time.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+    - if: inputs.ghc == '9.0.1' || inputs.ghc == '9.2.1'
+      name: (GHC 9.0/9.2) Use modified `cabal.project`
+      env:
+        GHCVER: ${{ inputs.ghc }}
+      run: |
+        # File has some protections preventing regular `rm`.
+        # (most probably sticky bit is set on $HOME)
+        # `&&` insures `rm -f` return is positive.
+        # Many platforms aslo have `alias cp='cp -i'`.
+        rm -f -v cabal.project && cp -v cabal-ghc${GHCVER//./}.project cabal.project
```
Anton-Latukha, line 45: Checking escaping in-build:
https://github.com/haskell/haskell-language-server/runs/4707587679?check_suite_focus=true#step:3:60:
```
removed 'cabal.project'
'cabal-ghc901.project' -> 'cabal.project'
```

Great, works as intended.

Anton-Latukha (approved):

Anton-Latukha:
For me, in the GitHub Action design is useful to note several things.

Great.

:sunglasses: :champagne: :clap: :smiley: :clap: :clap:

Anton-Latukha:
Looked at the several builds. What I saw looks normal ...

merging ...

Anton-Latukha:
Build was/is expecting:
Screenshot-2022-01-05-12:28:14
Restarted:
Screenshot-2022-01-05-12:28:54

But maybe the backend requirement names not aligned?

Anton-Latukha:
It may be due to `3.6` cabal setting removal from the matrix & so from the name:
Screenshot-2022-01-05-12:31:59

Screenshot-2022-01-05-12:33:05

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      shell: bash
+
+    - if: inputs.os == 'Windows' && inputs.ghc == '8.8.4'
+      name: (Windows,GHC 8.8) Modify `cabal.project` to workaround segfaults
```
michaelpj, line 49: Do we have an issue link for this? We should remember to remove this at some point...

Anton-Latukha:
This is actually what I arrived to with DHall, DHall solution essentially required to form a number of modules to reimport. But there is a whole hussle to convert it perfectly into GitHub `yaml` design caveats (` - '[ "ele.**", file.md, ... ]'` & stuff).

This way of solving under current & foreseeable future conditions is much neater.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+    - name: Hackage sources cache
+      uses: actions/cache@v2
+      env:
+        cache-name: hackage-sources
+      with:
+        path:  ${{ env.CABAL_PKGS_DIR }}
+        key:   ${{ env.cache-name }}-${{ env.INDEX_STATE }}
+        restore-keys: ${{ env.cache-name }}-
+
+    # To ensure we get the lastest hackage index without telying in the haskell action logic
```
michaelpj, line 88:
```suggestion
    # To ensure we get the lasest hackage index without relying on the haskell action logic
```

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      run: |
+        # Form: index-state: 2021-11-29T08:11:08Z
+        INDEX_STATE_ENTRY=$(grep index-state cabal.project)
+        # Form: 2021-11-29T08-11-08Z
+        INDEX_STATE1=$(echo "$INDEX_STATE_ENTRY" | cut -d' ' -f2 | tr ':' '-')
+        echo "INDEX_STATE=$INDEX_STATE1" >> $GITHUB_ENV
+      shell: bash
+
+    # We have to restore package sources before `cabal update`
+    # cause it overwrites the hackage index with the cached one
```
michaelpj, line 78:
```suggestion
    # because it overwrites the hackage index with the cached one
```

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      run: |
+        echo "CABAL_PKGS_DIR=C:\\cabal\\packages" >> $GITHUB_ENV
+      shell: bash
+    - if: ( inputs.os == 'Linux' ) || ( inputs.os == 'macOS' )
+      name: (Linux,macOS) Platform config
+      run: |
+        echo "CABAL_PKGS_DIR=~/.cabal/packages" >> $GITHUB_ENV
+      shell: bash
+
+    # Needs to be before Cache Cabal so the cache can detect changes to the modified cabal.project file
```
michaelpj, line 35:
```suggestion
    # Needs to be before the caching step so that the cache can detect changes to the modified cabal.project file
```

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+
+    - name: Form the package list ('cabal.project.freeze')
+      run: |
+        cabal v2-freeze && \
+        echo "" && \
+        echo 'Output:' && \
+        echo "" && \
+        cat 'cabal.project.freeze' && \
+        echo "" || \
+        echo 'WARNING: Could not produce the `freeze`.'
```
michaelpj, line 101: Hm, is there maybe a way to set `-e` globally so we can just fail if anything fails?

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+        cache-name: compiled-deps
+      with:
+        path: ${{ steps.HaskEnvSetup.outputs.cabal-store }}
+        key:  ${{ env.cache-name }}-${{ inputs.os }}-${{ inputs.ghc }}-${{ env.INDEX_STATE }}-${{ hashFiles('cabal.project.freeze') }}
+        restore-keys: |
+              ${{ env.cache-name }}-${{ inputs.os }}-${{ inputs.ghc }}-${{ env.INDEX_STATE }}-
+              ${{ env.cache-name }}-${{ inputs.os }}-${{ inputs.ghc }}-
+              ${{ env.cache-name }}-${{ inputs.os }}-
+
+    # We remove the freeze file cause it could interfere the build
```
michaelpj, line 117:
```suggestion
    # We remove the freeze file because it could interfere with the build
```

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+        echo "CABAL_PKGS_DIR=C:\\cabal\\packages" >> $GITHUB_ENV
+      shell: bash
+    - if: ( inputs.os == 'Linux' ) || ( inputs.os == 'macOS' )
+      name: (Linux,macOS) Platform config
+      run: |
+        echo "CABAL_PKGS_DIR=~/.cabal/packages" >> $GITHUB_ENV
+      shell: bash
+
+    # Needs to be before Cache Cabal so the cache can detect changes to the modified cabal.project file
+    - if: inputs.ghc == '9.0.1' || inputs.ghc == '9.2.1'
```
michaelpj, line 36: We could consider making this unconditional and just looking for the existence of a suffixed cabal.project for the current version. Then this would work automatically if we add/remove them in future. Perhaps too magical?

michaelpj (approved):
Deduplication looks amazing, just a few comments as I happened to be reading this file for the first time.

.github/actions/setup-build/action.yml
jneira: yeah, I think it is closed but we will remove when we drop support for ghc-8.8.4

.github/actions/setup-build/action.yml
jneira: latest?

.github/actions/setup-build/action.yml
jneira: afaik the shell has -e by default but we want the opposite thing: it should not fail so cache will fallback to only index state

.github/actions/setup-build/action.yml
michaelpj: :facepalm: yeah

.github/actions/setup-build/action.yml
michaelpj: But why? We can fail to produce the freeze file? Probably could do with a note explaining that if it's expected...

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+    - name: Hackage sources cache
+      uses: actions/cache@v2
+      env:
+        cache-name: hackage-sources
+      with:
+        path:  ${{ env.CABAL_PKGS_DIR }}
+        key:   ${{ env.cache-name }}-${{ env.INDEX_STATE }}
+        restore-keys: ${{ env.cache-name }}-
+
+    # To ensure we get the lastest hackage index without telying in the haskell action logic
```
jneira, line 88:
```suggestion
    # To ensure we get the latest hackage index without relying on the haskell action logic
```

.github/actions/setup-build/action.yml
jneira: good question, we had some issues producing it for bleeding-edge branches like ghc-9.2 but i *think* it should work now. However i am not sure if we have to be stricter and block ci. Maybe we could try the strict way and make it optional i we see there are more problems hard to fix quickly

michaelpj: SGTM.

Also, further dumb question: why do we need the freeze file? Given that we pin the index state, won't cabal produce a deterministic build plan anyway?

.github/actions/setup-build/action.yml
jneira: The issue was https://github.com/haskell/haskell-language-server/issues/86

.github/actions/setup-build/action.yml
jneira: i ve made it strict and trying the ghc-9.2 branch here: https://github.com/jneira/haskell-language-server/pull/59

jneira: It seems it is working: https://github.com/jneira/haskell-language-server/runs/4714665789?check_suite_focus=true

jneira: > SGTM.
> 
> Also, further dumb question: why do we need the freeze file? Given that we pin the index state, won't cabal produce a deterministic build plan anyway?

pr's can change dependencies removing, adding or updating bounds in any of the many .cabal files. This way we can have a dedicated cache taking in account those changes. We could hash all .cabal files though.

But otoh you can change things in the cabal.project (like constraints, flags,ghc option) which might affect dependencies. `cabal freeze` should record the relevant changes affecting dependencies. That way cosmetic or other changes not affecting dep resolution in cabal.project does not make miss the cache.

Anton-Latukha: Freeze file gets into account all projects & infers 1 package version set to rule them all.

Here `freeze` served the `caching` key detection purposes - so it was made fault-tolerant for the purpose of use.

There can be mutually exclusive dependencies, especially between `plugins`. Currently, most of them are enabled by default, so their package constraints should be solvable. But with the growing number of plugins - keeping all plugins package constraints up to date & mutually resolvable would be a challenge.

`freeze` in itself is a test, that all regular `build` targets in all projects can infer 1 set of versions of dependencies, so that is a dependency requirement test in itself & it probably should be made & placed so accordingly. Maybe now thinking on having CI to check with `freeze` for merge would seem viable.

And the additional thought - it seems a good strategy to check contributor PRs for consistency in update/dependency management. It would enforce also to have some backward compatibility of versions between projects, wich is both a good thing and may be a security concern, but security concern is when mainteiners can come into play. So for example, having the consistency means restricting from old versions in one place, would mean it can be at once removed in all other projects.

So indeed, having a `freeze` requirement seems as a feature.

.github/actions/setup-build/action.yml
jneira: Yeah, i considered it at somepoint but i forgot that possibility, thanks!

.github/actions/setup-build/action.yml
Anton-Latukha: @michaelpj
> Also, further dumb question: why do we need the freeze file? Given that we pin the index state, won't cabal produce a deterministic build plan anyway?

It is not a dumb question. It is a Cabal innerworks obscurities. `index-state` determines the Hackage deps API catalog, version of the dependency catalog available at that time. Hackage allows a lot of versions for every package at every moment, so the catalog is a tree structure `package/{set-of-all-versions}`. So under one `index-state` two build targets are free choose two mutually exclusive versions.
`freeze` requires all (`build`, *afaik*) package descriptions to be solvable as a whole.

jneira:
> All other workflows were not requiring this because they either have `post_job` & so merge checks depend on `post_job` and other workflows do not have matrix. (looking at this, this seems like a nice thing to abstract, `flags` probably needs to gain the `post_job` & CircleCI need also (because right now any GHC version change requires to readjust CircleCI merge checks) either to return 1 status or have a `post_job` that returns merge status product).

For now, i've changed the branch protection rules

PR #2565, opened 2022-01-05
support selection range lsp feature

### LINKED ISSUE #212: Implement selection range request lsp feature
Hi,

I started toying with the "Expand selection" (Alt + Shift + Right arrow by default on Linux) feature in VS Code and I noticed some quirks here and there. As usual, I am unsure if this is an HLS or Ghcide issue, sorry :sweat_smile: 

To reproduce:
```haskell
{-# LANGUAGE DataKinds #-}
module Foo where

import Prelude

newtype Foo = Foo {
    foo :: String
}

bar :: String -> Foo
bar "" = undefined
bar s = Foo {
    foo = s -- start with your cursor on this line, somewhere on the foo field
}

baz :: Bool -> ()
baz _ = ()
```

When using "Expand selection" feature several times, here's what happens:

1. Selects `foo`
2. Selects `foo = s -- start with your cursor on this line, somewhere on the foo field`
3. Selects `    foo = s -- start with your cursor on this line, somewhere on the foo field` (the left spaces are now included, but Github doesn't display them...)
4. Selects everything between the brackets excluded (including new lines)
5. Selects everything between the brackets included
6. :boom: Selects the pattern case of `bar` (i.e. does not include the empty string case)
7. :boom: Selects the whole file

IMHO 6 is weird, there should be additional steps:
* Between 5 and 6: Select the whole `Foo {}` expression
* TBD: Between 6 and 7: Select the whole implementation but not the signature (I am much less opinionated on this one)
* Between 6 and 7: Select the whole implementation, including the signature
* Between 6 and 7: Select all declarations (types, functions, etc.) but not imports
* Between 6 and 7: Select all declarations (types, functions, etc.) and imports, but not module declaration

kokobd:
this will fix #212

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2565"></a>

TODO
- [x] use the default structure of `HieAST` to generate selection ranges
- [x] point `lsp-types` version to master in both cabal and stack files
- [x] add tests
- [x] optimize the behavior for some special cases?
- [x] Fix CI (especially for GHC versions other than `8.10.7`, which I used during development)
- [x] update `lsp-types` to a new version on hackage before this is merged
- [x] update docs to include this plugin


cabal.project
```diff
@@ -38,6 +38,12 @@ package *
   ghc-options: -haddock
   test-show-details: direct
 
+source-repository-package
+    type: git
+    location: https://github.com/kokobd/lsp.git
```
jneira, line 43: minor thing, we could use lsp master, as your [pr ](https://github.com/haskell/lsp/pull/388) was merged into

michaelpj: Yeah, and we should do a release before this is done.

kokobd:
Currently the selection range is generated directly from the structure of `HieAST`, without considering what the ast node actually is. There are some bad cases:

1. When selecting a whole import statement, expanding the selection now won't select the whole area of all imports. Instead, the whole module (excluding the module header) is selected.
2. When selecting everything from a function definition (without signature), expanding the selection now won't select the function's signature and it's definition. Instead, the whole module (excluding the module header) is selected.

For case 1, there isn't a single node in ast containing all import statements. For case 2, signatures and definitions are distinct ast nodes (and actually they are not necessarily be placed nearby in source file)

If needed, special logic can be added to handle cases like this, but will incur extra complexity and compatibility issues (the actual structure of AST in GHC changes between versions?). Not sure whether to leave it as is, or handle the cases specially.

ghcide/src/Development/IDE/Plugin/HLS/GhcIde.hs
```diff
@@ -28,6 +29,7 @@ descriptors =
     CodeAction.fillHolePluginDescriptor "ghcide-code-actions-fill-holes",
     Completions.descriptor "ghcide-completions",
     TypeLenses.descriptor "ghcide-type-lenses",
+    SelectionRange.descriptor "ghcide-selection-range",
```
pepeiborra, line 32: Given that this has been set up as a new plugin, I would prefer if the plugin lived in its own package and ghcide didn't depend on it. This keeps ghcide small and makes for easier GHC upgrades (as mentioned in https://github.com/haskell/haskell-language-server/issues/212#issuecomment-1003618821)

ghcide/src/Development/IDE/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,76 @@
...
+    case filePathMaybe of
+        Nothing -> pure . Right . List $ []
+        Just filePath -> liftIO $ do
+            let (List positions) = _positions
+            selectionRanges <- runIdeAction "SelectionRange" (shakeExtras ide) $ getSelectionRanges filePath positions
+            pure . Right . List $ selectionRanges
+
+getSelectionRanges :: NormalizedFilePath -> [Position] -> IdeAction [SelectionRange]
+getSelectionRanges file positions = fmap (fromMaybe []) <$> runMaybeT $ do
+    (HAR{hieAst}, _) <- useE GetHieAst file
```
wz1000, line 54: You shouldn't ignore the `PositionMapping` and instead use `oldPositionToNew` to adjust all the output positions and `newPositionToOld` to adjust all the input positions.

kokobd: Thanks for pointing this out! I will fix this later.

ghcide/src/Development/IDE/Plugin/HLS/GhcIde.hs
kokobd: Ok, I will move this plugin to its own package. By the way, if I understand correctly, plugins in ghcide is somewhat 'legacy' and should be moved to their own packages eventually in the future if possible, right?

jneira: yeah, that is the long term plan, we should note it in the contributors guide

kokobd:
Recent progress:

+ move this plugin to a standalone package
+ add a basic test suite
+ apply `PositionMapping`
+ add the whole import block as a step in expanding/shrinking selection

Future work:

+ add the combination of function signature and its definition as a step in expanding/shrinking selection
+ fix CI

michaelpj:
> add the combination of function signature and its definition as a step in expanding/shrinking selection

FWIW I think it would be fine to merge a first version without this and then add it later. Up to you when you consider that this has enough features!

I'll try and do a `lsp-types` release this week so that's sorted...

ghcide/src/Development/IDE/Plugin/SelectionRange.hs
kokobd: There seems to be no such function as `oldPositionToNew` through. Having refered to other plugins, I use 
 functions like `toCurrentPosition` to apply position mapping to input and output. Resolving this conversation for now.

kokobd:
Currenly all the planned work is done, besides bumping lsp version. Marking it ready for review and waiting for GitHub actions result now.

jneira:
great work, I guess a LSP release would be useful /cc @michaelpj

kokobd:
A screen record:

https://user-images.githubusercontent.com/16440269/150301502-4c002605-9f8d-43f5-86d3-28846942c4ff.mov

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+descriptor plId = (defaultPluginDescriptor plId)
+    { pluginHandlers = mkPluginHandler STextDocumentSelectionRange selectionRangeHandler
+    }
+
+selectionRangeHandler :: IdeState -> PluginId -> SelectionRangeParams -> LspM c (Either ResponseError (List SelectionRange))
+selectionRangeHandler ide _ SelectionRangeParams{..} = do
+    let (TextDocumentIdentifier uri) = _textDocument
+    let filePathMaybe = toNormalizedFilePath' <$> uriToFilePath' uri
+    case filePathMaybe of
+        Nothing -> pure . Right . List $ []
```
michaelpj, line 59: Seems appropriate to return `Left` here?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+        Just filePath -> liftIO $ do
+            let (List positions) = _positions
+            selectionRanges <- runIdeAction "SelectionRange" (shakeExtras ide) $ getSelectionRanges filePath positions
+            pure . Right . List $ selectionRanges
+
+getSelectionRanges :: NormalizedFilePath -> [Position] -> IdeAction [SelectionRange]
+getSelectionRanges file positions = fmap (fromMaybe []) <$> runMaybeT $ do
```
michaelpj, line 66: Should we fail the request in some of the cases where we get `Nothing`? Clearly sometimes there are no appropriate selection ranges, but sometimes it might rather be that something went wrong trying to figure them out.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+  where
+    go acc (Node _ span [])       = fmap (span:) acc
+    go acc (Node _ span children) = concatMap (go (fmap (span:) acc)) children
+
+spansToSelectionRange :: [Span] -> Maybe SelectionRange
+spansToSelectionRange [] = Nothing
+spansToSelectionRange (span:spans) = Just $
+    SelectionRange {_range = realSrcSpanToRange span, _parent = spansToSelectionRange spans}
+
+simplifySpans :: [Span] -> [Span]
```
michaelpj, line 95: could use `nubOrd`? Technically it's a bit stronger than this, since it removes all duplicates, not just duplicates that are adjacent, but I think you could argue that you're only going to get adjacent duplicates anyway.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+simplifySpans = foldr go []
+  where
+    go x []     = [x]
+    go x (y:ys) = if x == y then y:ys else x:y:ys
+
+findSelectionRangesByPositions :: [SelectionRange] -> [Position] -> [SelectionRange]
+findSelectionRangesByPositions selectionRanges = fmap findByPosition
+  where
+    findByPosition p = fromMaybe SelectionRange{_range = Range p p, _parent = Nothing} $
```
michaelpj, line 104: personally I like type signatures for local bindings. Working out the return type of this requires staring at the RHS and the type signature of the main function...

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+    SelectionRange {_range = realSrcSpanToRange span, _parent = spansToSelectionRange spans}
+
+simplifySpans :: [Span] -> [Span]
+simplifySpans = foldr go []
+  where
+    go x []     = [x]
+    go x (y:ys) = if x == y then y:ys else x:y:ys
+
+findSelectionRangesByPositions :: [SelectionRange] -> [Position] -> [SelectionRange]
+findSelectionRangesByPositions selectionRanges = fmap findByPosition
```
michaelpj, line 102: I'd maybe just define the un-fmaped version and fmap it at the use site, but :shrug:

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+spansToSelectionRange (span:spans) = Just $
+    SelectionRange {_range = realSrcSpanToRange span, _parent = spansToSelectionRange spans}
+
+simplifySpans :: [Span] -> [Span]
+simplifySpans = foldr go []
+  where
+    go x []     = [x]
+    go x (y:ys) = if x == y then y:ys else x:y:ys
+
+findSelectionRangesByPositions :: [SelectionRange] -> [Position] -> [SelectionRange]
```
michaelpj, line 101: This could do with a comment, it's not totally obvious what it does: it finds the *first* selection range in the input whose range contains the given position (and then does that for the whole input list). Makes me wonder whether there could be two such selection ranges, does that matter?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+findSelectionRangesByPositions selectionRanges = fmap findByPosition
+  where
+    findByPosition p = fromMaybe SelectionRange{_range = Range p p, _parent = Nothing} $
+        find (isPositionInSelectionRange p) selectionRanges
+    isPositionInSelectionRange p SelectionRange{_range} =
+        let Range sp ep = _range in sp <= p && p <= ep
```
michaelpj, line 107: Hmm, we should probably export some range/position comparison utils from lsp-types...

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,107 @@
...
+            pure . Right . List $ selectionRanges
+
+getSelectionRanges :: NormalizedFilePath -> [Position] -> IdeAction [SelectionRange]
+getSelectionRanges file positions = fmap (fromMaybe []) <$> runMaybeT $ do
+    (HAR{hieAst, refMap}, positionMapping) <- useE GetHieAst file
+    positions' <- MaybeT . pure $ traverse (fromCurrentPosition positionMapping) positions
+    ast <- MaybeT . pure $ getAsts hieAst Map.!? (coerce. mkFastString . fromNormalizedFilePath) file
+    let ast' = runReader (preProcessAST ast) (PreProcessEnv refMap)
+    MaybeT . pure . traverse (toCurrentSelectionRange positionMapping) $
+        findSelectionRangesByPositions (astPathsLeafToRoot ast') positions'
```
michaelpj, line 72: These two lines in particular are tricky! A comment explaining what the idea is here would help

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
```diff
@@ -0,0 +1,107 @@
...
+                                                  realSrcSpanEnd,
+                                                  realSrcSpanStart)
+import           Development.IDE.GHC.Compat.Util (FastString)
+import           Prelude                         hiding (span)
+
+newtype PreProcessEnv a = PreProcessEnv
+    { preProcessEnvRefMap :: RefMap a
+    }
+
+-- | Make the AST more suitable for generating selection range.
```
michaelpj, line 33: Could briefly explain the changes and why we need them?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
```diff
@@ -0,0 +1,107 @@
...
+    merge [x] = Just x
+    merge xs  = Just (createVirtualNode xs)
+
+nodeIsImport :: HieAST a -> Bool
+nodeIsImport = isAnnotationInAstNode ("ImportDecl", "ImportDecl")
+
+createVirtualNode :: [HieAST a] -> HieAST a
+createVirtualNode nodes = mkAstNode (NodeInfo mempty mempty mempty) span' nodes
+  where
+    span' = mkRealSrcSpan (minimum locations) (maximum locations)
```
michaelpj, line 54: https://hackage.haskell.org/package/ghc-9.2.1/docs/GHC-Types-SrcLoc.html#v:combineRealSrcSpans

michaelpj: Maybe you can just fold up the source spans with that.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
```diff
@@ -0,0 +1,107 @@
...
+            Nothing   -> node':prev:others
+            Just comb -> comb:others
+
+mergeNearbySigDef :: RefMap a -> (HieAST a, HieAST a) -> Maybe (HieAST a)
+mergeNearbySigDef refMap (n1, n2) = do
+    if not (("TypeSig", "Sig") `isAnnotationInAstNode` n1 && ("FunBind", "HsBindLR") `isAnnotationInAstNode` n2)
+    then Nothing
+    else do
+        typeSigId <- identifierForTypeSig n1
+        refs <- Map.lookup typeSigId refMap
```
michaelpj, line 76: There's something clever going on here which is why you need this map, but I'm not sure what it is. Comment?

michaelpj (commented):
Code looks pretty good, could just do with some comments where there's cleverness happening. Might be good to get another look from @wz1000 for the various HIE AST manipulations.

michaelpj:
Published `lsp-types-1.4.0.1` with the fix.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: yeah, it should be. I will reconsider the error handling part

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: The error handling should be done better. I will reconsider that.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
kokobd: oh, this is really helpful

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: Oh, actually, this is a premature optimization. Though we only need to remove adjacent duplicates, as the list is not too long, the performance difference is small. For this case, duplicates should only occur in nearby positions, so the result will be same. This complexity is indeed unnecessary, I will replace it with `nubOrd`.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: This function filters the `[SelectionRange]`, preserving all `SelectionRange`s containing at least one of the `Positions`. (will add to comment later)
So, at first glance, I think there might exist better algorithm for this, at least better than the currently implemented simple traversal. That's why I keep it as a standalone function, rather than moving fmap to the caller site.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
kokobd: This is used to handle some cases not covered by the default tree structure of `HieAST`. Will supply more detailed explaination in comment.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
kokobd: Yes, indeed, this need a detailed explaination

michaelpj:
When you get back to this: we just merged some documentation improvements, so could you please:
- Update https://github.com/haskell/haskell-language-server/blob/master/docs/features.md like for the other features
- Add the new plugin to the table on https://github.com/haskell/haskell-language-server/blob/master/docs/supported-versions.md

kokobd:
Oh, sure. I will update docs in a future commit

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: @michaelpj Thanks for reading it so carefully. You got it right, the previous description is not accurate. I added more detailed explaination in the comment.

The performance concern makes sense somehow. If someone selects hundreds of positions in a batch edit, for a file containing tens of thousands of leaf nodes, it might become slow quickly. And will block the user's editing.

It is probably too early to optimize for things like that, for the very first version of this plugin. But at least for the possible optimizations I spotted during writing code, I would like to leave proper space for them. i.e. making the related code grouped together.

kokobd:
Recent update:

+ upgrade `lsp-types` to `1.4.0.1` (that version contains a bugfix)
+ add some comments to the code
+ return `ResponseError` incase of failures, instead of simply returning an empty list previously
+ list the plugin in Sphinx doc

Maybe it's ready for another review, to check if things are made right.

plugins/hls-haddock-comments-plugin/test/testdata/test/testdata/QualFunction.expected.hs
```diff
@@ -0,0 +1,6 @@
+module QualFunction where
+
+f :: (Show a, Show b) => a -- ^
```
michaelpj, line 3: Did you mean to include this?

plugins/hls-selection-range-plugin/hls-selection-range-plugin.cabal
```diff
@@ -0,0 +1,59 @@
...
+
+description:
+  Please see the README on GitHub at <https://github.com/haskell/haskell-language-server#readme>
+
+license:            Apache-2.0
+license-file:       LICENSE
+author:
+  https://github.com/haskell/haskell-language-server/contributors
+
+maintainer:
```
michaelpj, line 15: I think we're encouraging people to list themselves as author/maintainer in such places!

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
michaelpj: Hm yes, I hadn't even thought about performance. In theory this is O(n*m) for the two input lists, but TBH I agree that they're almost certainly small and it doesn't matter.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
```diff
@@ -0,0 +1,156 @@
...
+
+nodeIsImport :: HieAST a -> Bool
+nodeIsImport = isAnnotationInAstNode ("ImportDecl", "ImportDecl")
+
+createVirtualNode :: [HieAST a] -> Maybe (HieAST a)
+createVirtualNode [] = Nothing
+createVirtualNode children = Just $ mkAstNode (NodeInfo mempty mempty mempty) span' children
+  where
+    span' :: RealSrcSpan
+    span' = foldl1' combineRealSrcSpans . fmap nodeSpan $ children
```
michaelpj, line 76: Nit, but we're trying to avoid partial functions in HLS. You could make `createVirtualNode` take a `NonEmpty,` then it won't need to return `Maybe` and you can use a safe `fold1`.

michaelpj (approved):
Couple of nits, otherwise LGTM. Would still be nice to get a review from someone who knows about the HIE file stuff, but let's not block it on that.

plugins/hls-haddock-comments-plugin/test/testdata/test/testdata/QualFunction.expected.hs
kokobd: I guess the author of `hls-haddock-comments-plugin` forgot to commit this file... And the CI will never discover it, because the test will simply pass if there is no golden file. It is generated because I have run all tests locally.
As the test passes now, including this won't hurt, I think?

plugins/hls-selection-range-plugin/hls-selection-range-plugin.cabal
kokobd: Oh, that's great!

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
kokobd: Nice advice!

plugins/hls-haddock-comments-plugin/test/testdata/test/testdata/QualFunction.expected.hs
michaelpj: Oh I see, it's an expected file. Then yes, check it in!

kokobd:
From a new contributor's perspective, before reaching this point, most of the time was spent on figuring out how to use `HieAST a`. I implemented the ast manipulating functions through trial and error, so they may have many other bad cases not discovered. Hope it will have better docs in the future.

jneira:
> From a new contributor's perspective, before reaching this point, most of the time was spent on figuring out how to use `HieAST a`. I implemented the ast manipulating functions through trial and error, so they may have many other bad cases not discovered. Hope it will have better docs in the future.

Many thanks for the observation, that is a ghc type, right? i wonder what could we do here to improve the knowledge about until ghc get better docs

jneira (approved):
lgtm, many thanks for including anew one whole feature!

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/ASTPreProcess.hs
```diff
@@ -61,19 +63,18 @@ mergeImports node = pure $ node { nodeChildren = children }
...
 
-createVirtualNode :: [HieAST a] -> Maybe (HieAST a)
-createVirtualNode [] = Nothing
-createVirtualNode children = Just $ mkAstNode (NodeInfo mempty mempty mempty) span' children
+createVirtualNode :: NonEmpty (HieAST a) -> HieAST a
+createVirtualNode children = mkAstNode (NodeInfo mempty mempty mempty) span' (NonEmpty.toList children)
   where
     span' :: RealSrcSpan
-    span' = foldl1' combineRealSrcSpans . fmap nodeSpan $ children
+    span' = foldl1 combineRealSrcSpans . fmap nodeSpan $ children
```
michaelpj, line 77: I meant this one! https://hackage.haskell.org/package/semigroupoids-5.3.7/docs/Data-Semigroup-Foldable.html#v:fold1

The reason I'm being picky is that at some point I'm hoping we can just ban e.g. `foldl1` from `Foldable`. It is safe here, but it's easier if we can rely on people using the safe one from `Foldable1` when it's safe.

kokobd: Oh, thanks, I didn't know this function. Changed it to `foldlM1` in `Foldable1` though, as `fold1` requires a semigroup instance, and using the existing `Identity` monad is easier.

kokobd:
> Many thanks for the observation, that is a ghc type, right? i wonder what could we do here to improve the knowledge about until ghc get better docs

Oh, it lives in package `ghc`, but its module name `HieTypes` seems to suggest this is for the sake of **h**askell **i**de **e**ngine (the previous name of hls?), so I thought it is in the scope of haskell language server. (correct me if I'm wrong)

This is another topic though, but I think it would be helpful to have these information somewhere (or maybe it exists, but not found by me):

+ Difference between `GetParsedModule` and `GetHieAST`. Which one to pick up?
+ The possible values of those `FastString`s in `HieAST`'s annotations. (later I knew this string is generated from ghc's actual ast type, like `LHsDecl`)
+ The mysterious type 'a' in `HieAST a`. (`Type` & `TypeIndex`)

kokobd:
CI is green now, maybe it's ready for a merge?

pepeiborra:
> > Many thanks for the observation, that is a ghc type, right? i wonder what could we do here to improve the knowledge about until ghc get better docs
> 
> Oh, it lives in package `ghc`, but its module name `HieTypes` seems to suggest this is for the sake of **h**askell **i**de **e**ngine (the previous name of hls?), so I thought it is in the scope of haskell language server. (correct me if I'm wrong)
> 
Ping @wz1000 for this question

> * Difference between `GetParsedModule` and `GetHieAST`. Which one to pick up?

That's an interesting question. `HieAst` is designed to replace the need for the parsed AST in Haskell tooling, so I would say prefer `GetHieAST` when possible.

> * The possible values of those `FastString`s in `HieAST`'s annotations. (later I knew this string is generated from ghc's actual ast type, like `LHsDecl`)
> * The mysterious type 'a' in `HieAST a`. (`Type` & `TypeIndex`)

Ping @wz1000 for those questions.

plugins/hls-selection-range-plugin/hls-selection-range-plugin.cabal
```diff
@@ -0,0 +1,57 @@
...
+
+license:            Apache-2.0
+license-file:       LICENSE
+author: kokobd
+maintainer: kokobd
+
+category:           Development
+build-type:         Simple
+extra-source-files:
+  LICENSE
```
pepeiborra, line 18: The `testdata` files need to be included here

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,144 @@
...
+import           Prelude                                 hiding (span)
+
+descriptor :: PluginId -> PluginDescriptor IdeState
+descriptor plId = (defaultPluginDescriptor plId)
+    { pluginHandlers = mkPluginHandler STextDocumentSelectionRange selectionRangeHandler
+    }
+
+selectionRangeHandler :: IdeState -> PluginId -> SelectionRangeParams -> LspM c (Either ResponseError (List SelectionRange))
+selectionRangeHandler ide _ SelectionRangeParams{..} = do
+    liftIO $ logInfo logger $ "requesting selection range for file: " <> T.pack (show uri)
```
pepeiborra, line 64: Should this be `logDebug`?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,144 @@
...
+findSelectionRangesByPositions :: [SelectionRange] -- ^ all possible selection ranges
+                               -> [Position] -- ^ requested positions
+                               -> [SelectionRange]
+findSelectionRangesByPositions selectionRanges = fmap findByPosition
+    {-
+        Performance Tips:
+        Doing a linear search from the first selection range for each position is not optimal.
+        If it becomes too slow for a large file and many positions, you may optimize the implementation.
+        At least we don't need to search from the very beginning for each position, if the are sorted firstly.
+        Or maybe we could apply some techniques like binary search?
```
pepeiborra, line 135: Yes please, do consider at least benchmarking this in a follow-up PR. We do care about performance!  Unfortunately the benchmark suite doesn't currently support plugins (it's a benchmark suite for `ghcide`, not HLS). 

I wonder if binary search is really the solution here.  I suspect `HieAst` is isomorphic to a position indexed search tree, so the optimal approach would take advantage of this to find the solution  without building a list of all selection ranges.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -0,0 +1,144 @@
...
+    positions' <- maybeToExceptT "fail to apply position mapping to input positions" . MaybeT . pure $
+        traverse (fromCurrentPosition positionMapping) positions
+
+    ast <- maybeToExceptT "fail to get ast for current file" . MaybeT . pure $
+        -- in GHC 9, the 'FastString' in 'HieASTs' is replaced by a newtype wrapper around 'LexicalFastString'
+        -- so we use 'coerce' to make it work in both GHC 8 and 9
+        getAsts hieAst Map.!? (coerce . mkFastString . fromNormalizedFilePath) file
+
+    let ast' = runReader (preProcessAST ast) (PreProcessEnv refMap)
+    let selectionRanges = findSelectionRangesByPositions (astPathsLeafToRoot ast') positions'
```
pepeiborra, line 93: `(astPathsLeafToRoot ast'` is expensive, so we should cache its result and avoid doing it on every call. Please consider moving it to a new build rule `GetSelectionRanges`.

pepeiborra (commented):
Just a few more comments

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: Emm, yep, this may not be appropriate to place in info log

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: For performance issues, I will benchmark for large files in a future PR. As at least this action works without any delay in a file with 1000 lines on my local machine (intel 11th gen chip, not a fast one), so maybe it's ok for now?

pepeiborra: Do we understand at least how often the LSP client requests the selection ranges? Is it done on cursor move, or only when the user requests a selection expand?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: Oh, that's another optimization I didn't come up with. So if I get it correctly, there are two known optimizations now. (let n represents the number of selection ranges, m represents the number of positions)

+ For each position, we may treat HieAST as a position indexed tree to search it in O(log(n)).
+ For all positions, a searched position will narrow the search range for other positions. If we search the middle position first, then all positions before this position may only need to search in half of the tree (if both the tree and positions are balanced)

So the final complexity will be O(log(n) * log(m)).

If this looks right, I will update the comment to reflect these thoughts.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: > Do we understand at least how often the LSP client requests the selection ranges? Is it done on cursor move, or only when the user requests a selection expand?

Only when the user requests a selection expand. Even lesser frequent than that, in fact. If a user start expanding/shrinking from one position, and doesn't change the selection area manually, then this lsp method will only be called once. (I think that's why each `SelectionRange` is a linked list)

pepeiborra: Ok, if this is the case then the need to optimize is much less pressing.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
michaelpj: Right, at the moment (I think) you build all the possible selection ranges in the document, and then filter them by position, when you could fuse that into one pass and only build the paths which contain the target position or something.

kokobd: Maybe it's worth noting that, as selection ranges have shared parents, building up the list is O(n), where n is the number of nodes in the original AST. `astPathsLeafToRoot` basically clone the tree and make every pointer reversed. And functions in `ASTPreProcess` need to traverse through the AST, it's also O(n). So maybe this is not the only performance bottleneck, and we will need furthur benchmarks to make a conclusion.

eddiemundo: Not that it matters much but I think laziness saves a 1/2 constant factor on average for building the root to leaf paths. I think this holds true for the preprocessing phase too even with the `reverse` `foldl'` but not certain.

So if the preprocessing phase is lazy enough, then I think if we needed to upgrade the algorithm applying it to only `astPathsLeafToRoot` might suffice.

kokobd:
Added another issue #2632 to track the performance optimization

kokobd:
It seems that some recent changes to nix build files in master branch break this. Looking into that.

michaelpj:
Thanks again!

PR #2567, opened 2022-01-05
Add support for ghc 9.0.2

### LINKED ISSUE #2541: Please, support GHC 9.0.2

jneira:
* It misses the nix config
* We have to check the gitlab build
* Closes #2541 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2567"></a>


.github/workflows/hackage.yml
```diff
@@ -111,7 +112,7 @@ jobs:
           echo "allow-newer: Chart-diagrams:diagrams-core, SVGFonts:diagrams-core," >> cabal.project
 
       - name: "Add temporary needed allow-newer for ghc-9.0"
```
michaelpj, line 114: Just while I'm here: for this and the above step, don't we already have modified `cabal.project` files for exactly this kind of thing?

.github/workflows/test.yml
```diff
@@ -144,7 +147,7 @@ jobs:
 
         run: cabal test wrapper-test --test-options="$TEST_OPTS --rerun-log-file .tasty-rerun-log-wrapper" || cabal test wrapper-test --test-options="$TEST_OPTS  --rerun-log-file .tasty-rerun-log-wrapper" || cabal test wrapper-test --test-options="$TEST_OPTS  --rerun-log-file .tasty-rerun-log-wrapper"
 
-      - if: matrix.test && matrix.ghc != '9.0.1'
+      - if: matrix.test && matrix.ghc != '9.0.1' && matrix.ghc != '9.0.2'
         name: Test hls-brittany-plugin
         run: cabal test hls-brittany-plugin --test-options="$TEST_OPTS" || cabal test hls-brittany-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-brittany-plugin --test-options="$TEST_OPTS"
```
michaelpj, line 152: Can we define a `cabal test` command that abstracts this stuff out. Ouch...

.github/workflows/test.yml
```diff
@@ -144,7 +147,7 @@ jobs:
 
         run: cabal test wrapper-test --test-options="$TEST_OPTS --rerun-log-file .tasty-rerun-log-wrapper" || cabal test wrapper-test --test-options="$TEST_OPTS  --rerun-log-file .tasty-rerun-log-wrapper" || cabal test wrapper-test --test-options="$TEST_OPTS  --rerun-log-file .tasty-rerun-log-wrapper"
 
-      - if: matrix.test && matrix.ghc != '9.0.1'
+      - if: matrix.test && matrix.ghc != '9.0.1' && matrix.ghc != '9.0.2'
```
michaelpj, line 150: I wonder what would happen if we did this unconditionally but set the test suite to not be buildable on GHC 9 :thinking: might error, might just succeed...

.github/workflows/hackage.yml
jneira: well, it is a long story:
* we have to avoid any cabal.project tweaks (`allow-newer`, `source-repository-packages`) as hackage doesnt see them and without them the hackage build fails
* we want to make the packages tests and benchmarks work in hackage
* cabal solver needs all packages (including tests and benchmarks as we want enable them in general) be "solvable" although you dont build them effectively (like brittany for ghc-9.0.1)

* so we have to replicate the minimum possible of cabal.project tweaks in a controlled way to make the build work

But maybe there is another way to do it, o revise the initial premises

.github/workflows/test.yml
jneira: maybe with a shell routine? or you are thinking in other possiblesolution?

.github/workflows/test.yml
jneira: hmm i think it is good to build it even if the tests are gonna fail, or it will rot even with compile errors.
We are building all the test suites (with `tests: true`) and cabal doesnt have `runnable: fase` 😉

fendor:
There is no release of hie-bios with support for GHC 9.0.2 yet. Going to make one now.

.github/workflows/hackage.yml
michaelpj: Right, but this is for GHC 9, right? Where we *do* have `cabal.project` tweaks. Basically, I don't see how adding them here can be any better than adding them in `cabal-ghc9.project` or whatever?

.github/workflows/test.yml
michaelpj: Oh hm, we would actually need a function here. I was thinking you could do something like
```
echo "CABAL_TEST=cabal test ..." >> $GITHUB_ENV
```
but I don't know if you can do something like that for bash functions :thinking:

michaelpj: Ugly version would be to commit a small script that does that.

.github/workflows/test.yml
michaelpj: Yeah, fair.

jneira: An alternative is mark them as ignored or broken for that hls version using tasty modifiers, but we are not using them that way for now

.github/workflows/hackage.yml
jneira: Hmm, not sure if i am understanding, those tweaks are already in cabal-ghc901.project but we dont want to use the entire file, only the things needed by the solver which we know will not make the hackage build fail.
For example we are not building brittany for ghc-9, neither in the project nor in hackage so it is safe to add the `allow-newer`'s of brittany to make happy the solver cause the hackage build for ghc-9 of brittany is disabled by flags. 

Another alternative could be comment out the not buildable packages (like we do temporary, see for example the actual `cabal-ghc921.project`) but it would make more complex the actual `echo "packages: . ../../* ../../plugins/*" > cabal.project`

michaelpj: Ah I see, I think I just didn't understand the point either. Let me think about it a bit, and anyway it's irrelevant to this PR :)

wz1000:
We don't need a whole new `cabal.project` for GHC 9.0.2. The one for 9.0.1 will suffice with a few small tweaks.

pepeiborra:
Is there any way to avoid introducing a `cabal-ghc902.project` file? I see two reasonable options:
- Keep it, and delete `cabal-ghc901.project`
- Rename `cabal-ghc901.project` to `cabal-ghc9.project` and make it work for both

thomasjm:
Thanks for putting this together! FWIW I tried to build it with `stack build --stack-yaml stack-9.0.2.yaml` and it doesn't build. 

I had to add `unix` to `extra-deps`, and then it hit errors building `ghc-parser-ex`, `hie-bios`, `ormolu`, and `fourmolu`. Also a GHC panic building `primitive-unlifted`.

jneira:
> Thanks for putting this together! FWIW I tried to build it with `stack build --stack-yaml stack-9.0.2.yaml` and it doesn't build.
> 
> I had to add `unix` to `extra-deps`, and then it hit errors building `ghc-parser-ex`, `hie-bios`, `ormolu`, and `fourmolu`. Also a GHC panic building `primitive-unlifted`. [Logs](https://pastebin.com/nzu9y0nh)

thank for trying, circleci is asking unix too, will try to fix it asap

jneira:
> Is there any way to avoid introducing a `cabal-ghc902.project` file? I see two reasonable options:
> 
>     * Keep it, and delete `cabal-ghc901.project`
> 
>     * Rename `cabal-ghc901.project` to `cabal-ghc9.project` and make it work for both

i am for renaming to `cabal-ghc9.project`, there is a little chance of make the 9.0.1 build fail or fail in hackage due to 9.0.2 specific tweaks but i suppose is small enough to worht the deduplication

jneira:
after adding unix ci reached the error commented by @thomasjm for primitive-unlifted:

```
primitive-unlifted                > [ 7 of 18] Compiling Data.Primitive.Unlifted.MVar.Primops
primitive-unlifted                > ghc: panic! (the 'impossible' happened)
primitive-unlifted                >   (GHC version 9.0.2:
primitive-unlifted                > 	findSlot
primitive-unlifted                >   Can't find slot
primitive-unlifted                >   [PtrUnliftedSlot]
primitive-unlifted                >   [PtrLiftedSlot]
primitive-unlifted                >   Call stack:
primitive-unlifted                >       CallStack (from HasCallStack):
primitive-unlifted                >         callStackDoc, called at compiler/GHC/Utils/Outputable.hs:1230:37 in ghc:GHC.Utils.Outputable
primitive-unlifted                >         pprPanic, called at compiler/GHC/Types/RepType.hs:217:9 in ghc:GHC.Types.RepType
primitive-unlifted                > 
primitive-unlifted                > Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

jneira:
> after adding unix ci reached the error commented by @thomasjm for primitive-unlifted:
> 

issue in the ghc issue tracker: https://gitlab.haskell.org/ghc/ghc/-/issues/20908 :worried:

wz1000:
The solution is to use primitive-unlifted < 1.0

`primitive-unflifted-1.0` does some shady things which aren't semantically valid.

jneira:
> The solution is to use primitive-unlifted < 1.0
> 

thanks for the suggestion, it is not a direct dependency of hls so i guess it will be hard to make the hackage build works although we can force that version using cabal.prokect/stack.yaml

wz1000:
See https://github.com/metrix-ai/primitive-extras/issues/6

jneira:
Ok we hit the hie-bios error with cabal:

```
 src/HIE/Bios/Ghc/Gap.hs:260:7: error:
Error:     • Couldn't match expected type ‘IncludeSpecs’
                  with actual type ‘[String] -> IncludeSpecs’
    • Probable cause: ‘IncludeSpecs’ is applied to too few arguments
      In the ‘includePaths’ field of a record
      In the expression:
        df
          {includePaths = IncludeSpecs
                            (map f $ G.includePathsQuote (includePaths df))
                            (map f $ G.includePathsGlobal (includePaths df))}
      In an equation for ‘mapOverIncludePaths’:
          mapOverIncludePaths f df
            = df
                {includePaths = IncludeSpecs
                                  (map f $ G.includePathsQuote (includePaths df))
                                  (map f $ G.includePathsGlobal (includePaths df))}
    |
260 |       G.IncludeSpecs
    |       ^^^^^^^^^^^^^^...
```
so we have to upgrade hie-bios

In stack we got another error, already pointed out by @thomasjm :

```
ghc-lib-parser-ex                 > /tmp/stack-bcedea5ffccdf320/ghc-lib-parser-ex-9.0.0.6/src/Language/Haskell/GhclibParserEx/GHC/Settings/Config.hs:18:1: error:
ghc-lib-parser-ex                 >     Could not load module ‘GHC.Platform’
ghc-lib-parser-ex                 >     It is a member of the hidden package ‘ghc-boot-9.0.2’.
ghc-lib-parser-ex                 >     Perhaps you need to add ‘ghc-boot’ to the build-depends in your .cabal file.
ghc-lib-parser-ex                 >     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
ghc-lib-parser-ex                 >    |
ghc-lib-parser-ex                 > 18 | import GHC.Platform
ghc-lib-parser-ex                 >    | ^^^^^^^^^^^^^^^^^^^
```

wz1000:
Can we also drop stack-9.0.1.yaml? GHC 9.0.1 is not a version anyone should be using.

jneira:
> Can we also drop stack-9.0.1.yaml? GHC 9.0.1 is not a version anyone should be using.

well, we are planning remove all (but one or two) together and i would prefer to keep them consistent until that, having one stack.yaml for each supported ghc (we have also have stack-8.10.6.yaml). 
As noted previously it is needed to make work the install script and the 9.0.2 is still recent so users have to upgrade their instalaltions

jneira:
ghc-lib-parser-ex fails for stack with:

```
ghc-lib-parser-ex                 > /tmp/stack-83379216794bc4a2/ghc-lib-parser-ex-9.0.0.6/src/Language/Haskell/GhclibParserEx/GHC/Settings/Config.hs:18:1: error:
ghc-lib-parser-ex                 >     Could not load module ‘GHC.Platform’
ghc-lib-parser-ex                 >     It is a member of the hidden package ‘ghc-boot-9.0.2’.
ghc-lib-parser-ex                 >     Perhaps you need to add ‘ghc-boot’ to the build-depends in your .cabal file.
ghc-lib-parser-ex                 >     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
ghc-lib-parser-ex                 >    |
ghc-lib-parser-ex                 > 18 | import GHC.Platform
ghc-lib-parser-ex                 >    | ^^^^^^^^^^^^^^^^^^^
....
--  While building package ghc-lib-parser-ex-9.0.0.6 (scroll up to its section to see the error) using:
      /root/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_3.4.1.0_ghc-9.0.2 --builddir=.stack-work/dist/x86_64-linux/Cabal-3.4.1.0 build --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
```

but cabal is building succesfully ghc-lib-parser-ex-9.0.0.5

michaelpj:
> but cabal is building succesfully ghc-lib-parser-ex-9.0.0.5

That seems surprising since it's deprecated on hackage and the base bounds are set to be unsatisfiable :thinking:

jneira:
> > but cabal is building succesfully ghc-lib-parser-ex-9.0.0.5
> 
> That seems surprising since it's deprecated on hackage and the base bounds are set to be unsatisfiable thinking

you may be right, i dont find the build where i thought i've seen it

michaelpj:
I think you need to add `allow-newer: ghc-typelits-natnormalise:ghc-bignum` to `ghcide/test/data/plugin-knownnat/cabal.project`.

jneira:
> I think you need to add `allow-newer: ghc-typelits-natnormalise:ghc-bignum` to `ghcide/test/data/plugin-knownnat/cabal.project`.

yeah, thanks for the suggestion, i wonder if we could change it to not have similar problems in the future

.github/workflows/build.yml
```diff
@@ -54,12 +55,18 @@ jobs:
           echo -e 'package blaze-textual\n    flags: +integer-simple' >> cabal.project.local
         fi
 
-    - name: (GHC 9.0) Use modified cabal.project for GHC 9.0
-      if: ${{ matrix.ghc == '9.0.1' }}
-      run: cp cabal-ghc901.project cabal.project
-    - name: Use modified cabal.project for ghc9.2
-      if: ${{ matrix.ghc == '9.2.1' }}
-      run: cp cabal-ghc921.project cabal.project
+    - name: Use modified cabal.project
```
michaelpj, line 58: Wait, shouldn't this just be using the setup-build workflow from above?

cabal-ghc90.project
```diff
@@ -27,8 +27,6 @@ packages:
          ./plugins/hls-call-hierarchy-plugin
          ./plugins/hls-alternate-number-format-plugin
 
-with-compiler: ghc-9.0.1
```
michaelpj, line 30: I'm assuming this didn't matter? We won't have another GHC in scope or anything? (I don't think we will, but just checking)

.github/workflows/build.yml
jneira: yeah, it should but we were not using caching here, not sure why. I had that as TODO in my mind :-P

cabal-ghc90.project
jneira: the line only saved you `-w ghc-9.0.1`. Now the file should work for both ghc 9.0.1 and 9.0.2 and that line would make 9.0.2 in PATH (like we do in ci) be ignored

.github/workflows/build.yml
jneira: my idea is check if caching should be effectively avoided and separate the actual action in two, extracting the cache part, in the other case

Anton-Latukha: Yes. Neither of us initially planne to refactor `build`. `build` seems like particularly rarely used & important workflow & I left it to the more experienced people. Consolidating & solidifying the work between `caching`, `test`, `bench` already was enough task & scope. We always can share the work done into `build`, but at the start that approach was ineffective (consolidating between 4 targets at once may be exponentially harder task).

.github/workflows/hackage.yml
Anton-Latukha: Just a *nota bene*: the https://github.com/haskell/haskell-language-server/pull/2567/files#r779448472 is indeed a complex reason & seems like a case for a code comment/doc "why it was needed here".

.github/workflows/hackage.yml
```diff
@@ -103,22 +104,15 @@ jobs:
...
         run: |
           # TODO: remove when not needed
           cd $(ls -d ./incoming/${{ matrix.package }}-*)
           # For brittany
-          echo " brittany:base, brittany:ghc, brittany:ghc-boot-th, butcher:base, multistate:base, data-tree-print:base," >> cabal.project
-          # For floskell and stylish-haskell
-          echo " floskell:base, floskell:ghc-prim, stylish-haskell:Cabal,stylish-haskell:ghc-lib-parser," >> cabal.project
+          echo " butcher:base, multistate:base, data-tree-print:base," >> cabal.project
+          # For stylish-haskell
+          echo " stylish-haskell:Cabal,stylish-haskell:ghc-lib-parser,stylish-haskell:aeson" >> cabal.project
```
Anton-Latukha, line 115: Seems like the same reasoning from https://github.com/haskell/haskell-language-server/pull/2567/files#r779448472 to be documented

Seems like the sum of: https://github.com/haskell/haskell-language-server/pull/2567/files#r779448472 & https://github.com/haskell/haskell-language-server/pull/2567/files#r779120031 is enough to form a rough understanding.

But also `hackage` & `build` are a "deep cut" stuff for core maintainers, but even in that case doc may be beneficial to keep in the file.

.github/workflows/test.yml
Anton-Latukha: There is nothing ugly in the ugly shell script in the `.workflow` directory.

I would post an argument that supplying a function through `env` or through `.bashrc` would be dirtier.

Also, the GHC is in a build environment, so it can be `ghci` script also.

It would be:
```shell
#!/usr/bin/env sh

BUILD_TARGET_NAME=$1

TEST_OPTIONS=$2

cabal test "$BUILD_TARGET_NAME" --test-options="$TEST_OPTIONS" || cabal test "$BUILD_TARGET_NAME" --test-options="$TEST_OPTIONS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test "$BUILD_TARGET_NAME" --test-options="$TEST_OPTIONS"
```

```shell
[pyro@rogue haskell-language-server]$ bash -x ./draft.sh hls-brittany-plugin TOPT
+ BUILD_TARGET_NAME=hls-brittany-plugin
+ TEST_OPTIONS=TOPT
+ cabal test hls-brittany-plugin --test-options=TOPT
Resolving dependencies...
```

But I think this PR is already to big. Just an idea to refactor after merge.

Anton-Latukha: Reported in https://github.com/haskell/haskell-language-server/issues/2586

.github/workflows/hackage.yml
Anton-Latukha: Reported in https://github.com/haskell/haskell-language-server/issues/2586

.github/workflows/build.yml
Anton-Latukha: Reported in https://github.com/haskell/haskell-language-server/issues/2586

Anton-Latukha:
This `+0.0.1` update is a doozy. It is already big enough, the comments we made that are not related to the agenda - gathered into #2586.

.github/workflows/caching.yml
```diff
@@ -45,7 +45,7 @@ on:
     - cron: "25 2/8 * * *"
 
 env:
-  cabalBuild: "v2-build all --keep-going"
```
Anton-Latukha, line 48: *nota bene*, commit message explains this, this indeed needed to be removed currently, to simplify things to workaround hard `cabal` tooling/package description design difficulties.

.github/workflows/caching.yml
```diff
@@ -45,7 +45,7 @@ on:
...
+  cabalBuild: "v2-build --keep-going"
 
 jobs:
```
Anton-Latukha, line 50: *nota bene* commit message explains this, this indeed needed to be removed currently, to simplify things to workaround hard `cabal` tooling/package description design difficulties.

jneira: agree, will add the commit msg in the file

jneira:
As expected the caching workflow was succesful: https://github.com/jneira/haskell-language-server/runs/4804786986?check_suite_focus=true

jneira:
@michaelpj @Anton-Latukha I ve changed the caching workflow to build bench deps adding some comments (and in other workflow filed) hope it makes thing a little bit clearer

Anton-Latukha (approved):
I was Ok with previous (pre `9.2.1` changes merge), & I am Ok with the recently mentioned `caching` changes.

jneira:
The new caching workflow has been succesful here: https://github.com/jneira/haskell-language-server/actions/runs/1704688585

.github/workflows/build.yml
```diff
@@ -20,8 +20,9 @@ jobs:
       matrix:
         ghc:
           [ "9.2.1"
+          , "9.0.2"
           , "9.0.1"
-          , "8.10.7"
+          , '8.10.7'
```
jneira, line 25: Unnecessary change
```suggestion
          , "8.10.7"
```

michaelpj (approved):

Issue #2403, opened 2021-11-25
haskell-language-server-8.10.7: No match in record selector tcdLName


peterwicksstringfield:
I believe the partial record selection here:
```haskell
            TyClD _ x ->
                let generalCompls = [mkComp id cl (Just $ ppr $ tcdLName x)
                        | id <- listify (\(_ :: Located(IdP GhcPs)) -> True) x
```
(line 449, ghcide/src/Development/IDE/Plugin/Completions/Logic.hs)

is failing when HLS tries to do name completion for a type family.

Commenting out the entire "TcClD _ x" case avoids the crash.

C.f. https://hackage.haskell.org/package/ghc-8.10.2/docs/src/GHC.Hs.Decls.html#TyClDecl
(That's 8.10.2 documentation, not sure how much it matters.)
The FamDecl and XTyClDecl data constructors don't have a field tcdLName. 
The ClassDecl constructor doesn't have that field either, but the ClassDecl case is handled separately.

C.f. a03013ce1837353a59c85074d5cc8e4b51b14601

### Your environment
```sh
haskell-language-server version: 1.5.0.0 (GHC: 8.10.7) (PATH: /home/peter/.ghcup/bin/haskell-language-server-wrapper-1.5.0) (GIT hash: 311107eabbf0537e0c192b2c377d282505b4eff1)
Tool versions found on the $PATH
cabal:		3.6.2.0
stack:		2.7.3
ghc:		8.10.7
```

(I can also replicate the crash on a fresh build of master.)

Ubuntu
emacs
minimal example script provided

### Steps to reproduce

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeFamilies #-}

module MyLib where

type family Baaaar (bool :: Bool) a where
  Baaaar _ _ = Bool

class Foo a where
  foo :: a
  foo = Baa -- cursor here, wait for completion.
```

### Expected behaviour
Should see hover text and completions and other HLS stuff.

### Actual behaviour
HLS crashes with:
haskell-language-server-8.10.7: No match in record selector tcdLName

### Include debug information
<details>
<summary>
Debug output:
</summary>

```
peter@gtower:~/Notes/2021november21_optionals$ haskell-language-server-wrapper --debug .
No 'hie.yaml' found. Try to discover the project type!
[...]

PR #2576, opened 2022-01-10
test: Add regression tests for #2403

guibou:
Continuing the work on #2569, this is regression test for type family completion.

I've implemented the tests in `ghcide` and `haskell-language-server`, not sure if that's the way or if I should drop one.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2576"></a>

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -4528,6 +4528,15 @@ localCompletionTests = [
         ,("abcdefghi", CiFunction, "abcdefghi", True, False, Nothing)
         ],
     completionTest
+        "type family"
+        ["{-# LANGUAGE DataKinds, TypeFamilies #-}"
+        ,"type family Bar a"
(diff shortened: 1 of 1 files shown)


jneira:
many thanks for the test, I think it is enough to have it in ghcide

guibou:
I amended the commit to keep only the test in ghcide

jneira:
lgtm, many thanks
just in case did you get to verify the test against a hls version without the pr fixing it?

guibou:
> lgtm, many thanks just in case did you get to verify the test against a hls version without the pr fixing it?

Yes, I tested them and they were failing with a timeout.

At least, that's what I though, but I'm trying to reproduce it right now and I do observe the failure in the log, but tests are stuck. Perhaps the timeout is too long for my patience.

jneira (approved):

PR #2202, opened 2021-09-16
WIP: GHC-9.0 support for hls-tactics-plugin

anka-213:
It compiles and most tests succeed, but some fail. In particular, the ones where it should suggest `show` fails to find that as a possible solution and fails to find evidence for `Show a`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2202"></a>

PR #2581, opened 2022-01-12
GHC-9.0 support for hls-tactics-plugin

isovector:
This is a fork of @anka-213's #2202 that makes all the tactics tests pass! :tada: 

The only issue is what to do about 9.2 support, which is broken by the compatibility layer here.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2581"></a>


isovector:
OK, I think everything is good to go --- but having some issues getting Compat.Core to properly export TcType. Would someone more familiar with this scary module mind lending a hand?

pepeiborra (approved):
Awesome, thanks @isovector

test/functional/FunctionalCodeAction.hs
```diff
@@ -353,6 +355,12 @@ unusedTermTests = testGroup "unused term code actions" [
               $ Just CodeActionQuickFix `notElem` kinds
     ]
 
+expectFailIfGhc92 :: String -> TestTree -> TestTree
+expectFailIfGhc92 reason =
```
jneira, line 359: please, use the existing `knownBrokenForGhcVersions` like
https://github.com/haskell/haskell-language-server/blob/5714207bfe44b13b02726334f9caf883249e68a9/plugins/hls-hlint-plugin/test/Main.hs#L347

jneira: from https://github.com/haskell/haskell-language-server/blob/master/hls-test-utils/src/Test/Hls/Util.hs#L32

PR #2587, opened 2022-01-14
More work around next ghc-9.2.1 support

jneira:
* Updating the issue about i 've noted:
  * hls-class-hierarchy-plugin is being tested for ghc-9.2.1 but it is not included in the executable via flags, this includes it

https://github.com/haskell/haskell-language-server/blob/8804cbdf650e85ebb44adf39848d61454c26e8f5/.github/workflows/test.yml#L201-L203

https://github.com/haskell/haskell-language-server/blob/8804cbdf650e85ebb44adf39848d61454c26e8f5/cabal-ghc921.project#L54-L69

  * hls-module-name-plugin has a test suite but it not was being tested in ci :facepalm:, this includes even for ghc-9.2.1, if it does not work, will exclude from that version

### DIFF
--- a/cabal-ghc921.project
+++ b/cabal-ghc921.project
@@ -55,7 +55,6 @@ constraints:
     +ignore-plugins-ghc-bounds
     -alternateNumberFormat
     -brittany
-    -callhierarchy
     -class
     -eval
     -haddockComments
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -210,6 +210,10 @@ jobs:
         name: Test hls-hlint-plugin test suite
         run: cabal test hls-hlint-plugin --test-options="$TEST_OPTS" || cabal test hls-hlint-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-hlint-plugin --test-options="$TEST_OPTS"
 
+      - if: matrix.test
+        name: Test hls-module-name-plugin test suite
+        run: cabal test hls-module-name-plugin --test-options="$TEST_OPTS" || cabal test hls-module-name-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-module-name-plugin --test-options="$TEST_OPTS"
+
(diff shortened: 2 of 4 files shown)


cabal-ghc921.project
```diff
@@ -55,12 +55,10 @@ constraints:
     +ignore-plugins-ghc-bounds
```
michaelpj, line 55: Shouldn't we be able to go back to just turning this flag off and letting cabal pick the ones which say they can be included based on the ghc bounds in the cabal files?

jneira: I am afraid we can't, the ghc "bounds" are not a dependency on ghc in `build-depends` but a specific cabal syntax for conditionals using the current ghc, the condition sets the flag and then comes the solver phase.
For example it is possible to have conditions on ghc version without depending at all on ghc as a library

jneira: Maybe we could replace flags with ghc bounds in `build-depends` (maybe you were thinking on that?) but not all plugins depend on ghc directly but through ghcide and that dependency would be "artificial". I would say plugins should depend less directly on ghc (via hie-compat)

michaelpj: I meant:
- We have the `stylish-haskell` flag in `haskell-language-server.cabal`
- We depend on the `stylish-haskell` plugin if the flag is on *and* the GHC version is okay, unless we set the override flag
- Here we're turning off the `stylish-haskell` flag *and* setting the override flag.
- So... what if we stop setting either of them? Then the `stylish-haskell` plugin just won't be enabled because of the GHC version guard, which is what we want, right?

jneira: hmm, i think the idea was being able to override the ghc bound withouth changing the .cabal file only the cabal.project and change only the .cabal file when the development is finished and we are about to do the hackage release

so you could override it doing a cabal install from a hackage version if you local environment makes the plugin work somewhat

but maybe it is better to update directly the .cabal file yeah as that use case is in fact very rare, will try it, thanks!

jneira:
hls-module-name does not even compile with ghc-9.2.1, sad but good to know
and hls-class-hierarchy-plugin so :tada:

cabal-ghc921.project
jneira: hmm i think the explanation is we can have plugins working in github but no in hackage (due to allow-newers and remote-source-repo in the cabal.project)
so we have to make them not buildables by default for hackage but skip it in the cabal.project
does it make sense?

michaelpj (approved):

PR #2591, opened 2022-01-15
Turn HLS-wrapper into an LSP Server

### LINKED ISSUE #2589: Turn HLS-Wrapper into an LSP Server
**Is your feature request related to a problem? Please describe.**

When hls-wrapper realises it is going to fail for one of the following reasons:

* It can't figure out which GHC Version it needs
* It can not find a fitting HLS version for the project's GHC version

Then hls-wrapper writes something to stderr and crashes. This is bad UX as users are often not aware about the stderr output and if they are, have a hard time to read the output.

hls-wrapper should realise it is going to fail and become the LSP - Server to report to the client that we failed to launch. It can present the user a reasonable error message in combination with help instructions.

**Describe alternatives you've considered**

I was unable to come up with a better idea.

Maybe it is enough to send some json messages directly to stdout instead of a full-fledged `lsp` dependency.

fendor:
Introduces better error handling to HLS-wrapper to show LSP clients
dedicated error messages.
This should help users understand why their Language Server isn't
starting.

Closes #2589 

image

* [x] Sends Request to LSP client on failure
* [x] Can be restarted
* [ ] Documentation for Wrapper LSP

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2591"></a>

### DIFF
--- a/exe/Wrapper.hs
+++ b/exe/Wrapper.hs
@@ -1,15 +1,22 @@
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
 -- | This module is based on the hie-wrapper.sh script in
 -- https://github.com/alanz/vscode-hie-server
 module Main where
 
 import           Control.Monad.Extra
+import           Control.Monad.IO.Class
+import           Control.Monad.Trans.Except
 import           Data.Default
 import           Data.Foldable
 import           Data.List
+import           Data.Maybe
+import qualified Data.Text                  as T
+import qualified Data.Text.IO               as T
 import           Data.Void
-import qualified Development.IDE.Session as Session
-import qualified HIE.Bios.Environment    as HieBios
+import qualified Development.IDE.Session    as Session
+import qualified HIE.Bios.Environment       as HieBios
 import           HIE.Bios.Types
 import           Ide.Arguments
 import           Ide.Version
@@ -20,6 +27,7 @@ import           System.FilePath
 import           System.IO
 import           System.Info
 import           System.Process
+import           WrapperLspMain
 
 -- ---------------------------------------------------------------------
 
@@ -46,9 +54,17 @@ main = do
       BiosMode PrintCradleType ->
           print =<< findProjectCradle
 
-      _ -> launchHaskellLanguageServer args
+      _ -> launchHaskellLanguageServer args >>= \case
+        Right () -> pure ()
+        Left err -> do
+          T.hPutStrLn stderr "*** Startup ERROR"
+          T.hPutStrLn stderr (prettyError err NoShorten)
+          case args of
+            Ghcide ghcideArguments -> lspMain ghcideArguments (prettyError err Shorten)
+            _ -> pure ()
 
-launchHaskellLanguageServer :: Arguments -> IO ()
+
+launchHaskellLanguageServer :: Arguments -> IO (Either WrapperSetupError ())
 launchHaskellLanguageServer parsedArgs = do
   case parsedArgs of
     Ghcide GhcideArguments{..} -> whenJust argsCwd setCurrentDirectory
@@ -64,7 +80,10 @@ launchHaskellLanguageServer parsedArgs = do
 
   case parsedArgs of
     Ghcide GhcideArguments{..} ->
-      when argsProjectGhcVersion $ getRuntimeGhcVersion' cradle >>= putStrLn >> exitSuccess
+      when argsProjectGhcVersion $ do
+        runExceptT (getRuntimeGhcVersion' cradle) >>= \case
+          Right ghcVersion -> putStrLn ghcVersion >> exitSuccess
+          Left err -> T.putStrLn (prettyError err NoShorten) >> exitFailure
     _ -> pure ()
 
   progName <- getProgName
@@ -83,51 +102,53 @@ launchHaskellLanguageServer parsedArgs = do
   hPutStrLn stderr ""
   -- Get the ghc version -- this might fail!
   hPutStrLn stderr "Consulting the cradle to get project GHC version..."
-  ghcVersion <- getRuntimeGhcVersion' cradle
-  hPutStrLn stderr $ "Project GHC version: " ++ ghcVersion
+  runExceptT (getRuntimeGhcVersion' cradle) >>= \case
+    Left err -> pure $ Left err
+    Right ghcVersion -> do
+      hPutStrLn stderr $ "Project GHC version: " ++ ghcVersion
 
-  let
-    hlsBin = "haskell-language-server-" ++ ghcVersion
-    candidates' = [hlsBin, "haskell-language-server"]
-    candidates = map (++ exeExtension) candidates'
+      let
+        hlsBin = "haskell-language-server-" ++ ghcVersion
+        candidates' = [hlsBin, "haskell-language-server"]
+        candidates = map (++ exeExtension) candidates'
 
-  hPutStrLn stderr $ "haskell-language-server exe candidates: " ++ show candidates
+      hPutStrLn stderr $ "haskell-language-server exe candidates: " ++ show candidates
 
-  mexes <- traverse findExecutable candidates
+      mexes <- traverse findExecutable candidates
 
-  case asum mexes of
-    Nothing -> hPutStrLn stderr $ "Cannot find any haskell-language-server exe, looked for: " ++ intercalate ", " candidates
-    Just e -> do
-      hPutStrLn stderr $ "Launching haskell-language-server exe at:" ++ e
-      callProcess e args
+      case asum mexes of
+        Nothing -> pure $ Left $ NoLanguageServer ghcVersion candidates
+        Just e -> do
+          hPutStrLn stderr $ "Launching haskell-language-server exe at:" ++ e
+          callProcess e args
+          pure $ Right ()
 
--- | Version of 'getRuntimeGhcVersion' that dies if we can't get it, and also
--- checks to see if the tool is missing if it is one of
-getRuntimeGhcVersion' :: Show a => Cradle a -> IO String
+-- | Version of 'getRuntimeGhcVersion' that throws a 'WrapperSetupError' if we
+-- can't get it, and also checks if run-time tool dependencies are missing.
+getRuntimeGhcVersion' :: Cradle Void -> ExceptT WrapperSetupError IO String
 getRuntimeGhcVersion' cradle = do
 
+  let cradleName = actionName (cradleOptsProg cradle)
   -- See if the tool is installed
-  case actionName (cradleOptsProg cradle) of
+  case cradleName of
     Stack   -> checkToolExists "stack"
     Cabal   -> checkToolExists "cabal"
     Default -> checkToolExists "ghc"
     Direct  -> checkToolExists "ghc"
     _       -> pure ()
 
-  ghcVersionRes <- HieBios.getRuntimeGhcVersion cradle
+  ghcVersionRes <- liftIO $ HieBios.getRuntimeGhcVersion cradle
   case ghcVersionRes of
     CradleSuccess ver -> do
       return ver
-    CradleFail error -> die $ "Failed to get project GHC version:" ++ show error
-    CradleNone -> die "Failed get project GHC version, since we have a none cradle"
+    CradleFail error -> throwE $ FailedToObtainGhcVersion cradleName error
+    CradleNone -> throwE $ NoneCradleGhcVersion cradleName
   where
     checkToolExists exe = do
-      exists <- findExecutable exe
+      exists <- liftIO $ findExecutable exe
       case exists of
         Just _ -> pure ()
-        Nothing ->
-          die $ "Cradle requires " ++ exe ++ " but couldn't find it" ++ "\n"
-           ++ show cradle
+        Nothing -> throwE $ ToolRequirementMissing exe (actionName (cradleOptsProg cradle))
 
 findProjectCradle :: IO (Cradle Void)
 findProjectCradle = do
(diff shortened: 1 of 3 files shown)


fendor:
I was unable to write tests for this... so only a POC image to show I tested it at least locally.

fendor:
Another noteable improvement: the logs are better structured now:
```
Consulting the cradle to get project GHC version...
*** Startup ERROR
Failed to find the GHC version of this Cabal project.
Error when calling cabal --builddir=/home/hugin/.cache/hie-bios/dist-hls-036ebb56a4cc135651b7510a90f3136d v2-exec --with-compiler /home/hugin/.cache/hie-bios/wrapper-13a09b18ea883dd377d59db5e821a86b ghc -v0 -- --numeric-version

[__0] trying: ghcide-1.5.0.1 (user goal)
[__1] rejecting: ghcide:!test (constraint from config file, command line flag,
or user target requires opposite flag selection)
[__1] trying: ghcide:*test
[__2] trying: tasty-rerun-1.1.18 (dependency of ghcide *test)
[__3] trying: tagged-0.8.6.1 (dependency of tasty-rerun)
[__4] trying: template-haskell-2.18.0.0/installed-2.18.0.0 (dependency of
tagged)
[__5] trying: ghc-typelits-knownnat-0.7.6 (dependency of ghcide *test)
[__6] trying: ghc-typelits-natnormalise-0.7.6 (dependency of
ghc-typelits-knownnat)
[__7] next goal: base (dependency of ghcide)
[__7] rejecting: base-4.16.0.0/installed-4.16.0.0 (conflict:
ghc-typelits-natnormalise => ghc-bignum>=1.0 && <1.1, base =>
ghc-bignum==1.2/installed-1.2)
[__7] rejecting: base-4.16.0.0, base-4.15.1.0, base-4.15.0.0, base-4.14.3.0,
base-4.14.2.0, base-4.14.1.0, base-4.14.0.0, base-4.13.0.0, base-4.12.0.0,
base-4.11.1.0, base-4.11.0.0, base-4.10.1.0, base-4.10.0.0, base-4.9.1.0,
base-4.9.0.0, base-4.8.2.0, base-4.8.1.0, base-4.8.0.0, base-4.7.0.2,
base-4.7.0.1, base-4.7.0.0, base-4.6.0.1, base-4.6.0.0, base-4.5.1.0,
base-4.5.0.0, base-4.4.1.0, base-4.4.0.0, base-4.3.1.0, base-4.3.0.0,
base-4.2.0.2, base-4.2.0.1, base-4.2.0.0, base-4.1.0.0, base-4.0.0.0,
base-3.0.3.2, base-3.0.3.1 (constraint from non-upgradeable package requires
installed instance)
[__7] fail (backjumping, conflict set: base, ghc-typelits-natnormalise,
ghcide)
goals I've had most trouble fulfilling: base, ghc-typelits-natnormalise,
ghcide, ghc-typelits-knownnat, template-haskell, integer-gmp, tagged,
ghcide:test, tasty-rerun
Try running with --minimize-conflict-set to improve the error message.
```

exe/Wrapper.hs
```diff
@@ -46,9 +61,17 @@ main = do
...
-launchHaskellLanguageServer :: Arguments -> IO ()
+      _ -> launchHaskellLanguageServer args >>= \case
+        Right () -> pure ()
+        Left err -> do
+          T.hPutStrLn stderr "*** Startup ERROR"
+          T.hPutStrLn stderr (prettyError err NoShorten)
+          putWindowMessage J.MtError (prettyError err Shorten)
+          -- Wait for 45 seconds before shutdown, so we don't spam the same message
+          -- since some LSP clients attempt to re-launch the server.
+          threadDelay (45 * 1_000_000)
```
fendor, line 72: I think this kills VSCode "Restart Haskell Language Server" feature.

fendor: So, can we improve that? Should we maybe wait for a message from stdin or something like that?

pepeiborra: Instead of a `WindowShowMessage`, send a `SWindowShowMessageRequest` with a "Restart" `MessageActionItem`

fendor: That is a great idea, thanks!

fendor: Unfortunately, not sure if we can do that by easily by just parsing json messages... at some point it makes more sense in my opinion to use `lsp` and install a handler.

pepeiborra: Why not do that?

fendor: no reason other than thinking it *might* be overkill. I am going to use `lsp` now.

jneira:
Converted to draft until the lsp change

exe/Wrapper.hs
fendor: Fixed

fendor:
@jneira Can you please remove the Draft status? (I have no idea how to do that myself)

exe/Wrapper.hs
```diff
@@ -83,51 +102,53 @@ launchHaskellLanguageServer parsedArgs = do
...
+      case asum mexes of
+        Nothing -> pure $ Left $ NoLanguageServer ghcVersion candidates
+        Just e -> do
+          hPutStrLn stderr $ "Launching haskell-language-server exe at:" ++ e
+          callProcess e args
+          pure $ Right ()
+
+-- | Version of 'getRuntimeGhcVersion' that throws a 'WrapperSetupError' if we
+-- can't get it, and also checks if run-time tool dependencies are missing.
+getRuntimeGhcVersion' :: Cradle Void -> ExceptT WrapperSetupError IO String
```
fendor, line 128:
```suggestion
getRuntimeGhcVersion' :: MonadIO m => Cradle Void -> ExceptT WrapperSetupError m String
```

Maybe? Probably not worth it, though

exe/Wrapper.hs
```diff
@@ -142,3 +163,35 @@ findProjectCradle = do
...
+      "\n" <> T.pack (intercalate "\n" (cradleErrorStderr crdlError))
+prettyError (NoneCradleGhcVersion name) _ =
+  "Failed to get the GHC version of the " <> T.pack (show name) <>
+  " project, since we have a none cradle"
+prettyError (NoLanguageServer ghcVersion candidates) _ =
+  "Failed to find a HLS version for GHC " <> T.pack ghcVersion <>
+  "\nExecutable names we failed to find: " <> T.pack (intercalate "," candidates)
+prettyError (ToolRequirementMissing toolExe name) _ =
+  "This is a " <> T.pack (show name) <> " Project, but we failed to find \"" <>
+  T.pack toolExe <> "\" on the $PATH"
```
fendor, line 197: Maybe this can be a shortened version, and we can dump the Search Path to stderr, too?

exe/Wrapper.hs
```diff
@@ -142,3 +163,35 @@ findProjectCradle = do
...
+    Shorten ->
+      "\n" <> T.pack (fromMaybe "" . listToMaybe $ cradleErrorStderr crdlError)
+    NoShorten ->
+      "\n" <> T.pack (intercalate "\n" (cradleErrorStderr crdlError))
+prettyError (NoneCradleGhcVersion name) _ =
+  "Failed to get the GHC version of the " <> T.pack (show name) <>
+  " project, since we have a none cradle"
+prettyError (NoLanguageServer ghcVersion candidates) _ =
+  "Failed to find a HLS version for GHC " <> T.pack ghcVersion <>
+  "\nExecutable names we failed to find: " <> T.pack (intercalate "," candidates)
```
fendor, line 194: This wording feels bad

exe/WrapperLspMain.hs
```diff
@@ -0,0 +1,246 @@
...
+                , Ghcide.optTesting = Ghcide.IdeTesting argsTesting
+                }
+          }
+
+
+    inH <- Main.argsHandleIn realArguments
+    outH <- Main.argsHandleOut realArguments
+
+    runLanguageServer (Main.argsLspOptions realArguments) inH outH (Main.argsDefaultHlsConfig realArguments) getConfigFromNotification $ \env exitFun -> do
+        -- Send a message to the client to tell about our problems
```
fendor, line 69: WIP comment, will be changed

fendor (commented):
Most stuff is basically copy-pasted from ghcide's `runLangaugeServer` with all shake stuff stripped out.

exe/Wrapper.hs
drsooch: `FailedToObtainGhcVersion` and `NoneCradleGhcVersion` look they could be identical? i.e. something like `Failed to find GHC version for project: {name}` (unless name implies something else?) and then for `NoneCradle` keep the same but with `(NoneCradle)` or some other marker attached?

exe/WrapperLspMain.hs
```diff
@@ -0,0 +1,246 @@
...
+import           UnliftIO
+import           UnliftIO.Concurrent
+
+type ServerM c = ReaderT ReactorChan (LspM c)
+
+lspMain :: IdeArgs.GhcideArguments -> T.Text -> IO ()
+lspMain ghcideArgs@IdeArgs.GhcideArguments{..} msg = do
+    whenJust argsCwd IO.setCurrentDirectory
+    dir <- IO.getCurrentDirectory
+    LSP.setupLogger argsLogFile ["hls"]
```
drsooch, line 44: There were some earlier inconsistencies with pathing when `--cwd` was supplied. https://github.com/haskell/haskell-language-server/blob/47cb213ac70dabb567461ac93af6933f79c39f4f/ghcide/exe/Main.hs#L53-L56

exe/WrapperLspMain.hs
```diff
@@ -0,0 +1,246 @@
...
+                { Ghcide.optShakeProfiling = argsShakeProfiling
+                , Ghcide.optTesting = Ghcide.IdeTesting argsTesting
+                }
+          }
+
+
+    inH <- Main.argsHandleIn realArguments
+    outH <- Main.argsHandleOut realArguments
+
+    runLanguageServer (Main.argsLspOptions realArguments) inH outH (Main.argsDefaultHlsConfig realArguments) getConfigFromNotification $ \env exitFun -> do
```
pepeiborra, line 68: Duplicating `runLanguageServer` here is not great. This code should either:
1.  call the real `runLanguageServer` (you can always parameterise it even more) 
2. fork it (as done here) and clean up. 

In other words, I'd prefer if you could clean this up or just call the original

exe/WrapperLspMain.hs
```diff
@@ -0,0 +1,246 @@
...
+-- LSP Helper functions
+runLanguageServer
+    :: LSP.Options
+    -> Handle -- input
+    -> Handle -- output
+    -> Config
+    -> (Config -> Value -> Either T.Text Config)
+    -> (LSP.LanguageContextEnv Config -> IO () -> IO ())
+    -> IO ()
+runLanguageServer options inH outH defaultConfig onConfigurationChange onRun = do
```
pepeiborra, line 87: Why all these arguments?

```suggestion
runLanguageServer onRun = do
```

exe/WrapperLspMain.hs
```diff
@@ -0,0 +1,246 @@
...
+    let waitForCancel reqId = atomically $ do
+            cancelled <- readTVar cancelledRequests
+            unless (reqId `Set.member` cancelled) retry
+
+    -- Send everything over a channel, since you need to wait until after initialise before
+    -- LspFuncs is available
+    clientMsgChan :: Chan ReactorMessage <- newChan
+
+    let asyncHandlers = mconcat
+          [ cancelHandler cancelRequest
```
pepeiborra, line 126: Why handle cancellation?

exe/Wrapper.hs
fendor: They are slightly different though, we should improve the error message!
* In the case of a `none` cradle, we can not figure out the GHC version to use, thus can't decide which HLS version we want to launch. The outside world specifically instructed us not to load this project, probably via a `hie.yaml`
* FailedToObtainGhcVersion entails that some command has failed invocation, e.g. `cabal v2-exec -- ghc -v0 --numeric-version`, e.g. we attempted to load the project, but we couldn't.

drsooch: Ah I was paying close enough attention I didn't see the extra details added to

exe/WrapperLspMain.hs
fendor: calling the original is a bit tricky, as the original `runLanguageServer` handles the initialisation request and sends a response which we probably don't want to... If we don't care about that (when I think about it, we probably don't have to), we can re-use `runLanguageServer` with basically a startup continuation. 

I think we can re-use the code, we would open a shake database, etc... but that doesn't matter too much, right?

michaelpj:
I think we have to think up some way to do this without the massive duplication. I don't think we can justify all that just for some better reporting at startup :/

Another thing: how will running a whole LSP server here look to the client? Will we e.g. respond to their `initialize` request? But then what happens when the real HLS starts up, it will expect to get sent `initialize` also.

exe/WrapperLspMain.hs
pepeiborra: The shake database thing is just a legacy concept - we are not using Shake anymore so all it is now is a bunch of concurrent mutable hash maps.

pepeiborra:
> I just realise, we can't rely on that every lsp-client has the same behaviour as VSCode (e.g. try to restart the language server if it exits), which makes it a bit trickier... Afaict, the LSP server always parses the initialisation request (unless we can introduce some way to avoid that).

I wouldn't worry too much about this. If the LSP client doesn't try to restart, we have already notified the user that something was wrong and given them a strong hint about what to do

michaelpj:
I had a possibly insane idea. What if:
- We taught *all* the `haskell-language-server` binaries a `wrapper` subcommand that causes them to act as a wrapper.
- We build the "wrapper" by just building the normal executable, but statically and with all the plugin flags turned off to keep its footprint small.

Then we can hopefully avoid the duplication: because we'll reuse the LSP setup that we normally do. And we can probably inherit e.g. arguments for log files etc.

hasufell:
> I had a possibly insane idea. What if:
> 
> * We taught _all_ the `haskell-language-server` binaries a `wrapper` subcommand that causes them to act as a wrapper.
> * We build the "wrapper" by just building the normal executable, but statically and with all the plugin flags turned off to keep its footprint small.
> 
> Then we can hopefully avoid the duplication: because we'll reuse the LSP setup that we normally do. And we can probably inherit e.g. arguments for log files etc.

That requires either:

1. users to change `haskell-language-server-wrapper` to `haskell-language-server-wrapper --wrapper` in their LSP client config, or
2. a wrapper shell script that invokes it with `--wrapper`

michaelpj:
>That requires either...

True. But having a wrapper script that just calls `minimal-haskell-language-server --wrapper` doesn't seem so bad.

pepeiborra:
> I had a possibly insane idea. What if:
> 
> * We taught _all_ the `haskell-language-server` binaries a `wrapper` subcommand that causes them to act as a wrapper.
> * We build the "wrapper" by just building the normal executable, but statically and with all the plugin flags turned off to keep its footprint small.
> 
> Then we can hopefully avoid the duplication:

Initially I really liked this idea, but then I realised that the main HLS binary is just calling `Development.Ide.Main.defaultMain` with a list of plugins and a list of custom commands. 

So most of the duplication is already abstracted, and what's left is due to the need to inline `runLanguageServer` in order to prevent the wrapper from handling the `Initialize` request. This problem doesn't go away, the duplication would just move to the custom `wrapper` subcommand right? 

@fendor did you consider my suggestion in https://github.com/haskell/haskell-language-server/pull/2591#discussion_r797312138 yet?

michaelpj:
> So most of the duplication is already abstracted, and what's left is due to the need to inline runLanguageServer in order to prevent the wrapper from handling the Initialize request.

Hmm, good point. So maybe we just need to paramterize that by something to say "don't really start up" :thinking:

michaelpj:
... which is what you suggested already.

fendor:
> did you consider my suggestion in https://github.com/haskell/haskell-language-server/pull/2591#discussion_r797312138 yet?

Yes, I think it is a good idea to parameterise `runLanguageServer` more, but I didn't have time to implement that change.

hasufell:
> I had a possibly insane idea. What if:
> 
> * We taught _all_ the `haskell-language-server` binaries a `wrapper` subcommand that causes them to act as a wrapper.
> * We build the "wrapper" by just building the normal executable, but statically and with all the plugin flags turned off to keep its footprint small.
> 
> Then we can hopefully avoid the duplication: because we'll reuse the LSP setup that we normally do. And we can probably inherit e.g. arguments for log files etc.

Actually, instead of adding a subcommand, we can simply re-use the strategy many unix cli programs use: rely on `argv[0]` like busybox, git etc. In haskell, that's [`getProgName`](https://hackage.haskell.org/package/base-4.16.0.0/docs/System-Environment.html#v:getProgName) (with same caveats on windows).

fendor:
Closed in favour of #2960

PR #2593, opened 2022-01-15
[ghc-9.2] Fix rename plugin

pepeiborra:
The rename plugin relies on SYB to find located AST trees that need renaming. This breaks in GHC 9.2 because the representation of located AST trees has changed: it now includes exactprint annotations.

To make the rename plugin work in 9.2, we need to adjust the SYB queries to look for the new located AST types.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2593"></a>


jneira (approved):
lgtm, one more plugin for ghc-9.2.1, yay!

PR #2594, opened 2022-01-16
Fix module-name plugin on ghc-9.2.1

mrgutkun:
Actual fix mimics changes from #2503 ; config changes indiscriminately turn building with module-name on ghc-9.2.1 on.
I assume that bumping minor version of the plugin is useful for constraint configuration.
(AFAIU building 9.2.1 with stack is not ready yet, but why not turn module-name on in there too, in preparation.)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2594"></a>

### DIFF
--- a/cabal-ghc921.project
+++ b/cabal-ghc921.project
@@ -59,7 +59,7 @@ constraints:
     -eval
     -haddockComments
     -hlint
-    -moduleName
+    +moduleName
     -qualifyImportedNames
     -refineImports
     -retrie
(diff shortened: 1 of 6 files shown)


jneira (approved):
many thanks for including the plugin for 9.2.1, let's see if CI tests are happy and merge

cabal-ghc921.project
```diff
@@ -59,7 +59,7 @@ constraints:
     -eval
     -haddockComments
     -hlint
-    -moduleName
+    +moduleName
```
jneira, line 62: removing it would work top
```suggestion
```

jneira:
Tests are ok and changes looks good, mant thanks!

Issue #2193, opened 2021-09-14
Enable rename plugin


jneira:
* Currently it is disabled by default in the hls .cabal file:
https://github.com/haskell/haskell-language-server/blob/e1075e88f15e2c93e8968f69bacc33356d9db255/haskell-language-server.cabal#L134-L137
* Would be good to know what steps are needed to being able to enable it

OliverMadine:
It is just disabled due to the limited multi-cradle support.

I've had a temporary solution working for a while now (just loading all components listed in the `hie.yaml`) but we need [#304](https://github.com/haskell/hie-bios/pull/304) and #2009 to be merged first. Both of these PRs seems to be working and ready to merge to me.

michaelpj:
I think if we enabled it with the caveat that it will only work reliably across one component that would be fine. Perhaps slap "(experimental)" after all the code action titles :)

OliverMadine:
> I think if we enabled it with the caveat that it will only work reliably across one component that would be fine. Perhaps slap "(experimental)" after all the code action titles :)

There is no code action since we get the new name from the client. So I don't think we really have a great way of announcing this caveat to the user without them having to go through HLS logs etc.

It would be nice to have a vscode option so that we could enable the plugin for single-component use more easily, but I'm not really sure how to do this (especially without causing the plugin tests to fail). 
Also, hopefully we are close to some form of multi-component support so I'd rather just wait for that.

jneira:
i am afraid #2009 needs https://github.com/haskell/hie-bios/pull/301 which in turn needs the `show-build-info` feature which will be avilable in Cabal-3.8 afaik

OliverMadine:
> i am afraid #2009 needs [haskell/hie-bios#301](https://github.com/haskell/hie-bios/pull/301) which in turn needs the `show-build-info` feature which will be avilable in Cabal-3.8 afaik

It was my understanding that this PR is just concerned with 'loading multiple components' and we were going to use `show-build-info` to 'get the components to load' later.

fendor:
Sorry, that PR is indeed more a POC for what is possible with show-build-info, not of the general multiple components loading mechanism.

However, it is true that the general issue of loading multiple components can be done independently of show-build-info, just the semantics of the proposed NonEmpty needs to be discussed.

jneira:
[...]

Issue #304, opened 2020-08-09
cabal cannot find Main


HE7086:
I am getting this error on Main.hs, while other files worked fine  
cabal can build the executable and it runs fine. 
this only occurs with hls but not hie (previously, like several months ago)

![Screenshot_20200810_003143](https://user-images.githubusercontent.com/26016419/89743226-61958480-daa1-11ea-882c-94f4ae82a736.png)

a toy project can be found [here](https://github.com/HE7086/Fourier)
(note that -dynamic flag maybe not required on other distributions)

env: 
arch linux
neovim with coc
ghc 8.10.1, cabal 3.2.0.0, all dynamically linked
hls 0.2.2 from AUR

:CocInfo output:
```
## versions

vim version: NVIM v0.4.4
node version: v14.7.0
coc.nvim version: 0.0.78-bdd9a9e140
coc.nvim directory: /home/he/.config/nvim/plugged/coc.nvim
term: xterm-256color
platform: linux

## Output channel: languageserver.haskell

Module "/home/he/code/Fourier/src/a" is loaded by Cradle: 
Cradle {cradleRootDir = "/home/he/code/Fourier", cradleOptsProg = CradleAction: Cabal}
Run entered for haskell-language-server-wrapper
(haskell-language-server-wrapper) Version 0.2.2.0, 
Git revision e44f618c11b7978264a94beeee00d1f014867f6d (dirty) x86_64 ghc-8.10.1
Current directory: /home/he/code/Fourier/src
Operating system: linux
Arguments: ["--lsp"]
Cradle directory: /home/he/code/Fourier
Cradle type: Cabal
Consulting the cradle to get project GHC version...
Project GHC version: 8.10.1
haskell-language-server exe candidates: 
["haskell-language-server-8.10.1","haskell-language-server-8.10","haskell-language-server"]
Launching haskell-language-server exe at:/usr/bin/haskell-language-server-8.10.1
haskell-language-server version: 0.2.2.0 (GHC: 8.10.1) 
(PATH: /usr/bin/haskell-language-server-8.10.1) 
(GIT hash: e44f618c11b7978264a94beeee00d1f014867f6d)
Starting (haskell-language-server)LSP server... 
with arguments: LspArguments 
{
    argLSP = True, 
    argsCwd = Nothing, 
    argFiles = [], 
    argsShakeProfiling = Nothing, 
    argsTesting = False, 
    argsExamplePlugin = False, 
    argsDebugOn = False, 
[...]

PR #2598, opened 2022-01-17
Enable rename plugin

jneira:
* Make progress on #2193
* Only noting in docs it works inside component boundaries (it also will be noted in the release itself)
  * So it will *not* be disabled by default in config as suggested here: https://github.com/haskell/haskell-language-server/issues/2193#issuecomment-945697950
* Not sure if a warning in logs would be needed
//cc @OliverMadine

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -18,6 +18,8 @@ You can watch demos for some of these features [below](#demos).
 - [Call hierarchy support](#call-hierarchy)
 - [Qualify names from an import declaration](#qualify-imported-names) in your code
 - [Suggest alternate numeric formats](#alternate-number-formatting)
+- [Renaming](#renaming) definitions automatically across modules of *the same component* (library, executable, test suite, benchmark).
+  - Full renaming support for the entire project is being tracked in [this issue](https://github.com/haskell/haskell-language-server/issues/2193)
 
 ## Demos
 
@@ -56,3 +58,7 @@ You can watch demos for some of these features [below](#demos).
 ### Alternate Number Formatting
 
 ![Alternate Number Format Demo](../plugins/hls-alternate-number-format-plugin/HLSAll.gif)
+
+### Renaming
+
+![Rename Demo](https://user-images.githubusercontent.com/30090176/133072143-d7d03ec7-3db1-474e-ad5e-6f40d75ff7ab.gif)
--- a/.github/workflows/hackage.yml
+++ b/.github/workflows/hackage.yml
@@ -32,7 +32,7 @@ jobs:
                   "hls-refine-imports-plugin", "hls-retrie-plugin",
                   "hls-splice-plugin", "hls-tactics-plugin",
                   "hls-call-hierarchy-plugin", "hls-alternate-number-format-plugin",
-                  "hls-qualify-imported-names-plugin",
+                  "hls-qualify-imported-names-plugin", "hls-rename-plugin",
                   "haskell-language-server"]
         ghc: [ "9.0.2"
              , "8.10.7"
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -133,7 +133,7 @@ flag refineImports
 
 flag rename
   description: Enable rename plugin
-  default:     False
+  default:     True
   manual:      True
 
 flag retrie


docs/features.md
```diff
@@ -18,6 +18,8 @@ You can watch demos for some of these features [below](#demos).
 - [Call hierarchy support](#call-hierarchy)
 - [Qualify names from an import declaration](#qualify-imported-names) in your code
 - [Suggest alternate numeric formats](#alternate-number-formatting)
+- [Renaming](#renaming) definitions automatically across modules of *the same component* (library, executable, test suite, benchmark).
```
pepeiborra, line 21: This is not correct. Renaming works across all modules that have been indexed (in this or previous sessions).

jneira: Aha, thanks, please, could you note in which cases the lack of multicomponent support could affect the plugin behaviour? does it work for the components currently loaded? so if you open files from components forcing their loading, the renaming would work across those loaded components (and no others)?

pepeiborra: > could you note in which cases the lack of multicomponent support could affect the plugin behaviour? 

I already said: the plugin will rename all occurrences that have been indexed in this or previous IDE sessions.

One example where the plugin would fail to find some occurrences is the first time a project has been loaded in the IDE *ever*, and only a subset of the components are loaded.

> does it work for the components currently loaded? 

It works for all the components indexed, whether they are loaded or not.

> so if you open files from components forcing their loading, the renaming would work across those loaded components (and no others)?

If you open a file from a component, that will cause the component to be fully indexed. So assuming that you wait until indexing is done, the plugin will work for that component.

jneira: Ok so indexed as indexed in hiedb, i guess, thanks for the clarification
Afaiu, as we have the option `haskell.checkProject`, hls by default would index all the project opening any file from it, right?
If it is set to false the best way to ensure the last version of files are indexes would open all files where the symbol to berenamed i present.

michaelpj (approved):
approve modulo Pepe being happy about us getting the caveats right.

docs/features.md
OliverMadine: I think that the caveat should be that it only works for single component projects. It does not work for a single component within a multi-component project.

Indexing all components of files where the symbol is referenced before renaming should work (I did this manually when recording the [example gif](https://github.com/haskell/haskell-language-server/pull/2108#issuecomment-918075894)).
We could grep the workspace for the symbol to do this (since reference finding also won't work yet), however, we would need to be careful not to index files that don't have a proper cradle.

OliverMadine:
I think that the caveat should be that it only works for single component projects. It does not work for a single component within a multi-component project.

Indexing all components of files where the symbol is referenced before renaming should work (I did this manually when recording the [example gif](https://github.com/haskell/haskell-language-server/pull/2108#issuecomment-918075894)).
We could grep the workspace for the symbol to do this (since reference finding also won't work yet), however, we would need to be careful not to index files that don't have a proper cradle.

jneira:
Hmm, i would prefer to have a partial but consistent behaviour. If it depends on what files has been opened by the user, and the renaming is partial and potentially different between sessions the ux could be very bad.

@OliverMadine so the fact hls checks thenetire projectwould not make renaming work without opening all the files with symbol occurences?

> Afaiu, as we have the option haskell.checkProject, hls by default would index the entire the project opening any file from it, right?

docs/features.md
pepeiborra: > Afaiu, as we have the option haskell.checkProject, hls by default would index the entire the project opening any file from it, right?

If by project you mean component, then yes.

> If it is set to false the best way to ensure the last version of files are indexes would open all files where the symbol to berenamed i present.

If it is set to false, components won't be indexed in full on open. So you could just open all the files in the workspace.

OliverMadine:
> @OliverMadine so the fact hls checks thenetire projectwould not make renaming work without opening all the files with symbol occurences?

I've just tested by deleting my cache and trying to rename. It does not seem to index the entire project, but only the components for open files.

jneira:
Thanks, so i am not sure anymore if we can enable it at compile time without disabling by default at runtime. It would be acceptable ask for open all files with the symbol to make it work? i am sure there will be uses surprised to see the renaming works some times and other only partially.

@michaelpj thoughts?

michaelpj:
Okay, I give in, that does seem bad enough that we shouldn't do it. Do we have any path forwards for fixing the underlying problem?

michaelpj:
Can we write a rule that indexes everything in the project and then have rename depend on that? It'll be expensive but at least it will work.

jneira:
It sees hls does not know about all project components beforehand. The components are loaded via cradles on demand, while files are being opened. 
I guess @OliverMadine plan is do just that but hls needs #2009 to load all components in the startup, as noted in the issue.

OliverMadine:
> It seems hls does not know about all project components beforehand. The components are loaded via cradles on demand, while files are being opened.
> I guess @OliverMadine plan is do just that but hls needs #2009 to load all components in the startup, as noted in the issue.

I think that if we assume that the `hie.yaml` lists all the components, then we can extract them from there. We just need to be able to load multiple components at once (which I believe was the purpose of [#304](https://github.com/haskell/hie-bios/pull/304)).

jneira:
dont have time to continue working on this, feel free to reuse it when enabling definitle the plugin

PR #2600, opened 2022-01-17
[ghc-9.2] Fix qualify-imported-names plugin

mrgutkun:
ghc-9.2.1 assures me that `| otherwise` in `QualifyImportedNames.hs @ 100` is redundant; hope other ghcs agree.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2600"></a>

### DIFF
--- a/stack-9.2.1.yaml
+++ b/stack-9.2.1.yaml
@@ -100,9 +100,7 @@ flags:
     haddockComments: false
     hlint: false
     importLens: false
-    moduleName: true
     ormolu: false
-    qualifyImportedNames: true
     refineImports: false
     retrie: false
     splice: false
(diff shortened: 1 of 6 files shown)


stack-9.2.1.yaml
```diff
@@ -102,7 +102,7 @@ flags:
     importLens: false
     moduleName: true
     ormolu: false
-    qualifyImportedNames: false
+    qualifyImportedNames: true
```
jneira, line 105: we can remove the flag also for stack, as the default value is true

jneira (approved):
many thanks again, great work

stack-9.2.1.yaml
mrgutkun: Oh, I see. Will do.

eddiemundo:
Ah I didn't even realize the plugin didn't compile with 9.2. Thanks for that.

PR #2601, opened 2022-01-17
[ghc-9.2] Fix refine-imports plugin

mrgutkun:
`newRdrName` in Rename.hs seems to be a debugging leftover.


jneira (approved):
Fantastic, ghc-9.2.1 will be pretty complete thanks to your work

mrgutkun:
I.....want to be clear that I'm doing plugins with comparatively trivial fixes first, and expect to run out of these soon :)

jneira:
> I.....want to be clear that I'm doing plugins with comparatively trivial fixes first, and expect to run out of these soon :)

Sure, it is relatively easy but it is required work nevertheless. hlint, brittany and maybe others needs changes in their underlying main library so i would skip them for now

Issue #2327, opened 2021-11-03
Improve features list on website


GuillaumeDesforges:
The current list of features on the website https://haskell-language-server.readthedocs.io/en/latest/features.html is not very welcoming for beginners, who find themselves digging even the most basic information.

IMO we could get some inspiration from https://scalameta.org/metals/docs/

* table features vs editor support

* list of features on a single page
  * clear and concise description
  * demo gif
  * known limitations

jneira:
Wow that is a great example, many thanks for noting it, a pr inspired by will be very welcomed :slightly_smiling_face:

jneira:
@GuillaumeDesforges there is a pr improving the features section a lot, good enough to close this issue, It would be great if you could take a look

Anton-Latukha:
> demo gif

*Nota bene*: How to do quality GIFs Unix-way (at the time of this code - there was no quality apps for it):

So the hardcore way is:
Screen capture on Linux (or on any system that uses X11 server) can be as simple as:
https://github.com/Anton-Latukha/fish-functions/blob/master/xgrab-pulse.fish

Capturing audio is not important.

Lossless (`-crf 0`) x264 video capturing of the whole desktop works pretty great. Lossless video compression works great on encoding computer interfaces. It is also possible to capture only a part of the screen through the coordinate system. Or crop the video afterward & keep output lossless. Something like `-crf 20` can be used to enable lossy encoding, but lossy encoding can be more computationally demanding than lossless one.

After the video is produced, to convert video to GIF - the:
https://github.com/Anton-Latukha/fish-functions/blob/master/gifEnc.fish

can be used to produce a quality GIF. FPS 15 is mostly Ok setting, the `scale` is self-explanatory.

If GIF is still too massive, the `flags=lanczos` enables the lossy-noisy type of algorithm that produces small GIFs, it is seen at: https://github.com/michaelpj/haskell-language-server/blob/mpj/features-doc/docs/features.md#make-import-lists-fully-explicit-1

GuillaumeDesforges:
@Anton-Latukha what about Byzanz? https://github.com/rhcarvalho/byzanz-gui

PR #2603, opened 2022-01-19
Rework features documentation

michaelpj:
REMINDER: RTD provides a preview of the rendered documentation via a PR status.

This does a few things:
- Adds a clear listing of which LSP methods we support, and which we
don't (requested a couple of times recently).
- Lists every code action and code lens, which plugin provies them,
their status, and code action kind where applicable (apart from Wingman
because there are lots and I couldn't figure it all out).

Possibly some of this information belongs elsewhere, e.g. perhaps
instead of listing which versions don't have full support on this page,
we should instead add a page to the GHC support page that lists which
plugins work on which versions. Not sure.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2603"></a>

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -141,8 +141,6 @@ Rewrites imported names to be qualified.
  
 For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
 
-For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
-
 ### Refine import
 Provided by: `hls-refine-imports-plugin`
 
--- a/docs/features.md
+++ b/docs/features.md
@@ -1,58 +1,266 @@
 # Features
 
-You can watch demos for some of these features [below](#demos).
+This table gives a summary of the features that HLS supports.
+Many of these are standard LSP features, but a lot of special features are provided as [code actions](#code-actions) and [code lenses](#code-lenses).
 
-- Warning and error diagnostics from GHC
-- Type information and documentation on hover, [including your own comments](./configuration.md#how-to-show-local-documentation-on-hover).
-- Jump to definition: [for now only for local code definitions](https://github.com/haskell/haskell-language-server/issues/708)
-- Document symbols
-- Highlight references in document
-- Code completion
-- Formatting via [Brittany](https://github.com/lspitzner/brittany), [Floskell](https://github.com/ennocramer/floskell), [Fourmolu](https://github.com/fourmolu/fourmolu), [Ormolu](https://github.com/tweag/ormolu) or [Stylish Haskell](https://github.com/haskell/stylish-haskell)
-- [Code evaluation](#code-evaluation), see its [Tutorial](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/README.md)
-- [Integration with](#retrie-integration) [retrie](https://hackage.haskell.org/package/retrie), a powerful, easy-to-use codemodding tool
-- [Code lenses for explicit import lists](#explicit-import-lists)
-- [Generate functions from type signatures, and intelligently complete holes](#wingman) using [Wingman (tactics)](https://github.com/haskell/haskell-language-server/tree/master/plugins/hls-tactics-plugin)
-- [Integration](#hlint) with [hlint](https://github.com/ndmitchell/hlint), the most used haskell linter, to show diagnostics and apply hints via [apply-refact](https://github.com/mpickering/apply-refact)
-- [Module name suggestions](#module-names) for insertion or correction
-- [Call hierarchy support](#call-hierarchy)
-- [Qualify names from an import declaration](#qualify-imported-names) in your code
-- [Suggest alternate numeric formats](#alternate-number-formatting)
+| Feature                                             | Status                          | [LSP method](./what-is-hls.md#lsp-terminology)                                                    |
+|-----------------------------------------------------|---------------------------------|---------------------------------------------------------------------------------------------------|
+| [Diagnostics](#diagnostics)                         | Limited on some versions of GHC | `textDocument/publishDiagnostics`                                                                 |
+| [Hovers](#hovers)                                   | Working                         | `textDocument/hover`                                                                              |
+| [Jump to definition](#jump-to-definition)           | Working                         | `textDocument/definition`                                                                         |
+| [Jump to type definition](#jump-to-type-definition) | Working                         | `textDocument/typeDefinition`                                                                     |
+| [Find references](#find-references)                 | Working                         | `textDocument/references`                                                                         |
+| [Completions](#completions)                         | Working                         | `textDocument/completion`                                                                         |
+| [Formatting](#formatting)                           | Limited on some versions of GHC | `textDocument/formatting`, `textDocument/rangeFormatting`                                         |
+| [Document symbols](#document-symbols)               | Working                         | `textDocument/documentSymbol`                                                                     |
+| [Workspace symbols](#workspace-symbols)             | Working                         | `workspace/symbol`                                                                                |
+| [Call hierarchy](#call-hierarchy)                   | Working                         | `textDocument/prepareCallHierarchy`, `callHierarchy/incomingCalls`, `callHierarchy/outgoingCalls` |
+| [Highlight references](#highlight-references)       | Working                         | `textDocument/documentHighlight`                                                                  |
+| [Code actions](#code-actions)                       | Limited on some versions of GHC | `textDocument/codeAction`                                                                         |
+| [Code lenses](#code-lenses)                         | Limited on some versions of GHC | `textDocument/codeLens`                                                                           |
 
-## Demos
+The individual sections below also identify which [HLS plugin](./what-is-hls.md#hls-plugins) is responsible for providing the given functionality, which is useful if you want to raise an issue report or contribute!
 
-### Code evaluation
+## Diagnostics
 
-![Eval Demo](https://raw.githubusercontent.com/haskell/haskell-language-server/master/plugins/hls-eval-plugin/demo.gif)
+### GHC compiler errors and warnings
+Provided by: `ghcide`
 
-### Retrie integration
+Provides errors and warnings from GHC as diagnostics.
 
-![Retrie Demo](https://i.imgur.com/Ev7B87k.gif)
+### Hlint hints
+Status: Not supported on GHC 9.2
 
-### Explicit import lists
+Provided by: `hls-hlint-plugin`
 
-![Imports code lens Demo](https://imgur.com/pX9kvY4.gif)
+Provides hlint hints as diagnostics.
 
-### Wingman
+## Hovers
+Provided by: `ghcide`
 
-![Wingman Demo](https://user-images.githubusercontent.com/307223/92657198-3d4be400-f2a9-11ea-8ad3-f541c8eea891.gif)
+Type information and documentation on hover, [including from local definitions](./configuration.md#how-to-show-local-documentation-on-hover).
 
-### Hlint
+## Jump to definition
+Provided by: `ghcide`
 
-![Hlint Demo](https://user-images.githubusercontent.com/54035/110860028-8f9fa900-82bc-11eb-9fe5-6483d8bb95e6.gif)
+Jump to definition, [for now only for local code definitions](https://github.com/haskell/haskell-language-server/issues/708).
 
-### Module names
+## Jump to type definition
+Provided by: `ghcide`
 
-![Module Name Demo](https://user-images.githubusercontent.com/54035/110860755-78ad8680-82bd-11eb-9845-9ea4b1cc1f76.gif)
+## Find references
+Provided by: `ghcide`
 
-### Call hierarchy
+Find references to a name within the project.
+
+## Completions
+
+### Code completions
+Provided by: `ghcide`
+
+- Completion of names from qualified imports.
+- Completion of names from non-imported modules.
+
+### Pragma completions
+Provided by: `hls-pragmas-plugin`
+
+Completions for language pragmas.
+
+## Formatting
+Format your code with various Haskell code formatters.
+
+| Formatter       | Status                          | Provided by                  |
+|-----------------|---------------------------------|------------------------------|
+| Brittany        | Not supported on GHC 9.2        | `hls-brittany-plugin`        |
+| Floskell        | Working                         | `hls-floskell-plugin`        |
+| Fourmolu        | Working                         | `hls-fourmolu-plugin`        |
+| Ormolu          | Working                         | `hls-ormolu-plugin`          |
+| Stylish Haskell | Not supported on GHC 9.0 or 9.2 | `hls-stylish-haskell-plugin` |
+
+## Document symbols
+Provided by: `ghcide`
+
+Provides listing of the symbols defined in a module, used to power outline displays.
+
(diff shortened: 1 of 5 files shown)


michaelpj:
Also I would like to encourage plugin authors to provide slightly more substantial feature descriptions than my one-sentence summaries :sweat_smile:

michaelpj:
Possible changes I thought about:
- GHC version compatibility in a table on the GHC support page instead?
- Code action kinds for each code action in a table somewhere instead of spread out?

michaelpj:
Hmm, so maybe it would help to call out that you may need to consult your client to find out how to actually trigger these features?

jneira:
there is an issue in this repo asking for an improvement of the section changed here: https://github.com/haskell/haskell-language-server/issues/2327
not sure if the pr fullfills some of the suggestions done there (or if it could do it)

michaelpj:
I'm not convinced by the editor table, I think users should just consult the documentation for the editor, but otherwise I think this gets us a lot closer. Known limitations would be great... I just don't know the limitations in most cases. I think we'd need some help from plugin authors to make this significantly better.

jneira (approved):
great work as usual, I like the actual version but the alternatives sounds good as well

michaelpj:
Ah yeah, my concern with the GHC version support stuff being spread out is that it's harder to maintain. Which is maybe the killer argument. This information is no good if it's out-of-date, and the GHC version support page is naturally the place we're going to think to update when we do GHC version support stuff. So for that reason I think it should go there, maybe with a link from here.

GuillaumeDesforges:
Looks nice!

I've tried using Ctrl+F on some simple things I'd look for, like "hlint" or "explicit", and I got the info right away 👍 
I especially like the "missing features" part, which helps quickly figuring out why something can't be done (e.g. renaming).

docs/features.md
```diff
@@ -1,58 +1,266 @@
...
+| Signature help         | Unimplemented                                                                            | `textDocument/signatureHelp`                        |
+| Jump to declaration    | Unclear if useful                                                                        | `textDocument/declaration`                          |
+| Jump to implementation | Unclear if useful                                                                        | `textDocument/implementation`                       |
+| Renaming               | [Parital implementation](https://github.com/haskell/haskell-language-server/issues/2193) | `textDocument/rename`, `textDocument/prepareRename` |
+| Folding                | Unimplemented                                                                            | `textDocument/foldingRange`                         |
+| Selection range        | Unimplemented                                                                            | `textDocument/selectionRange`                       |
```
jneira, line 260: does it worth mention the wip pr #2565?

michaelpj: Eh, I'd wait until it's merged. It might never be, who knows!

michaelpj:
- Moved plugin version support table to GHC support page
- Added a few known limitations
- Removed fairly pointless "Status" column from main feature table

Anton-Latukha:
Personally, I think that in HLS docs it is sound to have plugin API info & examples how to develop them. Maybe include info/statuses on the closes built-in plugins, but not on all of them.

And lists of plugins, their description & status updates seem like should be in a separate repo. A repo that becomes a db on plugins & their support statuses.
That would allow to not get notifications & to not review the plugin list status update PRs.

Also woul'd be good if plugins could themselves tell which thing they support. That again goes to the coupling, if plugins are bundles - to tell is it supported or not - the closest people who have current updates on that info become core team responcibility.

What do you think on this?

michaelpj:
> 1) plugin API info 2) examples on how to develop them

I don't think either of those are appropriate for a features page aimed at users of HLS, though?

As for the rest... the point of this PR is just to make the features page better, not redo the entire way that plugins are organized and documented!

eddiemundo:
Some of the plugins have more substantial READMEs, should we inline some or most of that information here?

michaelpj:
> Some of the plugins have more substantial READMEs, should we inline some or most of that information here?

Maybe. Probably some! I think we should keep this page relatively lightweight, and maybe if there are plugins that have lots to say (e.g. I'm sure there is lots to say about Wingman), we could add a documentation page specifically for them.

eddiemundo:
Ah OK I'm thinking now we could just link to the readmes from this page if they have any.

docs/features.md
```diff
@@ -1,58 +1,255 @@
...
+Make import lists fully explicit (same as the code lens).
+
 ### Qualify imported names
+Provided by: `hls-qualify-imported-names-plugin`
+
+Code action kind: `quickfix`
+
+Rewrites imported names to be qualified.
 
 ![Qualify Imported Names Demo](../plugins/hls-qualify-imported-names-plugin/qualify-imported-names-demo.gif)
```
eddiemundo, line 141:
```suggestion
For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).

```

docs/features.md
```diff
@@ -1,58 +1,256 @@
...
+For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
```
eddiemundo, line 141:
```suggestion

For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
```

docs/features.md
```diff
@@ -138,6 +138,8 @@ Code action kind: `quickfix`
 Rewrites imported names to be qualified.
 
 ![Qualify Imported Names Demo](../plugins/hls-qualify-imported-names-plugin/qualify-imported-names-demo.gif)
+ 
+For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
 
 For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
```
eddiemundo, line 144:
```suggestion
```

docs/features.md
```diff
@@ -1,58 +1,259 @@
...
+
+Code action kind: `quickfix`
+
+Rewrites imported names to be qualified.
 
 ![Qualify Imported Names Demo](../plugins/hls-qualify-imported-names-plugin/qualify-imported-names-demo.gif)
+ 
+For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
+
+For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
```
eddiemundo, line 144:
```suggestion
```

docs/features.md
```diff
@@ -1,58 +1,257 @@
...
 ### Qualify imported names
+Provided by: `hls-qualify-imported-names-plugin`
+
+Code action kind: `quickfix`
+
+Rewrites imported names to be qualified.
 
 ![Qualify Imported Names Demo](../plugins/hls-qualify-imported-names-plugin/qualify-imported-names-demo.gif)
+ 
+For usage see the ![readme](plugins/hls-qualify-imported-names-plugin/README.md).
```
eddiemundo, line 142:
```suggestion
For usage see the ![readme](/plugins/hls-qualify-imported-names-plugin/README.md).
```

docs/features.md
```diff
@@ -1,58 +1,257 @@
...
 ### Qualify imported names
+Provided by: `hls-qualify-imported-names-plugin`
+
+Code action kind: `quickfix`
+
+Rewrites imported names to be qualified.
 
 ![Qualify Imported Names Demo](../plugins/hls-qualify-imported-names-plugin/qualify-imported-names-demo.gif)
+ 
+For usage see the ![readme](/plugins/hls-qualify-imported-names-plugin/README.md).
```
eddiemundo, line 142:
```suggestion
For usage see the ![readme](../plugins/hls-qualify-imported-names-plugin/README.md).
```

malteneuss:
> someone in the vscode extension asked recently for info like this (also key bindings but that is totally editor specific)
> haskell/vscode-haskell#525

Regarding documentation of key-bindings in HLS clients/editors, which is important for end-users, i'd like to propose the following idea as some future improvement to minimize, repeated manual work:

Could we mold the feature information into structured data, and then generate this page/table based on that?
Basically, provide (LSP Protocoll functions <-> Descriptions, Status, other Info). Then a client like vscode can provide (keybindings in vscode <-> LSP Protocoll functions), and combine both to generate a (keybindings in vscode <-> Descriptions, Status, other Info) Table on their documentation page.

That could be repeated for every HLS client in other editors.

GuillaumeDesforges:
Would it be possible to add a link to each plugin's homepage, that that users can go learn more about them?

michaelpj:
> Would it be possible to add a link to each plugin's homepage, so that users can go learn more about them?

@eddiemundo has been doing that, which is great. At the moment the existence of these is a bit spotty. I'd like to make them a more "proper" part of the documentation build, but that can happen later.

> Could we mold the feature information into structured data, and then generate this page/table based on that?
Basically, provide (LSP Protocoll functions <-> Descriptions, Status, other Info). Then a client like vscode can provide (keybindings in vscode <-> LSP Protocoll functions), and combine both to generate a (keybindings in vscode <-> Descriptions, Status, other Info) Table on their documentation page.

I guess my feeling is that the point of the LSP is that you get to do #clients + #servers work rather than #clients*#servers, and so I kind of want the same for docs: let the user put the server doc and the client doc together and do the "join" themselves. But then that is less user friendly.

Also we *can't* give keybindings for emacs in general since they're mapped to all sorts of things by various emacs distributions!

Anyway, a few additional requests in this thread, I'll make some issues for them so they don't get lost.

michaelpj:
Better plugin doc: https://github.com/haskell/haskell-language-server/issues/2066
Keybindings: https://github.com/haskell/haskell-language-server/issues/2613

docs/supported-versions.md
```diff
@@ -26,6 +26,30 @@ GHC versions not in the list have never been supported by HLS, or are not planne
 
 The policy for when we deprecate support for versions of GHC is given below. The table reflects that, but we may decide to deviate from it for good reasons.
 
+Additionally, some plugins are not supported on some GHC versions, as shown in the following table.
+
+| Plugin                              | Unsupported versions            |
```
Anton-Latukha, line 31: Mention that it is GHC versions.

Also stating in HLS docs in article "Supported versions" that some plugins correspond to some Unsupported GHC versions - it raises a question, it is HLS not supports them for this GHCs?

"Additionally, some plugins are not supported on some GHC versions, as shown in the following table." only magnifies that belief: plugins are really are unsupported by HLS for particular versions. When the implication is in a different direction. It is plugins that do not support new GHCs & HLS releases that support those GHCs.

Stating: "Bounds on the support of GHC releases" & declaring the bounds below as Cabal bounds - would align the knowledge & understanding.

The first culprit of a particular plugin - is that it needs to support new GHC, which means support HLS API for that release, which means person goes - looks at `.cabal` description of the plugin in the mainline branch, clones & tweaks the boundary & observes the result & starts from there, & goes lurking on according reports/PRs.

Anton-Latukha:
https://github.com/haskell/haskell-language-server/pull/2603#issuecomment-1016797883 well, sometimes a typo changes the reading.

The main point there - is that, as I think of it, plugins better to not be included into the documentation of HLS at all, only info on how to develop them & have "for additional features/to expend HLS - use plugins: <link to the plugin catalog>".

---

https://github.com/haskell/haskell-language-server/pull/2603#issuecomment-1016826913

> As for the rest... the point of this PR is just to make the features page better, not redo the entire way that plugins are organized and documented!

Nothing is more permanent than the temporary. Starting to catalog feature-by-feature catalog plugin abilities on `Features` page, means that users would grow to expect that all feature information cataloged there in this form, and that in short order plugins & users would start opening PRs to add their info into this article, which means permanent maintenance of the page, until design changes.

Besides even the scope of `haskell-language-server` is anthropologically perceived as development, not documentation. Because of the code development scope, people tend to treat repo as a code development repository and so completely forget about documentation.

"Features" page can be useful to state the basic functionality or the core projects of HLS & it is a great way to empirically transmit the structural knowledge/understanding. It is one thing to put in words what `ghcide` is & the other to know domain of user interface things that it does, knowing that it handles so and so - means that this code processing is present & the agenda of `ghcide` & so agenda aligns and code is in there & improvements to it would be accepted. It is overall also important that it forms the overall understanding of what `ghcide` does when to look from the user, which deepens the HLS stack understanding. But when all features are mentioned & `ghcide` is mentioned between a lot of plugins - that mixes-up the understanding & so `ghcide` agenda, scope & functionality as a whole stays not understood by the reader.

I do not want to draw the process in my thoughts. Just wanted to drive attention to the points that nothing is more permanent than the temporary, and that design decisions and maintenance required by that design matters.

Just giving an observation.

docs/features.md
```diff
@@ -1,58 +1,257 @@
...
+## Code actions
+
+### Insert missing pragmas
+Provided by: `hls-pragma-plugin`
+
+Code action kind: `quickfix`
+
+Inserts missing pragmas needed by GHC.
+
+### Apply Hlint fixes
```
Anton-Latukha, line 116: This mentioning then probably needs to mention the caveat, that during doing that (for sure during the "apply all hints" - it frequently deletes CPP code blocks across the whole file.

Anton-Latukha: Sorry, this may be out of date, because it appears I still use:
```
hlint --version
HLint v3.2.7
```
While there is `3.3.6`.

malteneuss:
> I guess my feeling is that the point of the LSP is that you get to do #clients + #servers work rather than #clients*#servers, and so I kind of want the same for docs: let the user put the server doc and the client doc together and do the "join" themselves. But then that is less user friendly.

"#clients + #servers work .. for docs" is a nice summary what i had in mind. My proposal is to not let the users "join" that info (as you said as well); let the HLS clients do that (let them worry about their keybindings). But possibly we can make their join work much easier by agreeing on and providing features info as structured data, so that the join can be automated on their side.

pepeiborra:
This is super-awesome, thank you @michaelpj

docs/supported-versions.md
michaelpj: I'm sorry, I'm not sure I quite understand what you're saying. I will try and make the text clearer so that it's obvious that the lack of support is because the _plugin_ lacks support for that version of GHC.

michaelpj:
> so users would grow to expect that all plugin feature information cataloged there in this form, and that in short order plugins & users would start opening PRs to add their info into this article, which means permanent maintenance of the page, until design changes.

That all sounds pretty good to me. At the moment we have basically no content about these things, so having some is a distinct improvement.

> But when all features are mentioned & ghcide is mentioned between a lot of plugins - that prevents to empirically infer the ghcide understanding & so scope & functionality of it stay not understood by the reader.

I think we still have a disconnect about who the _audience_ of this page is. I think the audience of this page is end-users, i.e. people who _use_ HLS but don't contribute. Understanding the structure of HLS is of no use to them, but knowing that a feature is associated with `ghcide` might be of some use.

Explaining the structure can be usefully done as part of the _contributing_ docs, IMO.

docs/features.md
michaelpj: No, this is an `apply-refact` limitation AFAICT, and good to include. I've added it.

michaelpj:
Okay, I think this is good enough state now.

docs/supported-versions.md
Anton-Latukha: Yep.

Despite using English, sometimes I still do basic errors, like making a question while sequencing the words as a statement.

There is still a structural question, between HLS-plugin & GHC, there is a major structural part - HLS, but the table directs to compatibility with GHC. But that does not matter, users/contributors would raise the question if they would need it.

PR #2604, opened 2022-01-19
Correct typo in Ide.Arguments:listPluginsParser

tombusby:
s/avaliable/available


jneira (approved):
Good catch, many thanks!

PR #2606, opened 2022-01-19
Make work stack-9.2.1.yaml and enable `pedantic` (`-WError`) for cabal 

jneira:
* Going deeper underground:
1. Make stack build work for ghc-9.2.1
2. It turns out that it has enable `pedantick` for the hls packahe so i had to fix somewarnins in `func-test`
3. I remembered @michaelpj and me talked about make consistent the use of the flag
4. Enable the flag for the default cabalbuild
5. Fix a good bunch of warning everywhere 😆 , several of them in tactics


cabal.project
```diff
@@ -34,6 +34,8 @@ optional-packages: vendored/*/*.cabal
 
 tests: true
 
+flags: +pedantic
```
michaelpj, line 37: I'm not sure we should do this here, rather only do it in CI. It's *really* annoying to develop with `-Werror` on!

plugins/hls-tactics-plugin/src/Wingman/Metaprogramming/Parser/Documentation.hs
```diff
@@ -6,8 +6,8 @@ import           Data.Functor ((<&>))
 import           Data.List (sortOn)
 import           Data.String (IsString)
 import           Data.Text (Text)
-import           Data.Text.Prettyprint.Doc hiding (parens)
-import           Data.Text.Prettyprint.Doc.Render.String (renderString)
+import           Prettyprinter hiding (parens)
```
michaelpj, line 9: We tried this before! https://github.com/haskell/haskell-language-server/pull/2389

michaelpj: Maybe we should just turn off deprecation warnings on these files locally?

stack-9.0.2.yaml
```diff
@@ -20,7 +20,7 @@ packages:
 - ./plugins/hls-retrie-plugin
 - ./plugins/hls-splice-plugin
 # - ./plugins/hls-tactics-plugin
```
isovector, line 22: i guess we should enable tactics here too!

isovector (approved):
lgtm, thanks

stack-9.0.2.yaml
jneira: oh good catch

plugins/hls-tactics-plugin/src/Wingman/Metaprogramming/Parser/Documentation.hs
jneira: ouch, luckily you has noted it 😅 
maybe we can add cpp on prettyprinter version

cabal.project
jneira: hmm i prefer doing it right step by step instead a final cleaning out, but fair enough

cabal-ghc921.project
```diff
@@ -27,16 +27,6 @@ packages:
          ./plugins/hls-call-hierarchy-plugin
          ./plugins/hls-alternate-number-format-plugin
 
-repository head.hackage.ghc.haskell.org
```
jneira, line 30: Ah, i also tried to remove head.hackage to see we can take ride of

michaelpj:
Why not just put `-Wno-deprecations` on the files that have the prettyprinter imports?

jneira:
> Why not just put `-Wno-deprecations` on the files that have the prettyprinter imports?

Well it will silence other possible deprecations, but i dont like the CPP. 
The right one would be removing the deprecation warning only for those definitions but ghc does not have those things.

michaelpj:
I think deprecations are comparatively rare, so IMO it's less bad than the CPP :shrug:

cabal.project
michaelpj: Running with `-Werror` means that if you have a binding that's temporarily unused then it's an error and nothing else will compile. It's really pretty bad IMO.

jneira:
stack build does not work anymore, since https://app.circleci.com/pipelines/github/jneira/haskell-language-server/1506/workflows/bf168e6e-0c8e-4784-95a1-c98d555c5358/jobs/12418 😢

jneira:
I am trying to fix the build release workflow here and i am afraid we have a new dependency which needs integer-gmp. As we are building the binaries for linux with alpine where some ghcs are using integer-simple, the build fails:

```
Resolving dependencies...
[__0] trying: ghcide-1.5.0.1 (user goal)
[__1] trying: lsp-types-1.4.0.0 (dependency of ghcide)
[__2] trying: mod-0.1.2.2 (dependency of lsp-types)
[__3] next goal: integer-gmp (dependency of mod)
[__3] rejecting: integer-gmp-1.0.3.0, integer-gmp-1.0.2.0,
integer-gmp-1.0.1.0, integer-gmp-1.0.0.1, integer-gmp-1.0.0.0,
integer-gmp-0.5.1.0 (only already installed instances can be used)
[__3] fail (backjumping, conflict set: integer-gmp, mod)
goals I've had most trouble fulfilling: base, ghcide, integer-gmp, lsp-types,
mod
```
afaics [mod ](https://hackage.haskell.org/package/mod) has no flag to switch to `integer-simple` like we are doing for `blaze-textual`

//cc @michaelpj @Bodigrim

michaelpj:
I suggest we leave the integer-simple issue for another PR since this is getting big.

.github/workflows/flags.yml
```diff
@@ -57,20 +61,19 @@ jobs:
...
-        run: cabal v2-build hls-test-utils --flags="pedantic"
-
-      - name: Build
+      - name: Build `ghcide` with flags
         run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe"
 
+      # we have to clean up warnings for 9.0 and 9.2 before enable -WAll
+      - if: matrix.ghc != '9.0.2' && matrix.ghc != '9.2.1'
+        name: Build with pedantic (-WError)
+        run: cabal v2-build --flags="pedantic"
```
michaelpj, line 75: This will rebuild everything, I think? do we want that? seems like it'll make the workflow slow. Maybe it's okay.

jneira: only local packages (as `--flags` only affect them), even the ones without the flag as we have one package in the root of the local dependency tree with the flag available

jneira:
> I suggest we leave the integer-simple issue for another PR since this is getting big.

yeah i've already accumulated too many things here

jneira:
@michaelpj it is the pr ok to merge (ignoring fo now the release build failure)?

michaelpj (approved):

PR #2608, opened 2022-01-19
Wingman: Add support for getting the current selector

isovector:
## selector

arguments: none.  
deterministic.

> Use the currently-being-built field selector.

### Example

> In the context of `f = Blah { bar = _ }`.
Given:

```haskell
_ 
```

running  `selector` will produce:

```haskell
bar
```

### Example

> In the context of `f = Blah { bar = Foo { foo = _ } }`.
Given:

```haskell
_ 
```

running  `selector` will produce:

```haskell
foo
```
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2608"><img src="https://gitpod.io/button/open-in-gitpod.svg"/></a>

### CODE REVIEW

**Comment by isovector** [comment]
This should also have support for the current data con

**Comment by isovector** [comment]
and it should also give the selector type to the selector:

```haskell
                  . maybe id ( withNewSelector 
                             . uncurry Selector 
                             . second (CType . mkVisFunTy (mkTyConTy $ dataConTyCon dc) . unCType)
                             ) fld

```

jneira (approved):

michaelpj:
Genuine test failure?

isovector:
Yeah, genuine failure. Not sure why yet.

PR #2609, opened 2022-01-19
Wingman: Stringify command

isovector:
## stringify

arguments: tactic.  
deterministic.

> Pretty print the result of a tactic as a string literal.

### Example

> The tactic parameter is run in the same environment as `stringify`, except its goal is a fresh type variable. 
Given:

```haskell
_ :: String
```

running  `stringify (ctor False)` will produce:

```haskell
"False"
```


isovector:
This should support overloaded strings

jneira (approved):

michaelpj:
Has conflicts.

PR #2614, opened 2022-01-20
Use helpers from lsp to do code action prefixing

michaelpj:
Nicer version as promised in
https://github.com/haskell/haskell-language-server/pull/2146.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2614"></a>


jneira (approved):
nice, many thanks to fulfill the promise 😄

PR #2615, opened 2022-01-20
Use ghc+integer-gmp for alpine linux build release

jneira:
* [After diagnose the relese build was failing](https://github.com/haskell/haskell-language-server/pull/2606#issuecomment-1017227934) for alpine+ghc with integer-simple due to a new package requiring integer-gmp (https://hackage.haskell.org/package/mod)
* Pinged @maerwald and kindly added in ghcup the missinf ghc versions for alpine with integer-gmp (https://github.com/haskell/ghcup-metadata/pull/15)
* So we can remove the existing workaround here and let haskell/action download the new bindists

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2615"></a>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,8 +43,8 @@ jobs:
 
     - name: Disable tests and bechmarks
       run: |
-        echo -e 'tests: false' >> cabal.project.local
-        echo -e 'benchmarks: false' >> cabal.project.local
+        echo "tests: false" >> cabal.project.local
+        echo "benchmarks: false" >> cabal.project.local
 
     - uses: ./.github/actions/setup-build
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
-    container: ${{ (startsWith(matrix.os,'ubuntu') && 'alpine:3.12') || '' }}
+    container: ${{ (startsWith(matrix.os,'ubuntu') && matrix.ghc != '9.2.1' && 'alpine:3.12') || '' }}
     defaults:
       run:
         shell: ${{ (startsWith(matrix.os,'windows') && 'bash') || 'sh' }}
@@ -35,7 +35,7 @@ jobs:
 
     steps:
     - name: Install system dependencies
-      if: runner.os == 'Linux'
+      if: runner.os == 'Linux' && matrix.ghc != '9.2.1'
       run: |
         apk add --no-cache curl gcc g++ gmp-dev ncurses-dev libffi-dev make xz gzip tar perl git bash sudo binutils-gold
         apk add --no-cache zlib zlib-dev zlib-static gmp gmp-dev ncurses-static
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,7 +98,7 @@ jobs:
         fi
 
     - name: (not check) Upload server to release
-      if: ${{ !contains(github.ref_name, 'check') }}
+      if: ${{ !contains(github.ref_name, 'check') && github.event.release.upload_url != ''}}
       uses: actions/upload-release-asset@v1.0.2
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -138,7 +138,7 @@ jobs:
         fi
 
     - name: (GHC 8.10, not check) Upload wrapper to the release
-      if: ${{ matrix.ghc == '8.10.7' && !contains(github.ref_name, 'check') }}
+      if: ${{ matrix.ghc == '8.10.7' && !contains(github.ref_name, 'check') && github.event.release.upload_url != '' }}
       uses: actions/upload-release-asset@v1.0.2
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -172,7 +172,7 @@ jobs:
                       HEAD
 
       - name: (not check) Upload source tarball to the release
-        if: ${{ !contains(github.ref_name, 'check') }}
+        if: ${{ !contains(github.ref_name, 'check')  && github.event.release.upload_url != '' }}
         uses: actions/upload-release-asset@v1.0.2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -219,7 +219,7 @@ jobs:
           tar -czpf haskell-language-server.tar.gz *
 
       - name: (not check) Upload binaries tarball to the release
-        if: ${{ !contains(github.ref_name, 'check') }}
+        if: ${{ !contains(github.ref_name, 'check')  && github.event.release.upload_url != '' }}
         uses: actions/upload-release-asset@v1.0.2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


jneira:
* Checking the release build here: https://github.com/jneira/haskell-language-server/actions/runs/1725135759

jneira:
The build has progressed but it has failed for linux and ghc-9.2.1 with

```
src/Ide/Version.hs:24:21: error:
Error:     • Exception when trying to run compile-time code:
        PATH: getEnv: does not exist (no environment variable)
      Code: gitCommitCount
    • In the untyped splice: $gitCommitCount
   |
24 |   let commitCount = $gitCommitCount
   |                     ^^^^^^^^^^^^^^^
cabal: Failed to build hls-1.5.1.0 (which is required by exe:hls from
hls-1.5.1.0).
```
🤔

hasufell:
> The build has progressed but it has failed for linux and ghc-9.2.1 with
> 
> ```
> src/Ide/Version.hs:24:21: error:
> Error:     • Exception when trying to run compile-time code:
>         PATH: getEnv: does not exist (no environment variable)
>       Code: gitCommitCount
>     • In the untyped splice: $gitCommitCount
>    |
> 24 |   let commitCount = $gitCommitCount
>    |                     ^^^^^^^^^^^^^^^
> cabal: Failed to build hls-1.5.1.0 (which is required by exe:hls from
> hls-1.5.1.0).
> ```
> 
> thinking

Sounds like this part is failing: https://github.com/acfoltzer/gitrev/blob/450a7abf444260e95791649489705e54650c0232/src/Development/GitRev.hs#L65

The environment is borked (doesn't contain the `PATH` env var).

jneira:
it does (https://github.com/jneira/haskell-language-server/runs/4889676070?check_suite_focus=true)

```
Path: /__t/stack/2.7.3/x64:/github/home/.ghcup/bin:/github/home/.ghcup/ghc/9.2.1/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

the build is successful for all ghcs < 9.2.1, using the same docker image for all, and the build for ghc-9.2.1 is successful in Ubuntu so I would say it is a weird interaction between both

jneira:
well the workflow between the Ubuntu build and the alpine one 8s not exactly the same but I don't see any change which could cause the bug
will try to use Ubuntu in the same workflow to check it though (and as possible workaround)

also we could trace the value of the PATH env bar in our code but getExecutable is pretty standard

jneira:
> the build for ghc-9.2.1 is successful in Ubuntu so I would say it is a weird interaction between both

well the ghc version for alpine is not exactly the same than the Ubuntu one

jneira:
As first sanity check i am gonna invalidate the cache created in ubuntu

jneira:
A fresh cache did not fix it: https://github.com/jneira/haskell-language-server/runs/4893126248?check_suite_focus=true

jneira:
As expected changing alpine for ubuntu for the ghc-9.2.1 has been succesful so the workflow is not the culprit: https://github.com/jneira/haskell-language-server/runs/4893355328?check_suite_focus=true

At this point i think we could consider use ubuntu for ghc-9.2.1 (it is experimental anyways) until we fix the bug for alpine + ghc-9.2.1. That way we will unblock the incoming release.

//cc @hasufell @pepeiborra @michaelpj

michaelpj:
@Bodigrim made a release of `mod` with `integer-simple` (https://github.com/Bodigrim/mod/issues/13#issuecomment-1017944515), so if we bump the index-state we may find that cabal can now get a build plan when using `integer-simple`.

michaelpj:
Personally, the commit count suffix on the version string doesn't seem that useful to me, we could consider just binning that.

hasufell:
> @Bodigrim made a release of `mod` with `integer-simple` (https://github.com/Bodigrim/mod/issues/13#issuecomment-1017944515), so if we bump the index-state we may find that cabal can now get a build plan when using `integer-simple`.

Probably won't, because I already updated ghcup alpine bindists to integer-gmp.

pepeiborra:
I support dropping the dependency if it helps.

jneira:
All `gitrev` definitions needs find git in the env var PATH throught `findExecutable "git"` . We are actually using it to get the git commit here:

https://github.com/haskell/haskell-language-server/blob/b937baa03cf310109aafc0d4835cdff1240ffd3e/src/Ide/Arguments.hs#L141-L150

and i would like to keep the commit when doing `hls --version`

The alternative `githash` dont use `findExecutable` though but does a raw `(proc "git" args) { cwd = Just root }` so maybe it would work :shrug:

jneira:
Meanwhile i've reproduced the ghc bug in a fresh alpine-3.12 vm installing ghc with ghcup:

imagen

wonderful

jneira:
It worths to note that ghc-9.2.1 in alpine seems to be fragile:
* "Alpine segfaults many tests in 9.2.1" https://gitlab.haskell.org/ghc/ghc/-/issues/20266
* "Alpine ci job is broken" https://gitlab.haskell.org/ghc/ghc/-/issues/20574

jneira:
Trying githash in alpine+9.2.1 here: https://github.com/jneira/haskell-language-server/runs/4896130017?check_suite_focus=true

jneira:
Ok, githash worked, yay! thanks @hasufell for the suggestion
I stand a little bit disturbed by the bug but whatever

hasufell:
@jneira I've updated the ghc 9.2.1 alpine bindist in ghcup. `getEnv` should work now.

jneira:
@hasufell many thanks! i've reran the build workflow and it should use the new ghc bindist: https://github.com/haskell/haskell-language-server/actions/runs/1728847536

jneira:
> @jneira I've updated the ghc 9.2.1 alpine bindist in ghcup. `getEnv` should work now.

i've double checked in my vm and it works as intended 👍

jneira:
Well i think we can keep githash cause as commented by @hasufell it is being actively maintained so the pr could be merged if you agree

pepeiborra (approved):
The githash change looks fine to me, but I haven't reviewed the CI changes. Maybe @Anton-Latukha can take a look.

michaelpj (approved):

Anton-Latukha:
Well, as I came around to look at it... To review it, I needed/need to fully learn & understand the `build` workflow, so time.

Seems good. Would submit PRs if would find something.

.github/workflows/build.yml
```diff
@@ -97,36 +84,36 @@ jobs:
...
       uses: actions/upload-artifact@v2
       with:
-        name: haskell-language-server-${{ runner.OS }}-${{ matrix.ghc }}${{env.EXE_EXT}}.${{ steps.compress_server_binary.outputs.extension }}
+        name: haskell-language-server-${{ runner.os }}-${{ matrix.ghc }}${{ env.EXE_EXT }}.${{ steps.compress_server_binary.outputs.extension }}
         path: ${{ steps.compress_server_binary.outputs.path }}
 
-    - name: (GHC 8.10) Build the wrapper
+    - name: Build the wrapper
       if: matrix.ghc == '8.10.7'
       run: cabal build exe:hls-wrapper -O2 $LINUX_CABAL_ARGS
```
pepeiborra, line 106: does this need to be `$CABAL_ARGS` ?

jneira: omg 😧

jneira: that should have failed, as @hasufell suggested, setting the appropriate bash flag

PR #2619, opened 2022-01-21
Add initial CODEOWNERS file

michaelpj:
Context: https://github.com/haskell/haskell-language-server/discussions/2588

The goal of this is to ensure that every plugin has at least one person associated with it who takes some responsibility for maintaining it over time.

The meaning of the CODEOWNERS file is: if you are listed on the RHS, you will be tagged for review on any PR touching the files on the LHS. Your review will not be required, so you won't have to look at every trivial change, but ideally you would be able to find time to look at any substantial changes.

**The users listed here are entirely my guesses, please do not be offended if you're not in there and you think you should be, or vice versa.** I've mostly just picked the original creator of the plugin, which isn't necessarily the right person. Please say if you do not want to be on the list, or if you think there is someone else who would want to be on the list. It's okay if we end up with some plugins that nobody claims responsibility for, at least then we know we have a problem :)

Additionally, my intention is to create a `hls-plugin-maintainers` Github team whose membership is the union of all the people listed here by plugins, so their attention can be drawn to e.g. GHC version compatibility work.

Mentioning all the people in this file for comment: @pepeiborra @drsooch @fendor @July541 @Ailrun @alanz @georgefst @jneira @tittoassini @berberman @eddiemundo @rayshih @OliverMadine @konn @isovector @wz1000 @Anton-Latukha @hasufell 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2619"></a>

PR #2621, opened 2022-01-21
Flake ghc 92

guibou:
This is (partial) flake support for GHC 9.2.

This is in a limited state, especially:

- I have not repaired the support for a few plugins.
- `nix develop .#....-dev` target is not working, it pulls a few dependencies which are not overiden. I'll give another pass soon.

But the cool thing is that it allows to build `haskell-language-server` from flake in another project. I've just updated my work project to GHC 9.2 and I'm enjoying HLS thank to this flake.

Tell me if you prefer to merge "partial" support or if I need to work on the currently broken tasks. I cannot guarantee that I'll allocate time to this soon.

However, users can find this MR and point their flake to this commit and immediately get HLS in their project.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2621"></a>

### DIFF
--- a/configuration-ghc-921.nix
+++ b/configuration-ghc-921.nix
@@ -1,0 +1,42 @@
+# nix version of cabal-ghc901.project
+{ pkgs, hackage }:
+
+let
+  disabledPlugins = [
+    "hls-brittany-plugin"
+    "hls-stylish-haskell-plugin"
+  ];
+
+  hpkgsOverride = hself: hsuper:
+    with pkgs.haskell.lib;
+    {
+      fourmolu = hself.callCabal2nix "fourmolu" hackage.fourmolu {};
+      primitive-extras = hself.primitive-extras_0_10_1_2;
+      ghc-exactprint = hself.callCabal2nix "ghc-exactprint" hackage.ghc-exactprint {};
+      constraints-extras = hself.callCabal2nix "constraints-extras" hackage.constraints-extras {};
+      retrie = hself.callCabal2nix "retrie" hackage.retrie {};
+      hlint = doJailbreak (hself.callCabal2nix "hlint" hackage.hlint {});
+
+      # Re-generate HLS drv excluding some plugins
+      haskell-language-server =
+        hself.callCabal2nixWithOptions "haskell-language-server" ./.
+        (pkgs.lib.concatStringsSep " " [
+          "-f-brittany"
+          "-f-stylishhaskell"
+          "-f-hlint"
+          "-f-haddockComments"
+          "-f-alternateNumberFormat"
+          "-f-eval"
+        ]) { };
+
+      # YOLO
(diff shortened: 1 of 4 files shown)


guibou:
For reference, here is the command I used in my project to upgrade HLS, where `hls` is the name of my flake input:

```
nix flake lock --override-input hls github:guibou/haskell-language-server/flake_ghc_92
```

michaelpj (approved):
Are you interested in maintaining these going forwards?

teto:
that may be the problem you are referring to but putting i there just in case since it's an area that under active development in nix so we not have the same results. With nix 2.6.0pre20211217_6e6e998
```
nix develop .#haskell-language-server-921-dev          
AWS error 'RequestTimeTooSkewed' (The difference between the request time and the current time is too large.), will retry in 0 ms
error: cannot fetch input 'path:./flake_hackage?narHash=sha256-gVzFEW0y%2fuPHJaGJ6w7VZc3QDhH6latvmYHJuzaXDa0=' because it uses a relative path
(use '--show-trace' to show detailed location information)
```

Cool work o/

michaelpj:
Ah, that's a good point. Do flakes at relative paths work reliably? I remember there being issues with that.

guibou:
> Ah, that's a good point. Do flakes at relative paths work reliably? I remember there being issues with that.

Initially I moved everything to a relative path flake because I wanted to have a flake only for the hackage dependencies and remove boilerplate, I tried that;

image

Unfortunately, it does not work, because flake does not "evaluate" nix language in the `input` section.

So actually, we can move everything back in the main `flake.nix` file. Let me do that tomorrow.

> Are you interested in maintaining these going forwards?

Yes sure.

michaelpj:
I added you to the CODEOWNERS in https://github.com/haskell/haskell-language-server/pull/2619.

smunix:
@guibou - I'm seeing this issue with the flake use at a relative path; is it possibly related to this fix?

```sh
/e/nixos (main ± u+4) ∫ sudo nixos-rebuild test --cores 10 --max-jobs 5                                             [21:08:41] 
building the system configuration...
warning: Git tree '/etc/nixos' is dirty
error: cannot fetch input 'path:./flake_hackage?narHash=sha256-%2fF3itZJSZDnaxRGpkTffHDkxC7PloYqxVTW%2f%2fbP3i20=' because it uses a relative path
(use '--show-trace' to show detailed location information)
```
I use flake with home-manager and [that used to work before](https://github.com/smunix/dratrion.nixos/blob/059c92ad8a2faddb3c6e4b10b5fab0e059f8bb1f/smunix.home.nix#L57), but it no longer does.

guibou:
@smunix yes, that's unfortunately an issue. I'm working on a fix right now. Sorry for the inconvenience.

PR #2624, opened 2022-01-22
Fix #2612 - hlint plugin - Apply fixities to parsed source before sending to apply-refact

### LINKED ISSUE #2612: Hlint "Expected type not found at the location specified in the refact file." error in certain circumstances when refactoring
### Your environment

Which OS do you use:
\- Windows
Which LSP client (editor/plugin) do you use:
\- VS Code
\- GHC 8.10.7, Cabal 3.6.2.0

### Steps to reproduce
Create a new .hs file with this code:
```haskell
import Data.List (sortOn)
import Control.Arrow ((&&&))
import Data.Ord (Down(Down))

functionA :: [String] -> [(Char,Int)]
functionA = reverse . sortOn snd . map (head &&& length)

functionB :: [String] -> [(Char,Int)]
functionB = reverse . sortOn snd . map (head &&& length) . id
```
Both functions will be highlighted by hlint telling you to avoid using the reverse
Use the quick fix feature (`Ctrl+.` in VS Code) to apply the refactor on each function

### Expected behaviour
The error should not appear, and the `avoid reverse` refactor should produce the same code with either function:
```haskell
functionB = sortOn (Down . snd) . map (head &&& length) . id
```

### Actual behaviour
Using the quick fix while the cursor is on the body of `functionA` will refactor the code as expected
However using the Quick Fix while the cursor is on the body of `functionB` will produce the follwing error:
```
hlint:applyOne: "Expected type not found at the location specified in the refact file.\n Expected type: HsExpr (GhcPass 'Parsed)\n Location: C:\\Users\\<username>\\Personal\\Hobbies\\Programming\\test.hs:9:36-61: inappropriate type"
```
Additonally the `redundant id` suggested refactor doesn't do anything at all in this circumstance.

### Include debug information
Extended error message from the logs:
```
[2022-01-20 02:12:51.997] [exthost] [error] Error: hlint:applyOne: "Expected type not found at the location specified in the refact file.\n  Expected type: HsExpr (GhcPass 'Parsed)\n  Location: C:\\Users\\<username>\\Personal\\Hobbies\\Programming\\test.hs:9:36-61: inappropriate type"
	at c:\Users\<username>\.vscode\extensions\haskell.haskell-1.8.0\dist\extension.js:2:1005073
	at c:\Users\<username>\.vscode\extensions\haskell.haskell-1.8.0\dist\extension.js:2:1005367
	at Immediate.<anonymous> (c:\Users\<username>\.vscode\extensions\haskell.haskell-1.8.0\dist\extension.js:2:1005732)
	at processImmediate (internal/timers.js:461:21) 7804:hlint:applyOne undefined
```

eddiemundo:
Turns out we weren't applying the standard fixities to the parsed source before handing it over to `applyRefactorings'`. Standalone hlint works because it calls the refactor executable which does apply the fixities before doing the refactor.

This also fixes a class of issues where applying the hlint refactor does nothing.

Closed the previous PR because I rebased wrong.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2624"></a>

### DIFF
--- a/plugins/hls-hlint-plugin/test/Main.hs
+++ b/plugins/hls-hlint-plugin/test/Main.hs
@@ -39,7 +39,7 @@ getIgnoreHintText :: T.Text -> T.Text
 getIgnoreHintText name = "Ignore hint \"" <> name <> "\" in this module"
 
 getApplyHintText :: T.Text -> T.Text
-getApplyHintText name = "Apply hint \"Avoid reverse\""  -- "Apply hint \"" <> name <> "\""
+getApplyHintText name = "Apply hint \"" <> name <> "\""
 
 ignoreHintTests :: TestTree
 ignoreHintTests = testGroup "hlint ignore hint tests"
--- a/plugins/hls-hlint-plugin/test/Main.hs
+++ b/plugins/hls-hlint-plugin/test/Main.hs
@@ -7,6 +7,7 @@ module Main
   ) where
 
 import           Control.Lens            ((^.))
+import           Control.Monad           (when)
 import           Data.Aeson              (Value (..), object, toJSON, (.=))
 import           Data.Functor            (void)
 import           Data.List               (find)
@@ -398,7 +399,7 @@ ignoreHintGoldenTest testCaseName goldenFilename point hintName =
   goldenTest testCaseName goldenFilename point (getIgnoreHintText hintName)
 
 applyHintGoldenTest :: TestName -> FilePath -> Point -> T.Text -> TestTree
-applyHintGoldenTest testCaseName goldenFilename point hintName =
+applyHintGoldenTest testCaseName goldenFilename point hintName = do
   goldenTest testCaseName goldenFilename point (getApplyHintText hintName)
 
 goldenTest :: TestName -> FilePath -> Point -> T.Text -> TestTree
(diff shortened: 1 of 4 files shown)


jneira (approved):
wow, good catch, many thanks for the fix

jneira:
the new pedantic build in action

```
src/Ide/Plugin/Hlint.hs:55:1: error: [-Wunused-imports, -Werror=unused-imports] 

Error: The qualified import of ‘Refact.Fixity’ is redundant 

except perhaps to import instances from ‘Refact.Fixity’ 

To import instances alone, use: import Refact.Fixity() 

| 

55 | import qualified Refact.Fixity as Refact 

| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -30,26 +32,40 @@ tests = testGroup "hlint" [
...
     , configTests
     , ignoreHintTests
+    , applyHintTests
     ]
 
 getIgnoreHintText :: T.Text -> T.Text
 getIgnoreHintText name = "Ignore hint \"" <> name <> "\" in this module"
 
+getApplyHintText :: T.Text -> T.Text
+getApplyHintText name = "Apply hint \"Avoid reverse\""  -- "Apply hint \"" <> name <> "\""
```
michaelpj, line 42: Commented out code, seems like this is left from debugging?

michaelpj (changes requested):

plugins/hls-hlint-plugin/test/Main.hs
eddiemundo: Yeah that was a bad mistake.

michaelpj (approved):

PR #2625, opened 2022-01-22
fix positionInRange

### LINKED ISSUE #2618: positionInRange function in PluginUtils doesn't work as its name suggests
Look closely at this function:

https://github.com/haskell/haskell-language-server/blob/b937baa03cf310109aafc0d4835cdff1240ffd3e/hls-plugin-api/src/Ide/PluginUtils.hs#L222-L227

If this range has only 1 line, and the position lies in the same line as the range, but **after** the end of the range, then,
the second branch will fail, but the third branch will succeed, which is not expected, I think.
### Your environment

Which OS do you use:
> Ubuntu 20.04

### Steps to reproduce

I wrote a testcase for that
```
tests :: TestTree
tests = testGroup "PluginUtils"
    [ testCase "positionInRange" $
        positionInRange (Position 1 10) (Range (Position 1 1) (Position 1 3)) @?= False
    ]
```

### Expected behaviour

`Position 1 10` is not in the range of `Position 1 1` and `Position 1 3`. The above testcase should pass.

### Actual behaviour
The testcase fails
image

kokobd:
fix #2618

As `hls-plugin-api` doesn't have a test suite now, I add one. I think common utilities like this should have standalone unit tests. Maybe this can serve as a starting point.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2625"></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -134,6 +134,10 @@ jobs:
         run: cabal test ghcide --test-options="$TEST_OPTS" || cabal test ghcide --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test ghcide --test-options="$TEST_OPTS"
 
       - if: matrix.test
+        name: Test hls-plugin-api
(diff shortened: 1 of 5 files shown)


pepeiborra (approved):
Thank you! This is much better.

To make sure that CI runs the test suite, please add it to `.github/workflows/test.yml`

michaelpj (approved):
Great!

Issue #758, opened 2020-03-09
Failing ghcide tests in windows


jneira:
I've run the test suite locally and in [azure ci](https://github.com/digital-asset/ghcide/issues/455) and i've got those failing tests:
```
 cpp
  cpp-error:                                        FAIL (4.16s)
diagnostics
  add missing module (non workspace):               FAIL
watched files
  non workspace file:                               FAIL
```
Full test log of failing tests here: https://gist.github.com/jneira/afe77151a540c5a27943d7661eefe3fa

PR #2627, opened 2022-01-22
Enable ghcide tests in windows

jneira:
* Check status of https://github.com/haskell/haskell-language-server/issues/758


jneira:
will open this in my fork to not waste ci resources here

Issue #2583, opened 2022-01-13
Migration to text-2.*


pepeiborra:
There are at least two dependencies that block this migration:

1. `lsp-types`: https://github.com/haskell/lsp/issues/391
2. `bytestring-encoding`: https://github.com/msakai/bytestring-encoding/issues/6

The first one is obviously a hard requirement, but for the second one there's only one call site so we could consider dropping it:

https://github.com/haskell/haskell-language-server/blob/d48a1df86640d0d8eaa3a08403301d7be101fa53/ghcide/src/Development/IDE/Core/Rules.hs#L525-L537

drsooch:
Looks like `retrie` needs it's text bounds updated as well?

Bodigrim:
`bytestring-encoding` is avoidable: `T.decode encoding <$> BS.readFile` is just `Data.Text.IO.readFile`, and `T.decode encoding` below can be replaced with `T.decodeUtf8`, because Haskell sources are always in UTF-8. 

I've successfully built HLS against `text-2.0` with this patch: https://github.com/Bodigrim/haskell-language-server/commit/e14ea011f6386911f678223e87c0b519f7843a3e

pepeiborra:
@Bodigrim would you be able to send a PR? I will gladly accept it

pepeiborra:
> Looks like `retrie` needs it's text bounds updated as well?

Opened https://github.com/facebookincubator/retrie/issues/42

Bodigrim:
@pepeiborra sorry, I'm trying really hard to keep myself away from diving into one more project ;) All you need is this:
```diff
+import qualified Data.Text.IO                                 as T
 import           Data.Time                                    (UTCTime (..))
 import           Data.Tuple.Extra
 import           Development.IDE.Core.Compile
@@ -530,10 +530,10 @@ persistentHieFileRule = addPersistentRule GetHieAst $ \file -> runMaybeT $ do
   (currentSource,ver) <- liftIO $ do
     mvf <- getVirtualFile vfs $ filePathToUri' file
     case mvf of
-      Nothing -> (,Nothing) . T.decode encoding <$> BS.readFile (fromNormalizedFilePath file)
+      Nothing -> (,Nothing) <$> T.readFile (fromNormalizedFilePath file)
       Just vf -> pure (Rope.toText $ _text vf, Just $ _lsp_version vf)
   let refmap = Compat.generateReferencesMap . Compat.getAsts . Compat.hie_asts $ res
[...]

PR #2628, opened 2022-01-23
Drop bytestring-encoding

pepeiborra:
Applying the patch suggested by @Bodigrim https://github.com/haskell/haskell-language-server/issues/2583#issuecomment-1019375604

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2628"></a>


jneira (approved):
it seems a real win-win, thanks!

Issue #2602, opened 2022-01-18
Data.Text.Internal.Fusion.Common.index: Index too large


fendor:
### Your environment

Which OS do you use: Ubuntu
Which LSP client (editor/plugin) do you use: VSCode
Describe your project (alternative: link to the project):
bios project with custom `hie.yaml`. I do not believe the project itself is important, but it does use TH.

Noticeably, this HLS version is installed via nix, is based on the hackage release 1.5.1, and applies the patch https://github.com/haskell/haskell-language-server/pull/2359.

### Steps to reproduce

Open HLS and perform some arbitrary editing of files.

### Expected behaviour

HLS should be happy and report diagnostics.

### Actual behaviour

Stops responding and shows no diagnostics any more. Logs contain:
```
haskell-language-server: Data.Text.Internal.Fusion.Common.index: Index too large
CallStack (from HasCallStack):
  error, called at libraries/text/src/Data/Text/Internal/Fusion/Common.hs:943:24 in text-1.2.4.1:Data.Text.Internal.Fusion.Common
```

### Include debug information

Since HLS has been installed via nix, I am not sure at the moment, whether an outdated dependency might cause the issue.
The error was thrown while all non-essential plugins were disabled via the VSCode settings.

jneira:
I've just seen this issue: https://github.com/haskell/haskell-language-server/issues/2149 which is reporting the same error 😟

fendor:
Ah, I knew I've seen this before! Unfortunately, no solution yet.
Going to run with profiling mode to get a proper stack-trace.

fendor:
According to #2149, the first commit with this issue is: https://github.com/haskell/haskell-language-server/commit/9c40dcff1b989b14e05b5aebe96f3da78c6ea25c

drsooch:
Just ran into this. Can't tell when the error fired (wasn't paying attention). I've been working on `TypeLenses.hs` from `ghcide/src/development/ide/plugin/`. On Emacs with all plugins enabled (plus a WIP plugin).

fendor:
Were you using a remote plugin?

Bodigrim:
https://github.com/haskell/haskell-language-server/blob/ffd0f346d88131f0238d266cb59ef2cf21329edb/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs#L733-L736
This is wrong, should be `backtickIndex >= T.length line`.

Bodigrim:
[...]

PR #2629, opened 2022-01-23
Fix off by one indexing error in openingBacktick

pepeiborra:
Apply the fix suggested by @Bodigrim in https://github.com/haskell/haskell-language-server/issues/2602#issuecomment-1019344264

Fixes https://github.com/haskell/haskell-language-server/issues/2602

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2629"></a>


jneira (approved):
lgtm

PR #2630, opened 2022-01-23
Avoid out-of-bounds index access

### LINKED ISSUE #2602: Data.Text.Internal.Fusion.Common.index: Index too large

fendor:
Closes #2602 

Thanks to @Bodigrim for finding this code!

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2630"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -732,7 +732,7 @@ isUsedAsInfix line prefixMod prefixText pos
 
 openingBacktick :: T.Text -> T.Text -> T.Text -> Position -> Bool
 openingBacktick line prefixModule prefixText Position { _character=(fromIntegral -> c) }
-  | backtickIndex < 0 || backtickIndex > T.length line = False
+  | backtickIndex < 0 || backtickIndex >= T.length line = False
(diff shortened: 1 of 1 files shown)


fendor:
Superseded by #2629

PR #2631, opened 2022-01-23
Specialize ghcide indent style to .hs

mrgutkun:
This way `ghcide/.editorconfig` reproduces the root `.editorconfig`, except for line length restrictions.

Motivation: same as root, `ghcide/` contains some YAML files, which are de facto indented with 2 spaces.


pepeiborra (approved):

pepeiborra:
Accepted, thanks

PR #2637, opened 2022-01-24
Use T.decodeUtf8 + BS.readFile instead of T.readFile

Bodigrim:
As documentation (https://hackage.haskell.org/package/text-2.0/docs/Data-Text-IO.html#v:readFile) explains, `T.readFile` is almost always a mistake. It's highly unlikely that files we try to read with `T.readFile` are in a locale encoding and not in UTF-8. Moreover, with `text-2.0` `fmap T.decodeUtf8 . BS.readFile` is up to 10x faster than `T.readFile`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2637"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -91,7 +91,6 @@ import           Data.Maybe
 import qualified Data.Rope.UTF16                              as Rope
 import qualified Data.Set                                     as Set
 import qualified Data.Text                                    as T
-import qualified Data.Text.IO                                 as T
 import qualified Data.Text.Encoding                           as T
 import           Data.Time                                    (UTCTime (..))
 import           Data.Tuple.Extra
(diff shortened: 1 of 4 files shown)


pepeiborra (approved):
Sounds like we need to deprecate T.readFile and introduce T.readFileUtf8 ?

jneira:
> It's highly unlikely that files we try to read with T.readFile are in a locale encoding and not in UTF-8

But it is more than possible that some users, specially in windows are using a local encoding, even without being aware of it.
I guess this pr would break hls for them
so imo at least a warning should be added in the changelog (will do myself)
Ideally a LSP alert warning should be added when loading a file if we detect its encoding is not utf8, will open an issue about

jneira:
Opened #2641

jneira (approved):
lgtm, thanks!

Bodigrim:
> But it is more than possible that some users, specially in windows are using a local encoding, even without being aware of it.
> I guess this pr would break hls for them

No, it's actually vice versa. This PR **fixes** HLS for people, whose locale is not UTF-8. GHC [mandates](https://downloads.haskell.org/ghc/9.2.1/docs/html/users_guide/separate_compilation.html?highlight=utf#haskell-source-files) that source files are in UTF-8 only, but before this PR HLS would try to use a locale encoding instead.

Issue #2617, opened 2022-01-21
Restore consistency in ci using alpine in all workflows


jneira:
* As commented here: https://github.com/haskell/haskell-language-server/issues/2461#issuecomment-991686414
* Sadly my pessimistic concern has come true and we hit two errors in the alpine build for the release that could have been caught before if the regular test workflow uses the same env to make the build (now uses raw ubuntu):
  * First the inclusion of a new dependency using integer-gmp without flag to switch break the build. Fortunately @maerwald added missing alpine ghc bindists against integer-gmp (see https://github.com/haskell/haskell-language-server/pull/2606#issuecomment-1017227934)
  * And then we have hit a weird bug only present in alpine + ghc-9.2.1 (and no in ubuntu + ghc-9.2.1) (see https://github.com/haskell/haskell-language-server/pull/2615#issuecomment-1017919238)
* So it has become clear that we need restore the consistency between the workflows (including tests and benchmarks) and run all of them on alpine for the linux variant

michaelpj:
I'm *more* worried about the first one. If we're going to force some build configuration (e.g. using `integer-simple`) in some workflow, we should do it in all of them, otherwise things will be inconsistent.

I'm less worried about GHC bugs. They happen from time to time, but it's less likely than solver issues IMO.

jneira:
Well the first one is resolved, as .ghcup downloads ghc versions linked against integer-gmp for all our supported ghc versions. So i removed the particular build configuration for `integer-simple` and afaics the build configuration is the same now between workflows (#2615)

jneira:
> I'm less worried about GHC bugs. They happen from time to time, but it's less likely than solver issues IMO.

More frequent than desirable i am afraid. And more frequent in not so widely used systems like alpine i guess

jneira:
From https://github.com/haskell/haskell-language-server/pull/2615#issuecomment-1018402230

> It worths to note that ghc-9.2.1 in alpine seems to be fragile:
>  * "Alpine segfaults many tests in 9.2.1" https://gitlab.haskell.org/ghc/ghc/-/issues/20266
>  * "Alpine ci job is broken" https://gitlab.haskell.org/ghc/ghc/-/issues/20574

btw, someone has done some dogfooding with the alpine based hls executables? do we know they work for sure? They have not passed the test suite

jneira:
> btw, someone has done some dogfooding with the alpine based hls executables? do we know they work for sure? They have not passed the test suite

[...]

PR #2638, opened 2022-01-24
Disable alpine build by default

jneira:
* As commented in https://github.com/haskell/haskell-language-server/issues/2617#issuecomment-1019289588
* It seems binaries from the alpine build cant load files containing template haskell, checked in wsl/ubuntu by myself and in native ubuntu by @kokobd

```
Unexpected usage error
Dynamic loading not supported
```

imagen

* That behaviour was not noted explicitly in the pr (#2463) or the issue about (#2461)
* We have pending the release of 1.6.0 for January
* So this pr disables it until we clarificate the situation to unblock the release

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2638"></a>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,11 @@ name: Builds
 
 on:
   workflow_dispatch:
+    inputs:
+      alpine:
+        description: 'Use alpine to build a fully static executable for linux'
+        required: false
+        default: 'false'
   release:
     types: [created]
   push:
@@ -11,7 +16,7 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
-    container: ${{ (startsWith(matrix.os,'ubuntu') && 'alpine:3.12') || '' }}
+    container: ${{ (github.event.inputs.alpine == 'true' && startsWith(matrix.os,'ubuntu') && 'alpine:3.12') || '' }}
     defaults:
       run:
         shell: ${{ (startsWith(matrix.os,'windows') && 'bash') || 'sh' }}
@@ -35,7 +40,7 @@ jobs:
 
     steps:
     - name: Install system dependencies
-      if: runner.os == 'Linux'
+      if: github.event.inputs.alpine == 'true' && runner.os == 'Linux'
       run: |
         apk add --no-cache curl gcc g++ gmp-dev ncurses-dev libffi-dev make xz gzip tar perl git bash sudo binutils-gold
         apk add --no-cache zlib zlib-dev zlib-static gmp gmp-dev ncurses-static
--- a/.gitlab/ci.sh
+++ b/.gitlab/ci.sh
@@ -53,7 +53,7 @@ fi
 # Shorten binary names
 sed -i.bak -e 's/haskell-language-server/hls/g' \
 	   -e 's/haskell_language_server/hls/g' \
-	   haskell-language-server.cabal cabal.project
+	   haskell-language-server.cabal $CABAL_PROJECT
 sed -i.bak -e 's/Paths_haskell_language_server/Paths_hls/g' \
 	   src/**/*.hs exe/*.hs
 
@@ -77,4 +77,3 @@ cp "$(cabal list-bin ${args[@]} exe:hls-wrapper)" "$CI_PROJECT_DIR/out/haskell-l
 cp dist-newstyle/cache/plan.json "$CI_PROJECT_DIR/out/plan.json"
 
 cd "$CI_PROJECT_DIR/out/"
-
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -259,7 +259,7 @@ common tactic
     cpp-options: -Dtactic
 
 common hlint
-  if flag(hlint)
+  if flag(hlint) && (impl(ghc < 9.2.1) || flag(ignore-plugins-ghc-bounds))
     build-depends: hls-hlint-plugin ^>=1.0.0.2
     cpp-options: -Dhlint
 


pepeiborra:
Uh, good catch. We really should have dogfooded the Musl binaries. 

But I don't understand what you want to release. We know that statically linking glibc is not a good idea, so let's investigate a bit more before reverting to that. 

1. Could the problem have been caught earlier by running the test suite in the build.yml CI script?
2. Are all ghc versions affected ?
3. Are the gitlab binaries affected?

wz1000:
I'm not sure why the alpine binaries don't work, I will investigate. The alternative for now is to do what I suggested in #2431 and compile against glibc, but don't statically link it in. This can be done by using `executable-dynamic: False` instead of `executable-static: True`. We can then do what GHC does and distribute a different binary for each supported distro/glibc version.

cabal-ghc921.project
```diff
@@ -42,7 +42,7 @@ index-state: 2022-01-21T11:23:29Z
 
 constraints:
   -- These plugins don't build/work on GHC92 yet
-  haskell-language-server
+  hls
```
jneira, line 45: ugh reverting asap

jneira:
> But I don't understand what you want to release. We know that statically linking glibc is not a good idea, so let's investigate a bit more before reverting to that.

I want to release the actual binary flavour as it will not break more use cases than the previous one. And i would prefer to make the release within January if possible to at least keep it bimonthly so if investigate and fix the problem will take more time we could do it after the release (and make another one before two months with the new binary)

> 
>     1. Could the problem have been caught earlier by running the test suite in the build.yml CI script?

Sure, i expressed my concerns about when it was added and opened https://github.com/haskell/haskell-language-server/issues/2617 to restore consistence in ci using the same build in all workflows (including benchmarks)

>     2. Are all ghc versions affected ?

@kokobd and me tested 8.10.7, the default in ghcup and more used one, but we could test other ones

>     3. Are the gitlab binaries affected?

i did not test them neither, there is no user report here or in the ghcup issue tracker afaik, maybe @hasufell knows

jneira:
> I'm not sure why the alpine binaries don't work, I will investigate. The alternative for now is to do what I suggested in #2431 and compile against glibc, but don't statically link it in. This can be done by using `executable-dynamic: False` instead of `executable-static: True`. We can then do what GHC does and distribute a different binary for each supported distro/glibc version.

That alternative could be tested easily changing the rest of workflows. But our distribution channels other than direct download (vsocde auto download and ghcup) doesn not take in account the linux distro/version so they should be changed (and that will take more time)

jneira:
Will try to adress #2617 in another pr to testthe alpine binary in our ci (but i guess it will break catastrophically the test suite)

wz1000:
> But our distribution channels other than direct download (vsocde auto download and ghcup) doesn not take in account the linux distro/version so they should be changed (and that will take more time)

GHCup can easily take distros into account when we add a metadata file for the next HLS release: https://github.com/haskell/ghcup-metadata/blob/master/ghcup-0.0.6.yaml#L2698

Instead of adding a single entry for `Linux_UnknownLinux`, multiple entries will need to be added for each distro version, as they already are for GHC.

jneira:
> GHCup can easily take distros into account when we add a metadata file for the next HLS release: https://github.com/haskell/ghcup-metadata/blob/master/ghcup-0.0.6.yaml#L2698

yeah, it already takes in account for ghc for necessity, but i will make the hls specific config more complex and the release harder as it is done semimanually
and changes in vscode will be everything but trivial

wz1000:
>  changes in vscode will be everything but trivial

I think we can re-use GHCup to download and install HLS in vscode, installing it(both GHCup and HLS) locally in the extension storage if it does not exist. I will try to write a patch to do this.

jneira:
> > changes in vscode will be everything but trivial
> 
> I think we can re-use GHCup to download and install HLS in vscode, installing it(both GHCup and HLS) locally in the extension storage if it does not exist. I will try to write a patch to do this.

We have an issue in the repo about: https://github.com/haskell/vscode-haskell/issues/483
It does not talk about download hls itself though. Add a new piece to the puzzle could bring tricky interactions though (and add another point of failure), we shoudl do it carefully.

wz1000:
OK, for this release what if we distribute generic "linux" binaries statically compiled with `glibc`(which will be used by the extension) and also distribute distro specific binaries that dynamically link against `glibc` which can be used by ghcup and also directly downloaded from the release page?

pepeiborra:
> Sure, i expressed my concerns about when it was added and opened #2617 to restore consistence in ci using the same build in all workflows (including benchmarks)

This is not enough, we need to run the testsuite in the `build.yml` workflow. But that can be added later, once we've sorted the alpine problem

pepeiborra:
> OK, for this release what if we distribute generic "linux" binaries statically linked with `glibc`(which will be used by the extension) and also distribute distro specific binaries that dynamically link against `glibc` which can be used by ghcup and also directly downloaded from the release page?

While I don't like the idea of having two flavours of binaries to support, it can be done if we add another user warning to highlight the statically linked glibc and document it, as in 

https://github.com/haskell/haskell-language-server/blob/4386396624180de112d04335a75334d04781110c/ghcide/src/Development/IDE/Core/Rules.hs#L828-L837

wz1000:
The `dlopen` failure "Dynamic loading not supported" happens because existing GHC alpine distributions are not statically linked, so GHC still tries to call `dlopen`. This should be fixed in future GHC releases by https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5147

pepeiborra:
> I'm not sure why the alpine binaries don't work, I will investigate. 

How does the GHC linker/loader detect that MUsl is being used, and is statically linked? It would need to in order to avoid calling dlopen.

pepeiborra (approved):
Approving since it seems clear that we need to go back to glibc until GHC is set up to detect a static musl binary.

jneira:
> The `dlopen` failure "Dynamic loading not supported" happens because existing GHC alpine distributions are not statically linked, so GHC still tries to call `dlopen`. This should be fixed in future GHC releases by https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5147

well ghc-9.2.1 was statically linked in alpine, no? and it produced problems in the ghc ci and bug like the described one in https://github.com/haskell/haskell-language-server/pull/2615#issuecomment-1018385710 (ghc ticket https://gitlab.haskell.org/ghc/ghc/-/issues/20977) so @hasufell changed the official ghc bindist for alpine+9.2.1 with a dynamic (or non static? i am a bit lost) in ghcup

jneira:
> > Sure, i expressed my concerns about when it was added and opened #2617 to restore consistence in ci using the same build in all workflows (including benchmarks)
> 
> This is not enough, we need to run the testsuite in the `build.yml` workflow. But that can be added later, once we've sorted the alpine problem

well given the actual master branch protection rules, there is no untested commit in master so it would serve to test a release made from a commit out of master, a very bad idea in its own. In exchange ci would be more complex, needing copy the entire test workflow or make use of [reused workflows](https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#reusing-workflows)

jneira:
Want to note that it is possible to trigger manual runs of the build workflow setting the alpine parameter to true

imagen

PR #2640, opened 2022-01-25
Use alpine in all sensible workflows

### LINKED ISSUE #2617: Restore consistency in ci using alpine in all workflows

jneira:
* Closes #2617
* Will test and bench the binaries to help trace and fix the issue with template haskell
* I've deleted the rerun workflow as it is not being used in practice


jneira:
better than i expected, the test suite failed in a th expression as expected:

```
 FAIL (1.68s)
        test/exe/Main.hs:3930:
        failed to find: `Identity` in hover message:
        
        '''haskell
        id :: forall a. a -> a
        '''
        
        *Defined in 'GHC.Base'*
         *(base-4.15.1.0)*
        
        
        [Documentation](file:///github/home/.ghcup/ghc/9.0.2/share/doc/ghc-9.0.2/html/libraries/base/GHC-Base.html#v:id)
        [Source](file:///github/home/.ghcup/ghc/9.0.2/share/doc/ghc-9.0.2/html/libraries/base/src/GHC.Base.html#id)
        
        * * *
        
        '''haskell
        _ :: Q Exp -> Q Exp
        '''
```

jneira:
For 9.2.1 the ghcide passed but a stack invocation failed (maybe stack does not work in alpine?):

```
Failed to get project GHC version:CradleError {cradleErrorDependencies = [], cradleErrorExitCode = ExitFailure 1, cradleErrorStderr = ["Error when calling stack setup --silent","",""]}
    stack with ghc 8.10.7:                    FAIL
      Exception: readCreateProcess: hls-wrapper "--project-ghc-version" (exit 1): failed
      Use -p '/stack with ghc 8.10.7/' to rerun this test only.
No 'hie.yaml' found. Try to discover the project type!
Failed to get project GHC version:CradleError {cradleErrorDependencies = [], cradleErrorExitCode = ExitFailure 1, cradleErrorStderr = ["Error when calling stack setup --silent","",""]}
    stack with ghc 8.8.4:                     FAIL
      Exception: readCreateProcess: hls-wrapper "--project-ghc-version" (exit 1): failed
      Use -p '/stack with ghc 8.8.4/' to rerun this test only.
```

Anton-Latukha:
Maybe install into Alpine `coreutils` to make it POSIX compliant.

Or simply lets remember that Alpine runs on `BusyBox` - that should be enough & leverage that specific to also test HLS running on `BusyBox`.

jneira:
This is not applicable to actual release scripts

Issue #2633, opened 2022-01-24
*** stack smashing detected ***: terminated


ilyakooo0:
### Your environment

Which OS do you use:
- Linux 6273dde0ec9a 5.10.91 NixOS x86_64 GNU/Linux
- VS Code 1.63.2
- Stack

The project has quite a bit of TH: autoexporter and embedding files in binaries.

### Steps to reproduce

I honestly don't know what causes this, but it happens every time I open my private project.

### Expected behaviour

It does not crash

### Actual behaviour

It crashes

### Include debug information

<details>
<summary>
Logs from the VS Code Haskell extension:
</summary>

```
*** stack smashing detected ***: terminated
haskell-language-server-wrapper: callProcess: /home/vscode/.local/bin/haskell-language-server "--lsp" (exit -6): failed
[Error - 8:00:21 AM] Connection to server got closed. Server will not be restarted.
[Error - 8:00:21 AM] Request textDocument/codeAction failed.
Error: Connection got disposed.
	at Object.dispose (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1010893)
	at Object.dispose (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1090976)
	at k.handleConnectionClosed (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1091189)
	at k.handleConnectionClosed (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1151535)
	at t (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1089278)
	at n.invoke (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1012506)
	at a.fire (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1013267)
	at Y (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1000151)
	at n.invoke (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1012506)
	at a.fire (/home/vscode/.vscode-server/extensions/haskell.haskell-1.8.0/dist/extension.js:2:1013267)
[...]

Issue #2596, opened 2022-01-17
Frequent random segfaults after Ubuntu 21.10 upgrade


jkaye2012:
### Description

HLS has been working flawlessly for me for months. Today, I upgraded my system from 21.04 to 21.10, and HLS is segfaulting. I'm not sure what could cause this, though I'm guessing it could be related to the glibc toolchain upgrade (to 2.34, https://discourse.ubuntu.com/t/impish-indri-release-notes/21951).

I have tried reinstalling HLS, nuking all stack caches, fully rebuilding, and even tried a few hand-written cradles to debug locally. I have been unable to identify anything that could be causing this.

### Your environment

Which OS do you use:
Ubuntu 21.10

Which LSP client (editor/plugin) do you use:
Emacs + lsp-mode

Describe your project (alternative: link to the project):
Stack project, basic auto-detected configuration (no explicit hie.yaml)

### Steps to reproduce
1. Install hls via ghcup
2. Run the language server

### Expected behaviour
The language server should analyze the code.

### Actual behaviour
The language server segfaults more than 90% of the time.

### Include debug information

Please let me know what other specific information I can provide to aid debugging.

Here are three subsequent runs (no changes made in the interim):

First run, partial completion:

```
jkaye@jkaye-linux ~/git/adaptive (master) $ haskell-language-server-8.10.7 src
haskell-language-server version: 1.5.1.0 (GHC: 8.10.7) (PATH: /home/jkaye/.ghcup/bin/haskell-language-server-8.10.7~1.5.1) (GIT hash: 745ef26f406dbdd5e4a538585f8519af9f1ccb09)
 ghcide setup tester in /home/jkaye/git/adaptive.
Report bugs at https://github.com/haskell/haskell-language-server/issues

Step 1/4: Finding files to test in /home/jkaye/git/adaptive
Found 16 files

Step 2/4: Looking for hie.yaml files that control setup
Found 1 cradle
  ()

Step 3/4: Initializing the IDE

Step 4/4: Type checking the files
2022-01-16 17:20:59.703536477 [ThreadId 94] INFO hls:	Consulting the cradle for "src/Adaptive/Utility/Error.hs"
2022-01-16 17:20:59.703669999 [ThreadId 94] WARNING hls:	No [cradle](https://github.com/mpickering/hie-bios#hie-bios) found for src/Adaptive/Utility/Error.hs.
 Proceeding with [implicit cradle](https://hackage.haskell.org/package/implicit-hie).
[...]

Issue #2340, opened 2021-11-10
Haskell extension hangs in VSCode


barnabasmolnar:
Linked issue on the IHP tracker: https://github.com/digitallyinduced/ihp/issues/1209

When using HLS as part of [IHP](https://github.com/digitallyinduced/ihp), either locally inside a WSL2 distro or via a Gitpod instance, the Haskell extension in VSCode hangs for a lot of time before becoming operational. For me, it's usually 4 minutes or so but a friend of mine on a weaker PC reported loading times up to 20-30 minutes.

I've provided some logs below.
Some of the actions that seem to have taken a lot of time are:

Wingman.getMetaprogramsAtSpan.TypeCheck (took 2m17s)
Wingman.codeLensProvider.GetBindings (took 3m34s)
RefineImports (took 3m34s)

When I'm using stack to manage my own Haskell projects, the extension seems to work fine and near instantaneously. Admittedly those projects are far smaller and have fewer files and everything... but ~4 minutes still seems very excessive to me.

Would be curious if anyone had similar issues and whether there might be a fix to speed things up.

<details>
<summary>Click to expand log</summary>

```log
Found "/home/roy/Developer/personal/bookmark-app/hie.yaml" for "/home/roy/Developer/personal/bookmark-app/a"
Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 1.4.0.0 x86_64 ghc-8.10.7
Current directory: /home/roy/Developer/personal/bookmark-app
Operating system: linux
Arguments: ["--lsp"]
Cradle directory: /home/roy/Developer/personal/bookmark-app
Cradle type: Bios

Tool versions found on the $PATH
cabal:		3.6.0.0
stack:		Not found
ghc:		8.10.7


Consulting the cradle to get project GHC version...
Project GHC version: 8.10.7
haskell-language-server exe candidates: ["haskell-language-server-8.10.7","haskell-language-server"]
Launching haskell-language-server exe at:/nix/store/kq65wynaxf8f0kmrx698npsrdg63g8fy-ghc-8.10.7-with-packages/bin/haskell-language-server-8.10.7
haskell-language-server version: 1.4.0.0 (GHC: 8.10.7) (PATH: /nix/store/xpwc04nd8g9vxjvjs3p4mwbi8vfqh00r-haskell-language-server-1.4.0.0/bin/haskell-language-server)
Starting (haskell-language-server)LSP server...
[...]

Issue #2054, opened 2021-07-31
Segmentation fault opening project with ghc-9.0.1 in debian 9


jneira:
* From discod user mallory
* os and ghc: linux kernel 4.4.0-21-generic and installed ghc with ghcup using the deb9 download
* Output of calling hls in the cli

```
~/.config/Code/User/globalStorage/haskell.haskell/haskell-language-server-1.3.0-linux-9.0.1 --logfile=./log.txt

haskell-language-server version: 1.3.0.0 (GHC: 9.0.1) (PATH: /home/foo/.config/Code/User/globalStorage/haskell.haskell/haskell-language-server-1.3.0-linux-9.0.1) (GIT hash: e7c5e90b6df5dff2760d76169eddaea3bdd6a831)
 ghcide setup tester in /home/foo/Documents/git/haskell/project.
Report bugs at https://github.com/haskell/haskell-language-server/issues

Step 1/4: Finding files to test in /home/foo/Documents/git/haskell/project
Found 2 files

Step 2/4: Looking for hie.yaml files that control setup
Found 1 cradle
  ()

Step 3/4: Initializing the IDE

Step 4/4: Type checking the files
Output from setting up the cradle Cradle {cradleRootDir = "/home/foo/Documents/git/haskell/project", cradleOptsProg = CradleAction: Cabal}
Output from setting up the cradle Cradle {cradleRootDir = "/home/foo/Documents/git/haskell/project", cradleOptsProg = CradleAction: Cabal}
> Build profile: -w ghc-9.0.1 -O1
> In order, the following will be built (use -v for more details):
>  - git-backup-0.1.0.0 (exe:git-backup) (first run)
> Preprocessing executable 'git-backup' for git-backup-0.1.0.0..
Output from setting up the cradle Cradle {cradleRootDir = "/home/foo/Documents/git/haskell/project", cradleOptsProg = CradleAction: Cabal}
Segmentation fault
```
* lsp output

```
2021-07-31 12:45:26.767378645 [ThreadId 24] INFO hls:    Consulting the cradle for "Setup.hs"
2021-07-31 12:45:26.767525619 [ThreadId 24] WARNING hls:    No [cradle](https://github.com/mpickering/hie-bios#hie-bios) found for Setup.hs.
 Proceeding with [implicit cradle](https://hackage.haskell.org/package/implicit-hie).
You should ignore this message, unless you see a 'Multi Cradle: No prefixes matched' error.
[...]

PR #2642, opened 2022-01-25
Prepare 1.6.0 release

### LINKED ISSUE #2424: January release 1.6.0.0
The machine never stops! Time to plan the new release.

* will include two new brand plugins
* and the continuous work of @pepeiborra improving performance
* [x] it will remove support for ghc-8.10.5
* [x] it should include the artifacts from gitlab, hopefully in a automated way (see https://github.com/haskell/ghcup-metadata/pull/11 and #2420
* [x] as noted by @hasufell the version of the gh release should match the hackage one (so 1.6.0.0)

as always if anyone feels strong enough to drive the release or collaborate with, help will be very welcomed :smiley_cat: 

-----------------------------

### prerelease sanity checks

- [x] trigger manually the hackage workflow *without* uploading the packages: https://github.com/jneira/haskell-language-server/actions/runs/1751828521
- [X] trigger manually the build workflow : https://github.com/jneira/haskell-language-server/actions/runs/1753316308
- [x] create a prerelease tag `1.6.0-check-gitlab` and push it to the [project repo in gitlab](https://gitlab.haskell.org/haskell/haskell-language-server) to check the build is fine: https://gitlab.haskell.org/haskell/haskell-language-server/-/pipelines/46483

### github release

- [x] generate the list of pull requests finished since the last release using the [haskell script](https://github.com/haskell/haskell-language-server/blob/master/GenChangelogs.hs) in the project root.
  Nix users should run command `gen-hls-changelogs` (a wrapper of the script) in nix-shell instead.
- [x] add that list to the actual [Changelog](https://github.com/haskell/haskell-language-server/blob/master/ChangeLog.md) with a description of the release.
- [x] bump up versions of changed packages. All are optional but [haskell-language-server itself](https://github.com/haskell/haskell-language-server/blob/master/haskell-language-server.cabal).
- [x] create the tag and make an initial prerelease to trigger the ci workflow (see details below)
- [x] contact ghcup team (#haskell-ghcup irc channel or via its [repo](https://github.com/haskell/ghcup-metadata)) to try to sync our release and its inclusion in ghcup
- [x] in the github release edit page, check the attached binaries and the release description (usually the changelog entry) and uncheck the prerelease box
- [ ] make public the release in the usual social channels (not required but useful to spread the word :slightly_smiling_face:):
  - [x] irc
  - [x] matrix
  - [x] twitter
  - [x] discord
  - [x] discourse
  - [x] reddit

### hackage release

- [x] bump up package versions following the [pvp specification](https://pvp.haskell.org/) if they are not already updated. You could use [policeman](https://github.com/kowainik/policeman) to help with this step.
- [x] create 1.6.0-hackage branch to trigger the hackage github workflow which will upload all changed packages to hackage as candidates
- [ ] check manually candidates in hackage
- [ ] publish them definitely

### ghcup release

- [x] push the release tag to the [haskell-language-server gitlab repo](https://gitlab.haskell.org/haskell/haskell-language-server) to trigger the build of ghcup specific artifacts
- [x] download specific artifacts only available in the gitlab build and compute their sha256sum
- [x] upload them to the github release and complete the SHA256SUMS file
- [x] change ghcup metadata to include the new release in https://github.com/haskell/ghcup-metadata
  -  example pull request [here](https://github.com/haskell/ghcup-metadata/pull/11)
  - pull request: https://github.com/haskell/ghcup-metadata/pull/19

jneira:
* Closes #2424

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2642"></a>


.github/workflows/hackage.yml
```diff
@@ -109,6 +107,7 @@ jobs:
         run: |
           cd $(ls -d ./incoming/${{ matrix.package }}-*)
           # For brittany
+          echo "allow-newer:" >> cabal.project
```
michaelpj, line 110: big hammer?

jneira: nope, below there are the concrete allow newer

jneira: nope, just below there are the concrete allow newers needed to make the solver happy, the section will look:

```
allow-newer:
 butcher:base, multistate:base, data-tree-print:base,
 stylish-haskell:Cabal,stylish-haskell:ghc-lib-parser,stylish-haskell:aeson
```

jneira: So brittany cant be built from hackage and ghc-9.0.2, will update the hls.cabal file if needed

michaelpj:
I guess we probably need to bump the versions of a bunch of plugins too?

jneira:
> I guess we probably need to bump the versions of a bunch of plugins too?

Yeah, i was just doing that. It is enforced by the hackage workflow: it fails if the local package being tested differs from the hackage version but has the same cabal version

ChangeLog.md
```diff
@@ -1,5 +1,285 @@
...
+- Finally hls [supports *ghc 9.2.1*](https://github.com/haskell/haskell-language-server/issues/2179)
+  - Including core features and [many plugins](https://haskell-language-server.readthedocs.io/en/latest/supported-versions.html#plugins-support-by-ghc-version))
+  - Thanks to a great collective effort coordinated by @pepeiborra and with the help of @wz100, @mpickering and @alanz among others
+- Hls now also [supports *ghc 9.0.2*](https://github.com/haskell/haskell-language-server/issues/297) with support for all plugins but the stylish-haskell formatter
+- And many many fixes and performance improvement
+
+### Deprecation notice for 1.6.0
+
+- As we noted in the previous release we have dropped support for ghc versions 8.10.5 and 8.8.3 in *this release*
+  - The advise is upgrade ghc to the last minor version: 8.8.4 or 8.10.7
```
jhrcek, line 20:
```suggestion
  - We recommend upgrading ghc to the last minor version: 8.8.4 or 8.10.7
```

ChangeLog.md
```diff
@@ -1,5 +1,285 @@
...
+  - You can read more about ghc deprecation policy and schedule [here](https://haskell-language-server.readthedocs.io/en/latest/supported-versions.html)
+- *After this release*:
+  - [We will remove all project stack.yaml's](https://github.com/haskell/haskell-language-server/issues/2533) but two: one for last lts and other for nightly. Temporary we could keep one more stack yaml when nightly upgrades the ghc version, to help in the transition
+  - [We will remove the install script](https://github.com/haskell/haskell-language-server/issues/2491) which lives [here](https://github.com/haskell/haskell-language-server/tree/master/install)
+    - The advise is install hls from source using `ghcup`. Download it and run `ghcup compile hls --help` to get moreinfo about.
```
jhrcek, line 25:
```suggestion
    - If you want to install hls from source we recommend using `ghcup`. Download it and run `ghcup compile hls --help` to get more info about.
```

docs/supported-versions.md
```diff
@@ -28,11 +28,14 @@ The policy for when we deprecate support for versions of GHC is given below. The
 
 Additionally, some plugins do not have support for some GHC versions, as shown in the following table.
 As such, the functionality provided by those plugins is not available in HLS when using a GHC version which they do not support.
+Sometimes a plugin can be supported in prebuilt binaries but no in a hls binary installed from hackage.
```
jhrcek, line 31:
```suggestion
Sometimes a plugin can be supported in prebuilt binaries but not in a hls binary installed from hackage.
```

jhrcek (commented):
Few stylistic suggestions..

ChangeLog.md
```diff
@@ -1,5 +1,290 @@
...
+
+Time for a new and exciting hls release:
+
+- It includes *three* brand new plugins:
+  - [Alternate number literals](https://haskell-language-server.readthedocs.io/en/latest/features.html#convert-numbers-to-alternative-formats) thanks to @drsooch
+  - [Qualify imported names](https://haskell-language-server.readthedocs.io/en/latest/features.html#qualify-imported-names) thanks to @eddiemundo
+  - New plugin to support [selection range](https://haskell-language-server.readthedocs.io/en/latest/features.html#selection-range) (aka double click text selection) thanks to @kokobd
+- Finally hls [supports *ghc 9.2.1*](https://github.com/haskell/haskell-language-server/issues/2179)
+  - Including core features and [many plugins](https://haskell-language-server.readthedocs.io/en/latest/supported-versions.html#plugins-support-by-ghc-version))
+  - Thanks to a great collective effort coordinated by @pepeiborra and with the help of @wz100, @mpickering and @alanz among others
```
pepeiborra, line 13:
```suggestion
  - Thanks to a great collective effort coordinated by @pepeiborra and with the help of @wz1000, @mpickering and @alanz among others
```

ChangeLog.md
```diff
@@ -1,5 +1,290 @@
...
+([#2482](https://github.com/haskell/haskell-language-server/pull/2482)) by @Ailrun
+- Delete some dead or deprecated settings
+([#2481](https://github.com/haskell/haskell-language-server/pull/2481)) by @michaelpj
+- Class plugin bump up
+([#2475](https://github.com/haskell/haskell-language-server/pull/2475)) by @Ailrun
+- Fix some pragma completion cases
+([#2474](https://github.com/haskell/haskell-language-server/pull/2474)) by @Ailrun
+- Minor org to contribution doc
+([#2472](https://github.com/haskell/haskell-language-server/pull/2472)) by @Anton-Latukha
+- Warn if TH and Mac and static binary
```
pepeiborra, line 203: Since we have rolled back the Musl change, we should tweak this warning to fire not only in Mac but also in Linux

pepeiborra (changes requested):

fendor (approved):
LGTM, thanks for taking care of the release!

ChangeLog.md
jneira: agree, will do in this pr

jneira:
@jhrcek thanks for corrections 🙂

ChangeLog.md
```diff
@@ -1,5 +1,290 @@
...
+
+- As we noted in the previous release we have dropped support for ghc versions 8.10.5 and 8.8.3 in *this release*
+  - We recommend upgrading ghc to the last minor version: 8.8.4 or 8.10.7
+  - You can read more about ghc deprecation policy and schedule [here](https://haskell-language-server.readthedocs.io/en/latest/supported-versions.html)
+- *After this release*:
+  - [We will remove all project stack.yaml's](https://github.com/haskell/haskell-language-server/issues/2533) but two: one for last lts and other for nightly. Temporary we could keep one more stack yaml when nightly upgrades the ghc version, to help in the transition
+  - [We will remove the install script](https://github.com/haskell/haskell-language-server/issues/2491) which lives [here](https://github.com/haskell/haskell-language-server/tree/master/install)
+    - If you want to install hls from source we recommend using `ghcup`. Download it and run `ghcup compile hls --help` to get more info about.
+    - Alternatively users could install hls manually following [these instructions](https://haskell-language-server.readthedocs.io/en/latest/installation.html#installation-from-source)
```
Ailrun, line 27: To make this more clear, I think it would be better to add a deprecation note at the beginning of [this doc](https://haskell-language-server.readthedocs.io/en/latest/installation.html#install-specific-ghc-version).
If a user scrolls down to this next section, he/she may get confused.

jneira: hmm, the plan would be replace installation instructions removing all references to install script. I thought do it just after the release but you are right if someone follows the link in the meanwhile it will see contradictory assertions

ChangeLog.md
jneira: done with 539f00, i've relaxed the warning as th evaluation works in some cases for linux

.github/actions/setup-build/action.yml
```diff
@@ -15,6 +15,11 @@ inputs:
     description: "To allow make unique the cache key"
     required: false
     default: ""
+  shorten-hls:
+    description: "Wether we have to apply shorten hls name"
```
eddiemundo, line 19:
```suggestion
    description: "Whether we have to shorten haskell language server to hls"
```

.github/workflows/hackage.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   check-and-upload-tarballs:
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: ${{ !contains(github.ref_name, 'check') && github.event.inputs.upload-candidates != 'true' }}
+      fail-fast: ${{ github.event.inputs.upload-candidates == 'true' }}
```
eddiemundo, line 23: Not sure what any of this does but just making sure `!=` is supposed to change to `==`?

eddiemundo (approved):

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -824,12 +824,12 @@ isHiFileStableRule = defineEarlyCutoff $ RuleNoDiagnostics $ \IsHiFileStable f -
...
 displayTHWarning
-  | isMac && not hostIsDynamic = do
+  | not isWindows && not hostIsDynamic = do
       LSP.sendNotification SWindowShowMessage $
         ShowMessageParams MtInfo $ T.unwords
-          [ "This HLS binary does not support Template Haskell."
+          [ "This HLS binary might not support Template Haskell."
           , "Follow the [instructions](" <> templateHaskellInstructions <> ")"
-          , "to build an HLS binary with support for Template Haskell."
+          , "to build an HLS binary with reliable support for Template Haskell."
```
pepeiborra, line 832: I'm not sure about the wording changes, it seems to me that it will generate more confusion than it's worth. 

Just looking at the last couple of days in the issue tracker to see how recurrent the problem is:

* https://github.com/haskell/haskell-language-server/issues/2596
* https://github.com/haskell/haskell-language-server/issues/2054
* https://github.com/haskell/haskell-language-server/issues/2340
* https://github.com/haskell/haskell-language-server/issues/1497
* https://github.com/haskell/haskell-language-server/issues/2633

ChangeLog.md
jneira: i started to write the new install doc but i dont get it right easily so i think is better to remove the reference for now

ghcide/src/Development/IDE/Core/Rules.hs
jneira: Well, the fact it *works* for some cases.
Not sure in how many cases is working cause happy users dont open issues but maybe more than we expected.
We are gonna lie to those users and make them install from source when it would not be necessary.

Maybe we can rewrite it to be precise, recommending strongly building from source without ignoring it could work in some environments

.github/workflows/hackage.yml
jneira: Thanks for noting it. Yeah it was wrong and this fixes it.

drsooch (approved):

kokobd (approved):

ChangeLog.md
```diff
@@ -1,5 +1,289 @@
...
+Time for a new and exciting hls release:
+
+- It includes *three* brand new plugins:
+  - [Alternate number literals](https://haskell-language-server.readthedocs.io/en/latest/features.html#convert-numbers-to-alternative-formats) thanks to @drsooch
+  - [Qualify imported names](https://haskell-language-server.readthedocs.io/en/latest/features.html#qualify-imported-names) thanks to @eddiemundo
+  - New plugin to support [selection range](https://haskell-language-server.readthedocs.io/en/latest/features.html#selection-range) (aka double click text selection) thanks to @kokobd
+- Finally hls [supports *ghc 9.2.1*](https://github.com/haskell/haskell-language-server/issues/2179)
+  - Including core features and [many plugins](https://haskell-language-server.readthedocs.io/en/latest/supported-versions.html#plugins-support-by-ghc-version))
+  - Thanks to a great collective effort coordinated by @pepeiborra and with the help of @wz1000, @mpickering and @alanz among others
+- Hls now also [supports *ghc 9.0.2*](https://github.com/haskell/haskell-language-server/issues/297) with support for all plugins but the stylish-haskell formatter
```
Ailrun, line 14: As stylish-haskell remains as a "special" case, maybe it is worth noting that the support is blocked by an upstream issue? Or even better, with a link to the issue: https://github.com/haskell/stylish-haskell/issues/378

Ailrun: Also,
```suggestion
- Hls now also [supports *ghc 9.0.2*](https://github.com/haskell/haskell-language-server/issues/297) with all plugins but the stylish-haskell formatter
```
the repetition of "support" sounds a bit redundant.

Ailrun (approved):
Thank you, always, for all your effort to cut releases. Here is one minor comment

ghcide/src/Development/IDE/Core/Rules.hs
jneira: Ok i changed the first verb from `might`to `could` to denote the binary *likely* will not work with th:
```
"This HLS binary could not support Template Haskell."
          , "Follow this [instructions](" <> templateHaskellInstructions <> ")"
          , "to build an HLS binary with reliable support for Template Haskell."
```

ghcide/ghcide.cabal
```diff
@@ -60,7 +60,7 @@ library
         haddock-library >= 1.8 && < 1.11,
         hashable,
         hie-compat ^>= 0.2.0.0,
-        hls-plugin-api ^>= 1.2.0.2,
+        hls-plugin-api ^>= 1.2.0.2 || ^>= 1.3,
```
michaelpj, line 63: Any reason not to drop the old constraints in the places where you've done this?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -824,11 +824,11 @@ isHiFileStableRule = defineEarlyCutoff $ RuleNoDiagnostics $ \IsHiFileStable f -
...
 displayTHWarning :: LspT c IO ()
 displayTHWarning
-  | isMac && not hostIsDynamic = do
+  | not isWindows && not hostIsDynamic = do
       LSP.sendNotification SWindowShowMessage $
         ShowMessageParams MtInfo $ T.unwords
-          [ "This HLS binary does not support Template Haskell."
-          , "Follow the [instructions](" <> templateHaskellInstructions <> ")"
+          [ "This HLS binary could not support Template Haskell."
+          , "Follow this [instructions](" <> templateHaskellInstructions <> ")"
```
michaelpj, line 831: I think the old version read better to me.

michaelpj (approved):

.github/actions/setup-build/action.yml
```diff
@@ -65,10 +70,11 @@ runs:
       shell: bash
 
     # Shorten binary names as a workaround for filepath length limits in Windows,
```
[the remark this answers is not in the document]
jneira: And would simplify ci, and there would not be problems with long paths in windows, the name would be consistent with ghcup, etc, etc
I knoe it is not constructive but i cant resist to remember i suggested hls as executable name when all this started (like the executable for haskell-ide-engine was hie)
But now it would break the actual usage and who knows how many scripts and configurations out there so maybe it is too late.
If we decide to do the change it should have to be done very very carefully and gradually, supporting both names in the meanwhile 😓

docs/features.md
```diff
@@ -192,61 +218,71 @@ Converts numeric literals to different formats.
...
 ### Evaluation code snippets in comments
+
 Provided by: `hls-eval-plugin`
 
 Evaluates code blocks in comments with a click. [Tutorial](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/README.md).
 
 ![Eval Demo](https://raw.githubusercontent.com/haskell/haskell-language-server/master/plugins/hls-eval-plugin/demo.gif)
 
-### Make import lists fully explicit
+### Make import lists fully explicit code lens
```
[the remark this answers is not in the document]
jneira: yeah, i apply several changes suggested by markdown-lint here and duplicate headers is one of its warnings

.github/actions/setup-build/action.yml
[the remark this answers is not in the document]
drsooch: A simple symlink should resolve most of the script/config issues no?

ghcide/src/Development/IDE/Core/Rules.hs
pepeiborra: I still think that `could` is still confusing, to be honest. 

Taking a step back, your main issue here is that you don't want users to build HLS from source unnecessarily. Why is that so bad? One obvious problem is that `cabal install` will place the binary under `~/.cabal/bin` and will not annotate it with the ghc version number, which means that the vscode extension will not find it. 

So the problem is that instructing users to build from source leads to a bad experience. Should we focus on addressing that, rather than on this wording?

ghcide/src/Development/IDE/Core/Rules.hs
jneira: the does -> could part? or the -> this? the first one is a semantic change cause th could work for some cases

ghcide/ghcide.cabal
jneira: No really, i changed all packages to use the same hls-graph, hls-plugin-api and ghcide bounds in all of them and make updates easier. The bounds includes the current version and the previous one but i did not any test if the prev version works or does not (or if the package works with older versions)

ghcide/src/Development/IDE/Core/Rules.hs
michaelpj: Then I'd use "might" there

ghcide/ghcide.cabal
michaelpj: AFAICT, the only thing we can make promises about is "the current state of the repo works", so I think including the previous versions isn't reliable (and does it benefit anyone?).

jneira: agree will keep only the lastest one

ghcide/src/Development/IDE/Core/Rules.hs
jneira: I used might but I read somewhere that might denotes less probability and after @pepeiborra concern I wanted to make stronger the fact it could not work

.github/workflows/hackage.yml
jneira: No, it is a new cabal.project created on the fly. I think i broke it in a previous pr and i detected the error here

.github/actions/setup-build/action.yml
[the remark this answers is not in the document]
drsooch: Yeah I don't know all the gory details with windows as I don't use it myself. Then the downstream itself I'm not sure how ghcup (?) would handle the symlinks and if that itself breaks anything? Seems like something to push down the road.

ghcide/src/Development/IDE/Core/Rules.hs
pepeiborra: Since statically linking glibc is unsupported, I don't know a good way of making the warning more precise.

.github/actions/setup-build/action.yml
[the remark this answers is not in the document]
jneira: > Then the downstream itself I'm not sure how ghcup (?) would handle the symlinks and if that itself breaks anything?

Using tiny executable wrappers called shims

ghcide/src/Development/IDE/Core/Rules.hs
jneira: I am afraid my knowledge of those low level details is very limited so i cant help much.
If we cant be precise with the warning i still think we should be cautious with the message.
Users with a working th installation watching that message will lose a little bit of confidence in our error handling.

But well, as you still think the original wording is better will restore it.

ghcide/src/Development/IDE/Core/Rules.hs
jneira: original version restored

jneira:
@pepeiborra would be this ready to go?

pepeiborra (approved):

pepeiborra:
Thanks @jneira for cutting a new release!

jneira:
Hi all, i firmly think that the next release *must* be driven by other maintainer. It will be good for the project as other maintainers than me should be able to do the relase. I hope the documentation about would be complete enough to do it.
Of cource will help in the process as much as i can.
Thanks!

PR #2644, opened 2022-01-27
Wingman: Fix #1879

### LINKED ISSUE #1879: Wingman produces silly arbitrary instance for single-constructor types
```haskell

data CourseData = CD
  { cd_dur :: Dur
  , cd_end :: Time
  }

instance Arbitrary CourseData where
  arbitrary
    = let terminal = [(CD <$> arbitrary) <*> arbitrary]
      in
        sized
          $ (\ n
               -> case n <= 1 of
                    True -> oneof terminal
                    False -> oneof $ ([] <> terminal))
```

MorrowM:
Adds a special case in order to fix #1879

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2644"></a>

### DIFF
--- a/plugins/hls-tactics-plugin/src/Wingman/KnownStrategies/QuickCheck.hs
+++ b/plugins/hls-tactics-plugin/src/Wingman/KnownStrategies/QuickCheck.hs
@@ -43,8 +43,9 @@ deriveArbitrary = do
               mempty
               mempty
               mempty
-          $ noLoc $
-              let' [valBind (fromString "terminal") $ list $ fmap genExpr terminal] $
+          $ noLoc $ case terminal of
+            [onlyCon] -> genExpr onlyCon -- See #1879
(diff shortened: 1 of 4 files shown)


isovector:
Thanks for tackling this! Mind adding a quick test?

It's quick: add a new file with a hole, like this one:

https://github.com/haskell/haskell-language-server/blob/0905e682de8cc3e2bfe509313924cff0527b6791/plugins/hls-tactics-plugin/test/golden/GoldenArbitrary.hs

and then add the line, column and filename of the hole to this file:

https://github.com/haskell/haskell-language-server/blob/0905e682de8cc3e2bfe509313924cff0527b6791/plugins/hls-tactics-plugin/test/CodeAction/AutoSpec.hs#L74

running the tests will then generate the expected test output, and is good to check in.

Thanks so much!

isovector (approved):
LGTM, thanks so much!

PR #2639, opened 2022-01-25
Speed up fuzzy search

Bodigrim:
My intuition is that this approach should be faster, but I do not have data to back such claim. @pepeiborra any chance to test it with a huge list of completions?

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2639"></a>

PR #2651, opened 2022-01-29
Get rid of some head usages

jhrcek:
Perfect is the enemy of good.
Instead of replacing all partial function usages at once, I'll try to do it in series of smaller PRs.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2651"></a>


test/functional/Completion.hs
```diff
@@ -327,11 +330,21 @@ contextTests = testGroup "contexts" [
...
 
 shouldNotContainCompl :: [CompletionItem] -> T.Text -> Assertion
-compls `shouldNotContainCompl` x =
-    all ((/= x) . (^. label)) compls
-    @? "Should not contain completion: " ++ show x
+compls `shouldNotContainCompl` lbl =
+    all ((/= lbl) . (^. label)) compls
+    @? "Should not contain completion: " ++ show lbl
+
+getCopletionByLabel :: T.Text -> [CompletionItem] -> Session CompletionItem
```
guibou, line 342: Is the `Copletion` a typo? `getCompletionByLabel`

plugins/hls-pragmas-plugin/test/Main.hs
```diff
@@ -74,7 +75,10 @@ codeActionTest testComment fp actions =
     _ <- waitForDiagnosticsFrom doc
     cas <- map fromAction <$> getAllCodeActions doc
     mapM_ (\(action, contains) -> go action contains cas) actions
-    executeCodeAction $ head cas
+    action <- case cas of
```
guibou, line 78: Instead of explicit `case` everywhere, wouldn't you instead write a function, such as:

```haskell
unsafeHeadWithNoteM :: (HasCallStack, MonadIO m) => String -> [a] -> m a
unsafeHeadWithNoteM message l = case l of
   [] -> liftIO $ assertFailure message
   (x:_) -> pure a
```

Or something similar?

test/functional/Completion.hs
jhrcek: Good spot, fixed.

plugins/hls-pragmas-plugin/test/Main.hs
jhrcek: Similar function could indeed be used, but I'm not sure where to put it, in order for it to be usable by different test suites. In this particular test suite it could only be used once as the other case requires failure also on lists of length 2+

guibou: I actually did not realized that it was code in the test suite. I don't really care about "code quality", "deduplication" and "exception safety" in test suite, so you can consider my suggestion as not important.

I'll let a project owner continue the discussion.

jhrcek: No problem. I figured eliminating some partial function usages in tests can still be useful, as in this recent example:
https://github.com/haskell/haskell-language-server/pull/2639#issuecomment-1022727133

guibou: For tests, I'm wondering if running tests with `-prof` enabled may not gives free stacktrace by default. It will be a bit slower, but with nicer error messages.

Note: at work, we use `base-noprelude` + a custom `Prelude.hs` which just reexport `Prelude` minors `head` and a few others, just rewrite `head` with `HasCallstack` constraint. We changed nothing to the codebase and immediatly got better stacktrace. This can be another option.

jhrcek: We're also using base-noprelude at work, unfortunately that package hasn't been maintained for a long time (we had to fork it to be able to upgrade to ghc 8.10.7). Moreover switching to custom prelude approach would be much larger and more controversial change than simpler "let's just not use partial functions".

guibou: I do agree. Good luck to progress on that patch, that's a cool merge request.

michaelpj: I'm pro removing partial functions in tests still, but I don't think it's worth exerting too much effort on making it super tidy. We do have `hls-test-utils`, though.

plugins/hls-tactics-plugin/src/Wingman/Judgements.hs
```diff
@@ -34,7 +34,7 @@ buildHypothesis
   where
     go (occName -> occ, t)
       | Just ty <- t
-      , isAlpha . head . occNameString $ occ = Just $ HyInfo occ UserPrv $ CType ty
+      , maybe False isAlpha . listToMaybe . occNameString $ occ = Just $ HyInfo occ UserPrv $ CType ty
```
michaelpj, line 37: this is fine, but you might also like this version. Pattern guards are cool
```suggestion
      , (h:_) <- occNameString $ occ
      isAlpha h = Just $ HyInfo occ UserPrv $ CType ty
```

michaelpj (approved):

plugins/hls-pragmas-plugin/test/Main.hs
jhrcek: Indeed, hls-test-utils seems like a good place for that utility. I moved it there and reused in few more places.

plugins/hls-tactics-plugin/src/Wingman/Judgements.hs
jhrcek: Good point, I'm not used to using pattern guards so it didn't occur to me. Applied.

pepeiborra (approved):
Looks good to me. Thanks!

Ailrun (approved):
Thank you, looks good to me.

jneira (approved):
@jhrcek thank you, this would make a little bit easier add hlint config to make impossible introduce head again (#2634)

PR #2652, opened 2022-01-29
Move hackage back to flake.nix

guibou:
Apparently `flake-compat` does not work correctly with local flake.

Reported by @teto and @smunix in #2621.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2652"></a>


smunix:
Thanks @guibou! The issue is fixed indeed

jneira (approved):
lgtm, thank you, and there is a report confirming the fix so merging

PR #2655, opened 2022-01-29
Nix flake fix dev shells

guibou:
This is a work in progress.

This clean and fixs the `nix develop` shells for GHC 9.2 and GHC 9.0.

See individual commits for details, but in summary:

- `nix develop .#haskell-language-server-ghc921-dev` now works.
- When entering the shell, it creates an alias for `cabal` which points to the correct `cabal-xxx.project`

I'm planning on testing on GHC 9.0 soon too (but right now it is building the world of dependencies).

Future commits to this MR will (hopefully) includes:

- Support for GHC 9.0
- I'd like to restore the CI steps related to nix/flake, as an optionnal step which can be triggered on demand by developers. It will also populate the cachix cache.
- Additional cleanups. I'm trying to remove all the explicit "hacks" `if version == ...`

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2655"></a>

### DIFF
--- a/configuration-ghc-921.nix
+++ b/configuration-ghc-921.nix
@@ -5,6 +5,11 @@ let
   disabledPlugins = [
     "hls-brittany-plugin"
     "hls-stylish-haskell-plugin"
+    "hls-hlint-plugin"
+    "hls-haddock-comments-plugin"
+    "hls-alternate-number-format-plugin"
+    "hls-eval-plugin"
+    "hls-tactics-plugin"
   ];
 
   hpkgsOverride = hself: hsuper:
@@ -22,11 +27,12 @@ let
         hself.callCabal2nixWithOptions "haskell-language-server" ./.
         (pkgs.lib.concatStringsSep " " [
           "-f-brittany"
-          "-f-stylishhaskell"
+          "-f-stylishHaskell"
           "-f-hlint"
           "-f-haddockComments"
           "-f-alternateNumberFormat"
           "-f-eval"
+          "-f-tactics"
         ]) { };
 
       # YOLO
--- a/configuration-ghc-921.nix
+++ b/configuration-ghc-921.nix
@@ -19,6 +19,8 @@ let
   hpkgsOverride = hself: hsuper:
     with pkgs.haskell.lib;
     {
+      hlsDisabledPlugins = disabledPlugins;
+
       fourmolu = hself.callCabal2nix "fourmolu" inputs.fourmolu {};
       primitive-extras = hself.primitive-extras_0_10_1_2;
       ghc-exactprint = hself.callCabal2nix "ghc-exactprint" inputs.ghc-exactprint {};
--- a/configuration-ghc-921.nix
+++ b/configuration-ghc-921.nix
@@ -24,7 +24,10 @@ let
       ghc-exactprint = hself.callCabal2nix "ghc-exactprint" inputs.ghc-exactprint {};
       constraints-extras = hself.callCabal2nix "constraints-extras" inputs.constraints-extras {};
       retrie = hself.callCabal2nix "retrie" inputs.retrie {};
+
+      # Hlint is still broken
       hlint = doJailbreak (hself.callCabal2nix "hlint" inputs.hlint {});
+      hiedb = hself.hiedb_0_4_1_0;
 
       # Re-generate HLS drv excluding some plugins
       haskell-language-server =
--- a/configuration-ghc-921.nix
+++ b/configuration-ghc-921.nix
@@ -7,8 +7,6 @@ let
     "hls-stylish-haskell-plugin"
     "hls-hlint-plugin"
     "hls-haddock-comments-plugin"
-    "hls-alternate-number-format-plugin"
-    "hls-eval-plugin"
     "hls-tactics-plugin"
     # That one is not technically a plugin, but by putting it in this list, we
     # get it removed from the top level list of requirement and it is not pull
@@ -39,8 +37,6 @@ let
           "-f-stylishHaskell"
           "-f-hlint"
           "-f-haddockComments"
-          "-f-alternateNumberFormat"
-          "-f-eval"
           "-f-tactics"
         ]) { };
 
--- a/configuration-ghc-921.nix
+++ b/configuration-ghc-921.nix
@@ -10,6 +10,10 @@ let
     "hls-alternate-number-format-plugin"
     "hls-eval-plugin"
     "hls-tactics-plugin"
+    # That one is not technically a plugin, but by putting it in this list, we
+    # get it removed from the top level list of requirement and it is not pull
+    # in the nix shell.
+    "shake-bench"
   ];
 
   hpkgsOverride = hself: hsuper:
--- a/flake.nix
+++ b/flake.nix
@@ -315,10 +315,9 @@
         # Copied from https://github.com/NixOS/nixpkgs/blob/210784b7c8f3d926b7db73bdad085f4dc5d79418/pkgs/development/tools/haskell/haskell-language-server/withWrapper.nix#L16
         mkExe = hpkgs:
           with pkgs.haskell.lib;
-          (justStaticExecutables (overrideCabal hpkgs.haskell-language-server
+          (enableSharedExecutables (overrideCabal hpkgs.haskell-language-server
             (_: {
               postInstall = ''
-                remove-references-to -t ${hpkgs.ghc} $out/bin/haskell-language-server
                 remove-references-to -t ${hpkgs.shake.data} $out/bin/haskell-language-server
                 remove-references-to -t ${hpkgs.js-jquery.data} $out/bin/haskell-language-server
                 remove-references-to -t ${hpkgs.js-dgtable.data} $out/bin/haskell-language-server
--- a/flake.nix
+++ b/flake.nix
@@ -251,22 +251,32 @@
                   removeAttrs hlsSources ghc921Config.disabledPlugins
                 else
                   hlsSources));
-            buildInputs = [ gmp zlib ncurses capstone tracy (gen-hls-changelogs hpkgs) pythonWithPackages ]
-              ++ (with hpkgs; [
-                cabal-install
-                hie-bios
+            # For theses tools packages, we use ghcDefault
+            # This removes a rebuild with a different GHC version
+            # Theses programs are tools, used as binary, independently of the
+            # version of GHC.
+            # The drawback of this approach is that our shell may pull two GHC
+            # version in scope (the `ghcDefault` one, and the one defined in
+            # `hpkgs`.)
+            # The advantage is that we won't have to rebuild theses tools (and
+            # dependencies) with a recent GHC which may not be supported by
+            # them.
+            buildInputs = [ gmp zlib ncurses capstone tracy (gen-hls-changelogs ghcDefault) pythonWithPackages ]
+              ++ [
+                pkgs.cabal-install
+                ghcDefault.hie-bios
                 hlint
                 # ormolu
                 # stylish-haskell
-                opentelemetry-extra
-              ]);
+                ghcDefault.opentelemetry-extra
+              ];
 
             src = null;
             shellHook = ''
               export LD_LIBRARY_PATH=${gmp}/lib:${zlib}/lib:${ncurses}/lib:${capstone}/lib
               export DYLD_LIBRARY_PATH=${gmp}/lib:${zlib}/lib:${ncurses}/lib:${capstone}/lib
               export PATH=$PATH:$HOME/.local/bin
-              ${if hpkgs.ghc.version != "9.0.1" then (pre-commit-check hpkgs).shellHook else ""}
+              ${(pre-commit-check ghcDefault).shellHook}
             '';
           };
         # Create a hls executable
--- a/flake.nix
+++ b/flake.nix
@@ -245,12 +245,8 @@
             packages = p:
               with builtins;
               map (name: p.${name}) (attrNames
-                (if hpkgs.ghc.version == "9.0.1" then
-                  removeAttrs hlsSources ghc901Config.disabledPlugins
-                else if hpkgs.ghc.version == "9.2.1" then
-                  removeAttrs hlsSources ghc921Config.disabledPlugins
-                else
-                  hlsSources));
+              # Disable dependencies should not be part of the shell.
+              (removeAttrs hlsSources (hpkgs.hlsDisabledPlugins or [])));
             # For theses tools packages, we use ghcDefault
             # This removes a rebuild with a different GHC version
             # Theses programs are tools, used as binary, independently of the
--- a/flake.nix
+++ b/flake.nix
@@ -365,7 +365,7 @@
           all-haskell-language-server = linkFarmFromDrvs "all-haskell-language-server" (lib.unique (builtins.attrValues allPackages));
 
           # Same for all shells 
-          all-dev-shells = linkFarmFromDrvs "all-dev-shells" (builtins.map (shell: shell.inputDerivation) (lib.unique (builtins.attrValues devShells)));
+          all-nix-dev-shells = linkFarmFromDrvs "all-dev-shells" (builtins.map (shell: shell.inputDerivation) (lib.unique (builtins.attrValues [nixDevShells.haskell-language-server-dev-nix])));
 
           all-simple-dev-shells = linkFarmFromDrvs "all-dev-shells" (builtins.map (shell: shell.inputDerivation) (lib.unique (builtins.attrValues simpleDevShells)));
           docs = docs;
--- a/flake.nix
+++ b/flake.nix
@@ -236,46 +236,78 @@
           dontInstall = true;
         };
 
+        mkDevShell = hpkgs: cabalProject: with pkgs; mkShell {
+          # For theses tools packages, we use ghcDefault
+          # This removes a rebuild with a different GHC version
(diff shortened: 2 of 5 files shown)


michaelpj:
> When entering the shell, it creates an alias for cabal which points to the correct cabal-xxx.project

I'm a little wary of this. It's nice if you know it's there, but it's hidden magic and adds a source of confusion ("I can't cabal build with 9.0", "I can".... "oh, one of us was using the magic cabal wrapper").

guibou:
> > When entering the shell, it creates an alias for cabal which points to the correct cabal-xxx.project
> 
> I'm a little wary of this. It's nice if you know it's there, but it's hidden magic and adds a source of confusion ("I can't cabal build with 9.0", "I can".... "oh, one of us was using the magic cabal wrapper").

I agree.

What about if the alias is called `cabal_ghc92`, so you must explicitly use `cabal_ghc92 build`, and a message is printed when entering the shell, such as:

```
Cabal won't be able to build your project without using the project file `cabal-xxx.project`. An alias `cabal_ghcxxx` was automatically created for you and uses this cabal project. Use it like:

cabal_ghcxxx build
```

michaelpj:
That sounds fine to me, although at that point I might as well type `--project-file=...` myself :) The warning is nice though!

guibou:
A few changes:

- As suggested by @michaelpj, the alias when entering the nix-shell is not `cabal`, but `cabal_project`. A message warns the user about the need to use a special `cabal.project` file, as well about the existence of the alias.
- I simplified the shell. Previously, it was trying to build the haskell dependencies using `nix`. However, this was problematic  because:
  
  * Once in the shell, `cabal build` may decide to use different version, and hence was rebuilding. So we were building dependencies twices.
  * Theses dependencies may fail to build, and the nix-shell may not start.
  * It clutters the nix setup, because we had to detect the Haskell dependencies used by hls, and manually remove the broken plugins. The new approach is simpler, it just starts a shell with `ghc`, `cabal` and a few C dependencies (zlib, ...).

flake.nix
```diff
@@ -236,39 +236,60 @@
...
+          # version in scope (the default one, and the one defined in
+          # `hpkgs`.)
+          # The advantage is that we won't have to rebuild theses tools (and
+          # dependencies) with a recent GHC which may not be supported by
+          # them.
+          buildInputs = [
+            # our compiling toolchain
+            hpkgs.ghc
+            pkgs.cabal-install
+            # @guibou: I'm not sure hie-bios is needed
```
michaelpj, line 254: I think it isn't.

flake.nix
```diff
@@ -236,39 +236,60 @@
...
+          # For theses tools packages, we use ghcDefault
+          # This removes a rebuild with a different GHC version
+          # Theses programs are tools, used as binary, independently of the
+          # version of GHC.
+          # The drawback of this approach is that our shell may pull two GHC
+          # version in scope (the default one, and the one defined in
+          # `hpkgs`.)
+          # The advantage is that we won't have to rebuild theses tools (and
+          # dependencies) with a recent GHC which may not be supported by
+          # them.
```
michaelpj, line 249: :+1:

michaelpj (approved):
Personally, I'm in favour of what you do here, just having the dev shell contain the system dependencies and letting cabal build the rest. In fact, that's what I do already (I don't use the existing dev shells for that reason!). However, possibly some people (@berberman ?) like the way it's done currently.

flake.nix
```diff
@@ -236,39 +236,60 @@
...
-            doBenchmark = true;
-            packages = p:
-              with builtins;
-              map (name: p.${name}) (attrNames
-                (if hpkgs.ghc.version == "9.0.1" then
-                  removeAttrs hlsSources ghc901Config.disabledPlugins
-                else if hpkgs.ghc.version == "9.2.1" then
-                  removeAttrs hlsSources ghc921Config.disabledPlugins
-                else
-                  hlsSources));
```
berberman, line 253: I think one basic motivation of using nix is that we don't want to wast time building these Haskell dependencies locally. Yes, the build failures happen, and it took a lot of efforts to maintain nix files. The initial idea here was to rely on nix as much as possible, so that we could build only few packages by cabal. Instead of completely cutting down haskell dependencies in nix shell, is there any chance to find a compromise that tries best to take the advantages of nix, but being conservative so that we can never fail to enter the shell?

guibou: The problem is that unless we perfectly select the packages version from nix so they match what `cabal` will select when creating its build plan, we cannot ensure that cabal won't try to rebuild dependencies inside the shell.

Unless we know how to force `cabal` to use what's available in the package database provided by nix. I don't know how to do that.

Another solution would be to use https://github.com/input-output-hk/haskell.nix and let it parses the cabal build plan so we can be sure that we'll fetch the dependencies which will be used by cabal. This would be the perfect solution, because it will also mean that `nix build` will uses the same dependencies as the one used by `cabal build`.

michaelpj: I think that given how frequently the HLS deps change (or even just the index-state), trying to keep up with things manually is doomed to failure. Something like `haskell.nix` that does it automatically can work (I use it at work), but I think is too complex for this project.

Letting cabal build the Haskell deps is dumb and a little slow, but requires almost no maintenance and tends to work.

guibou: > Unless we know how to force cabal to use what's available in the package database provided by nix. I don't know how to do that.

Actually, @berberman, you may know more than me about that. For some reasons, `cabal build` in `nix-shell` (or `nix develop`) won't fetch/build from hackage. However, it does for `nix develop .#haskell-language-server-ghcXXX-dev` for ghc 9.0 and 9.2. For example, it fetch and build `lsp-1.4.0.0` from hackage, even if the same package with same version is available inside `ghc-pkg list`. I may have missed something here, but how do I force `cabal` to use the package that is already available in my "local nix" package database?

guibou: One option may be to keep both options available, so users can decide to either pick the basic shell or use the advanced one which pre-populate the nix cache with all the Haskell dependencies.

At least, if the dependencies changes, dev will still be able to use the basic shell? And it will test our motivation to maintain the precise list of dependencies in nix. (I'll try to do it, contrary to what this MR may let you think, I'm highly in favor of doing as much as possible with nix, just that I exhausted the amount of time available to fix this, and I instead decided to try a simpler approach).

@berberman, @michaelpj, do you agree if I keep both approach? However I'd like to set the "basic" shell as the default one, so we get some guarantee that it will always be usable?

jneira: Sorry know nothing about nix but if you keep two version i think a good way to keep them up to date would be add both to ci

michaelpj: The one with all the deps has almost always been broken, if that's a useful data point.

I think in principle you can force cabal to use only what's in the package-db by setting `CABAL_CONFIG` to something that doesn't have any remote repositories. But then you're really out of luck if it doesn't work.

michaelpj: I'm okay with having both, I just don't want to maintain the complicated one :p

guibou: Super.

jneira:
the nix build in master is failing since https://github.com/haskell/haskell-language-server/actions/runs/1778521865 (the merge of #2668):

```
error: builder for '/nix/store/5hz103hms77w33chs4170kws3h5mxp63-haskell-language-server-1.6.1.0.drv' failed with exit code 1;
       last 10 log lines:
       > post-installation fixup
       > shrinking RPATHs of ELF executables and libraries in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0
       > shrinking /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server-wrapper
       > shrinking /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server
       > shrinking /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/lib/ghc-8.10.7/x86_64-linux-ghc-8.10.7/libHShaskell-language-server-1.6.1.0-KvnnT3SgOLdLTY2wE4rDC1-ghc8.10.7.so
       > strip is /nix/store/xiq6j4jsyj351p8q3yw9cg1hdqp9m685-gcc-wrapper-10.3.0/bin/strip
       > stripping (with command strip and flags -S) in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/lib  /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin
       > patching script interpreter paths in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0
       > checking for references to /build/ in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0...
       > RPATH of binary /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server contains a forbidden reference to /build/
       For full logs, run 'nix log /nix/store/5hz103hms77w33chs4170kws3h5mxp63-haskell-language-server-1.6.1.0.drv'.
Error: Process completed with exit code 1.
```

flake.nix
guibou: I've pushed a change where both type of shell exists.

guibou:
I've pushed a change with both type of shell (i.e. simple deps where cabal builds the dependencies, or complete deps).

I've also changed the CI so it builds HLS and shells for all flavors of them.

guibou:
> the nix build in master is failing since https://github.com/haskell/haskell-language-server/actions/runs/1778521865 (the merge of #2668):
> 
> ```
> error: builder for '/nix/store/5hz103hms77w33chs4170kws3h5mxp63-haskell-language-server-1.6.1.0.drv' failed with exit code 1;
>        last 10 log lines:
>        > post-installation fixup
>        > shrinking RPATHs of ELF executables and libraries in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0
>        > shrinking /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server-wrapper
>        > shrinking /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server
>        > shrinking /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/lib/ghc-8.10.7/x86_64-linux-ghc-8.10.7/libHShaskell-language-server-1.6.1.0-KvnnT3SgOLdLTY2wE4rDC1-ghc8.10.7.so
>        > strip is /nix/store/xiq6j4jsyj351p8q3yw9cg1hdqp9m685-gcc-wrapper-10.3.0/bin/strip
>        > stripping (with command strip and flags -S) in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/lib  /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin
>        > patching script interpreter paths in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0
>        > checking for references to /build/ in /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0...
>        > RPATH of binary /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server contains a forbidden reference to /build/
>        For full logs, run 'nix log /nix/store/5hz103hms77w33chs4170kws3h5mxp63-haskell-language-server-1.6.1.0.drv'.
> Error: Process completed with exit code 1.
> ```

Yes:

```
       > RPATH of binary /nix/store/y2m3d2c0a1wa1j0p7skgvfvzm9sqnh5d-haskell-language-server-1.6.1.0/bin/haskell-language-server contains a forbidden reference to /build/
```

I'm having a look.

guibou:
@jneira the build is now fixed, but now that HLS is not statically linked anymore, it does have the nix `ghc` package as a dependency, which bumps the uncompressed closure size to `6 GiB`.

I initially thought that it should not be an issue, because people are actually using HLS to develop with GHC, so chances are that they also pull `ghc` in their dev environment. However, due to how `nix flake` works, they may use a different `nixpkgs` clone, and hence use a different `ghc`.

configuration-ghc-921.nix
```diff
@@ -5,28 +5,43 @@ let
...
     "hls-stylish-haskell-plugin"
+    "hls-hlint-plugin"
+    "hls-haddock-comments-plugin"
+    "hls-alternate-number-format-plugin"
+    "hls-eval-plugin"
+    "hls-tactics-plugin"
+    # That one is not technically a plugin, but by putting it in this list, we
+    # get it removed from the top level list of requirement and it is not pull
+    # in the nix shell.
+    "shake-bench"
```
Ailrun, line 16: I think the class plugin should also be disabled. Is it buildable (at least it wasn't last time I tried to make it work with GHC9.2)?

guibou: This list is mostly here to disable the pulling of dependencies which are broken. Considering that the shell and binaries can be built, I don't see the point of removing dependencies for `class` from the shell.

michaelpj (approved):

guibou:
I've disable nix dev shell building for all but the current GHC version. (i.e. that's strictly more than what was previously tested). Let's see if it passes test like that, then we'll merge and later I'll try to fix it.

guibou:
I don't understand the latest failing job. Could we merge with that?

michaelpj:
Now trying to actually use this:

```
> nix develop .#haskell-language-server-ghc921-dev 

error: flake 'git+file:///home/michael/projects/forks/haskell-language-server' does not provide attribute 'devShells.x86_64-linux.haskell-language-server-ghc921-dev', 'packages.x86_64-linux.haskell-language-server-ghc921-dev', 'legacyPackages.x86_64-linux.haskell-language-server-ghc921-dev' or 'haskell-language-server-ghc921-dev'
```

```
> nix develop
> cabal build ghcide

session-loader/Development/IDE/Session/VersionCheck.hs:17:24: error:
    • Exception when trying to run compile-time code:
        I could not find a GHC installation at /nix/store/2dgrlf17sdjg3vjgs637xbj8kfmd3yxw-ghc-8.10.7/lib/ghc-8.10.7. Please do a clean rebuild and/or reinstall GHC.
  error, called at src/GHC/Check.hs:188:7 in ghc-check-0.5.0.6-29a80c8b99e2ee9e01c4d3d0858431da25b86b57be8c1c611f95e6a5d1788945:GHC.Check
      Code: makeGhcVersionChecker
              (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR")
    • In the Template Haskell splice
        $$(makeGhcVersionChecker
             (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
      In the expression:
        $$(makeGhcVersionChecker
             (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
      In an equation for ‘ghcVersionChecker’:
          ghcVersionChecker
            = $$(makeGhcVersionChecker
                   (fromMaybe GHC.Paths.libdir <$> lookupEnv "NIX_GHC_LIBDIR"))
```

:shrug:

michaelpj:
Setting `NIX_GHC_LIBDIR` helped me with the second one.

guibou:
I'm sorry.

I tried your second example and it works fine for me. Have you tried `cabal clean` first? But first, could you `grep -R 2dgrlf17sdjg3vjgs637xbj8kfmd3yxw-ghc-8.10.7 *`. I'm suspecting that there is an hidden / leftover state somewhere. For one reason, I do not have `/nix/store/2dgrlf17sdjg3vjgs637xbj8kfmd3yxw-ghc-8.10.7` on my machine, so you have a reference to something that my invocation of `nix-shell` or `nix-develop` did not pulled.

For the first one

```
nix develop .#haskell-language-server-ghc921-dev 
```

I can indeed reproduce the problem ;( However it works for me with `nix develop .\#haskell-language-server-921-dev` (see the missing `ghc` in my invocation). Is that just an error from your side, or did you read this line from a documentation / example / something and that there is something to fix?

michaelpj:
No need to apologize, just reporting!

> I'm suspecting that there is an hidden / leftover state somewhere.

Yes, I suspect you may be right. `cabal clean` didn't fix it, but I suspect the path may actually have crept into one of the packages in the `cabal` store...

> Is that just an error from your side, or did you read this line from a documentation / example / something and that there is something to fix?

Aha, that does it! I just blindly copied it out of the PR description because I was trying to figure out how to trigger one of the other shells. I should have checked more closely! It does indeed work with the right attribute name :facepalm:

guibou:
> Aha, that does it! I just blindly copied it out of the PR description because I was trying to figure out how to trigger one of the other shells. I should have checked more closely! It does indeed work with the right attribute name facepalm

I use auto completion (with zsh or bash) to get the available entry points. `nix flake show` is also supposed to do that, but I had never been able to get it to work.

Issue #2653, opened 2022-01-29
Modify Sublime Text section in configuration section of documents


pennychase:
I think it would be useful to include a link to the [client configuration section](https://lsp.sublimetext.io/guides/client_configuration/) of the Sublime Text LSP documentation in the Sublime Text section of the HLS client configuration documentation. In particular, the LSP documentation explains how to configure HLS's options within ST (I had wanted to turn off one of the plugins and had no idea how until I chanced upon the ST LSP docs). 

Also, Sublime Text now supports much more of HLS than what is listed in the documentation, so maybe that should be removed. The only  plugins I haven't been able to get to work are eval and Wingman code actions (and I may check in the ST LSP maintainers about Wingman).

michaelpj:
I think we're low on ST users, so it would be lovely if someone who actually uses it like yourself could send us a patch. Much more likely to be right that way :)

PR #2656, opened 2022-01-29
Updated Sublime Text configuration information

pennychase:
Updated the Sublime Text editor-specific configuration information to align with the information in [the Sublime Text LSP docs](https://lsp.sublimetext.io) and also provides a link to the ST LSP docs. This should fix Issue #2653.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2656"></a>


michaelpj (approved):
Great, thank you!

Issue #2436, opened 2021-12-03
Add change type signature quick fix


jneira:
When you change the body of a function and the type signature does not match you get the obvious compiler error:

F.e. for

```haskell
myfunc :: String
myfunc = forM
```

you get

```
• Couldn't match type ‘t0 a0 -> (a0 -> m0 b0) -> m0 (t0 b0)’
                 with ‘[Char]’
  Expected type: String
    Actual type: t0 a0 -> (a0 -> m0 b0) -> m0 (t0 b0)
• Probable cause: ‘forM’ is applied to too few arguments
  In the expression: forM
  In an equation for ‘myfunc’: myfunc = forM
```

It would be nice to have a quick fix (better than a code lens imo) which changes the type signature to match the body new type, something like `Change type signature to: (Traversable t, Monad m) =>t a -> (a -> m b) -> m (t b)`

That type is already shown on hover.

**Additional context**

* We already have a code lens to add the code signature, so this feature could reuse a lot from that one
* That feature is common in other ides like eclipse or intellij

ncfavier:
Would love to see this as well: code lenses are distracting.

michaelpj:
Note that there's already a code action for adding a missing signature, if you don't like the lens.

ncfavier:
I couldn't find it. When clicking on the light bulb that appears next to a top-level definition, I only see retrie's fold/unfold actions.

michaelpj:
Works for me! Unless `lsp-mode` is doing something weird and listing code lenses as code actions also?

jneira:
> Works for me! Unless `lsp-mode` is doing something weird and listing code lenses as code actions also?

I think that is the case, no code action in vscode for me neither

![imagen](https://user-images.githubusercontent.com/54035/145209595-e01f8678-f042-4865-8080-8fd79d420f2f.png)

jneira:
I would say code lenses for *missing* type signature is one of the more appropiate uses of lenses. It displays just the code which will be added if you click it, no replaces or subsitutions and helps you even if you dont add them.
But i agree even for missing type signatures it can be invasive and should have an alternative in the server itself.

@michaelpj what do you think to move upstream the code lenses as code actions thing of emacs lsp mode. Unconditionally or only if user disable the code lenses explicitly.

michaelpj:
[...]

Issue #2014, opened 2021-07-13
Integrate GHC's new diagnostic infrastructure


adinapoli:
Hello folks,

This is a meta ticket, feel free to close it if you think this is not the appropriate place to discuss the subject.

_Summary: GHC changed how diagnostics are emitted, HLS can benefit from this, help needed to integrate the work and feedback welcome to improve GHC to best serve HLS._

Well-Typed has been working together with @goldfirere for the past few months on what used to be GHC proposal [306](https://github.com/ghc-proposals/ghc-proposals/pull/306), i.e. refactor GHC so that it would emit richer Haskell types instead of `SDoc`s for its diagnostic messages. Since the original GHC ticket (cfr.  [ghc#18516](https://gitlab.haskell.org/ghc/ghc/-/issues/18516)) and the inception of our work, we tried to do our best to keep in mind HLS and Haskell IDEs in general as the primary consumers/beneficiaries of this work. The [Wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/Errors-as-(structured)-values) goes a bit deeper into the original design work, if people are interested.

I think we are now at a stage where it would be nice to start thinking how we can collectively integrate this into HLS and what we can learn from this exercise so that we can improve what we have already in GHC before 9.4 (that would be nice but not strictly required, I guess).

### 30 seconds overview

To give you a 30 seconds overview of GHC's `HEAD`, we now have a typeclass:

```hs
class Diagnostic a where
  diagnosticMessage :: a -> DecoratedSDoc
  diagnosticReason  :: a -> DiagnosticReason
  diagnosticHints   :: a -> [GhcHint]
```

And a bunch of types that implement instances for it, which forms a hierarchy:

```hs
data GhcMessage where
  -- | A message from the parsing phase.
  GhcPsMessage      :: PsMessage -> GhcMessage
  -- | A message from typecheck/renaming phase.
  GhcTcRnMessage    :: TcRnMessage -> GhcMessage
  -- | A message from the desugaring (HsToCore) phase.
  GhcDsMessage      :: DsMessage -> GhcMessage
  -- | A message from the driver.
  GhcDriverMessage  :: DriverMessage -> GhcMessage
  -- | An \"escape\" hatch which can be used when we don't know the source of
  -- the message or if the message is not one of the typed ones. 
  GhcUnknownMessage :: forall a. (Diagnostic a, Typeable a) => a -> GhcMessage
```

[...]

PR #2660, opened 2022-01-31
Change Type Signature Plugin

drsooch:
This plugin is designed to take care of #2436 . This a very rough first draft, with a basic outline targeting one specific GHC Type Error message (see `error.txt` for an example). There are certain instances of this error message that __WILL NOT__ work in the current implementation. I just wanted to get a first look and any potential help for further expansion.


ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,129 @@
...
+generateAction :: SigName p => Uri -> [LHsDecl (GhcPass p)] -> Diagnostic -> Maybe (Command |? CodeAction)
+generateAction uri decls diag = diagnosticToChangeSig uri decls diag >>= changeSigToCodeAction
+
+-- | Convert a diagnostic into a ChangeSignature and add the proper SrcSpan
+diagnosticToChangeSig :: SigName p => Uri -> [LHsDecl (GhcPass p)] -> Diagnostic -> Maybe ChangeSignature
+diagnosticToChangeSig uri decls diag = addSrcSpan decls <$> matchingDiagnostic uri diag
+
+-- | If a diagnostic has the proper message create a ChangeSignature from it
+matchingDiagnostic :: Uri -> Diagnostic -> Maybe ChangeSignature
+matchingDiagnostic uri diag@Diagnostic{_message} = unwrapMatch $ _message =~ ("Expected type: (.+)\n +Actual type: (.+)\n.*\n +In an equation for ‘(.+)’" :: Text)
```
jneira, line 79: minor nitpick: i would move the message to where bindings under `expectedMsg` or something alike

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -243,6 +246,14 @@ allLspCmdIds pid commands = concatMap go commands
 
 -- ---------------------------------------------------------------------
 
+getNormalizedFilePath :: Monad m => PluginId -> TextDocumentIdentifier -> ExceptT String m NormalizedFilePath
```
jneira, line 249: i guess something like this is being used everywhere and a nice refactoring would be use it (forother pr)

drsooch: Read my mind!!!

drsooch:
Does anyone have a preference as to where this plugin should go (if the functionality expands out a bit more?) Separate module? Leave it in `ghcide`?

pepeiborra:
> Does anyone have a preference as to where this plugin should go (if the functionality expands out a bit more?) Separate module? Leave it in `ghcide`?

Move it outside please - standalone plugins have better isolation, error handling, run in parallel, and are much easier to upgrade gradually when there is a new GHC version

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+diagnosticToChangeSig uri decls diag = addSrcSpan decls <$> matchingDiagnostic uri diag
+
+-- | If a diagnostic has the proper message create a ChangeSignature from it
+matchingDiagnostic :: Uri -> Diagnostic -> Maybe ChangeSignature
+matchingDiagnostic uri diag@Diagnostic{_message} = unwrapMatch $ _message =~ expectedMessage
+    where
+        expectedMessage = "Expected type: (.+)\n +Actual type: (.+)\n.*\n +In an equation for ‘(.+)’" :: Text
+        unwrapMatch :: (Text, Text, Text, [Text]) -> Maybe ChangeSignature
+        unwrapMatch (_, _, _, [exp, act, name]) = Just $ ChangeSignature exp act name Nothing diag uri
+        unwrapMatch _                           = Nothing
```
pepeiborra, line 84: This looks fine, but GHC 9.4 offers a typed api for diagnostics which hopefully removes the need to parse error messages. See #2014

pepeiborra (commented):
Looks fine to me

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
drsooch: Do you think it would be better to hold off until the new Diagnostic changes are integrated? Reading through the issue it seems like a big effort so maybe it would be better just to start now and migrate to the new format down the road?

pepeiborra: No need to wait, we still want present GHCs to be able to use this plugin

pepeiborra (approved):

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+      pure $ List actions
+
+getDecls :: MonadIO m => IdeState -> NormalizedFilePath -> ExceptT String m [LHsDecl GhcPs]
+getDecls state = handleMaybeM "Error: Could not get Parsed Module"
+    . liftIO
+    . fmap (fmap (hsmodDecls . unLoc . pm_parsed_source))
+    . runAction "changeSignature.GetParsedModule" state
+    . use GetParsedModule
+
+data ChangeSignature = ChangeSignature { expectedType :: Text
```
michaelpj, line 54: haddock?

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+                                       , declName     :: Text
+                                       , declSrcSpan  :: Maybe SrcSpan
+                                       , diagnostic   :: Diagnostic
+                                       , uri          :: Uri
+                                       }
+
+-- Needed to trackdown OccNames in signatures
+type SigName p = (HasOccName (IdP (GhcPass p)))
+
+-- | Generate CodeActions from a list of Diagnostics
```
michaelpj, line 65:
```suggestion
-- | Generate CodeActions from a list of Diagnostics given the module URI and declarations.
```

& below

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+generateActions :: SigName p => Uri -> [LHsDecl (GhcPass p)] -> [Diagnostic] -> [Command |? CodeAction]
+generateActions uri = mapMaybe . generateAction uri
```
michaelpj, line 67: I'd probably just inline this :shrug:

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+
+-- | Convert a diagnostic into a ChangeSignature and add the proper SrcSpan
+diagnosticToChangeSig :: SigName p => Uri -> [LHsDecl (GhcPass p)] -> Diagnostic -> Maybe ChangeSignature
+diagnosticToChangeSig uri decls diag = addSrcSpan decls <$> matchingDiagnostic uri diag
+
+-- | If a diagnostic has the proper message create a ChangeSignature from it
+matchingDiagnostic :: Uri -> Diagnostic -> Maybe ChangeSignature
+matchingDiagnostic uri diag@Diagnostic{_message} = unwrapMatch $ _message =~ expectedMessage
+    where
+        expectedMessage = "Expected type: (.+)\n +Actual type: (.+)\n.*\n +In an equation for ‘(.+)’" :: Text
```
michaelpj, line 81: I wonder if it's worth adding a unit test or two just for the regex match with a couple of expected error messages? It seems like that's often the most fallible part of these plugins, so it might be nice to have something that asserts: yes, it really does match errors like this and that.

Obviously that doesn't help us if the error messages change, but we have the tests with real GHCs for that!

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+        unwrapMatch :: (Text, Text, Text, [Text]) -> Maybe ChangeSignature
+        unwrapMatch (_, _, _, [exp, act, name]) = Just $ ChangeSignature exp act name Nothing diag uri
+        unwrapMatch _                           = Nothing
+
+-- | Given a String with the name of a declaration, find that declarations type signature location
+-- This is a modified version of functions found in Development.IDE.Plugin.CodeAction
```
michaelpj, line 87: could you add a sentence saying why the other one doesn't do what you need? could we fix it to make it more useful?

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+
+changeSigToCodeAction :: ChangeSignature -> Maybe (Command |? CodeAction)
+-- Does not generate a Code action if declSrcSpan is Nothing
+changeSigToCodeAction ChangeSignature{..} = declSrcSpan *> Just (InR CodeAction { _title       = mkChangeSigTitle declName actualType
+                                                                                , _kind        = Just CodeActionQuickFix
+                                                                                , _diagnostics = Just $ List [diagnostic]
+                                                                                , _isPreferred = Nothing
+                                                                                , _disabled    = Nothing
+                                                                                -- This CAN but probably never will be Nothing
+                                                                                , _edit        = mkChangeSigEdit uri declSrcSpan (mkNewSignature declName actualType)
```
michaelpj, line 109: It seems a little awkward to pass in a `Maybe SrcSpan` here when you should know it's there by now. Why not write this with a do block?
```
do 
  srcSpan <- declSrcSpan
  pure $ InR CodeAction { ... , mkChangeSigEdit uri srcSpan ... }
```
and then `mkChangeEdit` can take a non-`Maybe` `SrcSpan`.

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
```diff
@@ -0,0 +1,130 @@
...
+        let txtEdit = TextEdit (realSrcSpanToRange ss) replacement
+            changes = Just $ Map.singleton uri (List [txtEdit])
+        in Just $ WorkspaceEdit changes Nothing Nothing
+mkChangeSigEdit _ _ _                                   = Nothing
+
+mkNewSignature :: Text -> Text -> Text
+mkNewSignature declName actualType = declName <> " :: " <> actualType
+
+--------------------------------------------------------------------------------
+-- test :: Set.Set Int -> Int
```
michaelpj, line 127: commented out code

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
michaelpj: Also, I thought this would do the wrong thing for some error messages, right? In the issue you pointed out that sometimes the error will highlight a *part* of the type that's wrong. We don't want to then go in and replace the whole signature in that case! Could do with some tests for the cases we shouldn't pick up and maybe a comment explaining why we think we're safe.

michaelpj (approved):
Generally looks good!

Since Pepe mentioned the structured errors infrastructure, I wonder whether this is actually a good opportunity to make a feature request upstream! I know Richard has been looking for feedback from us about what would actually be useful in structured errors. Maybe we could open a GHC issue explaining what we're trying to do and asking if there would be a good way to include extra information in structured type-mismatch errors so we could always work out the "actual" type of the enclosing binding or something.

ghcide/src/Development/IDE/Plugin/ChangeTypeAction.hs
drsooch: Yeah, I hadn't dealt with most of the edge cases yet. I was just curious if someone much smarter than I knew of some mechanism of the GHC API that would be useful to look at. It seems like this is the only way until the new Diagnostics in `9.4`. 

And yes I have local unit tests just haven't bothered pushing them yet.

drsooch:
https://gitlab.haskell.org/ghc/ghc/-/issues/21063 -- Linking this here

drsooch:
Going to mark this as ready for review, there are a few things I'd like to clean up and I've left relevant comments.

plugins/hls-change-type-signature-plugin/README.md
```diff
@@ -0,0 +1,15 @@
+# Change Type Signature Plugin
```
drsooch, line 1: Whoops... Need to add some documentation here.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,237 @@
...
+
+codeActionHandler :: PluginMethodHandler IdeState TextDocumentCodeAction
+codeActionHandler ideState plId CodeActionParams {_textDocument = TextDocumentIdentifier uri, _context = CodeActionContext (List diags) _} = response $ do
+      nfp <- getNormalizedFilePath plId (TextDocumentIdentifier uri)
+      decls <- getDecls ideState nfp
+      let sigs = [ sigToText ts | L (RealSrcSpan rss _) (SigD _ ts@(TypeSig _ idsSig _)) <- decls]
+      let actions = mapMaybe (generateAction uri decls) diags
+      liftIO $ logDebug (ideLogger ideState) $ T.pack $ "CTS Diags: " <> show diags
+      liftIO $ logDebug (ideLogger ideState) $ T.pack $ "CTS Actions: " <> show actions
+      liftIO $ logDebug (ideLogger ideState) $ T.pack $ "CTS Sigs: " <> show sigs
```
drsooch, line 59: Going to delete these.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,237 @@
...
+-- | List of regexes that match various Error Messages
+errorMessageRegexes :: [Text]
+errorMessageRegexes = [ -- be sure to add new Error Messages Regexes at the bottom to not fail any existing tests
+    "Expected type: (" <> typeSigRegex <> ")\n +Actual type: (" <> typeSigRegex <> ")\n(.|\n)+In an equation for ‘(.+)’"
+    , "Couldn't match expected type ‘(" <> typeSigRegex <> ")’ with actual type ‘(" <> typeSigRegex <> ")’\n(.|\n)+In an equation for ‘(.+)’"
+    ]
+    where
+        -- The brackets in position 1 and 2 MUST BE IN THAT ORDER
+        -- Any other combination of brackets seems to fail
+        typeSigRegex = "[][a-zA-Z0-9 ->]+"
```
drsooch, line 144: Need to flesh this out more. Maybe I should just go with `(.*)\n`?

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,237 @@
...
+-- This function returns the actual location of the signature rather than the actual signature
+-- We also don't have access to `fun_id` or other actual `id` so we must use string compare instead
+findSigLocOfStringDecl :: SigName => [LHsDecl GhcPs] -> String -> Maybe (RealSrcSpan, DefinedSig)
+findSigLocOfStringDecl decls declName = do
+    -- can we simplify this logic? Just want to make sure ghcSig is a Just value
+    (rss, Just ghcSig) <- listToMaybe [ (locA rss, sigToText ts)
+                                        | L (RealSrcSpan rss _) (SigD _ ts@(TypeSig _ idsSig _)) <- decls,
+                                        any ((==) declName . occNameString . occName . unLoc) idsSig
+                                      ]
+    pure (rss, ghcSig)
```
drsooch, line 158: Is there a way to remove the last line? `pure (rss, ghcSig)`

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,237 @@
...
+
+freshUniqueVars :: [Text]
+freshUniqueVars = map T.singleton validVars <> [ a <> b | a <- freshUniqueVars, b <- freshUniqueVars ]
+    where
+        validVars = filter (`notElem` ['f', 'm', 't']) ['a'..'z']
+
+
+replaceWithFreshVar :: ([Text], Text) -> Text -> ([Text], Text)
+replaceWithFreshVar (replacement:rest, haystack) needle = (rest, T.replace needle replacement haystack)
+replaceWithFreshVar _ _ = error "How do I handle this..., the pattern match for infinite list shouldn't fail"
```
drsooch, line 237: The replacement list is always infinite, is there a way to encode this so I don't need to have the error match case?

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,237 @@
...
+        let txtEdit = TextEdit (realSrcSpanToRange ss) replacement
+            changes = Just $ Map.singleton uri (List [txtEdit])
+        in WorkspaceEdit changes Nothing Nothing
+
+mkNewSignature :: Text -> Text -> Text
+mkNewSignature declName actualType = declName <> " :: " <> tidyActualType actualType
+
+-- | Cleans up type signatures with oddly-named type variables (a0, m0 ...)
+tidyActualType :: Text -> Text
+tidyActualType actualType =
```
drsooch, line 204: Is this actually more complex than it's worth?

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,237 @@
...
+    (rss, Just ghcSig) <- listToMaybe [ (locA rss, sigToText ts)
+                                        | L (RealSrcSpan rss _) (SigD _ ts@(TypeSig _ idsSig _)) <- decls,
+                                        any ((==) declName . occNameString . occName . unLoc) idsSig
+                                      ]
+    pure (rss, ghcSig)
+
+-- | Pretty Print the Type Signature (to validate GHC Error Message)
+sigToText :: Sig GhcPs -> Maybe Text
+sigToText = \case
+  ts@TypeSig {} -> stripSignature (T.pack $ showSDocUnsafe $ ppr ts)
```
drsooch, line 163: Is there a better way than `showSDocUnsafe`? What is unsafe about it?

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
pepeiborra: `sequence`

pepeiborra, line 153:
```suggestion
    sequence <$> listToMaybe [ (locA rss, sigToText ts)
                                        | L (RealSrcSpan rss _) (SigD _ ts@(TypeSig _ idsSig _)) <- decls,
                                        any ((==) declName . occNameString . occName . unLoc) idsSig
                                      ]
```

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
pepeiborra: It ignores the `DynFlags` that control formatting and instead uses some default values.
* We want to ignore the formatting Dynflags, otherwise user set Cabal flags could break the code action!
* We don't want to blindly accept default values as those can change over GHC versions and could break the code action!

pepeiborra: I would recommend using `prettyPrint`:

https://github.com/haskell/haskell-language-server/blob/4386396624180de112d04335a75334d04781110c/ghcide/src/Development/IDE/GHC/Util.hs#L135

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
pepeiborra: yes

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+-- | The signature provided by GHC Error Message (Actual type)
+type ActualSig = Text
+-- | The signature defined by the user
+type DefinedSig = Text
+
+-- | DataType that encodes the necessary information for changing a type signature
+data ChangeSignature = ChangeSignature {
+                         -- | The expected type based on Signature
+                         expectedType :: ExpectedSig
+                                       ,
```
michaelpj, line 79: I've never seen this style of comma placement before! Up to you, but I think either before or after the field on the same line is more typical.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+    -- Find the definition and it's location
+    (declSrcSpan, ghcSig) <- findSigLocOfStringDecl decls (T.unpack declName)
+    -- Make sure the given "Actual Type" is a full signature
+    isValidMessage expectedType ghcSig
+    pure $ ChangeSignature{..}
+
+-- | Does the GHC Error Message give us a Signature we can use?
+-- We only want to change signatures when the "expected signature" given by GHC
+-- matches the one given by the user
+isValidMessage :: DefinedSig -> ExpectedSig -> Maybe ()
```
michaelpj, line 116: I think slightly more idiomatic to return a bool and use `guard` at the call site.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+isValidMessage defSig expSig = if defSig == expSig then Just () else Nothing
+
+-- | If a diagnostic has the proper message create a ChangeSignature from it
+matchingDiagnostic :: Diagnostic -> Maybe (ExpectedSig, ActualSig, DeclName)
+matchingDiagnostic diag@Diagnostic{_message} = case map (unwrapMatch . (=~) _message) errorMessageRegexes of
+                                                       []  -> Nothing
+                                                       css -> join $ find isJust css
```
michaelpj, line 123:
```suggestion
matchingDiagnostic diag@Diagnostic{_message} = asum $ map (unwrapMatch . (=~) _message) errorMessageRegexes
```

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+        unwrapMatch _                              = Nothing
+
+-- | List of regexes that match various Error Messages
+errorMessageRegexes :: [Text]
+errorMessageRegexes = [ -- be sure to add new Error Messages Regexes at the bottom to not fail any existing tests
+    "Expected type: (" <> typeSigRegex <> ")\n +Actual type: (" <> typeSigRegex <> ")\n(.|\n)+In an equation for ‘(.+)’"
+    , "Couldn't match expected type ‘(" <> typeSigRegex <> ")’ with actual type ‘(" <> typeSigRegex <> ")’\n(.|\n)+In an equation for ‘(.+)’"
+    ]
+    where
+        -- The brackets in position 1 and 2 MUST BE IN THAT ORDER
```
michaelpj, line 137: Mysterious.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+        -- Any other combination of brackets seems to fail
+        typeSigRegex = "[][a-zA-Z0-9 ->]+"
+
+-- | Given a String with the name of a declaration, find that declarations and give back
+-- the type signature location and the full signature
+-- This is a modified version of functions found in Development.IDE.Plugin.CodeAction
+-- This function returns the actual location of the signature rather than the actual signature
+-- We also don't have access to `fun_id` or other actual `id` so we must use string compare instead
```
michaelpj, line 145: Hmmmmm. So this could potentially go wrong if we had two functions with the same name, perhaps one of them is a local function. And you try and change signature for the second one? Maybe worth a test.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+findSigLocOfStringDecl :: SigName => [LHsDecl GhcPs] -> String -> Maybe (RealSrcSpan, DefinedSig)
+findSigLocOfStringDecl decls declName = do
+    -- can we simplify this logic? Just want to make sure ghcSig is a Just value
+    (RealSrcSpan rss _, Just ghcSig) <- listToMaybe [ (locA ss, sigToText ts)
```
michaelpj, line 149: Ugh, this is a partial pattern match, does this call `fail` or something if it doesn't match? ugly.

michaelpj: At least don't do the match here and add a `ghcSig <- mayGhcSig` line below or something. Haven't processed the rest enough to think of a simplification.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+                                                               , _diagnostics = Just $ List [diagnostic]
+                                                               , _isPreferred = Nothing
+                                                               , _disabled    = Nothing
+                                                               , _edit        = Just $ mkChangeSigEdit uri declSrcSpan (mkNewSignature declName actualType)
+                                                               , _command     = Nothing
+                                                               , _xdata       = Nothing
+                                                               }
+
+mkChangeSigTitle :: Text -> Text -> Text
+mkChangeSigTitle declName actualType = "change signature for ‘" <> declName <> "’ to: " <> actualType
```
michaelpj, line 186: Captialize first letter? Not sure if we do this consistently.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+
+mkChangeSigEdit :: Uri -> RealSrcSpan -> Text -> WorkspaceEdit
+mkChangeSigEdit uri ss replacement =
+        let txtEdit = TextEdit (realSrcSpanToRange ss) replacement
+            changes = Just $ Map.singleton uri (List [txtEdit])
+        in WorkspaceEdit changes Nothing Nothing
+
+mkNewSignature :: Text -> Text -> Text
+mkNewSignature declName actualType = declName <> " :: " <> actualType
```
michaelpj, line 195: Perhaps stupid question, but should we be using exactprint somehow to generate stuff like this?

michaelpj: For example: what happens if there are pragmas or something attached to the declaration? will they get clobbered? If they're included in the decl span they might do.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -0,0 +1,247 @@
...
+
+-- | Cleans up type signatures with oddly-named type variables (a0, m0 ...)
```
michaelpj, line 197: I'm a little unclear why we need to do this. We're using the signature from the message, right? Are they often bad?

plugins/hls-change-type-signature-plugin/test/Main.hs
```diff
@@ -0,0 +1,138 @@
...
+    where
+        codeActionsFromLocs doc = concat <$> mapM (getCodeActions doc . uncurry pointRange) locs
+
+findChangeTypeActions :: [Command |? CodeAction] -> Session [CodeAction]
+findChangeTypeActions = pure . filter isChangeTypeAction . rights . map toEither
+    where
+        isChangeTypeAction CodeAction{_kind} = case _kind of
+          Nothing -> False
+          Just kind -> case kind of
+            CodeActionQuickFix -> True
```
michaelpj, line 117: FYI you can give your code action a kind "under" quickfix if you want, then you can match against it with https://hackage.haskell.org/package/lsp-types-1.4.0.1/docs/Language-LSP-Types.html#v:codeActionKindSubsumes

plugins/hls-change-type-signature-plugin/test/testdata/TErrorGivenPartialSignature.hs
```diff
@@ -0,0 +1,4 @@
+module ErrorGivenPartialSignature where
```
michaelpj, line 1: missing expected file?

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
drsooch: Honestly, I just noticed it myself. I think I used a code action to generate comments and that's how it ended up

michaelpj: Maybe a bug report for the haddock comment insertion plugin!

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
drsooch: Yes, Pepe already mentioned there is no reason to tidy the signatures up. I just went overboard. I'm going to drop it.

plugins/hls-change-type-signature-plugin/test/testdata/TErrorGivenPartialSignature.hs
drsooch: No this one shouldn't generate any CodeAction as the error doesn't provide enough information

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
drsooch: Yeah, I reached out to the maintainers of regex-tdfa. I was always under the assumption that you could escape things in bracket expressions, but apparently not.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
drsooch: I'll look into it

plugins/hls-change-type-signature-plugin/test/Main.hs
drsooch: Are you saying something like `CodeActionUnknown "quickfix.changeSignature"`

michaelpj: Yeah (also it even has an `IsString` instance). There's no real logic to what to put as the bits underneath, use your judgement.

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
drsooch: I did some testing and I don't see any issues with Language Pragmas near signatures. I'm also not really familiar with `exactprint`. Considering we have to parse the GHC message to get the signature it seems easier to just to build it up ourselves?

michaelpj: I was thinking more pragmas that might get attached to the declaration like `{-# INLINE #-}`.

I'm also not really familiar with it: I just note that if the chunk of AST you're replacing includes stuff that isn't just the signature, you'll clobber it. I think this is the sort of thing exactprint solves, but I also haven't used it in anger, perhaps @pepeiborra can weigh in on whether this is something to worry about?

pepeiborra: Using exactprint to edit the AST is always the principled thing to do, we already have helpers to edit existing type signatures:

https://github.com/haskell/haskell-language-server/blob/4b8f9d53bed3e775ccd120d9a2471ae374a88db3/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs#L16-L18

But admittedly, exactprint is really hard to use, and the api changes in 9.2 make it even harder!

drsooch: > I was thinking more pragmas that might get attached to the declaration like `{-# INLINE #-}`.

Yes sorry, not sure why I said Language Pragma. Those are the ones I meant!

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
drsooch: The only way this would fail is if the functions have identical signatures. (There are some edge cases where there are no issues). I'd argue that bindings that have the same signatures and name are an extreme edge case.

docs/features.md
```diff
@@ -237,6 +237,22 @@ Provides a variety of code actions for interactive code development, see <https:
...
 
+### Change Type Signature
+
+Provided by: `hls-change-type-signature-plugin`
+
+Code action kind: `quickfix`
+
+Change/Update a type signature to match implementation.
+
+Status: Until GHC 9.4, the implementation is ad-hoc and relies on GHC error messages to create a new signature. Not all GHC error messages are supported.
```
michaelpj, line 248: Maybe in a "Known limitations" section like a few others have?

michaelpj (approved):
On tiny doc nit otherwise LGTM!

drsooch:
Sorry for this marathon PR. I pushed a change to handling local signatures for better support. New tests aren't passing locally, just going to see if it'll work in CI. I'll revisit this tonight.

michaelpj:
This looks green but I wasn't sure if you still have work you want to do or not!

drsooch:
Yes this is all set!

pepeiborra:
Ooops, we forgot to add the test suite to CI

https://github.com/haskell/haskell-language-server/blob/0d4f98eff2acc4e5df2d2bdd6111b04f91d1ec38/.github/workflows/test.yml#L0-L1

Issue #2659, opened 2022-01-30
Issues around th warning (`This HLS binary does not support Template Haskell`) and build hls with dynamic linking


jneira:
After adding the warning about TH a bunch of issues have come up:

- The instructions to build in the troubleshooting guide were incomplete, it did not include specific commands for stack or `ghcup compile hls`: i tried to fix that with #2647
- Users with env where th was working fine has been compelled to build hls from source only to take ride of the warning (but well, we should have expected this i guess). 
- If you build hls with `--enable-executable-dynamic` (which links dynamically only the system libs) instead of `--ghc-options="-dynamic"` (which additionally links dynamically all haskell deps) the warning does not go away. However i thought the first option was sufficient to workaround th problems (maybe i am wrong but i infer that from comments like [this](https://github.com/haskell/haskell-language-server/issues/1982#issuecomment-873396136))
- As reported in irc, `stack install haskell-language-server --ghc-options="dynamic"` seems to not work out of the box. It seems you have to build it both statically and dynamically (i am trying to reproduce)
- An user in macOS has got this error trying `cabal install pkg:haskell-language-server --ghc-options="-dynamic"`:

```
src/Distribution/Extra/Doctest.hs:34:8: error:
    Could not find module ‘Prelude’
    Perhaps you haven't installed the "p_dyn" libraries for package ‘base-4.14.3.0’?
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
34 | module Distribution.Extra.Doctest (
```

jneira:
> As reported in irc, stack install haskell-language-server --ghc-options="dynamic" seems to not work out of the box. It seems you have to build it both statically and dynamically (i am trying to reproduce)

I've reproduced the error in wsl ubuntu with `stack install haskell-language-server --ghc-options="-dynamic"`

```
hiedb                             > [1 of 9] Compiling HieDb.Compat
hiedb                             >
hiedb                             > /tmp/stack-6de8f3e7cd60c56b/hiedb-0.4.1.0/src/HieDb/Compat.hs:84:1: error:
[...]

PR #2668, opened 2022-01-31
Add the -dynamic flag and update build instructions

pepeiborra:
Addresses some of the issues with building HLS locally flagged in #2659

Reasons to add the `-dynamic` flag in the `haskell-language-server.cabal` descriptor:

* `cabal install haskell-language-server --enable-executable-dynamic` doesn't work as it doesn't actually link the dynamic `rts`. Note that `cabal build --enable-executable-dynamic` works fine, so this is an issue with `cabal install`.
* `cabal install haskell-language-server --ghc-option=-dynamic` works, but it applies the `dynamic` flag to all the dependencies, which takes really long as it plays badly with the Cabal store. It will also fail if the user has enabled profiling builds since GHC doesn't ship with a profiling dynamic `rts`.

I was reluctant to do this previously since `-dynamic` might not work in all the supported operating systems, e.g. Windows. Let's see what happens...


hasufell:
How does a user disable this flag now?

pepeiborra:
> How does a user disable this flag now?

I added a Cabal flag that we can use for release builds. Why would a user need to disable this flag though?

michaelpj:
Can we preserve the reasoning from the PR description somewhere? Perhaps in the cabal file where the flag is defined? This stuff is quite fiddly and if we don't write it down we'll forget why we're doing what we're doing.

hasufell:
> > How does a user disable this flag now?
> 
> I added a Cabal flag that we can use for release builds. Why would a user need to disable this flag though?

I don't know. But assumptions often explode into one's face, so better prepare for an exit strategy 😅

.github/workflows/build.yml
```diff
@@ -48,6 +48,11 @@ jobs:
         echo "tests: false" >> cabal.project.local
         echo "benchmarks: false" >> cabal.project.local
 
+    - name: Disable -dynamic
```
michaelpj, line 51: Why do we disable it here?

michaelpj (commented):
I haven't been following the discussion enough to know if this is actually the right thing to do :sweat_smile:

.github/workflows/build.yml
hasufell: Release builds are static, but we tell users to build dynamic. ..it's all quite confusing 😄

Afaiu static builds work fine as long as you use an official GHC bindist (does it work across ubuntu vs fedora bindists?)...otherwise TH may break.

pepeiborra: We want to build static binaries for ease of distribution. Sadly static binaries are no good for loading code dynamically, which is needed for compiling TH code.

jneira (changes requested):
sorry but I am not sure about this, it breaks the default build for windows and windows users must disable the flag explicitly to make it work

so at very least I would ask to make the flag only applicable in non windows os via a cabal condition `if !os(windows)`

otoh at this point I am not sure if this is needed for sure to get a hls binary which works for TH

are we 100% sure a binary dinamically linked in the target os only against system libs (the default link mode in Linux!) does not work? maybe simply build hls in the target os is the key, as we are compiling exactly with same ghc used in the projects

the fact we did not get to reproduce the th issue in CI makes very difficult get a solution

so I think it worths to investigate a little bit more the issue before enable this which could open a new whole kind of issues

jneira:
This was suggested in this comment: https://github.com/haskell/haskell-language-server/issues/2659#issuecomment-1026248862

> So adding the -dynamic flag in the Cabal descriptor really is the only way to get what we want.

Not sure if can trust cabal but maybe there would be another way, adding to a `cabal.project` or `cabal.project.local`:

```
package haskell-language-server
  ghc-options: -dynamic
```

only for linux and macos of course

pepeiborra:
> sorry but I am not sure about this, it breaks the default build for windows and windows users must disable the flag explicitly to make it work
> so at very least I would ask to make the flag only applicable in non windows os via a cabal condition `if !os(windows)`

The only reason I didn't disable it for Windows initially is because I wanted to see CI fail. The Cabal condition is added now.

> otoh at this point I am not sure if this is needed for sure to get a hls binary which works for TH
> 
> are we 100% sure a binary dinamically linked in the target os only against system libs (the default link mode in Linux!) does not work? maybe simply build hls in the target os is the key, as we are compiling exactly with same ghc used in the projects
> 
> the fact we did not get to reproduce the th issue in CI makes very difficult get a solution
> 
> so I think it worths to investigate a little bit more the issue before enable this which could open a new whole kind of issues

I think your concerns are unfounded. We are simply aligning HLS with GHC:
```
$ ghc +RTS --info                                                            
 [("GHC RTS", "YES")
 ,("GHC version", "8.10.7")
 ,("RTS way", "rts_thr_dyn")
 ```

I suppose what you are asking for is a repro. It shouldn't be too difficult to create one just by going through the myriad of existing TH issues.

jneira:
> The only reason I didn't disable it for Windows initially is because I wanted to see CI fail. The Cabal condition is added now.

Nice, many thanks.

> I suppose what you are asking for is a repro. It shouldn't be too difficult to create one just by going through the myriad of existing TH issues.

Yeah it would make our steps towards the solution in a more confident way for sure.
Adding such repro, make it fail with a default hls binary and work with the dynamically linked one would be really great

jneira (approved):
Ok this is clearly a improvement over the actual situation and make build instructions clearer.
Moreover it makes the build work for cabal and stack.

But it is a workaround over a cabal bug and a stack limitation (stack would need a package entry in the stack,yaml with ghc-options) and it should not be needed, as already commented in the flag (well the cabal part)

I really hope this would not open the door to other bugs, but it has been [tested in stack succesfully](https://github.com/haskell/haskell-language-server/issues/2659#issuecomment-1026786626), where `--ghc-options="dynamic"` failed.

Linux is not my primary os and my knowledge of linux is limited so i am gonna trust my co-maintainers living in linux

Thanks!

.github/workflows/build.yml
jneira: > Afaiu static builds work fine as long as you use an official GHC bindist (does it work across ubuntu vs fedora bindists?)...otherwise TH may break.

Yeah i guess those are the users who had th working but have to close the warning or build hls from source without needing it.

pepeiborra:
@jneira can we bump the version number of `haskell-language-server.cabal` and upload it to Hackage?

jneira:
@MergifyIO backport 1.6.1.1-hackage

Issue #2673, opened 2022-02-01
Wrong sort order in autocomplete list.


Akhristenko:
### Your environment

Which OS do you use: Ubuntu 20.04
Which LSP client (editor/plugin) do you use: VS Code
Describe your project (alternative: link to the project): stack new-template with added bytestring dependency

### Steps to reproduce
Modify file Main.hs to this:
```
module Main where

import qualified Data.ByteString as BS

main :: IO ()
main = do
    BS.read
    pure ()
```
place cursor after `BS.read` and invoke autocomplete.

### Expected behaviour
Function `readFile` from `Data.ByteString` on first position, before all other functions that not from `Data.ByteString`

### Actual behaviour
Function `readFile`on fourth position.
The first three positions are occupied by functions `read` from `GHC.IO.Device`, `Prelude`, `Text.Read`.

jneira:
hi thanks for the bug report, it is similar to #2082 that was marked closed, maybe incorrectly
or it is a regression

pepeiborra:
What version of HLS are you using ?

Akhristenko:
I'm using haskell-language-server-1.6.1.0-linux-8.10.7
This bug appears only in some cases:
if we invoke autocompletion on `BS.re` then sequence be correct (`readFile` from `Data.ByteStirng`, `replicate` from `Data.ByteString`, `reverse` from `Data.ByteString`, functions from other modules)
but if we invoke autocompletion on `BS.read` then sequence be incorrect (`read` from `GHC.IO.Device`, `read` from `Prelude`, `read` from `Text.Read`, `readFile` from `Data.ByteString`, functions from other modules)

pepeiborra:
I can't seem to repro this

Akhristenko:
I dumped the vscode and hls conversations and found that the functions are in the correct order. The reordering happens somewhere on the side of the vscode. So, this is not haskell-language-server issue.

jneira:
oh, that is unfortunate if we can't track the root cause, maybe something still could be done in the sever part to make the client behave correctly

maybe it worths a issue in the vscode extension? do you mind if I reopen and move this there?

Akhristenko:
> do you mind if I reopen and move this there?

I don't mind

I found additional info, that can help reproduce the bug.
In my client i have in settings this line: `"haskell.maxCompletions": 10000`
One of my friends, after setting this parameter, also observes a bug

PR #2676, opened 2022-02-01
Add a test for #2673

pepeiborra:
I cannot reproduce #2673 locally. Let's see if CI can

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2676"></a>


jneira (approved):
I guess the tests is useful in its own and we can merge this

Issue #2678, opened 2022-02-01
CI: Check/rewrite carrige return 


Anton-Latukha:
As diff in https://github.com/haskell/haskell-language-server/pull/2597/files shows - all touched files were rewritten fully (every line in number of files got `^M`, most probably due to Windows line break).

As a result, after merge that created additional challenges doing rebases or working with those files, since there are no longer diffs in the touched files.

Would be useful to prevent such diffs in the future.

drsooch:
Looks like the pre-commit hook in the contributing docs has support for mixed line endings https://pre-commit.com/hooks . Not familiar myself with the tool, so not sure how easy it is to add.

jneira:
I am suffering it continuously, the repo has unix line endings as default in 

https://github.com/haskell/haskell-language-server/blob/47cb213ac70dabb567461ac93af6933f79c39f4f/.editorconfig#L7

I have the .editorconfig extension installed in vscode so the project uses LF by default even for new files

And i am a good boy and have `core.autocrlf=input` in my git config instead the default `auto`

So the hook formats my code correctly, changes new lines to CRLF and then the commit fail, then i revisit the file, save it again to make the editor correct line endings to LF and redo the commit

:worried: 

//cc @bradrn

bradrn:
Oh dear, sorry if I messed everything up! I merely ran the pre-commit hook and committed what it altered, including all the line ending changes. I also have had issues with line endings in my repos (thanks to `core.autocrlf=auto`), and I still am not quite comfortable with how everything works together in this regard, so I’m probably not the best person to help here.

jneira:
> Oh dear, sorry if I messed everything up! I merely ran the pre-commit hook and committed what it altered, including all the line ending changes. I also have had issues with line endings in my repos (thanks to `core.autocrlf=auto`), and I still am not quite comfortable with how everything works together in this regard, so I’m probably not the best person to help here.

dont worry thaose are the little annoyances of develop in windows, we are in minority, so please, dont let me alone :-P

I would suggest you switch to `core.autocrlf=input` and install the appropiate plugin in your editor to honour `.editorconfig`.
It would be useful beyond the line endings thing and for other projects too. Nowadays you can stick to LF in windows as almost all editors lets you use them (but notepad :-P? not sure maybe it can now).

[...]

PR #2679, opened 2022-02-02
Add pre-commit hook for cleaning up mixed-line endings

drsooch:
Figured I'd offer a solution for #2678. I don't use nix so I'm not sure if there is anything that needs to be done to integrate in nix (also not sure if it's even necessary). Most of the file commits are the line endings with a few interspersed stylish-haskell changes.


plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
```diff
@@ -59,9 +59,9 @@ showDiff (First w)  = "WAS " <> w
 showDiff (Second w) = "NOW " <> w
 showDiff (Both w _) = w
 
-testCheck :: Bool -> (Section, Test) -> [T.Text] -> [T.Text]
-testCheck diff (section, test) out
-    | not diff || null (testOutput test) || sectionLanguage section == Plain = out
+testCheck :: (Section, Test) -> [T.Text] -> [T.Text]
+testCheck (section, test) out
+    | null (testOutput test) || sectionLanguage section == Plain = out
```
drsooch, line 64: This concerns me I have no idea what this is from....

drsooch: Nevermind I mis-merged...

michaelpj:
Hmm, isn't there a git setting that sorts this out for you? Something with `git config core.autocrlf`?

jneira:
> Hmm, isn't there a git setting that sorts this out for you? Something with `git config core.autocrlf`?

yeah `core.autocrlf=input` to ensure you use LF endings (the used ones in the remote repo) but:
* git suggests you `core.autocrlf=auto` which keep CRLF in your local and replaces them with LF on commit
  * that has its own caveats: if you share those files from the local copy (for example generating a tarball to upload to hackage) they still will have CRLF and that will make tests in hackage fail. Yes, that happened: https://github.com/haskell/haskell-language-server/issues/1808
* although we have line endings enforced in `.editorconfig` most editors does not honour it by default (vscode amongs them) you have to install a specific extension (as mentioned here: https://github.com/haskell/haskell-language-server/blob/master/docs/contributing/contributing.md#style-guidelines)
  * so you have to remember install the plugin

So maybe add a precommit hook to enforce it is not a bad idea

ghcide/src/Development/IDE/Core/UseStale.hs
```diff
@@ -1,162 +1,162 @@
-{-# LANGUAGE DerivingVia    #-}
```
jneira, line 1: omg all those files had incorrect line endings???? :facepalm:

docs/contributing/contributing.md
```diff
@@ -199,6 +199,17 @@ If you don't want to use [nix](https://nixos.org/guides/install-nix.html), you c
...
+    {
+       "repo": "https://github.com/pre-commit/pre-commit-hooks",
+       "rev": "v4.1.0",
+       "hooks": [
+          {
+            "id": "mixed-line-ending",
+            "args": ["--fix", "lf"],
+            "exclude": "test/testdata/.*CRLF*.hs$"
+          }
+       ]
```
jneira, line 212: what about create a `.pre-commit-config.json` in the repo to make a little bit more easier and disvoerable? docs could link to the file in that case

jneira:
> So maybe add a precommit hook to enforce it is not a bad idea

otoh you have to remember install the hook :shrug:

ghcide/src/Development/IDE/Core/UseStale.hs
jneira: yeah, as described in the issue: https://github.com/haskell/haskell-language-server/issues/2678

jneira:
What about fix the formatting existing hook to make it use LF in its formatting (or at least not change them to CRLF!!!)? we would have the two features with one hook

docs/contributing/contributing.md
drsooch: I can do that as well!

drsooch:
> What about fix the formatting existing hook to make it use LF in its formatting (or at least not change them to CRLF!!!)? we would have the two features with one hook
Ooo I didn't realize `stylish-haskell` could fix line-ending9s. I guess the only downside is it won't get run on a bunch of excluded files? Thankfully, Brittany is the only formatter with tests related to CRLF.

docs/contributing/contributing.md
drsooch: Although, the pre-commit hook is part of `.gitignore`. I can envision some users use their own personal pre-commit hook and having one in source control may cause issues?

Anton-Latukha:
Special thanks for converting those files to the initial line ending. Converting back - would solve Git diff rebases. I thought that is the first thing to do on return, before thinking further, it is already done here. Thank you.

jneira:
> > What about fix the formatting existing hook to make it use LF in its formatting (or at least not change them to CRLF!!!)? we would have the two features with one hook
> 
> Ooo I didn't realize `stylish-haskell` could fix line-endings. I guess the only downside is it won't get run on a bunch of excluded files?

not sure if it could modify line endings but it should not change the existing ones, like it is doing for me lf -> crlf
that alone would be an improvement

.gitignore
```diff
@@ -32,7 +32,7 @@ test/testdata/**/hie.yaml
 .shake/
 
 # pre-commit-hook.nix
-.pre-commit-config.yaml
+#.pre-commit-config.yaml
```
Ailrun, line 35: Is there a way to put the config into nix setting?

drsooch: I don't use nix so I have no idea, any help would be appreciative.

jneira: nix doesn't support windows so I suppose it is not necessary?

jneira (approved):
thanks!

.gitignore
Ailrun: Oh, makes sense.

Anton-Latukha (approved):

PR #2681, opened 2022-02-02
Prepare 1.6.1.1 (only hackage release)

jneira:
* We will use the fourth position for this release cause it will not have a github/gitlab/ghcup release, only a hackage one
* It tries to fix the build workflow, broken since #2668 (see https://github.com/haskell/haskell-language-server/actions/runs/1779913965)
  * I think the cause has been mix the dynamic build in the test workflow and the static one in the build one
  * The cabal store restored from the cache, populated by test/caching workflow makes the static build fail
  * The solution is moving the flag setting to the generic action used for all workflows
* This will be backported to the 1.6.1.1-hackage branch, created on the 1.6.1.0 one

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2681"></a>

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -63,6 +63,10 @@ runs:
         fi
       shell: bash
 
+    # The default build for haskell-language-server executable is dynamic for linux and macOS
+    # to get a executable which works for Template Haskell
+    # However we continue providing full static executables in releases so we have to disable it
+    # *for all workflows*, including test, flags and release builds
     - name: Disable -dynamic
       run: |
         echo -e "package haskell-language-server\n  flags: -dynamic" >> cabal.project.local
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -65,7 +65,7 @@ runs:
 
     - name: Disable -dynamic
       run: |
-        echo "package haskell-language-server\n  flags: -dynamic" >> cabal.project.local
+        echo -e "package haskell-language-server\n  flags: -dynamic" >> cabal.project.local
       shell: bash
 
     - if: inputs.os == 'Windows' && inputs.ghc == '8.8.4'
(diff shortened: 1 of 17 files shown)


jneira:
@Mergifyio backport 1.6.1.1-hackage

pepeiborra (approved):

.github/actions/setup-build/action.yml
```diff
@@ -63,10 +63,15 @@ runs:
         fi
       shell: bash
 
+    - name: Disable -dynamic
```
michaelpj, line 66: I still think this needs a comment.

jneira: there is one in the .cabal file, but maybe could be refer it here

jneira: finally added a full comment with ci specific caveats

.github/actions/setup-build/action.yml
```diff
@@ -63,6 +63,10 @@ runs:
         fi
       shell: bash
 
+    # The default build for haskell-language-server executable is dynamic for linux and macOS
```
michaelpj, line 66: Great, thank you.

michaelpj (approved):

jneira:
I changed the title of the troubleshooting section, breaking the url of the th warning. I had to correct it in the original pr
To have hardcoded urls in the code is a real PITA

michaelpj:
> To have hardcoded urls in the code is a real PITA

Maybe it should just link to the documentation root and then say in prose where to look? e.g. "Check the [documentation](<link>), look for BLAH in the troubleshooting section"

jneira:
> > To have hardcoded urls in the code is a real PITA
> 
> Maybe it should just link to the documentation root and then say in prose where to look? e.g. "Check the [documentation](<link>), look for BLAH in the troubleshooting section"

buy it would be longer and less direct... a  more specific and dedicated URL + redirection in docs would serve as the needed indirection, rtd allows those things?

jneira:
https://hackage.haskell.org/package/haskell-language-server-1.6.1.1 released

PR #2686, opened 2022-02-02
Fix flaky boot def test

### LINKED ISSUE #2677: Flaky 9.0.2 boot-def test
I've seen this 2 or 3 times I think. 9.0.2 ubuntu-latest. I'm gonna try and look at it.
```
ghcide
  code actions
    extend import actions
      with checkAll
        extend single line import with type: OK (2.42s)
  boot
    boot-def-test:                           FAIL
      Last message received:
      {
          "jsonrpc": "2.0",
          "method": "kick/done",
          "params": [
              "/tmp/extra-dir-65308127230959/C.hs"
          ]
      }
      Use -p '/boot-def-test/' to rerun this test only.
```

eddiemundo:
Should fix #2677

I could not reproduce it locally, but I found in the logs that it is possible for `getHover` to eat a `ghcide/reference/ready` custom message. If that happens the `skipManyTill` `ghcide/references/ready` will be starving until timeout. Not sure why this happens more often (or only?) on the 9.0.2 ubuntu-latest tests.

The fix skips messages until the hover response or `ghcide/reference/ready`, and depending on which was received skips messages until receiving the other.

In all the other places I could find where something similar happens they either do `getHover` or `skipManyTill` `ghcide/references/ready` but not both, and so that's why we don't see these problems in other tests.

As a side note at first I didn't realize more than one ghcide process was being run so the logs were very confusing.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2686"></a>

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -5553,6 +5553,13 @@ bootTests = testGroup "boot"
         -- Dirty the cache
         liftIO $ runInDir dir $ do
             cDoc <- createDoc cPath "haskell" cSource
+            -- We send a hover request then wait for either the hover response or
+            -- `ghcide/reference/ready` notification.
+            -- Once we receive one of the above, we wait for the other that we
+            -- haven't received yet.
+            -- If we don't wait for the `ready` notification it is possible 
+            -- that the `getDefinitions` request/response in the outer ghcide 
+            -- session will find no definitions.
             let hoverParams = HoverParams cDoc (Position 4 3) Nothing
             hoverRequestId <- sendRequest STextDocumentHover hoverParams
             let parseReadyMessage = satisfy $ \case
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -5553,7 +5553,17 @@ bootTests = testGroup "boot"
         -- Dirty the cache
         liftIO $ runInDir dir $ do
             cDoc <- createDoc cPath "haskell" cSource
-            _ <- getHover cDoc $ Position 4 3
+            let hoverParams = HoverParams cDoc (Position 4 3) Nothing
+            hoverRequestId <- sendRequest STextDocumentHover hoverParams
+            inBetweenMsgs <- manyTill anyMessage (responseForId STextDocumentHover hoverRequestId)
+            let isReadyMsg = \case
+                  FromServerMess (SCustomMethod "ghcide/reference/ready") (NotMess NotificationMessage{_params = params}) ->
+                    case fromJSON params of
+                      A.Success fp | equalFilePath fp cPath -> Just ()
+                      _ -> Nothing
+                  _ -> Nothing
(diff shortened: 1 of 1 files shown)


pepeiborra:
> Should fix #2677
> 
> I could not reproduce it locally, but I found in the logs that it is possible for `getHover` to eat a `ghcide/reference/ready` custom message. If that happens the `skipManyTill` `ghcide/references/ready` will be starving until timeout. Not sure why this happens more often (or only?) on the 9.0.2 ubuntu-latest tests.

Good catch, `getHover` (and all other lsp-test request-response primitives) definitely eat messages.

ghcide/test/exe/Main.hs
```diff
@@ -5553,12 +5553,17 @@ bootTests = testGroup "boot"
...
-                _ -> Nothing
+            let hoverParams = HoverParams cDoc (Position 4 3) Nothing
+            hoverRequestId <- sendRequest STextDocumentHover hoverParams
+            inBetweenMsgs <- manyTill anyMessage (responseForId STextDocumentHover hoverRequestId)
+            let isReadyMsg = \case
+                  FromServerMess (SCustomMethod "ghcide/reference/ready") (NotMess NotificationMessage{_params = params}) ->
+                    case fromJSON params of
+                      A.Success fp | equalFilePath fp cPath -> Just ()
+                      _ -> Nothing
+                  _ -> Nothing
```
pepeiborra, line 5564: The idiomatic thing here is to skip all messages except for `Either` the hover response or the reference ready notification.

pepeiborra (approved):
Thank you!!!

michaelpj:
Can we get a comment in the code? It's a bit non-obvious what's going on there for future people.

Is there something systematic we should be doing to make this happen less often? Should we alter `getHover` & friends somehow?

eddiemundo:
> Can we get a comment in the code? It's a bit non-obvious what's going on there for future people.
 
Yeah, I'll add a comment.

> Is there something systematic we should be doing to make this happen less often? Should we alter `getHover` & friends somehow?

The `getX` then `skipUntilY` pattern has caused some flaky tests in the past (and this one) because I guess we aren't fully aware of the possible orders messages can go to the client.

Something that might be better in some cases is a `waitForMsgSet` that can return the set of messages in the order specified, but where the order of the set doesn't matter.

So maybe
```haskell
waitForMsgSet :: NonEmptyList (FromServerMessage -> Bool) -> Session (NonEmptyList FromServerMessage)
```
and a bunch of predicates defined for common types like the hover response, definition response etc. The idea is that it will wait for all of these predicates to be true in any order, and return messages corresponding to each predicate in the same order as the predicates were listed.

Not sure if this is idiomatic regarding the parser type, maybe there is a higher level of doing the same thing, but I think something like this would be convenient.

michaelpj:
I don't really know enough about how the tests are written to say, but I encourage you to keep thinking about it!

Issue #2414, opened 2021-11-29
Minor "import action" formatting bug


googleson78:
### HLS `--version`
```sh
tauren :: /tmp » haskell-language-server-wrapper --version
haskell-language-server version: 1.5.0.0 (GHC: 8.10.7) (PATH: /home/googleson78/.ghcup/bin/haskell-language-server-wrapper-1.5.0) (GIT hash: 311107eabbf0537e0c192b2c377d282505b4eff1)
```
### Your environment
Output of `haskell-language-server --probe-tools` or `haskell-language-server-wrapper --probe-tools`:

```sh
tauren :: /tmp » haskell-language-server-wrapper --probe-tools
haskell-language-server version: 1.5.0.0 (GHC: 8.10.7) (PATH: /home/googleson78/.ghcup/bin/haskell-language-server-wrapper-1.5.0) (GIT hash: 311107eabbf0537e0c192b2c377d282505b4eff1)
Tool versions found on the $PATH
cabal:          3.6.2.0
stack:          2.7.3
ghc:            8.10.7
```
Which OS do you use:
Debian

Which lsp-client do you use:
nvim, LanguageClient-neovim

Describe your project:
Reproduces with a standalone file and in a stack project

### Steps to reproduce

Try to use the import action for `Int64` with these file contents.
```haskell
module Asdf
  (f)
  where

f :: Int64 -> Int64
f = id
```
### Expected behaviour
```haskell
module Asdf
  (f)
  where

import Data.Int (Int64)

f :: Int64 -> Int64
f = id
```
(or some other valid result)
### Actual behaviour
```haskell
module Lol
  (f)
import Data.Int (Int64)
  where

f :: Int64 -> Int64
f = id
```

### Include debug information

<details>
<summary>
Debug output:
</summary>

```
2021-11-29 16:18:45.503918081 [ThreadId 4] INFO hls:	Starting LSP server...
2021-11-29 16:18:45.504138425 [ThreadId 4] INFO hls:	If you are seeing this in a terminal, you probably should have run WITHOUT the --lsp option!
2021-11-29 16:18:45.505988303 [ThreadId 9] INFO hls:	Started LSP server in 0.00s
2021-11-29 16:18:45.509847454 [ThreadId 9] DEBUG hls:	setInitialDynFlags cradle: Cradle {cradleRootDir = "/tmp", cradleOptsProg = CradleAction: Default}
[...]

PR #2694, opened 2022-02-06
rewrite import completion with ghc-exactprint

kokobd:
This will solve #2414.

TODO List:
- [x] handle the exact case of #2414 properly
- [ ] support other existing cases (make tests pass)
- [ ] support ghc 9.2
- [ ] minimize the `TextEdit` sent to language client

I will request code review when it's feature complete.

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -53,7 +53,8 @@ import           Development.IDE.GHC.Util                          (prettyPrint,
                                                                     traceAst,
                                                                     unsafePrintSDoc)
 import           Development.IDE.Plugin.CodeAction.Args
-import           Development.IDE.Plugin.CodeAction.ExactPrint
+import           Development.IDE.Plugin.CodeAction.ExactPrint      hiding
+                                                                   (newImport)
 import           Development.IDE.Plugin.CodeAction.PositionIndexed
 import           Development.IDE.Plugin.TypeLenses                 (suggestSignature)
 import           Development.IDE.Spans.Common
--- a/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
@@ -16,11 +16,13 @@ module Development.IDE.Plugin.CodeAction.ExactPrint (
   -- * Utilities
   appendConstraint,
   removeConstraint,
+  newImport,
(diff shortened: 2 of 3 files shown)


kokobd:
Both `Completions.hs` and `CodeAction.hs` has some code for adding new imports. I will try to eliminate this duplication.

kokobd:
If we want to know where to insert the new import declaration, we will have to exact print the whole `Located HsModule`, because the only `Located` structure containing a `Located ImportDecl` is the whole module.

Then I encountered a problem. In hls, the exactprint doesn't preserve lines starting with `#!`. As a quick test, you can try to extend import for this code (extending import already uses exactprint, both in master branch and latest release)
```haskell
import
#! nix-shell --pure -i runghc -p "haskellPackages.ghcWithPackages (hp: with hp; [ turtle ])"
    Data.Monoid ( Monoid )

class Semigroup a => SomeData a
instance SomeData All
```
If we place the cursor on `All` and select code action "Add All to the import list of Data.Monoid", then the code becomes
```haskell
import

    Data.Monoid ( Monoid, All )

class Semigroup a => SomeData a
instance SomeData All
```

I'm wondering if losing the `#!` line is intended and having some special considerations, or if it's just a previously unknown bug that need to be fixed.

pepeiborra:
> Then I encountered a problem (at least for ghc 8.10.7. I haven't tested it on ghc 9.2). In hls, the exactprint doesn't preserve lines starting with `#!`.

I believe ghc-exactprint comes with a round tripping test suite. Have you been able to repro the issue using it?

kokobd:
@pepeiborra I've tried `ghc-exactprint` 0.6.4 in a standalone project. It exactprints perfectly. This problem seems to be related to hls (maybe we have some wrong configs for parsing).

I'm still investigating and trying to fix it in hls. However as changing fundamental parts like parsing might potentially affect other parts of hls, any related context information is welcome ;)

kokobd:
I might have identified the root cause of the shebang exact printing issue. Created another draft PR #2725 to fix it seperately.

pepeiborra:
@kokobd still around? Would be nice to land this

kokobd:
Sorry for being absent these days. It has taken me a few months to get rid of my tiresome job.

I will probably fix this in the next few days, but less aggressively. Using exact print info to print the whole module incurs too much risk (performance, compatibility, etc.) and a lot of code change. Instead, using exact print information to locate the `where` keyword of the module is an easier way to fix this issue.

kokobd:
Closing in favor of #2981.

PR #2697, opened 2022-02-08
fix: handle comma in extend import list with ghc 9.2

### LINKED ISSUE #2662: Add ctor to import list code action broken with ghc 9.2
Add to import list code action is broken with GHC 9.2, because it does not includes comma between symbols:

HLSAddImportMissingComa

### Your environment

Which OS do you use: Linux
Which LSP client (editor/plugin) do you use: Neovim + lsp_config
Describe your project (alternative: link to the project): standalone file

### Steps to reproduce

Test with the following file:

```haskell
import Data.List.NonEmpty (fromList)

a = case fromList [1,2,3] of
      a :| l -> "boo"
```

Error should complain about `:|` not being in scope, and a code action should propose to add `:|` to the import list of `Data.List.NonEmpty`. You can execute the code action.

### Expected behaviour

Code action should change the import list to something valid, such as `(fromList, NonEmpty((:|)))`.

### Actual behaviour

Executing the code action changes the import list from `(fromList)` to `(fromListNonEmpty((:|))` effectively missing the comma, generating a parse error.

guibou:
The comma annotation was missing.

Close #2662.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2697"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
@@ -49,7 +49,7 @@ import           GHC (AddEpAnn (..), AnnContext (..), AnnParen (..),
                       DeltaPos (SameLine), EpAnn (..), EpaLocation (EpaDelta),
                       IsUnicodeSyntax (NormalSyntax),
                       NameAdornment (NameParens), NameAnn (..), addAnns, ann, emptyComments,
-                      reAnnL, AnnList (..))
+                      reAnnL, AnnList (..), TrailingAnn (AddCommaAnn), addTrailingAnnToA)
 #endif
 import           Language.LSP.Types
 import Development.IDE.GHC.Util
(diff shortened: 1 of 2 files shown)


configuration-ghc-921.nix
```diff
@@ -26,7 +26,6 @@ let
           "-f-hlint"
           "-f-haddockComments"
           "-f-alternateNumberFormat"
-          "-f-eval"
```
guibou, line 29: That's a mistake.

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
```diff
@@ -490,11 +490,22 @@ extendImportViaParent df parent child (L l it@ImportDecl{..})
...
 #else
-          x :: LIE GhcPs = reLocA $ L l'' $ IEThingWith listAnn parentLIE NoIEWildcard [childLIE]
           listAnn = epAnn srcParent [AddEpAnn AnnOpenP (epl 1), AddEpAnn AnnCloseP (epl 0)]
+          x :: LIE GhcPs = reLocA $ L l'' $ IEThingWith listAnn parentLIE NoIEWildcard [childLIE]
+
+      x <- pure $ setEntryDP x (SameLine $ if (not (null pre)) then 1 else 0)
+
+      let
+
+          fixLast = if not (null pre) then first addComma else id
```
pepeiborra, line 503:
```suggestion
      let hasSibling = not (null pre)
      x <- pure $ setEntryDP x (SameLine $ if hasSibling then 1 else 0)

      let

          fixLast = if hasSibling then first addComma else id
```

pepeiborra: There is an opportunity to factor some logic out in the same module:

https://github.com/haskell/haskell-language-server/blob/7af4e587fe5696524821390dd381bd024a8ca14b/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs#L378-L381

pepeiborra (approved):
Thank you!!!

Issue #2692, opened 2022-02-05
none-cradle is not respected by HLS


tchoutri:
### Your environment

Which OS do you use: Fedora 34

HLS version: haskell-language-server version: 1.6.1.1 (GHC: 8.10.7)

Which LSP client (editor/plugin) do you use: Neovim's LSP client

Describe your project: https://github.com/tchoutri/pg-entity

### Steps to reproduce

run `haskell-language-server-wrapper --debug` in the project with the following `hie.yaml`

```yaml
cradle:                                                   
  multi:
    - path: "./example"                                   
      config:
        cradle:
          none:
    - path: "./"                                          
      config:
        cradle:
          cabal:
            - path: "./src"
              component: "lib:pg-entity"

            - path: "./test"
              component: "pg-entity:test:entity-test"
```

### Expected behaviour
HLS ignores the files located in `./example`

### Actual behaviour
HLS tries to process the files in `./example` and fails. Meanwhile, `hie-bios` has no trouble determining that these files have no cradle:

```
❯ hie-bios debug example/app/Main.hs       
No cradle
```

### Include debug information

```
2022-02-05 18:28:54.720122789 [ThreadId 383] DEBUG hls: hlint:getIdeas:file:NormalizedFilePath "/home/tchoutri/Projects/pg-entity/src/Database/PostgreSQL/Entity/Internal/Unsafe.hs"
2022-02-05 18:28:54.72018192 [ThreadId 385] DEBUG hls:  hlint:getIdeas:file:NormalizedFilePath "/home/tchoutri/Projects/pg-entity/test/Main.hs"
2022-02-05 18:28:54.720231304 [ThreadId 387] DEBUG hls: hlint:getIdeas:file:NormalizedFilePath "/home/tchoutri/Projects/pg-entity/src/Database/PostgreSQL/Entity/DBT.hs"
2022-02-05 18:28:54.720280258 [ThreadId 389] DEBUG hls: hlint:getIdeas:file:NormalizedFilePath "/home/tchoutri/Projects/pg-entity/src/Database/PostgreSQL/Entity.hs"  
2022-02-05 18:28:54.72033031 [ThreadId 391] DEBUG hls:  hlint:getIdeas:file:NormalizedFilePath "/home/tchoutri/Projects/pg-entity/src/Database/PostgreSQL/Entity/Internal/QQ.hs"
[...]

PR #2701, opened 2022-02-10
Print info message when ignoring a file due to a none cradle

ThomasCrevoisier:
:wave: Out of curiosity, I was investigating https://github.com/haskell/haskell-language-server/issues/2692.

From the debug logs, the `Left []` gives the impression that we failed but `none` cradles are correctly taken into account.

This is a small patch to makes things a bit explicit for the user.


ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -619,11 +621,15 @@ cradleToOptsAndLibDir logger cradle file = do
                 CradleFail err       -> return (Left [err])
                 -- For the None cradle perhaps we still want to report an Info
```
michaelpj, line 622: Now you're doing it, so you could get rid of these comments?

ThomasCrevoisier: Sure :+1: Fixed and squashed.

michaelpj (approved):

Issue #2663, opened 2022-01-31
Automate manual steps of the gitlab release


jneira:
The actual manual steps are:
1. download the six extra tarballs
2. download GitHub SHA256SUMS
3. compute the `sha256sum --tag haskell-language-server* >> SHA256SUMS`
4. upload the new six tarballs and replace SHA256SUMS in the GitHub release

so we have to find a way to

  * - [ ] download SHA256SUMS from GitHub to gitlab machine using GH API
  * - [ ] update it with all new checksums (maybe in one unique env which should access tarballs from all jobs)
  * - [ ] upload all artifacts to GitHub release using GH api (will need a GH Auth token in secrets)

cc @hasufell

hasufell:
https://stackoverflow.com/questions/37786539/how-to-upload-github-asset-file-using-curl

hasufell:
Storing secrets in gitlab seems to be a bit more annoying: https://gitlab.haskell.org/help/ci/secrets/index.md

PR #2707, opened 2022-02-12
Fix releasing

hasufell:
Tested here: https://github.com/hasufell/haskell-language-server/releases

----

The idea here is that we pull the archives from the gitlab API instead of implementing pushing. That means a pipeline with the same tag as the release will have to have succeeded fully on gitlab CI.

----

Related: https://github.com/haskell/haskell-language-server/issues/2663

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2707"></a>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,152 +1,27 @@
-name: Builds
+name: Release
 
 on:
   workflow_dispatch:
     inputs:
-      alpine:
-        description: 'Use alpine to build a fully static executable for linux'
+      version:
+        description: 'HLS version'
+        required: true
+        type: string
+      tag:
+        description: 'HLS tag'
         required: false
-        default: 'false'
+        type: string
+      project:
+        description: 'Gitlab project number'
+        type: number
+        required: false
+        default: 1180
   release:
     types: [created]
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
-    container: ${{ (github.event.inputs.alpine == 'true' && startsWith(matrix.os,'ubuntu') && 'alpine:3.12') || '' }}
-    defaults:
-      run:
-        shell: ${{ (startsWith(matrix.os,'windows') && 'bash') || 'sh' }}
-    strategy:
-      fail-fast: false
-      matrix:
-        ghc:
-          [ "9.2.1"
-          , "9.0.2"
-          , "9.0.1"
-          , "8.10.7"
-          , "8.10.6"
-          , "8.8.4"
-          , "8.6.5"
-          ]
-        os: [ "ubuntu-18.04"
-            , "macOS-latest"
-            , "windows-latest"
-            ]
-        cabal: ['3.6']
-
-    steps:
-    - name: Install system dependencies
-      if: github.event.inputs.alpine == 'true' && runner.os == 'Linux'
-      run: |
-        apk add --no-cache curl gcc g++ gmp-dev ncurses-dev libffi-dev make xz gzip tar perl git bash sudo binutils-gold
-        apk add --no-cache zlib zlib-dev zlib-static gmp gmp-dev ncurses-static
-    - uses: actions/checkout@v2
-
-    - name: Disable tests and bechmarks
-      run: |
-        echo "tests: false" >> cabal.project.local
-        echo "benchmarks: false" >> cabal.project.local
-
-    - uses: ./.github/actions/setup-build
-      with:
-        ghc: ${{ matrix.ghc }}
-        os:  ${{ runner.os }}
-
-    - name: (Windows) Platform specifics
-      if: runner.os == 'Windows'
-      run: |
-        echo "EXE_EXT=.exe" >> $GITHUB_ENV
-
-    - name: (Linux) Platform specifics
-      if: runner.os == 'Linux'
-      run: |
-        echo "CABAL_ARGS=--enable-executable-static --ghc-options=-split-sections" >> $GITHUB_ENV
-
-    - name: Build the server
-      # Try building it twice in case of flakey builds on Windows
-      run: |
-        cabal build exe:hls -O2 $CABAL_ARGS || cabal build exe:hls -O2 $CABAL_ARGS -j1
-
-    - name: Compress server binary
-      id: compress_server_binary
-      run: |
-        HLS_BUILD=$(find dist-newstyle \( -name 'hls' -o -name 'hls.exe' \) -type f)
-        HLS=haskell-language-server-${{ matrix.ghc }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -3,150 +3,25 @@ name: Builds
...
+      tag:
+        description: 'HLS tag'
         required: false
-        default: 'false'
+        type: string
+      project:
+        description: 'Gitlab project number'
+        type: number
+        required: false
+        default: 1178
```
hasufell, line 18: This well need to be changed to `1180`, which is the project id of https://gitlab.haskell.org/haskell/haskell-language-server

I set this to my fork for testing.

.github/workflows/build.yml
```diff
@@ -189,43 +64,85 @@ jobs:
...
+          if [ -n "${{ github.event.inputs.project }}" ] ; then
+            proj=${{ github.event.inputs.project }}
           else
-            gzip -d *.gz
+            proj=1178
+          fi
+
+          if [ -n "${{ github.event.release.tag_name }}" ] ; then
+            tag=${{ github.event.release.tag_name }}
+            ver=1.6.1.1
```
hasufell, line 81: This is also just set for testing... should be `ver=${tag/v/}`  as below

.github/workflows/build.yml
```diff
@@ -189,43 +64,85 @@ jobs:
...
-          if [[ "${{ matrix.target-os }}" == "Windows" ]]; then
-            7z x "*.zip"
-            rm *.zip
+          set -eux
+
+          if [ -n "${{ github.event.inputs.project }}" ] ; then
+            proj=${{ github.event.inputs.project }}
           else
-            gzip -d *.gz
+            proj=1178
```
hasufell, line 76: testing value

.github/workflows/build.yml
```diff
@@ -189,43 +64,85 @@ jobs:
...
+          ver: ${{ github.event.inputs.version }}
+          tag: ${{ github.event.inputs.tag }}
+          proj: ${{ github.event.inputs.project }}
 
       - name: Upload binaries tarball to the release
         if: ${{ github.event.release.upload_url != '' }}
-        uses: actions/upload-release-asset@v1.0.2
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: actions/github-script@v2
```
hasufell, line 124: Because `upload-release-asset` does not support wildcards

hasufell:
@jneira

jneira:
sorry dont have time to review the pr, surely some other maintainer(s) could do it

Anton-Latukha:
Currently reviewed it 2 times (had 2 passes). The second one - read the code.

I am not adept in AWK language.

Honestly - to understand PR better structurally (where what how goes) & run inference inside - the third & maybe fourth passes are needed.

Of course that it is a single commit with a big Bourne shell diff. I understand - subject matter makes making separate commits difficult, but it also becomes harder to review & internalize what was done precisely & the meaning of it. It is a long way of saying - to give a proper review - I need another pass. Seems pretty straightforward & understandable, though.

Seen that you've done testing during the work process.

I can review better & say/not say something more meaningful later in the day.

hasufell:
> I am not adept in AWK language.

That's just some foo that I might remove later, because the job names on gitlab differ from the tarball names (and we need both for download via API). That can be equalized, but requires another round of gitlab CI, which I wanted to skip for testing.

> Seems pretty straightforward & understandable, though.

Yeah, we just drop the static builds from github CI, pull all artifacts from gitlab and then upload them to the github release. We use this rather convoluted github `script` action, because the [upload-release-asset action](https://github.com/actions/upload-release-asset) is deprecated and doesn't support wildcards.

Anton-Latukha:
No. It is mine issue. I just as many was postponing learning AWK for years. At this time & level I probably need to know some AWK.

Yes, I understood that it is just reshuffling of fields, between probably job & the tarball names.

Anton-Latukha:
> We use this rather convoluted github script action, because the [upload-release-asset action](https://github.com/actions/upload-release-asset) is deprecated and doesn't support wildcards.

Thank you for complete clarification.

As [upload-release-asset action](https://github.com/actions/upload-release-asset) mentions was https://github.com/softprops/action-gh-release considered instead? It may be looked into after the PR merge.

Anton-Latukha:
The link to the resulting workflow run is: https://github.com/hasufell/haskell-language-server/runs/5169934711?check_suite_focus=true

.github/workflows/build.yml
```diff
@@ -189,58 +64,85 @@ jobs:
...
-          # move the binaries for the specific platform into bin/
-          mkdir bin
-          mv haskell-language-server-${{ matrix.target-os }}-*/* bin
-          mv haskell-language-server-wrapper-${{ matrix.target-os }}.*/* bin
-          # decompress them
-          cd bin
-          if [[ "${{ matrix.target-os }}" == "Windows" ]]; then
-            7z x "*.zip"
-            rm *.zip
+          set -eux
```
Anton-Latukha, line 71: Oh.

Thank you for setting the `x` here, great idea.

.github/workflows/build.yml
```diff
@@ -189,58 +64,85 @@ jobs:
...
+            for (let file of await fs.readdir('.')) {
+              console.log('uploading', file);
+
+              await github.repos.uploadReleaseAsset({
+                owner, repo,
+                release_id: ${{ github.event.release.id }},
+                name: file,
+                data: await fs.readFile(`./${file}`)
+              });
+            }
```
Anton-Latukha, line 140: Probably some mostly copy-paste JS script to replace the cause.

Anton-Latukha (approved):
Reviewed.

The situation & solution is understandable and straightforward.

Only had a couple of the most minimal improvements to the code, but the code is overall of high quality.

The use of the `set -x` in the shell script would help when some naming changes or API changes. I think for the prolonged shell parts it is a good idea to keep the `x` flag on.

With that https://github.com/hasufell/haskell-language-server/runs/5169934711?check_suite_focus=true works & artifacts are uploaded - I think my review is enough to mandate the merge.

Anton-Latukha:
[skip circleci]

Anton-Latukha:
@hasufell, should we report to evaluate https://github.com/softprops/action-gh-release for later?

hasufell:
> @hasufell, should we report to evaluate use of https://github.com/softprops/action-gh-release later, or it was considered/does not implement the requirements.

Afaiu that action creates a release, which we don't want. We create a release and trigger a build instead. I don't think it's worth investigating that action.

Anton-Latukha:
Thank you @hasufell for this.

Issue #2672, opened 2022-02-01
1.6.1.0 cli execution doesn't work


googleson78:
### Your environment

Which OS do you use:
Debian testing
Which LSP client (editor/plugin) do you use:
Raw invocation
Describe your project (alternative: link to the project):
https://github.com/polysemy-research/polysemy

HLS 1.6.1.0 from ghcup
```sh
> haskell-language-server-wrapper --version
haskell-language-server version: 1.6.1.0 (GHC: 8.10.7) (PATH: /home/googleson78/.ghcup/bin/haskell-language-server-wrapper-1.6.1.0) (GIT hash: f4022c5bb8530cd306c53b941878244bf27a5d41)
```

### Steps to reproduce
Clone the project and run `haskell-language-server-wrapper`

### Expected behaviour
Loads the project.

### Actual behaviour
Fails with multiple instances of something like
```
  Maybe.fromJust: Nothing
  CallStack (from HasCallStack):
  error, called at libraries/base/Data/Maybe.hs:148:21 in base:Data.Maybe
  fromJust, called at src/Development/IDE/Core/Rules.hs:842:51 in
  ghcide-1.6.0.0-inplace:Development.IDE.Core.Rules
```

### Include debug information
[hls.log](https://github.com/haskell/haskell-language-server/files/7977592/hls.log)
[hls-stderr.log](https://github.com/haskell/haskell-language-server/files/7977599/hls-stderr.log)

fendor:
I can reproduce it on the cli as described, but in VSCode everything seems to be fine 0.o.

googleson78:
Has calling the cli perhaps gotten more complicated in some way? (some env needs to be provided etc)

jneira:
@googleson78 does it work for you in the editor?

googleson78:
Yup, thanks. I didn't even try it in an editor when I saw the cli invocation doesn't work tbh. 

I rely on the cli invocation to provide debug info and to also to troubleshoot issues if something isn't working from inside my editor. What do I need to do to be able to use it?

fendor:
It should work, so there is at least one bug somewhere. 

I can trigger the same error on the hie-bios code-base, but everything *seems* to load fine, so this is likely to be a bug in the cli loader.

solomon-b:
FYI I also am hitting this bug and am not using polysemy:

```
Hidden:   no
Range:    1:1-2:1
Source:   compiler
Severity: DsError
Message: 
[...]

PR #2719, opened 2022-02-18
Fix crash for non-LSP modes wrt #2627

hasufell:
https://github.com/haskell/haskell-language-server/issues/2672#issuecomment-1044260987


fendor (approved):
LGTM

michaelpj (approved):

Issue #1835, opened 2021-05-15
hls-test-utils-1.0.0.0 doesn't build with hspec-core 2.8


felixonmars:
### Your environment
Which OS do you use:
Arch Linux x86_64

### Steps to reproduce
Build hls-test-utils 1.0.0.0 with hspec-core 2.8

### Actual behaviour
```
Configuring hls-test-utils-1.0.0.0...
Preprocessing library for hls-test-utils-1.0.0.0..
Building library for hls-test-utils-1.0.0.0..
[1 of 2] Compiling Test.Hls.Util    ( src/Test/Hls/Util.hs, dist/build/Test/Hls/Util.dyn_o )

src/Test/Hls/Util.hs:188:32: error:
    ‘configOutputFile’ is not a (visible) constructor field name
    |
188 |                              , configOutputFile = Right $ subdir </> "results.xml"
    |                                ^^^^^^^^^^^^^^^^
```

It was removed in https://github.com/hspec/hspec/commit/86ad5be451d849e7711c342374798bfdfcbd4d33#diff-f0f01b0b2162c19117bc77a50a8986ec24c6553b1e09c83213cac5c6760b270f

jneira:
Thanks for point it out. So as first action we have to add an upper bound and make a hackage revision to make it buildable

berberman:
Afaik `getHspecFormattedConfig` is no longer used, and only tactics plugin is using `hspec` in test suite. We should clear out `hspec` related things from `hls-test-utils`.

jneira:
Agree, but the upper bound an hackage revision will be needed.

jneira:
Revision done: https://hackage.haskell.org/package/hls-test-utils-1.0.0.0/revision/1.cabal

PR #2721, opened 2022-02-18
Remove `getHspecFormattedConfig` which is no longer used

hololeap:
Remove hspec dependency

Closes: https://github.com/haskell/haskell-language-server/issues/1835

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2721"></a>


pepeiborra (approved):
Thank you very much. I will take one fewer dependency any day!

hololeap:
I should have switched back to a draft when I realized I could remove `hspec` instead of bumping it.

PR #2733, opened 2022-02-22
Fix hls-class-plugin on ghc-9.2

July541:
Let me play with ghc-9.x :)

- [x] Fix codeAction
- [x] Fix edit command

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2733"></a>


pepeiborra:
Is this ready for review?

July541:
> Is this ready for review?

Not yet.

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -38,6 +38,12 @@ import           Language.LSP.Server
 import           Language.LSP.Types
 import qualified Language.LSP.Types.Lens                 as J
 
+#if MIN_VERSION_ghc(9,2,0)
+import           GHC.Data.Maybe (rightToMaybe)
```
pepeiborra, line 42: I'd rather avoid this import and instead write `either (const Nothing) Just`

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -63,25 +69,69 @@ addMethodPlaceholders state AddMinimalMethodsParams{..} = do
...
+        addWhere (InstD xInstD (ClsInstD ext decl@ClsInstDecl{..})) =
+          let ((EpAnn entry anns comments), key) = cid_ext
+          in InstD xInstD (ClsInstD ext decl {
+            cid_ext = (EpAnn
+                        entry
+                        (AddEpAnn AnnWhere (EpaDelta (SameLine 1) []) : anns)
+                        comments
+                      , key)
+          })
+        addWhere decl = decl
```
pepeiborra, line 119: can you add a brief comment explaining what this code does? It's kinda hard to read and therefore maintain

July541: Explained the effect of the function and the trace of the parameter.

plugins/hls-class-plugin/test/Main.hs
```diff
@@ -1,8 +1,10 @@
+{-# LANGUAGE CPP               #-}
```
Ailrun, line 1: Why do we need CPP here?

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -168,13 +217,17 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
         HAR {hieAst = hf} ->
           pure
             $ head . head
+#if MIN_VERSION_ghc(9,2,0)
+            $ pointCommand hf (fromJust (fromCurrentRange pmap range) ^. J.start & J.character +~ 1)
```
Ailrun, line 221: Could you leave some comment on this difference? I mean `+~ 1` and `-~ 1`.

plugins/hls-class-plugin/test/Main.hs
July541: Removed:)

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
July541: I don't know, it just works after `+~ 1` and `-~ 1`.  The `-~ 1` is introduced by the first commit, I even don't know why it is required:(

As reference:
for file
```haskell
module T1 where

data X = X

instance Eq X where

```
The diag gives `Just (Range {_start = Position {_line = 4, _character = 9}, _end = Position {_line = 4, _character = 13}})` for both ghc-8.10.7 and ghc-9.2.1, then `pointCommand` run with the position `Position {_line = 4, _character = 10}` (ghc-9.2.1) and the position ` Position {_line = 4, _character =  8}`(ghc-8.10.7).

This means, on ghc-8.10.7, we search the position at the space before `Eq X`, and on ghc-9.2.1, we search the position at the character `q`.

July541:
ghc9.0 installation failed, can anyone please restart that?

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
Ailrun: @July541 `-~ 1` is to make sure that the target contains `Eq` by cancelling https://github.com/haskell/haskell-language-server/blob/e8951c9152dbbd5a0a41b280ad76fc9e952164ab/ghcide/src/Development/IDE/Spans/AtPoint.hs#L414

I'm just not sure how `+~ 1` works

July541: @Ailrun Hello, I investigated the whole pipe:

1. (`+~ 1`) and (`-~ 1`) are totally different aspects, for (`-~ 1`), the [`TyThing`](https://hackage.haskell.org/package/ghc-9.2.1/docs/GHC-Types-TyThing.html#t:TyThing) we match is `AConLike` https://github.com/haskell/haskell-language-server/blob/ecec5396037cfd87da6f79091391d0ca2a194d0c/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs#L231-L233. for (`+~ 1`) we match `ATyCon` https://github.com/haskell/haskell-language-server/blob/ef218f579bcea6dbdc571220c8388fcda6e6db87/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs#L195. This means if we match `ATyCon`, both ghc8 and ghc9 are workable.
2. The real problem makes `codeAction` fail is the extra identifiers, https://github.com/haskell/haskell-language-server/blob/ecec5396037cfd87da6f79091391d0ca2a194d0c/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs#L218-L224 Take `(Eq a) => Eq (X a)` as example, the  `pointCommand` result get `Eq` before, and in ghc9 it returns both `Eq` and `a`, which makes `head . head` fail.

July541:
hls-rename-plugin failed unexpectedly

July541:
> hls-rename-plugin failed unexpectedly

Once more:(

Ailrun (approved):
LGTM :)

PR #2735, opened 2022-02-23
Track file versions accurately.

wz1000:
This patch does two things:

1. It allows us to track the versions of `Values` which don't come from the VFS, as long as those
   particular `Values` depended on the `GetModificationTime` rule
   This is necessary for the recompilation avoidance scheme implemented in #2316

2. It removes the VFSHandle type and instead relies on snapshots of the VFS state taken on every rebuild
   of the shake session to ensure that we see a consistent VFS state throughout each individual build.

With regards to 2, this is necessary because the lsp library mutates its VFS file store as changes come
in. This can lead to scenarios where the HLS build session can see inconsistent views of the VFS.
One such scenario is.

1. HLS build starts, with VFS state A
2. LSP Change notification comes in and lsp updates its internal VFS state to B
3. HLS build continues, now consulting VFS state B
4. lsp calls the HLS file change handler, interrupting the build and restarting it.
   However, the build might have completed, or cached results computed using an
   inconsistent VFS state.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2735"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -58,6 +58,7 @@ module Development.IDE.Core.Shake(
     setPriority,
     ideLogger,
     actionLogger,
+    getVirtualFile,
     FileVersion(..),
     Priority(..),
     updatePositionMapping,
@@ -73,7 +74,6 @@ module Development.IDE.Core.Shake(
     IndexQueue,
     HieDb,
     HieDbWriter(..),
-    VFSHandle(..),
     addPersistentRule,
     garbageCollectDirtyKeys,
     garbageCollectDirtyKeysOlderThan,
@@ -94,7 +94,6 @@ import qualified Data.HashMap.Strict                    as HMap
 import           Data.Hashable
 import           Data.List.Extra                        (foldl', partition,
                                                          takeEnd)
-import           Data.Map.Strict                        (Map)
 import qualified Data.Map.Strict                        as Map
 import           Data.Maybe
 import qualified Data.SortedList                        as SL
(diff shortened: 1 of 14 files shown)


ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1088,24 +1105,26 @@ defineEarlyCutoff' doDiagnostics cmp key file old mode action = do
         res <- case val of
             Just res -> return res
             Nothing -> do
+                staleV <- liftIO $ atomicallyNamed "define -read 3" $ getValues state key file <&> \case
+                  Nothing -> Failed False
+                  Just (Succeeded ver v, _) -> Stale Nothing ver v
+                  Just (Stale d ver v, _)   -> Stale d ver v
+                  Just (Failed b, _)        -> Failed b
```
pepeiborra, line 1112: We don't need to compute `staleV` when the rule runs successfully, right?

pepeiborra (approved):

ghcide/src/Development/IDE/Core/Shake.hs
wz1000: True, but together with https://github.com/haskell/haskell-language-server/pull/2316/commits/3273b87205b011f7acccc1c7a534d72d481054f1 in #2316, we will need to compute it in all cases. We can close this PR and merge as a batch with #2316 if that is any better?

pepeiborra: I don't see any problem with changing it again in #2316

michaelpj (commented):
Few comments, some might be worth addressing.

Issue #2716, opened 2022-02-16
complex type variables in hover with GHC 9.2


guibou:
Type signature in hover with GHC 9.2 are longer what they used to be.

See for example:

![image](https://user-images.githubusercontent.com/9705357/154348988-2f4a62f9-bed2-4fc8-8f16-c0ebb0442676.png)

See the `a_aI0wB`. With GHC 8.10.7, it was `a`:

![image](https://user-images.githubusercontent.com/9705357/154349226-d671835d-bf8d-4212-8392-8917b2bfd855.png)

### Your environment

Which OS do you use:
Linux
Which LSP client (editor/plugin) do you use:
neovim + lsp-config
Describe your project (alternative: link to the project):

Huge proprietary codebase built with bazel with a custom HLS hie setup.

### Steps to reproduce

Use hover on a type. However that's not always clear, because sometimes I can observe "short" type variable. In my example, it appen on `class` and `newtype` which have type parameters.

### Expected behaviour

Type variable should be short. They were used to be named as the type variable in the associated `newtype` or `class`.

### Actual behaviour

Type variables are long and complex. The starts with the original name, but a noisy suffix is appended.

michaelpj:
It seems likely that something changed in GHC upstream.

hsyl20:
The following code looks incorrect:
https://github.com/haskell/haskell-language-server/blob/bb73e32b54a8b195b289b17cc6ef6901ad298e2f/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs#L68 

Because of https://gitlab.haskell.org/ghc/ghc/-/commit/8d401e50ce35525dcdc6a21fcf952c3cbf3a72d4, a name is always printed with its unique except with "user" style, but `defaultSDocContext` defaults to use "dump" style. 

Something like this should fix it:

```haskell
renderWithContext (defaultSDocContext { sdocSuppressUniques = True, sdocStyle = defaultUserStyle }) . ppr
```

Kleidukos:
Well, that's a golden test that's just waiting to get written. :)

michaelpj:
I wrote a failing test, PR incoming.

> The following code looks incorrect...

Unfortunately, that's not it. I think the real issue is that we *pervasively* use `showSDocUnsafe`, re-exported from `ghc`, which prints uniques. I don't really know what to do about this, we have dozens of wrappers around various `SDoc` printing functions, it's somewhat horrifying.

PR #2742, opened 2022-03-01
Add failing test for variables in hovers

michaelpj:
Witnesses #2716.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2742"></a>


pepeiborra (approved):

ghcide/test/exe/Main.hs
```diff
@@ -4165,6 +4166,11 @@ findDefinitionAndHoverTests = let
   , test  no     yes    outL45     outSig        "top-level signature              #767"
   , test  broken broken innL48     innSig        "inner     signature              #767"
   , test  no     yes    holeL60    hleInfo       "hole without internal name       #831"
+  , if ghcVersion >= GHC92 then
+        -- Broken on GHC 9.2 and above due to printing of uniques
+        test  no     yes    holeL65    []        "hole with variable"
```
pepeiborra, line 4171: I cannot remember what these parameters mean, but at least the 2nd or 3rd should be `broken` right?

michaelpj: `broken` doesn't work there. I don't really understand how these work, so I just put an empty set of expectations...

PR #2750, opened 2022-03-05
LSP window message log recorder

pepeiborra:
This PR adds a new `Recorder` that emits LSP `WindowShowMessage` notifications, and then uses it to surface errors to the user. 

image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2750"></a>

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -16,7 +16,7 @@ import           Ide.Arguments                (Arguments (..),
 import           Ide.Main                     (defaultMain)
 import qualified Ide.Main                     as IdeMain
 import qualified Plugins
-import           Prettyprinter                (Pretty (pretty))
+import           Prettyprinter                (Pretty (pretty), vcat)
 import Development.IDE.Plugin.LSPWindowShowMessageRecorder (makeLspShowMessageRecorder)
 import Data.Text (Text)
 import Ide.PluginUtils (pluginDescToIdePlugins)
@@ -59,4 +59,10 @@ main = do
       defaultMain (cmapWithPrio LogIdeMain recorder) args (pluginDescToIdePlugins [lspRecorderPlugin] <> plugins)
 
 renderDoc :: Doc a -> Text
-renderDoc = renderStrict . layoutPretty defaultLayoutOptions
+renderDoc d = renderStrict $ layoutPretty defaultLayoutOptions $ vcat
+    ["Unhandled exception, please check your setup and/or the [issue tracker](" <> issueTrackerUrl <> "): "
+    ,d
+    ]
+
+issueTrackerUrl :: Doc a
+issueTrackerUrl = "https://github.com/haskell/haskell-language-server/issues"
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -5,11 +5,11 @@
 module Main(main) where
 
 import           Data.Function                ((&))
-import           Development.IDE.Types.Logger (Priority (Debug, Info),
+import           Development.IDE.Types.Logger (Priority (Debug, Info, Error),
                                                WithPriority (WithPriority, priority),
                                                cfilter, cmapWithPrio,
                                                makeDefaultStderrRecorder,
-                                               withDefaultRecorder)
+                                               withDefaultRecorder, renderStrict, layoutPretty, defaultLayoutOptions, Doc)
 import           Ide.Arguments                (Arguments (..),
                                                GhcideArguments (..),
                                                getArguments)
@@ -17,6 +17,9 @@ import           Ide.Main                     (defaultMain)
 import qualified Ide.Main                     as IdeMain
 import qualified Plugins
 import           Prettyprinter                (Pretty (pretty))
+import Development.IDE.Plugin.LSPWindowShowMessageRecorder (makeLspShowMessageRecorder)
+import Data.Text (Text)
+import Ide.PluginUtils (pluginDescToIdePlugins)
 
 data Log
   = LogIdeMain IdeMain.Log
@@ -33,6 +36,7 @@ main = do
     -- parser to get logging arguments first or do more complicated things
     pluginCliRecorder <- cmapWithPrio pretty <$> makeDefaultStderrRecorder Nothing Info
     args <- getArguments "haskell-language-server" (Plugins.idePlugins (cmapWithPrio LogPlugins pluginCliRecorder) False)
+    (lspRecorder, lspRecorderPlugin) <- makeLspShowMessageRecorder
 
     let (minPriority, logFilePath, includeExamplePlugins) =
           case args of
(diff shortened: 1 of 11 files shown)


eddiemundo (approved):
So this is what you meant by using a plugin to get the env... nice. Additionally, the issue of how to choose which specific messages to send to window is sidestepped by sending all the ones with Error.

The `ghcide/exe` version doesn't have the issue tracker url message but that's fine since most people will be using hls rather than ghcide.

There are a few other places that use SWindowMessage to add to the logging TODO list:
splice plugin
tactics plugin
fourmolu plugin
retrie plugin
extendImportHandler plugin in Completions.hs
displayTHWarning in Rules.hs

Sorry for not being too active, been distracted recently.

pepeiborra:
> There are a few other places that use SWindowMessage to add to the logging TODO list: splice plugin tactics plugin fourmolu plugin retrie plugin extendImportHandler plugin in Completions.hs displayTHWarning in Rules.hs

Thanks. I'll take a look at those and see if they can be changed to log errors meaningfully.

pepeiborra:
>  Additionally, the issue of how to choose which specific messages to send to window is sidestepped by sending all the ones with Error.

Yes - this is a very coarse way of selecting what messages get shown. 

The finer way to do this (and what we do in Sigma IDE where we have lots of telemetry) is to do a case analysis on the entire log type to filter exactly what things we want to record (and how).

exe/Main.hs
```diff
@@ -42,9 +46,23 @@ main = do
...
+                & cmapWithPrio renderDoc
+            ]
+        plugins = Plugins.idePlugins (cmapWithPrio LogPlugins recorder) includeExamplePlugins
 
-      defaultMain (cmapWithPrio LogIdeMain recorder) args (Plugins.idePlugins (cmapWithPrio LogPlugins recorder) includeExamplePlugins)
+      defaultMain (cmapWithPrio LogIdeMain recorder) args (pluginDescToIdePlugins [lspRecorderPlugin] <> plugins)
+
+renderDoc :: Doc a -> Text
+renderDoc d = renderStrict $ layoutPretty defaultLayoutOptions $ vcat
+    ["Unhandled exception, please check your setup and/or the [issue tracker](" <> issueTrackerUrl <> "): "
```
michaelpj, line 63: It's not necessarily an exception, right? It could be any other error.

ghcide/src/Development/IDE/Plugin/LSPWindowShowMessageRecorder.hs
```diff
@@ -0,0 +1,58 @@
...
+makeLspShowMessageRecorder = do
+  envRef <- newIORef Nothing
+  -- messages logged before the LSP stream is initialized will be sent when it is
+  backLogRef <- newIORef []
+  let recorder = Recorder $ \it -> do
+        mbenv <- liftIO $ readIORef envRef
+        case mbenv of
+          Nothing -> liftIO $ atomicModifyIORef'_ backLogRef (it :)
+          Just env -> sendMsg env it
+      -- the plugin captures the language context, so it can be used to send messages
```
michaelpj, line 30: I don't really like this approach. It's extremely indirect. My branch just has a function `LanguageContextEnv config -> Recorder ...` and then we have to pass it along until we reach `initialize`.

michaelpj: I think we could relatively easily adapt this to just pass the `IORef` to be filled in rather than going via a plugin...

ghcide/src/Development/IDE/Plugin/LSPWindowShowMessageRecorder.hs
```diff
@@ -0,0 +1,58 @@
...
+import qualified Language.LSP.Server as LSP
+import Language.LSP.Types (MessageType (..), SMethod (SInitialized, SWindowShowMessage), ShowMessageParams (..))
+
+-- | Creates a recorder that logs to the LSP stream via WindowShowMessage notifications.
+--   The recorder won't attempt to send messages until the LSP stream is initialized.
+makeLspShowMessageRecorder ::
+  IO (Recorder (WithPriority Text), PluginDescriptor c)
+makeLspShowMessageRecorder = do
+  envRef <- newIORef Nothing
+  -- messages logged before the LSP stream is initialized will be sent when it is
```
michaelpj, line 23: This is neat, my version just didn't record until it gets setup when the LSP server is ready.

ghcide/src/Development/IDE/Plugin/LSPWindowShowMessageRecorder.hs
```diff
@@ -0,0 +1,58 @@
...
+          { pluginNotificationHandlers = mkPluginNotificationHandler SInitialized $ \_ _ _ -> do
+              env <- getLspEnv
+              liftIO $ writeIORef envRef $ Just env
+              -- flush the backlog
+              backLog <- liftIO $ atomicModifyIORef' backLogRef ([],)
+              for_ (reverse backLog) $ sendMsg env
+          }
+  return (recorder, plugin)
+
+sendMsg :: MonadUnliftIO m => LanguageContextEnv config -> WithPriority Text -> m ()
```
michaelpj, line 42: You don't actually need the `MonadUnliftIO` constraint: just declare this function to run in `IO`, and call `liftIO` at the use site. Then you can get rid of all the other stuff.

michaelpj (commented):
I also have a branch that implements this, here's the commit: https://github.com/michaelpj/haskell-language-server/commit/57ac4a155df2d8f67226eade851990e3771a7f2d

The main difference is the way we get the `LanguageContextEnv` to the `Recorder`. I'm really not a fan of the plugin trick, I think we should just pass it in. I don't mind the `IORef` way, but I think we could just pass it down as an ordinary parameter instead of smuggling it in like this.

I also want to include a recorder that sends `window/logMessage` notifications, but I'll do that separately.

exe/Main.hs
pepeiborra: True, I copied this message over from an earlier exception site - will change to error

ghcide/src/Development/IDE/Plugin/LSPWindowShowMessageRecorder.hs
pepeiborra: Please send a PR! I suspect that the plugin solution is more modular, as it abstracts away the passing of the mutable state around.

exe/Main.hs
michaelpj: And indeed, since we send it as an error message, we don't need to tell the user again that it's an error. Including the issue link is maybe still hepful.

ghcide/src/Development/IDE/Plugin/LSPWindowShowMessageRecorder.hs
michaelpj: Yes, but it does so by using a mechanism that really isn't designed for that. Now the recorder mechanism is intertwined with the plugin mechanism, which it should have nothing to do with that. We shouldn't have to fix the loggers if we change how plugins work; that seems anti-modular to me.

michaelpj: I'll take a look.

ghcide/src/Development/IDE/Plugin/LSPWindowShowMessageRecorder.hs
pepeiborra: Good catch - I'll drop the instance before @wz1000 gets a chance to abuse it :)

PR #2752, opened 2022-03-05
Fix test failure for AlternateNumberFormat

drsooch:
This fixes alternate number format test-suite for 9.2.1 and enables it in github workflows. The plugin is already enabled in 9.2.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2752"></a>

### DIFF
--- a/plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
+++ b/plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
@@ -7,7 +7,6 @@ module Ide.Plugin.AlternateNumberFormat (descriptor, Log(..)) where
 import           Control.Lens                    ((^.))
 import           Control.Monad.Except            (ExceptT, MonadIO, liftIO)
 import qualified Data.HashMap.Strict             as HashMap
-import           Data.List                       (intercalate)
 import           Data.Text                       (Text)
 import qualified Data.Text                       as T
 import           Development.IDE                 (GetParsedModule (GetParsedModule),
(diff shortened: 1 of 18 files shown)


plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Literals.hs
```diff
@@ -13,7 +14,7 @@ import           Data.Maybe                    (maybeToList)
 import           Data.Text                     (Text)
 import qualified Data.Text                     as T
 import           Development.IDE.GHC.Compat    hiding (getSrcSpan)
-import           Development.IDE.GHC.Util      (unsafePrintSDoc)
+import           Development.IDE.GHC.Util      (prettyPrint, unsafePrintSDoc)
```
Ailrun, line 17: Why do we need this? I don't see any use of this function

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -7,6 +7,7 @@ module Ide.Plugin.AlternateNumberFormat (descriptor, Log(..)) where
 import           Control.Lens                    ((^.))
 import           Control.Monad.Except            (ExceptT, MonadIO, liftIO)
 import qualified Data.HashMap.Strict             as HashMap
+import           Data.List                       (intercalate)
```
Ailrun, line 10: Do we have a use of `interclate` in this file?

Ailrun (approved):
Thank you for the update! Here are some minor comments:

michaelpj:
Failures on 8.8, looks like maybe this should be using some helpers from `Compat`?

drsooch:
> Failures on 8.8, looks like maybe this should be using some helpers from `Compat`?

Yeah, looks like `LPat` is not actually located in 8.8. Fixed.

PR #2753, opened 2022-03-05
Estimate file versions safely

pepeiborra:
For a long time, `defineEarlyCutoff` has been accessing the `Values` store directly
to compute `GetModificationTime` values instead of calling `use`, breaking the
invariant. The values are used to associate the rule result to a file version,
which gets recorded in the `Value` as well as used as the key in the Diagnostics
store.

The problem here is that the `GetModificationTime` rule computes a new version and
mutates the Values store, so if `defineEarlyCutoff` peeks in the store before
`GetModificationTime` has run, it will grab the old version. This leads to lost
diagnostics and potentially to misversioned `Values`.

Fixing the problem is tricky, because we cannot simply use `GetModificationTime`
inside `defineEarlyCutoff` for all rules. There are three issues with this:
1. Creating a dependency on `GetModificationTime`: if everything depends on it,
then we lose the ability to do early cutoff
2. Creating cycles in the build graph, since `GetModificationTime` has
dependencies itself. Because hls-graph doesn't implement cycle detection (Shake
did), it is a nightmare to debug these cycles.
3. Creating overhead, since `GetModificationTime` calls the file system for non
FOIs and in the past this was very expensive for projects with large cartesian
product of module paths and source folders

To work around these I had to introduce a new hls-graph primitive,
`applyWithoutDependency`, as well as add a bunch more fragile type tests on the key
type to decide on whether to use `GetModificationTime` or peek into the values
store. The type casts could be cleaned up by introducing a type class, but I'm
not sure the end result would be any better.

To understand the issue and debug the implementation of the fix, I added a
number of opentelemety traces which I'm leaving in place in case they could be
useful in the future.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2753"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -1156,6 +1156,7 @@ defineEarlyCutoff' doDiagnostics cmp key file old mode action = do
         -> NormalizedFilePath
         -> Action (Maybe FileVersion)
     estimateFileVersionUnsafely state _k v fp
+        | fp == emptyFilePath = pure Nothing
         | Just Refl <- eqT @k @GetModificationTime = pure v
         -- GetModificationTime depends on these rules, so avoid creating a cycle
         | Just Refl <- eqT @k @AddWatchedFile = liftIO getFromStoreUnsafely
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -166,6 +166,7 @@ import qualified "list-t" ListT
 import           OpenTelemetry.Eventlog
 import qualified StmContainers.Map                      as STM
 import           System.FilePath                        hiding (makeRelative)
+import           System.IO.Unsafe (unsafePerformIO)
 import           System.Time.Extra
 
 data Log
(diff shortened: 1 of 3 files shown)


ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1144,117 +1143,78 @@ defineEarlyCutoff' doDiagnostics cmp key file old mode action = do
...
+         . IdeRule k v
+        => Values
+        -> k
+        -> Maybe v
+        -> NormalizedFilePath
+        -> Action (Maybe FileVersion)
+    estimateFileVersionUnsafely state _k v fp
+        | Just Refl <- eqT @k @GetModificationTime = pure v
+        -- GetModificationTime depends on these rules, so avoid creating a cycle
+        | Just Refl <- eqT @k @AddWatchedFile = liftIO getFromStoreUnsafely
```
wz1000, line 1161: How do these work? If `GetModificationTime` depends on these, how can we rely on the value in the store to be accurate?

pepeiborra: We can't get the current value because it doesn't exist yet, but it doesn't matter too much because these rules do not produce diagnostics

wz1000: Maybe we can return `Nothing` then?

pepeiborra: Yes, that's another option. I'm not sure what the implications would be, other than no position mapping information.

wz1000 (approved):

michaelpj:
It would be nice if the excellent description of the problem and the solution was recorded somewhere more durable!

PR #2754, opened 2022-03-06
Get rid of cabal-ghc90.project

pepeiborra:


fendor (approved):
LGTM, awesome that we can remove extra-configuration!

michaelpj:
I'm confused that this works: doesn't `stylish-haskell` still not support GHC 9? So won't the main configuration still not work on GHC 9? Evidently it does... I just don't understand how...

pepeiborra:
The GHC 9 build works, but it looks like the GHC 8.10 build breaks due to the `allow-newer`, sadly.

pepeiborra:
Stylish-haskell lacks support for `ghc-lib-parser-9.0`, and the lack of conditional constructs in `cabal.project` makes it impossible to work around this with a single project file:

1. `cabal build exe:haskell-language-server` fails even though `hls-stylish-haskell-plugin` is not a dependency, because the build plan still includes it since it's one of the packages in the project list
2. I don't know any way to remove or disable `hls-stylish-haskell-plugin` from the project list for GHC 9.0

PR #2758, opened 2022-03-06
Refactor LSP logger and log via window/logMessage also

michaelpj:
Also adds a logger that forwards on messages via `window/logMessage`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2758"></a>


ghcide/src/Development/IDE/Main.hs
```diff
@@ -289,9 +289,14 @@ testing recorder logger =
...
-defaultMain :: Recorder (WithPriority Log) -> Arguments -> IO ()
-defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats recorder) fun
+defaultMain
+    :: Recorder (WithPriority Log)
+    -> Maybe (MVar (LSP.LanguageContextEnv Config))
+    -- ^ Variable to be filled with the LSP environment, useful for tools that need this outside
+    -- the scope of runLanguageServer
+    -> Arguments
+    -> IO ()
+defaultMain recorder lspEnvVar Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats recorder) fun
```
pepeiborra, line 299: The `MVar` is an implementation detail, better hide it behind a callback, or a hook to ran an `Lsp` action on initialisation.

And please move it to the `Arguments` record, to keep the API simple and minimise the breakage

pepeiborra (changes requested):

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -290,3 +303,46 @@ textWithPriorityToText columns WithPriority{ priority, callStack_, payload } = d
...
+  backlog <- newTQueueIO
+  pure $ Recorder $ \it -> do
+        marg <- liftIO $ tryReadMVar argVar
+        case marg of
+          Nothing -> liftIO $ atomically $ writeTQueue backlog it
+          Just arg -> do
+              let recorder = recFun arg
+              toRecord <- liftIO $ atomically $ flushTQueue backlog
+              for_ toRecord (logger_ recorder)
+              logger_ recorder it
```
pepeiborra, line 322: So the backlog doesn't get flushed until the recorder gets called, right?  
This isn't good, specially if the recorder only gets called on error messages

ghcide/src/Development/IDE/Main.hs
michaelpj: I wasn't sure about whether to put it in `Arguments`, in some places there seems to be an assumption that it reflects CLI arguments. But yeah, it seems better in there.

ghcide/src/Development/IDE/Types/Logger.hs
michaelpj: Great point! All the more reason to have a callback: then I think we can indeed do this more nicely.

wz1000:
Please keep in mind what happens with the command line interface, where there is no LSP environment.

michaelpj:
> Please keep in mind what happens with the command line interface, where there is no LSP environment.

That should be fine, although it does suggest to me that we should use a bounded queue for the backlog: otherwise it will just accumulate logs forever because it can never send them.

ghcide/src/Development/IDE/Main.hs
```diff
@@ -232,6 +231,7 @@ data Arguments = Arguments
     , argsHandleIn              :: IO Handle
     , argsHandleOut             :: IO Handle
     , argsThreads               :: Maybe Natural
+    , argsLspEnvCb              :: LSP.LanguageContextEnv Config -> IO () -- ^ Callback to run once the LSP environment is setup
```
pepeiborra, line 234: This is a bit open ended - at what time is the callback run? Importantly, is it before or after LSP initialisation?

michaelpj: At the moment it only promises to be called when we have a `LanguageContextEnv`, but in fact it will be called during initialization. I agree it's not terribly precise, but I don't think representing it as a response to a LSP message is much more precise: it's not really a message handler!

ghcide/src/Development/IDE/Main.hs
michaelpj: I ended up putting it in the `Arguments` record for `ghcide` but not HLS: the HLS `Arguments` are very tightly coupled to argument parsing, expect to be `Show`-able etc, so it really didn't fit.

ghcide/src/Development/IDE/Main.hs
pepeiborra: It is important to be more precise because the LSP server is restricted in what it can do before initialisation. So to be safe, I think the callback should only be called **after** LSP initialisation is complete

michaelpj: You know what, maybe you were just right to begin with.
- A sensible time to do this _is_ when you get `Initialized`
- The way that we have to add notification handlers is plugins, so adding a plugin ad hoc like this is maybe not crazy

pepeiborra: I tend to agree 😊

michaelpj:
Okay, this is now a much more modest refactoring that still uses the plugin technique to install the handler, and adds logging via `window/logMessage`.

Also I officially give @pepeiborra 1 I-told-you-so point ;)

pepeiborra (approved):

PR #2761, opened 2022-03-07
Fix Change Type Signature Plugin test suite for 9.2.1

drsooch:
This should fix the test errors for 9.2.1 and adds new test cases specifically for 9.2.1.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2761"></a>


pepeiborra (approved):
Thanks!

PR #599, opened 2020-11-14
Fix fourmolu plugin inconsistent formatting

zweimach:
fix-fourmolu
~~**The Problem**: fourmolu plugin sets its indentation size from the editor (inferred from source file tab size or EditorConfig) when no `fourmolu.yaml` found. In other words: different formatting results between the plugin and the cli.~~

~~This little fix will prevent fourmolu from getting the editor's `tabSize`.~~
~~Hopefully, it will solve the problem.~~

**EDIT**:
> > And when I test it with fourmolu.yaml, it prioritizes the editor's indentation level over the config file even without EditorConfig.
>
> Ah, that is the wrong way round. Should be a one line fix.

 It fixes fourmolu plugin's use of FormattingOptions.

PR #691, opened 2020-12-21
Introduce generic config for plugins

alanz:
Make it possible to provide config for a plugin in a regular way, by
using a namespace in the json config space. So we have

```
haskell.plugin.hlint.globalOn
haskell.plugin.importLens.globalOn
```

It is also possible to have finer-grain config, so the individual
parts of a plugin can also be separately enabled/disabled.

Closes #513

PR #2763, opened 2022-03-07
Add an option to run Fourmolu via the CLI interface of a separate binary, rather than the bundled library

georgefst:
This makes it easier to set up an environment (e.g. a Nix shell) which uses the same version of Fourmolu in HLS as from the command line.

It can also be useful when Fourmolu has updated and we don't want to wait a month for an HLS release, or re-compile. Or when one wants to try out a development version of Fourmolu.

# To do
- [x] Add tests
  - ~~This will require a non-trivial refactor to parameterise `runSessionWithServerFormatterCLI` and `goldenWithHaskellDocFormatterCLI` over a `Config` arg, so I'd like some confirmation that I'm on the right track first.~~

# Future work
- Add config option to editor plugins e.g. VSCode
    - ~~We need to note that the option currently only affects Fourmolu~~
    - We should make clear that the binary needs to be on `PATH`
    - What should we call this option?
- Implement similar functionality for other formatters?
    - I'm personally happy to handle Ormolu, but I'm not very familiar with the others
    - @tfausak would seemingly be interested in this being implemented for Brittany: https://github.com/lspitzner/brittany/pull/356#issuecomment-955096825
- We _could_ add an option to specify the exact path to a binary, rather than requiring it to be in `PATH`. I'm not sure this would be useful though since we already have the `haskell.serverEnvironment` option. It would also make it harder to have a single option for all formatters. EDIT: is `haskell.serverEnvironment` specific to VSCode?

# Minor issues
- `lspPrinterOpts` is unused. I doubt anyone actually cares. We can't just pass `--indentation` as this would take priority over options from config files, which is not what we want (see #599).

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2763"></a>


michaelpj:
I wonder whether we really want this as a generic option. An alternative would be to just provide an additional formatting provider, say, `fourmolu-cli` which would call out to the CLI. That way we also don't end up with illegal states when people set that option but pick a formatting provider which doesn't support it.

e.g. what happens if you are trying out formatters and you do:
- Set provider to `fourmolu`
- Hmm, maybe I should use the CLI, set that option
- Hmm, maybe I'll try `stylish-haskell` (which doesn't support it)
- It breaks?

> This will require a non-trivial refactor to parameterise runSessionWithServerFormatterCLI and goldenWithHaskellDocFormatterCLI over a Config arg, so I'd like some confirmation that I'm on the right track first.

I think my suggestion would also solve that: there would simply be some tests for a different formatting provider.

> lspPrinterOpts is unused. I doubt anyone actually cares.

Maybe it doesn't matter that much in this case, but I imagine this sort of thing matters more for e.g. passing the right GHC extensions to tools that need to know that.

That's one advantage of my other proposal to do `haskell-language-server format foo`: you can be sure to use actually the same options as would be provided by HLS.

> We could add an option to specify the exact path to a binary, rather than requiring it to be in PATH. I'm not sure this would be useful though since we already have the haskell.serverEnvironment option.

Not all clients are VSCode. Putting it on the PATH is not onerous, but if having an option to give the path is cheap, I think it's nice. Again, no problem since I'm suggesting not having a single option ;)

georgefst:
Thanks @michaelpj. Agreed on all points. I'll implement it as a separate plugin.

> > lspPrinterOpts is unused. I doubt anyone actually cares.
> 
> Maybe it doesn't matter that much in this case, but I imagine this sort of thing matters more for e.g. passing the right GHC extensions to tools that need to know that.

It's fine. GHC options etc. come via `fileOpts`. `lspPrinterOpts` only deals with tab size set by the editor itself, outside of HLS.

I may have to think about factoring out `convertDynFlags` somewhere (where?), rather than duplicating it in a third place (it originally came from the Ormolu plugin.

georgefst:
Actually, I think we'd get most of the benefits, along with easier maintenance, by just making it a plugin-specific option. I'd completely missed that we now have #691.

georgefst:
Right. I think I'm using `Properties` correctly, but I can't seem to affect any change via editor settings. i.e. adding `"haskell.plugin.fourmolu.cli": true` in VSCode has no effect. Is there another step needed, or is that just not the right string?

georgefst:
I also can't find a nice way to set this new plugin-specific option in the tests. I could refactor `Fourmolu.hs` to expose a `Bool` argument to `descriptor` instead, but it's non-trivial...

georgefst:
Somehow [3d9f964](https://github.com/haskell/haskell-language-server/pull/2763/commits/3d9f9641d37778e7a554cf7f477022a370690d12) has broken the functionality completely. Odd.

michaelpj:
> Actually, I think we'd get most of the benefits, along with easier maintenance, by just making it a plugin-specific option.

Yes, I think you're right. The only issue would be that you might end up with illegal combinations of plugin options? e.g. if you have some other `fourmolu` option that can't actually take effect when you're using the CLI version. Just a hypothetical worry, though.

> adding "haskell.plugin.fourmolu.cli": true in VSCode has no effect.

I _think_ the VSCode extension also needs to be extended so it knows about the new setting, while the config is "generic" on the HLS side, I don't think is on the VSCode side. I don't really know how it works, though.

I'd probably just test it in a grimy way by setting it to true by default :laughing:

michaelpj:
> It's not immediately obvious to me how to achieve this, and I'm keen to get this PR in to the impending release if possible. So I'll save it for future work.

Isn't it just a matter of adding another plugin-specific option? "Path to fourmolu binary (only used if `cli` option is set)"

georgefst:
> > It's not immediately obvious to me how to achieve this, and I'm keen to get this PR in to the impending release if possible. So I'll save it for future work.
> 
> Isn't it just a matter of adding another plugin-specific option? "Path to fourmolu binary (only used if `cli` option is set)"

Yes, I'm sure you're right. I started writing that before I'd implemented the first option.

> The only issue would be that you might end up with illegal combinations of plugin options?

I don't anticipate that being an issue any time soon.

> I'd probably just test it in a grimy way by setting it to true by default 😆

Yeah, don't worry, I've already done that, and it works nicely. Well, it did before the CWD change, which I haven't been able to investigate yet.

I should be able to get this finished up later today.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -2,91 +2,127 @@
...
+                        ( proc "fourmolu" $
+                            ["-d"]
+                                <> catMaybes
+                                    [ ("--start-line=" <>) . show <$> regionStartLine region
+                                    , ("--end-line=" <>) . show <$> regionEndLine region
+                                    ]
+                                <> map ("-o" <>) fileOpts
+                        ){cwd = Just $ takeDirectory fp'}
+                        contents
+                T.hPutStrLn stderr err
```
michaelpj, line 74: this should get logged properly

hls-test-utils/src/Test/Hls.hs
```diff
@@ -152,11 +154,14 @@ goldenWithHaskellDocFormatter plugin formatter title testDataDir path desc ext a
 runSessionWithServer :: PluginDescriptor IdeState -> FilePath -> Session a -> IO a
 runSessionWithServer plugin = runSessionWithServer' [plugin] def def fullCaps
 
-runSessionWithServerFormatter :: PluginDescriptor IdeState -> String -> FilePath -> Session a -> IO a
-runSessionWithServerFormatter plugin formatter =
+runSessionWithServerFormatter :: PluginDescriptor IdeState -> String -> PluginConfig -> FilePath -> Session a -> IO a
```
michaelpj, line 157: Maybe document this function while you're here

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -2,91 +2,127 @@
...
-                        fillMissingPrinterOpts
-                            (printerOpts <> lspPrinterOpts)
-                            defaultPrinterOpts
-                    }
+properties :: Properties '[ 'PropertyKey "cli" 'TBoolean]
+properties =
+    emptyProperties
+        & defineBooleanProperty
+            #cli
+            "Call out to \"fourmolu\" executable, rather than using the bundled library"
```
michaelpj, line 49:
```suggestion
            "Call out to a \"fourmolu\" executable, rather than using the bundled library"
```

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -2,91 +2,127 @@
...
-                defaultConfig
-                    { cfgDynOptions = fileOpts
-                    , cfgRegion = region
-                    , cfgDebug = True
-                    , cfgPrinterOpts =
-                        fillMissingPrinterOpts
-                            (printerOpts <> lspPrinterOpts)
-                            defaultPrinterOpts
-                    }
+properties :: Properties '[ 'PropertyKey "cli" 'TBoolean]
```
michaelpj, line 44: bikeshedding: "external"? "shell"? I think "cli" seems fine though.

michaelpj (approved):
Implementation looks fine.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
georgefst: Yeah, surpisingly no word I could think of here seems perfect

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
georgefst: It's no worse than with the CLI version. I'll open a new issue to track better logging for both.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
georgefst: I've decided I do prefer "external": [fb3975e](https://github.com/haskell/haskell-language-server/pull/2763/commits/fb3975eedbf553457d46c74ea1500182aa3320b3). Not going to bother changing the variable names though.

georgefst:
As for:

> Add config option to editor plugins e.g. VSCode 

Is there anything I need to do? I see from https://github.com/haskell/haskell-language-server/pull/1576 that we have some auto-generation script. Is this part of the release process? Is any of this (i.e. the process of adding an option) documented anywhere?

michaelpj:
I have no idea. I suspect it isn't documented: I had a look at the CONTRIBUTING for `vscode-haskell` and it didn't say. Maybe you can write it down when you find out. @berberman might know?

burkettd:
This looks like this PR adds support for using a locally installed binary of fourmolu instead of the version bundled with hls.

I'm using hls 1.7.0.0, and installed fourmolu 0.6.0.0 on my path. I tried setting `"haskell.plugin.fourmolu.cli": true` in my coc config for neovim, but it doesn't seem to work because it doesn't seem to apply my `import-export-comma-style` setting which I think was added in 0.6.0.0.

What settings do I need to set to tell HLS to use a locally installed version of fourmolu?

georgefst:
> What settings do I need to set to tell HLS to use a locally installed version of fourmolu?

I genuinely don't know. My personal use case for this disappeared shortly after it was merged. I can only suggest subscribing to https://github.com/haskell/haskell-language-server/issues/2827.

cydparser:
The setting was later renamed to "external":
```json
{
  "haskell.formattingProvider": "fourmolu",
  "haskell.plugin.fourmolu.external": true
}
```

The setting was not respected by haskell-language-server 1.9.0.0

tablitza:
Yes, ~/.cabal/bin is on my emacs exec-path variable.

PR #2764, opened 2022-03-08
Avoid extra parens for wildcard type signature

### LINKED ISSUE #2674: Use type signature: Don't insert excess parentheses .
When completing a type signature HLS is very conservative and always adds parentheses around the actual type.

For example something like `myHead :: [a] -> _` will get a type of `myHead :: [a] -> (a)`. Note the parentheses around the a.

While I personally don't have a big problem with this. Many of my colleagues insist on them being removed. Which usually makes these signatures easier to read!

Ideally I would like the inserted type signature to be the one with the least parentheses. For example `Just _` will require braces if the inserted type is `Either Int Char`. But if it's just `Char` there is no need for them. But in something like `a -> _ -> ...` then we can get away without braces for `Either Int Char`!

It's likely not trivial to achieve this. But it would be really nice.

xsebek:
Avoid parentheses in simple cases (`a`, `Char`, `()`,...).

The good news is, that getting types inside tuples and lists will not force you to apply hint "Redundant brackets":
```haskell
g :: a -> _
g x = (x, const x)
--------------------------
-- before
g :: a -> ((a, b -> a))
--------------------------
-- after
g :: a -> (a, b -> a)
```

The caveat is that it is still context unaware (works on text only) and the common use case of `f :: _` is not improved.

- Hopefully closes #2674 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2764"></a>

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -1188,73 +1188,71 @@ renameActionTests = testGroup "rename actions"
 
 typeWildCardActionTests :: TestTree
 typeWildCardActionTests = testGroup "type wildcard actions"
-  [ testSession "global signature" $ do
-      let content = T.unlines
-            [ "module Testing where"
-            , "func :: _"
-            , "func x = x"
-            ]
-      doc <- createDoc "Testing.hs" "haskell" content
-      _ <- waitForDiagnostics
-      actionsOrCommands <- getAllCodeActions doc
-      let [addSignature] = [action | InR action@CodeAction { _title = actionTitle } <- actionsOrCommands
-                                   , "Use type signature" `T.isInfixOf` actionTitle
-                           ]
-      executeCodeAction addSignature
-      contentAfterAction <- documentContents doc
-      let expectedContentAfterAction = T.unlines
-            [ "module Testing where"
-            , "func :: (p -> p)"
-            , "func x = x"
-            ]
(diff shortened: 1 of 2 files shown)


michaelpj:
Could you add some tests for the examples you give in the PR description?

michaelpj (approved):
LGTM modulo adding some tests.

xsebek:
@michaelpj there are now tests for the tuple, list and unit type. :slightly_smiling_face:

xsebek:
(circleci did not finish so I rebased on master to give it a second chance)

ghcide/test/exe/Main.hs
```diff
@@ -1188,73 +1188,71 @@ renameActionTests = testGroup "rename actions"
...
+        , "func x ="
+        , "  let y :: Int"
+        , "      y = x * 2"
+        , "  in y"
+        ]
+  , testUseTypeSignature "multi-line message"
+        [ "func :: _"
+        , "func x y = x + y"
+        ]
+        [ "func :: (Integer -> Integer -> Integer)"
```
michaelpj, line 1215: Hmmm, I guess it's not easy to tell if the hole is the entirety of the type signature? That seems like the remaining slightly bad case, and also probably a relatively common one...

michaelpj: Anyway, that can be future work.

AndreasPK:
Thanks for this!

PR #2768, opened 2022-03-08
Review masking and add traces when things don't cancel timely

pepeiborra:


pepeiborra:
Who reviews changes made by the code owner? I need a volunteer to review this, thanks. @wz1000 @michaelpj @eddiemundo

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -303,6 +314,7 @@ mapConcurrentlyAIO_ _ [] = pure ()
 mapConcurrentlyAIO_ f [one] = liftIO $ justWait $ fmap f one
 mapConcurrentlyAIO_ f many = do
     ref <- AIO ask
+    -- mask to make sure we keep track of all the asyncs
     waits <- liftIO $ uninterruptibleMask $ \restore -> do
         waits <- liftIO $ traverse (waitOrSpawn . fmap (restore . f)) many
```
pepeiborra, line 319: This `restore` is potentially running in a different thread, as `waitOrSpawn` calls `async`. I wonder what the semantics of `restore` are when invoked in a different thread ...

wz1000: I think it is the expected semantics, since threads inherit their masking state from the parent.

wz1000 (approved):
Looks reasonable.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -706,6 +708,11 @@ shakeRestart recorder IdeState{..} reason acts =
         -- See https://github.com/haskell/ghcide/issues/79
         (\() -> do
           (,()) <$> newSession recorder shakeExtras shakeDb acts reason)
+    where
+        errorAfter :: Seconds -> Recorder (WithPriority Log) -> IO () -> IO ()
```
michaelpj, line 712: From the name I was expecting this to actually throw an error, but it just logs, surprisingly! Maybe rename.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -284,27 +288,36 @@ withRunInIO k = do
...
+    asyncs <- atomicModifyIORef' ref ([],)
+    -- interrupt all the asyncs without waiting
     mapM_ (\a -> throwTo (asyncThreadId a) AsyncCancelled) asyncs
-    mapM_ waitCatch asyncs
+    -- Wait until all the asyncs are done
+    -- But if it takes more than 10 seconds, log to stderr
+    unless (null asyncs) $ do
+        let loop = forever $ do
+                sleep 10
+                traceM "cleanupAsync: waiting for asyncs to finish"
```
michaelpj, line 301: Do we really want to use `Debug.Trace` here?

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -284,27 +288,36 @@ withRunInIO k = do
...
+        withAsyncWithUnmask (\unmask -> unmask loop) $ \_ ->
```
michaelpj, line 302: comment?

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -284,27 +288,36 @@ withRunInIO k = do
...
+waitOrSpawn :: Wait -> IO (Either (IO ()) (Async ()))
 waitOrSpawn (Wait io)  = pure $ Left io
-waitOrSpawn (Spawn io) = Right <$> async io
+waitOrSpawn (Spawn io) = Right <$> asyncWithUnmask (\unmask -> unmask io)
 
-mapConcurrentlyAIO_ :: (a -> IO ()) -> [Wait a] -> AIO ()
-mapConcurrentlyAIO_ _ [] = pure ()
-mapConcurrentlyAIO_ f [one] = liftIO $ justWait $ fmap f one
-mapConcurrentlyAIO_ f many = do
+mapConcurrentlyAIO_ :: [Wait] -> AIO ()
```
michaelpj, line 313: this isn't a map any more

michaelpj:
I guess I mean the functions operating on `AIO`. Perhaps they would exist in an equally gnarly form without `AIO`, I don't know. I trust your ability to write them, I don't trust my ability to review whether the masking is sensible :D

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -224,6 +226,7 @@ updateReverseDeps
     -> [Key] -- ^ Previous direct dependencies of Id
     -> HashSet Key -- ^ Current direct dependencies of Id
     -> IO ()
+-- mask to ensure that all the reverse dependencies are updated
```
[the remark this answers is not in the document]
pepeiborra: STM transactions can still be interrupted by exceptions, so the question is how badly do we want to record the reverse deps. The answer is very badly - failure to do so will lead to unsoundness, specially if this is the first build.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
pepeiborra: If you are seeing this trace, then most likely `hls-graph` is broken. 

Plus, I didn't want to thread a logger around....

pepeiborra:
`AIO` is just a thin wrapper around `IO`/`Async`. The gnarliness existed before `AIO`, it's evil magic contributed by @ndmitchell who, in turn, blames @simonmar.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
pepeiborra: Renamed, thanks!

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
pepeiborra: I've added a bunch of comments

ghcide/src/Development/IDE/Core/Shake.hs
pepeiborra: renamed, thanks

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -109,8 +109,7 @@ modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO ()
 modifyFileExists state changes = do
   FileExistsMapVar var <- getIdeGlobalState state
   -- Masked to ensure that the previous values are flushed together with the map update
-    -- update the map
-  mask_ $ join $ atomicallyNamed "modifyFileExists" $ do
+  join $ mask_ $ atomicallyNamed "modifyFileExists" $ do
```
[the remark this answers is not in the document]
michaelpj: But we can end up not calling `recordDirtyKeys`, since that happens in the IO action afterwards. Is that important? Maybe the comment could say.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
michaelpj: You're in IO, you could just print to stderr. It's a small difference, but I do think not having `Debug.Trace` in production code is good.

ghcide/src/Development/IDE/Core/FileExists.hs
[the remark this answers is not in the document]
pepeiborra: The IO is just logging, `recordDirtyKeys` mutates the collection inside the STM transaction. That said I appreciate how this is confusing and perhaps the `mask_` was better at the top level for clarity

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
pepeiborra: That's a fair point

PR #2769, opened 2022-03-09
Abbreviate explicit import code lenses

### LINKED ISSUE #2765: Abbreviate explicit import code lenses
It's cool that the code lens shows the code that will be present after the change, but for imports that import a lot of things, the resulting code lens is *massive*. In some cases I've had it occupy most of or all of my editor's viewport, which is pretty annoying. This also causes problems in some clients that don't deal so well with massive code lenses (e.g. https://github.com/emacs-lsp/lsp-haskell/issues/137).

I hypothesise that nobody really needs to see exactly what's going to be inserted by this code lens. I think it could be abbreviated:
- Radically, just show "Make this import fully explicit"
- Less radically, show "import Foo (<49 items>)" or something

michaelpj:
The tests currently don't check anything about the titles, I'm unsure
whether it's worth writing a test just for this.

Fixes #2765.

Demo: 
image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2769"></a>


pepeiborra:
You could extract the new logic into a pure function. Then you can test it easily without involving lsp-test.

pepeiborra (approved):

michaelpj:
Great point, I'll add a test for the abbreviation function.

michaelpj:
I added some tests, which was good, because it was broken :) one day I will not be surprised by this.

dariooddenino:
Hi, this is my version of hls:
```
❯ haskell-language-server-wrapper --version
haskell-language-server version: 1.6.1.1 (GHC: 9.0.2) (PATH: ~/.local/bin/haskell-language-server-wrapper) (GIT hash: 8466bc11a655f4cfbc8f228e527c2a7bf6762c7f)
```
which should already include this pr.

I'm still seeing incredibly long code lenses for explicit import. Am I missing anything?

Thanks!

michaelpj:
I don't think this PR is in a released version yet.

dariooddenino:
It looks like this was merged into master 21 days ago, and I manually built hls from a much more recent commit.
Now I feel like I'm missing something incredibly obvious. Sorry for the confusion.

michaelpj:
Can you provide a reproduction?

dariooddenino:
Sure!
https://github.com/dariooddenino/pru

This is what I see from `src/Handler/Home.hs`:

Screenshot from 2022-04-05 10-14-22

jhrcek:
The thing I'd check is whether the haskell-language-server-wrapper that you have in your PATH is the actual binary that is being run by your editor. Can you check some tree view of running processes and check the command that your editor used to start hls? E.g on my system I can use htop to see that my vscode started hls like this.
Screenshot from 2022-04-05 10-29-38
You will likely find out that your editor has started different binary than the one you're showing version of..

dariooddenino:
@jhrcek You are right.
I even checked emacs' path and tried running manually hls from within it, but didn't think about the obvious solution of checking the running process.
Sorry, and thanks to you all!

michaelpj:
Cannot reproduce indeed:
image

_However_ the "Refine import" lens is still massive and could do with this treatment.

PR #2770, opened 2022-03-09
Purge some more hslogger

michaelpj:
At this point we only really need it for `hie-bios`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2770"></a>


fendor:
@michaelpj We can purge it from hie-bios easily as well, the logic HLS uses, doesn't even use hslogger.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -125,13 +125,21 @@ import           System.Environment                                 (setEnv,
...
+  | LogUsingExtensions NormalizedFilePath [String] -- Extension is only imported conditionally, so we just stringify them
   deriving Show
 
 instance Pretty Log where
   pretty = \case
     LogShake log -> pretty log
+    LogApplying fp res -> "Applying hint(s) for" <+> viaShow fp <> ":" <+> viaShow res
+    LogGeneratedIdeas fp ideas -> "Generated hlint ideas for for" <+> viaShow fp <> ":" <+> viaShow ideas
+    LogUsingExtensions fp exts -> "Using extensions for " <+> viaShow fp <> ":" <+> pretty exts
+    LogGetIdeas fp -> "Getting hlint ideas for " <+> viaShow fp
```
fendor, line 142:
```suggestion
    LogUsingExtensions fp exts -> "Using extensions for" <+> viaShow fp <> ":" <+> pretty exts
    LogGetIdeas fp -> "Getting hlint ideas for" <+> viaShow fp 
```

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -584,7 +590,6 @@ applyHint ide nfp mhint =
       Right appliedFile -> do
         let uri = fromNormalizedUri (filePathToUri' nfp)
         let wsEdit = diffText' True (uri, oldContent) (T.pack appliedFile) IncludeDeletions
-        liftIO $ logm $ "hlint:applyHint:diff=" ++ show wsEdit
```
fendor, line 587: No replacement?

fendor (approved):
Pretty cool, LGTM!

michaelpj:
I was planning to just teach `hie-bios` to use the same kind of co-log based infrastructure as we are using now, it probably *should* be loggging!

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -125,13 +125,21 @@ import           System.Environment                                 (setEnv,
 import           Text.Regex.TDFA.Text                               ()
 -- ---------------------------------------------------------------------
 
-newtype Log
+data Log
   = LogShake Shake.Log
+  | LogApplying NormalizedFilePath (Either String WorkspaceEdit)
+  | LogGeneratedIdeas NormalizedFilePath [[Refact.Refactoring Refact.SrcSpan]]
+  | LogGetIdeas NormalizedFilePath
+  | LogUsingExtensions NormalizedFilePath [String] -- Extension is only imported conditionally, so we just stringify them
```
eddiemundo, line 133: I think the fields should be banged, although it probably doesn't matter since our current recorders will force the fields anyway, and there are composite types that prevent some forcing, and we probably will never buildup huge thunks of NormalizedFilePath or the composite types, and I noticed I didn't bang the wrapper constructor fields everywhere anyway.

eddiemundo (approved):
LGTM

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
michaelpj: I just added `StrictData` to the module, which is probably a sensible default these days anyway.

PR #2774, opened 2022-03-12
Fix a space leak in hls-graph

pepeiborra:
I found this space leak by running the edit benchmark with `+RTS -hc` to track down the monotonically increasing cost centre and then trying various things until ultimately locating the source of the leak. 

The thing that helped ultimately was an info table profile (`-hi`), only available with ghc 9.2.2 and requires building with `-finfo-table-map -fdistinct-constructor-tables`). Thanks @mpickering and @hasurahq for contributing this profiling mode!

Command line, for reference:

```
cabal build --project-file cabal-ghc921.project -w ghc-9.2 exe:ghcide && dist-newstyle/build/x86_64-osx/ghc-9.2.2/ghcide-1.6.0.1/x/ghcide-bench/build/ghcide-bench/ghcide-bench --ghcide dist-newstyle/build/x86_64-osx/ghc-9.2.2/ghcide-1.6.0.1/x/ghcide/build/ghcide/ghcide -s edit --ghcide-options "+RTS -hc  -l -RTS"
```

# Before

image
image
image

# After

image
image

### DIFF
--- a/hls-graph/src/Development/IDE/Graph/Internal/Types.hs
+++ b/hls-graph/src/Development/IDE/Graph/Internal/Types.hs
@@ -120,12 +120,12 @@ getDatabaseValues = atomically
                   . databaseValues
 
 data Status
-    = Clean Result
+    = Clean !Result
     | Dirty (Maybe Result)
     | Running {
         runningStep   :: !Step,
         runningWait   :: !(IO ()),
-        runningResult :: Result,
+        runningResult :: Result,     -- LAZY
         runningPrev   :: !(Maybe Result)
         }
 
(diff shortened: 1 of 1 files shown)


pepeiborra:
I found a second space leak, this time in the `ExportsMap` code in ghcide, using the same method.

# Before
image
image

# After 
image
image

michaelpj:
What do you think about just turning on StrictData everywhere?

fendor (approved):

pepeiborra:
> What do you think about just turning on StrictData everywhere?

I am a big fan.

Issue #1977, opened 2021-06-25
Capture stdout and stderr in eval plugin


anka-213:
### Steps to reproduce

Run the "Evaluate..." code lens on the following code 
```
-- >>> putStrLn "Hello"
```
or
```
-- >>> hPutStrLn stderr "Hello"
```

### Expected behaviour

It should put
```
-- Hello
```
in a comment below, just like it would do with normal return values.

### Actual behaviour

In the `stderr` case, the output is sent to the debug output together with the logs.
In the `stdout` case I believe the output is sent in the LSP channel directly to the editor as is. If you're lucky, the editor's LSP parser will handle the error gracefully instead of just crashing or something.

xsebek:
I would like to try implementing the `stdout`/`stderr` capture :slightly_smiling_face: 

In #2775 it was suggested by @pepeiborra to use `-fexternal-interpreter` from GHC(i).

So far I could not figure out from [GHC docs](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/external-interpreter) and code what happens to `stdout` when GHCi uses the external interpreter (which seems similar to our use case) and if it could be captured :confused:

anka-213:
I wonder if a simple redirect of stdout using `hDuplicateTo` like [silently](https://hackage.haskell.org/package/silently) does wouldn't actually work? As long as we make sure that all the output that should be sent to the actual stdout/stderr gets sent to the backup copy of the handles and not the ones currently named `stdout`/`stderr`. It should be easy to do for stdout, since (afaik) all the lsp communication is done at the same place, but might be more difficult to do for stderr, since code that writes to stderr is more spread out over the codebase.

There might also be an issue with this approach if we want to run multiple `eval` commands in parallel, since afaik, `stdout` is global per process.

`hDuplicateTo` is already done to prevent crashes from accidental prints: https://github.com/haskell/haskell-language-server/blob/65241229275e6a0950d6ca982456ed9819361b8a/ghcide/src/Development/IDE/Main.hs#L260-L273

Maybe all that's needed is to redirect `stderr` to a temporary buffer during eval and hope that there are no stray prints to stderr/stdout during that time. But of course, if we can get a more robust implementation using `-fexternal-interpreter`, that's better.

xsebek:
Hey @anka-213, thanks for pointing the redirection out, I was not aware of it!

The "global per process" part is exactly why we would like to use the external interpreter process as GHC and GHCi do.
[...]

PR #2775, opened 2022-03-12
Eval plugin: mark exceptions

### LINKED ISSUE #2353: Eval Plugin: different exception rendering than doctest
### Your environment

```
> haskell-language-server --probe-tools
haskell-language-server version: 1.4.0.0 (GHC: 8.10.7) (PATH: /nix/store/2sxjf20hy89835gxa07a80hqsg25aywq-haskell-language-server-1.4.0.0/bin/haskell-language-server)
cabal:		3.6.0.0
stack:		Not found
ghc:		8.10.7

```
Which OS do you use:

nixos
Which lsp-client do you use:

vscode

https://hackage.haskell.org/package/error
Contents of `hie.yaml`:
none

### Steps to reproduce

Use the eval plugin to evaluate a throwing bit of code, and then run `doctests` on the repo, which will complain that it comes to a different conclusion because the error formatting is different:

```
src/Data/Error.hs:195: failure in expression `unwrapIOError $ Left (newError "oh no!")'
expected: oh no!
 but got: *** Exception: oh no!
```

that is the eval plugin produced `expected: oh no!`, but doctests will produce `*** Exception: oh no!` and thus report an error when running.

The solution might be as easy as adding `*** Exception: ` to the beginning of every `displayException` string.

xsebek:
This change will* format exception results in the same way as doctest:
```haskell
>>> if error "-1" then 1 else 0
------------------
-- current:
-1
------------------
-- new:
*** Exception: -1
```

*) EDIT: if the configuration option `haskell.plugin.eval.config.exception` is set to `true`.

- Fixes #2353 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2775"></a>


xsebek:
Hi @jneira, there seems to be no code owner of the eval plugin - do you know who might be familiar with the code? :slightly_smiling_face:

pepeiborra (approved):
Looks good to me.

Ailrun (changes requested):
This no "Exception" message is intended AFAIK: https://github.com/haskell/haskell-language-server/tree/master/plugins/hls-eval-plugin#multiline-output

I think we cannot change this at least before we have somewhat "proper" handling of IO.

Ailrun:
C.C. @tittoassini

xsebek:
> This no "Exception" message is intended AFAIK: https://github.com/haskell/haskell-language-server/tree/master/plugins/hls-eval-plugin#multiline-output

Thanks, @Ailrun, I was not aware of this hack. I do find the missing message a bit confusing, but if someone is using `error` to emulate `putStr` then I can let this rest for a while.

pepeiborra:
I was not aware of this hack either! We should strive to make this hack unnecessary, some options come to mind: 

* Capturing output can be solved by using the external interpreter (`-fexternal-interpreter`). We should revisit using it for the eval plugin, but this might be challenging or even impossible.
* We could wrap all the user code in a custom exception type with a crafted Show instance that doesn't escape output. Then we catch all other exception types and add the "*** Exception" markup.

What do you think @Ailrun ?

Ailrun:
@pepeiborra I think the second would be more feasible. If we can handle IO better, that would be better, but that may take quite effort

xsebek:
Hi @pepeiborra, I am not entirely sure what you mean by the second option?

I imagine this could be hacked around by recognizing some prefix like `"PRETEND_I_AM_IO: "`.

If supporting this hack should involve more effort, I would rather attempt to fix the IO instead :slightly_smiling_face:

pepeiborra:
Sorry I wasn't clear: The only way to properly fix #1977 is by using the external interpreter, which is far from obvious.

xsebek:
Yes, I understood that the first option is fixing #1977 with an external interpreter though I have no idea how to do that (yet).

What I can not imagine is how to "wrap all the user code" which you suggested as the second option. :thinking:

pepeiborra:
Yes, I can see now that the 2nd option is not trivial either.

xsebek:
Another solution that comes to mind would be to make this configurable by the user.

Currently, the eval plugin has the `#diff` property for WAS/NOW (which is passed to `runTests` so it's almost there).

So I suggest adding an `#exception` property (`False` by default) with the hint:

> Enable marking exceptions with `*** Exception:` similarly to doctest and GHCi.

What do you think, @pepeiborra, @Ailrun? It seems relatively low effort and once we get better IO support, we could just turn it on by default or keep it off.  :slightly_smiling_face:

xsebek:
OK, the behaviour is now configurable via `haskell.plugin.eval.config.exception` (default `false`) and there is one more test for this configuration option.

Does this resolve your concerns @Ailrun? Btw. thanks for the scaffolding around `diff`. :slightly_smiling_face:

plugins/hls-eval-plugin/README.md
```diff
@@ -274,6 +279,8 @@ To display it properly, we can exploit the fact that the output of an error is d
 ]
 ```
 
+This assumes you did not turn on exception marking (see #)
```
Ailrun, line 282: It looks like a link is missing?

Ailrun (approved):
LGTM :) Here is one minor comment for documentation.

plugins/hls-eval-plugin/README.md
xsebek: Thanks, I was about to look up how to link sections and got distracted. :sweat_smile: 

Fixed :heavy_check_mark:

xsebek:
At the last minute, I remembered to check the docs for configuration options and added the eval section. :sweat_smile: 

I wonder if it needs to be added to vscode-haskell too?

michaelpj (approved):
Docs look fine to me. The way the config options are documented is suboptimal, but this doesn't make it worse!

I don't know what the deal is with exposing options in `vscode-haskell`...

PR #2781, opened 2022-03-14
Customizable TH warning

pepeiborra:
I want to change the `LspT` action that gets run the first time that we run into unsupported Template Haskell. This can also be used to completely disable the TH warning (programmatically) .

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2781"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -828,22 +828,11 @@ getModIfaceFromDiskAndIndexRule recorder =
 
   return (Just x)
 
-displayTHWarning :: LspT c IO ()
-displayTHWarning
-  | not isWindows && not hostIsDynamic = do
-      LSP.sendNotification SWindowShowMessage $
-        ShowMessageParams MtInfo $ T.unwords
-          [ "This HLS binary does not support Template Haskell."
-          , "Follow the [instructions](" <> templateHaskellInstructions <> ")"
-          , "to build an HLS binary with support for Template Haskell."
-          ]
-  | otherwise = return ()
-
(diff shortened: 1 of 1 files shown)


michaelpj:
Hmm, given the existence of the LSP `showMessage` recorder, we can *almost* replace this with just logging. But I think it's currently only showing logs at error severity, not warnings. It would be nicer if this was "just" a log, though.

pepeiborra:
> Hmm, given the existence of the LSP `showMessage` recorder, we can *almost* replace this with just logging. But I think it's currently only showing logs at error severity, not warnings. It would be nicer if this was "just" a log, though.

Agreed

michaelpj:
Would it be crazy to show warnings to the user as well? Or is that going to be too aggressive in general?

pepeiborra:
> Would it be crazy to show warnings to the user as well? Or is that going to be too aggressive in general?

Do we have any warning logs?

michaelpj:
Yes, quite a few.

The flip side of this is: how bad would it be if this message went only to the log? Note that now we will send it to the client log with `window/logMessage`, so it should be readily visible if anyone looks for diagnostics. Do we need it to be a "push" notification?

wz1000 (approved):

pepeiborra:
> Yes, quite a few.
> 
> The flip side of this is: how bad would it be if this message went only to the log? Note that now we will send it to the client log with `window/logMessage`, so it should be readily visible if anyone looks for diagnostics. Do we need it to be a "push" notification?

Sending the message to the log stream will provide useful diagnostics, but the goal of this message is to create awareness. I suspect that most HLS users, specially new ones, will not even bother to check the logs in the event of a crash.

Issue #2734, opened 2022-02-22
Feature: Quick fix to add all constructors to the import list


ejconlon:
Wingman case splits (which are amazing) often introduce many constructors that are not imported yet. Often one wants to do a wildcard import of all constructors, but the quick fix for imports only adds one constructor at a time to the import list. (Example: `Add MyDataType(MyFirstConstructor) to the import list...`) Quite often I just quick fix once, scroll up to the imports, and replace with `..`, which disrupts the flow of development. It would be nice to have a second quick fix option to add all constructors to the import list. (Example: `Add MyDataType(..) to the import list...`)

As always, apologies if I missed an existing issue for this, and thanks for the wonderful project.

July541:
We can close it if no more work on it.

PR #2782, opened 2022-03-15
Allow import all constructors

July541:
This is for #2734 

Brief intro:

With 
```haskell
Module A where
data Foo = Foo | Bar
```
it will suggest import `Foo(..)` now, like `import A(Foo(..))` and `Add Foo(..) to the import list of A`.

---

Note:
Support Now!
~~This doesn't work for record fields, ideally, we should support this.~~

Example:
```haskell
data Foo = Bar {a :: Int}
```
if we have `f = a`, we should also have `import A(Foo(..))`. Maybe we can extend
https://github.com/haskell/haskell-language-server/blob/621c2bb2173d5231082e5e4c4f157b2efb4f94c7/ghcide/src/Development/IDE/Types/Exports.hs#L64-L70
with a field to indicate `a`'s data type if `a` is a record field.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2782"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction.hs
@@ -1733,6 +1733,13 @@ data ImportStyle
       --
       -- @P@ and @?@ can be a data type and a constructor, a class and a method,
       -- a class and an associated type/data family, etc.
+
+    | ImportAllConstructors T.Text
+      -- ^ Import all constructors for a specific data type.
+      --
+      -- import M (P(..))
+      --
+      -- @P@ __must__ be a data type.
   deriving Show
 
 importStyles :: IdentInfo -> NonEmpty ImportStyle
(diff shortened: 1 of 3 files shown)


July541:
With `data A = B | C`,

```haskell
import A (A(B))
f = C
```
it will cause `import A (A(B, ..)` if we import all constructors. I'll check it later.

pepeiborra:
This is a good idea, thanks for working on it!

pepeiborra:
> if we have `f = a`, we should also have `import A(Foo(..))`. Maybe we can extend
> 
> https://github.com/haskell/haskell-language-server/blob/621c2bb2173d5231082e5e4c4f157b2efb4f94c7/ghcide/src/Development/IDE/Types/Exports.hs#L64-L70
> 
> 
> with a field to indicate `a`'s data type if `a` is a record field.

Did you consider checking if `isJust parent` ?

pepeiborra (approved):
Fantastic work, thank you @July541 !

PR #2787, opened 2022-03-18
Target GHC 9.2.2

pepeiborra:


July541:
Development with 9.2.1 on M1 is entirely broken, it spent me a lot of time and finally found 9.2.2 is the solution.

July541:
Development with 9.2.1 on M1 is entirely broken, it spent me a lot of time and finally found 9.2.2 is the solution.

michaelpj:
Looks like we need to deal with https://github.com/digital-asset/ghc-lib/issues/352

July541:
For reference, here script works for my M1 chip machine
```
>>> diff stack-9.2.1.yaml stack-9.2.2.yaml
2c2
< compiler: ghc-9.2.1
---
> compiler: ghc-9.2.2
49,50c49,50
< - ghc-lib-9.2.1.20220109
< - ghc-lib-parser-9.2.1.20220109
---
> - ghc-lib-9.2.2.20220307
> - ghc-lib-parser-9.2.2.20220307
92c92
< - ghc-boot-9.2.1
---
> # - ghc-boot-9.2.1
```

wz1000 (approved):

michaelpj:
The branch protection rules need to be changed to require the 9.2.2 circleci job instead of the 9.2.1 one.

.circleci/config.yml
```diff
@@ -105,9 +105,11 @@ jobs:
       - STACK_FILE: "stack-9.0.2.yaml"
     <<: *defaults
 
-  ghc-9.2.1:
+  ghc-9.2.2:
     environment:
-      - STACK_FILE: "stack-9.2.1.yaml"
+      # https://github.com/digital-asset/ghc-lib/issues/352
+      - CPATH: "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/ffi:$CPATH"
```
michaelpj, line 111: The workflow is broken with this: https://github.com/haskell/haskell-language-server/actions/runs/2012818868

michaelpj:
Apparently this PR broke the caching workflow and nobody noticed :/

.circleci/config.yml
michaelpj: Oh, apparently I did notice, but this wasn't fixed. It appears to be broken to this day. @pepeiborra do you remember what you did here?

July541: I've noticed this and #2936 is attempting to fix it up.

July541:
Caching works now :)

PR #2790, opened 2022-03-20
Provide all format suggestions in AlternatFormat Code Action

### LINKED ISSUE #2772: Provide All Alternate Number Formats regardless of Enabled Language Extensions
Currently, the alternate number format plugin only suggests formats when the Language Extension is enabled. 

So for example, with Haskell2010 set as the default language, the literal `15` will provide `0xF` and `0o17` as alternatives. 

With `BinaryLiterals` enabled the literal `15` will provide `0xF`, `0o17`, and `0b1111`. 

Instead the provided alternatives should span all possible formats and add the language extension if not already enabled.

drsooch:
Closes #2772

Previously the alternate format plugin would only generate code actions based on enabled extensions. With this PR we now provide all formats as code actions and when necessary insert the required extension.

This PR also refactors a portion of the logic internal to the plugin so we can track the Extensions needed. Overall, functionality remains the same.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2790"></a>


pepeiborra (approved):
The ghcide changes look fine to me

plugins/hls-alternate-number-format-plugin/test/testdata/TIntDtoND.expected.hs
```diff
@@ -2,4 +2,4 @@
 module TIntDtoND where
 
 convertMe :: Integer
-convertMe = 125.345e3
```
michaelpj, line 5: Did you mean to change this?

michaelpj (approved):

michaelpj:
Genuine test failure?

plugins/hls-alternate-number-format-plugin/test/testdata/TIntDtoND.expected.hs
drsooch: Yeah, I updated the way Integers are converted into NumDecimal Format

drsooch:
No, I forgot that files with no explicit `default-language` defaults to `GHC2021` which includes the extensions we are trying to add. The golden test expects the extension to be added.

michaelpj:
Great, and thanks for updating the docs nicely!

PR #2793, opened 2022-03-22
[hls-graph] do not Spawn on  refresh

pepeiborra:
Avoid spawning a new thread to schedule a build node dependencies. For performance reasons we want to spawn only the rule executions, while running the scheduler in the local thread.

Benchmark results: TBA

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2793"></a>


pepeiborra:
The Cabal edit example shows that this isn't an optimization on multi-core systems: even if we allocate 25% less, the amount of concurrency is crippled too much (from >3500 threads to 220), and the user time metric regresses noticeably. This is because `refresh` does block until the `key` dependencies are fully evaluated. In these cases we really want to spawn `refresh`, so that we can `refresh` all our dependencies concurrently.

This means that we want to spawn `refresh` in at least the following scenarios:
1. When we don't know anything about the `key`, i.e. the first time it is evaluated
2. When the `key` has dirty dependencies. But figuring this out is expensive.

Closing this PR for now

PR #2795, opened 2022-03-22
Troubleshooting: Add solution from #1882

chris-martin:


docs/troubleshooting.md
```diff
@@ -215,3 +215,7 @@ Due to some limitations in the interaction between HLS and `stack`, there are [i
 
 - The project has to be built succesfully *before* loading it with HLS to get components other than the library work.
 - Changes in the library are not automatically propagated to other components, especially in the presence of errors in the library. So you have to restart HLS in order for those components to be loaded correctly. The usual symptom is the editor showing errors like `Could not load module ...` or `Cannot satisfy -package ...`.
+
+### HLS stops with "Assertion failed"
```
michaelpj, line 219: I think we could expand this to be more general. The issue is something like "HLS fails to start and displays an error from cabal, which seems to be persistent", and the potential fix is "delete the cabal state used by HLS". I'm not quite sure how to phrase it but this is something to try for various kinds of persistent issues with cabal.

michaelpj:
I incorporated this into a recent PR.

PR #2802, opened 2022-03-27
Fix Show instance

pepeiborra:


ghcide/src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -421,7 +421,10 @@ newtype GhcSessionDeps = GhcSessionDeps_
         -- Required for interactive evaluation, but leads to more cache invalidations
         fullModSummary :: Bool
     }
-    deriving newtype (Eq, Show, Typeable, Hashable, NFData)
+    deriving newtype (Eq, Typeable, Hashable, NFData)
+
+instance Show GhcSessionDeps where
```
michaelpj, line 426: What's wrong with the derived one? Worth a comment.

pepeiborra: It prints True or False instead of the key name

pepeiborra: The problem is that the deriving strategy is new type instead of stock

michaelpj: So why not use the stock deriving?

pepeiborra: Oh the stock instance is very verbose and bloats traces. I have changed the custom instance to not ignore the boolean for good measure.

wz1000 (approved):

PR #2807, opened 2022-03-29
Fix `cabal install` commands for local HLS build in docs

9999years:
The existing command errors:

```
$ cabal install pkg:haskell-language-server
cabal: Internal error in target matching. It should always be possible to find
a syntax that's sufficiently qualified to give an unambiguous match. However
when matching 'pkg:haskell-language-server' we found haskell-language-server
(named-package) which does not have an unambiguous syntax. The possible syntax
and the targets they match are as follows:
'haskell-language-server' which matches haskell-language-server
(named-package)
'pkg:haskell-language-server' which matches haskell-language-server
(named-package), pkg:haskell-language-server (unknown-component),
:pkg:pkg:lib:pkg:file:haskell-language-server (unknown-file)
':pkg:haskell-language-server' which matches haskell-language-server
(named-package)
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2807"></a>

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -168,13 +168,13 @@ Dynamically linked binaries (including`ghci`) use the system linker instead of t
 The easiest way to obtain a dynamically linked HLS binary is to build HLS locally. With `cabal` this can be done as follows:
 
 ```bash
-cabal update && cabal install pkg:haskell-language-server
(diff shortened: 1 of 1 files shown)


fendor (approved):
Thank you for catching this!

PR #2809, opened 2022-03-29
Enable rename plugin

### LINKED ISSUE #2804: Enable the rename plugin on single component projects
Currently the rename plugin is disabled (and hence, not linked to the distributed binary) because it doesn't work reliably on multi-component projects (this is not the plugin's fault, but rather HLS / hie-bios). 

While there is work being done to address this, it's been almost a year since the plugin was contributed, and we regularly get requests for the functionality. So it would be nice to enable the plugin at least in single component projects.

OliverMadine:
The rename plugin is disabled due to limited multi-component support in the HLS / hie-bios. This limitation causes renaming across components to work partially (depending which files have been indexed).

---
### Within-module renaming
While the work on full [multi-component support](https://github.com/haskell/haskell-language-server/issues/2193) is continued, we can enable renaming of names within modules for use in both single- and multi-component projects.

I think that this is the better of the temporary solutions since it eliminates all partial behaviour, while still allowing the plugin to be of some use within multi-component projects.

The conditions required to perform a within-module rename are then:
- the source of the name is the current module (not an imported name)
- the current module must have an export list that does *not* contain the name

In the case that these are *not* met, we fail informatively.

---
### Cross-module plugin option
This also adds plugin option for renaming across modules. This can be used in single-component projects or when ensuring that all components that reference the name are indexed.

By default cross-module renaming is disabled. It can be enabled with the following configuration option:
```json 
"haskell.plugin.rename.config.crossModule": true
```

---
Closes #2804 
Closes haskell/vscode-haskell#560
Relevant issues: #2598

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2809"></a>

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Orphans.hs
+++ b/ghcide/src/Development/IDE/GHC/Orphans.hs
@@ -33,21 +33,17 @@ import           Retrie.ExactPrint          (Annotated)
 
 import           Development.IDE.GHC.Compat
 import           Development.IDE.GHC.Util
-import           Development.IDE.Types.Location
 
 import           Control.DeepSeq
 import           Data.Aeson
 import           Data.Bifunctor             (Bifunctor (..))
 import           Data.Hashable
-import           Data.Mod.Word
 import           Data.String                (IsString (fromString))
 #if MIN_VERSION_ghc(9,0,0)
 import          GHC.ByteCode.Types
 #else
 import          ByteCodeTypes
 #endif
-
-import           Language.LSP.Types        (UInt)
 
 -- Orphan instances for types from the GHC API.
 instance Show CoreModule where show = prettyPrint
@@ -143,11 +139,6 @@ instance Show ModuleName where
 instance Hashable ModuleName where
     hashWithSalt salt = hashWithSalt salt . show
 
-instance Hashable Location
-instance Hashable Range
-instance Hashable Position
-instance Hashable UInt
-instance Hashable (Mod a) where hash n = hash (unMod n)
 
 instance NFData a => NFData (IdentifierDetails a) where
     rnf (IdentifierDetails a b) = rnf a `seq` rnf (length b)
--- a/ghcide/src/Development/IDE/GHC/Orphans.hs
+++ b/ghcide/src/Development/IDE/GHC/Orphans.hs
@@ -33,17 +33,21 @@ import           Retrie.ExactPrint          (Annotated)
 
 import           Development.IDE.GHC.Compat
 import           Development.IDE.GHC.Util
+import           Development.IDE.Types.Location
 
 import           Control.DeepSeq
 import           Data.Aeson
 import           Data.Bifunctor             (Bifunctor (..))
 import           Data.Hashable
+import           Data.Mod.Word
 import           Data.String                (IsString (fromString))
 #if MIN_VERSION_ghc(9,0,0)
 import          GHC.ByteCode.Types
 #else
 import          ByteCodeTypes
 #endif
+
+import           Language.LSP.Types        (UInt)
 
 -- Orphan instances for types from the GHC API.
 instance Show CoreModule where show = prettyPrint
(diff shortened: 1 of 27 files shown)


pepeiborra (changes requested):
This makes no sense to me, it cripples the plugin artificially in an unrecoverable way. 

Sigma IDE (a custom LSP server built with HLS) includes the rename plugin since the Sigma codebase is single component. Isn't there a way to preserve the full feature set in single component projects?

pepeiborra:
> We could preserve the full feature set in single-component projects but I don't think we have a way of determining if we are in a multi-component project or not? 

I would suggest putting the behaviour in this PR behind a plugin option, enabled by default if that's what we want.

michaelpj:
Pepe, when you say it cripples it "unrecoverably", do you mean specifically that you can't turn it off for Sigma?

I agree that making it possible to get the "full" plugin is a good idea, and that probably means a plugin option.

michaelpj:
I think this is an okay step and provides at least *some* useful renaming to users (as opposed to nothing, currently). As and when we can support multiple components, we can just stop failing on non-local identifiers.

pepeiborra:
> Pepe, when you say it cripples it "unrecoverably", do you mean specifically that you can't turn it off for Sigma?

I mean that it's not possible to turn this change off, period. Anyone can do `cabal install haskell-language-server` with the `rename` flag to link this plugin and use it to do multi-component renames as long as they have indexed all the components.

michaelpj:
I totally agree that we shouldn't actually _lose_ the old functionality!

OliverMadine:
Thanks for the feedback @pepeiborra @michaelpj! I’ve [added a plugin option](https://github.com/haskell/haskell-language-server/pull/2809/commits/866d6f5b973d3817330abb3b68d3ff051f999c20) so that we can maintain the old functionality.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -1,188 +1,251 @@
...
 import           Ide.Types
-import           Language.Haskell.GHC.ExactPrint
 import           Language.LSP.Server
 import           Language.LSP.Types
 
+instance Hashable Location
+instance Hashable Range
+instance Hashable Position
+instance Hashable UInt
+instance Hashable (Mod a) where hash n = hash (unMod n)
```
pepeiborra, line 55: Where are these orphaned instances used? If not used please delete them, and if used they need to go into one of the Orphan modules in ghcide.

pepeiborra (changes requested):

pepeiborra:
I believe that an additional PR to https://github.com/haskell/vscode-haskell will be needed to expose the config option

OliverMadine:
> I believe that an additional PR to https://github.com/haskell/vscode-haskell will be needed to expose the config option

[Done!](https://github.com/haskell/vscode-haskell/pull/563)

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
OliverMadine: They’re used since I changed the list of name references to a HashSet for better lookup, although it didn’t have much of an impact on performance.

[I’ve moved them to Development/IDE/GHC/Orphans.hs](https://github.com/haskell/haskell-language-server/pull/2809/commits/bb91db031e7c1962084d9c69aa3761e2cd118b97)

ghcide/src/Development/IDE/GHC/Orphans.hs
```diff
@@ -139,6 +143,11 @@ instance Show ModuleName where
 instance Hashable ModuleName where
     hashWithSalt salt = hashWithSalt salt . show
 
+instance Hashable Location
+instance Hashable Range
+instance Hashable Position
+instance Hashable UInt
+instance Hashable (Mod a) where hash n = hash (unMod n)
```
pepeiborra, line 150: Hmm, `GHC.Orphans` is not a good place since none of these types is defined in the `ghc` package.

Actually, it's probably better to revert to the previous state with the instances  in the rename plugin, and send PRs to haskell/lsp and `umod` with the missing instances so that we can remove the orphan instances in the future.  

Thank you!

pepeiborra (changes requested):

ghcide/src/Development/IDE/GHC/Orphans.hs
OliverMadine: No problem, I've reverted the commit

pepeiborra (approved):
Thanks @OliverMadine for accepting all our suggestions and getting the rename plugin enabled at last!

PR #2810, opened 2022-03-30
Support hls-hlint-plugin and hls-stylish-haskell-plugin on ghc-9.2

July541:


July541:
Closed and continue in a smaller scope.

PR #2814, opened 2022-04-01
bench: Add more metrics

wz1000:
Add columns to keep track of total GHC rebuilds, time for first response and
average time per response

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2814"></a>


pepeiborra (approved):
Nice!

wz1000:
Some benchmarks here: https://well-typed.com/blog/2022/04/hls-performance/

(this is the `serialize-core` set of commits)

PR #2816, opened 2022-04-03
Make splice plugin compatible with GHC 9.2

eddiejessup:
I don't really know what I'm doing, but I did the minimal amount to make the splice plugin compatible with GHC 9.2.

I import from `GHC.Types.Error` to extract the warnings and errors, I'm not sure if that's okay, or if the idea is to avoid plugins depending directly on `GHC` modules?

Good chance I've done something wrong here, feedback welcome.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2816"></a>

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
@@ -476,7 +476,17 @@ graftDeclsWithM dst toDecls = Graft $ \dflags a -> do
     modifyDeclsT (fmap DL.toList . go) a
 
 
-class (Data ast, Default l, Typeable l, Outputable l, Outputable ast) => ASTElement l ast | ast -> l where
+-- In 9.2+, we need `Default l` to do `setPrecedingLines` on annotated elements.
+-- In older versions, we pass around annotations explicitly, so the instance isn't needed.
+class
+    ( Data ast
+    , Typeable l
+    , Outputable l
+    , Outputable ast
+#if MIN_VERSION_ghc(9,2,0)
+    , Default l
+#endif
+    ) => ASTElement l ast | ast -> l where
     parseAST :: Parser (LocatedAn l ast)
     maybeParensAST :: LocatedAn l ast -> LocatedAn l ast
     {- | Construct a 'Graft', replacing the node at the given 'SrcSpan' with
--- a/plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
@@ -25,6 +25,7 @@ module Development.IDE.GHC.ExactPrint
       Annotate,
       setPrecedingLinesT,
 #else
+      setPrecedingLines,
       addParens,
       addParensToCtxt,
       modifyAnns,
@@ -56,6 +57,7 @@ import           Control.Monad.Trans.Except
 import           Control.Monad.Zip
 import           Data.Bifunctor
 import           Data.Bool                               (bool)
+import           Data.Default                            (Default)
 import qualified Data.DList                              as DL
 import           Data.Either.Extra                       (mapLeft)
 import           Data.Foldable                           (Foldable (fold))
@@ -101,7 +103,13 @@ import           GHC                                     (EpAnn (..),
                                                           spanAsAnchor)
 import           GHC.Parser.Annotation                   (AnnContext (..),
                                                           DeltaPos (SameLine),
-                                                          EpaLocation (EpaDelta))
+                                                          EpaLocation (EpaDelta),
+                                                          deltaPos)
+#endif
+
+#if MIN_VERSION_ghc(9,2,0)
+setPrecedingLines :: Default t => LocatedAn t a -> Int -> Int -> LocatedAn t a
+setPrecedingLines ast n c = setEntryDP ast (deltaPos n c)
 #endif
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


pepeiborra:
> I import from GHC.Types.Error to extract the warnings and errors, I'm not sure if that's okay, or if the idea is to avoid plugins depending directly on GHC modules

We have a compatibility layer in `Development.IDE.GHC.Compat` but it's quite confusing and it doesn't help much with exactprint types (`LocatedAn` and friends)

pepeiborra:
@eddiejessup are you able to finish this PR?

eddiejessup:
Hey @pepeiborra, thanks for checking in on this and sorry for not updating the issue. I got this compiling but had some test failures. I spent a bunch of time trying to work them out, I think it's essentially compatibility problems with the new exact printing stuff, but I had a lot of headaches nailing things down.

Maybe what I can do is push what I have, show the failing tests and my understanding so far about what's going wrong? Then I can try to find some time to pick it up again, or if someone else wants to pick it up they can.

eddiejessup:
Hi, very sorry for the long delay on this, but I've found some time to dig into this more, I have all the tests passing on GHC 9.2, though I confess one change is quite ad hoc to get the tests passing.

Though it's to do with spacing, so given the staleness of this plugin, maybe even if we get some problems with extra/missing spaces, at this point maybe that's better than nothing.

Anyway I'm now working on making the changes back-compatible with < 9.2.

eddiejessup:
Some unrelated plugin tests seem to fail in CI, but they pass for me locally. I'm not sure if this is flakiness or something, would appreciate any input on this.

I think apart from that, this work is in a reviewable state

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -101,7 +103,13 @@ import           GHC                                     (EpAnn (..),
                                                           spanAsAnchor)
 import           GHC.Parser.Annotation                   (AnnContext (..),
                                                           DeltaPos (SameLine),
-                                                          EpaLocation (EpaDelta))
+                                                          EpaLocation (EpaDelta),
+                                                          deltaPos)
+#endif
+
+#if MIN_VERSION_ghc(9,2,0)
+setPrecedingLines :: Default t => LocatedAn t a -> Int -> Int -> LocatedAn t a
```
eddiejessup, line 111: This was removed from exactprint, but we still use it for sorting out spacing, so reimplement here in later versions

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -374,7 +382,7 @@ graftWithM dst trans = Graft $ \dflags a -> do
 #if MIN_VERSION_ghc(9,2,0)
                                 val'' <-
                                     hoistTransform (either Fail.fail pure) $
-                                        annotate dflags True $ maybeParensAST val'
+                                        annotate dflags False $ maybeParensAST val'
```
eddiejessup, line 385: I made this change because there was a failing test which was expanding a type signature to:

```
f ::  Int
```

instead of,

```
f :: Int
```

(note the extra space before `Int.)

I'm not sure what changed to make this differ from the pre-9.2 change, and it might not be the right fix to be honest, but since I think it just messes up formatting, and I find it very difficult to debug this exactprint annotation stuff, I think it's either this or nothing, unless someone else wants to dig in!

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -533,7 +552,7 @@ annotate dflags needs_space ast = do
     let rendered = render dflags ast
 #if MIN_VERSION_ghc(9,2,0)
     expr' <- lift $ mapLeft show $ parseAST dflags uniq rendered
-    pure expr'
+    pure $ setPrecedingLines expr' 0 (bool 0 1 needs_space)
```
eddiejessup, line 555: Change to match pre-9.2

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -542,6 +561,7 @@ annotate dflags needs_space ast = do
 
 -- | Given an 'LHsDecl', compute its exactprint annotations.
 annotateDecl :: DynFlags -> LHsDecl GhcPs -> TransformT (Either String) (LHsDecl GhcPs)
+#if !MIN_VERSION_ghc(9,2,0)
 -- The 'parseDecl' function fails to parse 'FunBind' 'ValD's which contain
```
eddiejessup, line 565: I'm leaning quite heavily on the tests here, but seems like this splitting then merging causes problems for on `>= 9.2`

konn:
Thank you for the tough work!
 I'm a bit busy these days, but will review the changes to Splice Plugin until this weekend.
I have no idea as for failure on 9.4+Win - I'll retry the job to see if it's just an example of flakey test behaviour.

eddiejessup:
Do I understand right that the rerun checks are passing? Also I see I need to rebase, I can do that

konn (approved):
LGTM, as for the core logic of Splice Plugin itself 👍 
Thank you for the great work!
I'm not so familiar with the recent changes to ExactPrint API, so it would be good to have another approval from a ExactPrint guy.

eddiejessup:
@konn thanks! Do the current reviewers include one or more Exactprint people? Or we should add someone?

michaelpj:
I'm not sure who is, but I think given the tests all pass this seems like you've probably done a good job. Perhaps @wz1000 or @pepeiborra might want to take a look, but I think we can merge for now and if they have any comments we can address them afterwards. Thanks for fixing this!

PR #2820, opened 2022-04-10
Support stylish-haskell with ghc9.0

July541:
I have a fork that include stylish-haskell compile with ghc 9.0 https://github.com/July541/stylish-haskell/tree/ghc-9.0, and this pr uses the fork directly.

CI looks well https://github.com/July541/stylish-haskell/actions/runs/2142857127

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2820"></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -190,7 +190,7 @@ jobs:
         name: Test hls-splice-plugin
         run: cabal test hls-splice-plugin --test-options="$TEST_OPTS" || cabal test hls-splice-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-splice-plugin --test-options="$TEST_OPTS"
 
-      - if: matrix.test && matrix.ghc != '9.0.1'  && matrix.ghc != '9.0.2' && matrix.ghc != '9.2.2'
+      - if: matrix.test && matrix.ghc != '9.2.2'
         name: Test hls-stylish-haskell-plugin
         run: cabal test hls-stylish-haskell-plugin --test-options="$TEST_OPTS" || cabal test hls-stylish-haskell-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-stylish-haskell-plugin --test-options="$TEST_OPTS"
 
--- a/cabal-ghc90.project
+++ b/cabal-ghc90.project
@@ -35,6 +35,13 @@ package *
   ghc-options: -haddock
   test-show-details: direct
 
+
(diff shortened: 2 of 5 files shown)


Ailrun (approved):
LGTM :)

Ailrun:
How hard will it be to make your patch compatible with the current stylish-haskell head (which supports ghc 9.2)? If it's not so hard, could you also send a PR there?

July541:
> How hard will it be to make your patch compatible with the current stylish-haskell head (which supports ghc 9.2)? If it's not so hard, could you also send a PR there?

Not easy, I developed based on v0.13.0.0, which has a bunch of different representations, current head has a clean structure with ghc9.2. I'd prefer someone can send this fork to hackage directly and this fork should exist as a branch in haskell/stylish-haskell, because it's just a compat that only works for HLS(Others will choose v0.14 ideally)

pepeiborra:
Let's take a step back:

* Stylish-haskell uses `ghc-lib` so it's not directly coupled to the ghc version. Concretely, `stylish-haskell-0.13` uses `ghc-lib-parser-8.10` and `stylish-haskell-0.14` uses `ghc-lib-parser-9.2`
* Moreover, other HLS plugins also have `ghc-lib-parser` requirements. Concretely, `hlint-3.3.6` requires `ghc-lib-9.0` and `hlint-3.2.8` requires `ghc-lib-parser-8.10`.
* `ghc-lib-parser` versions have ghc requirements themselves. From the [project](https://github.com/digital-asset/ghc-lib) readme:
| Ghc flavor | >= version | < version |
| ---------- |:----------:|:---------:|
| ghc-8.8.*  | 8.4.4      | 8.10.1    |
| ghc-8.10.* | 8.6.5      | 9.0.1     |
| ghc-9.0.1  | 8.8.1      | 9.2.1     |
| ghc-9.2.1  | 8.10.1     |           |
* **ghc 8.10.7** We can fallback to `hlint-3.2.8` and `stylish-haskell-0.13` which both accept `ghc-lib-parser-8.10.7`
* **ghc 9.0** Since `ghc-lib-parser-8.10.7` doesn't build with ghc 9.0, we cannot use `hlint-3.2.8` nor `stylish-haskell-0.13` anymore. We have to choose between `hlint-3.3.6` and `stylish-haskell-0.14`
* **ghc 9.2** Since `ghc-lib-parser-9.0` doesn't build with ghc 9.2, we cannot use `hlint-3.3.6`. But `stylish-haskell-0.14` is fine.

The future-proof and non-forky thing to do here is to evolve `hlint` to support ghc 9.2. Once that's available, we can link both `stylish-haskell` and `hlint` in all the ghc versions from 8.10.7.

pepeiborra:
That said, I am not against merging this PR, as it doesn't make anything worse and it means that we can distribute 9.0 binaries which support both plugins.

pepeiborra:
Relevant hlint issue: https://github.com/ndmitchell/hlint/issues/1359

July541:
It's worth closing this and waiting for hlint support before the next hls release.

PR #2826, opened 2022-04-12
Should no related code lens if the module name is correct

### LINKED ISSUE #2797: `Set module name to <name>` is suggested even if the module is already called `<name>`
### Example

set_module_name

### Version

GHC version: 9.2.1
HLS version: 1.6.1.0

July541:
Close #2797.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2826"></a>

### DIFF
--- a/plugins/hls-module-name-plugin/test/Main.hs
+++ b/plugins/hls-module-name-plugin/test/Main.hs
@@ -37,6 +37,12 @@ tests =
       [CodeLens { _command = Just c }] <- getCodeLenses doc
       executeCommand c
       void $ skipManyTill anyMessage (message SWorkspaceApplyEdit)
+  , testCase "Should no code lens if the module name is correct" $
+      runSessionWithServer moduleNamePlugin testDataDir $ do
+          doc <- openDoc "CorrectName.hs" "haskell"
+          lenses <- getCodeLenses doc
+          liftIO $ lenses @?= []
+          closeDoc doc
   ]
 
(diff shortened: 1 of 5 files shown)


test/functional/Progress.hs
```diff
@@ -33,7 +33,7 @@ tests =
                 let path = "diagnostics" </> "Foo.hs"
                 _ <- openDoc path "haskell"
                 expectProgressMessages [pack ("Setting up testdata (for " ++ path ++ ")"), "Processing", "Indexing"] []
-        ,  knownBrokenForGhcVersions [GHC92] "No evaluation status with GHC 9.2" $ requiresEvalPlugin $ testCase "eval plugin sends progress reports" $
```
July541, line 36: Not know how did it pass the previous tests.

plugins/hls-module-name-plugin/test/Main.hs
```diff
@@ -37,6 +37,12 @@ tests =
       [CodeLens { _command = Just c }] <- getCodeLenses doc
       executeCommand c
       void $ skipManyTill anyMessage (message SWorkspaceApplyEdit)
+  , testCase "Should no code lens if the module name is correct" $
```
michaelpj, line 40:
```suggestion
  , testCase "Should not show code lens if the module name is correct" $
```

michaelpj (approved):
Nice! Good bug hunting.

PR #2828, opened 2022-04-14
Restore concise type variables in ghc-9.2

### LINKED ISSUE #2716: complex type variables in hover with GHC 9.2

July541:
Close #2716.

GHC 9.2 changed the default `showSDocUnsafe` style from [`defaultUserStyle`](https://hackage.haskell.org/package/ghc-9.0.2/docs/src/GHC.Utils.Outputable.html#showSDoc) to [`defaultDumpStyle`](https://hackage.haskell.org/package/ghc-9.2.1/docs/src/GHC.Utils.Outputable.html#defaultSDocContext). This pr force `showGhc` with `defaultUserStyle`, hierarchy calling result shows the following functions are affected, but only [`atPoint`](https://github.com/haskell/haskell-language-server/blob/3c527431bc10879ffa80dfea23a80299992d87cc/ghcide/src/Development/IDE/Spans/AtPoint.hs#L249-L251)(The function returns type varibale) really render [`IfaceType`](https://hackage.haskell.org/package/ghc-9.2.1/docs/GHC-Iface-Type.html#t:IfaceType) type.

image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2828"></a>

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
@@ -13,6 +13,8 @@ module Development.IDE.GHC.Compat.Outputable (
     mkPrintUnqualified,
     mkPrintUnqualifiedDefault,
     PrintUnqualified(..),
+    defaultUserStyle,
+    withPprStyle,
     -- * Parser errors
     PsWarning,
     PsError,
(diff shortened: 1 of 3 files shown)


michaelpj:
Nice, this is a good fix, but I wonder if we need to audit all the uses of `showSDocUnsafe`, like @hsyl20 [suggested](https://github.com/haskell/haskell-language-server/issues/2716#issuecomment-1056652643)? Or do you think this covers the cases that matter?

michaelpj (approved):

PR #2830, opened 2022-04-16
Unify showSDocUnsafe

July541:
We choose `printOutputable :: Outputable a => a -> Text` finally.

---

Extends #2828, this pr wrapped `showSDocUnsafe` and  unified all `SDoc` printing functions into the following four functions
1. `showGhc` has `defaultUserStyle` and `suppressUnique`, we'd use this as much as we can to print ghc internal.
2. `showGhcWithUniques` is the alias of `showSDocUnsafe`, it may have different behavior under different ghc versions, like #2716.
3. `prettyPrint` sames as `showGhc` but return `String` instead of `Text`.
4. `prettyPrintWithUnique`, the `String` version of `showGhcWithUniques`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2830"></a>

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
@@ -13,6 +13,8 @@ module Development.IDE.GHC.Compat.Outputable (
     mkPrintUnqualified,
     mkPrintUnqualifiedDefault,
     PrintUnqualified(..),
+    defaultUserStyle,
+    withPprStyle,
     -- * Parser errors
     PsWarning,
     PsError,
@@ -43,7 +45,8 @@ import           GHC.Types.SourceError
 import           GHC.Types.SrcLoc
 import           GHC.Unit.State
 import           GHC.Utils.Error                 hiding (mkWarnMsg)
-import           GHC.Utils.Outputable
+import           GHC.Utils.Outputable            as Out hiding (defaultUserStyle)
+import qualified GHC.Utils.Outputable            as Out
 import           GHC.Utils.Panic
 #elif MIN_VERSION_ghc(9,0,0)
 import           GHC.Driver.Session
@@ -52,14 +55,16 @@ import           GHC.Types.Name.Reader           (GlobalRdrEnv)
 import           GHC.Types.SrcLoc
 import           GHC.Utils.Error                 as Err hiding (mkWarnMsg)
 import qualified GHC.Utils.Error                 as Err
-import           GHC.Utils.Outputable            as Out
+import           GHC.Utils.Outputable            as Out hiding (defaultUserStyle)
+import qualified GHC.Utils.Outputable            as Out
 #else
 import           Development.IDE.GHC.Compat.Core (GlobalRdrEnv)
 import           DynFlags
 import           ErrUtils                        hiding (mkWarnMsg)
 import qualified ErrUtils                        as Err
 import           HscTypes
-import           Outputable                      as Out
+import           Outputable                      as Out hiding (defaultUserStyle)
+import qualified Outputable                      as Out
 import           SrcLoc
 #endif
 
@@ -177,4 +182,11 @@ mkWarnMsg =
   const Error.mkWarnMsg
 #else
   Err.mkWarnMsg
+#endif
+
+defaultUserStyle :: PprStyle
+#if MIN_VERSION_ghc(9,0,0)
+defaultUserStyle = Out.defaultUserStyle
+#else
+defaultUserStyle = Out.defaultUserStyle unsafeGlobalDynFlags
 #endif
--- a/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
@@ -8,9 +8,7 @@ module Development.IDE.GHC.Compat.Outputable (
     showSDocForUser,
     ppr, pprPanic, text, vcat, (<+>), ($$), empty, hang, nest,
     printSDocQualifiedUnsafe,
-    printNameWithoutUniques,
     printWithoutUniques,
-    printSDocAllTheWay,
     mkPrintUnqualified,
     mkPrintUnqualifiedDefault,
     PrintUnqualified(..),
@@ -85,16 +83,6 @@ printWithoutUniques =
       dflags = unsafeGlobalDynFlags `gopt_set` Opt_SuppressUniques
 #endif
 
-printNameWithoutUniques :: Outputable a => a -> String
-printNameWithoutUniques =
-#if MIN_VERSION_ghc(9,2,0)
-  renderWithContext (defaultSDocContext { sdocSuppressUniques = True }) . ppr
-#else
-  printSDocAllTheWay dyn . ppr
-  where
-    dyn = unsafeGlobalDynFlags `gopt_set` Opt_SuppressUniques
-#endif
-
 printSDocQualifiedUnsafe :: PrintUnqualified -> SDoc -> String
 #if MIN_VERSION_ghc(9,2,0)
 printSDocQualifiedUnsafe unqual doc =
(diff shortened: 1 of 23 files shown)


ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -318,4 +314,45 @@ traceAst lbl x
...
+#endif
 
+-- | Print a GHC value by default `showSDocUnsafe`.
+--
+-- You may prefer `prettyPrint` unless you know what you are doing.
+--
+-- It internal using `unsafeGlobalDynFlags`.
+--
+-- `String` version of `showGhcWithUniques`.
+prettyPrintWithUniques :: Outputable a => a -> String
```
michaelpj, line 331: I think basically nobody should ever want to do this, right? there's only one use, so maybe inline it there?

ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -318,4 +314,45 @@ traceAst lbl x
...
+-- `String` version of `showGhc`.
+prettyPrint :: Outputable a => a -> String
+prettyPrint = printWithoutUniques
+
+-- | Print a GHC value by default `showSDocUnsafe`.
+--
+-- You may prefer `showGhc` unless you know what you are doing.
+--
+-- It internal using `unsafeGlobalDynFlags`.
+showGhcWithUniques :: Outputable a => a -> T.Text
```
michaelpj, line 349: this is unused, maybe let's not even expose it

ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
```diff
@@ -68,14 +67,20 @@ import qualified Outputable                      as Out
 import           SrcLoc
 #endif
 
-printNameWithoutUniques :: Outputable a => a -> String
-printNameWithoutUniques =
+printWithoutUniques :: Outputable a => a -> String
```
michaelpj, line 70: haddock?

ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -318,4 +314,45 @@ traceAst lbl x
...
+showGhcWithUniques = T.pack . showSDocUnsafe . ppr
+
+-- | Print a GHC value in `defaultUserStyle` without unique symbols.
+--
+-- This is the most common print utility, will print with a user friendly style like: `a_a4ME` as `a`.
+--
+-- It internal using `unsafeGlobalDynFlags`.
+showGhc :: Outputable a => a -> T.Text
```
michaelpj, line 357: In the interests of keeping things simple, I think we could get away with just this function. Replacing `prettyPrint` just requires `show $ showGhc x` at the call site.

Also, name bikeshedding: how about `printOutputable`?

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -250,10 +251,7 @@ defDocumentSymbol l = DocumentSymbol { .. } where
   _tags           = Nothing
 
 showRdrName :: RdrName -> Text
```
michaelpj, line 253: inline?

ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -178,7 +178,7 @@ unpackAvail mn
   | otherwise = const []
   where
     !mod = pack $ moduleNameString mn
-    f id@IdentInfo {..} = (pack (prettyPrint name), moduleNameText,[id])
+    f id@IdentInfo {..} = (pack (occNameString name), moduleNameText,[id])
```
michaelpj, line 181: why different here?

michaelpj (commented):
Wonderful, thank you!

isovector:
lgtm, thanks!

ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
```diff
@@ -91,14 +96,9 @@ printSDocQualifiedUnsafe unqual doc =
     showSDocForUser unsafeGlobalDynFlags unqual doc
 #endif
 
-printSDocAllTheWay :: DynFlags -> SDoc -> String
+
 #if MIN_VERSION_ghc(9,2,0)
-printSDocAllTheWay dflags sdoc = renderWithContext ctxt sdoc
```
Ailrun, line 96: Why this `#if` is still here? I mean, we can use the negated condition, cannot we?

Ailrun (commented):
Thank you for the update!

ghcide/src/Development/IDE/LSP/Outline.hs
July541: I'm just a little worried about whether we're too radical. It turns out we also have some `occNameString` and I keep it for semantic clarity, or you're thinking to rule them all with`printOutputable`?

ghcide/src/Development/IDE/Types/Exports.hs
July541: The same one, just a lonely clear name.

ghcide/src/Development/IDE/GHC/Util.hs
July541: I keeped `printOutputable :: Outputable a => a -> String` and `printOutputableText :: Outputable a => a -> Text`. I think we shouldn't pay for extra `O(n)` to `pack`  if `Text` is not required, and also, a lot of `pack $ printOutputable x` looks inconvenient.

It is only my personal aspect, I'm happy to revise them if you have different views:)

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -55,6 +55,7 @@ import           Data.Version                         (showVersion)
 import           Development.IDE.Types.Shake          (WithHieDb)
 import           HieDb                                hiding (pointCommand)
 import           System.Directory                     (doesFileExist)
+import Development.IDE.GHC.Util (printOutputableText)
```
Ailrun, line 59: Here is some ill-formatted code

ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
```diff
@@ -91,24 +100,15 @@ printSDocQualifiedUnsafe unqual doc =
...
 oldRenderWithStyle dflags sdoc sty = Out.renderWithStyle (initSDocContext dflags sty) sdoc
 oldMkUserStyle _ = Out.mkUserStyle
 oldMkErrStyle _ = Out.mkErrStyle
 
 oldFormatErrDoc :: DynFlags -> Err.ErrDoc -> Out.SDoc
 oldFormatErrDoc dflags = Err.formatErrDoc dummySDocContext
   where dummySDocContext = initSDocContext dflags Out.defaultUserStyle
-
-#else
+#elif !MIN_VERSION_ghc(9,2,0)
```
Ailrun, line 111: How about `(9,0,0)`? I think that's more clear.

Ailrun (approved):
Thanks, looks good to me :) Here are a few very minor comments:

ghcide/src/Development/IDE/GHC/Util.hs
michaelpj: I was thinking the other way around: just have the one that prints `Text` and let people `show` it if they need a `String`. `Text` by default is a better way anyway!

I think the cost of string conversions can be assumed to be totally trivial given the size of things we're talking about! I'm more concerned with the complexity of the codebase, so the fewer functions the better...

ghcide/src/Development/IDE/LSP/Outline.hs
michaelpj: I don't know - I don't know how this code works. I just noticed that it's trivial so maybe it should be inlined.

ghcide/src/Development/IDE/Types/Exports.hs
michaelpj: I have no idea if this is sensible or not, I don't know this bit of the code! Just wondering. Probably could do with a comment.

July541: Replaced with `printOutputable`

eddiemundo (approved):
I wanna bikeshed `printOutputable` to `prettyPrintAsText` since it's the same as `prettyPrint`, or oppositely `prettyPrint` to `printOutputableAsString`, but since you already did a change once, it's fine...

So all good.

michaelpj:
Temporarily taking out of the merge queue until the release is done.

PR #2832, opened 2022-04-17
nix: bump nixpkgs and use GHC 9.2.2

guibou:
I also renamed all configuratino from GHC-XYZ to GHC-XY to ease future
transitions.

Note: it means that users of the flake must use GHC 9.2.2 (for GHC 9.2). In the future we may want to allow a limited flexibility where user may also use another minor version of GHC (when they use a different nixpkgs checkout as the one we are using in this repo).


michaelpj (approved):
CI seems unhappy, I think you missed a few references.

teto:
I think this can be closed relative to the 9.2.3 bump that was merged already

PR #2834, opened 2022-04-18
Prepare release 1.7

wz1000:
Waiting for haskell.org to come back...

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2834"></a>

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,181 @@
 # Changelog for haskell-language-server
 
+## 1.7.0.0
+
+- Distribute dynamically linked binaries for HLS to avoid statically linking against GLIBC
+  and system libraries, and to avoid unpredictable failures due to subtle differences
+  between the GHC used to compile HLS and the GHC installed on the users machine
+  (@hasufell, #2675, #2431)
+
+- Improved recompilation avoidance in projects that make use of Template Haskell (#2316). See
+  the [blog post](https://well-typed.com/blog/2022/04/hls-performance/) for more details.
+  This release includes the `avoid-recompile` set of commits described in the blog post.
+
+- Support for GHC 9.2.2
+
+- Removal of HLS installer scripts as mentioned by the deprecation notice last release (#2773)
+
+- Many more improvements and bug fixed thanks to our contributors!
+
+### Pull requests merged for 1.6.1.1
+
+- Restore concise type variables in ghc-9.2
+([#2828](https://github.com/haskell/haskell-language-server/pull/2828)) by @July541
+- Should no related code lens if the module name is correct
+([#2826](https://github.com/haskell/haskell-language-server/pull/2826)) by @July541
+- Bump cachix/install-nix-action from 16 to 17
+([#2823](https://github.com/haskell/haskell-language-server/pull/2823)) by @dependabot[bot]
+- Bump actions/upload-artifact from 2 to 3
+([#2822](https://github.com/haskell/haskell-language-server/pull/2822)) by @dependabot[bot]
+- Bump actions/download-artifact from 2 to 3
+([#2821](https://github.com/haskell/haskell-language-server/pull/2821)) by @dependabot[bot]
+- bench: Add more metrics
+([#2814](https://github.com/haskell/haskell-language-server/pull/2814)) by @wz1000
+- Enable rename plugin
+([#2809](https://github.com/haskell/haskell-language-server/pull/2809)) by @OliverMadine
+- Fix `cabal install` commands for local HLS build in docs
+([#2807](https://github.com/haskell/haskell-language-server/pull/2807)) by @9999years
+- Bump actions/cache from 2 to 3
+([#2806](https://github.com/haskell/haskell-language-server/pull/2806)) by @dependabot[bot]
+- [hls-graph] Optimise waitConcurrently
+([#2805](https://github.com/haskell/haskell-language-server/pull/2805)) by @pepeiborra
+- [bench] track changes to hls-* projects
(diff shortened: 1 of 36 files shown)


konn (approved):
LGTM as for Splice Plugin!

GenChangelogs.hs
```diff
@@ -24,6 +24,7 @@ main = do
 
   lastDateStr <- last . lines <$> readProcess "git" ["show", "-s", "--format=%cI", "-1", last tags] ""
   lastDate <- zonedTimeToUTC <$> iso8601ParseM lastDateStr
+  print lastDate
```
fendor, line 27: What's this? Is this used?

fendor (approved):
LGTM, awesome!

fendor:
Actually one question, don't we have to bump the versions of all plugins? Since we are modifying their dependency requirements and have to re-upload them to hackage?

wz1000:
> Actually one question, don't we have to bump the versions of all plugins? Since we are modifying their dependency requirements and have to re-upload them to hackage?

No, I believe hackage revisions should be sufficient for the rest of the plugins.

GenChangelogs.hs
wz1000: Oh, this is just something I added to debug the script as I had already made a tag, and so it wasn't picking up all the PRs that were merged. I will revert it.

July541:
It would be great if we can bundle hls-stylish-haskell-plugin in ghc 9.2 with the new release. https://github.com/haskell/haskell-language-server/pull/2836

And, should we enabling it in ghc9.0 with a compromise solution? https://github.com/haskell/haskell-language-server/pull/2820

hasufell:
@wz1000 try

```sh
ghcup config set url-source '{"OwnSource": "https://mirror.sjtu.edu.cn/ghcup/yaml/ghcup/data/ghcup-0.0.6.yaml"}'
```

to enable a mirror.

wz1000:
> ghcup config set url-source '{"OwnSource": "https://mirror.sjtu.edu.cn/ghcup/yaml/ghcup/data/ghcup-0.0.6.yaml"}'

Thanks. See https://gitlab.haskell.org/haskell/haskell-language-server/-/pipelines/50432

wz1000:
> It would be great if we can bundle hls-stylish-haskell-plugin in ghc 9.2 with the new release. #2836

Sure , we can include this PR.

> And, should we enabling it in ghc9.0 with a compromise solution? #2820

Its not exactly clear to me what the compromise is.

July541:
> Its not exactly clear to me what the compromise is.

The stylish-haskell 9.0 branch is on my fork, migrate to the official repo is not cheap.

It is worth waiting if we are not eager. https://github.com/haskell/haskell-language-server/pull/2820#issuecomment-1094340713

pepeiborra:
> > Actually one question, don't we have to bump the versions of all plugins? Since we are modifying their dependency requirements and have to re-upload them to hackage?
> 
> No, I believe hackage revisions should be sufficient for the rest of the plugins.

`policeman` (build from HEAD) is great for reviewing version bumps, but sadly I have found it doesn't work very well for HLS. If you find a way to make it work nicely, please share!

michaelpj:
> No, I believe hackage revisions should be sufficient for the rest of the plugins.

Is this a good idea? Won't that make the previous version unbuildable? Since the versions of the plugins which worked with it before will now not work with it since you'll have mutated the version bounds.

ChangeLog.md
```diff
@@ -1,5 +1,181 @@
 # Changelog for haskell-language-server
 
+## 1.7.0.0
+
+- Distribute dynamically linked binaries for HLS to avoid statically linking against GLIBC
+  and system libraries, and to avoid unpredictable failures due to subtle differences
+  between the GHC used to compile HLS and the GHC installed on the users machine
+  (@hasufell, #2675, #2431)
```
Ailrun, line 8: We may need to change https://github.com/haskell/haskell-language-server/blob/9fca35d7277ff4e459b39e0c503931c9339fe394/docs/troubleshooting.md#static-binaries this doc as well

Ailrun (approved):
LGTM. Here's my minor comment:

pepeiborra:
> > No, I believe hackage revisions should be sufficient for the rest of the plugins.
> 
> Is this a good idea? Won't that make the previous version unbuildable? Since the versions of the plugins which worked with it before will now not work with it since you'll have mutated the version bounds.

I second this, previous versions would not become unbuildable if the change was only increasing the upper bound. I question the choice of clamped `^>` version bounds.

wz1000:
> > Is this a good idea? Won't that make the previous version unbuildable? Since the versions of the plugins which worked with it before will now not work with it since you'll have mutated the version bounds.
> 
> I second this, previous versions would not become unbuildable if the change was only increasing the upper bound. I question the choice of clamped `^>` version bounds.

Thanks for noticing this, I will relax the bounds on the plugins which haven't been bumped.

eddiemundo (approved):
LGTM!

isovector (approved):
lgtm for tactics

wz1000:
I have relaxed the bounds on all the packages for which the version wasn't bumped.

michaelpj (approved):

ChangeLog.md
wz1000: I don't think that is needed, since the same caveats about static binaries apply. The difference is that this time we aren't distributing static binaries but are distributing dynamic binaries. If someone tries to compile static binaries on their own then the caveat still applies (and they will get a warning popup).

Ailrun: @wz1000 I mean, the document mentions "local building" as the easiest way to get a dynamic binary, but now it's probably easier to just download the binary.

PR #2835, opened 2022-04-18
Fix flaky test 

### LINKED ISSUE #2831: Test "add missing module (non workspace)" is very flaky
It consistently fails on CI for #2789 but I can't reproduce it locally. I'm disabling it for now.

pepeiborra:
"add missing modules (non workspace)" became flaky after the recent changes to snapshot VFS data.

The test calls `createDoc` to open a `module B imports A` where `A` does not exist, then waits for the missing module error and calls `createDoc` to open a `module A`. Importantly, both `B` and `A` live outside the workspace and thus are not subject to `FileWatched` notifications. 

Unfortunately, the test wasn't testing the right thing because `createDoc` sends a `FileWatched` notification right before sending a `DidOpen` notification! This was causing a race condition, where the `FileWatched` notification triggers a build that may or may not run the `FileExists ModuleA` key:
* If it does, then the key will resolve to `False` since `ModuleA` exists neither in the file system nor in VFS, and the test will fail since the key will not get rebuilt again (nothing dirties it).
* If the build gets aborted before it does, then the key will get computed in the next build (for the `DidOpen` notification) and will return true because the file now exists in VFS.

I imagine that the test was passing before since the VFS state would get updated concurrently, and the race condition would not get triggered in practice.

To fix the test, we simply tell `lsp-test` not to send `FileWatched` notifications.

I also found an old comment stating that "VFS existence is not enough, a file must exist in disk". This is clearly not true anymore, so I have deleted it.

Fixes #2831 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2835"></a>

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -423,6 +423,10 @@ diagnosticTests = testGroup "diagnostics"
       _ <- createDoc "ModuleA.hs" "haskell" contentA
       expectDiagnostics [("ModuleB.hs", [])]
   , testCase "add missing module (non workspace)" $
+    -- By default lsp-test sends FileWatched notifications for all files, which we don't want
+    -- as non workspace modules will not be watched by the LSP server.
+    -- To work around this, we tell lsp-test that our client doesn't have the
+    -- FileWatched capability, which is enough to disable the notifications
     withTempDir $ \tmpDir -> runInDir'' lspTestCapsNoFileWatches tmpDir "." "." [] $ do
       let contentB = T.unlines
             [ "module ModuleB where"
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -95,7 +95,7 @@ import           System.Process.Extra                     (CreateProcess (cwd),
 import           Test.QuickCheck
 -- import Test.QuickCheck.Instances ()
 import           Control.Concurrent.Async
-import           Control.Lens                             (to, (^.))
+import           Control.Lens                             (to, (^.), (.~))
 import           Control.Monad.Extra                      (whenJust)
 import           Data.Function                            ((&))
 import           Data.IORef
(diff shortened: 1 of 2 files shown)


wz1000 (approved):
Thanks for looking into this! I ran the testsuite a couple of times but could not reproduce this failure locally at all.

wz1000 (approved):

ghcide/test/exe/Main.hs
```diff
@@ -421,16 +421,16 @@ diagnosticTests = testGroup "diagnostics"
       let contentA = T.unlines [ "module ModuleA where" ]
       _ <- createDoc "ModuleA.hs" "haskell" contentA
       expectDiagnostics [("ModuleB.hs", [])]
-  , ignoreTestBecause "Flaky #2831" $ testSessionWait "add missing module (non workspace)" $ do
+  , testSession' "add missing module (non workspace)" $ \tmpDir -> do
       -- need to canonicalize in Mac Os
```
michaelpj, line 425: dead comment

michaelpj:
> To fix the test, we simply tell lsp-test not to send FileWatched notifications.

I'm confused, that doesn't look like what the change does? And I don't see the comment you mentioned being deleted?

pepeiborra:
> I'm confused, that doesn't look like what the change does? And I don't see the comment you mentioned being deleted?

Fixed now, thanks for noticing this. I must have forgotten to `git push -f`

ghcide/test/exe/Main.hs
```diff
@@ -421,9 +422,8 @@ diagnosticTests = testGroup "diagnostics"
       let contentA = T.unlines [ "module ModuleA where" ]
       _ <- createDoc "ModuleA.hs" "haskell" contentA
       expectDiagnostics [("ModuleB.hs", [])]
-  , ignoreTestBecause "Flaky #2831" $ testSessionWait "add missing module (non workspace)" $ do
-      -- need to canonicalize in Mac Os
-      tmpDir <- liftIO $ canonicalizePath =<< getTemporaryDirectory
+  , testCase "add missing module (non workspace)" $
+    withTempDir $ \tmpDir -> runInDir'' lspTestCapsNoFileWatches tmpDir "." "." [] $ do
```
michaelpj, line 426: This definitely deserves a comment, it's very non-obvious why we would run this with restricted capabilities (after all, it took lots of debugging and reasoning for you to figure out that we needed it!).

michaelpj (approved):

Issue #2824, opened 2022-04-11
Completions broken for pre-qualified import


Jashweii:
```Haskell
import qualified Control.Monad as M(|)
                                    ^ cursor
```
Doesn't suggest imports from Control.Monad, unlike post-qualified and unqualified imports. 
I don't think it matters if "as" is present or not.

I get this on both: 
HLS current master / GHC 9.2.2 / WSL Ubuntu / vscode (WSL)
HLS 1.6.1.0 / GHC 9.2.1 / Windows / vscode

xsebek:
Hi @Jashweii, thanks for the bug report. :slightly_smiling_face: 

The issue seems to be with `ghcide` which provides the completion.

I wrote a test case in PR #2834, but I have not yet figured out where the problem is.

xsebek:
Mystery resolved!
```haskell
-- IDE/Plugin/Completions/Logic.hs
-- getCompletions
let moduleName = T.pack $ words (T.unpack fullLine) !! 1
```

Jashweii:
This is fixed for me as of hls-2.6.0.0 on ghc-9.8.1

PR #2838, opened 2022-04-18
Fix completion for qualified import

xsebek:
- add the missing test case for #2824
- fix how we get the module name considering it can be preceded by `qualified`
- use parsed context for import completions
  - add regression test for fixed multiline import
- refactor `getCompletions` function

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2838"></a>


michaelpj:
Can we mark this as failing somehow so we can merge it without the fix?

xsebek:
@michaelpj there are other tests marked as such, so it can surely be done.

The second one should be passing though, so I need to check what I am testing. :sweat_smile:

xsebek:
@michaelpj I marked it as expected to fail, but then I found the fix anyway :grinning:

ghcide/test/exe/Main.hs
```diff
@@ -4781,11 +4781,8 @@ completionNoCommandTest name src pos wanted = testSession name $ do
   docId <- createDoc "A.hs" "haskell" (T.unlines src)
   _ <- waitForDiagnostics
   compls <- getCompletions docId pos
-  let wantedC = find ( \case
-            CompletionItem {_insertText = Just x} -> wanted `T.isPrefixOf` x
-            _                                     -> False
-            ) compls
-  case wantedC of
+  let isPrefixOfInsertOrLabel ci = any (wanted `T.isPrefixOf`) [fromMaybe "" (ci ^. Lens.insertText), ci ^. Lens.label]
```
xsebek, line 4784: I made this helper function a little more lenient, but I think this is fine:

```haskell
CompletionItem
 { _label :: Text            -- ^ The label of this completion item. By default also
                             -- the text that is inserted when selecting this
                             -- completion.
 , _insertText :: Maybe Text -- ^ A string that should be inserted a
                             -- document when selecting this completion.
                             -- When `falsy` the label is used.
 -- ...
 }
```

xsebek:
Btw. I have no idea what is CI trying to tell me:
DA807326-035B-4D06-9D7B-E9B2F14E07C7

michaelpj:
I reran the failed jobs, let's see what happens.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -645,10 +648,10 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
   if
     -- TODO: handle multiline imports
     | "import " `T.isPrefixOf` fullLine
-      && (List.length (words (T.unpack fullLine)) >= 2)
-      && "(" `isInfixOf` T.unpack fullLine
+      && isJust (getModuleName fullLine)
+      && "(" `T.isInfixOf` fullLine
```
michaelpj, line 652: You can use pattern guards here to avoid the `fromJust`:
```
    | "import " `T.isPrefixOf` fullLine
    , Just moduleName <- getModuleName fullLine
    , "(" `T.isInfixOf` fullLine
```

ghcide/test/exe/Main.hs
michaelpj: Yeah, that seems reasonable.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -636,6 +637,8 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         , enteredQual `T.isPrefixOf` original label
         ]
 
+      getModuleName line = let ws = filter (/= "qualified") (T.words line)
+                            in if List.length ws >= 2 then Just (ws !! 1) else Nothing
```
michaelpj, line 641:
```suggestion
                            in ws !? 1
```
using https://hackage.haskell.org/package/extra-1.7.10/docs/Data-List-Extra.html#v:-33--63-

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -636,6 +637,8 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
         , enteredQual `T.isPrefixOf` original label
         ]
 
+      getModuleName line = let ws = filter (/= "qualified") (T.words line)
```
michaelpj, line 640: A comment explaining why this works would be good. Also, I don't really understand why we're doing sort-of-parsing here, don't we have a `ParsedModule` lying around?

michaelpj (approved):

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
xsebek: I forgot about pattern guards! [Brilliant!](https://youtu.be/cEpKcBkkVMY?t=8) (Nice that it's in Haskell2010 :slightly_smiling_face: )

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
xsebek: Nice :+1:

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
xsebek: `Maybe` we do. I added cases for when we can use the `ParsedModule` to get the module name. :slightly_smiling_face:

xsebek:
@michaelpj Good news, using the parsed context seems to have fixed the problem with multiline import completions! :+1:

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -560,24 +560,74 @@ getCompletions
...
+    moduleExportsMap
+    -- ------------------------------------------------------------------------
+    -- IMPORT MODULENAME (NAM|)
+    | Just (ImportListContext moduleName) <- maybeContext
+    = moduleImportListCompletions moduleName
+
+    | Just (ImportHidingContext moduleName) <- maybeContext
+    = moduleImportListCompletions moduleName
+
+    -- TODO: Is manual parsing ever needed or is context always present for module?
```
michaelpj, line 582: cc @pepeiborra . I don't know how common it is to not have a `ParsedModule` at all. Won't we almost certainly get a stale one? That seems like it should be good enough for e.g. the import context!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -587,20 +637,33 @@ getCompletions plId ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls, qu
           $ (if T.null enteredQual then id else mapMaybe (T.stripPrefix enteredQual))
             allModNamesAsNS
 
-      filtCompls = Fuzzy.filter chunkSize maxC prefixText ctxCompls (label . snd)
-        where
-
-          mcc = case maybe_parsed of
+      -- ----------------------------------------
+      -- Note: correct the cursorPos by moving
```
michaelpj, line 641: this comment doesn't seem to be in the right place

michaelpj (commented):
> Good news, using the parsed context seems to have fixed the problem with multiline import completions! 

Could we add a regression test for this?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
xsebek: Well, it never was. It used to be:
```Haskell
{- COMMENT -}
pos = cursorPos prefixInfo
```
It probably hints at something going on in the `maybeContext` right below or `filterCompls` below it.

xsebek:
The CI smiles on me this day! :sunny: 

@pepeiborra could you please check the comments above? :slightly_smiling_face:

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
pepeiborra: The caller is using `useWithStaleFast GetParsedModule`: We will always get a parsed module.

xsebek: @pepeiborra At least for import completions that seems* to be true, but if I tried limiting this entire function to `Just (parsedModule, positionMapping)` then some tests did fail. :confused: 
<sub>*) I am a bit afraid we just don't test this enough.</sub>

xsebek: @pepeiborra Well CI proved me wrong - HLS has a test that does not pass without manual parsing:
```
haskell-language-server
  completions
    import function completions:        FAIL (3.23s)
      src/Test/Hls/Util.hs:386:
      Completion with label "Alternative" not found in ["ALRTransitionalBit","ALoad" {- many more -}]
      Use -p '/import function completions/' to rerun this test only.
```
Confusingly though in CI it did pick up some `Alternative` that was a struct:
```
 haskell-language-server
  completions
    import function completions: FAIL (2.09s)
      test/functional/Completion.hs:135:
      expected: Just CiFunction
       but got: Just CiStruct
```

xsebek:
After inspecting the CI (see [comment above](https://github.com/haskell/haskell-language-server/pull/2838#discussion_r889601561)), I decided to keep the manual parsing.

@pepeiborra @michaelpj If the CI passes, I think it is ready to merge. :eyes:

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
pepeiborra: The only place where there won't be a parsed module is at file open, before a module has been parsed at least once. This test suite (`test/functional/Completion.hs`) is accidentally exercising that use case non-deterministically, because it doesn't wait for progress done. In comparison, the completion tests in the ghcide test suite (`ghcide/test/exe/Main.hs`) carefully wait for progress done.

xsebek: Thanks for the explanation, I will see if I can get the tests to pass with only parsed module completions :+1:

michaelpj:
I restarted the tests, and now the 8.6.5 one fails as well in the same way. It also doesn't fail except on ubuntu. Perhaps it's just flaky, and we're missing a call to wait for something somewhere?

pepeiborra:
Consistent failure:

```

Building test suite 'func-test' for hls-1.7.0.0..
285
Running 1 test suites...
286
Test suite func-test: RUNNING...
287
haskell-language-server
288
  completions
289
    import function completions: FAIL (2.28s)
290
      test/functional/Completion.hs:144:
291
      expected: Just CiFunction
292
       but got: Just CiStruct
293

294
1 out of 1 tests failed (2.28s)
295
Test suite func-test: FAIL

```

xsebek:
Sorry for leaving this open for so long!

@pepeiborra I fixed the test failing on `CiStruct` but after rebase on `upstream/master` I could not find the test again. 😕

I will wait to see what the CI says this time and hopefully finally get this merged.

fendor:
> I fixed the test failing on CiStruct but after rebase on upstream/master I could not find the test again.

@xsebek That's my fault, we cleaned up the `func-test` suite, and apparently I misjudged that the completion tests are redundant.

xsebek:
@fendor thanks for the quick response. 👍

The CI did not finish well:
* on macOS all tests were skipped, but are shown as green
* ubuntu with GHC 9.2 failed on one flaky test `sends indefinite progress notifications`
* Nix failed on ubuntu because "No space left on device"
* Nix failed on macOS because of `fourmolu` dependency

Any tips to make the CI green?

fendor:
We don't run the tests on macos (honestly, I don't know why, this will surely bite us at some point) and nix CI is optional. It will actually be reduced #3804 quite soon, so don't worry about that.

I restarted the test suite for the flaky test, sounds like this PR *is* green and doesn't require your input at the moment.

fendor (approved):
The changes look good to me. Am I correct in the assumption, the way this actually fixes the issue is by using `maybeContext` parsing the module and reliably giving a module name?

This PR has the trade-off that we sometimes can't suggest a module correctly if there is a parse error in the file, for example,

```haskell
module Test where 

import
import 
--      ^ Put your cursor here
```
will not suggest imports. That's a change in behaviour, but I don't think it is a blocking one.

I think we should document, perhaps in the function documentation of `getCompletions`, which classes of completions are dependent on an existing parsed module and which are not?

xsebek:
@fendor Yes, I switched to using the (hopefully) parsed context, which makes it work in more situations including multiline.

Maintaining the ad-hoc parsing does not seem worthwhile to me. I assume GHC does not stop on first parse error, so anything that it can manage to parse we will be able to use for suggestions. Your example seems to me like something GHC could (and should) handle better than HLS.

The rest of the completions seem to use a mix of HIE AST (`recordDotSyntaxCompls`) and ad-hoc text parsing (`enteredQual`). I don't know the detail, so my guess might be more misleading than useful. 😕

fendor:
> I assume GHC does not stop on first parse error

GHC *does* stop after the first parse error.

I think that's fine by me. A Haskell file shouldn't have parse errors in it any way.

PR #2844, opened 2022-04-20
Don't refresh HieDb while querying hierarchy results

### LINKED ISSUE #2843: hls-call-hierarchy-plugin crashed in ghcide directory
hls-call-hierarchy-plugin sometimes crashed while querying in ghcide directory, logs here:

```2022-04-20T08:11:15.236744Z | Debug | Finished: CallHierarchy.prepareHierarchy Took: 0.00s
Finished: CallHierarchy.prepareHierarchy Took: 0.00s
GHC error in desugarer lookup in Development.IDE.Core.FileStore:
  Failed to load interface for ‘Development.IDE.Core.RuleTypes’
  It is not a module in the current program, or in any known package.
[Error - 4:11:15 PM] Request callHierarchy/incomingCalls failed.
  Message: Exception in plugin PluginId "callHierarchy"while processing SCallHierarchyIncomingCalls: haskell-language-server: panic! (the 'impossible' happened)
  (GHC version 9.2.2:
	initDs

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug

  Code: -32603 
2022-04-20T08:11:15.465907Z | Debug | Finished: CallHierarchy.incomingCalls Took: 0.01s
Finished: CallHierarchy.incomingCalls Took: 0.01s
```

HLS: master
GHC: 9.2.2

July541:
These legacy codes are used to force update `hiedb` while querying hierarchy results, deleting now since users should save their changes before querying call hierarchy.

Will close #2843.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2844"></a>

### DIFF
--- a/plugins/hls-call-hierarchy-plugin/README.md
+++ b/plugins/hls-call-hierarchy-plugin/README.md
@@ -23,6 +23,8 @@ Enabled by default. You can disable it in your editor settings whenever you like
 ```
 
 ## Change log
+### 1.0.3.0
+Remove force update `HieDb` logic in queries.
 ### 1.0.1.0
 - Support call from a type signature.
 - Support call from a function pattern.
--- a/plugins/hls-call-hierarchy-plugin/hls-call-hierarchy-plugin.cabal
+++ b/plugins/hls-call-hierarchy-plugin/hls-call-hierarchy-plugin.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               hls-call-hierarchy-plugin
-version:            1.0.2.0
+version:            1.0.3.0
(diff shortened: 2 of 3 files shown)


michaelpj:
Do you have a test case that reproduces the problem that we could use as a regression test?

July541:
Sorry but I can't, I only met this in ghcide directory. I guess it requires a big scale code base, and something pushed into hiedb is incorrect, apparently, the code I'm deleting tried to operate hiedb. Since users should save their changes before querying call hierarchy, we don't need to update hiedb forcibly in `hls-call-hierarchy-plugin`.

wz1000 (approved):

PR #2848, opened 2022-04-22
Fix Completion document format

### LINKED ISSUE #683: Completion provenance for local definition is Just ModuleName instead of just ModuleName
To reproduce:

Create the following module:

```haskell
module Foo
  where

bar = ()

foo = ba
```

- Trigger completion for `ba`
- Observe that the provenance of `bar` is `Just Foo` instead of just `Foo`

image

July541:
This pr contributes to repairing broken markdown format in completion documents, and close #683 since this bug does not exist.

The function `spanDocToMarkdown` is both used in completion and hover, hover looks well, let's look at the tests.

Here are some comparations:

---

### Scenario 1
old1
new1

---

### Scenario 2
old2
new2

---

### Scenario 3
old3
new3

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2848"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Spans/Common.hs
+++ b/ghcide/src/Development/IDE/Spans/Common.hs
@@ -78,6 +78,10 @@ data SpanDocUris =
 emptySpanDoc :: SpanDoc
 emptySpanDoc = SpanDocText [] (SpanDocUris Nothing Nothing)
 
+-- | Convert `SpanDoc` to Markdown format.
+--
+-- Return a list `Text` includes haddock, document uri and source code uri,
+-- each item can be empty and must end with '\\n' if exist.
 spanDocToMarkdown :: SpanDoc -> [T.Text]
 spanDocToMarkdown = \case
     (SpanDocString docs uris) ->
@@ -87,7 +91,7 @@ spanDocToMarkdown = \case
   where
     go [] uris = spanDocUrisToMarkdown uris
     go txt uris = init txt <> [render (last txt)] <> spanDocUrisToMarkdown uris
-    -- If the doc is not end with an `'\n'`, we append it.
+    -- If the doc is not end with an '\n', we append it.
     render txt
       | T.null txt = txt
       | T.last txt == '\n' = txt
--- a/ghcide/src/Development/IDE/Spans/Common.hs
+++ b/ghcide/src/Development/IDE/Spans/Common.hs
@@ -81,7 +81,16 @@ emptySpanDoc = SpanDocText [] (SpanDocUris Nothing Nothing)
 -- | Convert `SpanDoc` to Markdown format.
 --
 -- Return a list `Text` includes haddock, document uri and source code uri,
--- each item can be empty and must end with '\\n' if exist.
+-- each item can be empty and must end with '\\n' if exist. This is to prevent
+-- subsequent render problem caused by the missing newline.
+--
+-- Example:
+--
+-- For return value ["xxxx","yyyy"], if we concat the list with inserting
+-- a separate line(note by "---\n"),
+-- it will result "xxxx---\nyyyy" and can't be rendered as a normal doc.
+-- Therefore we check every item in the value to make sure they all end with '\\n',
+-- this makes "xxxx\n---\nyyy\n" and can be rendered correctly.
 spanDocToMarkdown :: SpanDoc -> [T.Text]
 spanDocToMarkdown = \case
     (SpanDocString docs uris) ->
@@ -89,8 +98,8 @@ spanDocToMarkdown = \case
         in  go [doc] uris
     (SpanDocText txt uris) -> go txt uris
   where
-    go [] uris = spanDocUrisToMarkdown uris
-    go txt uris = init txt <> [render (last txt)] <> spanDocUrisToMarkdown uris
+    go [] uris = render <$> spanDocUrisToMarkdown uris
+    go txt uris = init txt <> [render (last txt)] <> (render <$> spanDocUrisToMarkdown uris)
     -- If the doc is not end with an '\n', we append it.
     render txt
       | T.null txt = txt
--- a/ghcide/src/Development/IDE/Spans/Common.hs
+++ b/ghcide/src/Development/IDE/Spans/Common.hs
@@ -62,7 +62,9 @@ safeTyThingId _                                = Nothing
 -- Possible documentation for an element in the code
 data SpanDoc
   = SpanDocString HsDocString SpanDocUris
+    -- ^ Extern module doc
   | SpanDocText   [T.Text] SpanDocUris
+    -- ^ Local module doc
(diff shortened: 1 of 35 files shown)


ghcide/test/exe/Main.hs
```diff
@@ -5230,6 +5231,87 @@ projectCompletionTests =
...
+  , broken $ testSession "extern mulit line doc" $ do
+      doc <- createDoc "A.hs" "haskell" $ T.unlines
+        [ "module A where"
+        , "foo = i"
+        ]
+      let expected = "*Imported from 'Prelude'*\n* * *\n\n\nIdentity function. \n```haskell\nid x = x\n```\n* * *\n[Documentation](file:"
+      test doc (Position 1 7) "id" (Just $ T.length expected) [expected]
+  ]
+  where
+    broken = knownBrokenForGhcVersions [GHC90, GHC92] "Completion doc doesn't support ghc9"
```
michaelpj, line 5301: :o do we have a ticket for this?

ghcide/src/Development/IDE/Spans/Common.hs
```diff
@@ -77,12 +79,19 @@ emptySpanDoc :: SpanDoc
...
-spanDocToMarkdown (SpanDocText txt uris) = txt <> ["\n"] <> spanDocUrisToMarkdown uris
+spanDocToMarkdown = \case
+    (SpanDocString docs uris) ->
+        let doc = T.pack $ haddockToMarkdown $ H.toRegular $ H._doc $ H.parseParas Nothing $ unpackHDS docs
+        in  go [doc] uris
+    (SpanDocText txt uris) -> go txt uris
+  where
+    go [] uris = spanDocUrisToMarkdown uris
+    go txt uris = init txt <> [render (last txt)] <> spanDocUrisToMarkdown uris
+    -- If the doc is not end with an `'\n'`, we append it.
```
michaelpj, line 90: This logic is a bit confusing, because we have a collection of lines, some of them ending with newlines. The old comment *is* correct, we do need two newlines to get a paragraph break in markdown, but that works out here because `T.unlines [ "foo\n", "bar" ]` will end up with two newlines. Confusing. 

In fact, should the lines _ever_ have trailing newlines? That seems odd.

I wonder if it would be simpler if this function just dealt with a single `Text` representing the whole piece of doc. AFAICT it would be fine if `spanDocToMarkdown` returned just a `Text`.

ghcide/test/exe/Main.hs
July541: No, I just found this https://github.com/haskell/haskell-language-server/blob/da0379b78caff2a3bc72e08b97b9f608fe73d50e/ghcide/src/Development/IDE/Spans/Documentation.hs#L104-L122

, ghc9.0 test fails on my machine, so I marked it as broken

ghcide/src/Development/IDE/Spans/Common.hs
July541: 1. Two newlines to get a new paragraph is correct, but one `\n` is enough to get a section separator(xxx\n***\nxxx)
2. My idea is: we just ensure the last line should end up with `\n`, so the separator can rendered correctly. https://github.com/haskell/haskell-language-server/blob/da0379b78caff2a3bc72e08b97b9f608fe73d50e/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs#L211-L213

July541: >I wonder if it would be simpler if this function just dealt with a single Text representing the whole piece of doc. AFAICT it would be fine if spanDocToMarkdown returned just a Text.

Unfortunately, `spanDocToMarkdown` is both used in completion and hover, and they have different UI(completion uses section separator, hover uses new line), so we can't return a `Text` in `spanDocToMarkdown`. Furthermore, the final documentation is the combination of `Defined in`, haddock, and source link.

pepeiborra: Does it return a list of sections then? Please add a Haddock comment clarifying what `spanDocToMarkdown` returns

ghcide/test/exe/Main.hs
pepeiborra: Note that this code path is only used to get documentation for local completions (from the same module). I dont' think that it is being used in your test.

Are you in Mac OS? Some ghc 9.x distributions are shipping without interface haddocks

July541:
Failure reason:
- windows-ghc9.2 can't deal local doc (ubuntu-ghc9.2 works fine)
- ubuntu-ghc9.0 can't deal local doc

I'll remove local doc test instead of just testing whether the defined doc(`Defined at`) ends with '\n', it weakens the test power:(

ghcide/src/Development/IDE/Spans/Common.hs
July541: Done

July541:
> Failure reason:
> 
> * windows-ghc9.2 can't deal with extern doc (ubuntu-ghc9.2 works fine)
> * ubuntu-ghc9.0 can't deal with local doc
> 
> I'll remove doc test instead of just testing whether the defined doc(`Defined at`) ends with '\n', it weakens the test power:(

I began to doubt the necessity of so many tests. In fact, it was just testing whether the "defined doc" ended with '\ n'.

pepeiborra:
> Failure reason:
> 
> * windows-ghc9.2 can't deal with extern doc (ubuntu-ghc9.2 works fine)

This could be the same as Mac OS, where the ghc 9.2.1 and ghc 9.0.2 distributions shipped without haddock docs in interfaces: https://gitlab.haskell.org/ghc/ghc/-/issues/20903

> * ubuntu-ghc9.0 can't deal with local doc
> 
> I'll remove doc test instead of just testing whether the defined doc(`Defined at`) ends with '\n', it weakens the test power:(

Preferably keep the full test, mark it as broken, and then add a reduced test that only checks for the final `'\n'`.

July541:
OK, doc tests should exist, I'll check each ghc and platform.

ghcide/src/Development/IDE/Spans/Common.hs
```diff
@@ -76,13 +78,24 @@ data SpanDocUris =
 emptySpanDoc :: SpanDoc
 emptySpanDoc = SpanDocText [] (SpanDocUris Nothing Nothing)
 
+-- | Convert `SpanDoc` to Markdown format.
+--
+-- Return a list `Text` includes haddock, document uri and source code uri,
+-- each item can be empty and must end with '\\n' if exist.
```
pepeiborra, line 84: It would be good to explain why each item must end with a newline.

Also, I don't think all the items end with a newline - the ones produced by `spanDocUrisToMarkdown` don't, do they ?

pepeiborra (approved):
Thanks for the fix, and for reworking the broken tests support in the testsuite 😍 !

PR #2850, opened 2022-04-23
Add scripts for releases and final 1.7 tweaks

wz1000:
1. Disable 8.8.4 build in windows gitlab CI
   The job fails for inscrutable reason, binaries will have to be manually compiled in a VM.
2. Bump hls-test-utils to 1.3
   Also need to bump the minor versions of all packages which were not bumped since 1.6.1.1
   and had changes to their testsuite.
3. Add release scripts to download, hash, sign and upload releases to downloads.haskell.org


pepeiborra (approved):
Thanks!

fendor (approved):
LGTM, thank you for the insane amount of time you are putting in!

eddiemundo (approved):
Yes, thank you!

isovector (approved):

.gitlab/tar.sh
```diff
@@ -10,7 +10,7 @@ ls -la out/
 TARBALL_PREFIX="haskell-language-server"
 case "${TARBALL_EXT}" in
     zip)
-		HLS_VERSION="$("$CI_PROJECT_DIR/out/haskell-language-server-wrapper" --numeric-version)"
+		HLS_VERSION="$("$CI_PROJECT_DIR/out/haskell-language-server-8.10.7" --numeric-version)"
```
hasufell, line 13: Why this change? This will break CI once 8.10.7 is deprecated, no?

GNUmakefile
```diff
@@ -66,7 +66,7 @@ define set_rpath
 endef
 
 hls: bindist/ghcs
-	for ghc in $(shell [ -e "bindist/ghcs-`uname`" ] && cat "bindist/ghcs-`uname`" || cat "bindist/ghcs") ; do \
+	for ghc in $(shell [ -e "bindist/ghcs-`uname -o`" ] && cat "bindist/ghcs-`uname -o`" || cat "bindist/ghcs") ; do \
```
hasufell, line 69: `-o` is not a posix option afais: https://pubs.opengroup.org/onlinepubs/009695399/utilities/uname.html

release/upload.sh
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env bash
```
hasufell, line 1: Note that `/usr/bin/env` is not guaranteed by posix either, contrary to popular belief of nix users.

hasufell (commented):
I only skimmed over the python/bash scripts. Since they're not mandatory, I guess it's fine.

.gitlab/tar.sh
wz1000: If hls-wrapper is compiled with 9.2.2 then the job fails because it can't find libgcc_s_seh-1.dll

hasufell: Yeah, this should have been fixed upstream weeks ago.

GNUmakefile
wz1000: I tried it on Windows and FreeBSD, and it seems like the only way to detect mingw without a version number so I think it is OK.

PR #2853, opened 2022-04-25
nix flake update & bump ghc 921 -> 922 (nix only)

smunix:
ghc921 is no longer available on `nixpkgs-unstable`,  so moving to ghc922

Edit: see <a href="https://github.com/NixOS/nixpkgs/pull/163426">haskell.compiler: ghc921 -> ghc922</a>

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2853"></a>


guibou:
Please, continue your MR or take ownership of #2832, I'm not sure I'll have time to fix #2832 soon.

smunix:
It looks like this is working now, but the nix build phases are taking long to complete. This can be explained by the fact we're building off of new source versions that aren't in cachix caches yet. 

Also, the `build all binaries` stage builds all the dependencies found for multiple GHC versions: 8.8.4, 8.10.7, 9.0.2, 9.2.2. Each of these versions are no longer hosted into nix caches, so we're building all of these dependencies from scratch again and again. 

Lastly, for each of the steps above, we do build each Haskell module twice: the profiling and the non-profiling (this is set so by default in nix).

Anyone has suggestions for what can be done to reduce the build time?

image

cabal-ghc92.project
```diff
@@ -39,7 +39,7 @@ package *
 
 write-ghc-environment-files: never
 
-index-state: 2022-04-27T09:22:49Z
+index-state: 2022-04-30T21:02:45Z
```
smunix, line 42: needed to pull the latest `ptr-poker`

hie-compat/hie-compat.cabal
```diff
@@ -24,8 +24,8 @@ library
   default-language:    Haskell2010
   build-depends:
      base < 4.17, array, bytestring, containers, directory, filepath, transformers
-  if flag(ghc-lib)
-    build-depends: ghc-lib
+  if flag(ghc-lib) && impl(ghc < 9)
```
smunix, line 27: This (and other .hs related changes) was introduced after a rebase from one of @pepeiborra branch.

smunix:
> It looks like this is working now, but the nix build phases are taking long to complete. 
> ...
> Anyone has suggestions for what can be done to reduce the build time?

These builds got cancelled after running for 6 hours.

guibou:
> It looks like this is working now, but the nix build phases are taking long to complete. This can be explained by the fact we're building off of new source versions that aren't in cachix caches yet.
> 
> Also, the `build all binaries` stage builds all the dependencies found for multiple GHC versions: 8.8.4, 8.10.7, 9.0.2, 9.2.2. Each of these versions are no longer hosted into nix caches, so we're building all of these dependencies from scratch again and again.
> 
> Lastly, for each of the steps above, we do build each Haskell module twice: the profiling and the non-profiling (this is set so by default in nix).
> 
> Anyone has suggestions for what can be done to reduce the build time?

- Bump all minors version to something supported by nixpkgs
- Drop any version of GHC which is not supported anymore by nixpkgs (at least `ghc` is in the cache)
- Disable profiling builds, at least for all versions but one. Profiled builds won't be used by users of the flake, so it can only benefits developers. Developers won't profile with old version of GHC, and if they want to profile, they can still run the "simple" `nix-shell` and let `cabal` build all the profiled environment.

michaelpj:
There is a HLS cachix. The workflow is configured to push to it, but I suspect that we're not getting to that step because the build is taking too long to get there. Possibly one of us could manually build this branch and push to the cachix. I don't know how to do that or who has the keys, though.

smunix:
Screenshot_20220502-202858

The `build all binaries` stage didn't push anything to cachix.

smunix:
I removed all builds for `884` and `8107`, and kept the `build all binaries` stage only, with the hope to see it push the binaries through cachix, it didn't. It looks like this stage doesn't contribute to caching. At least, the builds were green this time.
I'll deactivate the `build all binaries` run from the workflow for now.
image

michaelpj:
Hmm, then maybe something is wrong with the cachix setup? It should be pushing things there.

smunix:
> Hmm, then maybe something is wrong with the cachix setup? It should be pushing things there.

It may also be that only specific accounts have that token available to push to haskell-language-server. Mine certainly can't push to cachix.

michaelpj:
I think the GHA runners should be able to, although you probably can't do it locally yourself...

michaelpj:
It *looks* like the cachix push step is succeeding, but I also don't get anything when I try to build locally.

I'm a bit baffled, perhaps we can summon @domenkozar for ideas on debugging the cachix setup?

domenkozar:
If the person submitting the PR doesn't have write permissions then nothing will be pushed.

michaelpj:
Well, that explains it! @domenkozar it would be great if the CLI said something in this case, it looks like we just get empty output in the logs. A "not pushing since I don't have keys" message would have been handy!

Okay, I'm going to try something:
- Close this PR
- Open a new PR (as me) from the same branch
    - That way you should still be able to push to the same branch and have it update.
- See if that does anything with caching.

michaelpj:
Argh, it's skipping the job because it thinks it should be a no-op, and I can't push to your branch. Could you push a commit please @smunix ?

michaelpj:
https://github.com/haskell/haskell-language-server/pull/2892

PR #2840, opened 2022-04-19
Update plugin support table

michaelpj:
I linked someone to it on IRC and I noticed that it was out of date.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2840"></a>

PR #2854, opened 2022-04-25
Support hls-hlint-plugin and hls-stylish-plugin to ghc9.2

July541:


pepeiborra:
What about 9.0? Should be able to play the same trick

July541:
> What about 9.0? Should be able to play the same trick

Haha, sorry for my unclear words, I just want to express that this pr will support hlint-plugin and stylish-haskell-plugin to 9.2(include any version less than 9.2)

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -128,7 +128,7 @@ addMethodPlaceholders state AddMinimalMethodsParams{..} = do
         addWhere decl = decl
 
         newLine (L l e) =
-          let dp = deltaPos 1 (indent + 1) -- Not sure why there need one more space
```
michaelpj, line 131: unrelated?

cabal-ghc92.project
```diff
@@ -39,18 +39,26 @@ package *
...
-    -hlint
     -retrie
     -splice
     -tactic,
+  hls-hlint-plugin +ghc-lib,
+  ghc-lib-parser-ex -auto,
+  hlint +ghc-lib,
+  stylish-haskell +ghc-lib
+
+source-repository-package
```
michaelpj, line 58: Why do we need this? when can we remove it?

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -23,7 +23,7 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.5
+    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6
```
michaelpj, line 26: Do we need to disallow 0.5 or should we just add 0.6?

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -181,7 +181,8 @@ suggestionsTests =
         doc <- openDoc "IgnoreAnnHlint.hs" "haskell"
         expectNoMoreDiagnostics 3 doc "hlint"
 
-    , testCase "apply-refact preserve regular comments" $ runHlintSession "" $ do
+    , knownBrokenForGhcVersions [GHC92] "apply-refact has different behavior on v0.10.0.0" $
```
michaelpj, line 184: Can we make the expected output conditional on the version instead?

michaelpj (commented):
Can you also update the plugin support table in the docs when this is ready? (https://github.com/haskell/haskell-language-server/pull/2840 brings it up to date with the current state, but this PR makes it better!)

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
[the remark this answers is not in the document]
July541: It was introduced by `ghc-exactprint-1.4` and fixed in `ghc-exactprint-1.5`.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
July541: 0.5 and 0.6 have the same requirements but 0.6 has bug fix, I don't think we have a reason to allow 0.5. https://github.com/fourmolu/fourmolu/blob/master/CHANGELOG.md

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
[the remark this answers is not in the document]
michaelpj: Should we require `ghc-exactprint >= 1.5` for this plugin, then?

cabal-ghc92.project
July541: Due to the limitation of `hls-hlint-plugin` code structure, we must force `ghc-lib-parser` uses `ghc-lib` instead of ghc to make compile, this repo adds a flag to allow us to force use `ghc-lib` in `stylish-haskell`(or is there another way to do this with out edit the package?)

---
> Why do we have to use `ghc-lib`?

1. We have two code path in hls-hlint-plugin, includes using `ghc` and using `ghc-lib`. the `ghc` codebase relies on our compiled module by `getParsedModuleWithComments` which has some `ann` operation that ghc-9.2 has totally different views, we can make `ghc-9.2` also use this after some revise(Not much work, but will introduce more complex compat).  The `ghc-lib` codebase parses the source code directly and we can use this for any ghc-lib easily.
2. We use `apply-refact` to apply hlint, and for ghc before 9.2, it requires `apply-refact-0.9` and in ghc-9.2, it needs `apply-refact-0.10`. Since in `ghc` codebase we pass parsed module to apply-refact, it works in 0.9 but not in 0.10.  `apply-refact-0.10` uses an internal `IORef`(https://github.com/mpickering/apply-refact/blob/cb8464af74ce9c1cee26fa4862c713bb5ff96114/src/Refact/Internal.hs#L751-L755) and the only entrance is parse the source by `apply-refact`. (https://github.com/mpickering/apply-refact/blob/cb8464af74ce9c1cee26fa4862c713bb5ff96114/src/Refact/Internal.hs#L698)

---
I think we can remove this after the new `stylish-haskell` with `ghc-lib` flag is released.

July541: It looks we can turn to `ghc-lib` and abandon `ghc` in `hls-hlint-plugin`, can we?

July541: Pr created https://github.com/haskell/stylish-haskell/pull/402

plugins/hls-class-plugin/hls-class-plugin.cabal
```diff
@@ -28,6 +28,7 @@ library
     , base            >=4.12 && <5
     , containers
     , ghc
+    -- Require ghc-exactprint >= 1.5 on ghc9.2
```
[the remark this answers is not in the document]
July541: > oops, I think we can only add a constraint in .project or stack-x.yaml since ghc-exactprint-0.6 is required from ghc8 to ghc9.0. I'll add a comment here.

Unfuntunately, just a comment.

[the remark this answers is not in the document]
michaelpj: You could have a cabal conditional based on the GHC version? Maybe it's not worth it.

July541:
I think this pr is ready to be merged. Everything is ok except the nix config, we have a working pr to upgrade it, https://github.com/haskell/haskell-language-server/pull/2853, we can add nix contents after nix scripts upgrade.

eddiemundo (approved):
LGTM. Can't comment on the `ghc-lib` stuff because I never really understood it.

michaelpj (approved):

PR #2866, opened 2022-04-29
Consolidate all cabal.project files

pepeiborra:
Consolidate all the various project files into a single one that aggregates all the `allow-newer` and `constraints` from the existing project files.
- `cabal.project`
- `cabal-ghc90.project` 
- `cabal-ghc92.project` 

This is possible now that our dependencies (stylish-haskell, hlint, etc.) have moved forward to `ghc-lib` 9.2 and that all the per-ghc-version cabal flags have been eliminated (#2867).

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2866"></a>


cabal.project
```diff
@@ -37,17 +37,16 @@ optional-packages: vendored/*/*.cabal
...
 constraints:
     hyphenation +embed
 
 allow-newer:
-    -- for shake-bench
-    Chart-diagrams:diagrams-core,
-    SVGFonts:diagrams-core
+  stylish-haskell:Cabal,
+  stylish-haskell:ghc-lib-parser,
+  stylish-haskell:aeson
```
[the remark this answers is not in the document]
pepeiborra: WIP - I am looking to delete `cabal-ghc92.project` as well

.github/workflows/test.yml
```diff
@@ -113,13 +113,6 @@ jobs:
           ghc: ${{ matrix.ghc }}
           os:  ${{ runner.os }}
 
-      # repeating builds to workaround segfaults in windows and ghc-8.8.4
-      - name: Build
```
michaelpj, line 117: Previously we could have had plugins which _built_ on GHC X (and hence weren't commented out in the special `cabal.project`), but whose _tests_ didn't pass (and hence weren't included in the testing workflow. These plugins would have been built by this step, which would have ensured that they kept compiling, even if the tests didn't pass.

Now I think we will only build stuff that's necessary for the tests that we try and run, so if we don't test a plugin we also won't ensure that it keeps building.

Also I think we won't check compilation of the main HLS executable either?

pepeiborra: This is a good observation, thanks! 

I deleted this step before figuring out that we can use the `buildable` field to selectively disable projects per ghc version. With that trick, hopefully this step can stay and we can avoid all the pitfalls that you discovered

michaelpj: I have a feeling that that won't work. I think I suggested this to @jneira in the past and he convinced me that it wouldn't work. I can't remember why though >:(

pepeiborra: I can't see the reason why it wouldn't work. CI is failing because the Cabal plan it computes is wrong for 9.0.1, but I get the right plan locally so it might be a CI bug

michaelpj: I think maybe it had something to do with Cabal not taking `buildable` into account when making a plan, so it won't be clever enough to try setting the `splice` flag to `false` if the splice plugin isn't buildable, it'll just fail. Worth checking though.

pepeiborra: It doesn't need to be:

https://github.com/haskell/haskell-language-server/blob/4b475b0adf665d159b1de54f260eeb84bfdf58fe/haskell-language-server.cabal#L237-L239

michaelpj: Ah, so you're going to keep the existing machinery? Maybe the thing that wouldn't work was a more ambitious thing: I'd though that maybe we could *automatically* have plugins be enabled or disabled depending on whether the plugin package was buildable given the rest of the configuration or not. So then we'd only have to put the logic for deciding buildability in the plugin package, which would be nice.

pepeiborra: One step at a time..

pepeiborra:
@michaelpj the Nix CI is failing, but I am going to leave it to you (after this PR has landed). Deal?

plugins/hls-brittany-plugin/hls-brittany-plugin.cabal
```diff
@@ -17,6 +17,10 @@ extra-source-files:
   test/testdata/**/*.hs
 
 library
+  if impl(ghc >= 9.2)
+    buildable: False
+  else
+    buildable: True
```
fendor, line 23: This is not ideal, afaict, this spreads disabling plugins over multiple files, but it looks like we have a hard time doing better than this.
Is https://github.com/haskell/cabal/pull/7783 good enough to allow us to centralise this again?

fendor (approved):
LGTM, awesome!

plugins/hls-brittany-plugin/hls-brittany-plugin.cabal
pepeiborra: I don't know whether it's good enough, but I agree that buildabilty with ghc x.y would sit better in the project file

michaelpj:
> the Nix CI is failing

It's already broken, you're not making it any worse!

plugins/hls-brittany-plugin/hls-brittany-plugin.cabal
michaelpj: I actually think it's better in the plugin files! "Does this plugin work on GHC X" feels like a concern of the plugin.

cabal-ghc92.project
```diff
@@ -1,93 +0,0 @@
...
-
-constraints:
-  -- These plugins don't build/work on GHC92 yet
-  haskell-language-server
-    +ignore-plugins-ghc-bounds
-    -brittany
-    -haddockComments
-    -retrie
-    -splice
-    -tactic,
```
michaelpj, line 52: Shouldn't the retrie, splice, and tactics plugins also get `buildable` conditionals?

michaelpj: In fact, how is the CI passing?? Did they just start working without us noticing?

michaelpj: Oh, it's because we still have the GHC conditionals for inclusion in `haskell-language-server.cabal`

pepeiborra: Yeah you are right - I'm not sure how `cabal build` decides what to build, so far I'm just adding the `buildable` conditionals on demand

PR #2867, opened 2022-04-29
simplify hlint plugin Cabal descriptor

pepeiborra:
While working on #2866 I noticed that `hls-hlint-plugin`requires per-ghc-version flags. This prevents `cabal install haskell-language-server` from working across ghc versions.

I think we can get rid of these flags:
- the `ghc-lib` flag is unnecessary, there is no downside to using ghc-lib always for maximum compatibility
- the `hlint34` flag is unnecessary, cabal should be able to use the right version

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2867"></a>

### DIFF
--- a/cabal-ghc92.project
+++ b/cabal-ghc92.project
@@ -34,7 +34,8 @@ with-compiler: ghc-9.2.2
 tests: true
 
 package *
-  ghc-options: -haddock
+  -- ghc 8.10 cannot build ghc-lib 9.2 with --haddock
+  -- ghc-options: -haddock
   test-show-details: direct
 
 write-ghc-environment-files: never
@@ -50,8 +51,6 @@ constraints:
     -retrie
     -splice
     -tactic,
-  hls-hlint-plugin +ghc-lib,
-  -- # Use ghc-lib force instead of ghc itself
   ghc-lib-parser-ex -auto,
   hlint +ghc-lib,
   stylish-haskell +ghc-lib
--- a/cabal-ghc92.project
+++ b/cabal-ghc92.project
@@ -34,8 +34,7 @@ with-compiler: ghc-9.2.2
 tests: true
 
 package *
-  -- ghc 8.10 cannot build ghc-lib 9.2 with --haddock
-  -- ghc-options: -haddock
+  ghc-options: -haddock
   test-show-details: direct
 
 write-ghc-environment-files: never
--- a/cabal.project
+++ b/cabal.project
@@ -43,7 +43,7 @@ package *
 
 write-ghc-environment-files: never
 
-index-state: 2022-03-08T10:53:01Z
+index-state: 2022-04-27T09:22:49Z
 
 constraints:
     hyphenation +embed,
--- a/cabal.project
+++ b/cabal.project
@@ -47,6 +47,8 @@ index-state: 2022-04-27T09:22:49Z
 
 constraints:
     hyphenation +embed,
+    -- remove this when hlint sets ghc-lib to true by default
+    -- https://github.com/ndmitchell/hlint/issues/1376
     hlint +ghc-lib
 
 allow-newer:
--- a/cabal.project
+++ b/cabal.project
@@ -37,7 +37,8 @@ optional-packages: vendored/*/*.cabal
 tests: true
 
 package *
-  ghc-options: -haddock
+  -- ghc 8.10 cannot build ghc-lib 9.2 with --haddock
+  -- ghc-options: -haddock
   test-show-details: direct
 
 write-ghc-environment-files: never
--- a/cabal.project
+++ b/cabal.project
@@ -46,7 +46,8 @@ write-ghc-environment-files: never
 index-state: 2022-03-08T10:53:01Z
 
 constraints:
-    hyphenation +embed
+    hyphenation +embed,
+    hlint +ghc-lib
 
 allow-newer:
     -- for shake-bench
--- a/plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
+++ b/plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
@@ -25,19 +25,6 @@ flag pedantic
   default:     False
   manual:      True
 
-flag ghc-lib
-  default:     False
-  manual:      True
-  description:
-    Force dependency on ghc-lib-parser even if GHC API in the ghc package is supported
-
-flag hlint34
-  default:     True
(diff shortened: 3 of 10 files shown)


pepeiborra:
To make `cabal install haskell-language-server` truly possible, `hlint` needs to default to use `ghc-lib` consistently across ghc versions. I have opened https://github.com/ndmitchell/hlint/issues/1376 for this

michaelpj:
Ah, ignore the last comment, I just read the linked issue in hlint.

cabal-ghc92.project
```diff
@@ -34,7 +34,8 @@ with-compiler: ghc-9.2.2
 tests: true
 
 package *
-  ghc-options: -haddock
+  -- ghc 8.10 cannot build ghc-lib 9.2 with --haddock
```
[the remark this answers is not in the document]
pepeiborra: Honest mistake

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -69,41 +56,11 @@ library
     , transformers
     , unordered-containers
     , apply-refact          >=0.9.0.0
-    -- can be removed if https://github.com/ndmitchell/hlint/pull/1325#issue-1077062712 is merged
```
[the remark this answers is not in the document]
pepeiborra: Because it lies. The referenced issue got merged but I tried removing the dependency and found that it's still needed

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -69,41 +56,11 @@ library
...
-
-    if (!flag(ghc-lib) && impl(ghc >=9.0.1) && impl(ghc <9.1.0))
-      build-depends: ghc      ==9.0.*
-    else
-      build-depends:
-        , ghc-lib            ^>=9.2
-        , ghc-lib-parser-ex  ^>=9.2
-        , ghc-lib-parser     ^>=9.2
-
-      cpp-options:   -DHLINT_ON_GHC_LIB
```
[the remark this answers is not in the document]
pepeiborra: I reserve the option to clean up later, once the dust settles

cabal.project
```diff
@@ -37,15 +37,17 @@ optional-packages: vendored/*/*.cabal
 tests: true
 
 package *
-  ghc-options: -haddock
+  -- ghc 8.10 cannot build ghc-lib 9.2 with --haddock
```
michaelpj, line 40: I'm still confused by this comment.

cabal.project
```diff
@@ -37,15 +37,17 @@ optional-packages: vendored/*/*.cabal
...
 
 write-ghc-environment-files: never
 
-index-state: 2022-03-08T10:53:01Z
+index-state: 2022-04-27T09:22:49Z
 
 constraints:
-    hyphenation +embed
+    hyphenation +embed,
+    hlint +ghc-lib
```
michaelpj, line 50: How does this affect our ability to install from Hackage? Do we need a comment here? "Remove when there is a release of hlint that uses this by default"?

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -69,41 +56,11 @@ library
...
-      build-depends: ghc >=8.10 && <9.0
-    else
-      build-depends:
-        , ghc
-        , ghc-lib            ^>=8.10.7.20210828
-        , ghc-lib-parser-ex  ^>=8.10
-
-      cpp-options:   -DHLINT_ON_GHC_LIB
-
+  cpp-options:   -DHLINT_ON_GHC_LIB
```
michaelpj, line 63: I do think we should remove this and its uses before we merge this.

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -371,7 +371,7 @@ disableHlint = sendConfigurationChanged $ toJSON $ def { Plugin.plugins = Map.fr
 -- Although a given hlint version supports one direct ghc, we could use several versions of hlint
 -- each one supporting a different ghc version. It should be a temporary situation though.
 knownBrokenForHlintOnGhcLib :: String -> TestTree -> TestTree
-knownBrokenForHlintOnGhcLib = knownBrokenForGhcVersions [GHC86, GHC88, GHC90, GHC92]
+knownBrokenForHlintOnGhcLib = expectFailBecause
```
michaelpj, line 374: inline?

michaelpj (approved):

cabal.project
pepeiborra: ghc-lib 9.2 contains illegal haddock markup which triggers a build error in ghc 8.10. This was demoted to a warning in ghc 9.x

This is only a problem now because previously we fixed hlint to v3.2 when building ghc 8.10.7 and this problem doesn't arise

cabal.project
pepeiborra: This change removes the need to have per-ghc-version flags, enabling a single `cabal.project` file. And when `hlint` sets this flag by default, it will enable `cabal install haskell-language-server` across all ghc versions.
I'll add a comment.

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
pepeiborra: I wonder if this option is still relevant for some users of ghcide as a library. 

Sigma IDE doesn't currently integrate the hlint plugin, but if it did we'd almost certainly want to use the non ghc-lib version

plugins/hls-hlint-plugin/test/Main.hs
pepeiborra: I'll keep it for documentation until the non-ghc-lib code path is fully removed

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
michaelpj: I'm confused. There is no option now, we just unconditionally set this. If you want to preserve the flag (and the corresponding CPP), then sure, but I don't see why we'd have this half-way version?

pepeiborra: Of course, good point. For now let's preserve the flag with a default of True

PR #2868, opened 2022-04-29
Compatibility with older aeson releases

pepeiborra:
I want to build the call hierarchy plugin in a stockage LTS 16.11 system, which has an older version of `aeson`

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2868"></a>


July541 (approved):
LGTM

PR #2870, opened 2022-04-30
Fix hanging redundant import on Unicode function 

### LINKED ISSUE #2859: Redundant symbol containing Unicode character may cause HLS to hang
### Your environment

* Which OS do you use: Ubuntu 20.04.2 LTS

* Which LSP client (editor/plugin) do you use: VSCode + vscode-haskell

* Describe your project (alternative: link to the project): A simple project using GHC 9.02 consisting of the following two modules:

  + `Types.hs`:
    ```hs
    module Types where

    ε :: Double
    ε = 0.5

    data A = A
    ```

  + `Lib.hs`
    ```hs
    {-# OPTIONS_GHC -Wall #-}
    module Lib where
    import Types (A (..), ε)

    a :: A
    a = A
    ```

### Steps to reproduce
1. Open `Lib.hs` and wait a moment.
2. Warning on redundant import of `ε` gets reported and HLS stops responding.

### Expected behaviour
HLS should continue working and provide some code action to remove `ε` from the list.

### Actual behaviour
HLS hangs and stop responding indefinitely.

If we rename `ε` so that it doesn't include any  unicode character, then the panic goes away.
Also note that removing the definition of `a` from `Lib.hs` and stop importing `A(..)`, then HLS won't panic anymore - it seems that at least one extra imported item are needed to reproduce the situation. In other words, the following `Lib.hs` won't make HLS panic:

```hs
{-# OPTIONS_GHC -Wall #-}
module Lib where

import Types (ε)
```

### Include debug information

Debug Output here:

[panic.log](https://github.com/haskell/haskell-language-server/files/8567693/panic.log)

The following `panic` lines from the above illustrates the situation:

```
haskell-language-server-9.0.2: haskell-language-server-9.0.2: panic! (the 'impossible' happened)
  (GHC version 9.0.2:
	charType: '\949'
```

`\949` corresponds to `ε`.

drsooch:
Closes #2859

Revert partial changes from #2483

For redundant import actions, the code path ends up calling `wrapOperatorInParen` which under the hood makes a call to `is_ident` which is a low-level `Char -> Bool`. The function will panic when it's input falls out of ASCII range.

The fix is just to match directly on the '_' character. Outside of this codepath, the original
functionality has been restored (using `isAlpha`).

Also removes the now unnecessary `is_ident` compat layer.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2870"></a>

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Compat/Util.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat/Util.hs
@@ -69,8 +69,6 @@ module Development.IDE.GHC.Compat.Util (
     stringToStringBuffer,
     nextChar,
     atEnd,
-    -- * Char
-    is_ident
     ) where
 
 #if MIN_VERSION_ghc(9,0,0)
@@ -83,7 +81,6 @@ import           GHC.Data.FastString
 import           GHC.Data.Maybe
 import           GHC.Data.Pair
 import           GHC.Data.StringBuffer
-import           GHC.Parser.CharClass    (is_ident)
 import           GHC.Types.Unique
 import           GHC.Types.Unique.DFM
 import           GHC.Utils.Fingerprint
@@ -93,7 +90,6 @@ import           GHC.Utils.Panic         hiding (try)
 #else
 import           Bag
 import           BooleanFormula
-import           Ctype                   (is_ident)
 import           EnumSet
 import qualified Exception
 import           FastString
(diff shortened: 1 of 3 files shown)


pepeiborra (approved):
Is it worth adding a test?

konn:
> Is it worth adding a test?

I think that it makes sense to add a simple test based on the repro I reported in #2859 to make sure that the reported issue is actually solved by this PR.

drsooch:
Will add one yes

drsooch:
Regression test properly fails. Will timeout after some time, hope this is reasonable?

michaelpj (approved):

PR #2877, opened 2022-04-30
Restore compat. with prettyprinter 1.6

pepeiborra:
I intend to continue supporting LTS 16 for a little longer.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2877"></a>

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -1,5 +1,6 @@
 -- Copyright (c) 2019 The DAML Authors. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP               #-}
 {-# LANGUAGE NamedFieldPuns    #-}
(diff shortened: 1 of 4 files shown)


exe/Main.hs
```diff
@@ -1,5 +1,6 @@
 -- Copyright (c) 2019 The DAML Authors. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP               #-}
```
michaelpj, line 3: What we've been doing elsewhere is just to use the old imports and accept the deprecation warnings. Up to you, but I think the CPP is more painful than the warning.

pepeiborra: The `CPP` is more forward compatible and sets an example on how to do it

michaelpj (approved):

PR #2885, opened 2022-05-01
no longer disable -dynamic in CI

pepeiborra:


pepeiborra:
> Why?

We are no longer distributing static binaries, so why disable -dynamic in CI anymore?

michaelpj (approved):

PR #2886, opened 2022-05-01
Prefer Data.HashSet.member to Data.Foldable.elem

sergv:


fendor (approved):
LGTM, thanks!

pepeiborra:
What is the motivation to prefer `HashSet.member` over `Data.Foldable.elem`?

jhrcek:
`HashSet.member` is `O(log n)` wherease `Data.Foldable.elem` needs to traverse the whole structure (at best linear in number of elements). For example relude even uses type errors to warn you about using elem with sets: https://hackage.haskell.org/package/relude-1.0.0.1/docs/Relude-Foldable-Fold.html#v:elem

pepeiborra:
> I suspect it's because `HashSet.member` is `O(log n)` whereas `Data.Foldable.elem` needs to traverse the whole structure (at best linear in number of elements). For example relude even uses type errors to warn you about using elem with sets: https://hackage.haskell.org/package/relude-1.0.0.1/docs/Relude-Foldable-Fold.html#v:elem

Fair enough, `elem` lacks `Hashable` powers so it cannot do anything smart

sergv:
Indeed, `elem` is linear since it must work with only `Eq` constraint.

PR #2887, opened 2022-05-02
Add .txt files as extra-source-files for hls-change-type-signature-plugin

cdepillabout:
`hls-change-type-signature-plugin` appears to have a few `.txt` files that are necessary for running tests, but those files haven't been added to `extra-source-files` in `hls-change-type-signature-plugin`.

This appears to be causing errors when we try to build `hls-change-type-signature-plugin` in Nixpkgs: https://hydra.nixos.org/build/175155441/nixlog/1.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2887"></a>


pepeiborra (approved):
Thank you!

PR #2884, opened 2022-05-01
hls-pragmas-plugin requires ghcide >= 1.7

Bodigrim:
Otherwise `Development.IDE.printOutputable` is not available.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2884"></a>

PR #2889, opened 2022-05-02
Modify ghcide requirements of hls-change-type-signature-plugin

July541:
Reference #2884, I think hls-change-type-signature-plugin should also require ghcide >=1.7 since it uses `printOutputable`.

https://github.com/haskell/haskell-language-server/blob/6d319ba24d60d87677913a9c1a71279959c1b2c2/plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs#L134-L138

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2889"></a>


pepeiborra (approved):
Thanks @July541 I'll do a Hackage revision as well

Issue #2794, opened 2022-03-22
Support `hindent` for formatting


vidhanio:

michaelpj:
I think this would just require someone to write a plug-in for it.

uhbif19:
PR in progress: https://github.com/haskell/haskell-language-server/pull/2896

PR #2896, opened 2022-05-07
#2794 Initial HIndent support

uhbif19:
I am not quite sure if I am using HIndent API correctly, but it does format something.

Currently I do not handle config, cuz not sure where to call `getConfig :: IO Config` (https://github.com/mihaimaruseac/hindent/blob/master/src/main/Main.hs#L77).

There is some story with `[Extension]` param. As far as I understand providing `Config` should be enough. (https://github.com/mihaimaruseac/hindent/blob/master/src/HIndent.hs#L71)


pepeiborra:
You will need to add the test suite to the GitHub CI test.yml

pepeiborra:
> Currently I do not handle config, cuz not sure where to call `getConfig :: IO Config` (https://github.com/mihaimaruseac/hindent/blob/master/src/main/Main.hs#L77).

You can setup a handler for the LSP Initialised event and call it there:

https://github.com/haskell/haskell-language-server/blob/65fbc5e41cf407ce98a03f5faa0a1b93c995b7d6/exe/Main.hs#L54-L57

michaelpj:
I think it would also be fine to call it in the handler, that's what the `stylish-haskell` plugin does, for example.

uhbif19:
Looks like some yak-shaving required to make this to build: https://github.com/mihaimaruseac/hindent/issues/583

uhbif19:
Looks like `getConfig` is not exported from hindent. So I can either try to change hindent or copy `getConfig` with `Path.File` and bunch of dependencies. Which way is better?

michaelpj:
> So I can either try to change hindent

That would be ideal. Maybe open an issue and see what the maintainer thinks is the best way of doing it. People are typically happy to be included in HLS, so hopefully you'll get a response :)

uhbif19:
Issue created: https://github.com/mihaimaruseac/hindent/issues/585

michaelpj:
Looks like upstream isn't very responsive at the minute. Perhaps copying `getConfig` is the least bad option in the short term?

uhbif19:
@michaelpj Maybe I will write universal config-looking function working for both HIndent and Stan?

uhbif19:
@michaelpj 

`getConfig` was added and tests use configuration file now.

Where is still git-dependencty in cabal project, cuz build fix is not released yet.

cabal.project
```diff
@@ -29,6 +29,12 @@ packages:
          ./plugins/hls-selection-range-plugin
          ./plugins/hls-change-type-signature-plugin
          ./plugins/hls-gadt-plugin
+         ./plugins/hls-hident-plugin
+
+source-repository-package
```
michaelpj, line 34: What do you need to remove this? Do you think it's going to happen soon? If not, add a comment explaining why this is here and when we can remove it.

plugins/hls-hident-plugin/src/Ide/Plugin/HIndent.hs
```diff
@@ -0,0 +1,67 @@
...
+  config <- liftIO getHIndentConfig
+  let result = HIndent.reformat config extensions (Just file) (TE.encodeUtf8 selectedContents)
+  convertToTextEdit result range
+    where
+      builderToText = TE.decodeUtf8 . B.concat . BL.toChunks . toLazyByteString
+      convertToTextEdit result range = case result of
+        Left  err -> return $ Left $ responseError $ T.pack $ "hident: " ++ err
+        Right new -> return $ Right $ J.List [TextEdit range (builderToText new)]
+
+-- Copied from
```
michaelpj, line 52: Link to the ticket and explain that we can remove this once upstream gives us a better way.

plugins/hls-hident-plugin/src/Path/Find.hs
```diff
@@ -0,0 +1,98 @@
+{-# LANGUAGE DataKinds #-}
+
+-- | Finding files.
+
+-- Copied from HIndent, where it was lifted from Stack.
```
michaelpj, line 5: :see_no_evil:

michaelpj (approved):
LGTM, just a couple of comment requests.

plugins/hls-hident-plugin/src/Ide/Plugin/HIndent.hs
uhbif19: Done.

cabal.project
uhbif19: Done.

michaelpj:
Sorry, just one last thing I missed before - docs! Can you add this plugin to the "Features" doc and also add a line for it about GHC version support on the GHC version compatibility page.

uhbif19:
@michaelpj Circleci CI tasks are stalled. Can I do something to fix this?

uhbif19:
@michaelpj Circleci CI tasks are stalled. Can I do something to fix this?

uhbif19:
@michaelpj Circleci CI tasks are stalled. Can I do something to fix this?

michaelpj:
Not sure tbh!

michaelpj:
They seem to be passing on other jobs, maybe just try pushing a trivial commit and see if they start up this time?

uhbif19:
@michaelpj Tried. Seems like it still does not work.

michaelpj:
It does just seem to be this PR, I really have no idea why. Try making another PR with the same content?

uhbif19:
@michaelpj These PR seems stuck too:

https://github.com/haskell/haskell-language-server/pull/2949
https://github.com/haskell/haskell-language-server/pull/2908

pepeiborra:
@uhbif19 are you able to wrap up this PR?

uhbif19:
@pepeiborra Will try to fix in next couple of days.

uhbif19:
@michaelpj Another strange CI (bug?): "Could not find a usable config.yml because you deleted the CircleCI OAuth app."

michaelpj:
You are oddly cursed, nobody else seems to have trouble with CircleCI :sweat_smile:

michaelpj:
Okay, looks like a genuine failure on stackage-lts16 now.

uhbif19:
This looks like hindent problem, I created an issue: https://github.com/mihaimaruseac/hindent/issues/589

uhbif19:
Now all CI jobs getting stuck.

pepeiborra:
@uhbif19 are you able to resolve the conflicts? Hopefully that will unstuck CI

uhbif19:
Lol, I forgot about this PR. Does anyone need it?

Issue #2261, opened 2021-10-06
GADT syntax


georgefst:
An action on a datatype declaration for converting to GADT syntax would be awesome.

For large datatypes, this can be painful to do manually, despite being totally mechanical. And there are edge cases that make it difficult to use simple scripts or keyboard macros.

PR #2899, opened 2022-05-09
A plugin for GADT syntax converter

July541:
This plugin is for #2261

~~Currently only work for ghc-9.2.~~ Fully support now!

Known limits:
- ~~Origin comments missing(I hope gives comments up since they are not easy to put in a proper place).~~
~~(It's not handy, I'll try to keep some comments)~~
It's hard to trace comments, so I decide to put it on hold.

--- 

gadt

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2899"></a>


plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
```diff
@@ -0,0 +1,135 @@
...
+            $ filter (inRange range) hsDecls
+        exts = (toList . extensionFlags . ms_hspp_opts . pm_mod_summary) pm
+    pure (decls, exts)
+
+-- Copy from hls-alternate-number-format-plugin
+getNextPragma :: MonadIO m => IdeState -> NormalizedFilePath -> ExceptT String m NextPragmaInfo
+getNextPragma state nfp = handleMaybeM "Error: Could not get NextPragmaInfo" $ do
+      ghcSession <- liftIO $ runAction "GADT.GhcSession" state $ useWithStale GhcSession nfp
+      (_, fileContents) <- liftIO $ runAction "GADT.GetFileContents" state $ getFileContents nfp
+      case ghcSession of
```
July541, line 133: We can't move this to hls-plugin-api unless hls-plugin-api has dependency of ghcide

plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
```diff
@@ -0,0 +1,303 @@
...
+        removeExtraEmptyLine s = case stripInfix "\n\n" s of
+            Just (x, xs) -> x <> "\n" <> xs
+            Nothing      -> s
+
+#endif
+
+#if MIN_VERSION_ghc(9,2,1)
+wrap :: HsType GP -> GenLocated (Anno (HsType GP)) (HsType GP)
+wrap = wrapXRec @GP
+wrap' = wrapXRec @GP
```
July541, line 273: Here is something about type family, I don't know how to make it one...

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
```diff
@@ -0,0 +1,135 @@
...
+        mkAction :: T.Text -> Command |? CodeAction
+        mkAction name = InR CodeAction{..}
+            where
+                _title = "Convert \"" <> name <> "\" to GADT syntax"
+                _kind = Just CodeActionRefactorRewrite
+                _diagnostics = Nothing
+                _isPreferred = Nothing
+                _disabled = Nothing
+                _edit = Nothing
+                _command = Just
```
michaelpj, line 104: Why are you doing the indirection via a command? You could just compute the edit in the code action handler and return it here!

michaelpj: I don't think your command handler is significantly more expensive - I think it also mostly relies on the parsed module, which you also need here.

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
```diff
@@ -0,0 +1,135 @@
...
+                    $ mkLspCommand plId toGADTSyntaxCommandId _title (Just [toJSON mkParam])
+                _xdata = Nothing
+
+        mkParam = ToGADTParams (doc ^. L.uri) range txt
+
+-- | Get all H98 decls in the given range, and enabled extensions
+getInRangeH98DeclsAndExts :: (Monad (t IO), MonadTrans t) =>
```
michaelpj, line 111: Generally I think you could make your life a bit easier with some `mtl`: you could have `(MonadError String m, MonadIO m)` here and just use `mtl` functions instead of lots of explicit lifting. I think that's true throughout.

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
michaelpj: I think it would be good to move the pragma-insertion logic to `hls-plugin-api` from `ghcide`, since it's a very common thing to want to do. Ideally we could wrap it all up in something like `insertPragmaIfNotPresent` or something.

plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
```diff
@@ -0,0 +1,303 @@
...
+We use `printOutputable` to print H98 data decl as GADT syntax,
+this print is not perfect, it will:
+
+1. Make data name and the `where` key word in different lines.
+2. Make the whole data decl prints in one line if there is only one data constructor.
+3. The ident size of every data constructor depends on its origin
+   format, and may have different ident size between constructors.
+
+Hence, we first use `printOutputable` to get an initial GADT syntax,
+then use `ghc-exactprint` to parse the initial result, and finally
```
michaelpj, line 164: Why aren't we just using exactprint for everything? :thinking:

plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
```diff
@@ -0,0 +1,303 @@
+{-# LANGUAGE CPP              #-}
+{-# LANGUAGE GADTs            #-}
+{-# LANGUAGE LambdaCase       #-}
+{-# LANGUAGE PatternSynonyms  #-}
+{-# LANGUAGE RecordWildCards  #-}
+{-# LANGUAGE TupleSections    #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-missing-signatures #-}
+module Ide.Plugin.GHC where
```
michaelpj, line 9: I haven't really reviewed this, it seems complicated and we should probably get one of our exactprint experts to take a look.

michaelpj (commented):
I'm a little worried that the re-formatting logic looks quite complicated and version-dependent, is this going to be a pain to maintain?

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
```diff
@@ -0,0 +1,135 @@
...
+    } deriving (Generic, ToJSON, FromJSON)
+
+toGADTSyntaxCommandId :: CommandId
+toGADTSyntaxCommandId = "GADT.toGADT"
+
+-- | A command replaces H98 data decl with GADT decl in place
+toGADTCommand :: CommandFunction IdeState ToGADTParams
+toGADTCommand state ToGADTParams{..} = response $ do
+    let plId = PluginId pluginIdText
+    nfp <- getNormalizedFilePath plId uri
```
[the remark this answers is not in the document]
July541: Continue on https://github.com/haskell/haskell-language-server/pull/2906

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
July541: I think users won't use this plugin in most cases, it's unfair to give GADT results every time for these cases.

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
July541: Thank you for pointing it out! I'm not familiar with mtl indeed.

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -384,3 +386,10 @@ getCompletionByLabel desiredLabel compls =
         Nothing -> liftIO . assertFailure $
             "Completion with label " <> show desiredLabel
             <> " not found in " <> show (fmap (^. L.label) compls)
+
+-- ---------------------------------------------------------------------
+-- Run with a canonicalized temp dir
+withTempDir :: (FilePath -> IO a) -> IO a
+withTempDir f = System.IO.Extra.withTempDir $ \dir -> do
+  dir' <- canonicalizePath dir
+  f dir'
```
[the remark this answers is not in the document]
July541: Because I need run pragma insertion tests in an isolated directory without `hie.yaml` explicitly since its behavior is not the same between ghc-9.2 and others(ghc-9.2 enabled `GADTSyntax` implicitly).

https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0380-ghc2021.rst#proposed-change

July541:
> Why aren't we just using exactprint for everything? 🤔

I'm trying to rewrite this, it may not be very struggle.

hls-test-utils/src/Test/Hls/Util.hs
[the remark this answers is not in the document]
Ailrun: I mean, why do we need canonicalization?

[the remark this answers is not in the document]
July541: It's from ghcide's test. https://github.com/haskell/haskell-language-server/blob/b1bf5499155e259341e4868863b0fd743b6bd65f/ghcide/test/exe/Main.hs#L6975-L6981

July541:
I decide to keep the current design without a better substitute.

Why?

1. comments with their anchor will be broken after we trans to GADT.
```haskell
data A where -- comment
-- comment
{- comment -} B | C {-comment-} |
D -- comment
deriving ()
```
Something like the above is hard to track its semantic place.

2. Take the following code as example
```haskell
data (Eq a) => A a = B a
```
after we convert to GADT syntax, it would be
```haskell
data           A a   BEqaaAa
```
while using `exactPrint` directly. So we need `printOutputable` to convert it to
```haskell
data A a
  where
    B :: (Eq a) => a -> A a
```
and use this to continue later pretty adjustment.

July541:
> I'm a little worried that the re-formatting logic looks quite complicated and version-dependent, is this going to be a pain to maintain?

The re-formatting relies on the behavior of `showSDoc`, we'd adjust the pretty function if `showSDoc` has any changes. Seeing that its behavior is stable crossing ghc-8.6 to ghc-9.2, I don't think we have much pain in future.

July541:
> I think it would be good to move the pragma-insertion logic to hls-plugin-api from ghcide, since it's a very common thing to want to do. Ideally we could wrap it all up in something like insertPragmaIfNotPresent or something.

Can't agree more, but currently, pragma insertion functions have dependencies on `Development.IDE.GHC.Compat` and `Development.IDE`. we can't do this unless we moved compat part to a new package and extract some util functions from ghcide to hls-plugin-api.

plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
pepeiborra: I'm not an exactprint expert either, but I have no concerns about the code in this module since it is well commented and isolated in this plugin

pepeiborra:
> > I think it would be good to move the pragma-insertion logic to hls-plugin-api from ghcide, since it's a very common thing to want to do. Ideally we could wrap it all up in something like insertPragmaIfNotPresent or something.
> 
> Can't agree more

I am not following this conversation, just wanted to make a few points: 

1. `his-plugin-api` is an interface package with little or no implementation details, whereas `ghcide` is a library for building ghc-api based LSP servers and is full of implementation details. 

2. The main motivation to keep things out of `ghcide` is to make it easier to upgrade to new GHC versions, and the direction of travel is usually `ghcide  ---> <plugin>`. 

3. It's also fine to have dependencies across plugins,

Based on those points, I would rather extract the pragma insertion logic to `hls-pragma-plugin` rather than to `hls-plugin-api`.

July541:
> > > I think it would be good to move the pragma-insertion logic to hls-plugin-api from ghcide, since it's a very common thing to want to do. Ideally we could wrap it all up in something like insertPragmaIfNotPresent or something.
> > 
> > 
> > Can't agree more
> 
> I am not following this conversation, just wanted to make a few points:
> 
> 1. `his-plugin-api` is an interface package with little or no implementation details, whereas `ghcide` is a library for building ghc-api based LSP servers and is full of implementation details.
> 2. The main motivation to keep things out of `ghcide` is to make it easier to upgrade to new GHC versions, and the direction of travel is usually `ghcide  ---> <plugin>`.
> 3. It's also fine to have dependencies across plugins,
> 
> Based on those points, I would rather extract the pragma insertion logic to `hls-pragma-plugin` rather than to `hls-plugin-api`.

Agree with your words. But maybe we can't extract pragma insertion logic from ghcide to hls-pragma-plugin, because hls-pragma-plugin has dependency of ghcide.

pepeiborra:
> Agree with your words. But maybe we can't extract pragma insertion logic from ghcide to hls-pragma-plugin, because hls-pragma-plugin has dependency of ghcide.

I don't see why this would be a problem

> I hope it's better to have a package like hls-utils to accommodate util functions. So we can have hls-utils ---> ghcide ---> <plugin>.

This is a fine option too.

hls-test-utils/src/Test/Hls/Util.hs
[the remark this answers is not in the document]
Ailrun: Oh... because of macOS. I think it's probably helpful to refer that comment or add that comment.

michaelpj:
For some reason I thought we already had `hls-plugin-utils`. In fact, `hls-plugin-api` contains lots of stuff that belongs in a "plugin utils" package!

> comments with their anchor will be broken after we trans to GADT.

This is non-haddock comments, right? Haddock comments should be attached to their piece of AST so shouldn't be too hard to put in the right place? I'm inclined to not worry about non-Haddock comments and just have that be a limitation of the plugin. Or have them all printed above the new declaration.

> Take the following code as example

I don't understand this example. Why would we get nonsense from exactprint like you wrote? Surely if we construct a correct AST then exactprint should emit it properly??

July541:
> Haddock comments should be attached to their piece of AST so shouldn't be too hard to put in the right place

All comments with `RealSrcSpan` may lose their correct place after converting to GADT. We must attach anchor operation to describe its new place, but I don't have a proper solution.

> I'm inclined to not worry about non-Haddock comments and just have that be a limitation of the plugin.

Agree.

> Or have them all printed above the new declaration.

Comments without their target are nonsense, I prefer abandon them.

> Why would we get nonsense from exactprint like you wrote?

It is because we printed an incomplete GADT decl, it only has a semantic presentation but all place info is for H98 decl, that's why I use `printOutputable` to get a correct semantic GADT decl and then we do some patch on the the new GADT decl to make it pretty.

michaelpj:
I really don't understand what's going on, sorry :/ But it seems like we *should* be able to mess with the AST and print it out again nicely, I thought that was half the point of exactprint...

Maybe @alanz can help us>

July541:
Something that can convert ast's all absolute locations to relative locations is much helpful.

I have been thinking about edit data constructors one by one instead of edit the whole data decl, maybe it can keep comments for us, I'll try this later.

July541:
Don't have a feasible solution, I want to merge this and left improvement in the future.

cc @michaelpj

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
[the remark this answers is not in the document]
michaelpj: I still think making the change here to pass it when we construct the descriptor makes more sense.

plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs
michaelpj: Did you figure out whether there was anything sensible we can do to share code here? Or do you want to leave it as a follow-up?

michaelpj (commented):
Couple of nits, but otherwise I'm happy so long as you're on the hook for maintaining it, @July541 :laughing:

michaelpj (approved):

July541:
> Couple of nits, but otherwise I'm happy so long as you're on the hook for maintaining it, @July541 😆

Sure, I'm glad to improve it if we have any better solution.

> I still think making the change here to pass it when we construct the descriptor makes more sense.

I forgot it, thank you point it out!

> Did you figure out whether there was anything sensible we can do to share code here? Or do you want to leave it as a follow-up?

Currently not, let's leave it here, I have the same code in hls-class-plugin. I'll follow https://github.com/haskell/haskell-language-server/pull/2899#issuecomment-1126922217, it's a big work but worth doing.

Issue #258, opened 2020-07-31
[Feature] Provide Stan integration


sir4ur0n:
I propose HLS supports [Stan](https://hackage.haskell.org/package/stan-0.0.1.0) integration, what do you think?

Ailrun:
I think it would be great if possible, but as they use .hie file, we cannot use them for GHC < 8.8...

Sir4ur0n:
> P.S. You are at the Haskell Love Conference, aren't you? :)

Yes, and I wanted to link the existing HLS-Stan integration issue to others. That's how I noticed there is no issue yet :stuck_out_tongue:

fendor:
Hi, I started with this, but have no particular branch yet.
I once had a somewhat working impl, but we need https://github.com/digital-asset/ghcide/pull/701, too.

googleson78:
https://github.com/haskell/ghcide/pull/701 is now in `hls` master, what additional work needs to be done to get this working? @fendor

jneira:
A required step for sure would be create a plugin to provide stan diagnostics and refactorings (?). I think `hls-hlint-plugin` could be a good starting point.

googleson78:
I was wondering if @fendor s work is something that's "almost done", and just needs a bit of cleanup.

fendor:
@googleson78 The work got stalled, and it was not that much, so we should rather re-write it. I don't have anything workable at the moment :(

uhbif19:
I will try to work on this.

As far as I understand, it required to add new Rule, which will call Stan `runAnalysis` for each project file.

fendor:
Right, but ideally we re-use the .HIE files HLS is already generating, so we save some work.

uhbif19:
Stan requires `HieTypes.HieFile` while HLS `GetHieAst` returns `HieAstResult`, which contains only `Module`, not the whole file.
Is there any way to get `HieTypes.HieFile` out of HLS?

pepeiborra:
HLS writes `.hie` files to disk, so you can definitely get a `HieFile` out of HLS although it's far from trivial. Something like this should do it:

```haskell
getHieFile :: NormalizedFilePath -> Action (Maybe HieFile)
getHieFile nfp = runMaybeT $ do
    HAR{hieAst} <- MaybeT $ use GetHieAst nfp
    tmr <- MaybeT $ use TypeCheck nfp
    ghc <- MaybeT $ use GhcSession nfp
[...]

Issue #2827, opened 2022-04-12
Editor support for Fourmolu CLI option


georgefst:
As for:

> Add config option to editor plugins e.g. VSCode 

Is there anything I need to do? I see from https://github.com/haskell/haskell-language-server/pull/1576 that we have some auto-generation script. Is this part of the release process? Is any of this (i.e. the process of adding an option) documented anywhere?

_Originally posted by @georgefst in https://github.com/haskell/haskell-language-server/issues/2763#issuecomment-1066140768_

georgefst:
The functionality added in #244 is in a weird state, of being implemented but not supported by editor integrations. It would be great to get some help from someone who knows more about how plugin-specific options work. I suspect it's simple for anyone who knows the system, but I don't know how to proceed by myself.

More generally, we could really do with a documented process for this. I'm sure other contributors will want to add plugin options in future.

pepeiborra:
I think @berberman contributed the support for plugin-specific options. What do you need help with?

berberman:
#1576 provided `--vscode-extension-schema` for generating available client options, which can be used in `package.json` of `vscode-haskell`. Yes, this should be a part of the release process of `vscode-haskell`, and iirc @jneira used to do that. It would be better to have a documentation for this. However, for other editors, there is no automated generating support, because they usually have their own formats of configurations, and I think they always can translate the vscode one to their formats if needed.

michaelpj:
@fendor seems to be doing the vscode plugin release.

It would be great if someone could write some documentation in the `vscode-haskell` readme about how to add new HLS options there, since this question comes up a lot.

fendor:
I tried to look into and I am honestly a little confused myself, can someone help me out  writing some nice docs for this task?

georgefst:
> #1576 provided --vscode-extension-schema

FYI, there's no `--`, it's just `haskell-language-server vscode-extension-schema`.

Anyway, I finally looked in to this. https://github.com/haskell/haskell-language-server/pull/3284 combined with https://github.com/haskell/vscode-haskell/pull/736 should solve the problem at hand. But I'm reluctant to close this thread since it's now being used to track something more general: documenting what plugin writers need to do to add full support for any plugin options.

PS. The Fourmolu CLI option can be used in the latest release of HLS with VSCode by adding `"haskell.plugin.fourmolu.config.external": true` to the VSCode settings JSON file manually. It doesn't matter that neither of the aformentioned PRs have been merged.

[...]

PR #2908, opened 2022-05-16
Stan integration #258

uhbif19:
#258

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2908"></a>


plugins/hls-stan-plugin/test/testdata/test.hs
```diff
@@ -0,0 +1,4 @@
+orderPair x y
+    | x < y = 1
+    | x > y = 2
+    | otherwise = 3
```
Ailrun, line 4: The test directory contains some compiler outputs, such as .o and .hi.

uhbif19: Fixed. Shouldn't they be in gitignore? I can add them in this or separate PR.

Ailrun: I think we (most of HLS contributors) rarely compile files directly so they aren't in .gitignore (the file contains stack's and cabal's working directories only)

CODEOWNERS
```diff
@@ -28,6 +28,7 @@
 /plugins/hls-splice-plugin @konn
 /plugins/hls-stylish-haskell-plugin @Ailrun
 /plugins/hls-tactics-plugin @isovector
+/plugins/hls-stan-plugin @uhbif19
```
michaelpj, line 31: Only works if you've got write access... I've invited you :)

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -0,0 +1,119 @@
...
+instance NFData GetStanDiagnostics
+
+type instance RuleResult GetStanDiagnostics = ()
+
+rules :: Recorder (WithPriority Log) -> Rules ()
+rules recorder = do
+  define (cmapWithPrio LogShake recorder) $
+    \GetStanDiagnostics file -> do
+      hie <- fromJust <$> getHieFile file
+      let enabledInspections = HM.fromList [(LSP.fromNormalizedFilePath file, inspectionsIds)]
```
michaelpj, line 74: Is this idea to make this configurable? Comment?

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -0,0 +1,119 @@
...
+            _severity = Just LSP.DsHint,
+            _code = Just (LSP.InR $ unId (inspectionId inspection)),
+            _source = Just "stan",
+            _message = message,
+            _relatedInformation = Nothing,
+            _tags = Nothing
+          }
+      )
+      where
+        inspection = fromJust $ HM.lookup observationInspectionId inspectionsMap
```
michaelpj, line 99: Please don't use partial functions, this can take down HLS entirely. You can log an error and just omit the case you can't handle, just don't `error`.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -0,0 +1,119 @@
...
+        -- Lookin similar to Stan CLI output
```
michaelpj, line 100:
```suggestion
        -- Looking similar to Stan CLI output
```

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -0,0 +1,119 @@
...
+        message :: T.Text
+        message =
+          T.unlines $
+            [ " ✲ Name:        " <> inspectionName inspection,
+              " ✲ Description: " <> inspectionDescription inspection,
+              "Possible solutions:"
+            ]
+              ++ map ("  - " <>) (inspectionSolution inspection)
+
+getHieFile :: NormalizedFilePath -> Action (Maybe HieFile)
```
michaelpj, line 110: This seems like something we should provide alongside the other HIE stuff? WDYT @wz1000 ?

michaelpj: At least this needs some comments.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: We can enable Inspections by category: https://github.com/kowainik/stan/blob/main/src/Stan/Inspection/All.hs#L31.

Or we can add `excludeInspections` option.

Also Stan may look for cabal config to look for extensions. Is it possible to ask HLS for cabal file?

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: @michaelpj For me this function signature looks pretty self-explanatory. What else could be said about it?

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: But this function cant fail, because inspection ID always should be from inspection id list.

uhbif19: Fixed.

uhbif19:
I do not understand why CI fails. Plugin is disabled for GHC < 9, but Cabal build still fails: https://github.com/haskell/haskell-language-server/pull/2908/files#diff-aaf5dad5ddf1d2c16a2e8034c6435719f6eb3327cfe1bcd042df5a84b939afe3R282

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
michaelpj: Here are a list of questions that I have about this function after reading it:
- Why isn't this function provided with the other HIE file functions we have? Is it doing something suspicious? Is there a reason we don't want to provide HIE files? Is it working against the direction of travel with hiedb?
- How does it work? Are we sure we're constructing HIE files correctly? Can the information we get from the rules be inconsistent with what's coming from hiedb? 

I'm sure you thought about a bunch of these things when you were writing it, but all that knowledge is going to be lost unless it gets written down! :)

michaelpj:
@pepeiborra changed how you have a plugin that's only enabled on some versions recently. I'm not sure what the right way is off the top of my head (@pepeiborra if you write something here I'll add it to the contributing docs).

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: This function was written by @pepeiborra , so I do not know the answers https://github.com/haskell/haskell-language-server/issues/258#issuecomment-1126890590

pepeiborra:
You need to set the `buildable` field of your library and possibly the test-suite too.Example

https://github.com/haskell/haskell-language-server/blob/8cbefb71bc3aff7c19d4874607b6a374f2307fc3/plugins/hls-haddock-comments-plugin/hls-haddock-comments-plugin.cabal#L20-L24

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
pepeiborra: Those are valid questions, and @wz1000 is the right person to review rather than me. Also, perhaps move this function to the `ghcide` package and refactor other pieces of similar code around the `Core/Compile.hs` and `Core/Rules.hs` modules.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
July541: Maybe a configure file like https://github.com/kowainik/stan/blob/main/.stan.toml?

michaelpj: I'm not saying we should necessarily make it configurable now, but it would be good to write something down here about it.

e.g.
```
-- In the future we may want to control which inspections are enabled based on ???
```

> Also Stan may look for cabal config to look for extensions.

HLS knows the extensions enabled for a file directly, if `stan` will let you set them.

uhbif19: > HLS knows the extensions enabled for a file directly

I guess Stan waits for cabal enabled extensions, not extensions enabled directly, cuz function used for collecting extensions is named `createCabalExtensionsMap`. https://github.com/kowainik/stan/blob/main/src/Stan.hs#L74

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -0,0 +1,120 @@
...
+    void $ uses GetStanDiagnostics $ HM.keys files
+  where
+    analysisToDiagnostics :: NormalizedFilePath -> Analysis -> [FileDiagnostic]
+    analysisToDiagnostics file = mapMaybe (observationToDianostic file) . toList . analysisObservations
+    observationToDianostic :: NormalizedFilePath -> Observation -> Maybe FileDiagnostic
+    observationToDianostic file (Observation {observationSrcSpan, observationInspectionId}) =
+      do
+        inspection <- HM.lookup observationInspectionId inspectionsMap
+        let
+          -- Looking similar to Stan CLI output
```
Ailrun, line 89: Is there a reason to avoid `prettyShowInspection` of `Stan.Inspection`? If so, could you expand this comment with the reason?

uhbif19: `prettyShowInspection` included "Id" (which is already in _code) and "Severity" (which could be placed in _severety), which are not needed to show here, as I think.

michaelpj: Worth a comment.

uhbif19: Comment done.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: Comment done.

haskell-language-server.cabal
```diff
@@ -352,7 +362,7 @@ executable haskell-language-server
                     , rename
                     , retrie
                     , tactic
-                    , hlint
+                    , stan
```
July541, line 365: It looks you replaced hlint with stan?

July541:
Can we have a document as #2066 mentioned?

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -0,0 +1,131 @@
...
+          -- We do not use `prettyShowInspection` cuz Id and Categories is redundant here
```
michaelpj, line 96: You're including "categories", though?

michaelpj:
Yes, please do add something to the "Features" doc.

I would like @wz1000 to look at the hiedb stuff before we merge this.

uhbif19:
Marked as draft, to rebase before merging.

uhbif19:
I tried to use plugin, but it does not show anything to me.

Does something in VS Code Haskell need to be tweaked for this to work?
https://github.com/haskell/vscode-haskell/blob/e81160dbdb6be0ac8493f6dbc0983d27fa096ac7/package.json#L346

pepeiborra:
> Does something in VS Code Haskell need to be tweaked for this to work?

Not at all. I am in my phone right now, but if the test suite passes you should see the diagnostics in VSCode already.

michaelpj:
I'm not sure that's true. I think all plugins automatically get a "globalOn" option that needs to be set to enable them. I'm not sure if there's a way to turn it on by default when testing (@berberman ?).

For vscode, we have an ongoing issue that nobody seems to be sure what the deal is with options in the vscode extension (see https://github.com/haskell/haskell-language-server/issues/2827). It would be great if someone could find out and help us document it.

July541:
> if the test suite passes you should see the diagnostics in VSCode already.

Agree this.

Noticed you are using hint severity,it may not showy as you expected in vscode.

berberman:
> I'm not sure that's true. I think all plugins automatically get a "globalOn" option that needs to be set to enable them. I'm not sure if there's a way to turn it on by default when testing (@berberman ?).

If you don't specify `globalOn` in the lsp config for a plugin, which can be set by `sendNotification SWorkspaceDidChangeConfiguration DidChangeConfigurationParams ...` during the test, the default value `True` will be used:
https://github.com/haskell/haskell-language-server/blob/310e6a4683c40dc5d93ffaf12c45144dc8e079a4/hls-plugin-api/src/Ide/Plugin/Config.hs#L116-L118

So no worries about `globalOn`.

> For vscode, we have an ongoing issue that nobody seems to be sure what the deal is with options in the vscode extension (see #2827). It would be great if someone could find out and help us document it.

Yes, I can write a documentation for that later.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: @wz1000 Please, tell how to refactor this function?

pepeiborra:
is this still in progress or ready for review?

pepeiborra:
@uhbif19  there are stack build failures

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
pepeiborra: > Also, perhaps move this function to the ghcide package and refactor other pieces of similar code around the Core/Compile.hs and Core/Rules.hs modules.

Did you have a chance to look into this refactoring?

pepeiborra (approved):
Looks good to me, but it would be nice to make the suggested refactoring

.github/workflows/test.yml
```diff
@@ -226,6 +226,10 @@ jobs:
         name: Test hls-hlint-plugin test suite
         run: cabal test hls-hlint-plugin --test-options="$TEST_OPTS" || cabal test hls-hlint-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-hlint-plugin --test-options="$TEST_OPTS"
 
+      - if: matrix.test && matrix.ghc != '9.0.1' && matrix.ghc != '9.0.2' && matrix.ghc != '9.2.2'
```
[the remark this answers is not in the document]
uhbif19: Stan does not support 9.x

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
uhbif19: I thought, I was waiting on suggestion where to place the function.

Now I moved `getHieFile` to Development.IDE.Core.Rules.

As for `refactor other pieces of similar code around the Core/Compile.hs and Core/Rules.hs modules` I do not understand what you are talking about. Can you explain what you want to be performed?

pepeiborra: I believe that we have code to create HieFile in Core.Rules and/or Core.Compile

It would be nice to merge it with getHieFile in order to ensure that it stays working.

uhbif19: You mean create separate rule for HieFile?

pepeiborra: No, just abstracting and reusing the same code everywhere

uhbif19: @pepeiborra I do not know which code should be abstracted.

pepeiborra: Fair enough

uhbif19:
@michaelpj CircleCI stuck again.

PR #2911, opened 2022-05-17
Fix hover format

### LINKED ISSUE #2863: Hover content has different behavior between ghc-8.10.7 and ghc-9.2.2
On ghc-8.10.7
image

On ghc-9.2.2
image

Obviously, a newline is missing, I believe it was involved by https://github.com/haskell/haskell-language-server/pull/2848

July541:
Close #2863.

It is because markdown newline requires two '\n' but a separate section only needs one '\n'.

image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2911"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Spans/Common.hs
+++ b/ghcide/src/Development/IDE/Spans/Common.hs
@@ -51,9 +51,7 @@ safeTyThingId _                                = Nothing
 -- Possible documentation for an element in the code
 data SpanDoc
   = SpanDocString HsDocString SpanDocUris
-    -- ^ Extern module doc
   | SpanDocText   [T.Text] SpanDocUris
-    -- ^ Local module doc
   deriving stock (Eq, Show, Generic)
   deriving anyclass NFData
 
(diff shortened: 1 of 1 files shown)


ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -231,11 +231,14 @@ atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ p
...
         prettyName (Right n, dets) = T.unlines $
           wrapHaskell (printOutputable n <> maybe "" (" :: " <>) ((prettyType <$> identType dets) <|> maybeKind))
-          : definedAt n
-          ++ maybeToList (prettyPackageName n)
+          : maybeToList (pretty (definedAt n) (prettyPackageName n))
           ++ catMaybes [ T.unlines . spanDocToMarkdown <$> lookupNameEnv dm n
                        ]
           where maybeKind = fmap printOutputable $ safeTyThingType =<< lookupNameEnv km n
+                pretty Nothing Nothing = Nothing
+                pretty (Just define) Nothing = Just $ define <> "\n"
```
michaelpj, line 239: Completely independent of this PR, but I wonder if we'd make our lives easier by using `prettyprinter` for some of this stuff :thinking:

July541:
> Can we get some comments explaining the place where we need to be careful about newlines? Otherwise we're going to lose the knowledge you just generated!

Yes! all markdown-based content should be treat with care, I'll push more comment to explain what happened now.

ghcide/src/Development/IDE/Spans/AtPoint.hs
July541: Unfortunately, I have little knowledge about `prettyprinter`... Maybe I'll improve this someday or any other can pick this...

michaelpj (approved):

Issue #1423, opened 2021-02-21
Import suggestions for out of scope type class methods in instance declarations


Jashweii:
Example:
```Haskell
data X = X
instance Semigroup X where
  (<>) _ _ = X
  -- stimes is not imported
  stimes _ _ = X
```
HLS should suggest importing stimes (it does if instead used in an expression), and should probably import when using the "add placeholders for <type class complete set>". (Note it could be a qualified import.)

(GHC 8.10.4, cabal 3.2.0.0, stack 2.5.1 (unused), hls 0.9.0.0 (built from latest source), vscode)

July541:
Close and continue in a smaller scope. #3017

PR #2920, opened 2022-05-23
hls-class-plugin enhancement

### LINKED ISSUE #1157: [Feature request] Type sigs on class plugin
Would it be possible to look at the current source file, and if `InstanceSigs` is enabled, also add type signatures to the methods that the "Add placeholders" action generates?

I would add this too, but I don't know
* how to get the type information
* how to get "currently enabled extensions" (if it's possible to do so)

July541:
This pr includes:
1. Add code lens to show missing type signatures for instance methods.
2. Add `with signature` code actions.
3. Add `InstanceSigs` pragma if needed.

Close #1157 

codelens
codeactions


plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -0,0 +1,121 @@
...
+
+data BindInfo = BindInfo
+    { bindSpan     :: SrcSpan
+      -- ^ SrcSpan of the whole binding
+    , bindNameSpan :: SrcSpan
+      -- ^ SrcSpan of the binding name
+    }
+
+rules :: Recorder (WithPriority Log) -> Rules ()
+rules recorder = do
```
Ailrun, line 83: Could you explain why do you think that it would be better to introduce a new rule?

Ailrun (commented):
I will review in more detail once I come back from my vac. However, just one high level Q.

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
July541: Because rules have cache? I'm not sure about that.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                        title = mkTitle $ fst <$> methodGroup
+                        titleWithSig = mkTitleWithSig $ fst <$> methodGroup
+
+                mkTitle methodGroup
+                    = "Add placeholders for "
+                        <> mconcat (intersperse ", " (fmap (\m -> "'" <> m <> "'") methodGroup))
+
+                mkTitleWithSig methodGroup = mkTitle methodGroup <> " with signature(s)"
+
+                mkCmdParams methodGroup withSig =
```
michaelpj, line 121: these functions could live with the `AddMinimalMethodsParams` type, perhaps?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+
+        findClassIdentifier docPath range = do
+            (hieAstResult, pmap) <- handleMaybeM "Unable to GetHieAst"
+                . liftIO
+                . runAction "classplugin.findClassIdentifier.GetHieAst" state
+                $ useWithStale GetHieAst docPath
+            case hieAstResult of
+                HAR {hieAst = hf} ->
+                    pure
+                        $ head . head
```
michaelpj, line 144: can we avoid the partial functions please?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                        $ pointCommand hf (fromJust (fromCurrentRange pmap range) ^. J.start & J.character -~ 1)
```
michaelpj, line 145: more!

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                        ( (Map.keys . Map.filter isClassNodeIdentifier . getNodeIds)
+                            <=< nodeChildren
+                        )
+
+        findClassFromIdentifier docPath (Right name) = do
+            (hscEnv -> hscenv, _) <- handleMaybeM "Unable to GhcSessionDeps"
```
michaelpj, line 151: This view pattern seems weird. Isn't it just renaming the variable? Can't you just do that with a normal pattern i.e. `(hscenv, _) <- ...`?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                . liftIO
+                . runAction "classplugin.findClassFromIdentifier.GhcSessionDeps" state
+                $ useWithStale GhcSessionDeps docPath
+            (tmrTypechecked -> thisMod, _) <- handleMaybeM "Unable to TypeCheck"
```
michaelpj, line 155: ditto

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                $ useWithStale TypeCheck docPath
+            handleMaybeM "Error in TcEnv"
+                . liftIO
+                . fmap snd
+                . initTcWithGbl hscenv thisMod ghostSpan $ do
+                    tcthing <- tcLookup name
+                    case tcthing of
+                        AGlobal (AConLike (RealDataCon con))
+                            | Just cls <- tyConClass_maybe (dataConOrigTyCon con) -> pure cls
+                        _ -> panic "Ide.Plugin.Class.findClassFromIdentifier"
```
michaelpj, line 167: :thinking:

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeLens.hs
```diff
@@ -0,0 +1,133 @@
...
+            $ use GetInstanceBindTypeSigs nfp
+
+        pragmaInsertion <- insertPragmaIfNotPresent state nfp InstanceSigs
+
+        let (hsGroup, _, _, _) = tmrRenamed tmr
+            tycls = hs_tyclds hsGroup
+            -- class instance decls
+            insts = mapMaybe (getClsInstD . unLoc) $ concatMap group_instds tycls
+            -- Declared instance methods without signatures
+            bindInfos = concatMap getBindSpanWithoutSig insts
```
michaelpj, line 48: I don't normally suggest this, but perhaps some of these things would be nicer with list comprehensions:
```
bindInfos = [ bind | bind <- getBindSpanWithoutSig inst, Just inst <- getClsInstD $ unLoc inst_d, inst_d <- group_instds tycls ]
```

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeLens.hs
```diff
@@ -0,0 +1,133 @@
...
+            targetSigs = matchBind bindInfos binds
+            codeLens =
+                (\x@(range, title) ->
```
michaelpj, line 51: personally I try and avoid parenthesized expressions that span multiple lines since it makes the reader keep a "parentheses stack" in their head, which is a bit of a pain.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeLens.hs
```diff
@@ -0,0 +1,133 @@
...
+
+        pure $ List codeLens
+    where
+        uri = _textDocument ^. J.uri
+
+        -- Match Binds with their signatures
+        matchBind :: [BindInfo] -> [InstanceBindTypeSig] -> [InstanceBindTypeSig]
+        matchBind binds = map go
+            where
+                update bind sp = bind {bindDefSpan = Just sp}
```
michaelpj, line 65: why?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+
+-- Return (name text, signature text)
+minDefToMethodGroups :: Range -> [InstanceBindTypeSig] -> BooleanFormula Name -> [[(T.Text, T.Text)]]
+minDefToMethodGroups range sigs = go
+    where
+        go (Var mn)   = [[ (T.pack . occNameString . occName $ mn, bindRendered sig)
+                        | sig <- sigs
+                        , inRange range (getSrcSpan $ bindName sig)
+                        , printOutputable mn == T.drop (T.length bindingPrefix) (printOutputable (bindName sig))
+                        ]]
```
Ailrun, line 184: Why is this changed to require extra conditions?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                    <$> _documentChanges
+            , ..
+            }
+
+        workspaceEdit caps old new
+            = diffText caps (uri, old) new IncludeDeletions
+
+-- |
+-- This implementation is ad-hoc in a sense that the diagnostic detection mechanism is
+-- sensitive to the format of diagnostic messages from GHC.
```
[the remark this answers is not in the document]
July541: We can get rid of the dependency on diagnostic, will do it later.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+            pure $ concatMap mkAction $ minDefToMethodGroups range sigs . classMinimalDef $ cls
+            where
+                range = diag ^. J.range
+
+                mkAction :: [(T.Text, T.Text)] -> [Command |? CodeAction]
+                mkAction methodGroup
+                    = [ mkCodeAction title
+                            $ mkLspCommand plId codeActionCommandId title
+                                (Just $ mkCmdParams methodGroup False)
+                    , mkCodeAction titleWithSig
```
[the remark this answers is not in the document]
July541: I have no preference here, both two ways are acceptable for me.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+                            $ mkLspCommand plId codeActionCommandId titleWithSig
+                                (Just $ mkCmdParams methodGroup True)
+                    ]
+                    where
+                        title = mkTitle $ fst <$> methodGroup
+                        titleWithSig = mkTitleWithSig $ fst <$> methodGroup
+
+                mkTitle methodGroup
+                    = "Add placeholders for "
+                        <> mconcat (intersperse ", " (fmap (\m -> "'" <> m <> "'") methodGroup))
```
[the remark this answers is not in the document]
July541: Looks acceptable in most cases, `"Add class method placeholders" ` is vague indeed.

image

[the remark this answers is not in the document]
michaelpj: What about `Foldable`? :trollface:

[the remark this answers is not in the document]
July541: Currently `hls-class-plugin` only supply MINIMAL definitions, `{-# MINIMAL foldMap | foldr #-}` for `Foldable`. But we should be careful when we start working for https://github.com/haskell/haskell-language-server/issues/1423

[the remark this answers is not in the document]
Ailrun: > Add class method placeholders
@michaelpj the problem of this shorthand happens when there are multiple minimal sets. When `f, g` and `g, h` are both minimal sets, we need to provide some info to distinguish them.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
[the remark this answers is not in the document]
Ailrun: What do you mean by "get rid of the dependency on diagnostic"? Inspecting all instances in a file?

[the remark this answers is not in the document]
July541: We've got all instance methods in the rule, it's not hard to filter unimplemented methods.

https://github.com/haskell/haskell-language-server/blob/a8cb453cb4ccd994d9d6294ad6c9a51c5e8efb3c/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs#L89-L114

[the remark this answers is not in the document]
Ailrun: Ah since this PR will anyway inspect all instances to give them types... makes sense.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
July541: Because we passed all binds from type check here, should filter valid ones.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -0,0 +1,187 @@
...
+        methodDiags = filter (\d -> isClassMethodWarning (d ^. J.message)) ghcDiags
+
+        mkActions docPath diag = do
+            ident <- findClassIdentifier docPath range
+            cls <- findClassFromIdentifier docPath ident
+            InstanceBindTypeSigsResult sigs <- handleMaybeM "Unable to GetInstanceBindTypeSigs"
+                $ liftIO
+                $ runAction "classplugin.codeAction.GetInstanceBindTypeSigs" state
+                $ use GetInstanceBindTypeSigs docPath
+            pure $ concatMap mkAction $ minDefToMethodGroups range sigs . classMinimalDef $ cls
```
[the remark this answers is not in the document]
July541: Keep `mkAction` here due to parameter dependency.

July541:
Sorry for the long delay, I think I've resolved all reviews.

michaelpj (approved):
Thanks for the persistent work :)

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
Ailrun: Oh this is only to pick sig... I was confused. Thanks.

PR #2921, opened 2022-05-23
Cleanup hls-hlint-plugin

July541:
Since we have turned to `ghc-lib-parser` on every ghc version, some code is not required.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2921"></a>


eddiemundo (approved):

michaelpj:
@pepeiborra didn't want to remove this initially, I don't know whether he thinks we should still keep it?

pepeiborra:
I am a bit reluctant to remove it until https://github.com/ndmitchell/hlint/issues/1376 is closed

PR #2937, opened 2022-06-01
chore(nix): bump nixpkgs to prevent glibc issues

teto:
I've had errors with the current HLS built via nix since it uses an older
nixpkgs than mine with a different glibc
```
Unexpected usage error
/nix/store/lyl6nysc3i3aqhj6shizjgj0ibnf1pvg-glibc-2.34-210/lib/libc.so.6: undefined symbol: _dl_audit_symbind_alt, version GLIBC_PRIVATE
```

NB: Maybe I should add pre-commit to the devShell since I got:
```
`pre-commit` not found.  Did you forget to activate your virtualenv?
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2937"></a>

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -54,11 +54,11 @@
     },
     "flake-utils": {
       "locked": {
-        "lastModified": 1649676176,
-        "narHash": "sha256-OWKJratjt2RW151VUlJPRALb7OU2S5s+f0vLj4o1bHM=",
+        "lastModified": 1653893745,
+        "narHash": "sha256-0jntwV3Z8//YwuOjzhV2sgJJPt+HY6KhU7VZUL0fKZQ=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "a4b154ebbdc88c8498a5c7b01589addc9e9cb678",
+        "rev": "1ed9fb1935d260de5fe1c2f7ee0ebaae17ed2fa1",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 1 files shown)


guibou:
Approved. The problem with this is that it will break nix flake for all users who are using older nixpkgs version.

A more robust solution would be to encourage users to use the nix flake `follows` feature.

michaelpj (changes requested):
The new `.pre-commit-config.yaml` is a link into your nix store, which seems definitely wrong...

michaelpj:
I don't think the flake is really a reliable way of getting HLS installed on your system for this reason amongst others. Most users just install it from `nixpkgs` directly.

teto:
I had failed to see that the pre-commit hook was versioned. Seems like pre-commit is installed via `nix develop .\#haskell-language-server-922-dev-nix` (I probably ran initally  just a `nix develop .#haskell-language-server-922`).

> I don't think the flake is really a reliable way of getting HLS installed on your system for this reason amongst others. Most users just install it from nixpkgs directly.

When it comes to software installation, I find nix is the most reliable. but true historically (with HIE too) there has been glibc issues with HLS, not sure why.  I sometimes need to fetch a HLS from master because of some bugfixes and flakes make it incredibly easy to test/update.

michaelpj (approved):
This is an improvement regardless!

PR #2941, opened 2022-06-05
hls-explicit-fixity-plugin

### LINKED ISSUE #2019: Feature request: Fixity information in hover
Pardon me if there's a better place to open this issue, or if it has been discussed already. I tried to look for similar suggestions but could not find any.

When dealing with infix operators -heavy expressions, it is sometimes difficult to tell what precedence various operators have. In some cases, different parenthesising can lead to semantically different but equally type-correct expressions.

It would be helpful if HLS could show the fixity of an operator along with its type in the hover. (Not very relevant screenshot follows)

image

July541:
Close #2019.

This plugin will show available fixity explicitly while hovering.

~~The basic code is done, tests will come soon...~~ Done

fixity1
fixity2

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2941"></a>


July541:
The license missing... will upload it tomorrow...

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,85 @@
...
+
+    -- Names at the position
+    let names = case har of
+            HAR _ asts _ _ _ -> concat
+                $ pointCommand asts pos
+                $ \ast -> mapMaybe (either (const Nothing) Just) $ M.keys $ getNodeIds ast
+
+    -- Get fixity from HscEnv for local defined operator will crash the plugin,
+    -- we first try to use ModIface to get fixities, and then use
+    -- HscEnv if ModIface doesn't available.
```
pepeiborra, line 45: Did you look into getting the fixity from the HieAst ?

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,85 @@
...
+        getFixityFromEnv nfp names = do
+            env <- fmap hscEnv
+                $ handleMaybeM "Unable to get GhcSession"
+                $ liftIO
+                $ runAction "ExplicitFixity.GhcSession" state
+                $ use GhcSession nfp
+            fixities <- liftIO
+                $ fmap (map (second fromJust))
+                -- Ignore all default fixity
+                $ fmap (filter (\(_, fixity) -> isJust fixity && fromJust fixity /= defaultFixity))
```
pepeiborra, line 73: mapMaybe would work better here

pepeiborra (commented):
I think you need to add the testsuite to CI

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
July541: I tried HieAst but found nothing, it would be appreciated if you can leave some hints :)

michaelpj:
I think we should consider rolling this into the main hover code. While theoretically we can have multiple hover plugins, the hovers get concatenated together which isn't the *best* way of doing things. I think fixity is uncontroversially useful enough that we could just make it part of the default hover. It might save a bunch of code, too.

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,90 @@
...
+            hi <- handleMaybeM "Unable to get ModIface"
+                $ liftIO
+                $ runAction "ExplicitFixity.GetModIface" state
+                $ use GetModIface nfp
+            let iface = hirModIface hi
+                fixities = filter (\(_, fixity) -> fixity /= defaultFixity)
+                    $ map (\name -> (name, mi_fix iface (occName name))) names
+            -- We don't have much fixities on one position,
+            -- so `nubOn` is acceptable.
+            pure $ toHover $ nubOn snd fixities
```
michaelpj, line 64: There's some shared code between these two, namely: filtering out the default fixity; filtering out duplicates; and converting to a hover. Maybe pull that out to the call site so these functions are simpler.

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,90 @@
...
+    har <- handleMaybeM "Unable to get HieResult"
+        $ liftIO
+        $ runAction "ExplicitFixity.HieResult" state
+        $ use GetHieAst nfp
+
+    -- Names at the position
+    let names = case har of
+            HAR _ asts _ _ _ -> concat
+                $ pointCommand asts pos
+                $ \ast -> mapMaybe (either (const Nothing) Just) $ M.keys $ getNodeIds ast
```
michaelpj, line 44: https://hackage.haskell.org/package/extra-1.7.10/docs/Data-Either-Extra.html#v:eitherToMaybe ?

We already have `extra` in our dependency closure.

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,90 @@
...
+    { pluginHandlers = mkPluginHandler STextDocumentHover hover
+    }
+
+hover :: PluginMethodHandler IdeState TextDocumentHover
+hover state plId (HoverParams (TextDocumentIdentifier uri) pos _) = pluginResponse $ do
+    nfp <- getNormalizedFilePath plId uri
+    har <- handleMaybeM "Unable to get HieResult"
+        $ liftIO
+        $ runAction "ExplicitFixity.HieResult" state
+        $ use GetHieAst nfp
```
michaelpj, line 38: This sequence sure does come up a lot :laughing:

July541:
> I think we should consider rolling this into the main hover code. While theoretically we can have multiple hover plugins, the hovers get concatenated together which isn't the _best_ way of doing things. I think fixity is uncontroversially useful enough that we could just make it part of the default hover. It might save a bunch of code, too.

It turns out I'm a little fear of the performance requirements of ghcide. And it not easy to make 'show fixity' configurable, I can move this to ghcide if we reach a consensus on this plugin's performance and decide it *must* be in hover.

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
July541: Since our new plugins turn to `ExceptT`, these boilerplate codes are ubiquitous...

michaelpj:
> And it not easy to make 'show fixity' configurable, I can move this to ghcide if we reach a consensus on this plugin's performance and decide it must be in hover.

Configuration is bad :) The best option is for it to be fast enough and on by default!

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
drsooch: I promise when I get some time I plan on tackling this 😃

July541:
I'll move this to ghcide later if nobody suspects the performance of the current implementation.

pepeiborra:
I am not worried about the performance, but I would prefer if we fixed how hover plugins get combined rather than putting all the hovers in ghcide.

michaelpj:
> I am not worried about the performance, but I would prefer if we fixed how hover plugins get combined rather than putting all the hovers in ghcide.

I think it's not too bad, I just think that in this instance, since it's such a core piece of information, we might as well put it in the main hover. If e.g. someone actually writes code to show strictness signatures in a hover, that should definitely go in a plugin!

pepeiborra (approved):
Maybe we should land this now and move it to ghcide once we are 100% happy with it

July541:
I've been busy with graduation these days, and I'll pick up my opened prs later :)

pepeiborra:
did we end up putting this hover in ghcide?

July541:
>  did we end up putting this hover in ghcide?

I initially want close this after #2973 is merged.

July541:
Continue in this thread because https://github.com/haskell/haskell-language-server/pull/2973#discussion_r917359794

exe/Plugins.hs
```diff
@@ -209,8 +214,12 @@ idePlugins recorder includeExamples = pluginDescToIdePlugins allPlugins
     -- The ghcide descriptors should come last so that the notification handlers
     -- (which restart the Shake build) run after everything else
       GhcIde.descriptors pluginRecorder
+#if explicitFixity
```
Ailrun, line 217: Could you leave a comment about why we want to put this after `GhcIde.descriptors`?

plugins/hls-explicit-fixity-plugin/test/Main.hs
```diff
@@ -0,0 +1,70 @@
...
+    waitForKickDone
+    h <- getHover doc pos
+    let expected' = "\n" <> sectionSeparator <> expected
+    case h of
+        Nothing -> liftIO $ assertFailure "No hover"
+        Just (Hover contents _) -> case contents of
+          HoverContentsMS _ -> liftIO $ assertFailure "Unexpected content type"
+          HoverContents (MarkupContent mk txt) -> do
+              liftIO
+                $ assertBool ("Failed to find " <> T.unpack expected <> " in hover message")
```
Ailrun, line 65: Might be more helpful for debugging if this provides `txt` as well?

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,160 @@
...
+selectSmallestContainingForFixityTree sp node
+    | sp `containsSpan` fNodeSpan node = Just node
+    | fNodeSpan node `containsSpan` sp = getFirst $ mconcat
+        [ foldMap (First . selectSmallestContainingForFixityTree sp) $ fNodeChildren node
+        , First (Just node)
+        ]
+    | otherwise = Nothing
+
+-- | Transferred from ghcide `pointCommand`
+findInTree :: FixityTrees -> Position -> (FixityTree -> a) -> [a]
```
Ailrun, line 75: This function has a fairly specific usage, so why do we need this general type? I mean, by using `FixityTree -> [a]` as an argument, we can remove one indirection of `Maybe` and `catMaybe`.

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -0,0 +1,160 @@
...
+    where
+        -- Names at the current ast node
+        names :: [Name]
+        names = mapMaybe eitherToMaybe $ M.keys $ getNodeIds ast
+
+        getFixities :: MonadIO m => m [(Name, Fixity)]
+        getFixities = liftIO
+            $ fmap (filter ((/= defaultFixity) . snd) . mapMaybe pickFixity)
+            $ forM names $ \name ->
+                (\(_, fixity) -> (name, fixity))
```
Ailrun, line 141: Minor nitpicking:
```suggestion
                ((,) name . snd)
```

or do you use this for the strictness?

Ailrun (approved):
Thank you for the implementation! Looks good to me overall.

Here are my two cents.

Ailrun (approved):

PR #1439, opened 2021-02-24
Reformat all files

Ailrun:
This is a successor of #1384.

However, some CPP modules do not work with any formatters (including `stylish-haskell`, `ormolu`, `brittany`), so I format them by manually removing and then re-adding CPP parts.

Issue #693, opened 2020-12-22
Should we lint, -wall and format our code base?


alanz:
I have recently spent some time in the code base again.

I can't help noticing that I have several hlint and other warnings, as well as a variety of formatting styles. Format differences naturally occur with multiple contributors of course.

Given the growing size and number of contributors, I think it would be good to enforce an `hlint` config, as well as require compiles to pass `-wall`, or some agreed set of warnings.  This could initially be done via a separate "allow-fail" CI build which we aim to bring into compliance overtime.

And perhaps we should agree to use a specific formatter, and format all the code with it.

konn:
> I can't help noticing that I have several hlint and other warnings, as well as a variety of formatting styles. Format differences naturally occur with multiple contributors of course.

> And perhaps we should agree to use a specific formatter, and format all the code with it.

Even though I'm not a maintainer, but as a spare-time contributor, I strongly agree with this point!
While I tried to make some contributions to HLS, I have already broken the indent styles by accident for some times.
The changes are trivial in some cases so that one can manually re-order import list or adjusting indentation, but sometimes it becomes rather tedious to maintain formatting coherency in per-module manner.

As a contributor, I feel more comfortable if there is project-wide formatter configuration and we can just let HLS to invoke them automatically. That saves essentially unnecessary works both for contributors and reviewers, reducing noise from reviwing process, IMHO.

alanz:
Contentious question: **which** formatter?

To my mind it should be one without requiring much config, or used with the default config.

jneira:
> Contentious question: which formatter?

What about making a poll? 😄 
I am personally fine with any of them, and even i am not too disturbed with the actual situation (although agree in it can save time for all of us). I've found more useful the more semantic oriented `-Wall` and hlint.

Ailrun:
I think in terms of reducing config, ormolu would be a good choice.

peterwicksstringfield:
Reformat the entire code base with whatever format is chosen, then re-run it on each PR before the unit tests. If there are changes they can be automatically committed and the build restarted. So any successful build is a fixed point of the formatter. I don't think it matters which format is chosen, just so long as the computer is the only one who has to actually think about it.

Wall and hlint sound good too.

ndmitchell:
The formatting inconsistency in Ghcide is significantly higher than most code bases, because it went through 4 evolutions and 4 authors before it even became public. I think whatever we do here should also be done in Ghcide too.

[...]

PR #2944, opened 2022-06-08
Support `fourmolu ^>= 0.7`

parsonsmatt:
This PR adds support for `fourmolu-0.7`

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2944"></a>

### DIFF
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -9,12 +9,18 @@ license:            Apache-2.0
 license-file:       LICENSE
 author:             The Haskell IDE Team
 copyright:          The Haskell IDE Team
+homepage:           https://github.com/haskell/haskell-language-server
+bug-reports:        https://github.com/haskell/haskell-language-server/issues
 maintainer:         alan.zimm@gmail.com
 category:           Development
 build-type:         Simple
 extra-source-files:
   LICENSE
   test/testdata/**/*.hs
+
+source-repository head
+  type:     git
+  location: git://github.com/haskell/haskell-language-server.git
 
 library
   exposed-modules:  Ide.Plugin.Fourmolu
@@ -23,7 +29,7 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6
+    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
     , ghc
     , ghc-boot-th
     , ghcide          ^>=1.7
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -29,7 +29,8 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
+    -- , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
+    , fourmolu        == 0.7.0.1
     , ghc
     , ghc-boot-th
     , ghcide          ^>=1.7
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -29,8 +29,7 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    -- , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
-    , fourmolu        == 0.7.0.1
+    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
     , ghc
     , ghc-boot-th
     , ghcide          ^>=1.7
--- a/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
+++ b/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE TypeApplications         #-}
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE OverloadedLabels         #-}
+{-# LANGUAGE CPP                      #-}
 
 module Ide.Plugin.Fourmolu (
     descriptor,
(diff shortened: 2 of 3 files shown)


plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -9,21 +9,27 @@ license:            Apache-2.0
 license-file:       LICENSE
 author:             The Haskell IDE Team
 copyright:          The Haskell IDE Team
+homepage:           https://github.com/haskell/haskell-language-server
+bug-reports:        https://github.com/haskell/haskell-language-server/issues
```
parsonsmatt, line 13: It took me a minute to figure out the Git repository for this package, so I added these fields to the `.cabal` file. Technically an unrelated change, would be easy to put somewhere else.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -1,9 +1,10 @@
+{-# LANGUAGE CPP                      #-}
```
parsonsmatt, line 1: This is used to provide backwards compatibility.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -1,9 +1,10 @@
...
+{-# LANGUAGE DataKinds                #-}
```
parsonsmatt, line 2: This extension is reshuffled because I ran `stylish-haskell` (which appears to be how this file is formatted) and it automatically did that change.

Same for the reordered imports below.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -23,16 +24,18 @@ import           Development.IDE.GHC.Compat      as Compat hiding (Cpp)
...
 import           Ide.Plugin.Properties
-import           Ide.PluginUtils                 (makeDiffTextEdit, usePropertyLsp)
+import           Ide.PluginUtils                 (makeDiffTextEdit,
+                                                  usePropertyLsp)
 import           Ide.Types
 import           Language.LSP.Server             hiding (defaultConfig)
 import           Language.LSP.Types
 import           Language.LSP.Types.Lens         (HasTabSize (tabSize))
 import           Ormolu
+import           Ormolu.Config
```
parsonsmatt, line 34: This is redundant for `fourmolu <= 0.7`. I didn't gate it behind CPP because CPP is ugly.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -9,21 +9,27 @@ license:            Apache-2.0
...
+
 library
   exposed-modules:  Ide.Plugin.Fourmolu
   hs-source-dirs:   src
   ghc-options:      -Wall
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6
+    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
```
georgefst, line 32: This is in the wrong commit, right? Not necessarily actually worth changing now.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
georgefst: So `stylish-haskell` isn't enforced by CI? If not, I'd be tempted to go back to `fourmolu`.

This file was formatted with `stylish-haskell` when it looked like that was becoming necessary (#693, #1439). But I lost track of where the decision on formatting this codebase ended up.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -29,7 +29,8 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
-    , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
+    -- , fourmolu        ^>=0.3 || ^>=0.4 || ^>= 0.6 || ^>= 0.7
```
georgefst: Eugh, I take it we still can't drop support for old versions due to `stylish-haskell`: https://github.com/haskell/haskell-language-server/pull/2254#discussion_r721196303?

(if we could, we could obviously remove all the CPP)

georgefst (approved):
Commit history is a bit weird, but this otherwise looks good!

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
parsonsmatt: I am regrettably a complete `git` noob and generally just squash+merge all my stuff 😅

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
pepeiborra: We have a pre-commit hook that runs stylish-haskell, mentioned in the contributing guidelines. But it is not enforced by CI:

https://github.com/haskell/haskell-language-server/blob/7431cea783f3b92d327635c017dd5de87b4b8774/docs/contributing/contributing.md#L171-L214

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
pepeiborra: stylish-haskell has been updated since #2254  and I think it's very likely that this can be removed. Do you want to check in a follow-up PR?

July541:
I checked the test detail and found all tests are run for fourmolu 0.3.0.0-0.6.0.0 :(

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
georgefst: Ha, I'm very much in the "git is a wonderful concept with a horrible CLI" camp. Since I finally got around to setting up good editor integration, I haven't looked back.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
georgefst: See https://github.com/haskell/haskell-language-server/pull/2950.

georgefst:
> I checked the test detail and found all tests are run for fourmolu 0.3.0.0-0.6.0.0 :(

Yep, https://github.com/haskell/haskell-language-server/pull/2950 fixes the test, and optimistically drops support for old versions (lets see if that passes CI...).

Issue #2940, opened 2022-06-04
Plugin: hls-cabal-plugin


fendor:
## Support for cabal files

This issue tracks the efforts to bring language server support for cabal files to HLS.
Cabal files are an integral part of Haskell projects, but currently, HLS can not help you at all with writing cabal files, there is no error if a construct is unknown, we can not auto-complete field names, we have no knowledge about modules, etc...

Adding first-class support cabal files is a huge amount of work, so we break it up into multiple sub-issues, that can be worked on individually.

#### Formatting of .cabal files

A nice-to-have thing is to be able to format cabal files, which helps to minimise bikeshedding about formatting, minimises git-diffs, etc...
Moreover, formatters can bring long-desired features, such as [auto-expand of exposed modules](https://oleg.fi/gists/posts/2019-08-11-cabal-fmt.html#extra-expand-exposed-modules-and-other-modules).
Thanks to @VeryMilkyJoe, we have https://github.com/haskell/haskell-language-server/pull/2047 which is already a POC of using `cabal-fmt` for formatting and needs to be pushed over the finish line.

`cabal format` is not suited for formatting as it [inlines common sections](https://github.com/haskell/cabal/issues/5734).

#### Completion
Assistance in writing cabal files through completion suggestions.
See #3664 for status and details.

##### Completion of field-names

Auto-completion of field-names is a must-have in a configuration language. What does this entail? Ideally you can just use auto-completion to figure out what fields are available in a specific section.
For example, we should be able to complete: `name`, `cabal-version`, `version`, `author`, `license`, `homepage`, and many more.

As the next step we want to be able to auto-complete within stanzas, such as library, executable, testsuite and benchmark.
In these stanzas, the available fields are different, thus auto-completion needs to be aware of its current location within the cabal file.

##### Completion of cabal values

This is the natural extension of the former: now we want to be able to aid users to fill-in the values in a cabal file and suggest sensible inputs.

For example, when users want to fill in the `cabal-version` not all values are valid, suggesting the ones that we know of (maybe even with short explanations) will make it trivial to write a cabal file.

Keys that we can easily autocomplete the values for: `cabal-version`, `license`, `license-file`, `build-type`, `default-language`, `testsuite > type` and `source-repository > type`.

Keys that require more work to autocomplete: `ghc-options`, `hs-source-dirs`, `exposed-modules`, `other-modules`, `main-is` and last but not least `build-depends`.

#### Snippets

[...]

PR #2047, opened 2021-07-28
Add formatting plugin for cabal files which uses cabal-fmt

VeryMilkyJoe:
This allows formatting of cabal files with hls.

~~Currently waiting for merges on [cabal-fmt](https://github.com/phadej/cabal-fmt/pull/33) and [stylish haskell](https://github.com/haskell/stylish-haskell/pull/375).~~ We gave up on this, also it makes maintenance harder as cabal-fmt supports only one Cabal version at a time.

Closes #183 and part of #2964.

Builds on top of #2945 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2047"></a>

PR #2945, opened 2022-06-09
Teach HLS about different file extensions

fendor:
This PR is the foundation for #2940 and as such, I'd be really happy if we could get a review here before ZuriHac :)
I am working on a minimal example Plugin, that should illustrate that stuff works by tomorrow.

The Idea is to teach `PluginDescriptors` about the file-extension.

This PR is currently untested and only cherry-picked from #2047, the example plugin will have to suffice as tests.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2945"></a>

### DIFF
--- a/exe/Plugins.hs
+++ b/exe/Plugins.hs
@@ -13,6 +13,7 @@ import           Development.IDE                   (IdeState)
 import qualified Development.IDE.Plugin.HLS.GhcIde as GhcIde
 import qualified Ide.Plugin.Example                as Example
 import qualified Ide.Plugin.Example2               as Example2
+import qualified Ide.Plugin.ExampleCabal           as ExampleCabal
 
 -- haskell-language-server optional plugins
 #if qualifyImportedNames
@@ -204,4 +205,5 @@ idePlugins recorder includeExamples = pluginDescToIdePlugins allPlugins
     examplePlugins =
       [Example.descriptor  pluginRecorder "eg"
       ,Example2.descriptor pluginRecorder "eg2"
+      ,ExampleCabal.descriptor pluginRecorder "ec"
       ]
--- a/ghcide/src/Development/IDE/Plugin/HLS.hs
+++ b/ghcide/src/Development/IDE/Plugin/HLS.hs
@@ -58,8 +58,8 @@ asGhcIdePlugin :: Recorder (WithPriority Log) -> IdePlugins IdeState -> Plugin C
 asGhcIdePlugin recorder (IdePlugins ls) =
     mkPlugin rulesPlugins HLS.pluginRules <>
     mkPlugin executeCommandPlugins HLS.pluginCommands <>
-    mkPlugin extensiblePlugins HLS.pluginHandlers <>
-    mkPlugin (extensibleNotificationPlugins recorder) HLS.pluginNotificationHandlers <>
+    mkPlugin extensiblePlugins id <>
+    mkPlugin (extensibleNotificationPlugins recorder) id <>
     mkPlugin dynFlagsPlugins HLS.pluginModifyDynflags
     where
 
@@ -153,55 +153,80 @@ executeCommandHandlers ecs = requestHandler SWorkspaceExecuteCommand execCmd
 
 -- ---------------------------------------------------------------------
 
-extensiblePlugins :: [(PluginId, PluginHandlers IdeState)] -> Plugin Config
+extensiblePlugins :: [(PluginId, PluginDescriptor IdeState)] -> Plugin Config
 extensiblePlugins xs = mempty { P.pluginHandlers = handlers }
   where
+    getPluginDescriptor pid = lookup pid xs
     IdeHandlers handlers' = foldMap bakePluginId xs
-    bakePluginId :: (PluginId, PluginHandlers IdeState) -> IdeHandlers
-    bakePluginId (pid,PluginHandlers hs) = IdeHandlers $ DMap.map
+    bakePluginId :: (PluginId, PluginDescriptor IdeState) -> IdeHandlers
+    bakePluginId (pid,pluginDesc) = IdeHandlers $ DMap.map
       (\(PluginHandler f) -> IdeHandler [(pid,f pid)])
       hs
+      where
+        PluginHandlers hs = HLS.pluginHandlers pluginDesc
     handlers = mconcat $ do
       (IdeMethod m :=> IdeHandler fs') <- DMap.assocs handlers'
       pure $ requestHandler m $ \ide params -> do
         config <- Ide.PluginUtils.getClientConfig
-        let fs = filter (\(pid,_) -> pluginEnabled m pid config) fs'
-        case nonEmpty fs of
-          Nothing -> pure $ Left $ ResponseError InvalidRequest
-            ("No plugin enabled for " <> T.pack (show m) <> ", available: " <> T.pack (show $ map fst fs))
-            Nothing
-          Just fs -> do
-            let msg e pid = "Exception in plugin " <> T.pack (show pid) <> "while processing " <> T.pack (show m) <> ": " <> T.pack (show e)
-            es <- runConcurrently msg (show m) fs ide params
-            let (errs,succs) = partitionEithers $ toList es
-            case nonEmpty succs of
-              Nothing -> pure $ Left $ combineErrors errs
-              Just xs -> do
-                caps <- LSP.getClientCapabilities
-                pure $ Right $ combineResponses m config caps params xs
+        let pluginInfo = map (\(pid,_) -> (pid, getPluginDescriptor pid)) fs'
+        cleanPluginInfo <- collectPluginDescriptors pluginInfo []
+        case cleanPluginInfo of
+          Left err -> pure $ Left err
+          Right pluginInfos -> do
+            let fs = map snd $ filter (\((_, desc), _) -> pluginEnabled m params desc config) (zip pluginInfos fs')
+            case nonEmpty fs of
+              Nothing -> pure $ Left $ ResponseError InvalidRequest
+                ("No plugin enabled for " <> T.pack (show m) <> ", available: " <> T.pack (show $ map fst fs))
+                Nothing
+              Just fs -> do
+                let msg e pid = "Exception in plugin " <> T.pack (show pid) <> "while processing " <> T.pack (show m) <> ": " <> T.pack (show e)
+                es <- runConcurrently msg (show m) fs ide params
+                let (errs,succs) = partitionEithers $ toList es
+                case nonEmpty succs of
+                  Nothing -> pure $ Left $ combineErrors errs
+                  Just xs -> do
+                    caps <- LSP.getClientCapabilities
+                    pure $ Right $ combineResponses m config caps params xs
+
+collectPluginDescriptors :: [(PluginId, Maybe (PluginDescriptor c))] -> [(PluginId, PluginDescriptor c)] -> LSP.LspM Config (Either ResponseError [(PluginId, PluginDescriptor c)])
+collectPluginDescriptors ((pid, Nothing):_) _ = pure $ Left $ ResponseError InvalidRequest
+        ("No plugindescriptor found for " <>  pidT <> ", available: ")
+        Nothing
+  where
+    PluginId pidT = pid
+collectPluginDescriptors ((pid, Just desc):xs) ys = collectPluginDescriptors xs (ys ++ [(pid, desc)])
+collectPluginDescriptors [] ys = pure $ Right ys
+
 -- ---------------------------------------------------------------------
 
-extensibleNotificationPlugins :: Recorder (WithPriority Log) -> [(PluginId, PluginNotificationHandlers IdeState)] -> Plugin Config
+extensibleNotificationPlugins :: Recorder (WithPriority Log) -> [(PluginId, PluginDescriptor IdeState)] -> Plugin Config
 extensibleNotificationPlugins recorder xs = mempty { P.pluginHandlers = handlers }
   where
+    getPluginDescriptor pid = lookup pid xs
     IdeNotificationHandlers handlers' = foldMap bakePluginId xs
-    bakePluginId :: (PluginId, PluginNotificationHandlers IdeState) -> IdeNotificationHandlers
-    bakePluginId (pid,PluginNotificationHandlers hs) = IdeNotificationHandlers $ DMap.map
+    bakePluginId :: (PluginId, PluginDescriptor IdeState) -> IdeNotificationHandlers
+    bakePluginId (pid,pluginDesc) = IdeNotificationHandlers $ DMap.map
       (\(PluginNotificationHandler f) -> IdeNotificationHandler [(pid,f pid)])
       hs
+      where PluginNotificationHandlers hs = HLS.pluginNotificationHandlers pluginDesc
     handlers = mconcat $ do
       (IdeNotification m :=> IdeNotificationHandler fs') <- DMap.assocs handlers'
       pure $ notificationHandler m $ \ide vfs params -> do
         config <- Ide.PluginUtils.getClientConfig
-        let fs = filter (\(pid,_) -> plcGlobalOn $ configForPlugin config pid) fs'
-        case nonEmpty fs of
-          Nothing -> do
-              logWith recorder Info LogNoEnabledPlugins
-              pure ()
-          Just fs -> do
-            -- We run the notifications in order, so the core ghcide provider
-            -- (which restarts the shake process) hopefully comes last
-              mapM_ (\(pid,f) -> otTracedProvider pid (fromString $ show m) $ f ide vfs params) fs
+        let pluginInfo = map (\(pid,_) -> (pid, getPluginDescriptor pid)) fs'
+        cleanPluginInfo <- collectPluginDescriptors pluginInfo []
+        case cleanPluginInfo of
+          Left _ -> pure ()
+          Right pluginInfos -> do
+            let fs = map snd $ filter (\((_, desc), _) -> pluginEnabled2 m params desc config) (zip pluginInfos fs')
+            case nonEmpty fs of
+                Nothing -> do
+                  logWith recorder Info LogNoEnabledPlugins
+                  pure ()
+                Just fs -> do
+                  -- We run the notifications in order, so the core ghcide provider
+                  -- (which restarts the shake process) hopefully comes last
+                  mapM_ (\(pid,f) -> otTracedProvider pid (fromString $ show m) $ f ide vfs params) fs
 
 -- ---------------------------------------------------------------------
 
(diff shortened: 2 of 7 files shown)


hls-plugin-api/src/Ide/Types.hs
```diff
@@ -177,11 +179,14 @@ class HasTracing (MessageParams m) => PluginMethod m where
   combineResponses _method _config _caps _params = sconcat
 
 instance PluginMethod TextDocumentCodeAction where
-  pluginEnabled _ = pluginEnabledConfig plcCodeActionsOn
+  pluginEnabled _ msgParams pluginDesc
+    | pluginResponsible uri pluginDesc = pluginEnabledConfig plcCodeActionsOn (pluginId pluginDesc)
+    | otherwise = const False
```
pepeiborra, line 184:
```suggestion
  pluginEnabled _ msgParams pluginDesc config =
    pluginResponsible uri pluginDesc && pluginEnabledConfig plcCodeActionsOn (pluginId pluginDesc) config
```

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -270,32 +300,82 @@ instance PluginMethod TextDocumentCompletion where
...
+instance PluginNotificationMethod TextDocumentDidOpen where
+
+instance PluginNotificationMethod TextDocumentDidChange where
+
+instance PluginNotificationMethod TextDocumentDidSave where
+
+instance PluginNotificationMethod TextDocumentDidClose where
+
+instance PluginNotificationMethod WorkspaceDidChangeWatchedFiles where
+  pluginEnabled2 _ _ desc conf = plcGlobalOn $ configForPlugin conf (pluginId desc)
```
pepeiborra, line 366: I'm a bit dubious about skipping some file notifications and not others. Why not filter the watched file notifications too?

pepeiborra (approved):
This looks good to me. Are you going to send a companion change to the VSCode Haskell extension to update the metadata?

fendor:
Yeah, the companion PR is here: https://github.com/haskell/vscode-haskell/pull/618
FYI, currently it doesn't work, .cabal files are type-checked atm :angel: I am trying to fix it today, if I don't manage to, I will just tell everyone they should base their work off this PR while I fix it.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -162,7 +164,7 @@ defaultConfigDescriptor = ConfigDescriptor True False (mkCustomConfig emptyPrope
 class HasTracing (MessageParams m) => PluginMethod m where
 
   -- | Parse the configuration to check if this plugin is enabled
-  pluginEnabled :: SMethod m -> PluginId -> Config -> Bool
+  pluginEnabled :: SMethod m -> MessageParams m -> PluginDescriptor c -> Config -> Bool
```
michaelpj, line 167: Is there a bit of an incoherence here? `PluginDescriptor`s can claim to be enabled only on certain file types, but the decision is actually made based on the message params, which need not include a URI. 

Maybe this is fine, we just need to document it. e.g. what happens for methods that _don't_ have a URI in the params? Plugins are on? or off?

hls-plugin-api/src/Ide/Types.hs
fendor: Yeah, I don't remember why it is this way...

hls-plugin-api/src/Ide/Types.hs
fendor: Thanks!

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -205,12 +209,31 @@ instance PluginMethod TextDocumentCodeAction where
         , Just caKind <- ca ^. kind = any (\k -> k `codeActionKindSubsumes` caKind) allowed
         | otherwise = False
 
+pluginResponsible :: Uri -> PluginDescriptor c -> Bool
```
michaelpj, line 212: haddock

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -270,32 +299,82 @@ instance PluginMethod TextDocumentCompletion where
...
 
 instance PluginMethod TextDocumentFormatting where
-  pluginEnabled _ pid conf = (PluginId $ formattingProvider conf) == pid
-  combineResponses _ _ _ _ (x :| _) = x
+  pluginEnabled STextDocumentFormatting msgParams pluginDesc conf =
+    pluginResponsible uri pluginDesc && PluginId (formattingProvider conf) == pid
+    where
+      uri = msgParams ^. J.textDocument . J.uri
+      pid = pluginId pluginDesc
+  combineResponses _ _ _ _ x = sconcat x
```
michaelpj, line 307: No! we only ever want to return formatting from one plugin! this is weird (I have an issue complaining about it), but I think this is worse.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -270,32 +299,82 @@ instance PluginMethod TextDocumentCompletion where
...
 
 instance PluginMethod TextDocumentPrepareCallHierarchy where
-  pluginEnabled _ = pluginEnabledConfig plcCallHierarchyOn
+  pluginEnabled _ msgParams pluginDesc conf = pluginResponsible uri pluginDesc
+      && pluginEnabledConfig plcCallHierarchyOn pid conf
+    where
+      uri = msgParams ^. J.textDocument . J.uri
+      pid = pluginId pluginDesc
 
 instance PluginMethod TextDocumentSelectionRange where
```
michaelpj, line 325: There is a document reference in the message params, should use it.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -270,32 +299,82 @@ instance PluginMethod TextDocumentCompletion where
...
-  pluginEnabled _ = pluginEnabledConfig plcSelectionRangeOn
+  pluginEnabled _ _ pluginDesc conf = pluginEnabledConfig plcSelectionRangeOn pid conf
+    where
+      pid = pluginId pluginDesc
   combineResponses _ _ _ _ (x :| _) = x
 
 instance PluginMethod CallHierarchyIncomingCalls where
```
michaelpj, line 331: no uri, but deserves a comment explicitly saying that

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -270,32 +299,82 @@ instance PluginMethod TextDocumentCompletion where
...
 
 instance PluginMethod CustomMethod where
-  pluginEnabled _ _ _ = True
+  pluginEnabled _ _ _ _ = True
   combineResponses _ _ _ _ (x :| _) = x
 
 -- ---------------------------------------------------------------------
 
+class HasTracing (MessageParams m) => PluginNotificationMethod (m :: Method FromClient Notification)  where
+  pluginEnabled2 :: SMethod m -> MessageParams m -> PluginDescriptor c -> Config -> Bool
```
michaelpj, line 348: we can have better names, surely. `pluginNotificationMethodEnabled`+`pluginMethodEnabled`?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -270,32 +299,82 @@ instance PluginMethod TextDocumentCompletion where
...
+
+  default pluginEnabled2 :: (HasTextDocument (MessageParams m) doc, HasUri doc Uri)
```
michaelpj, line 350: feels a bit weird having a default method here but not for the other one.

michaelpj: indeed, you could do something very similar there

hls-plugin-api/src/Ide/Types.hs
michaelpj: This also seems a bit risky, in that e.g. this means that HLS presumably won't track cabal files in its VFS, which... might be okay? Or might not be.

fendor: We do want it to track it in its VFS. But why does this line affect the VFS?

michaelpj: Um, maybe I'm getting confused. This will stop the `ghcide` file change notification handlers from picking it up, but I forgot that `lsp` installs its own handlers that maintain the VFS.

hls-plugin-api/src/Ide/Types.hs
fendor: Will remove, can be discussed when finishing #2047

hls-plugin-api/src/Ide/Types.hs
fendor: Intentional bad name, see the comment from here: https://github.com/haskell/haskell-language-server/pull/2047#discussion_r727115061

Help is appreciated to make it typecheck :)

hls-plugin-api/src/Ide/Types.hs
fendor: You think this is good practice? I can imagine other implementations as well

hls-plugin-api/src/Ide/Types.hs
fendor: Indeed, that might be incoherent... I think, the semantics of `pluginEnabled` change with this PR slightly. Previously, it was: "is this plugin disabled from the config?". Now it is "is this plugin disabled or is it even applicable for these parameters?"

Methods that have no URI are not language specific, thus, we only care whether they are disabled.

Since we are conflating two concepts, maybe a completely separate methods makes sense?

michaelpj: `canBeHandledByPlugin`? Since this is talking about a _method_ as a thing, and answering the question of whether a particular plugin wants to handle it.

IMO `pluginEnabled` is a misnomer even today, since it's not associated with the plugin itself but with the plugin method handler.

michaelpj: > Methods that have no URI are not language specific, thus, we only care whether they are disabled.

This doesn't seem obviously true to me. What about `WorkspaceSymbols`?

fendor: Indeed, WorkspaceSymbols will probably have to be handled by the function itself. Maybe such a clear cut separation won't be possible in this situation

michaelpj: I still think it's valuable, we just need to say what it does and doesn't do. "For plugin methods which include information about which document the request is for, requests which are not associated with a document that has one of the listed file types will not be dispatched to plugin methods for this plugin." or something

ghcide/test/exe/Main.hs
```diff
@@ -6714,7 +6714,7 @@ unitTests recorder logger = do
                 ] ++ Ghcide.descriptors (cmapWithPrio LogGhcIde recorder)
 
         testIde recorder (IDE.testing (cmapWithPrio LogIDEMain recorder) logger){IDE.argsHlsPlugins = plugins} $ do
-            _ <- createDoc "haskell" "A.hs" "module A where"
+            _ <- createDoc "A.hs" "haskell" "module A where"
```
fendor, line 6717: https://hackage.haskell.org/package/lsp-test-0.14.0.2/docs/Language-LSP-Test.html#v:createDoc

hls-plugin-api/src/Ide/Types.hs
fendor: It is documented now, quite thorough, do you have something like this in mind?

ghcide/src/Development/IDE/LSP/HoverDefinition.hs
```diff
@@ -53,18 +55,6 @@ foundHover :: (Maybe Range, [T.Text]) -> Maybe Hover
 foundHover (mbRange, contents) =
   Just $ Hover (HoverContents $ MarkupContent MkMarkdown $ T.intercalate sectionSeparator contents) mbRange
 
-setIdeHandlers :: LSP.Handlers (ServerM c)
```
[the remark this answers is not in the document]
fendor: This is necessary, since these IdeHandlers are wired into HLS. Otherwise, Hover will trigger a typecheck, consequentially typechecking the .cabal file.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -121,8 +120,7 @@ runLanguageServer recorder options inH outH getHieDbLoc defaultConfig onConfigur
             unless (reqId `Set.member` cancelled) retry
 
     let ideHandlers = mconcat
```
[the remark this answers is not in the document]
fendor: fair

shell.nix
```diff
@@ -1,11 +1,16 @@
...
-(import (
-  let
-    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
-  in fetchTarball {
-    url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
-    sha256 = lock.nodes.flake-compat.locked.narHash; }
-) {
-  src =  ./.;
-}).shellNix
+with (import <nixpkgs> {});
```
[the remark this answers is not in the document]
fendor: accident

ghcide/src/Development/IDE/LSP/HoverDefinition.hs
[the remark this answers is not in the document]
michaelpj: I mean: make a PR for removing the wired in handlers, then we can merge this one afterwards.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -205,21 +285,132 @@ instance PluginMethod TextDocumentCodeAction where
...
+  pluginEnabled _ _ pluginDesc conf = pluginEnabledConfig plcCallHierarchyOn pid conf
+    where
+      pid = pluginId pluginDesc
+
+instance PluginMethod Request CustomMethod where
+  pluginEnabled _ _ _ _ = True
+
+---
+instance PluginRequestMethod TextDocumentDefinition where
+  combineResponses _ _ _ _ (x :| _) = x
```
michaelpj, line 390: This seems suspicious.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -205,21 +285,132 @@ instance PluginMethod TextDocumentCodeAction where
...
+
+instance PluginRequestMethod TextDocumentTypeDefinition where
+  combineResponses _ _ _ _ (x :| _) = x
```
michaelpj, line 393: ditto

michaelpj (commented):
Oops I missed a file. Thanks for all the Haddock!

ghcide/src/Development/IDE/LSP/LanguageServer.hs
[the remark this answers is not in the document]
fendor: removed

ghcide/src/Development/IDE/LSP/HoverDefinition.hs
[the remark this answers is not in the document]
fendor: sure, incoming in a couple of minutes

exe/Plugins.hs
```diff
@@ -23,6 +24,10 @@ import qualified Ide.Plugin.QualifyImportedNames   as QualifyImportedNames
 import qualified Ide.Plugin.CallHierarchy          as CallHierarchy
 #endif
 
+#if cabal
```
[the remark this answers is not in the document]
fendor: proper cabal plugin removed.

The initial reasoning was, that people at ZuriHac can work in parallel on the plugin, if they all branch off from this PR and they don't need to do the scaffolding on their own.

hls-plugin-api/src/Ide/Types.hs
michaelpj: Probably you couldn't define a plugin that adds go-to-definition results. But actually merging them should be doable: these methods can return a list of results, so we should be able to do it. Can leave it for later, though.

exe/Plugins.hs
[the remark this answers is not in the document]
michaelpj: Makes sense for Zurihac indeed.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -58,8 +58,8 @@ asGhcIdePlugin :: Recorder (WithPriority Log) -> IdePlugins IdeState -> Plugin C
 asGhcIdePlugin recorder (IdePlugins ls) =
     mkPlugin rulesPlugins HLS.pluginRules <>
     mkPlugin executeCommandPlugins HLS.pluginCommands <>
-    mkPlugin extensiblePlugins HLS.pluginHandlers <>
-    mkPlugin (extensibleNotificationPlugins recorder) HLS.pluginNotificationHandlers <>
+    mkPlugin extensiblePlugins id <>
+    mkPlugin (extensibleNotificationPlugins recorder) id <>
```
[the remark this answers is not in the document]
fendor: I am not entirely sure... 

I don't think it is sensible, the replacement would be:

```haskell
    extensiblePlugins recorder ls <>
```

but according to the documentation of `mkPlugin`, this'd need to be rather:
```haskell
    (if null ls then mempty else extensiblePlugins recorder ls) <>
``` 
or something identical.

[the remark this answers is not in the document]
michaelpj: I think I just don't understand how these lines work, never mind.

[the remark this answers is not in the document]
fendor: actually, this looks like a very slightly subtle change we introduced here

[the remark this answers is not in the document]
fendor: I want to restore the previous behaviour before merging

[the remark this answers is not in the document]
fendor: Essentially, we need to find the PluginDescriptor for a given PluginId, because 'IdeHandler' only gives us 
```haskell
fs' :: [(PluginId,
  IdeState
  -> MessageParams a
  -> LspM
       Config (NonEmpty (Either ResponseError (ResponseResult a))))] 
```

That's why we look up the plid in the list of plugins that we know about, but since this unsafe (e.g. maybe the plid doesn't exist in the list of plugin descriptors that we have?) we have `[(PluginId, Maybe PluginDescriptor)]`. So, we clean this up and make this into `[(PluginId, PluginDescriptor)]` (via `collectPluginInfo`, maybe misnomer?). Afterwards, we can finally check whether this Handler of the plugin `plid` is really enabled for the given request. If so, run all the enabled handlers just as the usual. 
The additional code is only to make sure we obtain the `PluginDescriptor` given a plid safely and report an error if there is an internal inconsistency.

I am not sure how to properly refactor this, maybe `IdeHandler` should rather hold a `PluginDescriptor` instead of a `PluginId`, then we wouldn't need this round-tripping.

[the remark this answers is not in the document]
fendor: This did not introduce a change by accident, so it is safe to go. Maybe we can still refactor it?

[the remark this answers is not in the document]
fendor: Refactored, please let me know what you think

michaelpj (approved):
LGTM!

fendor:
@michaelpj Thank you for your time and thorough reviews!

PR #2949, opened 2022-06-12
Don't suggest imports from deprecated modules

### LINKED ISSUE #2415: Import extension plugin: don't suggest imports from deprecated modules
### Your environment

Which OS do you use:
NixOS
Which lsp-client do you use:
emacs
cabal.project

### Steps to reproduce

With `extra` in your package db:

```
module Test where

foo :: [a] -> Int -> a
foo l i = l |? i
```

Import extension plugin will suggest importing `|?` from `Extra`, but `Extra` is deprecated (you're supposed to use `Data.List.Extra`).

### Expected behaviour

Don't suggest imports from deprecated modules (just filter them out).

### Actual behaviour

It suggests importing `|?` from `Extra`.

amesgen:
Closes #2415


michaelpj:
This is a bit tricky, I think your behaviour seems reasonable, but we should capture the design in a comment somewhere.

Things are nicer for completion items, where you can explicitly tag them as deprecated...

michaelpj:
> I'd prefer to mark deprecated items as deprecated instead of hiding them

This PR is only about code actions (despite the branch name), so that doesn't apply. For completions, yeah, we should do that if we don't already.

amesgen:
> I'd prefer to mark deprecated items as `deprecated` instead of hiding them.

Thanks for pointing that out, I agree. Right now, this PR works by completely forgetting that deprecated identifiers exist, so I am now changing that such that the completion code can take that into account (right now, e.g. deprecated are now marked as such in the completions). 

Also, nice touch that the `deprecated` property is deprecated :smile: (in favor of `tags`).

pepeiborra:
Is this ready for review?

PR #2950, opened 2022-06-12
Fix Fourmolu 0.7 support

georgefst:


plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -66,7 +66,7 @@ provider plId ideState typ contents fp fo = withIndefiniteProgress title Cancell
                 (exitCode, out, err) <-
                     readCreateProcessWithExitCode
                         ( proc "fourmolu" $
-                            ["-d"]
+                            ["-d", "--no-cabal"]
```
michaelpj, line 69: Well, here's a reason for the "compile against the package" approach: we don't have a sane way of knowing whether or not to pass this flag, since it depends on which CLI version the user has installed!

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -63,10 +64,26 @@ provider plId ideState typ contents fp fo = withIndefiniteProgress title Cancell
...
+                    let version = do
+                            guard $ exitCode == ExitSuccess
+                            "fourmolu" : v : _ <- pure $ T.words out
+                            pure $ T.splitOn "." v
+                    case version of
+                        Just v -> pure CLIVersionInfo
+                            { noCabal = v >= ["0", "7"]
+                            }
+                        Nothing -> do
+                            T.hPutStrLn stderr "couldn't get Fourmolu version"
```
georgefst, line 78: This is bad error reporting, but I'm about to revamp the logging for the whole plugin anyway, since I've been told off by @fendor for printing directly to `stdout`!

georgefst:
There are some weird things going on with CI.

This PR appeared to be failing for a while, with some impenetrable error in one of the post-job tests. But that somehow sorted itself out without any intervention.

And then, after merging, there's [this failure](https://github.com/haskell/haskell-language-server/actions/runs/2489991419).

michaelpj:
It seemed like it was stuck, so I restarted some jobs, that's why!

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
georgefst: https://github.com/haskell/haskell-language-server/pull/2959

PR #2951, opened 2022-06-12
Drop support for pre-0.7 versions of the Fourmolu library

georgefst:
Can't be merged while we still support GHC < 8.10, due to the dependency on `ghc-lib-parser-9.2`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2951"></a>

### DIFF
--- a/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
+++ b/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
@@ -29,7 +29,7 @@ library
   build-depends:
     , base            >=4.12 && <5
     , filepath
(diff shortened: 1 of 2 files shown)


michaelpj:
Hmm, that's not going to happen soon probably. Per our deprecation policy, we're only going to drop support for 8.8 once Stackage LTS is on 9.2.

georgefst:
> Hmm, that's not going to happen soon probably. Per our deprecation policy, we're only going to drop support for 8.8 once Stackage LTS is on 9.2.

Indeed, hence the draft status. Seeing as I'd written the code before I realised we needed to be compatible so far back, I figured I may as well put it up.

georgefst:
This is pretty out-of-date after #3078. Which makes it slightly easier to maintain support for older versions anyway, be abstracting shims in to a separate module.

PR #2954, opened 2022-06-13
Parse .cabal files; show error and warning diagnostics

runeksvendsen:
Depends on https://github.com/haskell/haskell-language-server/pull/2945

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2954"></a>


runeksvendsen:
Wait... do all the plugins share the same dependency version? This plugin uses the parser from Cabal==3.6.2.0, and by looking at the CI logs it fails because it tries to use the same Cabal version for everything? :-/

michaelpj:
All of the Haskell packages are compiled together into a single build plan by Cabal, yes. But I don't think any other plugins impose version constraints on Cabal. Possibly there's something to do with the Cabal library being bundled with GHC? It wouldn't shock me if we were only able to enable this on newer GHCs which give us a newer cabal library.

CODEOWNERS
```diff
@@ -9,6 +9,7 @@
 # Plugins
 /plugins/hls-alternate-number-format-plugin @drsooch
 /plugins/hls-brittany-plugin @fendor
+/plugins/hls-cabal-plugin @fendor @runeksvendsen
```
kokobd, line 12: Give @runeksvendsen write access?

kokobd:
Mention it in the docs? For example section "Features", "Supported GHC Versions", etc.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
```diff
@@ -0,0 +1,73 @@
...
+  mkDiag fp (T.pack "parsing") DsError (toBeginningOfNextLine pos) msg
+  where
+    msg = T.pack $ showPError (fromNormalizedFilePath fp) err
+
+-- | Produce a diagnostic from a Cabal parser warning
+warningDiagnostic :: NormalizedFilePath -> Lib.PWarning -> FileDiagnostic
+warningDiagnostic fp warning@(Lib.PWarning _ pos _) =
+  mkDiag fp (T.pack "parsing") DsWarning (toBeginningOfNextLine pos) msg
+  where
+    msg = T.pack $ showPWarning (fromNormalizedFilePath fp) warning
```
kokobd, line 35: Can we simplify the warning message? I'm seeing
```
[parsing] [E] /home/kokobd/work/github.com/haskell/haskell-language-server/plugins/hls-cabal-plugin/hls-cabal-plugin.cabal:9:25:
unexpected Unknown SPDX license identifier: 'MIT2'
```
But I believe the file position is not needed, as the message already shows at that location.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Parse.hs
```diff
@@ -0,0 +1,46 @@
...
+    readAndParseFile' fpath = do
+        exists <- Dir.doesFileExist fpath
+        unless exists $
+            Exit.die $
+                "Error Parsing: file \"" ++ fpath ++ "\" doesn't exist. Cannot continue."
+        bs <- BS.readFile fpath
+        parseCabalFileContents bs
+
+parseCabalFileContents
+    :: BS.ByteString
```
hasufell, line 43:
```suggestion
    :: BS.ByteString -- ^ UTF-8 encoded bytestring
```

https://hackage.haskell.org/package/Cabal-3.6.3.0/docs/src/Distribution.PackageDescription.Parsec.html#parseGenericPackageDescription shows that it actually checks for it... and does some pretty fishy stuff if it isn't

plugins/hls-cabal-plugin/hls-cabal-plugin.cabal
```diff
@@ -0,0 +1,66 @@
...
+  exposed-modules:
+    Ide.Plugin.Cabal
+    Ide.Plugin.Cabal.Diagnostics
+    Ide.Plugin.Cabal.LicenseSuggest
+    Ide.Plugin.Cabal.Parse
+
+  build-depends:
+    , base                  >=4.12     && <5
+    , bytestring
+    , Cabal                 ^>=3.2 || ^>=3.4 || ^>=3.6
```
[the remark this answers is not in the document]
fendor: iirc, it would not be a problem to use the same cabal version for all GHC versions here, since Cabal 3.6 still compiles with GHC 8.8. However, stack makes it difficult, as we would have to either allow newer Cabal versions for all dependencies, or make the old resolver use the newer version for virtually all dependencies (e.g. packages that needed only version revisions). Don't know what do exactly do here, however, pre-built executables aren't build with stack so it should always use Cabal 3.6, afaict.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
fendor: They are pretty much unstructured, I am not entirely sure that's possible :( but we can give it a try, nevertheless.

pepeiborra:
@runeksvendsen are you able to wrap up this PR?

fendor:
@pepeiborra I am making sure this PR will be pushed over the finish line. At the moment, we struggle with indeterministic test failure. Also, I want to add more tests, before merging.

michaelpj:
What's the status of this now? It looks like there's a stack of dependent PRs, can we merge this as is and iterate?

fendor:
@michaelpj We made great progress over MuniHac, I had hoped I fixed the CI on monday, I have to check what's going wrong.

fendor:
@runeksvendsen are you interested in code ownership (in the sense of, you will be pinged if there are issues)?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
fendor: Resolved for a follow-up PR, so this is no longer blocks the other hls-cabal-plugin PRs.

plugins/hls-cabal-plugin/hls-cabal-plugin.cabal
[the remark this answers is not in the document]
michaelpj: I think this generally isn't too much of a problem: this plugin mostly operates on cabal files, and different versions of cabal usually treat cabal files the same; differences in behaviour are controlled with the `cabal-version` field. So I think we're okay. BUT worth a comment :)

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/LicenseSuggest.hs
```diff
@@ -0,0 +1,73 @@
...
+  mkCodeAction <$> licenseErrorSuggestion diag
+  where
+    mkCodeAction (original, suggestion) =
+      let
+        -- The Cabal parser does not output the _range_ of the incorrect license identifier,
+        -- only a single source code position. Consequently, in 'Ide.Plugin.Cabal.Diag.errorDiagnostic'
+        -- we define the range to be from the returned position the first column of the next line.
+        -- Since the "replace" code action replaces this range, we need to modify the range to
+        -- start at the first character of the invalid license identifier. We achieve this by
+        -- subtracting the length of the identifier from the beginning of the range.
```
michaelpj, line 45: good comment

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/LicenseSuggest.hs
```diff
@@ -0,0 +1,73 @@
...
+        -- 'Ide.Plugin.Cabal.Diag.errorDiagnostic' ends at the beginning of the following line.
+        tedit = [TextEdit (adjustRange $ _range diag) (suggestion <> "\n")]
+        edit  = WorkspaceEdit (Just $ Map.singleton uri $ List tedit) Nothing Nothing
+      in CodeAction title (Just CodeActionQuickFix) (Just $ List []) Nothing Nothing (Just edit) Nothing Nothing
+
+-- | Given a diagnostic returned by 'Ide.Plugin.Cabal.Diag.errorDiagnostic',
+--   if it represents an "Unknown SPDX license identifier"-error along
+--   with a suggestion then return the suggestion (after the "Do you mean"-text)
+--   along with the incorrect identifier.
+licenseErrorSuggestion
```
michaelpj, line 59: Could be worth a few simple unit tests for this function specifically.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
```diff
@@ -0,0 +1,73 @@
...
+import           Language.LSP.Types     (Diagnostic (..),
+                                         DiagnosticSeverity (..),
+                                         DiagnosticSource, NormalizedFilePath,
+                                         Position (Position), Range (Range),
+                                         fromNormalizedFilePath)
+
+-- | Produce a diagnostic from a Cabal parser error
+errorDiagnostic :: NormalizedFilePath -> Lib.PError -> FileDiagnostic
+errorDiagnostic fp err@(Lib.PError pos _) =
+  mkDiag fp (T.pack "parsing") DsError (toBeginningOfNextLine pos) msg
```
michaelpj, line 26: use `OverloadedStrings`?

michaelpj: I think this is not the most informative thing to put in the `source` field. Parsing what? I think this should probably just say "cabal".

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
```diff
@@ -0,0 +1,73 @@
...
+warningDiagnostic fp warning@(Lib.PWarning _ pos _) =
+  mkDiag fp (T.pack "parsing") DsWarning (toBeginningOfNextLine pos) msg
+  where
+    msg = T.pack $ showPWarning (fromNormalizedFilePath fp) warning
+
+-- | The Cabal parser does not output a _range_ for a warning/error,
+-- only a single source code 'Lib.Position'.
+-- We define the range to be _from_ this position
+-- _to_ the first column of the next line.
+toBeginningOfNextLine :: Lib.Position -> Range
```
michaelpj, line 41: Should we have a version of this that works on LSP positions and put it with the other position utilities? Seems like a thing other people might want.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Parse.hs
```diff
@@ -0,0 +1,46 @@
...
+import           Distribution.Types.Version                   (Version)
+import qualified System.Directory                             as Dir
+import qualified System.Exit                                  as Exit
+
+
+parseCabalFile
+    :: FilePath
+    -> IO ([PWarning], Either (Maybe Version, NonEmpty PError) GenericPackageDescription)
+parseCabalFile =
+    readAndParseFile'
```
michaelpj, line 32: why not inline this?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Parse.hs
```diff
@@ -0,0 +1,46 @@
...
+  where
+    readAndParseFile' fpath = do
+        exists <- Dir.doesFileExist fpath
+        unless exists $
+            Exit.die $
```
michaelpj, line 37: Surely not! Won't this kill HLS??? We definitely need to forbid this function in hlint lol

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+  where
+    log' = logWith recorder
+
+    whenUriFile :: Uri -> (NormalizedFilePath -> IO ()) -> IO ()
+    whenUriFile uri act = whenJust (LSP.uriToFilePath uri) $ act . toNormalizedFilePath'
+
+-- | Helper function to restart the shake session, specifically for modifying .cabal files.
+-- No special logic, just group up a bunch of functions you need for the base
+-- Notification Handlers.
+restartCabalShakeSession :: IdeState -> VFS.VFS -> NormalizedFilePath -> String -> IO ()
```
michaelpj, line 112: Nothing here seems to be cabal-specific, is this not a generic function we use elsewhere? Also why do we need it?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+    summarize (IsCabalFOI OnDisk)           = BS.singleton 1
+    summarize (IsCabalFOI (Modified False)) = BS.singleton 2
+    summarize (IsCabalFOI (Modified True))  = BS.singleton 3
+
+getCabalFilesOfInterestUntracked ::  Action (HashMap NormalizedFilePath FileOfInterestStatus)
+getCabalFilesOfInterestUntracked = do
+    OfInterestCabalVar var <- Shake.getIdeGlobalAction
+    liftIO $ readVar var
+
+addFileOfInterest :: IdeState -> NormalizedFilePath -> FileOfInterestStatus -> IO ()
```
michaelpj, line 235: I get that we have a separate bit of the IDE state for tracking the cabal files of interest, but it seems like the logic for everything else is the same as for Haskell files of interest. Can we not generalize it? It sees like the only thing we'd need is to be generic over a `IO (Var (HashMap NormalizedFilePath FileOfInterestStatus))` or something?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+cabalRules :: Recorder (WithPriority Log) -> Rules ()
+cabalRules recorder = do
+  -- Make sure we initialise the cabal files-of-interest.
+  ofInterestRules recorder
+  -- Rule to produce diagnostics for cabal files.
+  define (cmapWithPrio LogShake recorder) $ \ParseCabal file -> do
+    t <- use GetModificationTime file
+    log' Debug $ LogModificationTime file t
+    mVirtualFile <- Shake.getVirtualFile file
+    contents <- case mVirtualFile of
```
michaelpj, line 137: Do we not have a function for this? Seems pretty standard: get the VFS contents or else read from disk.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+type instance RuleResult ParseCabal = ()
+
+cabalRules :: Recorder (WithPriority Log) -> Rules ()
+cabalRules recorder = do
+  -- Make sure we initialise the cabal files-of-interest.
+  ofInterestRules recorder
+  -- Rule to produce diagnostics for cabal files.
+  define (cmapWithPrio LogShake recorder) $ \ParseCabal file -> do
+    t <- use GetModificationTime file
+    log' Debug $ LogModificationTime file t
```
michaelpj, line 135: I'm not sure what this log means. It doesn't seem like anything is being _modified_ here?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+      Nothing -> do
+        liftIO $ BS.readFile $ fromNormalizedFilePath file
+
+    (pWarnings, pm) <- liftIO $ Parse.parseCabalFileContents contents
+    let warningDiags = fmap (Diagnostics.warningDiagnostic file) pWarnings
+    case pm of
+      Left (_cabalVersion, pErrorNE) -> do
+        let errorDiags = NE.toList $ NE.map (Diagnostics.errorDiagnostic file) pErrorNE
+            allDiags = errorDiags <> warningDiags
+        log' Debug $ LogDiagnostics file allDiags
```
michaelpj, line 148: This logging seems very generic. Maybe we should put it in the shake code so it gets logged for every rule that produces diagnostics?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+        pure (allDiags, Nothing)
+      Right _ -> do
+        log' Debug $ LogDiagnostics file warningDiags
+        pure (warningDiags, Just ())
+
+  action $ do
+    -- Run the cabal kick. This code always runs when 'shakeRestart' is run.
+    -- Must be careful to not impede the performance too much. Crucial to
+    -- a snappy IDE experience.
```
michaelpj, line 157: obscure :thinking:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+    kick
+  where
+    log' = logWith recorder
+
+-- | This is the kick function for the cabal plugin.
+-- We run this action, whenever we need to restart the shake session, which triggers
```
michaelpj, line 163: But it's not called by `restartCabalShakeSession` or indeed anywhere nearby. This whole comment is mysterious :thinking:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -0,0 +1,251 @@
...
+-- Cabal file of Interest rules and global variable
+-- ----------------------------------------------------------------
+
+-- | Cabal files that are currently open in the lsp-client.
+-- Specific actions happen when these files are saved, closed or modified,
+-- such as generating diagnostics, re-parsing, etc...
+--
+-- We need to store the open files to parse them again if we restart the shake session.
+-- Restarting of the shake session happens whenever these files are modified.
+newtype OfInterestCabalVar = OfInterestCabalVar (Var (HashMap NormalizedFilePath FileOfInterestStatus))
```
michaelpj, line 195: I get that we do this for Haskell too... but I wonder why we can't e.g. just traverse the VFS and look for files matching the extension we care about?

plugins/hls-cabal-plugin/test/Main.hs
```diff
@@ -0,0 +1,145 @@
...
+        diags <- waitForDiagnosticsFromSource doc "parsing"
+        reduceDiag <- liftIO $ inspectDiagnostic diags ["Unknown SPDX license identifier: 'BSD3'"]
+        liftIO $ do
+            length diags @?= 1
+            reduceDiag ^. J.range @?= Range (Position 3 24) (Position 4 0)
+            reduceDiag ^. J.severity @?= Just DsError
+        [InR codeAction] <- getCodeActions doc (Range (Position 3 24) (Position 4 0))
+        executeCodeAction codeAction
+        contents <- documentContents doc
+        liftIO $ contents @?= Text.unlines
```
michaelpj, line 120: I'd probably just shove the expected output in a file, but whatever.

michaelpj (approved):
Needs documentation!

michaelpj:
i.e. something in `features.md`, plugin support status.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Parse.hs
fendor: LOL, good catch

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: true it is obscure, but ghcide and hlint do the same thing. This part of the codebase in ghcide is barely documented but absolutely crucial afaict.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: hlint and ghcide do the same thing, you can query for the `kick` function in ghcide.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: having a separate variable seemed like a better seperation to me. Otherwise, you need to remember to always call the correct function...

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
michaelpj: :cry:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/LicenseSuggest.hs
fendor: there are some integration tests, though. You are obviously right, adding two unit tests won't hurt at all, but let me be a horrible software engineer :angry:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Parse.hs
fendor: Code wasn't used :angel:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: Afaict, we do not, we have 
```haskell
moduleText :: (IsString e, MonadLsp c m) => Uri -> ExceptT e m Text
moduleText uri =
    handleMaybeM "mdlText" $
      (virtualFileText <$>)
          <$> getVirtualFile
              (toNormalizedUri uri)
 ```
 in hls-eval-plugin and 
 ```haskell
     provider :: forall m. FormattingMethod m => SMethod m -> PluginMethodHandler a m
    provider m ide _pid params
      | Just nfp <- uriToNormalizedFilePath $ toNormalizedUri uri = do
        mf <- getVirtualFile $ toNormalizedUri uri
        case mf of
          Just vf -> do
            let typ = case m of
                  STextDocumentFormatting -> FormatText
                  STextDocumentRangeFormatting -> FormatRange (params ^. J.range)
                  _ -> error "mkFormattingHandlers: impossible"
            f ide typ (virtualFileText vf) nfp opts
          Nothing -> pure $ Left $ responseError $ T.pack $ "Formatter plugin: could not get file contents for " ++ show uri
```

in hls-plugin-api. Not in the mood for refactoring that right now.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: what documentation would you like to see here? I am eager to share and discuss my finding about the internals of hls, but this is the best I can do in this situation

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
fendor: Added a function to hls-plugin-api

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: Not quite due to pattern matching, etc... I think we'd need a bit more infrastructure for that. Don't think that should be added here.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: Improved docs slightly.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: The intent is to say, that we log the modification time of the file we want to parse. Afaict, this makes sure that our rule depends on the last modification time of that file, and only runs the rule, if the dependency is dirty.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: It looks like, most  Core IDE functions don't use the virtual files at all... very confusing...

fendor: ghcide uses `getFileContentsRule` and `getFileContents` to abstract over `virtualFileText`.

fendor: It is a bit too complicated to re-use everything, but now it is better.

michaelpj: Do you want to quickly note down your findings in an issue or something so we can pick it up later if we want to?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
michaelpj: Okay, I kind of get that (but write it down!), I just find the log confusing. If it was `LogParsingCabalFile file t` that would make more sense to me.

michaelpj: And now it arguably doesn't tell us much, given that we should be relying on GetFileContents?

michaelpj (approved):
LGTM!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: I added a comment, but I am not sure we need more than that

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: In other words, we *do* have a function for it, a rule to be precise. The rule `GetFileContents` abstracts over reading file contents from the virtual file, but we have no rule that alternatively reads the file from disk if it cannot be found on the virtual disk. Not sure that requires an issue.

PR #2956, opened 2022-06-13
hls-class-plugin: Only create placeholders for unimplemented methods

akshaymankar:
Before this change, the `hls-class-plugin` would offer to create placeholders for methods which were already implemented. This change walks through the AST of the instance and figure out which methods exist and only offer to implement those which don't already exist.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2956"></a>

### DIFF
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE CPP               #-}
 {-# LANGUAGE DeriveAnyClass    #-}
 {-# LANGUAGE DeriveGeneric     #-}
+{-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
 {-# LANGUAGE TypeFamilies      #-}
@@ -17,11 +18,12 @@ import           Control.Monad.Trans.Class
 import           Control.Monad.Trans.Maybe
 import           Data.Aeson
 import           Data.Char
+import           Data.Either                             (rights)
 import           Data.List
 import qualified Data.Map.Strict                         as Map
 import           Data.Maybe
-import qualified Data.Text                               as T
 import qualified Data.Set                                as Set
+import qualified Data.Text                               as T
 import           Development.IDE                         hiding (pluginHandlers)
 import           Development.IDE.Core.PositionMapping    (fromCurrentRange,
                                                           toCurrentRange)
@@ -40,7 +42,7 @@ import           Language.LSP.Types
 import qualified Language.LSP.Types.Lens                 as J
 
 #if MIN_VERSION_ghc(9,2,0)
-import           GHC.Hs                                  (AnnsModule(AnnsModule))
+import           GHC.Hs                                  (AnnsModule (AnnsModule))
 import           GHC.Parser.Annotation
 #endif
 
@@ -190,9 +192,16 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
     methodDiags = filter (\d -> isClassMethodWarning (d ^. J.message)) ghcDiags
 
     mkActions docPath diag = do
-      ident <- findClassIdentifier docPath range
+      (HAR {hieAst = ast}, pmap) <-
+          MaybeT . runAction "classplugin" state $ useWithStale GetHieAst docPath
+      instancePosition <- MaybeT . pure $
+                          fromCurrentRange pmap range ^? _Just . J.start
+                          & fmap (J.character -~ 1)
+
+      ident <- findClassIdentifier ast instancePosition
       cls <- findClassFromIdentifier docPath ident
-      lift . traverse mkAction . minDefToMethodGroups . classMinimalDef $ cls
+      implemented <- findImplementedMethods ast instancePosition
+      lift . traverse (mkAction . (\\ implemented)) . minDefToMethodGroups . classMinimalDef $ cls
       where
         range = diag ^. J.range
 
@@ -212,16 +221,14 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
           = InR
           $ CodeAction title (Just CodeActionQuickFix) (Just (List [])) Nothing Nothing Nothing (Just cmd) Nothing
 
-    findClassIdentifier docPath range = do
-      (hieAstResult, pmap) <- MaybeT . runAction "classplugin" state $ useWithStale GetHieAst docPath
-      case hieAstResult of
-        HAR {hieAst = hf} ->
-          pure
-            $ head . head
-            $ pointCommand hf (fromJust (fromCurrentRange pmap range) ^. J.start & J.character -~ 1)
-              ( (Map.keys . Map.filter isClassNodeIdentifier . Compat.getNodeIds)
-                <=< nodeChildren
-              )
+    findClassIdentifier :: HieASTs a -> Position -> MaybeT IO (Either ModuleName Name)
+    findClassIdentifier ast instancePosition =
+      pure
+        $ head . head
+        $ pointCommand ast instancePosition
+          ( (Map.keys . Map.filter isClassNodeIdentifier . Compat.getNodeIds)
+            <=< nodeChildren
+          )
 
     findClassFromIdentifier docPath (Right name) = do
       (hscEnv -> hscenv, _) <- MaybeT . runAction "classplugin" state $ useWithStale GhcSessionDeps docPath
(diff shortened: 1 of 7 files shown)


michaelpj:
cc @July541

plugins/hls-class-plugin/test/Main.hs
```diff
@@ -14,18 +14,22 @@ import qualified Ide.Plugin.Class        as Class
 import qualified Language.LSP.Types.Lens as J
 import           System.FilePath
 import           Test.Hls
+import Development.IDE.Types.Logger
+import Data.Functor.Contravariant (contramap)
```
Ailrun, line 18: Could you format this? (with pre-commit hook or cmd line stylish-haskell)

Ailrun (approved):
LGTM overrall. Thank you for the contribution!

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -234,18 +247,38 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
           _ -> panic "Ide.Plugin.Class.findClassFromIdentifier"
     findClassFromIdentifier _ (Left _) = panic "Ide.Plugin.Class.findClassIdentifier"
 
+    findImplementedMethods :: NormalizedFilePath -> Range -> MaybeT IO [T.Text]
+    findImplementedMethods docPath range = do
+        (HAR {hieAst = hf}, pmap) <- MaybeT . runAction "classplugin" state $ useWithStale GetHieAst docPath
+        pure
+            $ concat
+            $ pointCommand hf (fromJust (fromCurrentRange pmap range) ^. J.start & J.character -~ 1)
```
July541, line 255: It would be great if we could reduce the partial function usage.

July541 (approved):
LGTM.

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -38,16 +41,23 @@ import           Language.Haskell.GHC.ExactPrint.Utils   (rs)
...
+import qualified Development.IDE.Core.Shake              as Shake
 
 #if MIN_VERSION_ghc(9,2,0)
 import           GHC.Hs                                  (AnnsModule(AnnsModule))
 import           GHC.Parser.Annotation
 #endif
 
-descriptor :: PluginId -> PluginDescriptor IdeState
-descriptor plId = (defaultPluginDescriptor plId)
+newtype Log = Log T.Text deriving Show
```
fendor, line 51: I think, it would be a good idea to change this to more explicit constructors instead of a generic text type. The idea being, that the structured logging can have added information later, but if it is unstructured using text, we lose this structured info. In particular, I propose to change it to the following Log type:

```suggestion
data Log 
  = LogImplementedMethods [T.Text]
  deriving Show
```
and changing the `Pretty` method to prepend the message we currently have in the log message further down.

fendor (dismissed):
One comment, otherwise LGTM!
Awesome addition!

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
michaelpj: especially since you're already returning maybes and lists, so you have multiple ways of dealing with missing things!

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -234,18 +247,38 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
...
 
+    findImplementedMethods :: NormalizedFilePath -> Range -> MaybeT IO [T.Text]
+    findImplementedMethods docPath range = do
+        (HAR {hieAst = hf}, pmap) <- MaybeT . runAction "classplugin" state $ useWithStale GetHieAst docPath
+        pure
+            $ concat
+            $ pointCommand hf (fromJust (fromCurrentRange pmap range) ^. J.start & J.character -~ 1)
+            $ map (T.pack . getOccString) . rights . findInstanceValBindIdentifiers
+
+    findInstanceValBindIdentifiers :: HieAST a -> [Identifier]
```
michaelpj, line 258: This could do with some haddock and/or explanation. what is it doing and why?

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
akshaymankar: Um, I am not sure which function is partial here, Can you please help me with that?

michaelpj: `fromJust` will crash the program if the argument is `Nothing`!

akshaymankar: Oh right, i copy-pasted it from above and didn't even see :see_no_evil:

akshaymankar: Fixed now.

plugins/hls-class-plugin/test/Main.hs
akshaymankar: Done.

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
akshaymankar: I also added name of the class to the log.

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
akshaymankar: I added an explanation, please let me know if it makes sense.

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -234,18 +253,55 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
...
+    findInstanceValBindIdentifiers ast
+      | Map.null nodeIds
+        || hasEvidenceBind nodeIds = concatMap findInstanceValBindIdentifiers (nodeChildren ast)
+      | otherwise = Map.keys
+                  . Map.filter (not . Set.null)
+                  . Map.map (Set.filter isInstanceValBind . identInfo)
+                  $ nodeIds
+      where
+          nodeIds = getNodeIds ast
+          hasEvidenceBind = not . Map.null . Map.filter (any isEvidenceBind . identInfo)
```
michaelpj, line 287:
```suggestion
          hasEvidenceBind m = any isEvidenceBind $ concatMap (identInfo) $ Map.elems m
```

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
michaelpj: Why can't we just transitively look for all the things that are `InstanceValBind`s underneath, rather than having to specifically say which things we'll recurse through? I'd have thought that the only `InstanceValBind`s in a instance declaration would be precisely the already-implemented methods?

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -234,18 +253,55 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
           _ -> panic "Ide.Plugin.Class.findClassFromIdentifier"
     findClassFromIdentifier _ (Left _) = panic "Ide.Plugin.Class.findClassIdentifier"
 
+    findImplementedMethods :: NormalizedFilePath -> Range -> MaybeT IO [T.Text]
+    findImplementedMethods docPath range = do
+        (HAR {hieAst = hf}, pmap) <- MaybeT . runAction "classplugin" state $ useWithStale GetHieAst docPath
```
michaelpj, line 258: Both this and `findClassIdentifier` run an action to get the HIE AST. I'm not 100% sure, but I suspect it may be more efficient to only do that once. Perhaps pull the action out of the two functions and just pass in the HIE ast to them?

michaelpj: Also, they both run a `pointCommand` over the same range. Maybe not worth the refactoring, but in principle you could just run one that returns both the things we're interested in.

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
```diff
@@ -234,18 +253,55 @@ codeAction state plId (CodeActionParams _ _ docId _ context) = liftIO $ fmap (fr
...
+    -- gets bound over the 'Match' which contains the 'InstanceValBind', so to
+    -- find the 'InstanceValBind' this function must look inside any
+    -- 'EvidenceVarBind's it finds.
+    findInstanceValBindIdentifiers :: HieAST a -> [Identifier]
+    findInstanceValBindIdentifiers ast
+      | Map.null nodeIds
+        || hasEvidenceBind nodeIds = concatMap findInstanceValBindIdentifiers (nodeChildren ast)
+      | otherwise = Map.keys
+                  . Map.filter (not . Set.null)
+                  . Map.map (Set.filter isInstanceValBind . identInfo)
```
Ailrun, line 283:
```suggestion
      | otherwise = Map.keys
                  . Map.filter (not . Set.null . Set.filter isInstanceValBind . identInfo)
```

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
akshaymankar: I couldn't use `concatMap` as `identiInfo` returns a `Set`, so I used `mconcat $ map identInfo $ ...`

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
akshaymankar: I reduced it further to `Map.filter (any isInstanceValBind . identInfo)`

plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
akshaymankar: Hmm, yeah makes the code much simpler: 

```haskell
    findInstanceValBindIdentifiers :: HieAST a -> [Identifier]
    findInstanceValBindIdentifiers ast =
        let valBindIds = Map.keys
                         . Map.filter (any isInstanceValBind . identInfo)
                         $ getNodeIds ast
        in valBindIds <> concatMap findInstanceValBindIdentifiers (nodeChildren ast)
```

I guess this will perform worse as this is now traversing more of the AST, but maybe this is not as bad?

akshaymankar: Actually, this also makes the code compatible with older GHCs, so I am going to do commit it.

michaelpj (approved):
Nice, and thanks for the tests!

Ailrun (approved):
Thank you for the contribution!

PR #2960, opened 2022-06-15
Turn HLS-wrapper into an LSP Server

smatting:
Introduces better error handling to HLS-wrapper to show LSP clients
dedicated error messages.
This should help users understand why their Language Server isn't
starting.

Closes https://github.com/haskell/haskell-language-server/issues/2589

Note: This is continuation of @fendor 's PR https://github.com/haskell/haskell-language-server/pull/2591 developed with his help at Zurihac with notable additions:
- `runLanuageServer` has been refactored to act as an entrypoint for both HLS LSP and the HLS-wrapper LSP. This shoud be true refactoring for the HLS LSP call site
- The HLS-wrapper LSP contains only the bare minimim setup and handlers
- some changes to the wording of error messages

I've manually tested the happy path (wrapper calls hls) and a error case in emacs, where restarting worked.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2960"></a>


ghcide/src/Development/IDE/Main.hs
```diff
@@ -11,127 +12,146 @@ module Development.IDE.Main
 ,testing
 ,Log(..)
 ) where
-import           Control.Concurrent.Extra              (withNumCapabilities)
```
smatting, line 14: This diff is way too big, I'm going to fix this

smatting: fixed it

pepeiborra (approved):
Looks good to me!

fendor (approved):
LGTM, thank you for pushing this through!

wz1000:
Is it still possible to use ghcide/hls without the wrapper?

wz1000:
I don't quite understand how LSP state (eg config, VFS state) updates which could potentially be intercepted by the wrapper are communicated to the main HLS process.

fendor:
HLS-wrapper only activates the LSP mode if there was a breaking start-up issue. For example, when we can't figure out the GHC version, consequentially can't start the main HLS process. The only way forward is still to restart HLS and the wrapper, but there is a UI pop-up that tells you this instead of silently not starting, in combination with a "Restart" button that may be displayed to the user.

ghcide/src/Development/IDE/Main.hs
smatting: This has been resolved: on the master branch there is a commit that ran stylish-haskell on that module

smatting:
I don't understand what test is failing and why. There were some merge conflicts in https://github.com/haskell/haskell-language-server/pull/2960/commits/e7dee82299bf8e27f8924bcb29f9c745c0705a7d but it was only code blocks that have been moved by this PR.

PR #2966, opened 2022-06-17
Add sig lens for where clauses

### LINKED ISSUE #2943: Code lens to display type signature for local bindings in where clauses
Reminded by https://github.com/haskell/haskell-language-server/issues/2019#issuecomment-1144528250, I think it would be great to have a configurable plugin to display a signature code lens for local bindings in where clauses.

Please let me know if there has more places need code lenses.

~~Edited: I'll try to pick bindings from [`RenamedSource`](https://hackage.haskell.org/package/ghc-9.0.2/docs/GHC.html#t:RenamedSource), please let me know if there is a more efficient way.~~ Should be TypeCheckedSource.

July541:
Close #2943.

image

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/TypeLenses.hs
+++ b/ghcide/src/Development/IDE/Plugin/TypeLenses.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE DeriveAnyClass   #-}
 {-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE OverloadedLists  #-}
 {-# LANGUAGE TypeFamilies     #-}
-
 -- | An HLS plugin to provide code lenses for type signatures
 module Development.IDE.Plugin.TypeLenses (
   descriptor,
@@ -15,23 +15,28 @@ module Development.IDE.Plugin.TypeLenses (
 
 import           Control.Concurrent.STM.Stats        (atomically)
 import           Control.DeepSeq                     (rwhnf)
-import           Control.Monad                       (mzero)
+import           Control.Lens                        ((^.))
+import           Control.Monad                       (forM, mzero)
 import           Control.Monad.Extra                 (whenMaybe)
 import           Control.Monad.IO.Class              (MonadIO (liftIO))
 import           Data.Aeson.Types                    (Value (..), toJSON)
 import qualified Data.Aeson.Types                    as A
+import           Data.Generics                       (GenericQ, everything, mkQ,
+                                                      something)
 import qualified Data.HashMap.Strict                 as Map
 import           Data.List                           (find)
-import           Data.Maybe                          (catMaybes)
+import           Data.Maybe                          (catMaybes, fromMaybe,
+                                                      mapMaybe, maybeToList)
+import           Data.String                         (IsString)
 import qualified Data.Text                           as T
 import           Development.IDE                     (GhcSession (..),
                                                       HscEnvEq (hscEnv),
                                                       RuleResult, Rules, define,
                                                       srcSpanToRange)
 import           Development.IDE.Core.Compile        (TcModuleResult (..))
+import           Development.IDE.Core.Rules          (IdeState, runAction)
 import           Development.IDE.Core.RuleTypes      (GetBindings (GetBindings),
                                                       TypeCheck (TypeCheck))
-import           Development.IDE.Core.Rules          (IdeState, runAction)
 import           Development.IDE.Core.Service        (getDiagnostics)
 import           Development.IDE.Core.Shake          (getHiddenDiagnostics, use)
 import qualified Development.IDE.Core.Shake          as Shake
@@ -49,13 +54,16 @@ import           Development.IDE.Types.Logger        (Pretty (pretty), Recorder,
 import           GHC.Generics                        (Generic)
 import           Ide.Plugin.Config                   (Config)
 import           Ide.Plugin.Properties
-import           Ide.PluginUtils                     (mkLspCommand,
+import           Ide.PluginUtils                     (getNormalizedFilePath,
+                                                      handleMaybeM,
+                                                      mkLspCommand,
+                                                      pluginResponse,
                                                       usePropertyLsp)
 import           Ide.Types                           (CommandFunction,
-                                                      CommandId (CommandId),
                                                       PluginCommand (PluginCommand),
                                                       PluginDescriptor (..),
                                                       PluginId,
+                                                      PluginMethodHandler,
                                                       configCustomConfig,
                                                       defaultConfigDescriptor,
                                                       defaultPluginDescriptor,
@@ -66,38 +74,46 @@ import           Language.LSP.Types                  (ApplyWorkspaceEditParams (
                                                       CodeLens (CodeLens),
                                                       CodeLensParams (CodeLensParams, _textDocument),
                                                       Diagnostic (..),
-                                                      List (..), ResponseError,
+                                                      List (..),
+                                                      Method (TextDocumentCodeLens),
+                                                      ResponseError,
                                                       SMethod (..),
                                                       TextDocumentIdentifier (TextDocumentIdentifier),
                                                       TextEdit (TextEdit),
                                                       WorkspaceEdit (WorkspaceEdit))
+import qualified Language.LSP.Types.Lens             as L
 import           Text.Regex.TDFA                     ((=~), (=~~))
 
 data Log = LogShake Shake.Log deriving Show
-
 instance Pretty Log where
   pretty = \case
     LogShake log -> pretty log
 
-typeLensCommandId :: T.Text
+typeLensCommandId :: IsString s => s
 typeLensCommandId = "typesignature.add"
 
 descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
 descriptor recorder plId =
   (defaultPluginDescriptor plId)
     { pluginHandlers = mkPluginHandler STextDocumentCodeLens codeLensProvider
-    , pluginCommands = [PluginCommand (CommandId typeLensCommandId) "adds a signature" commandHandler]
+        <> mkPluginHandler STextDocumentCodeLens whereClauseCodeLens
+    , pluginCommands = [PluginCommand typeLensCommandId "adds a signature" commandHandler]
     , pluginRules = rules recorder
     , pluginConfigDescriptor = defaultConfigDescriptor {configCustomConfig = mkCustomConfig properties}
     }
 
-properties :: Properties '[ 'PropertyKey "mode" ('TEnum Mode)]
+properties :: Properties
+  '[ 'PropertyKey "whereLensOn" 'TBoolean,
+     'PropertyKey "mode" ('TEnum Mode)]
 properties = emptyProperties
   & defineEnumProperty #mode "Control how type lenses are shown"
     [ (Always, "Always displays type lenses of global bindings")
     , (Exported, "Only display type lenses of exported global bindings")
     , (Diagnostics, "Follows error messages produced by GHC about missing signatures")
     ] Always
+  & defineBooleanProperty #whereLensOn
+    "Enable type lens on instance methods"
+    True
 
 codeLensProvider ::
   IdeState ->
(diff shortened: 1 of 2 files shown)


ghcide/test/exe/Main.hs
```diff
@@ -4230,6 +4231,52 @@ addSigLensesTests =
...
+                            ["whereLensOn" A..= A.Bool False]]]]]
+      doc <- createDoc "Sigs.hs" "haskell" content
+      waitForProgressDone
+      lenses <- getCodeLenses doc
+      liftIO $ length lenses @?= 0
+  , test "Simple" "    g = True" "    g :: Bool\n    g = True"
+  , test "Tuple" "    (g,h) = (id, True)" "    g :: a -> a\n    (g,h) = (id, True)"
+  , test "Operator" "    g = ($)" "    g :: (a -> b) -> a -> b\n    g = ($)"
+  , test "Infix" "    a `g` b = a" "    g :: p1 -> p -> p1\n    a `g` b = a"
+  , expectFail $ test "Typeclass" "    g a b = a + b" "    g :: Num a :: a -> a -> a\n    g a b = a + b"
```
July541, line 4261: typeclass constraints in local bindings are not displayed as top level bindings..

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+    }
+
+-- | All where clauses from type checked source.
+findWhereQ :: GenericQ [LHsLocalBinds GhcTc]
+findWhereQ = everything (<>) $ mkQ [] (pure . findWhere)
+    where
+        findWhere :: GRHSs GhcTc (LHsExpr GhcTc) -> LHsLocalBinds GhcTc
+        findWhere = grhssLocalBinds
+
+-- | Find all bindings for **one** where clasure.
```
michaelpj, line 358:
```suggestion
-- | Find all bindings for **one** where clause.
```

I didn't quite understand what this meant?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+data WhereBinding = WhereBinding
+    { bindingId  :: [Id]
+    -- ^ There may multiple ids for one expression.
+    -- e.g. @(a,b) = (1,True)@
+    , bindingLoc :: SrcSpan
+    -- ^ Location for the whole binding.
+    -- Here we use the this to render the type signature at the proper place.
+    --
+    -- Example: For @(a,b) = (1,True)@, it will print the signature after the
+    -- open parenthesis instead of the above of the whole expression.
```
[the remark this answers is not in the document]
michaelpj: The nice thing about this is that you then can have one lens per id, which I think simplifies the whole `WhereBindings` thing significantly. It's also nicer output: singatures on separate lines, rather than inline, and you can only specify a type for a subset of the ids if you only care about some of them.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+    }
+
+-- | Existed bindings in a where clause.
+data WhereBindings = WhereBindings
+    { bindings        :: [WhereBinding]
+    , existedSigNames :: [Name]
```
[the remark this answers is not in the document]
michaelpj: I guess it's hard to identify at the same time, because the signature could be arbitrarily far away in the AST!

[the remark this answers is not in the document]
michaelpj: worth a comment

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+            $ liftIO
+            $ runAction "codeLens.local.TypeCheck" state
+            $ use TypeCheck nfp
+        (hscEnv -> hsc) <- handleMaybeM "Unable to get GhcSession"
+            $ liftIO
+            $ runAction "codeLens.local.GhcSession" state
+            $ use GhcSession nfp
+        let tcGblEnv = tmrTypechecked tmr
+            rdrEnv = tcg_rdr_env tcGblEnv
+            typeCheckedSource = tcg_binds tcGblEnv
```
michaelpj, line 400: I wonder if it's worth trying to merge this with the other code that generates the lenses so we only do this stuff once. Not sure how expensive it is. But then maybe it's hard to manage them being enabled/disabled :thinking:

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+
+            wheres = findWhereQ typeCheckedSource
+            localBindings = mapMaybe findBindingsQ wheres
+
+            -- | Note there may multi ids for one binding
+            bindingToLenses ids span = case srcSpanToRange span of
+                Nothing -> pure []
```
michaelpj, line 407: this is a bit weird, right? maybe we should log here

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+                        $ initTcWithGbl hsc tcGblEnv ghostSpan
+                        $ bindToSig id hsc rdrEnv
+                    pure $ generateWhereLens plId range (T.pack sig)
+
+        lenses <- concat <$> sequence
+            [ bindingToLenses idsWithoutSig bindingLoc
+            | WhereBindings{..} <- localBindings
+            , let sigSpans = getSrcSpan <$> existedSigNames
+            , WhereBinding{..} <- bindings
+            , let idsWithoutSig = filter (\x -> getSrcSpan (idName x) `notElem` sigSpans) bindingId
```
michaelpj, line 419: this seems a bit odd to me: we're checking that a binding doesn't have a signature by comparing source spans. Why not just compare the ids directly?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -294,3 +324,116 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+                let cmd = mkLspCommand plId typeLensCommandId title (Just [toJSON (makeEdit range title)])
+                in  CodeLens range (Just cmd) Nothing
+
+            makeEdit :: Range -> T.Text -> WorkspaceEdit
+            makeEdit range text =
+                let startPos = range ^. L.start
+                    insertChar = startPos ^. L.character
+                    insertRange = Range startPos startPos
+                in WorkspaceEdit
+                    (pure [(uri, List [TextEdit insertRange (text <> "\n" <> T.replicate (fromIntegral insertChar) " ")])])
```
michaelpj, line 437: huh, do we really have to do this? insert spaces on the next line? I'm surprised there isn't something to handle that, and this seems to have the risk of breaking people who use tabs. worth a comment

michaelpj: There's `InsertTextMode` but only for completions :roll_eyes:

ghcide/test/exe/Main.hs
```diff
@@ -4230,6 +4231,52 @@ addSigLensesTests =
...
 
+addSigLensesForWhereClausesTests :: TestTree
+addSigLensesForWhereClausesTests = testGroup
+  "add signature for where clauses"
+  [ testSession "Disbled" $ do
+      let content = T.unlines
+            [ "module Sigs where"
+            , "f :: b"
+            , "f = undefined"
+            , "  where"
```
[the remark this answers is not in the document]
michaelpj: Would be nice to see tests for this and for a variety of "multiple ids in one pattern" cases

ghcide/test/exe/Main.hs
```diff
@@ -4230,6 +4231,52 @@ addSigLensesTests =
             ]
         ]
 
+addSigLensesForWhereClausesTests :: TestTree
+addSigLensesForWhereClausesTests = testGroup
+  "add signature for where clauses"
+  [ testSession "Disbled" $ do
```
michaelpj, line 4237: ?

ghcide/test/exe/Main.hs
```diff
@@ -4230,6 +4231,52 @@ addSigLensesTests =
...
+                ["plugin" A..= A.object
+                    ["ghcide-type-lenses" A..= A.object
+                        ["config" A..= A.object
+                            ["whereLensOn" A..= A.Bool False]]]]]
+      doc <- createDoc "Sigs.hs" "haskell" content
+      waitForProgressDone
+      lenses <- getCodeLenses doc
+      liftIO $ length lenses @?= 0
+  , test "Simple" "    g = True" "    g :: Bool\n    g = True"
+  , test "Tuple" "    (g,h) = (id, True)" "    g :: a -> a\n    (g,h) = (id, True)"
```
michaelpj, line 4258: nothing for h?

ghcide/test/exe/Main.hs
michaelpj: There's also an ugly question about `ScopedTypeVariables`. What do we do if the type references variables bound in the top level type signature? Worth a test at least.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -66,38 +74,46 @@ import           Language.LSP.Types                  (ApplyWorkspaceEditParams (
...
+  '[ 'PropertyKey "whereLensOn" 'TBoolean,
+     'PropertyKey "mode" ('TEnum Mode)]
 properties = emptyProperties
   & defineEnumProperty #mode "Control how type lenses are shown"
     [ (Always, "Always displays type lenses of global bindings")
     , (Exported, "Only display type lenses of exported global bindings")
     , (Diagnostics, "Follows error messages produced by GHC about missing signatures")
     ] Always
+  & defineBooleanProperty #whereLensOn
+    "Enable type lens on instance methods"
```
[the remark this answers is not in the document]
July541: It can be `localLensMode` when we have type signatures for let bindings also.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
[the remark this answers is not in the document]
July541: Currently, it behaves as you described.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
July541: Insert space to align the signature and its definition.

ghcide/test/exe/Main.hs
July541: It expects no lens if we disabled it, I revise this more clearly.

July541:
I rewrite the test to make it clear.

I increasingly find that I know little about type information in Haskell, I'm confused by the concept of `tidy type`, `open type`, `sigma type`...

ghcide/test/data/local-sig-lens/ScopedTypeVariables.expected.hs
```diff
@@ -0,0 +1,8 @@
+{-# LANGUAGE ExplicitForAll #-}
+module ScopedTypeVariables where
+
+f :: forall a b. a -> b -> (a, b)
+f aa bb = (aa, ida bb)
+    where
+        ida :: b -> b
+        ida = id
```
July541, line 8: It fails as you may guess...

ghcide/test/data/local-sig-lens/Typeclass.expected.hs
```diff
@@ -0,0 +1,7 @@
+module Typeclass where
+
+f :: a
+f = undefined
+    where
+        g :: Num a => a -> a -> a
+        g a b = a + b
```
July541, line 7: `Num a` lost here, the same method can show constraints for top-level bindings...

July541 (commented):
Marked two known failed tests which I can't tackle.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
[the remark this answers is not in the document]
michaelpj: Oh, I misunderstood the comment! The comment is saying that it *would* go in the wrong place if you didn't use the span for the whole binding. I get it now!

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
michaelpj: That seems surprising? I would really have expected that the _names_ of the identifiers would be the same? Or maybe there's a different comparison method you're supposed to use there?

ghcide/test/data/local-sig-lens/Typeclass.expected.hs
pepeiborra: Are you saying that the lens cannot deduce `Num a =>` here if the signature is not present? 
I find that ok, constraints belong at the top level, not in local type signatures

July541: Sorry for ignoring this...

> Are you saying that the lens cannot deduce Num a => here if the signature is not present?

Yes.

> I find that ok, constraints belong at the top level, not in local type signatures

But it's not correct without constraints.

July541:
Some notes from what I observed:

1. the symbol `r` has a different type of notation between its definition and application.
image

2. Following 1, I tried to print the ast for the local variable and top variable with the same expression `f a b = a <> b`, where I got a `FunTy`wrapped with a `ForAllTy` for top bindings, and for local bindings, there only a `ForAllTy`, that means we can't get a proper representation for local variables till now.

3. What  I am trying to do:
    1. There has a mystery function I haven't know that can give an explicit forall refer to its corresponding top bindings :(
    2. Extract type variable from top level bindings and then manually construct a forall binding with [mkForAllTy](https://hackage.haskell.org/package/ghc-9.0.2/docs/GHC-Core-Type.html#v:mkForAllTy), it's not easy and I also doubt its feasibility.
    3. We can accept this shortcoming by displaying these by **inlay hints** until we find a solution to build the correct type info.

PR #2973, opened 2022-06-20
Implement explicit fixity in ghcide

July541:
This is a substitute for #2941, which implement in ghcide.

Due to `AtPoint.atPoint` is a pure function, I insert the fixity in `Core/Action.hs` as an entrance.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2973"></a>


ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -203,6 +205,51 @@ gotoDefinition
...
+  where
+    fixityInfo :: HieAST a -> IO (Maybe Range, [T.Text])
+    fixityInfo ast = do
+      let range = realSrcSpanToRange $ nodeSpan ast
+          names = mapMaybe eitherToMaybe $ M.keys $ getNodeIds ast
+          fixities = getFixityFromModIface names
+
+      -- Get fixity from HscEnv for local defined operator will crash,
+      -- we first try to use ModIface to get fixities, and then use
+      -- HscEnv if ModIface doesn't available.
```
michaelpj, line 219: And the reason we can't just rely on the `ModIface` fixities is presumably because it will *only* have fixities for local definitions?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -203,6 +205,51 @@ gotoDefinition
 gotoDefinition withHieDb getHieFile ideOpts imports srcSpans pos
   = lift $ locationsAtPoint withHieDb getHieFile ideOpts imports pos srcSpans
 
+fixityAtPoint :: HieAstResult -> HiFileResult -> HscEnv -> Position -> IO (Maybe (Maybe Range, [T.Text]))
```
michaelpj, line 208: I think it wouldn't be crazy to make `atPoint` run in `MonadIO m`: almost every other function in here does.

ghcide/src/Development/IDE/Spans/AtPoint.hs
michaelpj: Do the tests cover both cases?

July541: > And the reason we can't just rely on the ModIface fixities is presumably because it will only have fixities for local definitions?

If I remember correctly, ModIface only has definitions from packages, and we rely on ghcenv to get fixities for local definitions, I'll confirm it again and comment here.

> Do the tests cover both cases?

The test data file from line 24 is for local definitions.

ghcide/src/Development/IDE/Spans/AtPoint.hs
July541: A little crazy to make a pure function to impure?😉

The pain point to making fixity query be impure is `runTcInteractive`, I feel it may have a pure method here, worth trying.

michaelpj: Sure, that's generally bad, but most everything here has to run in some monad because it's doing stuff with the GHC API. It's a nice coincidence that `atPoint` didn't need it, but I don't think it's fundamental.

July541:
It looks `useE` with `GetModIface` crashed the hover. Don't know how to adjust to pass the tests...

ghcide/test/exe/Main.hs
```diff
@@ -6523,6 +6524,57 @@ expectSameLocations actual expected = do
...
+    , hoverTest "seq" (Position 20 9) "infixr 0 `seq`"
+    , hoverTest "(<|>)" (Position 21 8) "infixl 3 `<|>`"
+    , hoverTest "fixity define" (Position 23 11) "infixr 7 `>>:`"
+    , hoverTest "record" (Position 28 10) "infix 9 `>>::`"
+    , hoverTest "wildcards" (Position 30 5) "infixr 7 `>>:`\n\ninfix 9 `>>::`"
+    , hoverTest "function" (Position 32 11) "infixl 1 `f`"
+    , hoverTest "signature" (Position 35 2) "infixr 9 `>>>:`"
+    , hoverTest "operator" (Position 36 2) "infixr 9 `>>>:`"
+    , hoverTest "escape" (Position 39 2) "infixl 3 `~\\:`"
+    -- It will cause error like: "Failed to load interface for \8216Fixity\8217\nIt is not a module in the current program, or in any known package."
```
July541, line 6553: Get fixity from locally defined module failed because something like `Failed to load interface for \8216Fixity\8217\nIt is not a module in the current program, or in any known package.` It is strange.

It's a trivial case and I think we have an available function even without this.

ghcide/src/Development/IDE/Core/Actions.hs
```diff
@@ -61,9 +61,24 @@ getAtPoint file pos = runMaybeT $ do
   (hf, mapping) <- useE GetHieAst file
   env <- hscEnv . fst <$> useE GhcSession file
   dkMap <- lift $ maybe (DKMap mempty mempty) fst <$> runMaybeT (useE GetDocMap file)
+  tcGblEnv <- tmrTypechecked . fst <$> useE TypeCheck file
```
pepeiborra, line 64: Using `TypeCheck` here is bad. It will make ghcide hovers slower, and it will break the ability to get hover information at cold startup from the local cache. 

Now, if fixity was in a plugin, it would still create all the problems above since `his-plugin-api` collects all the results before sending the response. But one can:
1. Disable the fixity plugin
2. Implement partial responses

I think this is a blocker. Workarounds:

1. Extract back to a plugin
2. Introduce a new build rule for fixity information, and give it a persistent implementation with `addPersistentRule`. The persistent implementation can return some placeholder or you can cache the fixity info in hiedb if desired.

pepeiborra (changes requested):

July541:
Implemented in #2941.

PR #2977, opened 2022-06-20
Implement PluginMethod for hard-wired in handlers

fendor:
Don't hard-wire handlers, but make sure all handlers are associated to a
PluginDescriptor, s.t. we can check that the handler is actually enabled
for a given request.

Extracted from #2945 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2977"></a>

### DIFF
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -205,6 +205,25 @@ instance PluginMethod TextDocumentCodeAction where
         , Just caKind <- ca ^. kind = any (\k -> k `codeActionKindSubsumes` caKind) allowed
         | otherwise = False
 
+instance PluginMethod TextDocumentDefinition where
+  pluginEnabled _ _ _ = True
+  combineResponses _ _ _ _ (x :| _) = x
+
+instance PluginMethod TextDocumentTypeDefinition where
+  pluginEnabled _ _ _ = True
+  combineResponses _ _ _ _ (x :| _) = x
+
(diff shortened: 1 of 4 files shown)


michaelpj (approved):

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -205,6 +205,25 @@ instance PluginMethod TextDocumentCodeAction where
         , Just caKind <- ca ^. kind = any (\k -> k `codeActionKindSubsumes` caKind) allowed
         | otherwise = False
 
+instance PluginMethod TextDocumentDefinition where
+  pluginEnabled _ _ _ = True
+  combineResponses _ _ _ _ (x :| _) = x
```
fendor, line 210: @michaelpj Don't you want to discuss the `combineResponses` function here?

michaelpj: I commented on the other PR: I think this could be made more sensible, but I'm happy to just have this for now. Or do you want to sort it out properly? I think it's not 100% trivial because of the weird type the spec gives it (`Location[] | LocationLink[] | null`), but in principle I think you could merge things (probably by demoting `LocationLink`s into `Location`s if you have to).

michaelpj: i.e. we could do it in a follow up PR so we unblock your other one.

fendor: Yeah, that sounds like it aligns better with my goals :sweat_smile:

PR #2986, opened 2022-06-24
Remove some partial fucntions from Shake.hs

michaelpj:
Also removes a partial pattern match from Action.hs.


fendor (approved):
LGTM, nice improvement!

michaelpj:
Some reproducible test failures on Windows? Baffling. I'm not sure what to do with that :thinking:

July541:
> Some reproducible test failures on Windows? Baffling. I'm not sure what to do with that 🤔

I've noticed this since #2920, it fails randomly and can be passed after some retry.

michaelpj:
Are you sure? It seems pretty repeatable.

July541:
On #2960, it failed at the previous two tries, and succeed the third time. https://github.com/haskell/haskell-language-server/runs/7057233778?check_suite_focus=true

But I really saw your test failed too much indeed.

July541:
It worked mysteriously :)

PR #2987, opened 2022-06-24
Add associated type families to local completions

gasparattila:
Currently, completing exported associated type families defined in the current module incorrectly adds an import, while non-exported ones do not show up at all. This PR fixes both of these issues.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2987"></a>


pepeiborra (approved):
Thank you!!

PR #2988, opened 2022-06-25
Log response errors returned from Plugins

drsooch:
In certain scenarios, Clients do not display `ResponseError`s from the server. This change will allow for __ALL__ `ResponseError`s to be visible in the logs for use in debugging. 

Similarly, I've dropped the "who" "what" "where" when semantics of the `throwPluginError` function. There's no need to figure out where the `ResponseError` came from, just leave it up to the caller to supply a good message. I want to explore what I've dubbed (to myself) `Logging Contexts`. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2988"></a>


ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -44,20 +47,48 @@ import           UnliftIO.Exception           (catchAny)
...
+data Log = LogPluginError ResponseError
+    deriving Show
 
 instance Pretty Log where
   pretty = \case
-    LogNoEnabledPlugins ->
-      "extensibleNotificationPlugins no enabled plugins"
+    LogPluginError err -> responseErrorToLogMessage err
+
+responseErrorToLogMessage :: ResponseError -> Doc a
```
michaelpj, line 57: `prettyResponseError`? Or even we should have a `Pretty ResponseError` instance somewhere...

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -169,19 +194,15 @@ extensiblePlugins recorder xs = mempty { P.pluginHandlers = handlers }
...
-              <> ", available: " <> T.pack (show $ map (\(plid,_,_) -> plid) fs)
-              )
-              Nothing
+          Nothing -> logAndReturnError recorder InvalidRequest (pluginNotEnabled m fs')
           Just fs -> do
             let msg e pid = "Exception in plugin " <> T.pack (show pid) <> "while processing " <> T.pack (show m) <> ": " <> T.pack (show e)
                 handlers = fmap (\(plid,_,handler) -> (plid,handler)) fs
             es <- runConcurrently msg (show m) handlers ide params
             let (errs,succs) = partitionEithers $ toList es
+            unless (null errs) $ forM_ errs $ \err -> logWith recorder Warning $ LogPluginError err
```
michaelpj, line 205: Why `Warning` here and `Error` elsewhere?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -253,10 +253,8 @@ getNormalizedFilePath (PluginId plId) uri = handleMaybe errMsg
         errMsg = T.unpack $ "Error(" <> plId <> "): converting " <> getUri uri <> " to NormalizedFilePath"
 
 -- ---------------------------------------------------------------------
-throwPluginError :: Monad m => PluginId -> String -> String -> ExceptT String m b
-throwPluginError (PluginId who) what where' = throwE msg
-    where
-        msg = (T.unpack who) <> " failed with " <> what <> " at " <> where'
+throwPluginError :: Monad m => String -> ExceptT String m b
```
michaelpj, line 256: inline?

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -44,20 +47,48 @@ import           UnliftIO.Exception           (catchAny)
...
+
+responseErrorToLogMessage :: ResponseError -> Doc a
+responseErrorToLogMessage err =  errorCode <> ":" <+> errorBody
+    where
+        errorCode = pretty $ show $ err ^. LSP.code
+        errorBody = pretty $ err ^. LSP.message
+
+-- various error message specific builders
+pluginNotEnabled :: SMethod m -> [(PluginId, b, a)] -> Text
+pluginNotEnabled method availPlugins = "No plugin enabled for " <> T.pack (show method) <> ", available: " <> T.pack (show $ map (\(plid,_,_) -> plid) availPlugins)
```
michaelpj, line 65: use `pretty` for the plugin ids, define an instance if there isn't one? Could even do some layout so they appear on subsequent lines if that's useful.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -44,20 +47,48 @@ import           UnliftIO.Exception           (catchAny)
...
+
+pluginDoesntExist :: PluginId -> Text
+pluginDoesntExist (PluginId pid) = "Plugin " <> pid <> " doesn't exist"
+
+commandDoesntExist :: CommandId -> PluginId -> [PluginCommand ideState] -> Text
+commandDoesntExist (CommandId com) (PluginId pid) legalCmds = "Command " <> com <> " isn't defined for plugin " <> pid <> ". Legal commands are: " <> T.pack (show $ map commandId legalCmds)
```
michaelpj, line 71: ditto?

michaelpj (approved):
Have you checked what happens when you us this? The one thing I'm a bit worried about is that I think we're calling `window/showMessage` for all error logs, so for a client that *does* show the `ResponseError` to users, we might get the message shown twice?

ghcide/src/Development/IDE/Plugin/HLS.hs
drsooch: I think my reasoning was that in the event `succs` has responses we should just mark errors as `Warnings` but I guess ANY `ResponseError` should be classified as an error.

hls-plugin-api/src/Ide/PluginUtils.hs
drsooch: I plan on coming back to this (assuming people are ok with my suggestion). I want to come up with some way to tie back log messages to their context. So for instance I'm thinking this would eventually return `ExceptT Context m b` where `Context` can provide some information for what happened. This `Context` should be usable in all locations where we log (so internal to `ghcide` or `shake` etc.) This way plugins can just give an error message and the plugin id is automatically attached.

ghcide/src/Development/IDE/Plugin/HLS.hs
drsooch: but yes I will update to Error

ghcide/src/Development/IDE/Plugin/HLS.hs
drsooch: I think an instance would be best, I'll see if it's possible

ghcide/src/Development/IDE/Plugin/HLS.hs
drsooch: Yeah, I hoisted these error message builders outside of the plugin handlers. Just wanted to clean up the surrounding code in the handlers. I'll go for the instance option and see if I can clean up some of these messages

ghcide/src/Development/IDE/Plugin/HLS.hs
michaelpj: Honestly, this code is kind of suspect. I'd be more comfortable with failing if any of them fail, rather than succeeding if any of them succeed :thinking:

hls-plugin-api/src/Ide/PluginUtils.hs
drsooch: I'm not sure It would be in a follow up PR and the main detractor is whatever the solution is, has to cross `ghcide` and `hls-plugin-api` package barrier -- which I'm not sure how to deal with

ghcide/src/Development/IDE/Plugin/HLS.hs
michaelpj: I guess that's true. Hopefully that will still happen, they'll just get more obvious logs...

ghcide/src/Development/IDE/Plugin/HLS.hs
July541: While looking into #2993, I noticed our `handleMaybe` and `pluginResponse` will raise an InternalError with a warning, like `2022-06-28T10:53:51.815585Z | Warning | InternalError: Unable to typecheak`.

We should have two kinds of handle functions to result in warning and error separately I think.

drsooch:
> Have you checked what happens when you us this? The one thing I'm a bit worried about is that I think we're calling `window/showMessage` for all error logs, so for a client that *does* show the `ResponseError` to users, we might get the message shown twice?

I've tested using VSCode and EMacs and I've yet to see anything shown twice. I will however note that maybe `Error` is the wrong log level. In certain scenarios if We fail to build a project the user will get spammed with `Error` notifications. I'd propose to move back to `Warning` for any `ResponseErrors` coming from plugins directly.

michaelpj:
> I will however note that maybe Error is the wrong log level. In certain scenarios if We fail to build a project the user will get spammed with Error notifications.

Okay, so the question is "what does Error mean?". I think the things we see here *are* errors, and so we should be able to log them as such. Perhaps the problem is that we report all Error logs to the user now, so Error carries the additional connotation of "... and it's bad enough that the user should see". Perhaps that's a mistake, which we could fix by e.g. having a level above Error (say Critical) for errors that are bad enough they should be shown to the user.

PR #2994, opened 2022-06-28
GHC 9.4 compatibility + Multiple Home Units

wz1000:
TODO

- CI for 9.4.1
- Fixing testsuite
- ExactPrint stuff (everything that needs it is disabled for now)
- Plugins

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2994"></a>


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -157,14 +157,21 @@ import qualified Development.IDE.Types.Shake as Shake
 import           Development.IDE.GHC.CoreFile
 import           Data.Time.Clock.POSIX             (posixSecondsToUTCTime, utcTimeToPOSIXSeconds)
 import Control.Monad.IO.Unlift
+#if MIN_VERSION_ghc(9,3,0)
+import GHC.Unit.Module.Graph
+import GHC.Unit.Env
+#endif
+import Debug.Trace
```
pepeiborra, line 164: do we need an hlint rule to ban `Debug.Trace` from landing? @michaelpj

pepeiborra:
It sounds like we should move all the code actions that depend on `ghc-exactprint` out of ghcide. Or simply move all the code actions out of ghcide, full stop.

ghcide/src/Development/IDE/Core/Rules.hs
michaelpj: :+1:

michaelpj: https://github.com/haskell/haskell-language-server/pull/3000

guibou:
I've added one commit which adds the required files for GHC 9.4 for nix.

isovector (approved):
LGTM for wingman

konn (approved):
Although I haven't tested in my environment, it seems ok as for the Splice Plugin.

Just curious: is this change to make the splice plugin dependent on `hls-refactor-plugin` also fixes the Splice Plugin also on GHC 9.2, or do we need an extra effort?

georgefst (approved):
Looks fine for my plugins (Ormolu+Fourmolu). Until https://github.com/tweag/ormolu/pull/885 lands, anyway.

wz1000:
> Just curious: is this change to make the splice plugin dependent on hls-refactor-plugin also fixes the Splice Plugin also on GHC 9.2, or do we need an extra effort?

Those changes were from https://github.com/haskell/haskell-language-server/pull/3091 which I rebased on. This patch doesn't add 9.4 support for any plugins, only the core ghcide part.

wz1000:
@guibou could you help with the nix failure?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -269,40 +283,69 @@ captureSplicesAndDeps TypecheckHelpers{..} env k = do
...
+                                        -- Non det OK as we will put it into maps later anyway
+                                       $ nonDetEltsUniqSet mods_transitive
+#endif
+
+#if MIN_VERSION_ghc(9,3,0)
+           ; moduleLocs <- readIORef (fcModuleCache $ hsc_FC hsc_env)
 #else
-                                                                      uid
+           ; moduleLocs <- readIORef (hsc_FC hsc_env)
 #endif
```
pepeiborra, line 328: This sets a new record.
Would it not be just better to have 3 versions of this code, one for each GHC api?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -685,8 +689,9 @@ getModuleGraphRule recorder = defineNoFile (cmapWithPrio LogShake recorder) $ \G
 typeCheckRuleDefinition
     :: HscEnv
     -> ParsedModule
+    -> NormalizedFilePath
     -> Action (IdeResult TcModuleResult)
-typeCheckRuleDefinition hsc pm = do
+typeCheckRuleDefinition hsc pm file = do
```
pepeiborra, line 694: `file` is not used anywhere right?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -772,9 +777,21 @@ ghcSessionDepsDefinition fullModSummary GhcSessionDepsConfig{..} env file = do
...
+            let inLoadOrder = map (\HiFileResult{..} -> HomeModInfo hirModIface hirModDetails Nothing) ifaces
+#if MIN_VERSION_ghc(9,3,0)
+            mss_imports <- uses_ GetLocatedImports (file : deps)
+            final_deps <- forM mss_imports $ \imports -> do
+              let fs = mapMaybe (fmap artifactFilePath . snd) imports
+              dep_mss <- map msrModSummary <$> if fullModSummary
+                    then uses_ GetModSummary fs
+                    else uses_ GetModSummaryWithoutTimestamps fs
+              return (map (NodeKey_Module . msKey) dep_mss)
+            ms <- msrModSummary <$> use_ GetModSummary file
```
pepeiborra, line 789: Can use `GetModSummaryWithoutTimestamps` when `fullModSummary` is false here?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -772,9 +777,21 @@ ghcSessionDepsDefinition fullModSummary GhcSessionDepsConfig{..} env file = do
 
             depSessions <- map hscEnv <$> uses_ (GhcSessionDeps_ fullModSummary) deps
             ifaces <- uses_ GetModIface deps
-
-            let            inLoadOrder = map (\HiFileResult{..} -> HomeModInfo hirModIface hirModDetails Nothing) ifaces
-            session' <- liftIO $ mergeEnvs hsc mss inLoadOrder depSessions
+            let inLoadOrder = map (\HiFileResult{..} -> HomeModInfo hirModIface hirModDetails Nothing) ifaces
+#if MIN_VERSION_ghc(9,3,0)
+            mss_imports <- uses_ GetLocatedImports (file : deps)
```
pepeiborra, line 782: A comment explaining why collecting the mod summaries of the direct dependencies here

pepeiborra (approved):
Minor comments only, as the slice of HLS that I understand shrinks with every GHC update. 

This is a massive amount of work, thank you @wz1000  !!!

ghcide/src/Development/IDE/Core/Rules.hs
wz1000: You were right, there was something fishy here. We were including the `ModSummary` for the current module in GHC 9.4 but not before. I've changed this to always include the current `ModSummary`, which allows us to not query imports of direct dependencies. We still need to get the `ModSummary`s of direct dependencies in order to know their `Module`s, which we need to reference in the `ModuleGraph`.

I've written a short note documenting all the assumptions we now make.

ghcide/src/Development/IDE/Core/Compile.hs
wz1000: I've reduced the nesting, is that any better?

guibou:
> @guibou could you help with the nix failure?

Yes, for sure, I'll have a look at it tomorrow.

PR #2995, opened 2022-06-28
Don't use typecheck rule for non FOIs in refine imports plugin

### LINKED ISSUE #2962: refineImports plugin causes excessive memory use
The refineImports plugin calls `TypeCheck` on all transitively imported modules which violates the important memory invariant that we only store TypeCheck results for files of interest. I recommend disabling this plugin by default until it is fixed.

My experience on GHC is that disabling this plugin causes start-up memory usage to decrease from 3GB to 1GB.

wz1000:
Also add an assertion to check that we never use non-FOI rules

Fixes #2962

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/2995"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -165,6 +165,7 @@ data Log
   | LogLoadingHieFileFail !FilePath !SomeException
   | LogLoadingHieFileSuccess !FilePath
   | LogExactPrint ExactPrint.Log
+  | LogTypecheckedFOI !NormalizedFilePath
   deriving Show
 
 instance Pretty Log where
@@ -182,6 +183,14 @@ instance Pretty Log where
     LogLoadingHieFileSuccess path ->
       "SUCCEEDED LOADING HIE FILE FOR" <+> pretty path
     LogExactPrint log -> pretty log
+    LogTypecheckedFOI path -> vcat
+      [ "WARNING: Typechecked a file which is not currently open in the editor:" <+> pretty (fromNormalizedFilePath path)
+      , "This can indicate a bug which results in excessive memory usage."
+      , "This may be a spurious warning if you have recently closed the file."
+      , "If you haven't opened this file recently, please file a report on the issue tracker mentioning"
+        <+> "the HLS version being used, the plugins enabled, and if possible the codebase and file which"
+        <+> "triggered this warning."
+      ]
 
 templateHaskellInstructions :: T.Text
 templateHaskellInstructions = "https://haskell-language-server.readthedocs.io/en/latest/troubleshooting.html#static-binaries"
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -650,7 +650,11 @@ typeCheckRule :: Recorder (WithPriority Log) -> Rules ()
 typeCheckRule recorder = define (cmapWithPrio LogShake recorder) $ \TypeCheck file -> do
     pm <- use_ GetParsedModule file
     hsc  <- hscEnv <$> use_ GhcSessionDeps file
-    typeCheckRuleDefinition hsc pm
+    foi <- use_ IsFileOfInterest file
+    -- We should only call the typecheck rule for files of interest.
+    -- Keeping typechecked modules in memory for other files is
+    -- very expensive.
+    assert (foi /= NotFOI) $ typeCheckRuleDefinition hsc pm
```
mpickering, line 657: I have this assertion locally and it often fails for me but not necessarily in bad ways, just seems there is a race?

michaelpj:
Ah, the same mysterious ghcide windows failure that I was getting! That makes me feel much better that it's not actually my fault :joy: Would be nice if we had any idea what's going on there though.

ghcide/src/Development/IDE/Core/Rules.hs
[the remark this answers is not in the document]
pepeiborra: Assertions are disabled by -O. So this assert won't trigger for users, and I doubt it will even trigger in CI

pepeiborra: Can you get a call stack?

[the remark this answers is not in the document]
michaelpj: Okay, then I think we should definitely log too! I would like to know if this is happening in general usage as well. A warning log would be fine.

michaelpj:
@wz1000 add a log and I think we can merge this?

ghcide/src/Development/IDE/Core/Rules.hs
wz1000: If you close a file while a typecheck has been queued, then on restart that typecheck will be executed, but the assertion will fail because it is not an FOI any longer.

I think this is harmless because it should be garbage collected soon after?

michaelpj: Sounds like a good reason to make it just a warning :)

wz1000: Yes, but I think the false positive rate can be very high to the point of being annoying. Maybe we should just silently log it, with a message that explains this caveat and asks for a bug report. 

Something to effect of "Warning: Typecheck executed for file that is not currently open. This is harmless if this file was recently open and has now been closed. If not, please report a bug."

pepeiborra: > If you close a file while a typecheck has been queued,

Can the check be moved to the code that schedules the typecheck?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -182,6 +183,14 @@ instance Pretty Log where
     LogLoadingHieFileSuccess path ->
       "SUCCEEDED LOADING HIE FILE FOR" <+> pretty path
     LogExactPrint log -> pretty log
+    LogTypecheckedFOI path -> vcat
+      [ "WARNING: Typechecked a file which is not currently open in the editor:" <+> pretty (fromNormalizedFilePath path)
```
michaelpj, line 187: Logging a message at INFO that begins with `WARNING` makes no sense. Use WARN!

wz1000: How does WARN manifest in the vscode UI?

michaelpj: It appears in the logs, and I think it currently gets shown to the user. But I'm pretty convinced we should turn that off, and I would _much_ rather do that than start weirdly misclassifying logs in order to get different behaviour.

michaelpj: Oh no, only errors get shown to the user at the moment. So it just goes into the log.

PR #3000, opened 2022-06-29
Hlint more partial functions, and Debug.Trace

michaelpj:
Turns out that for functions which are typically imported in multiple ways, you need to list both explicitly, and you can't combine them (see the linked HLint issue). So I did that and added some more exceptions.

I also banned `Debug.Trace`, removing a couple of dead usages and adding exceptions for the existing (!) usages.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3000"></a>

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -241,6 +241,27 @@
   - name: "GHC.Arr.!"
     within: []
 
+  # Tracing functions
+  # We ban an explicit list rather than the
+  # Debug.Trace, because that module also
+  # includes the eventlog tracing functions,
+  # which are legitimate to use.
+  - name: 
+    - Debug.Trace.trace
+    - Debug.Trace.traceId
+    - Debug.Trace.traceShow
+    - Debug.Trace.traceShowId
+    - Debug.Trace.traceStack
+    - Debug.Trace.traceIO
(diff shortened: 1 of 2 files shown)


wz1000:
Can we have global exceptions for `traceEvent`, `traceEventIO`, `traceMarker` and `traceMarkerIO`? These should cover some of the existing module exceptions.

michaelpj:
I'm not sure if I can do that easily, I might have to switch to explicitly listing the banned functions from `Debug.Trace`.

michaelpj:
I'm doing that, but in fact it doesn't let us get rid of any exceptions, since all of those modules also use non-eventlog tracing functions :shrug:

wz1000:
Maybe change those instances to use the existing log mechanism where it is easy? Perhaps a job for another time.

drsooch:
Is there no way to ignore whole directories? For example the test suites seem like it would be ok to allow certain partial functions

michaelpj:
Not sure, I think it works off them module name anyway. I'm also not too bothered by it - the explicit list is not actually that long!

drsooch:
Fair enough. Just looks tedious 😂

drsooch (approved):

Issue #3001, opened 2022-06-29
Docs: Video does not render for hls-selection-range-plugin in features


sloorush:
### Your environment

Which OS do you use: NA

Which LSP client (editor/plugin) do you use: NA

Describe your project (alternative: link to the project):  NA

### Steps to reproduce

Visit https://haskell-language-server.readthedocs.io/en/latest/features.html#selection-range

### Expected behaviour

The video should be rendered, all other videos on the page are .gif files.

### Actual behaviour

It is a .mov file and does not render in the documentation
![image](https://user-images.githubusercontent.com/54478821/176459171-cb7b0a82-cb8d-46e5-a29a-71552c8992cf.png)

### Include debug information

The file should be replaced by a gif instead [here](https://github.com/sloorush/haskell-language-server/blob/0b8c793dfdf0d6adb1d4704343fa512c7646a15a/docs/features.md?plain=1#L326)

sloorush:
Also the URL for the `shrink/expand selection` has been changed to https://code.visualstudio.com/docs/editor/codebasics#_shrinkexpand-selection over [here](https://github.com/sloorush/haskell-language-server/blob/0b8c793dfdf0d6adb1d4704343fa512c7646a15a/docs/features.md?plain=1#L323)

kokobd:
Fixed in master, closing.

Issue #2997, opened 2022-06-28
Re-exported symbol hover ghcide test is flaky on Windows


July541:
Recently windows tests are unstable due to https://github.com/haskell/haskell-language-server/blob/master/ghcide/test/exe/Main.hs#L4442

It's a test for the definition of reexported symbol. I've checked that both the expected result and the real result are all acceptable for goto definition behavior.

Some observations:
1. It failed most time.
2. Retry has a chance to pass the test.

Known affected prs:
1. #2995
2. #2988 
3. #2986 
4. #2960

PR #3003, opened 2022-06-30
refactor selection range plugin

### LINKED ISSUE #2996: Refactor selection range plugin
Previously, `hls-selection-range-plugin` converts `HieAST a` (from ghc) to `[SelectionRange]` (from lsp). This design is very tight-coupled. 

I'm planning to:
- Extract a new data type `CodeRange = CodeRange RealSrcSpan [CodeRange]`
- Split the process into two phases:
  1. Build `CodeRange` from `HieAST a` and comments (put this step in `Rules`)
  2. Derive `[SelectionRange]` from `CodeRange` (put this step in selection range lsp method handler)

I'll probably finish this in no more than two weeks. Then, "folding range" will simply become a use case of `CodeRange`. Adding it should be easier.

@sloorush Hope this will help your GSOC project (I'm assuming you have the same username on IRC and GitHub. Sorry if it's not the case). If you have any questions or advice, feel free to discuss them here or on the IRC channel.

kokobd:
This PR should close #2996 and #3001

In brief, `hls-selection-range-plugin` is changed to `hls-code-range-plugin`, and the code is rewritten so that adding
 "folding range" to this plugin becomes relatively easy.

@sloorush please check if the refactor satisfies your needs.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3003"></a>


plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -1,67 +1,66 @@
...
 
+data Log = LogCodeRange CodeRange.Log
+
+instance Pretty Log where
+    pretty log = case log of
+        LogCodeRange codeRangeLog -> pretty codeRangeLog
+
 selectionRangeHandler :: IdeState -> PluginId -> SelectionRangeParams -> LspM c (Either ResponseError (List SelectionRange))
 selectionRangeHandler ide _ SelectionRangeParams{..} = do
-    liftIO $ logDebug logger $ "requesting selection range for file: " <> T.pack (show uri)
```
michaelpj, line 64: did you decide this wasn't worth it or you just don't want to pass the recorder here?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -75,27 +74,45 @@ selectionRangeHandler ide _ SelectionRangeParams{..} = do
...
 
-    ast <- maybeToExceptT "fail to get ast for current file" . MaybeT . pure $
-        -- in GHC 9, the 'FastString' in 'HieASTs' is replaced by a newtype wrapper around 'LexicalFastString'
-        -- so we use 'coerce' to make it work in both GHC 8 and 9
-        getAsts hieAst Map.!? (coerce . mkFastString . fromNormalizedFilePath) file
-
-    let ast' = runReader (preProcessAST ast) (PreProcessEnv refMap)
-    let selectionRanges = findSelectionRangesByPositions (astPathsLeafToRoot ast') positions'
+    let selectionRanges = flip fmap positions' $ \pos ->
+            -- codeRange doesn't cover all portions of text in the file, so we need a default value
```
michaelpj, line 85: why? we're allowed to return an empty list of selection ranges in the response, wouldn't that be the appropriate response here?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -75,27 +74,45 @@ selectionRangeHandler ide _ SelectionRangeParams{..} = do
...
+            let defaultSelectionRange = SelectionRange (Range pos pos) Nothing
+             in reverseSelectionRange . fromMaybe defaultSelectionRange . findPosition' pos $ codeRange
 
     -- 'positionMapping' should be applied to the output ranges before returning them
     maybeToExceptT "fail to apply position mapping to output positions" . MaybeT . pure $
          traverse (toCurrentSelectionRange positionMapping) selectionRanges
 
+-- Find 'Position' in 'CodeRange'. Producing an inverse 'SelectionRange'
```
michaelpj, line 93:
```suggestion
-- | Find 'Position' in 'CodeRange'. Producing an inverse 'SelectionRange'
```

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
```diff
@@ -0,0 +1,109 @@
...
+        LogShake shakeLog                 -> pretty shakeLog
+        LogBadDependency badDependencyLog -> pretty badDependencyLog
+        LogNoAST                          -> "no HieAst exist for file"
+
+data BadDependencyLog = forall rule. Show rule => BadDependencyLog rule
+
+instance Pretty BadDependencyLog where
+    pretty (BadDependencyLog rule) = "can not get result from rule " <> pretty (show rule)
+
+data CodeRange = CodeRange Range [CodeRange]
```
michaelpj, line 54: haddock?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
```diff
@@ -0,0 +1,109 @@
...
+    deriving (Show, Generic)
+
+instance NFData CodeRange
```
michaelpj, line 57: can do `deriving anyclass NFData`

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
michaelpj: Maybe record syntax so we can have documented fields?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
```diff
@@ -0,0 +1,109 @@
...
+    define (cmapWithPrio LogShake recorder) $ \GetCodeRange file -> handleError recorder $ do
+        -- We need both 'HieAST' (for basic AST) and api annotations (for comments and some keywords).
+        -- See https://gitlab.haskell.org/ghc/ghc/-/wikis/api-annotations
+        HAR{hieAst, refMap} <- useExcept LogBadDependency GetHieAst file
+        ast <- maybeToExceptT LogNoAST . MaybeT . pure $
+            getAsts hieAst Map.!? (coerce . mkFastString . fromNormalizedFilePath) file
+        annPS <- useExcept LogBadDependency GetAnnotatedParsedSource file
+        pure $ buildCodeRange ast refMap annPS
+
+-- | Handle error in 'Action'. Returns an 'IdeResult' with no value and no diagnostics on error. (but writes log)
```
michaelpj, line 101: cc @drsooch

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
```diff
@@ -0,0 +1,109 @@
...
+-- | Like use, but report absense in 'ExceptT'
+useExcept :: IdeRule k v => (BadDependencyLog -> msg) -> k -> NormalizedFilePath -> ExceptT msg Action v
+useExcept f rule = maybeToExceptT (f (BadDependencyLog rule)) . MaybeT . use rule
+
+codeRangeRule :: Recorder (WithPriority Log) -> Rules ()
+codeRangeRule recorder =
+    define (cmapWithPrio LogShake recorder) $ \GetCodeRange file -> handleError recorder $ do
+        -- We need both 'HieAST' (for basic AST) and api annotations (for comments and some keywords).
+        -- See https://gitlab.haskell.org/ghc/ghc/-/wikis/api-annotations
+        HAR{hieAst, refMap} <- useExcept LogBadDependency GetHieAst file
```
michaelpj, line 95: This seems suspicious. I thought BadDependency was special, and we shouldn't be catching it like this?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
```diff
@@ -0,0 +1,109 @@
...
+    -- range don't need to care about 'HieAST'
+    -- TODO @sloorush actually use 'Annotated ParsedSource' to handle structures not in 'HieAST' properly (for example comments)
+    let ast' = runReader (preProcessAST ast) (PreProcessEnv refMap)
+    in simplify . astToCodeRange $ ast'
+
+astToCodeRange :: HieAST a -> CodeRange
+astToCodeRange (Node _ sp []) = CodeRange (realSrcSpanToRange sp) []
+astToCodeRange (Node _ sp children) = CodeRange (realSrcSpanToRange sp) (fmap astToCodeRange children)
+
+-- Remove redundant nodes in 'CodeRange' tree
```
michaelpj, line 71:
```suggestion
-- | Remove redundant nodes in 'CodeRange' tree
```

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
```diff
@@ -0,0 +1,109 @@
...
+simplify :: CodeRange -> CodeRange
+simplify r@(CodeRange range1 [CodeRange range2 children])
```
michaelpj, line 73: I would have assumed this function should be recursive? But it seems to only look at the very top level?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -75,27 +74,45 @@ selectionRangeHandler ide _ SelectionRangeParams{..} = do
...
+
+-- Reverse 'SelectionRange'. Just like 'reverse' for list.
+reverseSelectionRange :: SelectionRange -> SelectionRange
+reverseSelectionRange = go (SelectionRange invalidRange Nothing)
+  where
+    go :: SelectionRange -> SelectionRange -> SelectionRange
+    go acc (SelectionRange r Nothing) = SelectionRange r (checkRange acc)
+    go acc (SelectionRange r (Just parent)) = go (SelectionRange r (checkRange acc)) parent
+
+    checkRange :: SelectionRange -> Maybe SelectionRange
```
michaelpj, line 110: what's going on here? why does this happen?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -75,27 +74,45 @@ selectionRangeHandler ide _ SelectionRangeParams{..} = do
...
+            -- codeRange doesn't cover all portions of text in the file, so we need a default value
+            let defaultSelectionRange = SelectionRange (Range pos pos) Nothing
+             in reverseSelectionRange . fromMaybe defaultSelectionRange . findPosition' pos $ codeRange
 
     -- 'positionMapping' should be applied to the output ranges before returning them
     maybeToExceptT "fail to apply position mapping to output positions" . MaybeT . pure $
          traverse (toCurrentSelectionRange positionMapping) selectionRanges
 
+-- Find 'Position' in 'CodeRange'. Producing an inverse 'SelectionRange'
+findPosition' :: Position -> CodeRange -> Maybe SelectionRange
```
michaelpj, line 94: Could we return things in the right order from this function instead of needing `reverseSelectionRange`?

michaelpj:
cc @sloorush

michaelpj:
Oops, I started looking at this before I saw it was a draft, apologies if I made lots of comments that you were going to fix anyway :sweat_smile:

kokobd:
> Oops, I started looking at this before I saw it was a draft, apologies if I made lots of comments that you were going to fix anyway 😅

Thank you for your valuable suggestions! I'm a bit hurry, to make sure not to block sloorush for too long.

sloorush (commented):
Should we also rename the plugin to `hls-range-plugin` (as you suggested) in this PR?

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: I believe this information should be logged systematically for all handlers if needed. Currently, I do not need it anymore.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: The client can request multiple positions in a single request. (when there are multiple cursors). I don't want a single position breaks all of them.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: 1. Yes, maybe I should change comments to top-level functions to haddock.
2. `findPosition'` returns a top-down `SelectionRange`, while LSP expects a bottom-up one. I will check if it's possible to directly return a bottom-up one.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
kokobd: indeed

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
kokobd: Oh, I will look into that trick

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
kokobd: To clarify, `LogBadDependency` is just a log type constructor. It has no relation with `BadDependency` exception in  ghcide.

Do we prefer throwing an exception in this case? (indirectly through `use_` or `uses_`?)

I can't find too many usages of `use_`, `uses_` and `BadDependency` in plugins. It seems that most code just handles the `Maybe` manually, and returns an empty result on `Nothing`.

For example:
https://github.com/haskell/haskell-language-server/blob/510bd51e46fea8fb51ddfaa60bba505f0663497d/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs#L371-L375

https://github.com/haskell/haskell-language-server/blob/510bd51e46fea8fb51ddfaa60bba505f0663497d/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs#L399-L408

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
kokobd: Yep, the unit tests only catch the top-level case. I will fix it and add more tests.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
kokobd: This is basically trying to reverse a non-empty list (`SelectionRange` is basically a non-empty list). Unlike reversing an ordinary list, I can't find an appropriate initial value for `go`. Maybe I should use `Maybe`

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange.hs
```diff
@@ -75,27 +74,45 @@ selectionRangeHandler ide _ SelectionRangeParams{..} = do
...
+findPosition' pos (CodeRange range children) =
+    if positionInRange pos range
+    then Just $ case mapMaybe (findPosition' pos) children of
+            [childSelectionRange] -> SelectionRange range (Just childSelectionRange)
+            _                     -> SelectionRange range Nothing
```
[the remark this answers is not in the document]
kokobd: I will add some checks prior to this function, to make sure the whole `CodeRange` tree is valid.

michaelpj:
> Should we also rename the plugin to hls-range-plugin (as you suggested) in this PR?

Worth pointing out that we do publish the plugin package to Hackage. I think that's fine, though.

plugins/hls-selection-range-plugin/src/Ide/Plugin/SelectionRange/CodeRange.hs
michaelpj: Oh I see, I misunderstood. I think it's fine to log like you're doing here. I agree it's pretty inconsistent how people use these things (also of interest to @drsooch I think).

michaelpj: And maybe @July541

kokobd: In fact, I'm a bit unsure too. In the latest commit, I changed it to `use_`. Maybe we should discuss this, and:

1. If we decide to use exceptions, then we should advocate `use_` and `uses_`, and my previous ExceptT trick with a `BadDependency` log type is no longer necessary.
2. If we decide not to use exceptions, maybe we could mark `use_` and `uses_` as deprecated, and add a log type for `BadDependency` directly in `Shake.hs`.

kokobd:
Seems that this PR is also affected by https://github.com/haskell/haskell-language-server/issues/2997

kokobd:
Maybe this is ready for a review, while I'm adding unit tests to individual functions.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -0,0 +1,139 @@
...
+        | otherwise = do
+            let (left, right) = V.splitAt (V.length v `div` 2) v
+            startOfRight <- _start . _codeRange_range <$> V.headM right
+            if pos < startOfRight then binarySearchPos left else binarySearchPos right
+
+-- | Construct 'SelectionRange' from 'NonEmpty' 'Range'
+selectionRangeFromNonEmpty :: NonEmpty Range -> SelectionRange
+selectionRangeFromNonEmpty ranges
+    | (r, remaining) <- NonEmpty.uncons ranges =
+        SelectionRange r (fmap selectionRangeFromNonEmpty remaining)
```
michaelpj, line 130: I feel like this should be a fold of some kind, but not a big deal.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange/Rules.hs
```diff
@@ -0,0 +1,186 @@
...
+
+astToCodeRange :: HieAST a -> Writer [Log] CodeRange
+astToCodeRange (Node _ sp []) = pure $ CodeRange (realSrcSpanToRange sp) mempty CodeKindRegion
+astToCodeRange node@(Node _ sp children) = do
+    children' <- removeInterleaving . sort =<< traverse astToCodeRange children
+    let codeKind = if Just CustomNodeImportsGroup == isCustomNode node then CodeKindImports else CodeKindRegion
+    pure $ CodeRange (realSrcSpanToRange sp) (V.fromList children') codeKind
+
+-- | Remove interleaving of the list of 'CodeRange's.
+removeInterleaving :: [CodeRange] -> Writer [Log] [CodeRange]
```
michaelpj, line 117: I ran out of steam around here, but now this is so well factored, it might be nice to have a couple of unit tests for this function and `simplify`.

michaelpj (approved):

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange/Rules.hs
kokobd: Yes, I'm writing unit tests for every pure function like this

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -0,0 +1,139 @@
...
+findPosition pos root =
+    selectionRangeFromNonEmpty . NonEmpty.reverse -- SelectionRange requires a bottom-up order, so we need to reverse
+    <$> go root
+  where
+    -- Helper function for recursion. The range list is built top-down
+    go :: CodeRange -> Maybe (NonEmpty Range)
+    go node =
+        if positionInRange pos range
+        then case binarySearchPos children of
+            Just childContainingPos -> fmap (range NonEmpty.<|) (go childContainingPos)
```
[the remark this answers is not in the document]
kokobd: I fixed this. Many thanks!

kokobd:
I'm feeling good about this, merging.

PR #3008, opened 2022-07-01
Disable flaky test on Windows

### LINKED ISSUE #2997: Re-exported symbol hover ghcide test is flaky on Windows

michaelpj:
I don't think anyone is going to fix this soon, let's just disable it so
we can move on.

Fixes #2997.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3008"></a>

### DIFF
--- a/ghcide/test/exe/Main.hs
+++ b/ghcide/test/exe/Main.hs
@@ -4439,7 +4439,11 @@ findDefinitionAndHoverTests = let
   , test  no     yes    holeL65    hleInfo2      "hole with variable"
   , test  no     yes    cccL17     docLink       "Haddock html links"
   , testM yes    yes    imported   importedSig   "Imported symbol"
-  , testM yes    yes    reexported reexportedSig "Imported symbol (reexported)"
+  , if | isWindows ->
+        -- Flaky on Windows: https://github.com/haskell/haskell-language-server/issues/2997
+        testM no     yes    reexported reexportedSig "Imported symbol (reexported)"
+       | otherwise ->
+        testM yes    yes    reexported reexportedSig "Imported symbol (reexported)"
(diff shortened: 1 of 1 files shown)


ghcide/test/exe/Main.hs
```diff
@@ -4439,7 +4439,11 @@ findDefinitionAndHoverTests = let
   , test  no     yes    holeL65    hleInfo2      "hole with variable"
   , test  no     yes    cccL17     docLink       "Haddock html links"
   , testM yes    yes    imported   importedSig   "Imported symbol"
-  , testM yes    yes    reexported reexportedSig "Imported symbol (reexported)"
+  , if | isWindows ->
+        -- Flaky on Windows: https://github.com/haskell/haskell-language-server/issues/2997
+        testM broken yes    reexported reexportedSig "Imported symbol (reexported)"
```
July541, line 4444:
```suggestion
        testM no yes    reexported reexportedSig "Imported symbol (reexported)"
```

July541 (approved):
Should be `no` because it may succeed, we don't test it on Windows.

Otherwise, LGTM!

ghcide/test/exe/Main.hs
michaelpj: Ah, thanks!

PR #3013, opened 2022-07-03
Renaming of indirect references (RecordFieldPuns)

### LINKED ISSUE #2970: Rename plugin doesn't work with NamedFieldPuns
The environment doesn't matter here: just a plugin bug...

### Steps to reproduce
Rename the following `field` to something else using the rename plugin

```
newtype Foo = Foo { field :: Int }

unFoo :: Foo -> Int
unFoo Foo{field} = field
```

### Expected behaviour
I think the best UX should be that renames compose with bindings incurred by NamedFieldPuns

### Actual behaviour
The output is instead:

```
newtype Foo = Foo { bleh :: Int }

unFoo :: Foo -> Int
unFoo Foo{bleh} = field
```
i.e. the field reference in the binding site is renamed, but not the subsequent reference on the RHS of unFoo

OliverMadine:
### Issue
Consider this code snippet. While explaining this issue, I will refer to the occurrences of `field` on lines 3, 6, and 7 as A, B, and C respectively.

```haskell

newtype Foo = Foo { field :: Int }

unFoo :: Foo -> Int
unFoo Foo{field} 
    = field
```

The idea of a reference in HieDB is not transitive.
In this example, A references B and B references C but A does **not** reference C. This is because B is a punned name.

As a consequence, trying to rename A causes only A and B to be renamed, but we also want C to be renamed.

### Proposed Solution
Doing another pass of the references to find any punned names should be sufficient. 

I do **not** think that we can have doubly-indirect references, so we should only need to do one additional pass to find the transitive closure.

### Question
Are there other cases of having multiple names at the same position that I might have not properly considered?

Closes #2970

### DIFF
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
@@ -24,7 +24,7 @@ import           Data.Generics
 import           Data.Hashable
 import           Data.HashSet                          (HashSet)
 import qualified Data.HashSet                          as HS
-import           Data.List.Extra
+import           Data.List.Extra                       hiding (length)
 import qualified Data.Map                              as M
 import           Data.Maybe
 import           Data.Mod.Word
@@ -42,7 +42,6 @@ import           Development.IDE.GHC.ExactPrint
 import           Development.IDE.Spans.AtPoint
 import           Development.IDE.Types.Location
 import           HieDb.Query
-import           Ide.Plugin.Config
 import           Ide.Plugin.Properties
 import           Ide.PluginUtils
 import           Ide.Types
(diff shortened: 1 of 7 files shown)


michaelpj:
> The idea of a reference in HieDB is not transitive

Should we make it transitive? This might be easier at the db level, since it would just be a join of the table with itself...

> Are there other cases of having multiple names at the same position that I might have not properly considered?

The problem is that B is both a use of a name and a binding of a name. I can' think of any other instances of that...

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -65,16 +64,25 @@ descriptor pluginId = (defaultPluginDescriptor pluginId)
...
 renameProvider state pluginId (RenameParams (TextDocumentIdentifier uri) pos _prog newNameText) =
     pluginResponse $ do
-        nfp <- safeUriToNfp uri
-        oldName <- getNameAtPos state nfp pos
-        refLocs <- refsAtName state nfp oldName
+        nfp <- handleUriToNfp uri
+        directOldNames <- getNamesAtPos state nfp pos
+        directRefs <- concat <$> mapM (refsAtName state nfp) directOldNames
+
+        -- References of references to account for punned names (Indirect references)
```
michaelpj, line 71: You wrote a lovely explanation of this in the PR description that's going to probably get lost when you merge the PR, maybe move it into a longer comment in the code?

OliverMadine:
> > The idea of a reference in HieDB is not transitive
> 
> Should we make it transitive? This might be easier at the db level, since it would just be a join of the table with itself...

I think it is best not to do this in the DB. It seems like HieDB maintains the distinction between the name and the binding of the name. I do not think we want to join the table as it might be useful to keep this information.

In the example above, if i query for the references of `B`, HieDB gives 2 sets of references (one for the name and one for the binding). If we consider `B` as two distinct names, then references in this sense are transitive.

plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -21,6 +21,9 @@ tests = testGroup "Rename"
         rename doc (Position 0 15) "Op"
     , goldenWithRename "Exported function" "ExportedFunction" $ \doc ->
         rename doc (Position 2 1) "quux"
+    , ignoreForGhcVersions [GHC90, GHC92] "Record field refs broken for ghc > 9" $
+      goldenWithRename "Field Puns" "FieldPuns" $ \doc ->
+        rename doc (Position 7 13) "bleh"
```
pepeiborra, line 26: What is missing for GHC 9.x?

michaelpj: Write it down in the code so we remember! Even better, make a ticket and link to it.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -65,16 +64,27 @@ descriptor pluginId = (defaultPluginDescriptor pluginId)
...
     pluginResponse $ do
-        nfp <- safeUriToNfp uri
-        oldName <- getNameAtPos state nfp pos
-        refLocs <- refsAtName state nfp oldName
+        nfp <- handleUriToNfp uri
+        directOldNames <- getNamesAtPos state nfp pos
+        directRefs <- concat <$> mapM (refsAtName state nfp) directOldNames
+
+        {- References in HieDB are not necessarily transitive. With `NamedFieldPuns`, we can have
+           indirect references through punned names. To find the transitive closure, we do a pass of
```
michaelpj, line 72: "See test X for an example"?

michaelpj (approved):

plugins/hls-rename-plugin/test/Main.hs
OliverMadine: I made a [ticket](https://github.com/wz1000/HieDb/issues/45) and linked it in a comment.

PR #3015, opened 2022-07-04
re-enable haddock

kokobd:
Currently, we disable `-haddock` option in `cabal.project` because `ghc-lib` 9.2 doesn't build on GHC 8.10.
For now, we can remove the dependency on `ghc-lib` in `hls-hlint-plugin`, using `ghc-lib-parser` instead. (actually `hlint` itself doesn't depend on `ghc-lib`) Meanwhile, I've opened an issue in `ghc-lib`, see: https://github.com/digital-asset/ghc-lib/issues/391

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3015"></a>

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -38,8 +38,7 @@ optional-packages: vendored/*/*.cabal
 tests: true
 
 package *
-  -- ghc 8.10 cannot build ghc-lib 9.2 with --haddock
-  -- ghc-options: -haddock
+  ghc-options: -haddock
   test-show-details: direct
 
 write-ghc-environment-files: never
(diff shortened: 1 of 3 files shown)


kokobd:
> so it depends on `ghc-lib`

@michaelpj hlint depends on `ghc-lib-parser` actually, see: https://hackage.haskell.org/package/hlint-3.4/hlint.cabal

michaelpj:
So it does! Seems fine to me, but @pepeiborra might have thoughts since I know he did stuff in this area recently.

pepeiborra:
Looks great to me

pepeiborra (approved):

kokobd:
@mergifyio refresh

Issue #2732, opened 2022-02-22
OverloadedRecordDot support - Type Awareness / Completion


Dessix:
**Is your feature request related to a problem? Please describe.**

Given the following (GHC 9.2+) snippet:
```haskell
{-# LANGUAGE OverloadedRecordDot #-}
{-# LANGUAGE DuplicateRecordFields #-}
{-# LANGUAGE NoFieldSelectors #-}

data MyRecord = MyRecord
  { a :: String
  , b :: Integer
  } deriving (Eq, Show)

x :: MyRecord
x = MyRecord { a = "Hello", b = 12 }

y = x.a

z = x.b
```

HLS now understands (without record-dot-preprocessor!) that `y` is a `String` and `z` is an `Integer`, seen in both above-line hints and on hover.

Unfortunately, it has no type information available when hovering over the `a` in `x.a` or the `b` in `x.b`. Additionally, it lacks the ability to suggest- when entering a dot after a record type- the fields accessible on that type. 

**Describe the solution you'd like**

The ability to "dot into" types is both a small form of IDE-driven API exploration and a way to easily manage larger applications without needing a Haddock window on the side at all times.

Pressing '.' after a type (or manually opening auto-completion with the cursor directly following a dot following a type) should summon auto-completion with the context narrowed to the prior type's fields.

For a step further, but one worth considering when designing a solution, a useful addition would be allowing automatic import of the type information required if necessary, similar to when accepting a completion of an unimported `trait` function in [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/). Documentation on the HasField requirements is available [here (HasField constraints)](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/hasfield.html#solving-hasfield-constraints) and [here (OverloadedRecordDot)](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_record_dot.html). `NoFieldSelectors` may complicate this functionality.

**Describe alternatives you've considered**

I've built my glue-code in Rust instead, so far, to avoid having to manage large numbers of one-off record types in Haskell, but I'd love to unify my codebase, and Record Dot Syntax has been the selling point that would get me there, assuming the presence of IDE support. It appears that 9.2.1 support is now functioning, but dot-accesses appear opaque to the language server.

michaelpj:
[...]

PR #3016, opened 2022-07-04
Record Dot Hover Types

coltenwebb:
Working towards #2732 (reopened since I closed and force-pushed my last PR, which github dislikes)

This PR patches the way HieAST is generated by GHC. For performance reasons, not all nodes in the AST get expanded and checked for types. This caused the nodes corresponding to record-dot-syntax to not be included in the HieAST. This patch adds them back.

## Notes

This is what type info on hover looks like:

Code_ps0jgEHy0v

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3016"></a>


kokobd:
Do we have to introduce the large file with 2000+ lines? It looks similar to `GHC/Iface/Ext/Ast.hs`, could you explain it a bit?

coltenwebb:
You're right, the `HieAst.hs` in src-ghc92 is the same one in 9.2.x GHC except I inserted the line 789, to ensure the record dot syntax gets desugared. I agree that adding so much to the codebase is less than ideal. Maybe this explanation belongs in a banner comment at the top of the file? I'll also try to get this patch directly into GHC, but it may be better to keep this here for backwards compatibility with 9.2.{1,2}

michaelpj:
I think that's the nature of the `hie-compat` packages, they mostly consist of wholesale copies of files from GHC with small changes. I'm not sure if we can easily improve that, but comments are _definitely_ a good idea.

wz1000:
We would want a patch that is accepted for GHC HEAD before we can merge such changes to hie-compat. We have followed this policy for all the previous patches we have in `hie-compat`.

coltenwebb:
Should be good to go. I've added the comments to HieAst.hs, and the patch is merged into GHC master.

wz1000:
Could you add the patch to the readme of hie-compat?

wz1000 (approved):
Thanks, looks all good now.

michaelpj:
Just needs a test, I think?

coltenwebb:
I added three tests, there are additional tests for the record dot syntax types on the GHC side too.

coltenwebb:
✨

michaelpj:
Sweet!

ghcide/test/exe/Main.hs
```diff
@@ -4352,16 +4352,25 @@ findDefinitionAndHoverTests = let
...
-  typeDefinitionTests = [ tst (getTypeDefinitions, checkDefs) aaaL14 (pure tcData) "Saturated data con"
-                        , tst (getTypeDefinitions, checkDefs) aL20 (pure [ExpectNoDefinitions]) "Polymorphic variable"]
+    , testGroup "type-definition" typeDefinitionTests
+    , testGroup "hover-record-dot-syntax" recordDotSyntaxTests ]
+
+  typeDefinitionTests = [ tst (getTypeDefinitions, checkDefs) aaaL14 sourceFilePath (pure tcData) "Saturated data con"
+                        , tst (getTypeDefinitions, checkDefs) aL20 sourceFilePath (pure [ExpectNoDefinitions]) "Polymorphic variable"]
+  
+  recordDotSyntaxTests 
+    | ghcVersion == GHC92 =
```
michaelpj, line 4362: Ah, too late: should this check be more generous? We expect these tests to work on later versions also, right?

coltenwebb: Oh that's correct. I'll make a PR for that

Issue #2057, opened 2021-07-31
Code action order could prefer fixes over disabling warnings


andys8:
## TL;DR

Code actions are ordered in a way where the most likely option is on the bottom.

## Issue

I'm often using code actions to organize imports and remove the ones that aren't necessary. The option that is use the most often ("Remove all redundant imports") is the last option of the menu. And I think most of the time ["Disable ... warnings"](https://github.com/haskell/haskell-language-server/blob/ced934e66c8011d61ea934f16793247b8ab27132/plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs#L82-L86) is the first option. "Make all imports implicit" seems not relevant here.

One issue is that one needs to move to the end of the list each time. But for example in CoC (Vim) `coc-fix-current` defines the top code action as default and will execute it. So changing the order could be an improvement.

![image](https://user-images.githubusercontent.com/13085980/127753553-799d88cb-3af6-4b73-9550-7a26a2cd8311.png)

## Potential Improvement

Change the order of code actions to (try to) match the most likely behavior. E.g. move "disable warnings" to the bottom of the suggested code actions by default.

## Probe Tools

```
haskell-language-server version: 1.3.0.0 (GHC: 8.6.5) (PATH: /home/andreas/.local/bin/haskell-language-server-wrapper) (GIT hash: e7c5e90b6df5dff2760d76169eddaea3bdd6a831)
Tool versions found on the $PATH
cabal:          3.2.0.0
stack:          2.7.3
ghc:            8.10.4
```

### Config

https://github.com/andys8/dotfiles/blob/3c1d32349ffaf4c1ab840f7b340746846fb5e522/coc-settings.json#L34-L55

July541:
Hello, I can only get this hint, am I missing something?

![image](https://user-images.githubusercontent.com/16057632/127840633-fb111026-d77d-428b-a942-5656c5ecdf36.png)

jneira:
Mmm i think in vim all the code actions are shown, not only the associated with the diagnostic, in vscode you could see them clicking in the light bulb at the start of the expression

July541:
@jneira Seems no, all I can see while clicking the bulb is just the one hint.

July541:
It turns out I don't know when my hls doesn't work as before, I remember it can suggest removing unused import.

![image](https://user-images.githubusercontent.com/16057632/127842151-38a77262-30b2-4178-a5b6-9eaff75c4eb4.png)

[...]

PR #3018, opened 2022-07-05
Hlint: CodeAction with isPreferred

andys8:
> A quick fix should be marked preferred if it properly addresses the underlying error. A refactoring should be marked preferred if it is the most reasonable choice of actions to take.

For hlint it is the likeliest choice to apply the fix. Ignore the rule
for the complete module is preferred less.

Related to https://github.com/haskell/haskell-language-server/issues/2057

## Demo

The client - in my case Vim/Coc - can respect `isPreferred`. In case of CoC it will be reflected in the sort order of code actions.

### Before

screenshot-1657034414

### After

screenshot-1657034612

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3018"></a>

### DIFF
--- a/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
+++ b/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
@@ -450,19 +450,19 @@ diagnosticToCodeActions dynFlags fileContents pluginId documentId diagnostic
            , let applyHintTitle = "Apply hint \"" <> hint <> "\""
                  applyHintArguments = [toJSON (AOP (documentId ^. LSP.uri) start hint)]
                  applyHintCommand = mkLspCommand pluginId "applyOne" applyHintTitle (Just applyHintArguments) ->
-               Just (mkCodeAction applyHintTitle diagnostic Nothing (Just applyHintCommand))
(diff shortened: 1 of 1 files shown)


michaelpj:
Please include the rationale for why we're making this decision as a comment in the code. 

I think your argument is good, and probably extends to other code actions that are produced directly from diagnostics. The "add pragma" code action seems like another good candidate.

I'm not totally sure what our rubric should be, though. Arguably ignoring the hint also clearly deals with the diagnostic, maybe both should be preferred? I don't know. I think it's reasonable to assume that if people have the hint enabled globally they probably want to fix it rather than ignore it.

andys8:
> Please include the rationale for why we're making this decision as a comment in the code.

Sure thing [2a9a687](https://github.com/haskell/haskell-language-server/pull/3018/commits/2a9a6872b2a5982e4d15f2690b26b4f047ed7860)

> I think it's reasonable to assume that if people have the hint enabled globally they probably want to fix it rather than ignore it.

That would also be my main argument. I know the project needs to work for different projects and developers, but the main influence of `isPreferred` seems to express the "default" quick fix which in case of an installed and configured `hlint` would be to apply the quick fixes, rather than ignoring them all the time. My assumption is since where talking mostly about sort order this can save a bunch (millions 😄) of keystokes.

michaelpj (approved):

PR #3022, opened 2022-07-06
hls-plugin-api: add lower bounds

Bodigrim:
Before `opentelemetry-0.4` `SpanInFlight` is not exported.
Before `dependent-sum-0.7` `GCompare` is not coming from `some`, but is defined elsewhere, which leads to incompatibility with `lsp`.

As a Hackage trustee, I made a matching revision:
https://hackage.haskell.org/package/hls-plugin-api-1.4.0.0/revisions/

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3022"></a>

### DIFF
--- a/hls-plugin-api/hls-plugin-api.cabal
+++ b/hls-plugin-api/hls-plugin-api.cabal
@@ -39,7 +39,7 @@ library
     , containers
     , data-default
     , dependent-map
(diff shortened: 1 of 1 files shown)


michaelpj (approved):
Thank you!

Issue #2967, opened 2022-06-18
pre-commit hook is broken in nix


kokobd:
Currently, `nix develop` complains:
```
nix-pre-commit-hooks: updating /home/kokobd/work/github.com/kokobd/haskell-language-server repo
nix-pre-commit-hooks: WARNING: Refusing to install because of pre-existing .pre-commit-config.yaml
    1. Translate .pre-commit-config.yaml contents to the new syntax in your Nix file
        see https://github.com/hercules-ci/nix-pre-commit-hooks#getting-started
    2. remove .pre-commit-config.yaml
    3. add .pre-commit-config.yaml to .gitignore
```
This problem was introduced by #2679, where `.pre-commit-config.yaml` was checked in git.
But according to our documentation, `.pre-commit-config.yaml` should be managed by Nix. People do not use Nix can just manually install `pre-commit` hook and paste the file content into `.pre-commit-config.yaml`
https://github.com/haskell/haskell-language-server/blob/09968a156738d8bc196ad280e474c13bd36555c6/docs/contributing/contributing.md?plain=1#L171-L216

Now we have two options to compare:
1. Remove `pre-commit` config from Nix, and update the docs accordingly.
    - pros: Non-Nix users' life is made easier a little bit.
    - cons: Nix users will have to install pre-commit hooks manually.
2. Remove `.pre-commit-config.yaml` from Git, and let Nix manage it again.
    - pros: Nix users can benefit from the pre-commit hooks without any extra setup.
    - cons: Non-Nix users have to do one more manual step: paste the `.pre-commit-config.yaml`

I recommend the second option, as Non-Nix users will have some manual setups anyway, and we were on option 2 before #2679 was merged.

drsooch:
To be fair this has been merged for 5 months. No one has raised any objections or issues since. It would seem like nix users are in the minority. 

The original intention was to make it so users didn't have to copy the json from documentation. 

Is there a third option that can satisfy both parties?

drsooch:
I'm assuming the `nix` file of interest is `flake.nix` it looks like the pre-commit hook is out of date anyways. Not that it matters for nix users but we've added a `\r\n` to `\n` converter in the pre-commit hook as well. Seems odd nix can't handle a predefined pre commit hook configuration.

kokobd:
I noticed the last change to `flake.nix` was made by @michaelpj last month, so I assume you are using Nix. Would you mind providing some advice on this?

michaelpj:
[...]

PR #3024, opened 2022-07-08
remove all usages of pre-commit-check in nix

kokobd:
In #2967, @teto points out `nix flake show` is broken due to some usages of `pre-commit-check` were not removed. As `pre-commit-check` is already removed from flake input, I believe we can remove the remaining usages of it safely.

However, `nix flake show` and `nix flake check` still fail due to our implicit use of IFD(import from derivation), see: https://github.com/NixOS/nix/issues/4265. With IFD, these commands are trying to build all systems listed in `flake-utils.lib.eachSystem`.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3024"></a>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -374,7 +374,6 @@
 
         devShells = simpleDevShells // nixDevShells // {
           default = simpleDevShells.haskell-language-server-dev;
-          inherit (self.checks.${system}.pre-commit-check) shellHook;
         };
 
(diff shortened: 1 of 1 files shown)


teto:
that fixes it for me. @guibou

michaelpj (approved):

PR #3025, opened 2022-07-08
ghcide: lower bounds

Bodigrim:
Before `ghc-exactprint-1.4` some `instance Default` are missing.
`UnliftIO.Directory` appears only in `unliftio-0.2.6`.
`tasty-hunit < 0.10` does not re-export `HasCallStack`.


Bodigrim:
I made a matching revision:
https://hackage.haskell.org/package/ghcide-1.7.0.0/revisions/

PR #3029, opened 2022-07-09
hls-eval-plugin: add lower bound for parser-combinators

Bodigrim:
`manyTill_` was introduced only in `parser-combinators-1.2`.

Matching revision:
https://hackage.haskell.org/package/hls-eval-plugin-1.2.2.0/revisions/

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3029"></a>


pepeiborra (approved):
Thanks!

PR #3035, opened 2022-07-11
handle trailing comma in import list properly

### LINKED ISSUE #2958: Adding imports adds an extra comma
It is legal in Haskell to write an import list like this:

``` haskell
import Foo
  ( a,
    b,
    c,
  )
```

Note the trailing comma.  In particular, Ormolu formats files this way.  However, if I have an import list formatted this way, and ask HLS to add an import (for example, from the quick-actions list, or otherwise auto-importing something), I get this

``` haskell
import Foo
  ( a,
    b,
    c,, newImport
  )
```

Note the double comma at the end.  Of course, this is a syntax error.

kokobd:
fix #2958 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3035"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
+++ b/ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
@@ -32,8 +32,7 @@ import           Data.Data                             (Data)
 import           Data.Functor
 import           Data.Generics                         (listify)
 import qualified Data.Map.Strict                       as Map
-import           Data.Maybe                            (fromJust, isNothing,
-                                                        mapMaybe)
+import           Data.Maybe                            (fromJust, isNothing, mapMaybe, fromMaybe )
 import qualified Data.Text                             as T
 import           Development.IDE.GHC.Compat hiding (Annotation)
 import           Development.IDE.GHC.Error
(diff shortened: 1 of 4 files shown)


kokobd:
Strange. I can not reproduce this test failure locally. Both running tests and manually trying this feature give successful results.
```
extend multi line import with trailing comma:                                                     FAIL (1.67s)
          test/exe/Main.hs:[208](https://github.com/haskell/haskell-language-server/runs/7295318339?check_suite_focus=true#step:8:209)3:
          expected: "module ModuleB where\nimport ModuleA (stuffB, stuffA,\n               )\nmain = print (stuffA, stuffB)\n"
           but got: "module ModuleB where\nimport ModuleA (stuffB, stuffA\n               )\nmain = print (stuffA, stuffB)\n"
          Use -p '/without checkAll.extend multi line import with trailing comma/' to rerun this test only.
```

kokobd:
> I can not reproduce this test failure locally.

Okay, that's because ghc-exactprint behaves differently in GHC 9.0 (not 9.2!) and GHC 8.10.7, where the API is the same. But in GHC 9.0, an extra comma is attached to the "list", besides the one attached to the "list item". According to the docs, that behavior might be wrong.

Anyway, I finally got all compatibility issues handled.

georgefst:
Works for me on GHC 9.2.3, thanks! (but I won't approve this because I haven't looked at the code)

kokobd:
> Works for me on GHC 9.2.3

@georgefst Thanks for your testing!

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
```diff
@@ -367,17 +374,28 @@ extendImportTopLevel thing (L l it@ImportDecl{..})
...
-        when hasSibling $
-          addTrailingCommaT (last lies)
+        anns <- getAnnsT
+        maybe (pure ()) addTrailingCommaT (lastMaybe lies)
         addSimpleAnnT x (DP (0, if hasSibling then 1 else 0)) []
         addSimpleAnnT rdr dp00 [(G AnnVal, dp00)]
+
+        -- When the last item already has a trailing comma, we append a trailing comma to the new item.
+        let isAnnComma (G AnnComma, _) = True
+            isAnnComma _                  = False
```
fendor, line 384: nitpick: can we align this or remove the superfuous space?

fendor (approved):
LGTM!

ghcide/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
kokobd: Aha, my mistake. btw, stylish-haskell refuses to run on this file, because there are too many CPP, making it fail.

PR #3058, opened 2022-07-23
Feat: Folding Ranges

sloorush:
This PR adds the feature of Folding Ranges to the Range Plugin.

References:
- Folding ranges from haskell/lsp (https://github.com/haskell/lsp/blob/5422dd13f0362917e5981e07d60617ca6e233833/lsp-types/src/Language/LSP/Types/FoldingRange.hs)
- Folding ranges from LSP spec (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3058"></a>


sloorush:
@kokobd I have started to implement Folding Ranges. 

Is there any reference/documentation for Code Range that I can use to understand it better?

sloorush:
cc: @michaelpj

kokobd:
> Is there any reference/documentation for Code Range

@sloorush "code range" is just a concept I come up with to unify selection range and folding range. The motivation is to split source code into "ranges". Please just read the code. And by design, you should be able to traverse the `CodeRange` tree, producing `FoldingRange`s along the way.

Moreover, feel free to add more fields to `CodeRange` as you need (if you need to preserve some extra information from HieAST). Maybe you have noticed, I already added `CodeRangeKind`, which corresponds to `FoldingRangeKind`

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -153,24 +148,41 @@ findPosition pos root = go Nothing root
...
+createFoldingRange :: CodeRange -> Maybe FoldingRange
+createFoldingRange node1 = do
+    let range = _codeRange_range node1
+    -- let children = _codeRange_children node1
+    let Range startPos endPos = range
+    let Position lineStart _= startPos
+    let Position lineEnd _ = endPos
+    let codeRangeKind = _codeRange_kind node1
+
+    if codeRangeKind == CodeKindComment
```
kokobd, line 179: Tip: You can use pattern matching instead of `==`

hls-plugin-api/src/Ide/Plugin/Config.hs
```diff
@@ -156,6 +159,7 @@ instance A.FromJSON PluginConfig where
       <*> o .:? "completionOn"     .!= plcCompletionOn  def
       <*> o .:? "renameOn"         .!= plcRenameOn      def
       <*> o .:? "selectionRangeOn" .!= plcSelectionRangeOn def
+      <*> o .:? "selectionRangeOn" .!= plcFoldingRangeOn def
```
kokobd, line 162: did you mean `"foldingRangeOn"`?

kokobd:
It would be nice to post a screencast here and add some tests to `hls-code-range-plugin/test`

michaelpj:
Yeah, this definitely needs some tests.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj:
```
case codeRangeKind of
   CodeKindComment -> ...
   CodeKindImports -> ...
   ...
```

michaelpj: I'd also be tempted to pull out the construction of the range, and just have this return the kind, i.e. 
```
let frk <- case codeRangeKind of
    CodeKindComment -> Just FoldingRangeComment
    ...
    _ -> Nothing
case frk of
    Just _ -> Just (FoldingRange lineStart Nothing lineEnd Nothing frk)
    Nothing -> Nothing
```

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +148,42 @@ findPosition pos root = go Nothing root
...
+        then Just (FoldingRange lineStart Nothing lineEnd Nothing (Just FoldingRangeComment))
```
michaelpj, line 180: I'm pretty sure you do want to include the start and end characters here, you should have them.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +148,42 @@ findPosition pos root = go Nothing root
             startOfRight <- _start . _codeRange_range <$> V.headM right
             if pos < startOfRight then binarySearchPos left else binarySearchPos right
 
+findFoldingRanges :: CodeRange -> [Maybe FoldingRange]
```
michaelpj, line 151: These functions share a lot of code with the functions for creating selection ranges. I wonder if we can share some of it.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Start and end characters are optional, and it takes the start and end character of the line by default.

I wanted to try if it makes a difference. I'll check it.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: I made a separate function that converts crk to frk, the function looks a lot cleaner now.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -470,6 +477,9 @@ instance PluginRequestMethod TextDocumentPrepareCallHierarchy where
 instance PluginRequestMethod TextDocumentSelectionRange where
   combineResponses _ _ _ _ (x :| _) = x
 
+instance PluginRequestMethod TextDocumentFoldingRange where
+  combineResponses _ _ _ _ (x :| _) = x
```
kokobd, line 481: Should we instead concatenate the result, instead of always taking the first one?

michaelpj: Yes, definitely.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -1,13 +1,15 @@
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards   #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
-module Ide.Plugin.CodeRange (
-    descriptor
```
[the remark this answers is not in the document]
sloorush: The tab sizing was at 2 instead of 4, fixed it.

hls-plugin-api/src/Ide/Types.hs
sloorush: Yes! Makes sense

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -42,40 +44,67 @@ import           Ide.Types                            (PluginDescriptor (pluginH
...
+          <> mkPluginHandler STextDocumentFoldingRange foldingRangeHandler,
+      pluginRules = codeRangeRule (cmapWithPrio LogRules recorder)
     }
 
 data Log = LogRules Rules.Log
 
 instance Pretty Log where
-    pretty log = case log of
-        LogRules codeRangeLog -> pretty codeRangeLog
+  pretty log = case log of
```
kokobd, line 73: Please revert all indentation changes. Basically, all code is modified, making code review harder.

sloorush: I have removed all the formatting, will format it after the code is reviewed

sloorush:
> It would be nice to post a screencast here and add some tests to `hls-code-range-plugin/test`

Here is a screen recording of the feature.
folding-ranges

kokobd:
You are free to run `stylish-haskell` at any time. In fact, if you have followed [Style Guidelines](https://haskell-language-server.readthedocs.io/en/latest/contributing/contributing.html#style-guidelines) carefully, you should have installed a pre-commit hook, which runs `stylish-haskell` automatically for you.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: I think both of them now work differently. Can something be done now to share code between the two?

michaelpj: Yeah, now they're completely different!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +148,31 @@ findPosition pos root = go Nothing root
...
+   in frRoot ++ frChildren
+
+-- | Parses code range to folding range
+createFoldingRange :: CodeRange -> Maybe FoldingRange
+createFoldingRange node1 = do
+    let range = _codeRange_range node1
+    let Range startPos endPos = range
+    let Position lineStart _= startPos
+    let Position lineEnd _ = endPos
+    let codeRangeKind = _codeRange_kind node1
```
michaelpj, line 168:
```suggestion
createFoldingRange (CodeRange (Range (Position lineStart _) (Position lineEnd _) _ codeRangeKind)) = do
```

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +148,31 @@ findPosition pos root = go Nothing root
...
+
+    let frk = crkToFrk codeRangeKind
+
+    case frk of
+        Just _  -> Just (FoldingRange lineStart Nothing lineEnd Nothing frk)
```
michaelpj, line 173: I still think we should use the characters. Why not? We have them!

sloorush: Makes sense. There is nothing to lose with characters; I've added them.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Looks a lot cleaner this way. I've changed it.

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
```diff
@@ -50,5 +50,16 @@ testTree =
...
+        testGroup "findFoldingRanges" $
+            let check :: CodeRange -> [FoldingRange] -> Assertion
+                check codeRange = (findFoldingRanges codeRange @?=)
+
+                mkCodeRange :: Position -> Position -> V.Vector CodeRange -> CodeRange
+                mkCodeRange start end children = CodeRange (Range start end) children CodeKindRegion
+            in [
+                testCase "General" $ check
+                (mkCodeRange (Position 1 1) (Position 5 10) [])
+                [FoldingRange 1 (Just 1) 5 (Just 10) (Just FoldingRangeRegion)]
```
sloorush, line 63: @michaelpj @kokobd I was wondering what all test cases will be necessary to test folding ranges, or should I just add some general test cases of situations that happen(like switch cases, if else etc.)?

kokobd: Please add some test cases in `test/Main.hs` too. That file is for integration tests.

`Ide/Plugin/CodeRangeTest` contains unit tests. Your tests for `findFoldingRanges` is good ;) Maybe you could add a  test case for `createFoldingRange` too.

sloorush:
@kokobd, I have added the integration test but I'm not sure if this is the correct behaviour.

So the test case(`plugins/hls-code-range-plugin/test/testdata/folding-range/Function.golden.txt`) I have added is what the test is generating assume that it is correct, but logically thinking, I'm unsure as to why it is generating so many folding ranges for such a small function(`plugins/hls-code-range-plugin/test/testdata/folding-range/Function.hs`) which we can see in the screenshot has only 2 foldings. 

PTAL. Is it something that I am doing incorrectly?

image

kokobd:
> why it is generating so many folding ranges for such a small function(`plugins/hls-code-range-plugin/test/testdata/folding-range/Function.hs`) which we can see in the screenshot has only 2 foldings.

@sloorush It means that VSCode has some deduplication logic built in. You should implement the deduplication in your code, only preserving the code ranges you want in the folding range handler.

sloorush:
Amm @kokobd, But I am not able to understand the logic of how these ranges are being generated. Like it seems very random, from the start of a line to the middle of a function definition.

kokobd:
> Amm @kokobd, But I am not able to understand the logic of how these ranges are being generated. Like it seems very random, from the start of a line to the middle of a function definition.

@sloorush It corresponds to the AST. A single line of code doesn't mean a single node in AST. For example:

```
((4, 0) : (4, 25)) : isEven :: Integer -> Bool
((4, 0) : (4, 6)) : isEven
((4, 10) : (4, 25)) : Integer -> Bool
((4, 10) : (4, 17)) : Integer
((4, 21) : (4, 25)) : Bool
```

sloorush:
> @sloorush It means that VSCode has some deduplication logic built in. You should implement the deduplication in your code, only preserving the code ranges you want in the folding range handler.

@kokobd, I have fixed it:
- removed all folding ranges in which startline and endline were the same.
- filtered final result to have only the largest folding range that starts from one startline

kokobd:
I'll pull and try it out locally. You may merge origin/master into your branch first, as some changes to `hls-code-range-plugin` have been made on master.

sloorush:
Okay! Also should I squash my commits? or will it be done while merging

kokobd:
> Okay! Also should I squash my commits? or will it be done while merging

Mergify will squash and merge for us.

kokobd:
What happened? There are still merge conflicts, and it fails to build.

sloorush:
> What happened? There are still merge conflicts, and it fails to build. And almost everything is changed...
> 
> image

Oops, I messed up the merge, fixed it. Sorry for the request spam.

kokobd (changes requested):
I tried it out locally, and the overall functionality was fine ;) Just some more suggestions

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +150,34 @@ findPosition pos root = go Nothing root
...
+    else case frk of
+        Just _  -> Just (FoldingRange lineStart (Just charStart) lineEnd (Just charEnd) frk)
+        Nothing -> Nothing
+
+-- Removes all small foldings that start from the same line
+removeDupStartLineFoldings :: [FoldingRange] -> [FoldingRange]
+removeDupStartLineFoldings [] = []
+removeDupStartLineFoldings [x] = [x]
+removeDupStartLineFoldings (frx@(FoldingRange x _ _ _ _):xs@((FoldingRange y _ _ _ _):xs2))
+    | x == y = removeDupStartLineFoldings ([frx]++xs2)
```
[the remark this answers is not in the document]
sloorush: 1. Yes, the list is always sorted by having the biggest coderange first, but this is just my observation.
2. I've changed it, thanks!

kokobd (approved):

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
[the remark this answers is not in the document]
kokobd: Approved. As this is part of your GSOC project, let's wait for another approval from @michaelpj

docs/features.md
```diff
@@ -354,15 +365,15 @@ Known limitations:
...
-| Document color         | Unclear if useful                                                                        | `textDocument/documentColor`                        |
-| Color presentation     | Unclear if useful                                                                        | `textDocument/colorPresentation`                    |
-| Monikers               | Unclear if useful                                                                        | `textDocument/moniker`                              |
+| Feature                | Status            | [LSP method](./what-is-hls.md#lsp-terminology) |
+| ---------------------- | ----------------- | ---------------------------------------------- |
+| Signature help         | Unimplemented     | `textDocument/signatureHelp`                   |
+| Jump to declaration    | Unclear if useful | `textDocument/declaration`                     |
+| Jump to implementation | Unclear if useful | `textDocument/implementation`                  |
+| Folding                | Unimplemented     | `textDocument/foldingRange`                    |
```
michaelpj, line 373: you implemented it!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -68,6 +72,26 @@ instance Pretty Log where
...
+        uri :: Uri
+        TextDocumentIdentifier uri = _textDocument
+
+getFoldingRanges :: NormalizedFilePath -> ExceptT String IdeAction [FoldingRange]
+getFoldingRanges file = do
+    (codeRange, _) <- maybeToExceptT "fail to get code range" $ useE GetCodeRange file
+
+    let foldingRanges = findFoldingRanges codeRange
+
+    maybeToExceptT "Fail to generate folding range" (MaybeT . pure $ Just foldingRanges)
```
michaelpj, line 93:
```suggestion
    pure $ findFoldingRanges codeRange
```

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +150,34 @@ findPosition pos root = go Nothing root
...
 
+-- | Traverses through the code range and children to a folding ranges
+findFoldingRanges :: CodeRange -> [FoldingRange]
+findFoldingRanges r@(CodeRange _ children _) =
+  let frRoot :: [FoldingRange] = case createFoldingRange r of
+        Just x  -> [x]
+        Nothing -> []
+
+      frChildren :: [FoldingRange] = concat $ V.toList $ fmap findFoldingRanges children
+   in frRoot ++ frChildren
```
michaelpj, line 161: Worth a note that this isn't risky because `frRoot` has  length 0 or 1, otherwise it could be quadratic. Have to be careful with `++`. You could make this totally unambiguous by pushing this inside the `frRoot` bit, i.e.
```
let frChildren = ...
in case createFoldingRange r of
  Just x -> x:frChildren
  Nothing -> frChildren
```
not sure if that's better

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +150,34 @@ findPosition pos root = go Nothing root
...
+
+-- | Parses code range to folding range
+createFoldingRange :: CodeRange -> Maybe FoldingRange
+createFoldingRange (CodeRange (Range (Position lineStart charStart) (Position lineEnd charEnd)) _ ck) = do
+    let frk = crkToFrk ck
+    if lineStart == lineEnd
```
michaelpj, line 167: why? this sort of thing _really_ needs a comment, ideally with an example showing why it's problematic. You spent lots of time figuring this out, and all that knowledge is going to be lost unless you write it down!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +150,34 @@ findPosition pos root = go Nothing root
...
+        then Nothing
+    else case frk of
+        Just _  -> Just (FoldingRange lineStart (Just charStart) lineEnd (Just charEnd) frk)
+        Nothing -> Nothing
+
+-- Removes all small foldings that start from the same line
```
michaelpj, line 173: Why is this good? Give an example!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +150,34 @@ findPosition pos root = go Nothing root
...
+removeDupStartLineFoldings :: [FoldingRange] -> [FoldingRange]
+removeDupStartLineFoldings [] = []
+removeDupStartLineFoldings [x] = [x]
+removeDupStartLineFoldings (frx@(FoldingRange x _ _ _ _):xs@((FoldingRange y _ _ _ _):xs2))
```
michaelpj, line 177: This isn't a total pattern match, I think? Does GHC not give you a warning here? The pattern is also a bit hard to understand, I'd split it up, maybe just match on `(x:xs)` and then do the rest in let-bindings.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: And it removes _everything_ that starts on the same line? Why is that good?

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
```diff
@@ -50,5 +50,66 @@ testTree =
                         ( SelectionRange (Range (Position 1 1) (Position 5 10)) Nothing
                         )
                     )
-            ]
+            ],
+
+        testGroup "findFoldingRanges" $
```
michaelpj, line 55: This could also be in its own module: this module is `CodeRangeTest`, perhaps create `FoldingRangeTest`. Not super important though.

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
```diff
@@ -50,5 +50,66 @@ testTree =
...
+            mkCodeRange start end children crk = CodeRange (Range start end) children crk
+        in [
+            -- General test
+            testCase "Test General Code Block" $ check
+                (mkCodeRange (Position 1 1) (Position 5 10) [] CodeKindRegion)
+                (Just (FoldingRange 1 (Just 1) 5 (Just 10) (Just FoldingRangeRegion))),
+            -- General test
+            testCase "Test Same Start Line" $ check
+                (mkCodeRange (Position 1 1) (Position 1 10) [] CodeKindRegion)
+                Nothing
```
michaelpj, line 113: Why is this Nothing? I think this would be answered by more explanation about what you're doing with code ranges on the same line

michaelpj (changes requested):
Looking good!

Do we have tests for all the weird cases we observed during testing? I remember there was a case where if there was a gap between the type signature and the body then things went wrong. It's good to have tests for things that go wrong also, you can leave a TODO on them explaining what's needs to be fixed.

michaelpj:
Also it would probably be good to have some more substantial Haskell examples in the tests!

kokobd:
@michaelpj I always appreciate your unlimited energy on code reviewing! Many thanks!

michaelpj:
I'm just sorry I didn't get to it earlier! Badly timed holiday!

docs/features.md
sloorush: Should I just remove folding from here? Or should I mention implemented?

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
sloorush: So, Something in one line does not need folding like what will it exactly fold. So I am removing all the folding ranges that start and end on the same line. (This logic is also handled by vscode itself, double proofing it by removing them).

I'll add an explanation about this in the comments.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Okay, This makes sense, I had not thought about this

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: This is a total pattern match. 

I'll document how it works in the comments, I guess that should help, mainly understanding the destructuring should help. 

Also, I feel I should rename the x and y to something more self-explanatory. Should I change it?

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
sloorush: Module CodeRangeTest only had tests about selection range. 

So I thought of it as testing CodeRange implementation on selection and folding ranges.

Should I keep it the same or change it?

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: PTAL, if the explanation in the comments is adequate or not?

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: PTAL, if the explanation in the comments is adequate or not?

I wrote a nice explanation in the remove duplicate as well as the function where we parse the root and its children.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: I was also thinking if we can fit this function before we make the folding ranges but that does not seem very convenient. Will that have a significant efficiency advantage? 

It is not convenient because we don't have a place where we can easily judge that these ranges will be duplicates and they start from the same line (like a list of CodeRanges).

I guess the current way is easily maintainable and readable code-wise,

sloorush:
> Do we have tests for all the weird cases we observed during testing

I'll try to write those as well, They were mostly very random, I will check if there is a pattern.

docs/features.md
michaelpj: It's a table of unimplemented stuff, so remove it!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: It's a little tricky, but `++` is linear in the first argument, so any time you see `++` and a tree traversal you want to be careful!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: I think this way is fine!

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
michaelpj: Oh I see, good point. I forgot that "code range" is more general than "selection range"!

michaelpj:
> I'll try to write those as well, They were mostly very random, I will check if there is a pattern.

Even writing down random examples is useful, it gives people something to start with if they want to investigate further.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +148,61 @@ findPosition pos root = go Nothing root
...
+-- | Traverses through the code range and it children to a folding ranges.
+--
+-- It starts with the root node, converts that into a folding range then moves towards the children.
+-- It converts each child of each root node and parses it to folding range and moves to its children.
+findFoldingRanges :: CodeRange -> [FoldingRange]
+findFoldingRanges r@(CodeRange _ children _) =
+    let frChildren :: [FoldingRange] = concat $ V.toList $ fmap findFoldingRanges children
+    in case createFoldingRange r of
+        Just x  -> x:frChildren
+        Nothing -> []
```
michaelpj, line 160: This should e `frChildren`! Did the tests keep passing with this? If they did, you are missing some tests!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: Good explanation. Are we 100% sure this is true, though? I feel like I've seen 1-line folding ranges in emacs where it turns
```
foo = ...
```
into
```
foo ...
```
which is a bit weird but not obviously wrong. Maybe we should pass them through and just let the client decide whether or not to use them? You said vscode ignores them already.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +148,61 @@ findPosition pos root = go Nothing root
...
+        Just _  -> Just (FoldingRange lineStart (Just charStart) lineEnd (Just charEnd) frk)
+        Nothing -> Nothing
+
+-- | Removes all small foldings that start from the same line.
+--
+-- As we are converting nodes of the ast into folding ranges, there are multiple nodes starting from a single line.
+-- A single line of code doesn't mean a single node in AST, so this function removes all the nodes that have a duplicate
+-- start line, ie. they start from the same line.
+--
+-- This function preserves the largest folding range from the ranges that start from the same line.
```
michaelpj, line 184: I think this is the most important line. This is the property that you want: for each line, we only return one folding range, and it should be the largest amongst all the folding ranges we have.

If we specify the behaviour like that then it suggests other ways we could implement this function, e.g. go through the list maintaining a map from lines to the largest folding range so far that starts on that line, and then each time we see a range, check if it is bigger than the existing one. That would remove the way the current implementation depends on the order.

It also raises some questions: folding ranges can overlap, with neither being bigger than the other (e.g. two ranges on the same line with character ranges [0, 5] and [1,6]). Which one should we pick then?

Lastly, are we sure this is what we want? Your example is interesting: why do we assume that people will want to fold the function rather than the if expression? I wonder if we *can* leave this up to the client. It seems like the server's job should just be to report all the ranges it knows about, and then the client can potentially make the decision about which ones to show to the user. What happens in e.g. vscode if we return multiple folding ranges starting on the same line?

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Do people use this kind of foldings?

I guess we should let it be like this, there are a lot of foldings already and it might make things chaotic.

Also talked about this with @kokobd earlier

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: I’ll test it manually as well. There were no errors in the codebase as well as the tests.

michaelpj: Well, that tells us the case you're not testing: a code range with a kind that we can't handle, but which has children that could provide valid folding ranges. That's the case that will go wrong her.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
kokobd: Then maybe we should preserve the longest folding range on each line?

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
kokobd: > folding ranges can overlap

folding ranges correspond to the ranges of AST nodes. I can not come up with a case where two AST nodes overlap, while neither of them is the parent of the other.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: > Then maybe we should preserve the longest folding range on each line?

I guess we should do this

michaelpj: What happens if we send all the ranges to the client? I guess you tried it and it was bad? Worth writing it down so we know why we're doing it.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: Good point! So that gives us some information. We can have parents and children on the same line, though so that suggests that we know the following: code ranges only overlap if one of them entirely contains the other one. That's useful. We still can't totally order them because we could also have *non*-overlapping code ranges on the same line (which one is bigger?).

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Nothing happens, vscode handles it well.

But the code is difficult to test (reference: https://github.com/haskell/haskell-language-server/pull/3058#issuecomment-1242659776 and the commit above this)

And if there is a client that does not handle this, it might go bad, so handling it on hls itself

michaelpj: I think if the client handles it then it's better to just give the client all the information! We can change it later if it's a problem.

I agree it's more effort to verify that the test output is correct, but that's useful work: we're going to return lots of small folding ranges, we _should_ think carefully about whether they're all right!

sloorush: Should I keep the code there and commented or should i remove it for handling the extra folding ranges?

kokobd: Then I vote to remove the deduplicating code if that doesn't change the user experience. You should leave a comment explaining the situation.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: So CodeRangeKind can only be -->
```
-- | 'CodeKind' represents the kind of a code range
data CodeRangeKind =
    -- | ordinary code
    CodeKindRegion
    -- | the group of imports
  | CodeKindImports
  -- | a comment
  | CodeKindComment
    deriving (Show, Eq, Generic, NFData)
```

I am not sure how can I mention something that does not have a valid codekind.

michaelpj: Then why is there a `Nothing` case in `crkToFrk`? (https://github.com/haskell/haskell-language-server/pull/3058/files#diff-b72658b4ae69940ecdda95504f965e4ac60e70b0c7582360149b9bb52bd361f6R201) I guess that's actually a redundant match and it could be total. `crkToFrk` returning `Nothing` is the case that would trip this up.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: Yeah, I think we can just a) not remove folding ranges on a single line and b) not remove folding ranges contained in other folding ranges, and add a note saying "we think the client can handle this, if not we could change to remove these in future".

sloorush: Okay, I’ll remove both of the filters and add a comment explaining the situation and edit the tests.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Yes makes sense, I’ll remove this case

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Closing as this functionality has been left for the client to handle

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -126,6 +150,34 @@ findPosition pos root = go Nothing root
...
+createFoldingRange (CodeRange (Range (Position lineStart charStart) (Position lineEnd charEnd)) _ ck) = do
+    let frk = crkToFrk ck
+    if lineStart == lineEnd
+        then Nothing
+    else case frk of
+        Just _  -> Just (FoldingRange lineStart (Just charStart) lineEnd (Just charEnd) frk)
+        Nothing -> Nothing
+
+-- Removes all small foldings that start from the same line
+removeDupStartLineFoldings :: [FoldingRange] -> [FoldingRange]
```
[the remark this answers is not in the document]
sloorush: Closing as this functionality has been left for the client to handle

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
sloorush: Resolved

sloorush:
@michaelpj, All requested changes have been resolved, Please take a look.

cc: @kokobd

kokobd:
It will be better to collapse the export list rather than the whole file when clicking on the folding marker at line 6. See below:
image

@sloorush I think it's okay to handle this in a subsequent PR if you are eager to merge this one.

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
```diff
@@ -50,5 +50,70 @@ testTree =
                         ( SelectionRange (Range (Position 1 1) (Position 5 10)) Nothing
                         )
                     )
-            ]
+            ],
+
+        -- TODO: Some more tests can be added on strange cases like
```
kokobd, line 55: TODO?

plugins/hls-code-range-plugin/test/testdata/selection-range/Import.hs
```diff
@@ -1,4 +1,4 @@
 module MultiPositions where
 
 import Data.List (find)
-import qualified Data.Foldable (foldl, foldl')
+import qualified Data.Foldable (foldl, foldl')
```
kokobd, line 4: Modified this file by accident?

plugins/hls-code-range-plugin/test/Ide/Plugin/CodeRangeTest.hs
sloorush: Writing down known issues: https://github.com/haskell/haskell-language-server/pull/3058#issuecomment-1247838543

plugins/hls-code-range-plugin/test/testdata/selection-range/Import.hs
sloorush: Ah yes, stylish-haskell was adding white spaces in the import lines, and it was messing up the selection ranges defined in the tests, I'll revert it.

sloorush: Reverted it back to normal

kokobd: Oh, maybe we could run the formatter in a separate PR.

sloorush:
> It will be better to collapse the export list rather than the whole file when clicking on the folding marker at line 6. See below: image
> 
> @sloorush I think it's okay to handle this in a subsequent PR if you are eager to merge this one.

We will have to tinker with code ranges and why is this code range so big, Should we make a new issue to tackle this?

plugins/hls-code-range-plugin/test/testdata/selection-range/Import.hs
sloorush: Yep, We'll also have to edit the test case file.

kokobd:
> We will have to tinker with code ranges and why is this code range so big, Should we make a new issue to tackle this?

I recommend adding a test case in `testdata` for that, and see if it can be easily solved (for example if you can figure out what AST node it corresponds to, then filter it out somewhere.) If it's not so trivial, open an issue for that, and add a link to the issue in that test case.

michaelpj:
Yeah, the best thing if you ever find a weird case is to record it with a test. Helps to remember and make it obvious why it's weird, and have somewhere to start to fix it.

sloorush:
Not really sure why this is happening but it happens when there is an export, I'll check how exports work in the AST and write if I find something on this

In this function:
image

The folding ranges generated look like this:
```
((2, 4) : (8, 21))
((2, 4) : (2, 10))
((5, 0) : (8, 21))
((5, 0) : (5, 25))
((5, 0) : (5, 6))
((5, 10) : (5, 25))
((5, 10) : (5, 17))
((5, 21) : (5, 25))
((6, 0) : (8, 21))
((6, 0) : (6, 6))
((6, 7) : (6, 8))
((6, 9) : (8, 21))
((6, 11) : (8, 21))
((6, 14) : (6, 28))
((6, 14) : (6, 23))
((6, 14) : (6, 15))
((6, 16) : (6, 21))
((6, 22) : (6, 23))
((6, 24) : (6, 26))
((6, 27) : (6, 28))
((7, 16) : (7, 20))
((8, 16) : (8, 21))
```

The first one is causing the issue, This happens when there is an export(even a single line one) like:

function: 
image

folding ranges:
```
((2, 22) : (7, 21))
((2, 22) : (2, 28))
((4, 0) : (7, 21))
((4, 0) : (4, 25))
((4, 0) : (4, 6))
((4, 10) : (4, 25))
((4, 10) : (4, 17))
((4, 21) : (4, 25))
((5, 0) : (7, 21))
((5, 0) : (5, 6))
((5, 7) : (5, 8))
((5, 9) : (7, 21))
((5, 11) : (7, 21))
((5, 14) : (5, 28))
((5, 14) : (5, 23))
((5, 14) : (5, 15))
((5, 16) : (5, 21))
((5, 22) : (5, 23))
((5, 24) : (5, 26))
((5, 27) : (5, 28))
((6, 16) : (6, 20))
((7, 16) : (7, 21))
```

sloorush:
So for this file:

 

The first folding range seen on VSCode collapses everything after this line



The first code range looks like this:

```
{_codeRange_range = Range {_start = Position {_line = 2, _character = 16}, _end = Position {_line = 7, _character = 21}} ...}
```

Which has the root node range ((2,16):(7,21)) which is from the start of the export parenthesis to the end of the file(even when there are functions that are not exported, it is folded till the end of the file)

Full Output of CodeRanges of this function:
```
CodeRange {_codeRange_range = Range {_start = Position {_line = 2, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 2, _character = 16}, _end = Position {_line = 2, _character = 22}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 17}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 21}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 5, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 7}, _end = Position {_line = 5, _character = 8}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 9}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 11}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 6, _character = 16}, _end = Position {_line = 6, _character = 20}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 7, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 2, _character = 16}, _end = Position {_line = 2, _character = 22}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 17}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 21}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 5, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 7}, _end = Position {_line = 5, _character = 8}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 9}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 11}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 6, _character = 16}, _end = Position {_line = 6, _character = 20}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 7, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 17}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 21}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 4, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 17}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 21}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 10}, _end = Position {_line = 4, _character = 17}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 4, _character = 21}, _end = Position {_line = 4, _character = 25}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 5, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 7}, _end = Position {_line = 5, _character = 8}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 9}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 11}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 6, _character = 16}, _end = Position {_line = 6, _character = 20}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 7, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 5, _character = 6}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 7}, _end = Position {_line = 5, _character = 8}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 9}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 11}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 6, _character = 16}, _end = Position {_line = 6, _character = 20}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 7, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 11}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 6, _character = 16}, _end = Position {_line = 6, _character = 20}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 7, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion},CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 14}, _end = Position {_line = 5, _character = 15}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 16}, _end = Position {_line = 5, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 22}, _end = Position {_line = 5, _character = 23}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 24}, _end = Position {_line = 5, _character = 26}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 5, _character = 27}, _end = Position {_line = 5, _character = 28}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 6, _character = 16}, _end = Position {_line = 6, _character = 20}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
CodeRange {_codeRange_range = Range {_start = Position {_line = 7, _character = 16}, _end = Position {_line = 7, _character = 21}}, _codeRange_children = [], _codeRange_kind = CodeKindRegion}
```

I believe this is an issue dealing with CodeRange conversion from the AST. 

@michaelpj @kokobd

kokobd:
> I believe this is an issue dealing with CodeRange conversion from the AST.

@sloorush So you can just skip the root node when constructing folding ranges? I believe no one wants to fold the whole file. I wonder why it starts from line 2 instead of line 0, but that's another problem.

sloorush:
> I wonder why it starts from line 2 instead of line 0

For a file where there are no exports, The first folding range of the first function just folds the entire file.



> So you can just skip the root node when constructing folding ranges

I'll do this

sloorush:
All set! Should be good to go now finally!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -179,9 +180,13 @@ createFoldingRange :: CodeRange -> Maybe FoldingRange
 createFoldingRange (CodeRange (Range (Position lineStart charStart) (Position lineEnd charEnd)) _ ck) = do
     -- Type conversion of codeRangeKind to FoldingRangeKind
     let frk = crkToFrk ck
-
     Just (FoldingRange lineStart (Just charStart) lineEnd (Just charEnd) (Just frk))
 
+-- | Removes first node from folding ranges
+removeFirstNode :: [FoldingRange] -> [FoldingRange]
+removeFirstNode []     = []
+removeFirstNode (_:xs) = xs
```
kokobd, line 188: If you are worried that `tail` is not a total function, you can use [`drop 1`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#v:drop)

sloorush: Makes sense, changed it.

kokobd:
Added "merge me" label. It will be squashed and merged automatically when CI passes.

michaelpj (approved):

PR #3059, opened 2022-07-24
Run pre-commit hooks

lunaticare:
I could not create a commit locally with [pre-commit hook installed](https://github.com/haskell/haskell-language-server/blob/master/docs/contributing/contributing.md#formatter-pre-commit-hook). Ran it locally on the clean branch:

```sh
pre-commit run --all-files
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3059"></a>


Ailrun (approved):
LGTM. Thank you :)

plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.formatted_range.hs
```diff
@@ -1,4 +1,4 @@
-foo ::   Int   -> String-> IO ()
+foo ::   Int   -> String-> IO ()
```
Ailrun, line 1: Oops, this should not be formatted. Could you add this to the ignore file list?

Ailrun (changes requested):

plugins/hls-brittany-plugin/test/testdata/BrittanyCRLF.formatted_range.hs
Ailrun: Huh, it looks like this file is already in the ignored list... You still cannot make a commit without this change?

lunaticare: Will fix it, thank you for noticing!

lunaticare: Fixed!

Ailrun (approved):

PR #3060, opened 2022-07-24
Add Github precommit workflow

lunaticare:
Add Github workflow to check code formatting in pull requests.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3060"></a>

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,6 +15,7 @@ jobs:
       with:
         ghc: "9.2.3"
         os: ${{ runner.os }}
+        shorten-hls: false
 
     - name: "Install stylish-haskell"
       run: |
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,42 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


lunaticare:
@Ailrun now that #3059 is merged, we should be able to enforce pre-commit hooks by default :)

Ailrun:
Oh I see where that Brittany issue happened... It's because https://github.com/haskell/haskell-language-server/blob/347a7187f20242540926b9927c59a45b18798c67/.pre-commit-config.yaml#L27 starts with `.` (which I believe shouldn't)

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: ./.github/actions/setup-build
+      with:
+        ghc: "9.2.3"
+        os: ${{ runner.os }}
```
Ailrun, line 17: You should not shorten the project name for this, I believe. (Otherwise format can change)

Ailrun (commented):
It looks like there was a concurrent PR without formatting (for Stan integration). Could you format the code as well?

.github/workflows/pre-commit.yml
lunaticare: Fixed!

Ailrun (approved):
LGTM

PR #3065, opened 2022-07-29
add a prefix to plugin CPP definitions

### LINKED ISSUE #3064: can not build with profiling due to CPP definition name clash with C standard library
### Your environment

Which OS do you use?
Ubuntu
GHC 9.2.3, installed through ghcup

`cabal.project.local`:
```
ignore-project: False
profiling: True
executable-profiling: True
flags: -dynamic
```

### Steps to reproduce

Checkout master, run `cabal build exe:haskell-language-server`

### Expected behaviour

It should compiles

### Actual behaviour

```
<command-line>: error: expected identifier or ‘(’ before numeric constant
`gcc' failed in phase `C Compiler'. (Exit code: 1)
```

### Reason

Dig into the actual `gcc` command issued by `ghc --make`, we see:
```
-Xpreprocessor -Drename
```

But in `<stdio.h>`, there already exists a function called `rename`: https://en.cppreference.com/w/c/io/rename

So this finally lead to this line in the preprocessed C file (because `rename` is replaced by `1`)
```
extern int 1 (const char *__old, const char *__new) __attribute__ ((__nothrow__ , __leaf__));
```
Then of course C compilation fails

### Solution

I recommend adding a prefix to all CPP definitions in haskell-language-server executable. For example change `-Drename` to `-Dhls_rename`.

kokobd:
fix #3064 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3065"></a>

### DIFF
--- a/exe/Plugins.hs
+++ b/exe/Plugins.hs
@@ -16,101 +16,100 @@ import qualified Ide.Plugin.Example2               as Example2
 import qualified Ide.Plugin.ExampleCabal           as ExampleCabal
 
 -- haskell-language-server optional plugins
-#if qualifyImportedNames
+#if hls_qualifyImportedNames
 import qualified Ide.Plugin.QualifyImportedNames   as QualifyImportedNames
 #endif
 
-#if callHierarchy
+#if hls_callHierarchy
 import qualified Ide.Plugin.CallHierarchy          as CallHierarchy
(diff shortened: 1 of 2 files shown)


fendor (approved):
LGTM, thank you!

Issue #3046, opened 2022-07-18
Migrate to OsPath


kokobd:
We have a better alternative to FilePath now. See: https://hasufell.github.io/posts/2022-06-29-fixing-haskell-filepaths.html

What we should do:
1. Change `NormalizedFilePath` to `NormalizedOsPath` after the issue in lsp to support OsPath https://github.com/haskell/lsp/issues/445 is solved. Profiling it to make sure the performance improves.
2. Change other usages of `FilePath` to `OsPath`

However, this can not be done until `ghc-9.6` is released (ships the new `filepath` library), or all ghc shipped libraries are available on Hackage.
> Many GHC bundled libraries do not have new versions released to Hackage (for example ghci and hpc, their versions on hackage is very old). Thus, cabal solver fails if I add filepath ^>= 1.4.100.0 to a project that depends on ghc, which then depends on ghci. The installed ghci links against filepath-1.4.2.2 in this case, and cabal can not rebuild a ghci with filepath ^>= 1.4.100.0, as the source code is not found on hackage.

pepeiborra:
> Thus, cabal solver fails if I add filepath ^>= 1.4.100.0 to a project that depends on ghc, which then depends on ghci.

Oh bummer. Well, if we really want to, it should be ok to embed `OsPath` and combinators in a `filepath-compat` package.

hasufell:
https://gitlab.haskell.org/ghc/ghc/-/issues/21887#note_443922

It may be possible to provide updated GHC releases that include new filepath. But this is still in discussion.

kokobd:
I don't know why, but the overall memory usage decreased from 1.5G to 700M
The data is collected by launching `haskell-language-server` from command line with profiling RTS options enabled.
<img width="1137" alt="image" src="https://user-images.githubusercontent.com/16440269/181737382-e6057143-f387-4a34-8260-a8c448d4944a.png">
<img width="1137" alt="image" src="https://user-images.githubusercontent.com/16440269/181737424-bf63f673-514b-4af8-a6ce-7bbee848b654.png">

hasufell:
@kokobd you mean between String and OsPath?

kokobd:
> @kokobd you mean between String and OsPath?

@hasufell Yes. But when loading HLS executable into VSCode, I no longer witness such a big difference. I'm still investigating the reason.

pepeiborra:
[...]

PR #3072, opened 2022-08-01
upgrade lsp to 1.5

kokobd:
This is a prerequisite of #3046

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3072"></a>


kokobd:
I'm receiving a lot of logs like this if `--verbose` is enabled in `ghcide-tests`. And finally got OOM even with 64GB memory.
image

So disabling it for now

kokobd:
I see. In `lsp`, there is a debug log, printing response body:
https://github.com/haskell/lsp/blob/8b63438313828a31ba1065ea865e504d532d093e/lsp/src/Language/LSP/Server/Control.hs#L224-L234
And in HLS, our logger sends a copy of log to LSP client:
https://github.com/haskell/haskell-language-server/blob/347a7187f20242540926b9927c59a45b18798c67/exe/Main.hs#L72-L82

Then it goes into infinite recursion... So we should not enable debug logs in `lsp` package itself.

michaelpj:
Hmm, I thought I had convinced myself that wasn't a problem :| I'm not 100% sure what to do about this. It's nice to be able to log the responses from LSP, but we do indeed have a loop problem if we then forward them on like this.

kokobd:
> It's nice to be able to log the responses from LSP

And some other tests depend on the logs, I had to fix them one by one. This makes the PR harder than I initially thought.

kokobd:
Finally, the CI is almost green.

plugins/hls-floskell-plugin/src/Ide/Plugin/Floskell.hs
```diff
@@ -33,7 +33,7 @@ provider _ideState typ contents fp _ = liftIO $ do
     config <- findConfigOrDefault file
     let (range, selectedContents) = case typ of
           FormatText    -> (fullRange contents, contents)
-          FormatRange r -> (r, extractRange r contents)
+          FormatRange r -> (normalize r, extractRange r contents)
```
fendor, line 36: Why this change? Is that a bug, where the fix doesn't actually belong here?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -48,10 +48,6 @@ import           Ide.Types
 import           Language.LSP.Server
 import           Language.LSP.Types
 
-instance Hashable Location
-instance Hashable Range
-instance Hashable Position
-instance Hashable UInt
```
fendor, line 54: Instances have been added to lsp-types-1.5. This is not a question or complaint, just a note for the next reviewer

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -322,38 +314,31 @@ expectNoMoreDiagnostics timeout doc src = do
...
+    when (timeout > 0) $
         -- Give any further diagnostic messages time to arrive.
         liftIO $ sleep timeout
         -- Send a dummy message to provoke a response from the server.
         -- This guarantees that we have at least one message to
         -- process, so message won't block or timeout.
-        void $ Test.sendNotification (SCustomMethod "non-existent-method") A.Null
-    handleMessages
+    testId <- Test.sendRequest (SCustomMethod "test") (A.toJSON GetFilesOfInterest)
+    handleMessages testId
```
fendor, line 324: this sounds like it is going to perform some actual work (although very minor), seems dubious to me?

fendor (approved):
LGTM, outstanding work!

plugins/hls-floskell-plugin/src/Ide/Plugin/Floskell.hs
kokobd, line 36: It won't work without this change. I don't know why. All other formatters use `normalize` already.

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -322,38 +314,31 @@ expectNoMoreDiagnostics timeout doc src = do
 -- If timeout is 0 it will wait until the session timeout
 waitForDiagnosticsFromSourceWithTimeout :: Seconds -> TextDocumentIdentifier -> String -> Test.Session [Diagnostic]
 waitForDiagnosticsFromSourceWithTimeout timeout document source = do
```
kokobd, line 316: Previously, this function relied on logs to work. But the current test logger doesn't send logs to lsp client. I found a very similar function in `ghcide` which doesn't rely on logs, so changed this function accordingly.

plugins/hls-floskell-plugin/src/Ide/Plugin/Floskell.hs
fendor: I assumed, floskell performed the normalisation internally. That, or no one is running the tests.

hls-test-utils/src/Test/Hls/Util.hs
fendor: Right, but why are you sending `(A.toJSON GetFilesOfInterest)`. Seems like `A.Null` would express our intent more clearly?

plugins/hls-floskell-plugin/src/Ide/Plugin/Floskell.hs
kokobd: The test case "formats a range" failed.
I didn't find out the reason, but all other formatting plugins use `normalize`, and this fixed the problem.

hls-test-utils/src/Test/Hls/Util.hs
kokobd: I adjusted the indentation to 4 spaces so the diff is messier. This is previously `SCustomMethod "non-existent-method"`, and the client waits for a logging diagnostics from the server, which is no longer present. I think instead of adding the log back, change the test to not rely on a specific form of logger is better.

And there is a similar function in `ghcide` which is newer and doesn't rely on logs to function properly:
https://github.com/haskell/haskell-language-server/blob/347a7187f20242540926b9927c59a45b18798c67/ghcide/test/src/Development/IDE/Test.hs#L92-L101

hls-test-utils/src/Test/Hls/Util.hs
kokobd: Yes, `A.Null` is better, and is working!

plugins/hls-floskell-plugin/src/Ide/Plugin/Floskell.hs
fendor: good enough for me

ghcide/test/exe/Main.hs
```diff
@@ -6986,18 +6990,20 @@ positionMappingTests =
...
-                        pure (range, replacement, newPos)
-                forAll
-                    (suchThatMap gen
-                        (\(range, replacement, newPos) -> positionResultToMaybe $ (range, replacement, newPos,) <$> fromCurrent range replacement newPos)) $
-                    \(range, replacement, newPos, oldPos) ->
-                    toCurrent range replacement oldPos === PositionExact newPos
+                        pure (rope, range, replacement)
+                monadicIO $ forAllM gen
+                    $ \(rope, range, replacement) -> do
+                        newRope <- liftIO $ applyChange (toCologActionWithPrio $ cmapWithPrio LogVfs recorder) rope (TextDocumentContentChangeEvent (Just range) Nothing replacement)
```
michaelpj, line 7001: Why not pass `mempty :: LogAction Identity` here? then you don't need IO, and you'll just not get hte log messages which you don't care about anyway. Then I think you can keep the old code.

michaelpj (approved):

ghcide/src/Development/IDE/Types/Logger.hs
```diff
@@ -381,3 +380,14 @@ priorityToLsp =
     Info    -> MtInfo
     Warning -> MtWarning
     Error   -> MtError
+
+toCologActionWithPrio :: (MonadIO m, HasCallStack) => Recorder (WithPriority msg) -> LogAction m (WithSeverity msg)
```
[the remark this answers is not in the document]
kokobd: This is the first time `co-log-core` was introduced in HLS, so I think there isn't a existing function for this.

ghcide/test/exe/Main.hs
kokobd: Good advice!

PR #3078, opened 2022-08-06
Add support for Fourmolu 0.8

brandonchinn178:
Also includes some clean up. Will we always be supporting every version of Fourmolu, or will HLS ever remove support for old versions of Fourmolu?

cc @parsonsmatt @georgefst 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3078"></a>

### DIFF
--- a/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
+++ b/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
@@ -117,7 +117,7 @@ provider recorder plId ideState typ contents fp fo = withIndefiniteProgress titl
              in liftIO (loadConfigFile fp') >>= \case
                     ConfigLoaded file opts -> liftIO $ do
                         logWith recorder Info $ ConfigPath file
-                        format (toConfig opts)
+                        format opts
                     ConfigNotFound searchDirs -> liftIO $ do
                         logWith recorder Info $ NoConfigPath searchDirs
                         format emptyConfig
--- a/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
+++ b/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
@@ -16,7 +16,7 @@ import           Control.Exception               (IOException, try)
 import           Control.Lens                    ((^.))
 import           Control.Monad
 import           Control.Monad.IO.Class
-import           Data.Bifunctor                  (first)
+import           Data.Bifunctor                  (bimap, first)
 import           Data.Maybe
 import           Data.Text                       (Text)
 import qualified Data.Text                       as T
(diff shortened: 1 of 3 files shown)


plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu/Shim.hs
```diff
@@ -0,0 +1,68 @@
...
+    , cfgFileFixities = mempty
+    }
+#endif
+
+emptyConfig :: FourmoluConfig
+emptyConfig =
+  FourmoluConfig
+    { cfgFilePrinterOpts = mempty
+    , cfgFileFixities = mempty
+    }
```
brandonchinn178, line 46: will eventually be under a `#if !MIN_VERSION_fourmolu(0,8,1)` macro: https://github.com/fourmolu/fourmolu/pull/221

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -100,8 +100,8 @@ provider recorder plId ideState typ contents fp fo = withIndefiniteProgress titl
                         pure . Left . responseError $ "Fourmolu failed with exit code " <> T.pack (show n)
         else do
             let format fourmoluConfig =
-                    first (mkError . show)
-                        <$> try @OrmoluException (makeDiffTextEdit contents <$> ormolu config fp' (T.unpack contents))
+                    bimap (mkError . show) (makeDiffTextEdit contents)
+                        <$> try @OrmoluException (ormolu config fp' (T.unpack contents))
```
georgefst, line 103: Nice. Surpised I didn't notice that opportunity.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu/Shim.hs
```diff
@@ -0,0 +1,68 @@
+{-# LANGUAGE CPP #-}
+
+module Ide.Plugin.Fourmolu.Shim (
```
georgefst, line 3: Excellent. This is much cleaner.

georgefst (approved):
Thanks!

georgefst:
> Will we always be supporting every version of Fourmolu, or will HLS ever remove support for old versions of Fourmolu?

I don't see a reason to keep _really_ old versions around, especially if it means huge amounts of CPP. But it's likely that some people won't want to be forced in to updating Fourmolu (and thus probably reformatting their codebase) by an HLS update\*. Then again, part of the reason for adding the CLI mode was to give users that flexibility.

I've wanted to remove some backwards compatibility before, but it wasn't possible then because of constraints from HLS' other dependencies: https://github.com/haskell/haskell-language-server/pull/2951.

\* Although in practice, most people are getting HLS from bindists, so they don't have much choice without building from source instead.

brandonchinn178:
@georgefst By "building from source", do you mean fourmolu? Now that we release binaries on the releases page, is there some hls logic that can download fourmolu binaries where possible instead of building fourmolu? (Even better, make that the canonical way; dont bundle fourmolu in hls, just have hls provide a dropdown to install version of fourmolu)

Keeping old versions isnt terrible right now, because all the cpp is for the 0.7 upgrade. But I think maybe once 0.7 becomes two years old (or something) itd be nice to drop support

georgefst:
> @georgefst By "building from source", do you mean fourmolu?

I meant HLS, but I suppose that statement works for either, once you take the CLI option in to account.

We _could_ make downloading the binary the default, but I'm more confident of correctness when using the library. And it would go against the way every other HLS plugin is doing things (although that is not, in itself, a great argument against).

PR #3079, opened 2022-08-06
fix lsp-types benchmark

### LINKED ISSUE #3102: CI bench-unit job failures
The upload of the .cabal folder to the CI cache fails with a 400. Example below: 

https://github.com/haskell/haskell-language-server/runs/7818378102?check_suite_focus=true

pepeiborra:
The lsp-types benchmark definition needs some fixes in ghc 9.2

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3079"></a>

Also fixes #3102 by tar'ing the benchmark artefacts before upload

### DIFF
--- a/ghcide/bench/lib/Experiments.hs
+++ b/ghcide/bench/lib/Experiments.hs
@@ -24,8 +24,9 @@ module Experiments
 ) where
 import           Control.Applicative.Combinators (skipManyTill)
 import           Control.Exception.Safe          (IOException, handleAny, try)
-import           Control.Monad.Extra
-import           Control.Monad.Fail
+import Control.Monad.Extra
+    ( forM_, void, unless, forM, (&&^), allM, whenJust )
+import           Control.Monad.Fail              (MonadFail)
 import           Control.Monad.IO.Class
 import           Data.Aeson                      (Value (Null), toJSON)
 import           Data.Either                     (fromRight)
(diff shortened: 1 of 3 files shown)


ghcide/bench/lib/Experiments.hs
```diff
@@ -563,6 +563,9 @@ setup = do
                 writeFile
                     (path </> "cabal.project.local")
                     ""
+                -- Fix newlines, required for GHC 9.2 and lsp-types
```
kokobd, line 566: Why is this required? Does this mean HLS has some troubles when working on `lsp-types`?

kokobd (approved):

ghcide/bench/lib/Experiments.hs
pepeiborra: Honestly, I have no idea. The issue does not repro with ghc 8.10.7, which makes me think it's not an HLS bug. But it doesn't repro in `cabal repl` with ghc 9.2 either, so 🤷‍♂️🤷‍♂️🤷‍♂️

michaelpj:
Maybe we should switch to a more stable package for benchmarking :laughing:

pepeiborra:
> Maybe we should switch to a more stable package for benchmarking 😆

True, I chose `lsp-types` because it is guaranteed to be buildable by our CI, uses TH, and doesn't increase our dependency closure, but there are other options like `aeson` that also check those boxes

PR #3080, opened 2022-08-06
Feat: basic record dot completions

coltenwebb:
When a user dots a record type with OverloadedRecordDot enabled, shows all possible record selectors for that type.

Todo:
* write test
* make completion items more complete

Notes
* Requires fresh HieAst to work, which shouldn't be a problem since we are dotting an actively worked on file
* Doesn't support HasField (virtual fields)
* Moves getCompletionPrefix, PosPrefixInfo in from LSP.Language.VFS (and patches them to support records in addition to modules)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3080"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -48,7 +48,7 @@ import qualified Language.LSP.VFS                         as VFS
 import           Numeric.Natural
 import           Text.Fuzzy.Parallel                      (Scored (..))
 
-import qualified GHC.LanguageExtensions                       as LangExt
+import qualified GHC.LanguageExtensions                   as LangExt
 import           Language.LSP.Types
 
 data Log = LogShake Shake.Log deriving Show
@@ -167,8 +167,8 @@ getCompletionsLSP ide plId
                 let clientCaps = clientCapabilities $ shakeExtras ide
                     plugins = idePlugins $ shakeExtras ide
                 config <- getCompletionsConfig plId
-                
-                allCompletions <- liftIO $ getCompletions plId ideOpts cci' parsedMod astres bindMap pfix' clientCaps config moduleExports
+
+                allCompletions <- liftIO $ getCompletions plugins ideOpts cci' parsedMod astres bindMap pfix' clientCaps config moduleExports
                 pure $ InL (List $ orderedCompletions allCompletions)
               _ -> return (InL $ List [])
           _ -> return (InL $ List [])
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -151,24 +151,23 @@ getCompletionsLSP ide plId
 #endif
             ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
             astres <- case ms of
-              Just ms' -> if uses_overloaded_record_dot ms'
-                then useWithStaleFast GetHieAst npath
-                else return Nothing
-              Nothing -> return Nothing
+              Just ms' | uses_overloaded_record_dot ms'
+                ->  useWithStaleFast GetHieAst npath
+              _ -> return Nothing
 
             pure (opts, fmap (,pm,binds) compls, moduleExports, astres)
         case compls of
           Just (cci', parsedMod, bindMap) -> do
-            pfix <- getCompletionPrefix position cnts
+            let pfix = getCompletionPrefix position cnts
             case (pfix, completionContext) of
-              (Just (PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
+              ((PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
                 -> return (InL $ List [])
-              (Just pfix', _) -> do
+              (_, _) -> do
                 let clientCaps = clientCapabilities $ shakeExtras ide
                     plugins = idePlugins $ shakeExtras ide
                 config <- getCompletionsConfig plId
 
-                allCompletions <- liftIO $ getCompletions plugins ideOpts cci' parsedMod astres bindMap pfix' clientCaps config moduleExports
+                allCompletions <- liftIO $ getCompletions plugins ideOpts cci' parsedMod astres bindMap pfix clientCaps config moduleExports
                 pure $ InL (List $ orderedCompletions allCompletions)
               _ -> return (InL $ List [])
           _ -> return (InL $ List [])
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -48,7 +48,7 @@ import qualified Language.LSP.VFS                         as VFS
 import           Numeric.Natural
 import           Text.Fuzzy.Parallel                      (Scored (..))
 
-import qualified GHC.LanguageExtensions as LangExt
+import qualified GHC.LanguageExtensions                       as LangExt
 import           Language.LSP.Types
 
 data Log = LogShake Shake.Log deriving Show
@@ -142,12 +142,14 @@ getCompletionsLSP ide plId
                 exportsCompItems = foldMap (map (fromIdentInfo uri) . Set.toList) . Map.elems . getExportsMap $ exportsMap
                 exportsCompls = mempty{anyQualCompls = exportsCompItems}
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls <> Just lModules
-            
+
             -- get HieAst if OverloadedRecordDot is enabled
-            let uses_th_qq (ms_hspp_opts -> dflags) = xopt LangExt.OverloadedRecordDot dflags
+            let uses_overloaded_record_dot (ms_hspp_opts . msrModSummary -> dflags) = xopt LangExt.OverloadedRecordDot dflags
             ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
-            astres <- case ms of 
-              Just ms' -> if uses_th_qq . msrModSummary  $ ms' then useWithStaleFast GetHieAst npath else return Nothing
+            astres <- case ms of
+              Just ms' -> if uses_overloaded_record_dot ms'
+                then useWithStaleFast GetHieAst npath
+                else return Nothing
               Nothing -> return Nothing
 
             pure (opts, fmap (,pm,binds) compls, moduleExports, astres)
@@ -162,7 +164,7 @@ getCompletionsLSP ide plId
                     plugins = idePlugins $ shakeExtras ide
                 config <- getCompletionsConfig plId
                 
-                allCompletions <- liftIO $ getCompletions plugins ideOpts cci' parsedMod astres bindMap pfix' clientCaps config moduleExports
+                allCompletions <- liftIO $ getCompletions plId ideOpts cci' parsedMod astres bindMap pfix' clientCaps config moduleExports
                 pure $ InL (List $ orderedCompletions allCompletions)
               _ -> return (InL $ List [])
           _ -> return (InL $ List [])
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -48,6 +48,9 @@ import qualified Language.LSP.VFS                         as VFS
 import           Numeric.Natural
 import           Text.Fuzzy.Parallel                      (Scored (..))
 
+import qualified GHC.LanguageExtensions as LangExt
+import           Language.LSP.Types
+
 data Log = LogShake Shake.Log deriving Show
 
 instance Pretty Log where
@@ -120,7 +123,7 @@ getCompletionsLSP ide plId
     fmap Right $ case (contents, uriToFilePath' uri) of
       (Just cnts, Just path) -> do
         let npath = toNormalizedFilePath' path
-        (ideOpts, compls, moduleExports) <- liftIO $ runIdeAction "Completion" (shakeExtras ide) $ do
+        (ideOpts, compls, moduleExports, astres) <- liftIO $ runIdeAction "Completion" (shakeExtras ide) $ do
             opts <- liftIO $ getIdeOptionsIO $ shakeExtras ide
             localCompls <- useWithStaleFast LocalCompletions npath
             nonLocalCompls <- useWithStaleFast NonLocalCompletions npath
@@ -139,19 +142,27 @@ getCompletionsLSP ide plId
                 exportsCompItems = foldMap (map (fromIdentInfo uri) . Set.toList) . Map.elems . getExportsMap $ exportsMap
                 exportsCompls = mempty{anyQualCompls = exportsCompItems}
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls <> Just lModules
+            
+            -- get HieAst if OverloadedRecordDot is enabled
+            let uses_th_qq (ms_hspp_opts -> dflags) = xopt LangExt.OverloadedRecordDot dflags
+            ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
+            astres <- case ms of 
+              Just ms' -> if uses_th_qq . msrModSummary  $ ms' then useWithStaleFast GetHieAst npath else return Nothing
+              Nothing -> return Nothing
 
-            pure (opts, fmap (,pm,binds) compls, moduleExports)
+            pure (opts, fmap (,pm,binds) compls, moduleExports, astres)
         case compls of
           Just (cci', parsedMod, bindMap) -> do
-            pfix <- VFS.getCompletionPrefix position cnts
+            pfix <- getCompletionPrefix position cnts
             case (pfix, completionContext) of
-              (Just (VFS.PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
+              (Just (PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
                 -> return (InL $ List [])
               (Just pfix', _) -> do
                 let clientCaps = clientCapabilities $ shakeExtras ide
                     plugins = idePlugins $ shakeExtras ide
                 config <- getCompletionsConfig plId
-                allCompletions <- liftIO $ getCompletions plugins ideOpts cci' parsedMod bindMap pfix' clientCaps config moduleExports
+                
+                allCompletions <- liftIO $ getCompletions plugins ideOpts cci' parsedMod astres bindMap pfix' clientCaps config moduleExports
                 pure $ InL (List $ orderedCompletions allCompletions)
               _ -> return (InL $ List [])
           _ -> return (InL $ List [])
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -144,7 +144,11 @@ getCompletionsLSP ide plId
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls <> Just lModules
 
             -- get HieAst if OverloadedRecordDot is enabled
+#if MIN_VERSION_ghc(9,2,0)
             let uses_overloaded_record_dot (ms_hspp_opts . msrModSummary -> dflags) = xopt LangExt.OverloadedRecordDot dflags
+#else
+            let uses_overloaded_record_dot _ = False
+#endif
             ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
             astres <- case ms of
               Just ms' -> if uses_overloaded_record_dot ms'
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -10,6 +10,7 @@ module Development.IDE.Plugin.Completions.Logic (
 , localCompletionsForParsedModule
 , getCompletions
 , fromIdentInfo
+, getCompletionPrefix
 ) where
 
 import           Control.Applicative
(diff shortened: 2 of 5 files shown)


ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -139,18 +141,26 @@ getCompletionsLSP ide plId
                 exportsCompItems = foldMap (map (fromIdentInfo uri) . Set.toList) . Map.elems . getExportsMap $ exportsMap
                 exportsCompls = mempty{anyQualCompls = exportsCompItems}
             let compls = (fst <$> localCompls) <> (fst <$> nonLocalCompls) <> Just exportsCompls <> Just lModules
+            
+            -- get HieAst if OverloadedRecordDot is enabled
+            let uses_th_qq (ms_hspp_opts -> dflags) = xopt LangExt.OverloadedRecordDot dflags
+            ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
+            astres <- case ms of 
+              Just ms' -> if uses_th_qq . msrModSummary  $ ms' then useWithStaleFast GetHieAst npath else return Nothing
```
pepeiborra, line 149:
```suggestion
            let uses_overloaded_record_dot (ms_hspp_opts.msrModSummary -> dflags) = 
                  xopt LangExt.OverloadedRecordDot dflags
            ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
            astres <- case ms of 
              Just ms' -> if uses_overloaded_record_dot ms' 
                then useWithStaleFast GetHieAst npath 
                else return Nothing
```

ghcide/src/Development/IDE/Plugin/Completions/Types.hs
```diff
@@ -136,3 +137,25 @@ instance Monoid CachedCompletions where
 instance Semigroup CachedCompletions where
     CC a b c d e <> CC a' b' c' d' e' =
         CC (a<>a') (b<>b') (c<>c') (d<>d') (e<>e')
+
+
+-- moved here from Language.LSP.VHS
```
pepeiborra, line 142:
```suggestion
-- moved here from Language.LSP.VFS
```

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -884,3 +942,33 @@ mergeListsBy cmp all_lists = merge_lists all_lists
...
+            if | T.null beforePos -> Just ""
+               | T.last beforePos == ' ' -> Just "" -- don't count abc as the curword in 'abc '
+               | otherwise -> lastMaybe (T.words beforePos)
+
+        let parts = T.split (=='.')
+                      $ T.takeWhileEnd (\x -> isAlphaNum x || x `elem` ("._'"::String)) curWord
+        case reverse parts of
+          [] -> Nothing
+          (x:xs) -> do
+            let modParts = dropWhile (\_ -> False)
```
pepeiborra, line 968: what is the `dropWhile` for?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -884,3 +942,33 @@ mergeListsBy cmp all_lists = merge_lists all_lists
         []     -> []
         [xs]   -> xs
         lists' -> merge_lists lists'
+
+
+getCompletionPrefix :: (Monad m) => Position -> VFS.VirtualFile -> m (Maybe PosPrefixInfo)
+getCompletionPrefix pos@(Position l c) (VFS.VirtualFile _ _ ropetext) =
```
pepeiborra, line 948: Is this from Language.LSP.VFS? What have you changed?

michaelpj: It doesn't really belong there, so moving it here is good (I'd like to delete it from `lsp` afterwards).

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: left over from adapting getCompletionsPrefix to working with record dot, removed now

coltenwebb:
@pepeiborra Would you mind doing another review? It should be good to go.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -141,18 +142,32 @@ getCompletionsLSP ide plId
...
+            let uses_overloaded_record_dot (ms_hspp_opts . msrModSummary -> dflags) = xopt LangExt.OverloadedRecordDot dflags
+#else
+            let uses_overloaded_record_dot _ = False
+#endif
+            ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
+            astres <- case ms of
+              Just ms' -> if uses_overloaded_record_dot ms'
+                then useWithStaleFast GetHieAst npath
+                else return Nothing
+              Nothing -> return Nothing
```
pepeiborra, line 156:
```suggestion
              Just ms' | uses_overloaded_record_dot ms'
                 ->  useWithStaleFast GetHieAst npath
              _ -> return Nothing
```

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -890,3 +929,33 @@ mergeListsBy cmp all_lists = merge_lists all_lists
...
+      return $ Just $ fromMaybe (PosPrefixInfo "" "" "" pos) $ do -- Maybe monad
+        let headMaybe []    = Nothing
+            headMaybe (x:_) = Just x
+            lastMaybe []     = Nothing
+            lastMaybe [x]    = Just x
+            lastMaybe (_:xs) = lastMaybe xs
```
pepeiborra, line 941: Note that `headMaybe` = `Data.Maybe.maybeToList`

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -632,10 +670,11 @@ getCompletions plugins ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls,
...
 
-          compls = if T.null prefixModule
-            then map (notQual,) localCompls ++ map (qual,) unqualCompls ++ ((notQual,) . ($Nothing) <$> anyQualCompls)
-            else ((qual,) <$> Map.findWithDefault [] prefixModule (getQualCompls qualCompls))
-                 ++ ((notQual,) . ($ Just prefixModule) <$> anyQualCompls)
+          compls
+            | T.null prefixScope = map (notQual,) localCompls ++ map (qual,) unqualCompls ++ ((notQual,) . ($ Nothing) <$> anyQualCompls)
+            | not $ null recordDotSyntaxCompls = recordDotSyntaxCompls
+            | otherwise = ((qual,) <$> Map.findWithDefault [] prefixScope (getQualCompls qualCompls))
+                 ++ ((notQual,) . ($ Just prefixScope) <$> anyQualCompls)
```
pepeiborra, line 677: Should we union these two sources of completions when a prefix is present?

pepeiborra (approved):

test/functional/Completion.hs
```diff
@@ -84,6 +84,23 @@ tests = testGroup "completions" [
...
+            doc <- openDoc "RecordDotSyntax.hs" "haskell"
+
+            let te = TextEdit (Range (Position 25 0) (Position 25 5)) "z = x.a"
+            _ <- applyEdit doc te
+
+            compls <- getCompletions doc (Position 25 6)
+            item <- getCompletionByLabel "a" compls
+            liftIO $ do
+                item ^. label @?= "a"
+                --item ^. detail @?= Just "Data.List"  TODO
```
michaelpj, line 99: TODO?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -890,3 +929,33 @@ mergeListsBy cmp all_lists = merge_lists all_lists
         []     -> []
         [xs]   -> xs
         lists' -> merge_lists lists'
+
+
+getCompletionPrefix :: (Monad m) => Position -> VFS.VirtualFile -> m (Maybe PosPrefixInfo)
```
michaelpj, line 934: Needs haddock

michaelpj, line 934: Just requiring a `Monad` constraint is a bit suspicious: do you need that at all, can you not just return `Maybe PosPrefixInfo`?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -890,3 +929,33 @@ mergeListsBy cmp all_lists = merge_lists all_lists
...
+getCompletionPrefix pos@(Position l c) (VFS.VirtualFile _ _ ropetext) =
+      return $ Just $ fromMaybe (PosPrefixInfo "" "" "" pos) $ do -- Maybe monad
```
michaelpj, line 936: This is quite weird: you're unconditionally returning a `Just`, but you _have_ a `Maybe` already, but you're swapping out the `Nothing` case with a default value? Quite confusing. Maybe you just want to also drop the `Maybe` from the return type?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -890,3 +929,33 @@ mergeListsBy cmp all_lists = merge_lists all_lists
...
+        let headMaybe []    = Nothing
+            headMaybe (x:_) = Just x
+            lastMaybe []     = Nothing
+            lastMaybe [x]    = Just x
+            lastMaybe (_:xs) = lastMaybe xs
+
+        curLine <- headMaybe $ T.lines $ Rope.toText
```
michaelpj, line 943: what are you trying to do? worth explaining, much of this is not obvious

michaelpj: Also this is different from the `lsp` version now, what changed?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
michaelpj: I see, this was here before. Well, it makes no sense so we might as well improve it.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -606,6 +614,36 @@ getCompletions plugins ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls,
                   hpos = upperRange position'
               in getCContext lpos pm <|> getCContext hpos pm
 
+
+          -- we need the hieast to be fresh
+          -- not fresh, hasfield won't have a chance. it would to another larger change to ghc IfaceTyCon to contain record fields
```
michaelpj, line 619: This seems subtle and important: write more! You spent lots of time figuring this out, don't let the knowledge be lost!

michaelpj: Also, it's not clear what this comment relates to. Nothing here ensures that the AST is fresh, so what's going on?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -606,6 +614,36 @@ getCompletions plugins ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls,
...
+              nodeCompletions :: HieAST Type -> [(Bool, CompItem)]
+              nodeCompletions node = concatMap g (nodeType $ nodeInfo node)
+              g :: Type -> [(Bool, CompItem)]
+              g (TyConApp theTyCon _) = map (dotFieldSelectorToCompl (printOutputable $ GHC.tyConName theTyCon)) $ getSels theTyCon
+              g _ = []
+              getSels :: GHC.TyCon -> [T.Text]
+              getSels tycon = let f fieldLabel = printOutputable fieldLabel
+                              in map f $ tyConFieldLabels tycon
+              dotFieldSelectorToCompl :: T.Text -> T.Text -> (Bool, CompItem)
+              --dotFieldSelectorToCompl label = (True, CI CiVariable label (ImportedFrom T.empty) Nothing label Nothing emptySpanDoc False Nothing)
```
michaelpj, line 634: commented out code

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -606,6 +614,36 @@ getCompletions plugins ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls,
...
+              dotFieldSelectorToCompl recname label = (True, CI
+                { compKind = CiField
+                , insertText = label
+                , provenance = DefinedIn recname
+                , typeText = Nothing
+                , label = label
+                , isInfix = Nothing
+                , docs = emptySpanDoc
```
michaelpj, line 642: it should have doc, probably?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -606,6 +614,36 @@ getCompletions plugins ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls,
...
+                , isTypeCompl = False
+                , additionalTextEdits = Nothing
```
michaelpj, line 644: we might potentially need to import something?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -632,10 +670,11 @@ getCompletions plugins ideOpts CC {allModNamesAsNS, anyQualCompls, unqualCompls,
               ty = showForSnippet <$> typ
               thisModName = Local $ nameSrcSpan name
 
-          compls = if T.null prefixModule
-            then map (notQual,) localCompls ++ map (qual,) unqualCompls ++ ((notQual,) . ($Nothing) <$> anyQualCompls)
-            else ((qual,) <$> Map.findWithDefault [] prefixModule (getQualCompls qualCompls))
-                 ++ ((notQual,) . ($ Just prefixModule) <$> anyQualCompls)
+          compls
+            | T.null prefixScope = map (notQual,) localCompls ++ map (qual,) unqualCompls ++ ((notQual,) . ($ Nothing) <$> anyQualCompls)
```
michaelpj, line 674: I know this was here before, but I find this a bit too point-free and quite obscure. I would be happy if it became a bit less so...

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
michaelpj: Worth an explanation regardless!

ghcide/src/Development/IDE/Plugin/Completions/Types.hs
```diff
@@ -136,3 +137,25 @@ instance Monoid CachedCompletions where
 instance Semigroup CachedCompletions where
     CC a b c d e <> CC a' b' c' d' e' =
         CC (a<>a') (b<>b') (c<>c') (d<>d') (e<>e')
+
+
+-- moved here from Language.LSP.VFS
```
michaelpj, line 142: historical note, probably not necessary

test/testdata/completion/RecordDotSyntax.hs
```diff
@@ -0,0 +1,27 @@
...
+  , c2 :: MyChild
+  } deriving (Eq, Show)
+
+newtype MyChild = MyChild
+  { z :: String
+  } deriving (Eq, Show)
+
+x = MyRecord1 { a = "Hello", b = 12, c = MyChild { z = "there" } }
+
+y = x.a ++ show x.b
```
michaelpj, line 25: What about some nested completions? e.g. `x.y.z`?

michaelpj (changes requested):

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: The thinking here was that if there are record completions, they will be the only valid completions (since we get record completions from `myrecord.var`, and everything else would only work with something like `myrecord .somefn`, with a space, in which case record completions is null). I'll add a comment. Does that make sense though?

test/functional/Completion.hs
coltenwebb: The TODO was in regards to filling out more of the details field with where the type was defined. It turned out that due to HieAST, the fields' definitions weren't easy to get at (since it would give the types of the fields, but not the fields themselves). I added a comment to Logic.hs about this, and removed the TODO.

michaelpj: Maybe worth leaving a todo somewhere since there is something that we might want to improve later?

michaelpj: Or even a known broken test!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: Yep, not sure why it was like that

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: I'll document in comments

coltenwebb: But the reason for the change was different rope libraries, though it looks like the versions are the same now.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: Made the comment more detailed, and now in proper English. Basically when doing autocomplete its possible for the HieAST to be "stale" meaning its loaded from an hie file rather than generated. But when this happens, it's not possible to get the record fields. This only happens when the current file is first loaded, if an hie file was generated before. As soon as the user edits it, it's fresh again, so I never ran into trouble from this.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: I believe that other autocompletions don't provide docs atm, e.g. `Logic.hs:344`. Though we do provide them for snippets. I worry putting something like "record dot completion" would distract the developer. Did you have something else in mind?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: This is a todo. Since I get at the fields in a roundabout way I can't trivially adapt the existing `ExtendImport` to this. It requires a way to get the record's module, which I didn't see included in the type information I read. I'll leave a comment about this.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: Alright, refactored the anyQualCompls parts since those seemed the worst

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -140,18 +143,31 @@ getCompletionsLSP ide plId
...
 
-            pure (opts, fmap (,pm,binds) compls, moduleExports)
+            -- get HieAst if OverloadedRecordDot is enabled
+#if MIN_VERSION_ghc(9,2,0)
+            let uses_overloaded_record_dot (ms_hspp_opts . msrModSummary -> dflags) = xopt LangExt.OverloadedRecordDot dflags
+#else
+            let uses_overloaded_record_dot _ = False
+#endif
+            ms <- fmap fst <$> useWithStaleFast GetModSummaryWithoutTimestamps npath
+            astres <- case ms of
```
michaelpj, line 153: Possibly stupid question, but why guard here? What happens if we just guard getting the HIE AST on whether or not we have record dot?

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -140,18 +143,31 @@ getCompletionsLSP ide plId
...
+              _ -> return Nothing
+
+            pure (opts, fmap (,pm,binds) compls, moduleExports, astres)
         case compls of
           Just (cci', parsedMod, bindMap) -> do
-            pfix <- VFS.getCompletionPrefix position cnts
+            let pfix = getCompletionPrefix position cnts
             case (pfix, completionContext) of
-              (Just (VFS.PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
+              ((PosPrefixInfo _ "" _ _), Just CompletionContext { _triggerCharacter = Just "."})
```
michaelpj, line 163: OOI, what *is* happening in this case? I guess this is basically preventing us from giving completions when people write ` .` which usually corresponds to function composition?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
michaelpj: Well, fields have haddock usually, you know
```
data Foo {
  hello :: Int -- ^ This is the doc for foo
}
```
I would expect to get the field haddock in the completion doc!

This might get automatically sorted out for us with https://github.com/haskell/haskell-language-server/pull/3204, not sure.

michaelpj (approved):

michaelpj:
Looks like the new test failed on 9.0.2?

ghcide/src/Development/IDE/Plugin/Completions.hs
coltenwebb: If I understand you correctly, that's exactly what I'm trying to do here. Maybe there's a simpler way to do this?

ghcide/src/Development/IDE/Plugin/Completions.hs
coltenwebb: That's how I read this too. `getCompletionPrefix` doesn't find a prefix word before the cursor so it returns emptiness.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
coltenwebb: Ah that makes much more sense than what I was thinking. If that pr doesn't sort this out, it looks like it will provide the machinery to make it easier.

ghcide/src/Development/IDE/Plugin/Completions.hs
michaelpj: No, it was a stupid question: I missed that the `uses_overloaded_record_dot` predicate needs the module summary!

michaelpj:
Last thing: could you make some issues for any remaining user-facing problems, like maybe the documentation one?

PR #3085, opened 2022-08-10
Support ghc-9.2.4

July541:


.github/workflows/caching.yml
```diff
@@ -80,8 +80,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: [ "9.2.3"
```
michaelpj, line 83: I think we should also keep 9.2.3 here?

michaelpj (approved):

Issue #3089, opened 2022-08-11
Stan errors contains terminal color codes


andys8:
## Is your enhancement request related to a problem? Please describe.

Severity and Category use terminal color codes that aren't recognized in Vim/CoC. I assume that's also the case in other editors.

![image](https://user-images.githubusercontent.com/13085980/184101907-8bc89dbc-dab5-4be3-aabb-59cd6cda636b.png)

## Describe the solution you'd like

* Either allowing `stan` itself to configure/detect non interactive mode or skip coloring
* Or filter out color codes in the stan hls plugin

## Additional context

Currently using `a89ba412293327ec838c00d899ea9c6013c46aa3`.

```shell
haskell-language-server version: 1.7.0.0 (GHC: 9.2.4) (PATH: /home/andreas/.ghcup/bin/haskell-language-server-wrapper-1.7.0.1)
Tool versions found on the $PATH
cabal:          3.8.1.0
stack:          2.7.5
ghc:            9.2.4
```

In case the local stan installation is relevant:

```
Stan v0.0.0.0
 ➤ Git revision: 4fcf2e15d067f652c753272202fb33ef4200ccad
 ➤ Commit date:  Thu Jun 25 18:40:45 2020 +0100
```

Pinging @uhbif19

andys8:
It's easier than I initially thought. Pretty functions are adding terminal colors and are directly used. Replacing them with show instances solves the issue.

* https://github.com/kowainik/stan/blob/dcbc8682a964ae6f642ae37adbd46f6545d39aa6/src/Stan/Category.hs#L36-L38
* https://github.com/kowainik/stan/blob/dcbc8682a964ae6f642ae37adbd46f6545d39aa6/src/Stan/Severity.hs#L74-L76
* 
Opened #3090

andys8:
Solved by #3090

PR #3090, opened 2022-08-11
Stan: Avoid terminal colors in messages

andys8:
Solves #3089

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3090"></a>

### DIFF
--- a/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
+++ b/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
@@ -43,12 +43,10 @@ import           Ide.Types                      (PluginDescriptor (..),
                                                  defaultPluginDescriptor)
 import qualified Language.LSP.Types             as LSP
 import           Stan.Analysis                  (Analysis (..), runAnalysis)
-import           Stan.Category                  (prettyShowCategory)
 import           Stan.Core.Id                   (Id (..))
 import           Stan.Inspection                (Inspection (..))
 import           Stan.Inspection.All            (inspectionsIds, inspectionsMap)
 import           Stan.Observation               (Observation (..))
-import           Stan.Severity                  (prettyShowSeverity)
 
 descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
 descriptor recorder plId = (defaultPluginDescriptor plId) {pluginRules = rules recorder}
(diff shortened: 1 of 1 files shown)


michaelpj:
Build fails with 8.10.7, I think because that forces an older version of `stan` that doesn't expose `unCategory`?

uhbif19 (approved):

uhbif19:
@andys8 Stan does not work with GHC >8.10 (https://github.com/kowainik/stan/issues/423).

I does not understand how it seems to work with 9.2 in your case.

andys8:
Maybe it's just skipped in that case?

https://github.com/andys8/haskell-language-server/blob/f274919bccc02d139295dedfc5f2e4ec8db313fd/plugins/hls-stan-plugin/hls-stan-plugin.cabal#L26-L29

michaelpj:
Yeah, I think that means it was just unconditionally broken. There is a small mystery in that your original report said you were using GHC 9.2 which shouldn't have been possible!

andys8:
> There is a small mystery in that your original report said you were using GHC 9.2 which shouldn't have been possible!

Ah, I can help solve the mystery. The system ghc version is indeed 9.2, and I think also hls is build with it.

The project where the screenshot is from is a stack project using 8.10.7 though.

Still the output of `--probe-tools` in that directory looks like this. It also says "found on $PATH".

```
haskell-language-server-wrapper --probe-tools
haskell-language-server version: 1.7.0.0 (GHC: 9.2.4) (PATH: /home/andreas/.ghcup/bin/haskell-language-server-wrapper-1.7.0.1)
cabal:          3.8.1.0
stack:          2.7.5
ghc:            9.2.4
```

michaelpj:
Hmm, interesting. So for stack users this is potentially misleading, we really want to report the GHC that stack would use for the project or something.

andys8:
Is the failing benchmark check related to this pull request?

image

And as a followup to the "ghc mystery" we could add the output of `stack ghc -- --numeric-version` to the output of `--probe-tools`, see https://github.com/haskell/haskell-language-server/pull/3093

pepeiborra:
The bench failure is unrelated - our CI fails to upload the cache artifacts! I don't know why this happens, retried and hopefully it will go through but long term it would be great if someone could figure it out

PR #3091, opened 2022-08-11
Remove exactprint dependencies from ghcide by introducing hls-refactor-plugin.

wz1000:
All code actions have been moved to hls-refactor-plugin

Mostly straightforward, only slight complication was that completion auto imports
depends on exactprint, but I didn't want to remove all completion logic from ghcide
just for this.

Instead, I added logic to dynamically lookup the plugin that provides
the extend import command, so that auto imports work as expected when you have
hls-refactor-plugin enabled.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3091"></a>


pepeiborra (approved):
Blind approve

wz1000:
@pepeiborra you might want to look the changes to `showAstData` in `ghcide/src/Development/IDE/GHC/Dump.hs` and see if you are comfortable with those. We could add a flag to use exactprint for that function.

wz1000:
Also the async tests now use code lenses instead of code actions, I believe this still preserves the expected test behavior, is that correct @pepeiborra?

ghcide/src/Development/IDE/GHC/Dump.hs
```diff
@@ -21,18 +21,18 @@ import           GhcPlugins
...
         [
+-- #if MIN_VERSION_ghc(9,2,1)
+
+-- #else
 #if MIN_VERSION_ghc(9,2,1)
-            li (pre $ text (exactPrint a0)),
-            li (showAstDataHtml' a0),
+--             li (pre $ text (exactPrint a0)),
+--             li (showAstDataHtml' a0),
+--             li (nested "Raw" $ pre $ showAstData NoBlankSrcSpan NoBlankEpAnnotations a0)
```
fendor, line 35: Why are these still comments? Should probably be removed, if we are not going to use them anymore?

ghcide/src/Development/IDE/GHC/Dump.hs
```diff
@@ -21,18 +21,18 @@ import           GhcPlugins
...
-#else
 showAstDataHtml :: (Data a, Outputable a) => a -> SDoc
-#endif
 showAstDataHtml a0 = html $
     header $$
     body (tag' [("id",text (show @String "myUL"))] "ul" $ vcat
         [
+-- #if MIN_VERSION_ghc(9,2,1)
+
+-- #else
```
fendor, line 31: Remove?

ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -306,9 +306,9 @@ traceAst lbl x
         writeFile htmlDumpFileName $ renderDump htmlDump
         return $ unlines
             [prettyCallStack callStack ++ ":"
-#if MIN_VERSION_ghc(9,2,0)
-            , exactPrint x
-#endif
+-- #if MIN_VERSION_ghc(9,2,0)
+            -- , exactPrint x
+-- #endif
```
fendor, line 311: remove as well?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -241,6 +243,13 @@ allLspCmdIds pid commands = concatMap go commands
   where
     go (plid, cmds) = map (mkLspCmdId pid plid . commandId) cmds
 
+
+-- | Lookup the plugin that exposes a particular command
+lookupPluginId :: CommandId -> IdePlugins a -> Maybe PluginId
+lookupPluginId cmd (IdePlugins ls) = fst <$> find go ls
+  where
+    go (_, desc) = cmd `elem` map commandId (pluginCommands desc)
```
fendor, line 251: This looks inefficient, I hope this doesn't end up in any hot-loop.

fendor (approved):
Awesome, I think this split is great! Will make it much easier to get initial support for new GHC Versions!

I have some minor nitpicks, thank you very much for your work!

isovector (approved):
Tactics stuff lgtm

hls-plugin-api/src/Ide/PluginUtils.hs
wz1000: Good catch, thanks. I've made it so that it will only be called once per completion call.

ghcide/src/Development/IDE/GHC/Dump.hs
wz1000: I want to get @pepeiborra's opinion on this first.

wz1000:
@pepeiborra I've also replaced code actions in the benchmarks with code lenses. This is unfortunate, but we should fix this in the future by benchmarking the entire IDE with plugins included instead of just `ghcide`.

pepeiborra:
I'll take a proper look over the weekend

michaelpj:
I don't *love* the idea of dynamically looking up another plugin just in order to get access to some helper code. I think that could do with some clear documentation and a big warning saying not to use it unless you really really have to. I'm not sure what the "right" solution is: probably moving the completions logic to a plugin and pulling out a common dependency?

wz1000:
> I don't _love_ the idea of dynamically looking up another plugin just in order to get access to some helper code. I think that could do with some clear documentation and a big warning saying not to use it unless you really really have to. I'm not sure what the "right" solution is: probably moving the completions logic to a plugin and pulling out a common dependency?

It isn't really a function call, it decides whether to respond with an LSP `Command` depending on whether or not a plugin that provides the correct command is enabled.

michaelpj:
Also there's still an inverse dependency, in that `extendImportCommandId` is still defined in `ghcide`. So this won't work as a way of composing plugins together unless you have a central registry of the possible command ids to look for.

I'm not sure what the alternative is though, strings or something :/

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -217,9 +218,9 @@ mkCompl
             "line " <> printOutputable (srcLocLine loc) <> ", column " <> printOutputable (srcLocCol loc)
 
 
-mkAdditionalEditsCommand :: PluginId -> ExtendImport -> Command
-mkAdditionalEditsCommand pId edits =
-  mkLspCommand pId (CommandId extendImportCommandId) "extend import" (Just [toJSON edits])
+mkAdditionalEditsCommand :: Maybe PluginId -> ExtendImport -> Maybe Command
+mkAdditionalEditsCommand (Just pId) edits = Just $ mkLspCommand pId (CommandId extendImportCommandId) "extend import" (Just [toJSON edits])
```
pepeiborra, line 222: Would it make sense to introduce an abstraction for this dynamic lookup now?
Something like:
```
genLspCommand :: IdePlugins a -> CommandId -> Maybe (Text -> ... -> Command)
genLspCommand = fmap (mkLspCommand ...) (lookupPluginId ...)
```

hls-plugin-api/src/Ide/PluginUtils.hs
pepeiborra: Why not create an inverted index `(CommandId -> PluginId)`? Persisted inside `IdePlugins`

wz1000:
All the plugins that use the `GetAnnotatedParsedSource` rule now have to link it in via `pluginRules` to ensure they have access to it even if hls-refactor-plugin is disabled.

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -5,6 +5,14 @@
...
 {-# LANGUAGE RankNTypes             #-}
 {-# LANGUAGE TypeFamilies           #-}
+{-# LANGUAGE LambdaCase             #-}
+{-# LANGUAGE TupleSections          #-}
+{-# LANGUAGE ViewPatterns           #-}
+{-# LANGUAGE ScopedTypeVariables    #-}
+{-# LANGUAGE RecordWildCards        #-}
+{-# LANGUAGE TypeApplications       #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DeriveGeneric #-}
```
pepeiborra, line 15: I presume that these were all listed in the `ghcide.cabal` descriptor? I would prefer to copy that approach for `hls-refactor` too, these long lists of common extensions are silly.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -4,26 +4,41 @@
...
+{-# LANGUAGE BangPatterns          #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE ViewPatterns          #-}
+{-# LANGUAGE RecordWildCards       #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE TupleSections         #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE DuplicateRecordFields #-}
```
pepeiborra, line 20: 🤡

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -135,29 +158,117 @@ iePluginDescriptor plId =
...
+
+extendImportCommand :: PluginCommand IdeState
+extendImportCommand =
+  PluginCommand (CommandId extendImportCommandId) "additional edits for a completion" extendImportHandler
+
+extendImportHandler :: CommandFunction IdeState ExtendImport
+extendImportHandler ideState edit@ExtendImport {..} = do
+  res <- liftIO $ runMaybeT $ extendImportHandler' ideState edit
+  whenJust res $ \(nfp, wedit@WorkspaceEdit {_changes}) -> do
+    let (_, List (head -> TextEdit {_range})) = fromJust $ _changes >>= listToMaybe . Map.toList
```
pepeiborra, line 208: It would be good to copy over the hlint config from ghcide

pepeiborra (approved):
> look the changes to showAstData in `ghcide/src/Development/IDE/GHC/Dump.hs`

I would have extracted this function and `traceAst` too

> Also the async tests now use code lenses instead of code actions, I believe this still preserves the expected test behavior

Yes!

> I've also replaced code actions in the benchmarks with code lenses. This is unfortunate, but we should fix this in the future by benchmarking the entire IDE with plugins included instead of just ghcide

We should probably benchmark every plugin in isolation.

> If we are serious about composing plugins together, this is behaviour we probably want to support, so that plugins can pass off tasks to each other in such cases.

I agree

> Also there's still an inverse dependency, in that extendImportCommandId is still defined in ghcide. So this won't work as a way of composing plugins together unless you have a central registry of the possible command ids to look for.

This is only because completions are still inside ghcide. Once they are extracted to their own plugin, the dependency will be `hls-completions` -> `hls-refactor`. But this would mean `hls-completions` cannot be built until `hls-refactor` has been ported, so we would probably want to put the command id inside a new package `hls-refactor-command-ids`. Which can be done right now. @michaelpj what do you think?

> All the plugins that use the GetAnnotatedParsedSource rule now have to link it in via pluginRules to ensure they have access to it even if hls-refactor-plugin is disabled.

Yeah, this makes sense, just keep in mind that `his-graph` only deduplicates rules, not rules actions (created with `action :: Action a -> Rules a`).

michaelpj:
> But this would mean hls-completions cannot be built until hls-refactor has been ported, so we would probably want to put the command id inside a new package hls-refactor-command-ids. Which can be done right now. @michaelpj what do you think?

Well, I think we should think about what a world where we do this more often looks like. I think we'd have to have a package where all the command IDs for all plugins are defined. That wouldn't be so crazy: that's the "interface definition" for a plugin in this case, and having a separate interface definition which calling modules depend on isn't that unusual. I think I'd rather have one such package than one per plugin, though...

It's still quite ugly in that it means that we end up with a bit of logical coupling between all our plugins in this location. I don't have any better ideas, though. I would be happiest if we could just avoid doing it :sweat_smile: So I would prefer `hls-completions` -> `hls-refactor`, or better `hls-completions`/`hls-refactor` -> `hls-plugin-utils` or whatever. But that's significantly more work.

hls-plugin-api/src/Ide/PluginUtils.hs
wz1000: I've done this now.

ghcide/src/Development/IDE/GHC/Dump.hs
```diff
@@ -21,18 +21,18 @@ import           GhcPlugins
...
 #if MIN_VERSION_ghc(9,2,1)
-            li (pre $ text (exactPrint a0)),
-            li (showAstDataHtml' a0),
+--             li (pre $ text (exactPrint a0)),
```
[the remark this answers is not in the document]
wz1000: Done.

wz1000:
> so we would probably want to put the command id inside a new package hls-refactor-command-ids. Which can be done right now. @michaelpj what do you think?

I will do this in another MR. I think this one is ready to merge.

Issue #2358, opened 2021-11-16
HLS can't find symlinked Modules


fendor:
### Steps to reproduce

Project structure
```
.
├── _ghci
│   └── Sym.hs -> /home/hugin/Documents/haskell/hls-bug2/some_loc/Sym.hs
├── hie.yaml
├── some_loc
│   └── Sym.hs
└── src
    └── Foo.hs
```

where hie.yaml is:
```yaml
cradle:
  direct:
    arguments:
      - -i
      - -isrc
      - -i_ghci/
      - _ghci/Sym.hs
      - src/Foo.hs
```

and `Foo.hs` imports `Sym.hs`. 

The following script generates this project:
<details>
<summary>Generate project</summary>

```sh
#!/usr/bin/env bash

mkdir _ghci some_loc src
echo -e "module Foo where\n\nimport Sym\n" > src/Foo.hs
echo -e "module Sym where foo = \"\"" > some_loc/Sym.hs
ln -s $(pwd)/some_loc/Sym.hs _ghci/Sym.hs

echo "" > hie.yaml # create empty hie.yaml
echo "cradle:" >> hie.yaml
echo "  direct:" >> hie.yaml
echo "    arguments:" >> hie.yaml
echo "      - -i" >> hie.yaml
echo "      - -isrc" >> hie.yaml
echo "      - -i_ghci/" >> hie.yaml
echo "      - _ghci/Sym.hs" >> hie.yaml
echo "      - src/Foo.hs" >> hie.yaml
```
</details>

### Expected behaviour

HLS should be able to load this project, just like `ghc`:
```sh 
> ghc -i -isrc -i_ghci/ _ghci/Sym.hs src/Foo.hs
[1 of 2] Compiling Sym              ( _ghci/Sym.hs, _ghci/Sym.o )
[2 of 2] Compiling Foo              ( src/Foo.hs, src/Foo.o )
```

and `hie-bios`:
```sh
> hie-bios check src/Foo.hs
src/Foo.hs:3:1:Warning: The import of ‘Sym’ is redundant
  except perhaps to import instances from ‘Sym’
To import instances alone, use: import Sym()
```

### Actual behaviour

However, the following error is shown:
```sh
> haskell-language-server-8.10.7 src/Foo.hs
[...]

PR #3092, opened 2022-08-11
Fix #3047

### LINKED ISSUE #3047: incorrect suggestion for module name - missing first character of directory name
### Your environment

Which OS do you use?

debian 11 actualized
ghc 8.10.7 installed with ghcup
3.6.2.0 of the Cabal library 

I have in a directory Lib a file Example.hs in my project; it compiles without error.

In VScode I see a problem:

Set module name to `ib.Example` above the line
module Lib.Example

This is systematic and happens with all my modules.  

Version: 1.69.1
Commit: b06ae3b2d2dbfe28bca3134cc6be65935cdfea6a
Date: 2022-07-12T08:28:29.328Z
Electron: 18.3.5
Chromium: 100.0.4896.160
Node.js: 16.13.2
V8: 10.0.139.17-electron.0
OS: Linux x64 5.10.0-16-amd64

no  hie.yaml file - rely on automatic one

The Haskell Language server with the plugin for VScode works well - except it drops the first character of the directory for a suggested module name.

For example:

I have in a directory Lib a file Example.hs in my project; it compiles without error.

In VScode I see a problem:

Set module name to `ib.Example` above the line
module Lib.Example

This is systematic and happens with all my modules. Is there a problem in any of my setup files and which files should I check?

 

### Steps to reproduce

open a file, see the suggestion for a change in the module name.

 

### Expected behaviour

no suggestion - resp. a correct suggestion (which would then not be made). correct means with the first character of the directory name. (e.g. Lib, when directory is Lib and current suggestion is `ib`).

July541:
It is because of the difference between `hs-source-dirs: ./` and `hs-source-dirs: .`, take `/user/project/Lib/A.hs` as example,

| | `hs-source-dirs: ./` | `hs-source-dirs: .` |
| ---- | ---- | ---- |
| project path | /user/project/./ | /user/project/. |
| absolute project path | /user/project/ | /user/project |
| canonicalized project path |  /user/project | /user/project |

The canonicalized project path will be used as the prefix for `/user/project/Lib/A.hs`

Canonicalized it to prevent dropping one more character.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3092"></a>


plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
```diff
@@ -143,8 +160,8 @@ codeModuleName state nfp = runMaybeT $ do
   L (locA -> (RealSrcSpan l _)) m <- MaybeT . pure . hsmodName . unLoc $ pm_parsed_source pm
   pure (realSrcSpanToRange l, T.pack $ moduleNameString m)
 
--- traceAs :: Show a => String -> a -> a
--- traceAs lbl a = trace (lbl ++ " = " ++ show a) a
+newtype Log = ModuleNameLog T.Text deriving Show
```
fendor, line 163: If possible, you should rather use Named constructors instead of a generic "Text" field here, since we can process logs later on, disable some log messages and allow others. In my opinion, something like this would be much cleaner:

```suggestion
data Log
  = LogSourcePaths [FilePath] -- ^ I am not sure right now what the type is, just assuming fp for now.
  | LogModulePaths [FilePath]
  | LogCanonicalisedPaths [FilePath]
```

plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
```diff
@@ -118,22 +129,28 @@ action state uri =
...
+pathModuleNames recorder state normFilePath filePath
   | isLower . head $ takeFileName filePath = return ["Main"]
   | otherwise = do
       session <- runAction "ModuleName.ghcSession" state $ use_ GhcSession normFilePath
       srcPaths <- evalGhcEnv (hscEnvWithImportPaths session) $ importPaths <$> getSessionDynFlags
-      paths <- mapM makeAbsolute srcPaths
+      logWith recorder Debug (ModuleNameLog $ "ModuleName.srcpath: " <> T.pack (unlines srcPaths))
+
+      paths <- mapM canonicalizePath srcPaths
+      logWith recorder Debug (ModuleNameLog $ "ModuleName.paths: " <> T.pack (unlines paths))
```
fendor, line 141: Canonicalisation is, in my experience, too big of a hammer and behaves unexpectedly with symbolic links.

We basically removed its usage except for in tests and non-lsp usage. See https://github.com/haskell/haskell-language-server/pull/2359 and https://github.com/haskell/haskell-language-server/issues/2358 where we ran into issues. 

Would you mind looking into this test-case and check this patch doesn't cause any regressions?

fendor (changes requested):

fendor:
Excuse me for this drive-by review, I was just looking at it and found the usage of `canonicalizePath` might be tricky.

July541:
>  Excuse me for this drive-by review, I was just looking at it and found the usage of `canonicalizePath` might be tricky.

I'll try to evaluate the usage of `canonicalizePath`, at least find an alternative one.

plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
July541: I checked the project in #2358, it works fine for the current patch.

But to reduce concerns further, I'll try to investigate `normalise`, it seems workable.

plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
```diff
@@ -143,8 +163,20 @@ codeModuleName state nfp = runMaybeT $ do
...
+  deriving Show
+
+instance Pretty Log where
+    pretty log = "ModuleName." <> case log of
+        CorrectNames log     -> "CorrectNames" <> colon <+> pretty log
+        BestName log         -> "BestName" <> colon <+> pretty log
+        ModuleName log       -> "StatedNameMaybe" <> colon <+> pretty log
+        SrcPaths log         -> "SrcPaths" <> colon <+> pretty log
+        NormalisedPaths log  -> "NormalisedPaths" <> colon <+> pretty log
+        AbsoluteFilePath log -> "AbsoluteFilePath" <> colon <+> pretty log
```
fendor, line 182: Nitpick: inconsistent formatting, indent with 2 spaces?

fendor (approved):
LGTM, awesome!

PR #3093, opened 2022-08-11
Probe-tools: Print stack ghc version

andys8:
The ghc version on the $PATH is often not relevant for stack projects. The ghc version stack is using is printed in addition.

## Example Output

```
haskell-language-server version: 1.7.0.0 (GHC: 9.2.4) (PATH: [...]/haskell-language-server-wrapper) (GIT hash: 0e74593e8df9665067174ca27543006c9a6ad230)
cabal:          3.8.1.0
stack:          2.7.5
ghc:            9.2.4
stack ghc:      8.10.7
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3093"></a>

## Discussion

@michaelpj Followup to [this comment](https://github.com/haskell/haskell-language-server/pull/3090#issuecomment-1212146888) to sketch the idea for a change. Do you have something like this change in mind?

### DIFF
--- a/exe/Wrapper.hs
+++ b/exe/Wrapper.hs
@@ -14,7 +14,7 @@ module Main where
 import           Control.Monad.Extra
 import           Data.Char                          (isSpace)
 import           Data.Default
-import           Data.Either                        (fromRight)
+import           Data.Either.Extra                  (eitherToMaybe)
 import           Data.Foldable
 import           Data.List
 import           Data.Void
@@ -83,7 +83,7 @@ main = do
           putStrLn "Tool versions in your project"
           cradle <- findProjectCradle' False
           ghcVersion <- runExceptT $ getRuntimeGhcVersion' cradle
-          putStrLn $ "ghc:\t\t" ++ fromRight "Not found" ghcVersion
+          putStrLn $ showProgramVersion "ghc" $ mkVersion =<< eitherToMaybe ghcVersion
 
       VersionMode PrintVersion ->
           putStrLn hlsVer
--- a/exe/Wrapper.hs
+++ b/exe/Wrapper.hs
@@ -80,7 +80,7 @@ main = do
           putStrLn hlsVer
           putStrLn "Tool versions found on the $PATH"
           putStrLn $ showProgramVersionOfInterest programsOfInterest
-          putStrLn "Tool versions found by cradle"
+          putStrLn "Tool versions in your project"
           cradle <- findProjectCradle' False
           ghcVersion <- runExceptT $ getRuntimeGhcVersion' cradle
           putStrLn $ "ghc:\t\t" ++ fromRight "Not found" ghcVersion
--- a/exe/Wrapper.hs
+++ b/exe/Wrapper.hs
@@ -14,6 +14,7 @@ module Main where
 import           Control.Monad.Extra
 import           Data.Char                          (isSpace)
 import           Data.Default
+import           Data.Either                        (fromRight)
 import           Data.Foldable
 import           Data.List
 import           Data.Void
@@ -79,6 +80,10 @@ main = do
           putStrLn hlsVer
           putStrLn "Tool versions found on the $PATH"
           putStrLn $ showProgramVersionOfInterest programsOfInterest
+          putStrLn "Tool versions found by cradle"
+          cradle <- findProjectCradle' False
+          ghcVersion <- runExceptT $ getRuntimeGhcVersion' cradle
+          putStrLn $ "ghc:\t\t" ++ fromRight "Not found" ghcVersion
 
       VersionMode PrintVersion ->
           putStrLn hlsVer
--- a/src/Ide/Version.hs
+++ b/src/Ide/Version.hs
@@ -46,13 +46,17 @@ data ProgramsOfInterest = ProgramsOfInterest
 showProgramVersionOfInterest :: ProgramsOfInterest -> String
 showProgramVersionOfInterest ProgramsOfInterest {..} =
   unlines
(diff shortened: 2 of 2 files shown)


michaelpj:
Yeah, this looks nice! Do we think "stack ghc" is obvious enough? I kind of want to say "ghc provided by stack in this project" but that's a bit long.

fendor:
I like the idea in general, but I am worried `stack ghc` writes something stdout, killing LSP.

That's why run it in silent mode for https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L817

Also, one more issue, since we might trigger a download in this case, since the wrapper calls it: https://github.com/haskell/haskell-language-server/blob/master/exe/Wrapper.hs#L137

And if the command fails for any reason (e.g. can't install GHC because we are out of space), then we won't launch the hls-wrapper lsp service to report errors to the user and it will just crash.

andys8:
Thanks for the feedback @fendor . These definitely sound like valid points to consider. For my understanding: You mentioned standard output could kill the language server. Is the output and execution of `--probe-tools` relevant for the execution as language server via `--lsp`?

fendor:
No, it is not relevant, but truly great for debugging, since you tell people: Copy paste the logs from VSCode and link them. I suppose the issue also asks you to paste `haskell-language-server-wrapper --probe-tools`, but editors may affect how hls is launched, e.g. by modifying the PATH. I think, the only way to be sure to know which versions were used is to output them during execution as the lsp server.

andys8:
What do you have in mind as a solution to the mentioned problems?

* Calling stack in "silent mode": What did you mean by that?
* Not adding "stack ghc" at all because it depends on the directory again and could lead to the unwanted side effects?
* Only adding stack ghc to probe tools if hie has a stack cradle?
* Not using the same probe tools output in both places?


To repeat the core issue that should be addressed: For stack projects the users might post the output of `--probe-tools` in issues but it doesn't contain the ghc version used in the stack project.

fendor:
I am unfortunately not entirely sure how we can solve the issue here. I can imagine, there are multiple valid solutions.

*However*, I personally like your first/fourth suggestion. We can use `probeTools`, and if the cradle is a stack cradle (the respective call to it is just three lines later), we can add the compiler information. If the cradle is not a stack cradle, we can basically say that in the logs. That information is helpful as well.

andys8:
The more I start digging into the current state, the more I've the feeling I'm trying to reinvent something that already exists 😆

On hls launch there is a lot of debug output, including the ghc version of the cradle. Maybe it only needs to be moved around to be included in probe tools instead?

https://github.com/andys8/haskell-language-server/blob/511527c3e3e2160fe692159af947b814657e55b7/exe/Wrapper.hs#L125-L144

And there even is `haskell-language-server-wrapper --project-ghc-version`.

So reusing already existing functionality seems to make sense, but comes at the cost of adding the whole cradle business to the currently simple functionality in `Ide.Version`. While the two commands seem similar from the outside, the place where the information is available is quite deep inside the initialization. Still resusing seems to make sense 😉

fendor:
You are not strictly speaking re-inventing something, we are trying to show debug information at a stage, where we don't know the details yet.

The purpose of `probe-build-tools` is to show what tools and versions we are going to pick to discover the cradle type (e.g. the specifics of how to build our project). Found tools will influence what cradle we choose, e.g. if cabal or ghc are missing, but stack exists -> choose stack. The stack ghc version does not matter for that process.

The cradle type shows you, what has been chosen. Once we know that, we can print more specific information, such as stack ghc version. Before we know we are using a stack-cradle, it actually doesn't make sense at all to print the stack ghc version.

Thus, I'd make more sense to print cradle specific debug information. Then we have the following stages:
* Some details about how we are going to choose a cradle
* Actually choose the cradle type
* Details about a cradle, e.g. chosen ghc version, maybe stuff like that.

fendor:
I think the idea works well

exe/Wrapper.hs
```diff
@@ -79,6 +80,10 @@ main = do
           putStrLn hlsVer
           putStrLn "Tool versions found on the $PATH"
           putStrLn $ showProgramVersionOfInterest programsOfInterest
+          putStrLn "Tool versions found by cradle"
```
fendor, line 83:
```suggestion
          putStrLn "Tool versions in your project"
```

No average user knows what a cradle is, maybe we can avoid throwing vocabulary at users in this case.

exe/Wrapper.hs
```diff
@@ -79,6 +80,10 @@ main = do
...
+          cradle <- findProjectCradle' False
+          ghcVersion <- runExceptT $ getRuntimeGhcVersion' cradle
+          putStrLn $ "ghc:\t\t" ++ fromRight "Not found" ghcVersion
```
fendor, line 86: While you are at it, would you mind replacing `\t\t` with always the same amount of spaces (e.g., in Version.hs, too)? The tabulator doesn't render as I initially intended and it aligns baldy on some systems.

fendor (approved):
LGTM, minor comments

exe/Wrapper.hs
andys8: Would you prefer manually adding a handful of spaces for each line (depending on the string length) or can we drop indentation and use a single space everywhere?

fendor: I think it would be nicer if it was aligned, but I leave it up to you, since you are doing me the favour :)

andys8: Sure, changed it a bit to make sure the number of spaces is the same amount in both places: 6cfb60820ce25480310a34c7ca70546ce95f4076

fendor (approved):
LGTM, thanks!

PR #3103, opened 2022-08-13
Support fourmolu-0.8.1.0

brandonchinn178:
~Merging now: https://github.com/fourmolu/fourmolu/pull/226~

~This should be all that's required, but I'll test when released~

Tested that it builds locally with fourmolu 0.8.1

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3103"></a>


pepeiborra (approved):
Thanks!

PR #3107, opened 2022-08-15
Remove unused config in hls-class-plugin

July541:
While looking on #3105, I found type lens itself has a config filed, see `haskell-language-server generate-default-config`

Remove a duplicate one, although I don't know where is the another one from now.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3107"></a>

### DIFF
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class.hs
@@ -13,8 +13,6 @@ descriptor recorder plId = (defaultPluginDescriptor plId)
     , pluginRules = rules recorder
     , pluginHandlers = mkPluginHandler STextDocumentCodeAction (codeAction recorder)
         <> mkPluginHandler STextDocumentCodeLens codeLens
-    , pluginConfigDescriptor =
-        defaultConfigDescriptor { configCustomConfig = mkCustomConfig properties }
     }
 
 commands :: PluginId -> [PluginCommand IdeState]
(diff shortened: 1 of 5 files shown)


michaelpj:
I'm slightly confused, I think we probably *do* want a config field for the lenses? Or are there currently two or something?

July541:
Currently we have two config fields for lens,  `haskell-language-server generate-default-config` shows
```
"class": {
                "codeActionsOn": true,
                "codeLensOn": true,
                "config": {
                    "typelensOn": true
                }
            },
```

Turns out `codeLensOn` and `typelensOn` are the same one.

michaelpj:
Huh, do we generate a generic config option for any plugin that provides code lenses? That's pretty cool.

July541:
> Huh, do we generate a generic config option for any plugin that provides code lenses? That's pretty cool.

Obviously, it does! Although I don't know where the code is due to time limitation.

michaelpj:
@berberman ?

michaelpj:
Dumb question: does the other config option work? Can we keep the test and use the generic config option? Or maybe we can trust it, I'm just suspicious.

July541:
Both `codeLensOn` and `typelensOn` are available.

>  Can we keep the test and use the generic config option?

It's better if we can. I tried to find how to write the config, but failed as I said.

> Or maybe we can trust it

It's ok I think, all of our plugins don't have tests for main ability.

berberman:
> @berberman ?

Yes: https://github.com/haskell/haskell-language-server/blob/e55004a0cc9b727423a16cec4f5103a4df685eb6/hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs#L84-L94

So I think we no longer need the `typelensOn` in this case.

michaelpj:
We really need to document how this stuff works...

michaelpj (approved):

July541:
We should have a tech doc to land our tech details while developing hls, but not sure what should be included yet.

PR #3111, opened 2022-08-17
Do not send Heap Stats to the LSP log

pepeiborra:
It interferes with the ability of lsp-test to detect timeouts in e.g. benchmarks

The assumption is that logging them to stderr serves the purpose they
were designed for.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3111"></a>

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -64,12 +64,12 @@ main = do
               liftIO $ (cb1 <> cb2) env
           }
 
-    let (minPriority, logFilePath, includeExamplePlugins) =
+    let (argsTesting, minPriority, logFilePath, includeExamplePlugins) =
           case args of
             Ghcide GhcideArguments{ argsTesting, argsDebugOn, argsLogFile, argsExamplePlugin } ->
               let minPriority = if argsDebugOn || argsTesting then Debug else Info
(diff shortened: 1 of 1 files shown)


michaelpj:
Would it make sense to just disable the client logs when doing tests or benchmarks?

pepeiborra:
> Would it make sense to just disable the client logs when doing tests or benchmarks?

It's easy to do and I don't have a strong opinion

kokobd (approved):
> > Would it make sense to just disable the client logs when doing tests or benchmarks?
> 
> It's easy to do and I don't have a strong opinion

Then it looks good to me. Some users might be using this log as a quick measure to check memory consumption.

pepeiborra:
> > > Would it make sense to just disable the client logs when doing tests or benchmarks?
> 
> > 
> 
> > It's easy to do and I don't have a strong opinion
> 
> 
> 
> Then it looks good to me. Some users might be using this log as a quick measure to check memory consumption.
> 
> 
> 
> 
> 
> 

This PR doesn't remove the log from stderr, only from the LSP stream. We only started emitting those around 3 months ago so I doubt any users depend on them.

michaelpj:
I just think it makes our logging less consistent, which is confusing in future. But :shrug:

PR #3112, opened 2022-08-17
Deduplicate HLS plugins

pepeiborra:
Use a smart constructor to prevent duplicated plugins. We cannot use a
set since order matters for plugin execution.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3112"></a>

### DIFF
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -61,7 +61,7 @@ import           Data.Dependent.Map              (DMap)
 import qualified Data.Dependent.Map              as DMap
 import qualified Data.DList                      as DList
 import           Data.GADT.Compare
-import           Data.List.Extra (nubOrdOn)
+import           Data.List.Extra                 (nubOrdOn)
 import           Data.List.NonEmpty              (NonEmpty (..), toList)
 import qualified Data.Map                        as Map
 import           Data.Maybe
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -12,6 +12,7 @@
 {-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE NamedFieldPuns             #-}
 {-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE PatternSynonyms            #-}
 {-# LANGUAGE PolyKinds                  #-}
 {-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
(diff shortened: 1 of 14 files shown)


fendor (approved):
LGTM!

wz1000:
I think it would be better to use a Map, and then use an explicit priority (a simple `Int`) for ordering rather than implicitly depending on the order that we construct the list.

Perhaps we also want a richer graph structure in the future so we can express dependencies between plugins, but lets just go for explicit priorities for now.

pepeiborra:
> I think it would be better to use a Map, and then use an explicit priority (a simple `Int`) for ordering rather than implicitly depending on the order that we construct the list.

Agreed, explicit priorities are much better. Added now

wz1000 (approved):

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -76,9 +109,15 @@ import           Text.Regex.TDFA.Text            ()
 
 -- ---------------------------------------------------------------------
 
-newtype IdePlugins ideState = IdePlugins
-  { ipMap :: [(PluginId, PluginDescriptor ideState)]}
-  deriving newtype (Monoid, Semigroup)
+newtype IdePlugins ideState = IdePlugins_ { ipMap_ :: HashMap PluginId (PluginDescriptor ideState)}
+  deriving newtype (Semigroup, Monoid)
+
+-- | Smart constructor that deduplicates plugins
```
michaelpj, line 115: Is this comment accurate now? Or does it now give them in priority order?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -113,6 +152,8 @@ instance Show (IdeCommand st) where show _ = "<ide command>"
 data PluginDescriptor (ideState :: *) =
   PluginDescriptor { pluginId           :: !PluginId
                    -- ^ Unique identifier of the plugin.
+                   , pluginPriority     :: Natural
```
michaelpj, line 155: What happens if two plugins have the same priority?

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -42,6 +43,9 @@ descriptor :: Recorder (WithPriority Log) -> PluginDescriptor IdeState
 descriptor recorder = (defaultPluginDescriptor pluginId)
     { pluginRules = fixityRule recorder
     , pluginHandlers = mkPluginHandler STextDocumentHover hover
+    -- Make this plugin has a lower priority than ghcide's plugin to ensure
+    -- type info display first.
```
michaelpj, line 47: This is identifying a tricky interaction with `combineResponses`: since responses are combined in the same order that plugin handlers are called, having a higher priority means *both* that you get called first, *and* that your response appears first when we combine responses. We should document this, it's important!

Indeed, do we actually care about the order in which handlers are called, or is really only the combination of responses that we care about? Perhaps we should rather be having plugin handlers return a response + a priority, rather than associating it with the plugin? (thinking out loud here)

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
```diff
@@ -14,28 +14,30 @@ module Ide.Plugin.Pragmas
...
+import qualified Language.LSP.VFS                   as VFS
+import qualified Text.Fuzzy                         as Fuzzy
 
 -- ---------------------------------------------------------------------
 
 descriptor :: PluginId -> PluginDescriptor IdeState
 descriptor plId = (defaultPluginDescriptor plId)
   { pluginHandlers = mkPluginHandler J.STextDocumentCodeAction codeActionProvider
                   <> mkPluginHandler J.STextDocumentCompletion completion
+  , pluginPriority = ghcideCompletionsPluginPriority + 1
```
michaelpj, line 40: worth explaining

hls-plugin-api/src/Ide/Types.hs
pepeiborra: It still deduplicates. The comment is correct but incomplete

hls-plugin-api/src/Ide/Types.hs
pepeiborra: Unspecified order as decided by unordered-containers

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
pepeiborra: The ghcide-core plugin must run last'ish as it restarts the current build. Plug-ins can have side effects, so order is going to matter. This is not a nice pattern, so it would be good to offer pure alternatives. If anyone wants to contribute a better way to combine responses, I would be supportive.

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
pepeiborra: Argh, I forgot adding a line there.

PR #3116, opened 2022-08-20
Do not collect Hlint diagnostics when plugin is disabled

pepeiborra:
Fixes an oversight since actions in build rules are not implicitly gated in the same way that plugin handlers are.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3116"></a>


pepeiborra:
Abandoning as this is already working as expected

PR #3117, opened 2022-08-20
HLS benchmarks

pepeiborra:
Port the ghcide benchmark suite to HLS, adding support to benchmark plugin "configurations" independently.

This includes the following changes to the ghcide benchmark suite:
- Support for "configurations" which are defined as sets of plugin ids.
  The benchmark will be run with only these plugins enabled and all others disabled
- Support for configurable concurrency. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3117"></a>


haskell-language-server.cabal
```diff
@@ -366,13 +360,12 @@ common brittany
     build-depends: hls-brittany-plugin ^>= 1.0
     cpp-options: -Dhls_brittany
 
-executable haskell-language-server
+library plugins
```
fendor, line 363: this might make it impossible to use stack for loading HLS into HLS. I.e. HLS needs to be configured to use cabal for development with this change due to https://github.com/commercialhaskell/stack/issues/4564

pepeiborra: Finally time to drop the stack descriptors and the Circle CI

michaelpj:
Great!

Is there a reason not to delete the `ghcide` benchmarking code? Would make the diff simpler too, since I assume the benchmark runner fails are copies of the existing `ghcide` ones, and then it would just be a move.

pepeiborra:
> Great!
> 
> 
> 
> Is there a reason not to delete the `ghcide` benchmarking code? Would make the diff simpler too, since I assume the benchmark runner fails are copies of the existing `ghcide` ones, and then it would just be a move.

I am actively deleting it. But maybe I missed some code.

pepeiborra:
This is ready for review. CI is setup to run benchmarks for the plugin configurations defined in `bench/config.yaml`:

- *None*: no plugins
- *All*
- *Ghcide*: just the ghcide plugins
- *Core*: Ghcide plugins + call hierarchy + code range + eval + pragmas. This is an arbitrary selection
- *<plugin>*: configurations for each individual plugin, excluding formatting plugins since none of the benchmark experiments exercise `STextDocumentFormatting`.

https://github.com/haskell/haskell-language-server/blob/78ffc49c23eefa05226f6172739d99c3e523932e/bench/config.yaml#L83-L161

The results show that *hls-hlint-plugin* is, unsurprisingly, the most expensive plugin, followed closely by Tactics.

There is certainly margin for improvement:
1. Generated configurations: currently the list of configurations needs to be manually updated when new plugins are added (or removed). 
2. Conditional configurations: the 9.2 run is benchmarking a lot of plugins that are not included in the 9.2 build of HLS. These configurations are equivalent to the None configuration.
3. Sharding: currently it takes CI ~4h to run all the configurations for the Cabal example, ideally we would split the work across multiple machines.

bench/Main.hs
```diff
@@ -0,0 +1,282 @@
+
```
pepeiborra, line 1: `git` didn't notice, but this comes from `ghcide/bench/hist/Main.hs`

pepeiborra:
Example successful runs:

- https://github.com/haskell/haskell-language-server/runs/7937224383?check_suite_focus=true
- https://github.com/haskell/haskell-language-server/runs/7937224416?check_suite_focus=true

bench/Main.hs
pepeiborra, line 1: This file is a rename of `ghcide/bench/hist/Main.hs`, but for some reason Github doesn't show it

michaelpj:
> I am actively deleting it. But maybe I missed some code.

Excellent. There are still some files in `ghcide/bench`, though?

pepeiborra:
> > I am actively deleting it. But maybe I missed some code.
> 
> 
> 
> Excellent. There are still some files in `ghcide/bench`, though? 

The ghcide test suite references them, so a bit more work is required to move them

pepeiborra:
I took a look and `ghcide-bench` uses code from the ghcide test suite (`D.I.Test.Diagnostic`) that is quite tricky to extract. The path would be:  
1. move all the ghcide support test code to `hls-test-utils`, 
2. move `ghcide-bench` to its own package. 

Part 1 is a bit scary, I don't know how much work it will be.

michaelpj:
> The path would be:

Yeah, step 1 looks scary, especially because then the ghcide tests would depend on hls-test-utils which depends on ghcide? Not sure if that would work.

Would we need to do step 2? Wouldn't we just fold it into the HLS benchmark suite?

bench/README.md
```diff
@@ -1,22 +1,17 @@
...
-
-- `exe/Main.hs` - a standalone benchmark runner. Run with `stack run ghcide-bench`
-- `hist/Main.hs` - a Shake script for running the benchmark suite over a set of commits.
-  - Run with `stack bench ghcide` or `cabal bench ghcide`,
-  - Requires a `ghcide-bench` binary in the PATH (usually provided by stack/cabal),
-  - Calls `cabal` (or `stack`, configurable) internally to build the project,
-  - Driven by the `bench/config.yaml` configuration file.
-    By default it compares HEAD with "master"
+This folder contains a Shake script to simplify the performance analysis of HLS.
+It drives the `ghcide-bench` benchmark suite over a set of commits and experiments.
```
michaelpj, line 5:
```suggestion
It drives the `haskell-language-server:benchmark` benchmark suite over a set of commits and experiments.
```

bench/README.md
```diff
@@ -1,22 +1,17 @@
...
+To run it, use `cabal bench`.
+To configure it, edit `bench/config.yaml`.
+By default it compares HEAD with "origin/master"
 
 # Examples and experiments
 
 The benchmark suites runs a set of experiments (hover, completion, edit, etc.)
 over all the defined examples (currently Cabal and lsp-types). Examples are defined
-in `ghcide/bench/config.yaml` whereas experiments are coded in `ghcide/bench/lib/Experiments.hs`.
+in `bench/config.yaml` whereas experiments are coded in `ghcide/bench/lib/Experiments.hs`.
```
michaelpj, line 14: I guess it's true that the experiments are still defined in ghcide...

bench/config.yaml
```diff
@@ -90,27 +75,87 @@ versions:
...
+  - hlint
+  - importLens
+  - moduleName
+  - ormolu
+  - pragmas
+  - qualifyImportedNames
+  - refineImports
+  - rename
+  - stylish-haskell
+- alternateNumberFormat
```
michaelpj, line 129: weird indentation? I'm confused about this tbh.

exe/Main.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE CPP               #-}
 {-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PackageImports    #-}
```
michaelpj, line 6: dead?

michaelpj (approved):
Haven't looked at the actual benchmarking code, I trust you on that, just superficial comments.

michaelpj:
Looks like you're going ahead with a `ghcide-bench` package.

bench/config.yaml
pepeiborra: I'll add comments

haskell-language-server.cabal
```diff
@@ -412,8 +415,6 @@ executable haskell-language-server
     -- Enable collection of heap statistics
     "-with-rtsopts=-I0 -A128M -T"
     -Wno-unticked-promoted-constructors
-  if flag(pedantic)
```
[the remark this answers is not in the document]
pepeiborra: oops

test/functional/Diagnostic.hs
```diff
@@ -10,23 +10,8 @@ import           Test.Hls.Command
 -- ---------------------------------------------------------------------
 
 tests :: TestTree
-tests = testGroup "diagnostics providers" [
-        basicTests
-        , warningTests
-    ]
+tests = testGroup "diagnostics providers" [ warningTests ]
 
-basicTests :: TestTree
```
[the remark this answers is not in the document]
pepeiborra: We are losing the example plugins. There's hardly any point when we have a nice tutorial and 25? existing plugins that can serve as example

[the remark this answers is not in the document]
michaelpj: :+1: Might as well delete their source too then?

pepeiborra:
> Looks like you're going ahead with a `ghcide-bench` package.

Yeah, but I avoided the scary step 1 by duplicating some things from the ghcide test suite. Can be deleted when/if they are extracted to a helper package

fendor:
PR seems to have removed the example plugins, an accident?

PR #3120, opened 2022-08-23
Update hls-retrie-plugin to be usable with 9.2.4.

drsooch:
This is the first pass at getting hls-retrie-plugin enabled. Much of the
changes were updating to match the changes in the upstream `retrie`
package.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3120"></a>


drsooch:
@pepeiborra - Here's a first pass. Functionality seems to be the same. I don't see any major differences that would break anything.

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -267,13 +276,13 @@ describeRestriction restrictToOriginatingFile =
         if restrictToOriginatingFile then " in current file" else ""
 
 suggestTypeRewrites ::
-  (Outputable (IdP pass)) =>
+  (Outputable (IdP GhcRn)) =>
```
drsooch, line 279: I need to check again whether `GhcRn` is required. It was needed in a prior iteration.

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -416,12 +421,19 @@ callRetrie state session rewrites origin restrictToOriginatingFile = do
...
+        unsafeMkA (map (noLocA . toImportDecl) theImports) mempty 0
+#endif
 
   (originFixities, originParsedModule) <- reuseParsedModule origin
   retrie <-
     (\specs -> apply specs >> addImports annotatedImports)
       <$> parseRewriteSpecs
+#if MIN_VERSION_ghc(9,2,0)
+        libdir -- TODO: does this actualy get the proper libdir?
+#endif
```
drsooch, line 436: @pepeiborra this is the biggest question mark I have. I assume this should point to a user's libdir?

pepeiborra: This will get the HLS libdir, which will probably not work since it will point to a location in the build agent that compiled HLS...

pepeiborra: To get the user libdir, grab a GHC session and use https://hackage.haskell.org/package/ghc-9.4.2/docs/GHC-Driver-Session.html#v:topDir

plugins/hls-retrie-plugin/hls-retrie-plugin.cabal
```diff
@@ -25,6 +25,7 @@ library
     , extra
     , ghc
     , ghcide                ^>=1.7
+    , ghc-paths
```
pepeiborra, line 28: you don't want this

pepeiborra (approved):
Thank you so much for helping with this!

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
drsooch: Yeah thought as much, didn't know if ghc-paths used some trickery to get the actual path.

drsooch:
Is the intended operation for `getParsedModule` to fire multiple times? It looks like when a file contains CPP it will continuously run `getParsedModule`, until for whatever reason it will stop and finally create the command. 

Here's a sample output for running `unfold in the current file` on `qualify` in `Retrie.hs`: https://gist.github.com/drsooch/f63e9a45e6c17a4229c7e85bd2b79746

The `PARSED` and other lines, are just me trying to trace the logic. In the gist, you can see my memory usage explodes to ~10GB.

I compared with current master branch, and it appears this continuous parsing is a by-product of the file itself. So it doesn't look like this is a new issues.

pepeiborra:
> Is the intended operation for `getParsedModule` to fire multiple times? It looks like when a file contains CPP it will continuously run `getParsedModule`, until for whatever reason it will stop and finally create the command.
> 
> Here's a sample output for running `unfold in the current file` on `qualify` in `Retrie.hs`: https://gist.github.com/drsooch/f63e9a45e6c17a4229c7e85bd2b79746
> 
> The `PARSED` and other lines, are just me trying to trace the logic. In the gist, you can see my memory usage explodes to ~10GB.
> 
> I compared with current master branch, and it appears this continuous parsing is a by-product of the file itself. So it doesn't look like this is a new issues.

Yes, this is how `retrie` deals with CPP files. The memory usage is unfortunate, maybe there's room for improvement there

pepeiborra (approved):

kokobd (approved):
`hls-code-range-plugin` looks good to me.

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
drsooch: It is

drsooch:
Looks like there's some files that snuck through to master without formatting. Those are cleaned up here.

PR #3131, opened 2022-09-01
Reformat repository to clean up failing CI step

drsooch:
The pre-commit action appears to be failing. I did this in a separate PR but that didn't seem to catch everything. This PR should clean up all formatting problems so we can stop failing builds.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3131"></a>


michaelpj (approved):

drsooch:
@wz1000 to get stylish-haskell to parse properly, I've had to move some of the CPP blocks around. I tried to verify everything's good, but if you wouldn't mind double-checking.

kokobd:
pre-commit as a CI workflow was added one month ago. By default, it runs on `--all-files`. Maybe we can simply change its args to let it only run on changed files, like what we did before locally?

PR #3132, opened 2022-09-01
Enable support for 9.4 on windows

wz1000:
- Revert "drop windows for 9.4.2"
- Fix entropy to 0.4.1.10 to fix build on windows

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3132"></a>

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -51,6 +51,7 @@ write-ghc-environment-files: never
 index-state: 2022-08-15T06:53:13Z
 
 constraints:
+  entropy == 0.4.1.10,
   hyphenation +embed,
   -- remove this when hlint sets ghc-lib to true by default
   -- https://github.com/ndmitchell/hlint/issues/1376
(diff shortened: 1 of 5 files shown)


cabal.project
```diff
@@ -51,6 +51,7 @@ write-ghc-environment-files: never
 index-state: 2022-08-15T06:53:13Z
 
 constraints:
+  entropy == 0.4.1.10,
```
michaelpj, line 54: please add a comment explaining why this is here and when we can remove it

wz1000:
@guibou thanks for your help, but it seems like the Nix build is still failing, with a much worse error this time. I guess the last patch got merged because the Nix job is not required to merge.

cabal.project
wz1000: I've relaxed it to a lower bound and added a comment.

fendor (approved):
LGTM!

PR #3140, opened 2022-09-05
unescape printable characters

### LINKED ISSUE #3115: Unreadable unicode in type-level literals
### Your environment

* ghcup-compiled LSP from a recent-ish master
* VS Code extension v2.2.1
* GHC 9.2.4
* Ubuntu

### Steps to reproduce

```haskell
type family (label :: Symbol) ::: a where
  label ::: a = a

type Example = "привет" ::: Int

x :: Example
x = 4
```

Hover over `4` to see its type.

### Expected behaviour

The label should be preserved since it's a valid Haskell symbol: `_ :: "привет" ::: Int`.

### Actual behaviour

The label escaped into `_ :: "\1087\1088\1080\1074\1077\1090" ::: Int`.

Capture d’écran du 2022-08-19 15-19-08

kokobd:
This should fix #3115. It works. I will add tests later if the idea is okay.
Maybe a more proper fix is to modify the `Outputable` instance for `IfaceTyLit` in GHC?

Demo:
image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3140"></a>

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Util.hs
+++ b/ghcide/src/Development/IDE/GHC/Util.hs
@@ -81,6 +81,7 @@ import           GHC.IO.Exception
 import           GHC.IO.Handle.Internals
 import           GHC.IO.Handle.Types
 import           GHC.Stack
+import           Ide.PluginUtils                   (unescape)
 import           System.Environment.Blank          (getEnvDefault)
 import           System.FilePath
 import           System.IO.Unsafe
@@ -292,5 +293,5 @@ instance Outputable SDoc where
 --
 -- It internal using `showSDocUnsafe` with `unsafeGlobalDynFlags`.
 printOutputable :: Outputable a => a -> T.Text
-printOutputable = T.pack . printWithoutUniques
+printOutputable = unescape . T.pack . printWithoutUniques
 {-# INLINE printOutputable #-}
--- a/ghcide/src/Development/IDE/GHC/Util.hs
+++ b/ghcide/src/Development/IDE/GHC/Util.hs
@@ -289,9 +289,13 @@ instance Outputable SDoc where
 
 -- | Print a GHC value in `defaultUserStyle` without unique symbols.
 --
--- This is the most common print utility, will print with a user-friendly style like: `a_a4ME` as `a`.
+-- This is the most common print utility, and it will print with a user-friendly style like: `a_a4ME` as `a`.
 --
--- It internal using `showSDocUnsafe` with `unsafeGlobalDynFlags`.
+-- It uses `showSDocUnsafe` with `unsafeGlobalDynFlags` internally.
 printOutputable :: Outputable a => a -> T.Text
-printOutputable = unescape . T.pack . printWithoutUniques
+printOutputable =
+    -- IfaceTyLit from GHC.Iface.Type implements Outputable with 'show'.
+    -- Showing a String escapes non-ascii printable characters. We unescape it here.
+    -- More discussion at https://github.com/haskell/haskell-language-server/issues/3115.
+    unescape . T.pack . printWithoutUniques
 {-# INLINE printOutputable #-}
--- a/ghcide/src/Development/IDE/GHC/Util.hs
+++ b/ghcide/src/Development/IDE/GHC/Util.hs
@@ -288,10 +288,13 @@ instance Outputable SDoc where
 #endif
 
 -- | Print a GHC value in `defaultUserStyle` without unique symbols.
---
--- This is the most common print utility, and it will print with a user-friendly style like: `a_a4ME` as `a`.
---
 -- It uses `showSDocUnsafe` with `unsafeGlobalDynFlags` internally.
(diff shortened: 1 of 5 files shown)


michaelpj:
Is there not a more appropriate printing method from GHC we can use? Or are we just only given ones that escape?

kokobd:
> Is there not a more appropriate printing method from GHC we can use?

I'm not aware of one. Note that `SDoc` already contains the escaped string, so `SDocContext` and `OutputableP` won't help.

michaelpj:
Then I think the approach seems reasonable so long as we document it.

kokobd:
> Then I think the approach seems reasonable so long as we document it.

Done.

ghcide/src/Development/IDE/GHC/Util.hs
```diff
@@ -288,9 +289,13 @@ instance Outputable SDoc where
 
 -- | Print a GHC value in `defaultUserStyle` without unique symbols.
 --
--- This is the most common print utility, will print with a user-friendly style like: `a_a4ME` as `a`.
+-- This is the most common print utility, and it will print with a user-friendly style like: `a_a4ME` as `a`.
```
michaelpj, line 292: Maybe also include the fact that it unescapes characters? That's quite an important difference from other outputable functions. In fact... maybe we should use hlint to ban use of other outputable functions apart from this one?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -255,3 +260,35 @@ pluginResponse :: Monad m => ExceptT String m a -> m (Either ResponseError a)
...
+    case P.runParser escapedTextParser "inline" input of
+        Left _     -> input
+        Right strs -> T.pack strs
+
+-- | Parser for a string that contains double quotes. Returns unescaped string.
+escapedTextParser :: TextParser String
+escapedTextParser = do
+    xs <- P.many (P.try stringLiteral)
+    x <- P.manyTill P.anySingle P.eof -- consume characters after the final double quote
+    pure $ concat xs ++ x
```
michaelpj, line 282: It feels a little weird that you handle the content before the opening quote in `stringLiteral` but the content after the closing quote outside it. Maybe make that consistent?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -255,3 +260,35 @@ pluginResponse :: Monad m => ExceptT String m a -> m (Either ResponseError a)
...
+  where
+    stringLiteral :: TextParser String
+    stringLiteral = do
+        before <- P.manyTill P.anySingle (P.char '"') -- include any character before the first double quote
+        inside <- P.manyTill P.charLiteral (P.char '"')
```
michaelpj, line 287: What about `"hell\"ooo"`? i.e. escaped quotes probably shouldn't stop us

michaelpj: Hmm, from the tests it looks like you made this work but I'm not sure why it works...

hls-plugin-api/test/Ide/PluginUtilsTest.hs
```diff
@@ -1,13 +1,35 @@
...
 import           Test.Tasty
 import           Test.Tasty.HUnit
 
 tests :: TestTree
 tests = testGroup "PluginUtils"
-    [
+    [ unescapeTest
+    ]
+
+unescapeTest :: TestTree
```
michaelpj, line 19: :100:

michaelpj (approved):
Mostly just some nits.

hls-plugin-api/src/Ide/PluginUtils.hs
kokobd: `P.charLiteral` calls `lookAhead` internally. And this is an example from the haddock of `charLiteral`:
```
stringLiteral = char '"' >> manyTill L.charLiteral (char '"')
```

ghcide/src/Development/IDE/GHC/Util.hs
kokobd: > Maybe also include the fact that it unescapes characters?

Done.

> maybe we should use hlint to ban use of other outputable functions apart from this one?

TBH I don't know if this is configurable. I think the scheme is to use `printOutputable` by default, and also allow other print functions to be used when needed.

hls-plugin-api/src/Ide/PluginUtils.hs
kokobd: I rewrote `stringLiteral` to not include any content outside of the double quotes. I think that's more reasonable.

ghcide/src/Development/IDE/GHC/Util.hs
michaelpj: I mean, we can use `hlint` to ban functions, so we could do it. I don't know if it's a good idea, though.

PR #3143, opened 2022-09-06
Skip benchmarks when method is not enabled by plugin

wz1000:


wz1000:
This doesn't quite work because it turns out that we advertise capabilities even when all plugins that could provide support for a method are disabled, because we construct the map of `Handlers` that we send to `lsp` too early, before we have a chance to consult the config.

pepeiborra:
This is a nice idea, shame that it doesn't work!

PR #3145, opened 2022-09-07
Only run the pre-commit hook on changed files

drsooch:
Using an external action to grab all changed files, and pass that to the pre-commit check to reduce false-negatives.

### NOTES:

 Alternatively, could pass the arguments: `pre-commit ... --from-ref HEAD^ --to-ref HEAD` to run the action on the last commit. However,  will only capture a single commit, even if the branch received multiple commits on a push.

I'm not even sure this `get-diff-action` will even do what I expect.

**TESTING NOTES**:
1. first commit with no haskell source changes, worked as expected. Skipped all the steps when there were no diffs.
2. Committing a single haskell source change is matched.
3. Diffs across multiple commits are captured (looks like the `get-diff` action compares the PR to main branch so this will always be the case)
4. probably need to split the `GIT_DIFF` variable to pass files appropriately to pre-commit - done

This seems to work as expected  I refactored a bit to skip the pre-commit hook if there are no hs files in the changelog. (not that it saves any time).

This should enable us to be green for pre-commit check

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3145"></a>


Ailrun (commented):
Does `--from-ref origin/master` (or similar using `github.base_ref`) work? If so, I think it might be better to use that for pr, and for merged commit, we can use `HEAD^`.

drsooch:
> Does `--from-ref origin/master` (or similar using `github.base_ref`) work? If so, I think it might be better to use that for pr, and for merged commit, we can use `HEAD^`.

I believe the current 3rd party GitHub action does a simple compare to the branch we are merging to.

I guess it could just get tricky with all possible enumerations of which commit to count towards the change list. Pre-commit itself even suggested the following: `--from-ref HEAD^^^ --to-ref HEAD`.

Really what I was going for is anything new, we want to make sure we try to keep formatted. And didn't want to try to cover everything.

Ailrun:
> I guess it could just get tricky with all possible enumerations of which commit to count towards the change list.

I mean, if pre-commit supports target branch names like origin/master for `--from-ref`, it's not really hard, is it? Or you mean that pre-commit doesn't support that?

> I believe the current 3rd party GitHub action does a simple compare to the branch we are merging to.

IMO, if pre-commit already supports the feature, it'd be better not to introduce more dependency (which will be maintenance burden).

drsooch:
> > I guess it could just get tricky with all possible enumerations of which commit to count towards the change list.
> 
> I mean, if pre-commit supports target branch names like origin/master for `--from-ref`, it's not really hard, is it? Or you mean that pre-commit doesn't support that?

The docs I was reading, didn't use a specific ref like `origin/master` it used  `--from-ref HEAD^^^ --to-ref HEAD`. I can experiment with using a specific ref like master.


> 
> > I believe the current 3rd party GitHub action does a simple compare to the branch we are merging to.
> 
> IMO, if pre-commit already supports the feature, it'd be better not to introduce more dependency (which will be maintenance burden).

Definitely agree, I will try some other experiments when I can

PR #3149, opened 2022-09-12
support add-argument action

santiweight:
This action supports an action "Add argument `xxx` to function" whenever there is an undefined variable error:

```Haskell
foo = unknown
-- becomes
foo unknown = unknown
```

The only minor wrinkle is that

```Haskell
foo = True
   where
      bar = unknown
-- becomes
foo unknown = True
   where
      bar = unknown
```

But this is good enough for now as a matter of pragmatism.

As part of this MR, I fixed some behavior. Previously an unknown-type undefined variable did not offer a "define x" option, but now does:

```Haskell
foo = unknown
-- previously no code actions
-- now becomes
foo = unknown

unknown :: _
unknown = _
```

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3149"></a>


kokobd:
Thanks for your contribution!

For reference, below are the behavior of some other language servers/IDEs on this:
- rust-analyzer in VSCode: no quickfix
- TypeScript in VSCode: no quickfix
- Java in Intellij IDEA Community: has this quickfix, but provides a dialog to modify function signature
<details>
<summary>screenshots</summary>
image
image
image
image
</details>

I'm not quite sure about the design, maybe someone else will provide better insight.

santiweight:
Thanks for the screenshots! Very helpful context :) You note that Rust/TS have no quick fix. I assume you mean this as a positive right?

I am quite familiar with the IntelliJ interface from my previous job. I would love to support this workflow, but after I looked around the HLS and LSP issues I got the impression that such a pretty workflow would require too much work to justify. Is that you guys' experience? I really think our UX could be best in class given Haskell's type system, but I'm concerned about maintenance hell.

On a side note, it would be really great to have some sort of function:

```Haskell
alterAllUsages :: {- function -} RdrName -> (LHsExpr -> TransformT m LHsExpr) -> IDEState -> TransformT m IDEState
```

The idea being that when I add a function parameter to a definition, I can insert appropriate holes in all usages project wide. Does anyone have any thoughts on the feasibility of this? @OliverMadine seems like a candidate for such knowledge!

```Haskell
module One where

import Two

bar = foo True
----
module Two where

foo True = unknown False

----- BECOMES -----
module One where

import Two

bar = foo True _unknown
----
module Two where

foo True unknown = unknown False
```

OliverMadine:
> The idea being that when I add a function parameter to a definition, I can insert appropriate holes in all usages project wide. Does anyone have any thoughts on the feasibility of this? @OliverMadine seems like a candidate for such knowledge!

Yes, it seems very feasible to me!
Most of the logic could be extracted from the rename plugin. It would, however, be subject to the same multi-component limitation of the rename plugin (#2193)

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -623,6 +653,14 @@ eqSrcSpanA l r = leftmost_smallest l r == EQ
 #endif
 
 #if MIN_VERSION_ghc(9,2,0)
+
+spanContainsRange :: SrcSpan -> Range -> Bool
```
kokobd, line 657: Maybe you can use [`realSrcSpanToRange`](https://github.com/haskell/haskell-language-server/blob/163ebfe9d02d5aa9f2c83ef8d9db2e91cde31032/ghcide/src/Development/IDE/GHC/Error.hs#L76) and [`isSubrangeOf`](https://github.com/haskell/lsp/blob/a41954e356117d8f5cb64b20e1a9d32e8fc6a884/lsp-types/src/Language/LSP/Types/Location.hs#L89) ?

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -430,6 +434,32 @@ graftDecls dst decs0 = Graft $ \dflags a -> do
...
     modifyDeclsT (pure . DL.toList . go) a
 
+modifySmallestDeclWithM ::
+    forall a.
+    (HasDecls a) =>
+    (SrcSpan -> Bool) ->
+    (LHsDecl GhcPs -> TransformT (Either String) (Maybe [LHsDecl GhcPs])) ->
+    a ->
+    TransformT (Either String) a
+modifySmallestDeclWithM validSpan f a = do
```
pepeiborra, line 444: could you add Haddock docs for this and other new functions please?

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -623,6 +653,14 @@ eqSrcSpanA l r = leftmost_smallest l r == EQ
...
 
 #if MIN_VERSION_ghc(9,2,0)
+
+spanContainsRange :: SrcSpan -> Range -> Bool
+spanContainsRange srcSpan Range {..} =
+  srcSpan `spans` positionToTuple _start && srcSpan `spans` positionToTuple _end
+  where
+    positionToTuple :: Position -> (Int, Int)
+    positionToTuple (Position l c) = (fromIntegral l + 1, fromIntegral c)
+
```
pepeiborra, line 663: Searching for this type signature (flipped) within the codebase, I find 3 functions that seem to duplicate this logic.

Worth reusing one of them? 

Screenshot 2022-09-18 at 16 26 18

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -96,7 +96,9 @@ import           GHC                                               (AddEpAnn (Ad
                                                                     EpAnn (..),
                                                                     EpaLocation (..),
                                                                     LEpaComment,
-                                                                    LocatedA)
+                                                                    LocatedA, spans)
+import Language.Haskell.GHC.ExactPrint (runTransformFromT, noAnnSrcSpanDP1, runTransform, runTransformT)
+import GHC.Types.SrcLoc (generatedSrcSpan)
```
pepeiborra, line 101: You need to reformat this with stylish-haskell to pass the pre-commit-hook check

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -845,34 +848,93 @@ suggestReplaceIdentifier contents Diagnostic{_range=_range,..}
...
+          let newPat = L (noAnnSrcSpanDP1 generatedSrcSpan) $ VarPat NoExtField (noLocA unqualName)
+          pure $ L locMatch (Match xMatch ctxMatch (pats <> [newPat]) rhs)
+        insertArg = \case
+          (L locDecl (ValD xVal (FunBind xFunBind idFunBind mg coreFunBind))) -> do
+            mg' <- modifyMgMatchesT mg addArgToMatch
+            let decl' = L locDecl (ValD xVal (FunBind xFunBind idFunBind mg' coreFunBind))
+            pure $ Just [decl']
+          _ -> pure Nothing
+    case runTransformT $ modifySmallestDeclWithM (`spanContainsRange` range) insertArg (makeDeltaAst parsedSource) of
+      Left err -> error $ "Error when inserting argument: " <> err
```
pepeiborra, line 931: Avoid throwing an `error` here, as it will at best kill all other code actions and at worst kill all other plugins. Ideally you should return a `ResponseError` but that probably requires too much refactoring.

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
```diff
@@ -430,6 +434,32 @@ graftDecls dst decs0 = Graft $ \dflags a -> do
...
+    (HasDecls a) =>
+    (SrcSpan -> Bool) ->
+    (LHsDecl GhcPs -> TransformT (Either String) (Maybe [LHsDecl GhcPs])) ->
+    a ->
+    TransformT (Either String) a
+modifySmallestDeclWithM validSpan f a = do
+    let modifyMatchingDecl [] = pure DL.empty
+        modifyMatchingDecl (e@(L src _) : rest)
+            | validSpan $ locA src = do
+                decs' <- fromMaybe [e] <$> f e
```
pepeiborra, line 448: Why does `f` return a `Maybe`? Couldn't it simply return its input untouched?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -845,34 +848,93 @@ suggestReplaceIdentifier contents Diagnostic{_range=_range,..}
...
+          let unqualName = mkRdrUnqual $ mkVarOcc $ T.unpack name
+          let newPat = L (noAnnSrcSpanDP1 generatedSrcSpan) $ VarPat NoExtField (noLocA unqualName)
+          pure $ L locMatch (Match xMatch ctxMatch (pats <> [newPat]) rhs)
+        insertArg = \case
+          (L locDecl (ValD xVal (FunBind xFunBind idFunBind mg coreFunBind))) -> do
+            mg' <- modifyMgMatchesT mg addArgToMatch
+            let decl' = L locDecl (ValD xVal (FunBind xFunBind idFunBind mg' coreFunBind))
+            pure $ Just [decl']
+          _ -> pure Nothing
+    case runTransformT $ modifySmallestDeclWithM (`spanContainsRange` range) insertArg (makeDeltaAst parsedSource) of
```
pepeiborra, line 930: It would be sweet if the type signature was also extended if present

pepeiborra (changes requested):

drsooch:
There is a plug-in that attempts to update signatures when it can. `change-type-signature` It's pretty rudimentary and only works in a handful of cases where GHC provides enough information. When the upstream `lsp` package updates to the lsp spec 3.17. We should be able to get direct access to GHCs Error Message API and potentially be worthwhile to this case.

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
santiweight: I pulled this out into Development.IDE.GHC.Error, where similar functions are.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
santiweight: I totally agree. I've left a TODO above for the time being, but it'll be my priority after I finish this and a follow-up action.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
santiweight: Let's see if I was able to fix all the formatting issues...

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
santiweight: This was all a little silly. There was no need for the Maybe at all, so I reworked the passed-in function.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
santiweight: I'm unclear on what to do here. For the time being I used a trace call and returned no diffs. Is ResponseError part of the MonadLSP interface?

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
santiweight: Added some comments on both

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
santiweight: I was able to implement adding to the type signature here: https://github.com/santiweight/haskell-language-server/tree/add-arg-type-sig

It is rather complicated... consider altering multi-type-sig-decls of the form `foo, bar :: Int`. You have to split up the declarations.

I was able to make it work, but it'll take a fair amount of testing, and so I'd much rather keep this PR as an intermediate step

santiweight:
@drsooch I've thought about the change-type-sig plugin in this context. I think they're fairly different situations. In change-type-sig, I can only see occurrences where the new type signature is given. 

However, in this case, we do not have the full type signature: we only have the type of the argument we're adding. That means we do have to manually alter the HsSigType with the exact print api, which is far more complicated.

I implemented a working first attempt (it's surely very buggy for non-trivial layouts): https://github.com/santiweight/haskell-language-server/tree/add-arg-type-sig

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
fendor: Yes it is :)

santiweight: I summoned up the courage to reveal ResponseErrors in the CodeAction API. The only hold up is I don't know how to throw/report ResponseErrors with MonadLSP after some hoogling.

Anyone have any pointers? 😅 Note that I am returning a list of errors _and_ successes (since there are many potential code actions), so it should just be `logResponseError :: MonadLSP m => ResponseError -> m ()`, not throw

pepeiborra: I think it's fine to just use `logError` or `logWarn` and return nothing

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
pepeiborra: Fair

pepeiborra (approved):
Looks good to me modulo the build failures

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
santiweight: After some investigation, it appears that the only Recorder available to CodeActions is `Recorder (WithPriority E.Log)`, which is the logging from `Development.IDE.Core.Shake`. 

Which logger should I be using for code actions? I can't find a Logger or nonShake recorder in the parent functions...

plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs
santiweight: I forgot to respond. This worked perfectly thanks :)

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
pepeiborra: This is because this plugin was only recently extracted from ghcide core. The plugin needs to define its own Logger I guess.

michaelpj:
Can you add yourself to `CODEOWNERS`? You might also want to mention this in `features.md`.

santiweight:
It seems that the tests didn't run as part of CI? Besides that I've added myself to codeowners but I don't have write access, which is causing an error in the code owners file!

I'm happy to merge whenever now :) Thanks for the help getting this one through!

michaelpj:
> It seems that the tests didn't run as part of CI?

They should have, we're testing `hls-refactor-plugin` which should include the new stuff, right?

> Besides that I've added myself to codeowners but I don't have write access, which is causing an error in the code owners file!

I've invited you to the repo :)

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -881,34 +891,95 @@ suggestReplaceIdentifier contents Diagnostic{_range=_range,..}
         = [ ("Replace with ‘" <> name <> "’", [mkRenameEdit contents _range name]) | name <- renameSuggestions ]
     | otherwise = []
 
+matchVariableNotInScope :: T.Text -> Maybe (T.Text, Maybe T.Text)
```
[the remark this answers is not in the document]
michaelpj: Also, this module is also quite large, perhaps the add-action stuff could go in a separate module also?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -881,34 +891,95 @@ suggestReplaceIdentifier contents Diagnostic{_range=_range,..}
...
+#if MIN_VERSION_ghc(9,2,1)
+suggestAddArgument :: ParsedModule -> Diagnostic -> Either ResponseError [(T.Text, [TextEdit])]
+suggestAddArgument parsedModule Diagnostic {_message, _range}
+  | Just (name, typ) <- matchVariableNotInScope message = addArgumentAction parsedModule _range name typ
+  | Just (name, typ) <- matchFoundHoleIncludeUnderscore message = addArgumentAction parsedModule _range name (Just typ)
+  | otherwise = pure []
+  where
+    message = unifySpaces _message
+
+-- TODO use typ to modify type signature
```
michaelpj, line 960: are you going to do the TODO in this PR?

michaelpj (approved):
I took a fairly superficial look, but I think we can get this over the line :)

Thank you for digging into the code actions in `hls-refactor-plugin`. As Pepe mentioned, they only recently got extracted from `ghcide`, so they quite probably be tidied up more to be more in keeping with the style of more recent code action handlers if you felt moved to!

santiweight:
Thanks for the review @michaelpj 

A heads up on the status of this PR... I'm pretty happy with where it is right now, but I'm a little worried about my approach to using the Exactprint API, and so I'm going to attempt reimplementing the code in a better style (by following the examples in ghc-exactprint).

This MR isn't forgotten, just hopefully being replaced by something better...

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
santiweight: no - it's upcoming in a future PR :)

ghcide/src/Development/IDE/GHC/Error.hs
```diff
@@ -119,6 +121,9 @@ p `isInsideSrcSpan` r = case srcSpanToRange r of
   Just (Range sp ep) -> sp <= p && p <= ep
   _                  -> False
 
+spanContainsRange :: SrcSpan -> Range -> Bool
+spanContainsRange srcSpan range = maybe False (range `isSubrangeOf`) $ srcSpanToRange srcSpan
```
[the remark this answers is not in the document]
santiweight: I returned a Nothing and added a comment; I think it's clearer now...

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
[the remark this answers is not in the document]
santiweight: I actually already did this in a followup MR. Would you be okay with following up with this change (to avoid unnecessary conflicts)?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -881,34 +891,95 @@ suggestReplaceIdentifier contents Diagnostic{_range=_range,..}
...
-    | Just [name, typ] <- matchRegexUnifySpaces message "Found hole: _([^ ]+) :: ([^*•]+) Or perhaps"
-    , [(label, newDefinitionEdits)] <- newDefinitionAction ideOptions parsedModule _range name typ
-    = [(label, mkRenameEdit contents _range name : newDefinitionEdits)]
-    | otherwise = []
-    where
-      message = unifySpaces _message
+suggestNewDefinition ideOptions parsedModule contents Diagnostic {_message, _range}
+  | Just (name, typ) <- matchVariableNotInScope message =
+      newDefinitionAction ideOptions parsedModule _range name typ
+  | Just (name, typ) <- matchFoundHole message,
```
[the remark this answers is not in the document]
santiweight: Revised

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -2131,14 +2135,219 @@ insertNewDefinitionTests = testGroup "insert new definition actions"
...
+      liftIO $ contentAfterAction @?= T.unlines (txtB ++
+        [ ""
+        , "select :: _"
+        , "select = _"
+        ]
+        ++ txtB')
   ]
 
+#if MIN_VERSION_ghc(9,2,1)
+addFunctionArgumentTests :: TestTree
```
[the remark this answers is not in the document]
santiweight: I have already done this in a  followup change. Unfortunately fixing this now would cost me some time with conflicts, but I promise to follow up :)

santiweight:
I was hoping to upstream some of the changes here in ghc-exactprint before merging, but I think that process will take some time.

I've answered the comments from reviewers, and I now think this is ready to merge!

michaelpj:
I'm happy for you to continue improving things, especially since it sounds like you're definitely continuing to work in this area :)

PR #3155, opened 2022-09-14
Remove unused build-depends and install warnings

pepeiborra:


ghcide-bench/ghcide-bench.cabal
```diff
@@ -108,7 +108,7 @@ test-suite test
         implicit-hie:gen-hie
     main-is: Main.hs
     hs-source-dirs: test
-    ghc-options: -Wunused-packages
+    ghc-options: -Wunused-packages -Werror=unused-packages
```
michaelpj, line 111: I think we probably don't want the `Werror`? Or can we do this only in CI? any kind of `Werror` is a huge pain during development IMO

michaelpj: In fact, don't we already have a `pedantic` cabal flag that we set to enable `Werror` during CI?

michaelpj (approved):
Would be great to do this for the other packages too, but this is already good!

PR #3159, opened 2022-09-14
hls-hectare-plugin

pepeiborra:
Hectare is a type-driven code synthesis tool presented at ICFP by @jkoppel. @tritlo packaged it in a GHC plugin that packages the results as typed hole suggestions, and this is a simple HLS plugin that conveniently packages the GHC plugin. [Paper](https://arxiv.org/pdf/2206.07828.pdf)

The plugin is opt-in as it is disabled by default. This is because the Hectare search is triggered automatically for *all* typed holes as part of type checking, which means no diagnostics until the search is finished. To enable the plugin, you can set the following LSP property:
```
  "haskell.plugin.hectare.globalOn": true,
```
This will enable the Plugin immediately, no restart needed. However, Hectare seems to be quite buggy at the moment, check the HLS logs where exceptions will be dumped.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3159"></a>


jkoppel:
Correction: ECTAs are a new data structure and constraint solver for enumerative program synthesis, available from https://hackage.haskell.org/package/ecta . The tool is Hectare.

Tritlo:
I think then that the plugin I published might be misnamed. Since the package on Hackage is called `ecta`, I named the plugin the `ecta-plugin`. We could rename it `hectare-plugin`, but since it's already published to hackage, the old one will still be there. Let's at least make the name correct in HLS, @pepeiborra.

pepeiborra:
The plan here is to launch `hls-hectare-plugin` as an optional HLS plugin, disabled by default and enabled explicitly by the user. This requires #3193 to land first

jkoppel:
What's blocking this from getting merged?

pepeiborra:
https://github.com/haskell/haskell-language-server/pull/3193

docs/support/plugin-support.md
```diff
@@ -68,3 +68,4 @@ For example, a plugin to provide a formatter which has itself been abandoned has
 | `hls-stan-plugin`                   | 3    | 8.6, 9.0, 9.2, 9.4       |
 | `hls-retrie-plugin`                 | 3    | 9.2, 9.4                 |
 | `hls-splice-plugin`                 | 3    | 9.4                      |
+| `hls-hectare-plugin`                | 3    | 8.10.7                   |
```
michaelpj, line 71: Ah, maybe you got the semantics of this backwards: this is the list of _unsupported_ GHC versions.

haskell-language-server.cabal
```diff
@@ -241,6 +246,11 @@ common haddockComments
     build-depends: hls-haddock-comments-plugin ^>= 1.1
     cpp-options: -Dhls_haddockComments
 
+common hectare
+  if flag(hectare) && ((impl(ghc > 8.10) && impl(ghc < 9)) || flag(ignore-plugins-ghc-bounds))
```
michaelpj, line 250: Should this be `>= 8.10`?

plugins/hls-hectare-plugin/Ide/Plugin/Hectare.hs
```diff
@@ -0,0 +1,30 @@
...
+import           Data.Default
+import           Development.IDE.GHC.Compat
+import           Development.IDE.GHC.Compat.Util
+import qualified ECTA.Plugin                     as GHC
+import           Ide.Plugin.Config
+import           Ide.Types
+
+descriptor :: PluginId -> PluginDescriptor s
+descriptor pluginId = (defaultPluginDescriptor pluginId)
+  { pluginModifyDynflags = staticPlugin
```
michaelpj, line 12: So _all_ this does is enable the Hectare plugin? I'm a bit confused by this. I would expect that if users want a GHC plugin then they can just add it to their project themselves. It seems a a bit roundabout (and extra work for us) to have HLS plugin for every GHC plugin like that?

plugins/hls-hectare-plugin/hls-hectare-plugin.cabal
```diff
@@ -0,0 +1,30 @@
...
+license-file:       LICENSE
+author:             Pepe Iborra
+maintainer:         pepeiborra@gmail.com
+category:           Development
+build-type:         Simple
+extra-source-files:
+  LICENSE
+
+library
+  if impl(ghc >= 8.10) && impl(ghc < 9.0)
```
michaelpj, line 17: Given that we might drop support for 8.10.7 after the next major release, do we want to add plugins that only work on 8.10?

pepeiborra: Hectare itself should work for any GHC version, 
@Tritlo do you plan to upgrade the GHC plugin?  
cc @jkoppel

plugins/hls-hectare-plugin/Ide/Plugin/Hectare.hs
pepeiborra: This is a good question. The Hectare GHC plugin doesn't make programs type check, it does not add a static analysis pass, nor does it rewrite your program to make it faster. It is an IDE plugin, designed to help you write programs faster.

Why is it packaged as a GHC plugin instead of as an HLS plugin then? Well, I guess a GHC plugin is more reusable - anyone can import it in their project, not just HLS users. 

Why repackage it as an HLS plugin? 
 - For convenience, just as we repackage code formatters and linters. 
 - For discoverability, although that is a weaker argument since the plugin is disabled by default.

michaelpj: I think the comparison with formatters and linters is interesting:
- For formatters, you get a new way of interacting with the tool, including things like convenient range formatting
- For linters you get the diagnostics integrated with HLS

If we needed this plugin in order for the Hectare diagnostics to work I'd definitely be in favour - it's just that the Hectare developers have been so clever in getting their diagnostics into hole fit suggestions that we don't actually _need_ to do anything if the user turns on the plugin themselves. In no other case is the plugin actually _unnecessary_.

But I do take your argument about it being a pure IDE plugin. There wouldn't be much point enabling it in your cabal file, really!

All of which is to say: I don't know :shrug: so I guess if you're willing to maintain it we can include it!

pepeiborra:
> rebased the PR, ready to get merged now?

Haven't heard from @jkoppel and @Tritlo for a while, so I'll be closing this PR until further notice

jkoppel:
Matti and I have begun debugging.

PR #3163, opened 2022-09-14
Fixes the flake deps to align with cabal bounds

mjrussell:
Was trying to build HLS 1.8.0 with nix via the flake and noticed that the versions of the input deps were not in sync with the required cabal bounds. 

ghcide needs hie-bios-0.11.0 [here](https://github.com/haskell/haskell-language-server/blob/master/ghcide/ghcide.cabal#L108)

ghcide needs lsp-types-1.6.0.0 [here](https://github.com/haskell/haskell-language-server/blob/5a734dfb7f2ae12a1dc85a3bc0d0ef0a16125862/ghcide/ghcide.cabal#L72)

hls-plugin-api needs lsp-1.6.0.0 [here](https://github.com/haskell/haskell-language-server/blob/master/hls-plugin-api/hls-plugin-api.cabal#L52)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3163"></a>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,11 +21,11 @@
 
     # List of hackage dependencies
     lsp = {
-      url = "https://hackage.haskell.org/package/lsp-1.5.0.0/lsp-1.5.0.0.tar.gz";
+      url = "https://hackage.haskell.org/package/lsp-1.6.0.0/lsp-1.6.0.0.tar.gz";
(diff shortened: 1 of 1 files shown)


michaelpj:
It still doesn't work, but this is better.

PR #1235, opened 2021-01-19
Add code actions for disabling a warning in the current file

georgefst:
Previously known as haskell/ghcide#897.

Progress towards #705.

PR #3167, opened 2022-09-15
Refactor plugin: Prefer code action

andys8:
`_isPreferred` can influence the client side order of code actions. The idea is that an unused import is likely to be removed and less likely the warning will be disabled. Therefore actions to remove a single or all redundant imports should be preferred, so that the client can prioritize them higher.

Followup of <https://github.com/haskell/haskell-language-server/pull/3018>

## Demo

Here is an example in Vim/CoC. There is an import that isn't necessary. The list of code starts with "Disable unused-imports warnings", which is also the option that would be defaulted to. Instead removing the redundant import should be preferred and will influence the order (depends on the language server client).

image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3167"></a>


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -520,7 +520,7 @@ caRemoveInvalidExports m contents digs ctxDigs uri
         _documentChanges = Nothing
         _edit = Just WorkspaceEdit{..}
         _command = Nothing
-        _isPreferred = Nothing
+        _isPreferred = Just True
```
michaelpj, line 523: This is a policy decision. You made a good argument for why we should do this in the PR description, but all that information is going to be lost once this is merged. Write it down! I think this is a good place for a Note-style comment, e.g.

```
-- Note [Removing imports is preferred]
-- It's good to prefer the remove imports code action because...
...
   -- See Note [Removing imports is preferred]
   _isPreferred = Just True

andys8: > You made a good argument for why we should do this in the PR description, but all that information is going to be lost once this is merged

`git blame` on the line will show the explanation as part of the commit and refer to the pull request. I was thinking about a comment, but assumed we'll have many code actions that need to be adapted like this, and I thought comments would be two much. But if that's the preferred granularity and place for documentation, I'll add one here.


https://user-images.githubusercontent.com/13085980/190390388-72bfa8ee-97a5-4c96-a592-a9bc80c9ad25.mp4

georgefst:
When implementing https://github.com/haskell/haskell-language-server/pull/1235, I did think that the disable-warning actions should probably always be at the bottom of the list, but I couldn't work out how to do that. This looks good!

andys8:
@georgefst Cool, maybe we could (in a different PR) also set these code actions to `isPreferred = Just False` to explicitly mark them as not the preferred (default / go to) solution. But looks like there would be a bit of refactoring necessary in order to pass the information along.

https://github.com/haskell/haskell-language-server/pull/1235/files#diff-bf3e1eeb4b7bfd43c7bd24d10063ea6e23c0714d61ab5f5c62e7323ebf1282aaR231

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
michaelpj: Commit messages last until someone else touches that line. Plus, maybe someone wants to change it *and* change the reasoning. IMO there is no substitute for just writing things in the code!

michaelpj: More of my opinions if you care ;) https://www.michaelpj.com/blog/2022/04/24/on-commenting-code.html

andys8: Interesting, I can see arguments for both ways.

> Commit messages last until someone else touches that line.

"Put it in code" often comes from times where tooling was bad or didn't exist. Git blame / commits can be viewed before the change as well, and there is history and references to issue and PR. But this only works well for small incremental changes (and not large commits changing everything). So it depends on the project.

image

What I've seen more often is that code will be changed, but the comment will not be adapted and be misleading/wrong. In these cases newer commits overlaying older ones isn't necessarily bad.

The most important part is *that there is* is place where a decision is documented at all and it's somewhat consistently done in the project (format and granularity). Not really arguing to do anything differently here, just bringing up a different view 😀

pepeiborra:
Just to confirm, does this work in Vim/Coc? And have you tested with VSCode?

andys8:
Honestly, I don't know how all clients behave and I haven't tested in other editors, but I can tell what I know. That being said, the actual effect will likely depend on the client. But these changes add information of what the most reasonable choice of actions is. I think it would also make sense to mark all "disable something" actions as "not preferred", but often seems to need refactoring.

## Previous Change

I've tested a similar change in https://github.com/haskell/haskell-language-server/pull/3018, but this one not again. There are screenshots demonstrating the change in behavior with Vim and CoC.

## This Change

Tested this change by building it locally. Seems to show "Remove all redundant imports" before "Disable warning". Not sure why the "Remove import" is at the end, maybe they are created in different places.

image


## Language Server Protocol

LSP docs say:


 	 * A quick fix should be marked preferred if it properly addresses the
	 * underlying error. A refactoring should be marked preferred if it is the
	 * most reasonable choice of actions to take.
	 *
	 * @since 3.15.0

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/

## CoC Implementation

Here is the sorting implementation of coc (vim client). It seems to take the property into account.

https://github.com/neoclide/coc.nvim/blob/59fa3972e4395adf1852717011d57fbfa5c169b5/src/handler/codeActions.ts#L64-L70

andys8:
Rebased :heavy_check_mark:

michaelpj:
> I think it would also make sense to mark all "disable something" actions as "not preferred", but often seems to need refactoring.

I actually don't know if clients treat "no isPreferred field" differently from "isPreferred set to false". The LSP spec of course doesn't tell us anything like usual :roll_eyes: 

This LGTM, we've used this in the past for hlint and I think it's the right thing to do.

michaelpj (approved):

andys8:
@michaelpj I don't know either. In theory `isPreferred?: boolean` in type script allows 3 states because of its optionality: `true`, `false`, and `undefined`. But for example the [CoC implementation][coc] only checks for truthiness with `!b.isPreferred`. Therefore for CoC's current implementation `Just False` and `Nothing` in Haskell should result in the same behavior. But I don't know about other clients and if that's intended or random implementation.

It's a good point though. Adding `isPreferred: Just False` would be meaningful information, but we could end up not changing actual client behavior.

[coc]: https://github.com/neoclide/coc.nvim/blob/59fa3972e4395adf1852717011d57fbfa5c169b5/src/handler/codeActions.ts#L65

PR #3168, opened 2022-09-15
remove manual heap profiling from ghcide

pepeiborra:
The [memory profiling scheme](https://mpickering.github.io/ide/posts/2020-08-04-measuring-memory-size.html) was added by @mpardalos. It works by stopping the world and traversing the heap using the `heapsize` package, but it breaks if a GC starts halfway, leading to a segfault.

Nowadays we have a better tool for this sort of analysis, ghc-debug, which snapshots the heap and performs the analysis offline. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3168"></a>


fendor (approved):
LGTM! Doesn't it still have merits for debugging performance issues on ghc versions that have no ghc-debug support?

pepeiborra:
> (Is this PR misnamed?)

Oops. I have been using the VSCode GitHub extension to create PRs and maybe it automatically reuses prior titles

pepeiborra:
> LGTM! Doesn't it still have merits for debugging performance issues on ghc versions that have no ghc-debug support? Or doesnt this happen anyway?

Just use 9.2 or higher to analyse memory, there is no reason to support analysis in older versions

michaelpj:
Build failures on 9.4

PR #3133, opened 2022-09-01
nix: fix nix environment for GHC 9.4

guibou:
This is a work in progress, I'm trying to restore the build with nix.

- [X] restore `nix develop .\#haskell-language-server-941-dev` (light environment which provides cabal + ghc)
- [x] restore the build of all hackages dependencies with nix.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3133"></a>

PR #3175, opened 2022-09-15
Attempt to fully fix nix build

mjrussell:
Pins base-compat in flake inputs and update lock

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3175"></a>

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -48,6 +48,18 @@
         "url": "https://hackage.haskell.org/package/brittany-0.13.1.2/brittany-0.13.1.2.tar.gz"
       }
     },
+    "cereal": {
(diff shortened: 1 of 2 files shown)


mjrussell:
Ah my bad I see @guibou is already doing some of this and in likely a better approach

guibou:
> Ah my bad I see @guibou is already doing some of this and in likely a better approach

Sorry about that ;( I think my MR https://github.com/haskell/haskell-language-server/pull/3133 fixs most of the build.

mjrussell:
No worries! Glad someone is working on it 🙂

Issue #3157, opened 2022-09-14
Config: How to disable stan plugin


andys8:
Hey there, it's either currently not possible or I don't know how to exactly disable the stan plugin.

* Config is `"stan": { "globalOn": false }`
* [Docs](https://haskell-language-server.readthedocs.io/en/latest/configuration.html#generic-plugin-configuration) don't list `stan` plugin
* The default configuration (`haskell-language-server-wrapper generate-default-config`) doesn't have an entry for `stan`.

Is `globalOn` respected for this plugin?

![image](https://user-images.githubusercontent.com/13085980/190134161-025ae27a-eaba-4867-8fb6-9971a40b3fb0.png)

###  Vim

Still enabled, but expected to be off

![image](https://user-images.githubusercontent.com/13085980/190133625-629d1e75-d93e-4983-8ae2-218bd94a8be4.png)

### CLI

Still enabled, but expected to be off

![image](https://user-images.githubusercontent.com/13085980/190134550-4d9e4e45-f7a1-415b-854a-29e49aa087ab.png)

## Additional context

```
haskell-language-server-wrapper --probe-tools
haskell-language-server version: 1.8.0.0 (GHC: 9.2.4) (PATH: /home/andreas/.ghcup/bin/haskell-language-server-wrapper-1.8.0.0)
Tool versions found on the $PATH
cabal:          3.8.1.0
stack:          2.7.5
ghc:            9.2.4

Tool versions in your project
ghc:            8.10.7
```

kokobd:
I'm not familiar with the dynamic config, but you can add `--constraint "haskell-language-server -stan"` to your cabal command if you were building HLS from source. In this way, stan plugin will be disabled at compile time.

andys8:
I'm indeed compiling hls from scratch with `ghcup compile hls` and thanks for the hint. But I'm less interested in "fixing" this for me but to make sure configuration is working as expected for the project. Therefore I'm wondering if it can currently be disabled (like other plugins) and why it's missing in the documentation.

pepeiborra:
Have you tried `"haskell.plugin.span.globalOn": false` ?

andys8:
I'm starting to have doubts regarding the whole editor configuration. This way would not be the format `generate-default-config` suggests, but I'm having trouble to find any configuration that has an effect.

Could you point me to documentation showing how to configure Vim with Coc with plugin options?

[...]

PR #3179, opened 2022-09-16
Stan: Respect plugin configuration globalOn

andys8:
Respects `globalOn` in stan plugin configuration and won't show hints anymore if the plugin is disabled. Solves #3157

## Tested

```json
"stan": { "globalOn": false }
```

Could verify the `globalOn` setting will be respected with a local project.

## About the change

Solves the issue (respecting if the plugin is disabled) the same way it is done [in the hlint plugin][hlint], as hinted to by @pepeiborra in [this comment][2]  thanks

[hlint]: https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs#L199-L205
[2]: https://github.com/haskell/haskell-language-server/issues/3157#issuecomment-1247766075

@uhbif19 Since you wanted to take a look - hope it's okay and I ddin't take it away from you - is this fine? Any doubts? Also feel free to take over. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3179"></a>

### DIFF
--- a/plugins/hls-stan-plugin/hls-stan-plugin.cabal
+++ b/plugins/hls-stan-plugin/hls-stan-plugin.cabal
@@ -32,6 +32,7 @@ library
   build-depends:
       base
     , containers
+    , data-default
     , deepseq
     , hashable
     , hls-plugin-api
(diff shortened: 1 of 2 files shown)


pepeiborra (approved):
This will do, thanks!

michaelpj:
Indeed this looks good, @uhbif19 can tweak it afterwards if they like. Thanks!

PR #3180, opened 2022-09-16
Docs: List stan plugin

andys8:
Minor, but as mentioned in #3157 the stan plugin is not listed in list of existing plugins in the documentation. This change will add it.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3180"></a>


pepeiborra (approved):
We really should auto generate this and other lists of plugins.

PR #3186, opened 2022-09-17
Update supported GHC versions doc

michaelpj:


July541 (approved):
Thanks!

michaelpj:
Hang on, I missed one bit! Need to check which plugins are disabled for 9.4.

PR #3188, opened 2022-09-17
Fix broken call-hierarchy-plugin-tests for type signatures

July541:
I did some refactoring by the way. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3188"></a>


plugins/hls-call-hierarchy-plugin/src/Ide/Plugin/CallHierarchy/Internal.hs
```diff
@@ -91,98 +77,93 @@ patternBindInfo ctxs = listToMaybe [ctx       | ctx@PatternBind{} <- ctxs]
...
-    renderTyDecl = case ident of
-      Left _ -> Nothing
-      Right name -> case getNameBindingInClass name ssp (getAsts hf) of
-        Nothing -> Nothing
-        Just sp -> case resolveIntoCallHierarchy hf (realSrcSpanToRange sp ^. L.start) nfp of
-          Just (Just items) -> listToMaybe items
-          _                 -> Nothing
+        renderTyDecl = case ident of
+            Left _ -> Nothing
+            Right name -> case getNameBinding name (getAsts hf) of
```
July541, line 132: Here is the core.

plugins/hls-call-hierarchy-plugin/src/Ide/Plugin/CallHierarchy/Types.hs
```diff
@@ -13,36 +13,36 @@ data Vertex = Vertex {
...
 , occ    :: String
 , hieSrc :: FilePath
-, sl     :: Int
-, sc     :: Int
-, el     :: Int
-, ec     :: Int
+, sl     :: Int -- start line
+, sc     :: Int -- start character
+, el     :: Int -- end line
+, ec     :: Int -- end character
```
pepeiborra, line 19:
```suggestion
, sl     :: Int -- ^ start line
, sc     :: Int -- ^ start character
, el     :: Int -- ^ end line
, ec     :: Int -- ^ end character
```

pepeiborra (approved):
Looks reasonable

PR #3189, opened 2022-09-17
Add policy on plugin support tiers

michaelpj:
Following discussion in
https://github.com/haskell/haskell-language-server/discussions/3123.

This aims to make it more obvious what it means when we say we support a version of GHC.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3189"></a>


July541 (approved):
Very detailed description!

July541:
So, we can announce "fully" support of ghc-9.2 and drop the support of 8.6.5 and 8.8.4?

michaelpj:
> So, we can announce "fully" support of ghc-9.2 and drop the support of 8.6.5 and 8.8.4?

I'm not sure. We have _said_ that we won't do this until there is a 9.2 LTS stackage. I don't know if that's something we want to stick to, given that we're not using stack really for building or distribution any more. But we can discuss that separately, maybe at the HLS meeting?

pepeiborra:
> > So, we can announce "fully" support of ghc-9.2 and drop the support of 8.6.5 and 8.8.4?

> I'm not sure. We have said that we won't do this until there is a 9.2 LTS stackage. I don't know if that's something we want to stick to, given that we're not using stack really for building or distribution any more. But we can discuss that separately, maybe at the HLS meeting?

In case I don't make it to the HLS meeting, I am going to voice my concerns here:

1. We currently support and test GHC 8.6, 8.8, 8.10, 9.0, 9.2 and 9.4 
2. It is extremely expensive, in terms of code maintainability, CI, and release cost, to support a GHC version
3. In addition to user experience, we need to consider HLS sustainability as a fully volunteer-driven effort. The current situation with 6 GHC versions is imho leaning too far on the user side.

docs/support/plugin-support.md
```diff
@@ -0,0 +1,68 @@
...
+Hence a tier 3 plugin should ideally have some kind of plan for getting out of tier 3, either by getting the plugin to tier 2 or by sunsetting it.
+For example, a plugin to provide a formatter which has itself been abandoned has no hope of reaching tier 2, but may be gracefully sunset by only being supported for old versions of GHC, and deleted once those exit our GHC support window.
+
+## Current plugin support tiers
+
+| Plugin                              | Tier | Unsupported GHC versions |
+|-------------------------------------|------|--------------------------|
+| ghcide core plugins                 | 1    |                          |
+| `hls-call-hierarchy-plugin`         | 1    |                          |
+| `hls-code-ragne-plugin`             | 1    | 9.4                      |
```
pepeiborra, line 44:
```suggestion
| `hls-code-range-plugin`             | 1    | 9.4                      |
```

docs/support/plugin-support.md
```diff
@@ -0,0 +1,68 @@
...
+| `hls-qualify-imported-names-plugin` | 2    |                          |
+| `hls-ormolu-plugin`                 | 2    | 9.4                      |
+| `hls-rename-plugin`                 | 2    | 9.4                      |
+| `hls-refine-imports-plugin`         | 2    |                          |
+| `hls-stylish-haskell-plugin`        | 2    | 9.4                      |
+| `hls-tactics-plugin`                | 2    | 9.2, 9.4                 |
+| `hls-brittany-plugin`               | 3    | 9.2, 9.4                 |
+| `hls-haddock-comments-plugin`       | 3    | 9.2, 9.4                 |
+| `hls-stan-plugin`                   | 3    | 8.6, 9.0, 9.2, 9.4       |
+| `hls-retrie-plugin`                 | 3    | 9.2                      |
```
pepeiborra, line 67:
```suggestion
| `hls-retrie-plugin`                 | 3    | 9.2, 9.4                 |
```

docs/support/plugin-support.md
```diff
@@ -0,0 +1,68 @@
...
+Tier 3 plugins are maintained on a best-effort basis, often by irregular contributors.
+A plugin may have to be tier 3 despite being well-maintained if it depends on a tool (e.g. a formatter) which is not itself reliably updated for new GHC versions.
+
+Since we cannot make any guarantees that a tier 3 plugin will be working, they provide a bad experience for users.
+Hence a tier 3 plugin should ideally have some kind of plan for getting out of tier 3, either by getting the plugin to tier 2 or by sunsetting it.
+For example, a plugin to provide a formatter which has itself been abandoned has no hope of reaching tier 2, but may be gracefully sunset by only being supported for old versions of GHC, and deleted once those exit our GHC support window.
+
+## Current plugin support tiers
+
+| Plugin                              | Tier | Unsupported GHC versions |
```
pepeiborra, line 40: This table needs to be autogenerated, right? It should be not hard using the `cabal-syntax` package

pepeiborra (approved):
I can see how this is a good thing, but unless we automatically generate the plugins table, I can also see user issues in the future when the plugins table becomes outdated

docs/support/plugin-support.md
michaelpj: It's not currently, and we do have trouble keeping it up-to-date. But it's still useful information, and I volunteer to maintain it until/unless we get to a point that we can auto-generate it.

The tier information also is manual and I don't think could be emitted from the plugin. Conceivably we could record all this information in the plugins and then have a `haskell-language-server describe-plugins` command that dumped it out, and then do that for all versions...

docs/support/plugin-support.md
kokobd, line 44:
```suggestion
| `hls-code-ragne-plugin`             | 1    |                          |
```

It's supported.

michaelpj:
Let's give this a try at least.

PR #3193, opened 2022-09-18
Support optional plugins

pepeiborra:
Adds support for plugins that are not enabled by default

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3193"></a>


michaelpj:
This feels a little special-case-y. Doesn't the `Properties` machinery have a way to let the plugin set default values for its configuration? Could we extend that to cover the generic options somehow? @berberman ?

pepeiborra:
> This feels a little special-case-y. Doesn't the `Properties` machinery have a way to let the plugin set default values for its configuration? Could we extend that to cover the generic options somehow? @berberman ?

The `Properties` machinery is for plugin-specific options. For generic options, I think you will struggle to find a better way than this. What in it feels special-case-y?

michaelpj:
> What in it feels special-case-y?

In that we have to add a new special field to plugin descriptors just to control the default value of one of the generic plugin options. What if we want to, say, have a plugin for which code lenses are off by default? Add another field? Then again, it's not the end of the world to generalize later...

pepeiborra:
> In that we have to add a new special field to plugin descriptors just to control the default value of one of the generic plugin options. What if we want to, say, have a plugin for which code lenses are off by default? Add another field? Then again, it's not the end of the world to generalize later...

That's a good point, I have generalised the scheme to cover all the generic options.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -170,6 +170,8 @@ instance Show (IdeCommand st) where show _ = "<ide command>"
 data PluginDescriptor (ideState :: *) =
   PluginDescriptor { pluginId           :: !PluginId
                    -- ^ Unique identifier of the plugin.
+                   , pluginConfig       :: !PluginConfig
```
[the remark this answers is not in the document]
pepeiborra: This is a good question - I completely forgot about the custom config stuff. It probably makes sense to consolidate all the plugin config in a single place, will check.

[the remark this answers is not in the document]
michaelpj: It's a field of `PluginConfig`, so in principle you could set the defaults using this mechanism instead... but the other one is nice because it's typed, I guess. I don't really know how this stuff works properly.

pepeiborra:
Third attempt, let's see if this is the winner.

hls-plugin-api/src/Ide/Types.hs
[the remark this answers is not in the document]
pepeiborra: Me neither, but it seems well thought out...

pepeiborra:
It would be good to get this landed while I still remember the details...

hls-plugin-api/src/Ide/Plugin/Config.hs
```diff
@@ -79,7 +90,18 @@ parseConfig defValue = A.withObject "Config" $ \v -> do
         <*> (o .:? "checkProject" <|> v .:? "checkProject") .!= checkProject defValue
         <*> o .:? "formattingProvider"                      .!= formattingProvider defValue
         <*> o .:? "maxCompletions"                          .!= maxCompletions defValue
-        <*> o .:? "plugin"                                  .!= plugins defValue
+        <*> explicitParseFieldMaybe (parsePlugins $ plugins defValue) o "plugin" .!= plugins defValue
+
+parsePlugins :: Map.Map T.Text PluginConfig -> Value -> A.Parser (Map.Map T.Text PluginConfig)
+parsePlugins defValue = A.withObject "Config.plugins" $ \o -> do
+  let -- parseOne :: Key -> Value -> A.Parser (T.Text, PluginConfig)
```
michaelpj, line 97: any reason to not just leave the signature in?

hls-plugin-api/src/Ide/Plugin/Config.hs
```diff
@@ -54,18 +61,22 @@ data Config =
     , plugins            :: !(Map.Map T.Text PluginConfig)
     } deriving (Show,Eq)
 
-instance Default Config where
-  def = Config
+-- | Default configuration values
```
michaelpj, line 64: Maybe expand this comment a bit to explain the general approach? Given that we were both complaining about finding this bit hard to understand, maybe let's try to make the new stuff easy to follow :) perhaps something like:

```
Default configuration values. The default values for plugins can be passed by the caller so that plugin definitions can provide their own, custom, default values for configuration values.

File: hls-plugin-api/src/Ide/Plugin/Config.hs [hunk 4]
```diff
@@ -79,7 +90,18 @@ parseConfig defValue = A.withObject "Config" $ \v -> do
         <*> (o .:? "checkProject" <|> v .:? "checkProject") .!= checkProject defValue
         <*> o .:? "formattingProvider"                      .!= formattingProvider defValue
         <*> o .:? "maxCompletions"                          .!= maxCompletions defValue
-        <*> o .:? "plugin"                                  .!= plugins defValue
+        <*> explicitParseFieldMaybe (parsePlugins $ plugins defValue) o "plugin" .!= plugins defValue
+
+parsePlugins :: Map.Map T.Text PluginConfig -> Value -> A.Parser (Map.Map T.Text PluginConfig)
```
michaelpj, line 95:
```suggestion
-- | Parse the 'PluginConfig'. Since we need to fall back to default values if we do not find one in the input, we need the map of plugin-provided defaults, as in 'defConfig'.
parsePlugins :: Map.Map T.Text PluginConfig -> Value -> A.Parser (Map.Map T.Text PluginConfig)
```

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
```diff
@@ -61,12 +60,14 @@ pluginsToDefaultConfig IdePlugins {..} =
         --   "codeLensOn": true
         -- }
         --
-        genericDefaultConfig =
-          let x = ["diagnosticsOn" A..= True | configHasDiagnostics] <> nubOrd (mconcat (handlersToGenericDefaultConfig <$> handlers))
+        genericDefaultConfig
+          | Nothing <- configInitialGenericConfig = []
+          | Just config <- configInitialGenericConfig =
```
michaelpj, line 65: This feels like a weird way to write `case configInitialGenericConfig` :D

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -220,21 +221,22 @@ data CustomConfig = forall r. CustomConfig (Properties r)
 -- which can be inferred from handlers registered by the plugin.
 -- @config@ is called custom config, which is defined using 'Properties'.
 data ConfigDescriptor = ConfigDescriptor {
-  -- | Whether or not to generate generic configs.
-  configEnableGenericConfig :: Bool,
+  -- | Initial values for the generic config, if the plugin admits one
```
michaelpj, line 224:
```suggestion
  -- | Initial values for the generic config, if the plugin admits them.
```

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -220,21 +221,22 @@ data CustomConfig = forall r. CustomConfig (Properties r)
...
+  configCustomConfig         :: CustomConfig
 }
 
 mkCustomConfig :: Properties r -> CustomConfig
 mkCustomConfig = CustomConfig
 
 defaultConfigDescriptor :: ConfigDescriptor
-defaultConfigDescriptor = ConfigDescriptor True False (mkCustomConfig emptyProperties)
+defaultConfigDescriptor =
+    ConfigDescriptor (Just Data.Default.def) False (mkCustomConfig emptyProperties)
```
michaelpj, line 239: Hmm, this surprises me, I'd have expected the default value to be `Nothing` i.e. the plugin doesn't care?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -748,6 +751,16 @@ pluginEnabledConfig f pid config = plcGlobalOn pluginConfig && f pluginConfig
   where
     pluginConfig = configForPlugin config pid
 
+defConfigForPlugins :: IdePlugins ideState -> Config
```
michaelpj, line 754: haddock?

hls-test-utils/src/Test/Hls.hs
```diff
@@ -152,13 +151,13 @@ goldenWithHaskellDocFormatter plugin formatter conf title testDataDir path desc
...
 runSessionWithServer :: PluginDescriptor IdeState -> FilePath -> Session a -> IO a
-runSessionWithServer plugin = runSessionWithServer' [plugin] def def fullCaps
+runSessionWithServer plugin = runSessionWithServer' [plugin] (defConfigForPlugins $ IdePlugins [plugin]) def fullCaps
 
 runSessionWithServerFormatter :: PluginDescriptor IdeState -> String -> PluginConfig -> FilePath -> Session a -> IO a
 runSessionWithServerFormatter plugin formatter conf =
   runSessionWithServer'
     [plugin]
-    def
+    (defConfig mempty)
```
michaelpj, line 160: Why not `defConfigForPlugins $ IdePlugins [plugin]` again?

michaelpj (approved):
Sorry for taking so long to look at this. LGTM, mostly just some documentation nits.

pepeiborra:
> Sorry for taking so long to look at this. LGTM, mostly just some documentation nits.

Thanks for the review, I'll get to your comments in a couple of weeks

hls-plugin-api/src/Ide/Plugin/Config.hs
pepeiborra: it would require CPP, since `Key` is only defined in aeson 2.0

hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
```diff
@@ -61,13 +60,14 @@ pluginsToDefaultConfig IdePlugins {..} =
...
         --
-        genericDefaultConfig =
-          let x = ["diagnosticsOn" A..= True | configHasDiagnostics] <> nubOrd (mconcat (handlersToGenericDefaultConfig <$> handlers))
-           in case x of
-                -- if the plugin has only one capability, we produce globalOn instead of the specific one;
-                -- otherwise we don't produce globalOn at all
-                [_] -> ["globalOn" A..= True]
-                _   -> x
+        genericDefaultConfig = case configInitialGenericConfig of
+          config ->
```
fendor, line 64:
```suggestion
        genericDefaultConfig =
```

fendor (approved):
LGTM!

pepeiborra:
I found a problem with this scheme: if the user specifies config for *any* plugin, HLS will forget all the initial defaults. This would happen as follows:
1. at startup we seed the plugin config, a map of plugin configs, with the initial values for each plugin. 
2. at runtime the user custom config fills this map with one entry containing custom config for a specific plugin
3. HLS will parse the user custom config, forgetting the previous value with all the initial plugin configs

I added some tests to track this issue. To make the tests pass, I first considered merging the parsed plugin config map with the previous config map. I ended up implementing a slightly different solution:

- no longer initialise the plugin config map with the initial values
- when resolving the current plugin config, read from the plugin map and fall back to the initial defaults if no entry

I like this solution better because it no longer required initialising the config with the plugin defaults. The downside is that to read the current plugin config, we now need the plugin descriptor, not just the plugin id. This required a bunch of additional changes in plugin code.

PR #3196, opened 2022-09-18
Fix the nix 9.4.2 devshell to build every dep with nix

lf-:
Currently cabal tries to build HUnit and more with cabal instead of grabbing them from nix since nix provides old versions in the current ghc942 configuration. This is largely because nix simply does not have a usable package set for ghc942 (basically the configuration-ghc9.4.2.nix file was a stub).

I'm working on that upstream here: https://github.com/NixOS/nixpkgs/pull/191618

This PR fixes that today, but points to a nixpkgs fork, which is not necessarily desirable.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3196"></a>

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -214,17 +214,17 @@
     "lsp": {
       "flake": false,
       "locked": {
-        "lastModified": 1662291729,
-        "narHash": "sha256-KlL38v/75G9zrW7+IiUeiCxFfLJGm/EdFeWQRUikab8=",
+        "lastModified": 1663082924,
+        "narHash": "sha256-H2q53kVRJnxllZtnIN4JPETqCFVIM1WCJjj/iaSwYyk=",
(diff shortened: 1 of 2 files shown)


lf-:
> Said otherwise, is there a "robust" way to fix our nix develop experience so it always take the same packages as nix build, which are working.

Such a good question. I think what is happening is that the HLS packages are getting jailbroken (due to our hack to jailbreak and dontCheck everything) so they ignore the dependencies going into them, but crucially, jailbreak is part of the Nix *build* process, so it doesn't happen to the shell. 

The solution would then be to Not Do That. I think my package set is unjacked enough to try removing that hack without suffering too much pain. I'll try it when I next work on this.

michaelpj:
I'm assuming this is stale, please reopen if not!

PR #3171, opened 2022-09-15
Solve formatting issues (stylish-haskell, pre-commit CI)

andys8:
Because of the failing pre-commit hook https://github.com/haskell/haskell-language-server/issues/3170 a bunch of files are checked in, but not formatted as expected.

Note: The commit hook will still fail because of the first reason (stylish haskell parse errors, https://github.com/haskell/haskell-language-server/issues/3170)

Solves formatting issues

image

Solves some of the parse errors Currently results in ...

image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3171"></a>

Issue #3170, opened 2022-09-15
pre-commit hook is failing locally and on CI


andys8:
All open pull requests have a failing CI job because "pre-commit" will result in errors. 

![image](https://user-images.githubusercontent.com/13085980/190400170-4b083f0a-abd3-4a93-9693-ad5d8513dff9.png)

Solving the root cause or if not possible disabling the job for now would be great in order to avoid confusion (especially for new contributors)

### Steps to reproduce

Run CI, open a PR

Example: https://github.com/haskell/haskell-language-server/actions/runs/3060016524/jobs/4938045479

#### Stylish Haskell fails to parse some files

See #3131 and #3171

##### CI

```
ghcide/src/Development/IDE/GHC/Compat/Plugins.hs: <string>:53:3: error:
    parse error (possibly incorrect indentation or mismatched brackets)
ghcide/src/Development/IDE/GHC/Compat/Env.hs: <string>:101:1: error: parse error on input `import'
```

##### Locally (stylish-haskell 0.14.2.0)

```
ghcide/src/Development/IDE/GHC/Compat/Plugins.hs: <string>:53:3: error:
    parse error (possibly incorrect indentation or mismatched brackets)
plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs: <string>:7:1: error: parse error on input `#'
ghcide/src/Development/IDE/GHC/Compat/Env.hs: <string>:101:1: error: parse error on input `import'
plugins/hls-refactor-plugin/src/Development/IDE/GHC/Compat/ExactPrint.hs: <string>:5:1: error: parse error on input `#'
plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs: <string>:90:1: error: parse error on input `#'
```

#### Some files are not formatted

Fixed with https://github.com/haskell/haskell-language-server/pull/3171

andys8:
https://github.com/haskell/stylish-haskell/issues/409#issuecomment-1250887142

andys8:
CI passing 

https://github.com/haskell/haskell-language-server/pull/3199#issuecomment-1250969019

PR #3199, opened 2022-09-19
Stylish Haskell: CPP parse issues

andys8:
## Change

* CPP language extension
* Manually fix some files with duplication
* Fixed ignored filepaths in pre-commit config (ExactPrint was moved)

### Without this change

```shell
pre-commit run --show-diff-on-failure --all-files

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs: <string>:7:1: error: parse error on input `#'
plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/Util.hs: <string>:14:1: error: parse error on input `#'
plugins/hls-refactor-plugin/src/Development/IDE/GHC/Compat/ExactPrint.hs: <string>:5:1: error: parse error on input `#'
plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs: <string>:90:1: error: parse error on input `#'
```

### With this change

image

image

## System 

stylish-haskell 0.14.2.0

## Issue

* Related to #3170
* Followup of #3171
* If there would be a way to tell stylish-haskell to skip parts of a file, these kind of issues could easier be solved without mangling with the code. It's doesn't exist. See https://github.com/haskell/stylish-haskell/issues/409#issuecomment-1250887142

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3199"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Import/FindImports.hs
+++ b/ghcide/src/Development/IDE/Import/FindImports.hs
@@ -28,7 +28,7 @@ import           Data.List                         (isSuffixOf)
 import           Data.Maybe
 import           System.FilePath
 #if MIN_VERSION_ghc(9,3,0)
-import GHC.Types.PkgQual
+import           GHC.Types.PkgQual
 #endif
 
 data Import
@@ -123,11 +123,12 @@ locateModule env comp_info exts targetFor modName mbPkgName isSource = do
 #if MIN_VERSION_ghc(9,3,0)
     OtherPkg uid
       | Just dirs <- lookup uid import_paths
+          -> lookupLocal uid dirs
 #else
     Just pkgName
       | Just (uid, dirs) <- lookup (PackageName pkgName) import_paths
-#endif
           -> lookupLocal uid dirs
+#endif
       | otherwise -> lookupInPackageDB env
 #if MIN_VERSION_ghc(9,3,0)
     NoPkgQual -> do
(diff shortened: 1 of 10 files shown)


ghcide/src/Development/IDE/Import/FindImports.hs
```diff
@@ -123,11 +123,12 @@ locateModule env comp_info exts targetFor modName mbPkgName isSource = do
 #if MIN_VERSION_ghc(9,3,0)
     OtherPkg uid
       | Just dirs <- lookup uid import_paths
+          -> lookupLocal uid dirs
```
andys8, line 126: :information_source: I've seen that in many cases stylish-haskell has trouble to parse parts of code, because they are so partial and don't seem to make sense to the parser. Duplication of parts of the code - in order to make the blocks more complete - helps the parser to understand what's going on.

andys8:
@pepeiborra Any change `pre-commit` CI job can be required in order to prevent this to happen again in the future? It's way easier to fix (or wiggle around) a format issue right when you work on a file, rather than fail all - or now some - other PRs that have nothing to do with it.

image

pepeiborra:
> @pepeiborra Any change `pre-commit` CI job can be required in order to prevent this to happen again in the future? It's way easier to fix (or wiggle around) a format issue right when you work on a file, rather than fail all - or now some - other PRs that have nothing to do with it.

@andys8 thanks a lot for this fix, it is very appreciated. The pre-commit hook was contributed by @lunaticare and accepted by @Ailrun, maybe they can comment.

ghcide/src/Development/IDE/Import/FindImports.hs
pepeiborra: Thanks for the context!

pepeiborra:
My thoughts: 

We have some of the worst CPP code in Haskell-land, it's bad enough for contributors as it is. Having to fight CI to make stylish-haskell "parse" the CPP will make things worse. 

Ideally the pre-commit-hook would ignore parse errors, and leave those files untouched. Perhaps the exit code can be used to tell a parse error apart from an unformatted file?

pepeiborra (approved):

PR #3203, opened 2022-09-20
Sort vscode extension schema json by keys

fendor:
Makes it easier to copy and paste configurations into VSCode and reviewing what options have been added and removed. I was reminded that we need something like this and add this to the release checklist for HLS, since we forgot to add Stan options vscode.

Remove code-duplication, namely ghcide exe loses some capabilities, as it is destined to be removed sooner or later.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3203"></a>


ghcide/src/Development/IDE/Main.hs
```diff
@@ -191,8 +191,6 @@ data Command
     | Db {hieOptions ::  HieDb.Options, hieCommand :: HieDb.Command}
      -- ^ Run a command in the hiedb
     | LSP   -- ^ Run the LSP server
-    | PrintExtensionSchema
-    | PrintDefaultConfig
```
fendor, line 195: For anyone who was wondering, this was the actual codepath taken, instead of `VSCodeExtensionSchemaMode` from `src/Ide/Main.hs`.

fendor:
Actually, isn't there supposed to be a `haskell.plugins.stan.globalOn` configuration? Couldn't find it right now.

michaelpj (approved):

michaelpj:
I assume the sorting means that we do get options for the same plugin near each other because they will have the same prefix?

fendor:
yes, that's the idea, lexicographic sorting should take care of it. Unless my brain is messing up again

PR #3204, opened 2022-09-21
Implement completionItem/resolve

wz1000:
Implement `completionItem/resolve` to lazily resolve types and documentation for completion items.
This results in very big memory usage improvments while also improving the responsiveness of completions.
Usually when GHC typechecks code, it reads in and deserializes interfaces lazily, so if the code imports
a bunch of modules, the interfaces will only be read in and deserialized if a particular name is actually
required for typechecking.

However, when we greedily generate completion lists, the act of querying for types and documentation results
in sucking in and deserializing interface sections for every single name in scope. This is very bad for
memory usage as it negates virtually all the benefits of lazy deserialization.

Instead, we can compute types and documentation on demand for individual completion items using the `completionItem/resolve`
LSP functionality.

A downside to this is that we can no longer generate snippets for functions
arguments with placeholders mentioning the types because we no longer have this
information available when we generate the original completion, and it is not
supported by LSP to resolve the actual completion snippet using `completionItem/resolve`.

However, I don't personally consider this to be a very big deal because this
feature was unreliable (eg for things from the export map which weren't in
scope) and I found it to be quite annoying when it did work.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3204"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -12,21 +12,27 @@ module Development.IDE.Plugin.Completions
 import           Control.Concurrent.Async                 (concurrently)
 import           Control.Concurrent.STM.Stats             (readTVarIO)
 import           Control.Monad.IO.Class
+import           Control.Lens                            ((&), (.~))
 import qualified Data.HashMap.Strict                      as Map
 import qualified Data.HashSet                             as Set
+import           Data.Aeson
 import           Data.Maybe
 import qualified Data.Text                                as T
 import           Development.IDE.Core.PositionMapping
+import           Development.IDE.Core.Compile
 import           Development.IDE.Core.RuleTypes
 import           Development.IDE.Core.Service             hiding (Log, LogShake)
 import           Development.IDE.Core.Shake               hiding (Log)
 import qualified Development.IDE.Core.Shake               as Shake
 import           Development.IDE.GHC.Compat
+import           Development.IDE.GHC.Util
 import           Development.IDE.Graph
+import           Development.IDE.Spans.Common
+import           Development.IDE.Spans.Documentation
 import           Development.IDE.Plugin.Completions.Logic
 import           Development.IDE.Plugin.Completions.Types
 import           Development.IDE.Types.Exports
-import           Development.IDE.Types.HscEnvEq           (HscEnvEq (envPackageExports),
+import           Development.IDE.Types.HscEnvEq           (HscEnvEq (envPackageExports, envVisibleModuleNames),
                                                            hscEnv)
 import qualified Development.IDE.Types.KnownTargets       as KT
 import           Development.IDE.Types.Location
@@ -37,6 +43,8 @@ import           Development.IDE.Types.Logger             (Pretty (pretty),
 import           Ide.Types
 import qualified Language.LSP.Server                      as LSP
 import           Language.LSP.Types
+import qualified Language.LSP.Types.Lens         as J
+import qualified Language.LSP.VFS                         as VFS
 import           Numeric.Natural
 import           Text.Fuzzy.Parallel                      (Scored (..))
 
@@ -57,9 +65,11 @@ descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeSta
 descriptor recorder plId = (defaultPluginDescriptor plId)
   { pluginRules = produceCompletions recorder
   , pluginHandlers = mkPluginHandler STextDocumentCompletion getCompletionsLSP
+                  <> mkPluginHandler SCompletionItemResolve resolveCompletion
   , pluginConfigDescriptor = defaultConfigDescriptor {configCustomConfig = mkCustomConfig properties}
   , pluginPriority = ghcideCompletionsPluginPriority
   }
+
 
 produceCompletions :: Recorder (WithPriority Log) -> Rules ()
 produceCompletions recorder = do
@@ -85,8 +95,9 @@ produceCompletions recorder = do
               (global, inScope) <- liftIO $ tcRnImportDecls env (dropListFromImportDecl <$> msrImports) `concurrently` tcRnImportDecls env msrImports
               case (global, inScope) of
                   ((_, Just globalEnv), (_, Just inScopeEnv)) -> do
+                      visibleMods <- liftIO $ fmap (fromMaybe []) $ envVisibleModuleNames sess
                       let uri = fromNormalizedUri $ normalizedFilePathToUri file
-                      cdata <- liftIO $ cacheDataProducer uri sess (ms_mod msrModSummary) globalEnv inScopeEnv msrImports
+                      let cdata = cacheDataProducer uri visibleMods (ms_mod msrModSummary) globalEnv inScopeEnv msrImports
                       return ([], Just cdata)
                   (_diag, _) ->
                       return ([], Nothing)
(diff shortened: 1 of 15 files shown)


michaelpj:
Wow, this is a big deal!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -370,65 +316,63 @@ cacheDataProducer uri env curMod globalEnv inScopeEnv limports = do
...
               origMod = showModName (is_mod spec)
-          return (unqual,QualCompls qual)
+          in (unqual,QualCompls qual)
 
-      toCompItem :: Parent -> Module -> T.Text -> Name -> Maybe (LImportDecl GhcPs) -> IO [CompItem]
-      toCompItem par m mn n imp' = do
-        docs <- getDocumentationTryGhc packageState curMod n
+      toCompItem :: Parent -> Module -> T.Text -> Name -> Maybe (LImportDecl GhcPs) -> [CompItem]
+      toCompItem par m mn n imp' =
+        -- docs <- getDocumentationTryGhc packageState curMod n
```
michaelpj, line 352: commented out code

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -370,65 +316,63 @@ cacheDataProducer uri env curMod globalEnv inScopeEnv limports = do
 
       rdrElts = globalRdrEnvElts globalEnv
 
-      foldMapM :: (Foldable f, Monad m, Monoid b) => (a -> m b) -> f a -> m b
-      foldMapM f xs = foldr step return xs mempty where
-        step x r z = f x >>= \y -> r $! z `mappend` y
+      -- construct a map from Parents(type) to their fields
+      fieldMap = Map.fromListWith (++) $ flip mapMaybe rdrElts $ \elt -> do
```
michaelpj, line 320: Can this get big? Could this repeated `++` be a performance hit?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -840,6 +768,8 @@ mkRecordSnippetCompItem uri parent ctxStr compl importedFrom docs imp = r
                   importQual = getImportQual x,
                   newThing = ctxStr
                 }
+          , nameDetails = Nothing
+          , isLocalCompletion = True
```
michaelpj, line 772: Is this accurate? Doesn't it depend on whether the record is imported?

ghcide/src/Development/IDE/Plugin/Completions/Types.hs
```diff
@@ -94,13 +96,13 @@ data CompItem = CI
...
   , provenance          :: Provenance     -- ^ From where this item is imported from.
-  , typeText            :: Maybe T.Text   -- ^ Available type information.
   , label               :: T.Text         -- ^ Label to display to the user.
   , isInfix             :: Maybe Backtick -- ^ Did the completion happen
                                    -- in the context of an infix notation.
-  , docs                :: SpanDoc        -- ^ Available documentation.
   , isTypeCompl         :: Bool
   , additionalTextEdits :: Maybe ExtendImport
+  , nameDetails         :: Maybe NameDetails
+  , isLocalCompletion   :: Bool
```
michaelpj, line 105: haddock? what does this mean?

ghcide/src/Development/IDE/Plugin/Completions/Types.hs
```diff
@@ -136,3 +138,42 @@ instance Monoid CachedCompletions where
 instance Semigroup CachedCompletions where
     CC a b c d e <> CC a' b' c' d' e' =
         CC (a<>a') (b<>b') (c<>c') (d<>d') (e<>e')
+
+data NameDetails
```
michaelpj, line 142: haddock? what does this mean?

michaelpj: In particular, I think it would be good to explain broadly what you're doing _somewhere_. There's quite a bit of logic spread around: you capture some information in the first pass which you want during resolution, so you store it in this datatype and put it into the `data` field of the completion item... this stuff is not obvious!

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -106,6 +109,44 @@ dropListFromImportDecl iDecl = let
     f x = x
     in f <$> iDecl
 
+resolveCompletion :: IdeState -> PluginId -> CompletionItem -> LSP.LspM Config (Either ResponseError CompletionItem)
+resolveCompletion ide _ comp@CompletionItem{_detail,_documentation,_xdata}
+  | Just resolveData <- _xdata
+  , Success (uri, NameDetails mod occ) <- fromJSON resolveData
+  , Just file <- uriToNormalizedFilePath $ toNormalizedUri uri = do
```
michaelpj, line 116: I think each of these cases could probably at least log if they fail, so that we know what happened. At the moment it would be totally impossible to diagnose e.g. the JSON failing to decode, we'd just get no resolution.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -106,6 +109,44 @@ dropListFromImportDecl iDecl = let
...
+    mdkm <- liftIO $ runAction "Completion resolve" ide $ use GetDocMap file
+    case mdkm of
+      Nothing -> pure (Right comp)
+      Just (DKMap dm km) -> liftIO $ runAction "Completion resolve" ide $ do
```
michaelpj, line 120: I think you can make this kind of pattern simpler using the pattern we use elsewhere in the codebase using `pluginResponse` at the top level and `maybeToEither` and friends.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -106,6 +109,44 @@ dropListFromImportDecl iDecl = let
     f x = x
     in f <$> iDecl
 
+resolveCompletion :: IdeState -> PluginId -> CompletionItem -> LSP.LspM Config (Either ResponseError CompletionItem)
+resolveCompletion ide _ comp@CompletionItem{_detail,_documentation,_xdata}
+  | Just resolveData <- _xdata
+  , Success (uri, NameDetails mod occ) <- fromJSON resolveData
+  , Just file <- uriToNormalizedFilePath $ toNormalizedUri uri = do
+    mdkm <- liftIO $ runAction "Completion resolve" ide $ use GetDocMap file
```
michaelpj, line 117: Why is this in its own action, whereas below we have multiple `use`s in the same action? Worth a comment.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -106,6 +109,44 @@ dropListFromImportDecl iDecl = let
...
+          Just ty -> pure (safeTyThingType ty)
+          Nothing -> do
+            (safeTyThingType =<<) <$> liftIO (lookupName (hscEnv sess) cur_mod name)
+        let det1 = case typ of
+              Just ty -> Just (":: " <> printOutputable ty <> "\n")
+              Nothing -> Nothing
+            doc1 = case _documentation of
+              Just (CompletionDocMarkup (MarkupContent MkMarkdown old)) ->
+                CompletionDocMarkup $ MarkupContent MkMarkdown $ T.intercalate sectionSeparator (old:doc)
+              _ -> CompletionDocMarkup $ MarkupContent MkMarkdown $ T.intercalate sectionSeparator doc
```
michaelpj, line 143: Is this right? You do the same thing for `Nothing` and for `Just` but with different markup content?

michaelpj (commented):
Code looks good! I'm going to do my usual routine and question whether it's really the case that adding a complex feature like this requires 0 lines of explanatory comments :)

michaelpj:
Also: tests?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
wz1000: It's a map from type constructor to field names, so `[(Foo,[field1, field2,...]), ...]`. I guess it could get big if you have big records, but I don't think its any worse than any of the other listy code we have in completions.

michaelpj: Worth a comment, casual use of `++` makes me jumpy these days :D

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
pepeiborra: @wz1000 is this the line causing the eager deserialization of interfaces?

wz1000: This along with `lookupName` (the `lookupName` is slightly worse since it forces typechecking the iface which is also done lazily, `getDocumentation` will still load in the interface if it wasn't needed before and deserialize the documentation part).

michaelpj:
Oh one last thing: we definitely want a cautionary comment on the main completion handler warning people not to accidentally pull in information from interface files in future!

michaelpj:
https://github.com/haskell/haskell-language-server/pull/3080 gets the HIE file during completions, is that acceptable?

pepeiborra:
> https://github.com/haskell/haskell-language-server/pull/3080 gets the HIE file during completions, is that acceptable?

I am pretty confident getting the HIE file is 👍

michaelpj:
I merged the record dot completions PR, which has introduced some conflicts.

michaelpj (commented):
Previous comments still need addressing also.

isovector (approved):
lgtm for wingman

.github/workflows/test.yml
```diff
@@ -221,7 +221,7 @@ jobs:
         name: Test hls-explicit-imports-plugin test suite
         run: cabal test hls-explicit-imports-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-explicit-imports-plugin --test-options="$TEST_OPTS"
 
-      - if: matrix.test
+      - if: matrix.test && matrix.os != 'windows-latest'
```
michaelpj, line 224: why? is there an issue?

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
michaelpj: still commented out

michaelpj (approved):
Fine, although I do think there were a few more comments that could be addressed.

.github/workflows/test.yml
wz1000: call-hierarchy tests were very flaky, I've fixed the problem I think and removed this.

michaelpj:
pre-commit check is also failing, I think you're missing formatting

PR #3205, opened 2022-09-21
Improve hls-fixity-plugin

wz1000:
Simplify and improve the performance and memory usage behaviour or hls-fixity-plugin.

Also restore the property that hover is immediately available on startup using the cached hie file data and `IdeAction` mechanism.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3205"></a>


July541:
Many thanks for your improvement!

I'm not familiar with `IdeAction`, would you mind give a review? @pepeiborra

pepeiborra:
Thanks. Could you also enable the benchmark for this plugin, just so we can see the improvement? Maybe the hover benchmark needs extending though..

pepeiborra (approved):

PR #3213, opened 2022-09-22
Add diagnostics to Stan descriptor

pepeiborra:
This has effect only when generating the config schema and won't change anything for the Stan plugin schema at all

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3213"></a>


kokobd (approved):
LGTM

PR #1340, opened 2021-02-10
Add pragma after she-bang

ishmum123:
Draft PR for #555

PR #3216, opened 2022-09-23
hls-hlint-plugin: Update README.md

hololeap:
Remove warning about hlint restrictions, as #1340 has been merged.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3216"></a>


kokobd:
@hololeap Have you checked if that PR has been released to Hackage, and HLS is currently using a version of hlint containing that feature?

michaelpj:
I think master documentation should reflect master. We can start using RTD's versioned documentation feature if we want to make sure people see the documentation for released versions!

michaelpj:
We are using a newer hlint on master, I checked.

michaelpj (approved):

PR #3227, opened 2022-09-27
Use nixpkgs variants of Sphinx packages

### LINKED ISSUE #3211: Building documentation fails with a Python import error
Build of the documentation fails with the message provided in the "Actual behaviour" section. It fails regardless of the method used for building the documentation (i.e. both with `nix build .#docs` and with `cd docs; make html`).
### Your environment

Which OS do you use?
NixOS

### Steps to reproduce

```bash
nix build .#docs
```

### Expected behaviour

Documentations should be building

### Actual behaviour

```bash
error: builder for '/nix/store/8vfimsxaib5l3n77gap9vswgv6bhw0zx-hls-docs.drv' failed with exit code 2;
       last 10 log lines:
       > unpacking source archive /nix/store/2djd5r3qzm8ay4ygajk8rj0kf17hqmhz-source
       > source root is source
       > patching sources
       > configuring
       > no configure script, doing nothing
       > building
       > Running Sphinx v5.1.1
       >
       > Extension error:
       > Could not import extension myst_parser (exception: No module named 'myst_parser')
       For full logs, run 'nix log /nix/store/8vfimsxaib5l3n77gap9vswgv6bhw0zx-hls-docs.drv'.
```

ozkutuk:
Fixes #3211  

This PR replaces manually specified `myst-parser` and `sphinx_rtd_theme` packages with their counterparts already packaged on nixpkgs. With this change, `poetry2nix` is no more used, therefore that is removed as well.

The documentation build via Nix (`nix build .#docs`) still fails. However, the cause of failure is not related to the Nix infrastructure, but a warning in the documentation source. Nix build of the documentation promotes warnings to errors, hence the failure. We need to either lift this restriction, or actually fix the warning. Either way, that should be handled in a separate PR.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3227"></a>

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -84,21 +84,6 @@
         "type": "github"
       }
     },
-    "flake-utils_2": {
-      "locked": {
-        "lastModified": 1659877975,
-        "narHash": "sha256-zllb8aq3YO3h8B/U0/J1WBgAL8EX5yWf5pMj3G0NAmc=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "c0e246b9b83f637f4681389ecabcb2681b4f3af0",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
     "fourmolu": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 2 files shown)


michaelpj (approved):
Excellent, thank you. I think `myst_parser` was not in nixpkgs when I wrote this.

PR #3229, opened 2022-09-27
Fix error in code range

kokobd:
I'm seeing this popup every time an empty source file is opened, like below:
image
```haskell
module Empty where

```

`HieASTs` doesn't have an AST for empty files (by empty files, I mean those haskell source files without anything other than the module header)

I'm not sure if it's the problem of `GetHieAST`, but let's just decrease the error level in the code range plugin, and handle the case when `CodeRange` rule fails in a more tolerant way.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3229"></a>

### DIFF
--- a/plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
+++ b/plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
@@ -1,6 +1,7 @@
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE RecordWildCards     #-}
-{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE OverloadedStrings         #-}
+{-# LANGUAGE RecordWildCards           #-}
+{-# LANGUAGE ScopedTypeVariables       #-}
 
 module Ide.Plugin.CodeRange (
     descriptor
@@ -33,7 +34,9 @@ import           Development.IDE                      (Action, IdeAction,
 import           Development.IDE.Core.PositionMapping (PositionMapping,
                                                        fromCurrentPosition,
                                                        toCurrentRange)
-import           Development.IDE.Types.Logger         (Pretty (..))
+import           Development.IDE.Types.Logger         (Pretty (..),
+                                                       Priority (Warning),
+                                                       logWith)
 import           Ide.Plugin.CodeRange.Rules           (CodeRange (..),
                                                        GetCodeRange (..),
                                                        codeRangeRule, crkToFrk)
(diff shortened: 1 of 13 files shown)


michaelpj (commented):
What about keeping the structure of the current code, but using `ExceptT` over a custom error type. Then we can fail saying "couldn't get a dependency" or whatever, and then decide just below `pluginResponse` what to do about that, e.g. log or return an error or what.

(Generally, I think it would be nice if `pluginResponse` encouraged people to do this instead of just using a `String` as the error type and turning them all into response errors...)

kokobd:
> What about keeping the structure of the current code, but using `ExceptT` over a custom error type. Then we can fail saying "couldn't get a dependency" or whatever, and then decide just below `pluginResponse` what to do about that, e.g. log or return an error or what.
> 
> (Generally, I think it would be nice if `pluginResponse` encouraged people to do this instead of just using a `String` as the error type and turning them all into response errors...)

Good idea!

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -77,49 +77,62 @@ foldingRangeHandler ide _ FoldingRangeParams{..} = do
...
+    -- This might happen if the HieAst is not ready, so we give it a default value instead of throwing an error
+    showError SelectionRangeBadDependency = Nothing
+    showError SelectionRangeInputPositionMappingFailure = Just "failed to apply position mapping to input positions"
+    showError SelectionRangeOutputPositionMappingFailure = Just "failed to apply position mapping to output positions"
+
+data SelectionRangeError = SelectionRangeBadDependency
+                         | SelectionRangeInputPositionMappingFailure
+                         | SelectionRangeOutputPositionMappingFailure
+
+getSelectionRanges :: NormalizedFilePath -> [Position] -> ExceptT SelectionRangeError IdeAction [SelectionRange]
```
kokobd, line 120: @michaelpj Did I get it correctly?

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -77,49 +77,62 @@ foldingRangeHandler ide _ FoldingRangeParams{..} = do
...
-getSelectionRanges :: NormalizedFilePath -> [Position] -> ExceptT String IdeAction [SelectionRange]
+    runIdeAction' :: IdeAction (Either SelectionRangeError [SelectionRange]) -> LspT c IO (Either String [SelectionRange])
+    runIdeAction' action = do
+        result <- liftIO $ runIdeAction "SelectionRange" (shakeExtras ide) action
+        pure $ case result of
+            Left err   -> maybe (Right []) Left (showError err)
+            Right list -> Right list
+
+    showError :: SelectionRangeError -> Maybe String
+    -- This might happen if the HieAst is not ready, so we give it a default value instead of throwing an error
```
michaelpj, line 111: I'd probably just inline this function, then it would look less weird that you're kind of doing different things in the different branches?

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -77,49 +77,62 @@ foldingRangeHandler ide _ FoldingRangeParams{..} = do
...
+
+data SelectionRangeError = SelectionRangeBadDependency
+                         | SelectionRangeInputPositionMappingFailure
+                         | SelectionRangeOutputPositionMappingFailure
+
+getSelectionRanges :: NormalizedFilePath -> [Position] -> ExceptT SelectionRangeError IdeAction [SelectionRange]
 getSelectionRanges file positions = do
-    (codeRange, positionMapping) <- maybeToExceptT "fail to get code range" $ useE GetCodeRange file
+    (codeRange, positionMapping) <- maybeToExceptT SelectionRangeBadDependency . MaybeT $
+        useWithStaleFast GetCodeRange file
```
michaelpj, line 123: maybe we should include the dependency that failed in the error so we can log it later? I guess `useWithStaleFast` doesn't log if it fails?

michaelpj (approved):
Nice! Makes me want to refactor the other plugins :D

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
michaelpj: Right, I realised when thinking about this that a plugin returning a `ResponseError` can be "normal" and not necessarily something that HLS needs to log as an error. I still think it's good to log them so you know what's happening, but maybe it doesn't need to be an error log.

PR #3231, opened 2022-09-28
Improve memory characteristics of ExportsMap

wz1000:
Storing rendered names as `Text`, especially for parents, adds a lot of duplication to the ExportsMap.
Instead we store the `OccName`s directly, which have hash-consed symbols due stored as `FastStrings`
and render it out on demand (which is just decoding the UTF-8 FastString to UTF-16 text for text <2.0,
and essentially free on text >2.0).

This PR depends on #3204, ignore that commit while reviewing.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3231"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Types/Exports.hs
+++ b/ghcide/src/Development/IDE/Types/Exports.hs
@@ -8,27 +8,31 @@ module Development.IDE.Types.Exports
     rendered,
     moduleNameText,
     occNameText,
+    renderOcc,
+    mkTypeOcc,
+    mkVarOrDataOcc,
     isDatacon,
     createExportsMap,
     createExportsMapMg,
-    createExportsMapTc,
     buildModuleExportMapFrom,
     createExportsMapHieDb,
     size,
+    exportsMapSize,
     updateExportsMapMg
     ) where
 
 import           Control.DeepSeq             (NFData (..))
 import           Control.Monad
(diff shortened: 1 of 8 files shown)


ghcide/src/Development/IDE/Types/Exports.hs
```diff
@@ -158,49 +183,44 @@ type WithHieDb = forall a. (HieDb -> IO a) -> IO a
 createExportsMapHieDb :: WithHieDb -> IO ExportsMap
 createExportsMapHieDb withHieDb = do
     mods <- withHieDb getAllIndexedMods
-    idents <- forM (filter (nonInternalModules . modInfoName . hieModInfo) mods) $ \m -> do
+    idents' <- forM (filter (nonInternalModules . modInfoName . hieModInfo) mods) $ \m -> do
         let mn = modInfoName $ hieModInfo m
-            mText = pack $ moduleNameString mn
-        fmap (wrap . unwrap mText) <$> withHieDb (\hieDb -> getExportsForModule hieDb mn)
-    let exportsMap = Map.fromListWith (<>) (concat idents)
-    return $ ExportsMap exportsMap $ buildModuleExportMap (concat idents)
```
wz1000, line 166: This fixes a bug where the module exports map created from hiedb would be keyed by the identifier name instead of the module name. Caught because of using proper types instead of Text!

PR #3235, opened 2022-09-29
Move new imports down the code action list

kokobd:
There are so many suggestions for new imports sometimes, but usually, I just want to extend the existing import if it is possible. I have to scroll down to the list to find it. I think it would be reasonable to change the order a bit.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3235"></a>


michaelpj:
If we think there's an obvious best suggestion here, we could also give some of them `isPreferred`?

kokobd:
> If we think there's an obvious best suggestion here, we could also give some of them `isPreferred`?

I think we are not 100% sure. Even if an already imported module exports a symbol with our desired name, it might not be what the user wants. And there might be multiple modules exporting symbols with the same name.

pepeiborra (approved):

PR #3249, opened 2022-10-05
Simplify implementation of eval plugin

mpickering:
The plugin was implemented by calling "load" which circumvents all of HLSs caching mechanisms for interface files and linkables. Instead we should work like the other typechecking functions which get all the stuff we need using HLS rules and setup the HscEnv with all the state in the right places.

The key part to this is setting up all the HPT modules with linkables if they are depenedencies of the module we are trying to run a function from.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3249"></a>

### DIFF
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
@@ -29,74 +29,69 @@ import           Control.Exception                            (try)
 import qualified Control.Exception                            as E
 import           Control.Lens                                 (_1, _3, ix, (%~),
                                                                (<&>), (^.))
-import           Control.Monad                                (guard, join,
+import           Control.Monad                                (guard,
                                                                void, when)
 import           Control.Monad.IO.Class                       (MonadIO (liftIO))
 import           Control.Monad.Trans                          (lift)
 import           Control.Monad.Trans.Except                   (ExceptT (..))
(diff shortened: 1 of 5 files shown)


michaelpj:
Would you like to add yourself to CODEOWNERS for this plugin? :) Since it's currently totally unloved.

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs
```diff
@@ -761,22 +685,3 @@ parseGhciLikeCmd input = do
...
-        dflags3c =
-            foldl gopt_unset dflags3b $
-                concatMap (Compat.wayUnsetGeneralFlags platform) evalWays
-        dflags4 =
-            dflags3c
-                `gopt_set` Opt_ImplicitImportQualified
-                `gopt_set` Opt_IgnoreOptimChanges
-                `gopt_set` Opt_IgnoreHpcChanges
-                `gopt_unset` Opt_DiagnosticsShowCaret
-    Compat.hsc_dflags <$> Compat.initializePlugins (Compat.hscSetFlags dflags4 env)
```
pepeiborra, line 782: Why is it ok to skip this setup? It was copied from GHCi to try and replicate the same behaviour

pepeiborra (approved):
This looks great to me, but I'm far from an expert on the GHC interactive evaluation internals.

wz1000:
Superseded by #3391

PR #3250, opened 2022-10-05
Proof of concept: Diagrams plugin

edsko:
This PR adds a proof-of-concept plugin that can render diagrams when hovering over a (top-level) binding of the appropriate type. 

Perhaps it would be  nicer to generalize this over some kind of custom `Render` type class, but that might get a bit messy dependency wise (where would that `Render` type class be defined)?

It would also be much nicer if the diagram would be shown inline as a code lens rather than on hover, but it seems this is not possible with LSP?

With many thanks to @wz1000 and @mpickering !

image

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3250"></a>

### DIFF
--- a/plugins/hls-diagrams-plugin/src/Ide/Plugin/Diagrams.hs
+++ b/plugins/hls-diagrams-plugin/src/Ide/Plugin/Diagrams.hs
@@ -1,0 +1,332 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ide.Plugin.Diagrams (descriptor) where
+
+import Prelude hiding (mod)
+
+import Control.DeepSeq
+import Control.Monad.Except
+import Control.Monad.IO.Unlift
+import Data.Hashable
+import Data.List (intercalate)
+import Data.List (isPrefixOf)
+import Data.Maybe (mapMaybe)
+import Data.String
+import Data.Text (Text)
+import Data.Time (UTCTime)
+import Diagrams.Backend.SVG (B)
+import Diagrams.Prelude (Diagram)
+import GHC.Generics (Generic)
+import System.IO.Temp
+
+import qualified Data.Map          as Map
+import qualified Data.Text         as Text
+import qualified Diagrams.Prelude  as D
+
+import Development.IDE.Core.Rules (IdeState)
+import Development.IDE.GHC.Compat (HscEnv, Name, ModSummary, DynFlags, ModuleName, GhcMonad)
+import Development.IDE.Types.Logger
+import Language.LSP.Server (MonadLsp)
+
+import qualified Development.IDE                              as IDE
+import qualified Development.IDE.Core.Compile                 as IDE
+import qualified Development.IDE.Core.FileStore               as IDE
+import qualified Development.IDE.Core.Rules                   as IDE
+import qualified Development.IDE.GHC.Compat                   as IDE
+import qualified Development.IDE.Import.DependencyInformation as IDE
+import qualified Development.IDE.Spans.AtPoint                as IDE
+import qualified Development.IDE.GHC.Compat                   as GHC
+import qualified Development.IDE.GHC.Util                     as GHC
+import qualified Development.IDE.Core.Shake                   as Shake
+import qualified Development.IDE.Types.Shake                  as Shake
+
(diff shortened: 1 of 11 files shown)


plugins/hls-diagrams-plugin/src/Ide/Plugin/Diagrams.hs
```diff
@@ -0,0 +1,332 @@
...
+            else do
+              logWith recorder Info $ RenderingFailed
+              return Nothing
+          _otherwise ->
+            return Nothing
+      IDE.HieFromDisk{} ->
+        return Nothing
+  where
+    looksLikeDiagram :: IDE.Type -> Bool
+    looksLikeDiagram = ("Diagram" `isPrefixOf`) . IDE.printWithoutUniques
```
michaelpj, line 196: Do we not at least have the qualified name here?

plugins/hls-diagrams-plugin/src/Ide/Plugin/Diagrams.hs
```diff
@@ -0,0 +1,332 @@
...
+-- | Unload any modules that are out of date
+unloadOutOfDate :: MonadIO m => IdeState -> HscEnv -> CatchErrors c m ()
+unloadOutOfDate ide env = runAction ide $ do
+    current <- IDE.currentLinkables
+    liftIO $ GHC.unload env $ map (uncurry aux) $ GHC.moduleEnvToList current
+  where
+    aux :: GHC.NoCompat.Module -> UTCTime -> GHC.Linkable
+    aux mod time = GHC.LM time mod []
+
+initialiseSessionForEval ::
```
michaelpj, line 236: Can these bits be shared with the eval plugin? It looks like you already depend on it. GHC session management is fiddly at the best of times, it would be great if we can deduplicate it!

plugins/hls-diagrams-plugin/src/Ide/Plugin/Diagrams.hs
```diff
@@ -0,0 +1,332 @@
...
+          , show tempFile
+          , "(Diagrams.TwoD.Size.dims2D 100 100)"
+          , "x"
+          ]
+
+{-------------------------------------------------------------------------------
+  Example diagram
+-------------------------------------------------------------------------------}
+
+_myCircle :: Diagram B
```
michaelpj, line 331: dead?

michaelpj (commented):
Looks plausible! How's the compatibility story for `diagrams`?

I think hovers are the best place to put this, sadly.

Could do with some tests.

plugins/hls-diagrams-plugin/src/Ide/Plugin/Diagrams.hs
pepeiborra: I was going to ask the same thing. The Eval plugin already contains code to do interactive evaluation, so I think it would be best if hls-diagrams-plugin reused more of it!

pepeiborra (approved):
This looks good to me, modulo the code sharing with the Eval plugin. I'm happy to merge it as soon as you update the docs to list the new plugin. In particular:

- [ ] update `docs/features.md`
- [ ] update `docs/supported-versions.md`
- [ ]  add yourself to the CODEOWNERS file,

Thankyou!

edsko:
I'm happy to make those changes, but then we should wait for the dust to settle on https://github.com/haskell/haskell-language-server/pull/3249 first I guess. 

Also, if we are going to merge this,perhaps I should make a minor modification: instead of recognizing `diagrams` specifically, we could instead recognize something like

```haskell
// hls:renderimage
myImage :: FilePath -> IO ()
myImage = ... rendering code ...
```

or some such, so that we generalize away from `diagrams` and can render _anything_ that can write itself to a file? In this case of course it would be more important for people to opt-in to this; but perhaps this is too open to security exploits? (I'm happy to leave it specific to `diagrams` too, but some people were asking me about other kinds of rendering).

I also need to add a handle to clean up the temporary files on exit, I will do that too, along with the other suggestions. 

But yes, let's wait for Matt's eval simplification PR first (this was written in parallel to that last week).

edsko:
Sorry, I just realized that there is an obvious better way to do have the generalization without causing security exploits. The above signature is too general; we should just have

```haskell
-- hls:renderimage
myImage :: ByteString
myImage = .. rendering code ..
```

or something like that.

PR #3258, opened 2022-10-07
Support ghc 9.4 for hls-class-plugin

July541:


Ailrun (approved):
Does it work as-is? Sounds great :)

July541:
The test looks fine, and I've tested it on my machine, nothing has a problem.

I'm surprised too that it works without any code being modified :)

plugins/hls-class-plugin/test/Main.hs
```diff
@@ -96,7 +96,7 @@ codeLensTests recorder = testGroup
     , goldenCodeLens recorder "Apply code lens for local class" "LocalClassDefine" 0
     , goldenCodeLens recorder "Apply code lens on the same line" "Inline" 0
     , goldenCodeLens recorder "Don't insert pragma while existing" "CodeLensWithPragma" 0
-    , onlyRunForGhcVersions [GHC92] "Only ghc-9.2 enabled GHC2021 implicitly" $
+    , onlyRunForGhcVersions [GHC92, GHC94] "Only ghc-9.2 enabled GHC2021 implicitly" $
```
Ailrun, line 99: As CI failed, let me add one nitpicking here. The message may sound better with `"ghc-9.2 or above"` instead of `"ghc-9.2"`.

PR #3259, opened 2022-10-07
Hlint: A handful of fixes to hints

andys8:
## Goal

Less existing warnings in pull requests - by addressing some of the issues.

## Changes

* hlint fixes
* hlint config (https://github.com/haskell/haskell-language-server/pull/3259#discussion_r990197895)
* CI: `rwe/actions-hlint-run@v2`

## Review 

While most of the changes seem to be trivial (whitespace or parens removed), some changes prefer `extra` functions (as it has been configured). What I think is potentially critical is removing extensions as suggested by hlint: Is it possible that one of the removed extensions is actually changing functionality? If so, please comment.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3259"></a>


hls-plugin-api/src/Ide/Plugin/ConfigUtils.hs
```diff
@@ -106,7 +106,7 @@ pluginsToVSCodeExtensionSchema IdePlugins {..} = A.object $ mconcat $ singlePlug
         genericSchema =
           let x =
                 [toKey' "diagnosticsOn" A..= schemaEntry "diagnostics" | configHasDiagnostics]
-                  <> nub (mconcat (handlersToGenericSchema <$> handlers))
+                  <> Data.List.Extra.nubOrd (mconcat (handlersToGenericSchema <$> handlers))
```
andys8, line 109: :information_source: Because of rule defined here

https://github.com/andys8/haskell-language-server/blob/ebb58eec6fb9817c163879800cf0c17dcb25e460/.hlint.yaml#L47

hie-compat/src-ghc86/Compat/HieDebug.hs
```diff
@@ -129,9 +129,7 @@ validateScopes asts = M.foldrWithKey (\k a b -> valid k a ++ b) [] refMap
     valid (Right n) refs = concatMap inScope refs
       where
         mapRef = foldMap getScopeFromContext . identInfo . snd
-        scopes = case foldMap mapRef refs of
-          Just xs -> xs
-          Nothing -> []
+        scopes = Data.Maybe.fromMaybe [] (foldMap mapRef refs)
```
michaelpj, line 132: FWIW I personally disagree with this hlint rule pretty strongly. If anything I tend to replace `fromMaybe` and friends with explicit pattern matches unless there's something higher-order going on!

andys8: Main issue I think that needs to be addressed is the difference between what is configured in `.htlint.yaml` and the codebase. I myself am trying to be not opinionated in this PR. But I'm also not sure regarding defaults and capabilities of different hlint versions - and how hint levels are dealt with by different tools (hls, the github action in this repository and together with apply-refact). Currently I'm still looking into it.

michaelpj:
In terms of getting our codebase clean: we have some `hlint` rules set to `error`, and those _will_ fail the build, and hence they have a bunch of exceptions. Reducing the number of exceptions would be great: in that way the "error" rules are the ones we're really saying we don't want to happen ever.

I'm also very pro getting rid of warnings and pruning any rules that don't add value :+1:

drsooch (approved):
Lgtm on number formatting plugin

Issue #628, opened 2020-11-25
"Replace with" action removes ticks from TemplateHaskellQuotes usage


expipiplus1:
Oops, I've misspelled `break`

```haskell
{-# LANGUAGE TemplateHaskellQuotes #-}

module Foo where

import Language.Haskell.TH (Name)

foo :: Name
foo = 'bread
```

HLS suggests `Replace with ‘break’` (weird quotes btw), perfect:

```haskell
foo :: Name
foo = break
```

oh no, `Couldn't match expected type ‘Name’ with actual....`

jneira:
mmm I guess this one will not be too difficult, it could be related with the fact that in windows for older ghcs the identifiers in error msgs was surrounded by 'id```

sloorush:
Hey @jneira, I was trying to find a solution to this issue; I have managed to successfully use my hacked HLS. 

Could you guide me on where to look to solve this issue?

jneira:
@sloorush hi, thanks for working in a fix for this, from 

>  I have managed to successfully use my hacked HLS.

i understand you have a branch with the fix, in that case i would open a pr against this repo master branch, noting it will close this issue

sloorush:
By that I mean i was successfully able to use my custom build of HLS but I am unable to understand how to solve this issue.

It would be really helpful if you could guide me on how to solve it.

bendo:
Will look at this issue.

expipiplus1:
Thanks @bendo

fendor:
fixed by #3260

PR #3260, opened 2022-10-07
Fix action removes ticks from TemplateHaskellQuotes (#628)

bendo:


michaelpj (approved):
Could we add a bit of commentary? Evidently what's going on is that *sometimes* GHC will give us... the wrong name, or something? Or is the issue that there's some information about what _type_ of name it is, or how it's used, that we should be detecting and passing along? I would sort of expect this information to be in the GHC AST... but maybe we're not using that.

Basically, I'm clueless, and I wonder if there's some knowledge we should be recording here.

bendo:
@michaelpj I've added comment with some example as well. Please review. Thanks

michaelpj:
Great, thanks!

PR #3261, opened 2022-10-07
Add suggestions about licenses in cabal file

dyniec:
Depends very much on #2954

Old approach depends on error message from [cabal](https://github.com/haskell/cabal/blob/master/Cabal-syntax/src/Distribution/SPDX/LicenseId.hs#L559). So it was only capable of suggetsions on input `BSD2` and `BSD3`.
This approach uses license list from `cabal-syntax` package, but still depends on error message from cabal (though it extracts only original name, and uses that to find similar one in supported list)

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3261"></a>


dyniec:
oops, I forgot to update docs for `licenseErrorSuggestion`

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/LicenseSuggest.hs
```diff
@@ -52,22 +56,25 @@ licenseErrorAction uri diag =
...
+  T.Text
   -- ^ Output of 'Ide.Plugin.Cabal.Diag.errorDiagnostic'
-  -> Maybe (T.Text, T.Text)
+  -> [(T.Text, T.Text)]
   -- ^ (Original (incorrect) license identifier, suggested replacement)
-licenseErrorSuggestion message =
-  mSuggestion message >>= \case
-    [original, suggestion] -> Just (original, suggestion)
-    _                      -> Nothing
+licenseErrorSuggestion msg = take 10 $
```
michaelpj, line 72: Should we use our fuzzy matching module that we use elsewhere? I think it's overkill in some senses since we don't have a big list of completions here, but I think it's somewhat nice to use the same mechanism everywhere?

fendor: What fuzzy matching module do we have? This PR uses `simpleFilter` to fuzzily find best matches in the list of SPDX identifiers.

michaelpj: Oh right. We have `Text.Fuzzy.Parallel`, but I guess that's only supposed to be a faster version of `Text.Fuzzy`. Never mind.

michaelpj (approved):

fendor:
Still lacks some redesign of the tests, will get to it, maybe next weekend.

michaelpj:
Sorry, not clear if that means you want to merge it or not? Also should maybe have some feature doc?

fendor:
CI is red, it can't be merged outright, no? And the CI is red because the test-suite fails because it suggests more than previously expected.

michaelpj:
Ah I see, I'll let you deal with it then :)

fendor:
CI is probably green now, ready for final review, I think.

plugins/hls-cabal-plugin/test/Main.hs
```diff
@@ -150,8 +156,39 @@ pluginTests recorder = testGroup "Plugin Tests"
           , "    build-depends:    base"
           , "    default-language: Haskell2010"
           ]
+    , runCabalTestCaseSession "Apache-2.0" recorder "" $ do
+        doc <- openDoc "licenseCodeAction2.cabal" "cabal"
+        diags <- waitForDiagnosticsFromSource doc "cabal"
+        -- test if it supports typos in license name, here 'apahe'
+        reduceDiag <- liftIO $ inspectDiagnostic diags ["Unknown SPDX license identifier: 'APAHE'"]
+        traceShowM (reduceDiag)
```
michaelpj, line 164: good hlint

michaelpj (changes requested):

plugins/hls-cabal-plugin/test/Main.hs
fendor: yeah, yeah, my mistake :D

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/LicenseSuggest.hs
fendor: I did actually opt to `Text.Fuzzy.Parallel`, looks a bit cleaner, thank you for your suggestion!

plugins/hls-cabal-plugin/hls-cabal-plugin.cabal
```diff
@@ -58,6 +59,7 @@ library
     , stm
     , text
     , unordered-containers  >=0.2.10.0
+    , fuzzy                 >=0.1
```
fendor, line 62:
```suggestion
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/LicenseSuggest.hs
```diff
@@ -6,22 +6,28 @@
 module Ide.Plugin.Cabal.LicenseSuggest
 ( licenseErrorSuggestion
 , licenseErrorAction
+, licenseNames
```
fendor, line 9:
```suggestion
```

michaelpj (approved):

Issue #2901, opened 2022-05-09
Plugin tutorial outdated


uhbif19:
Section about `Plugins are “linked” in the Main module, so we will need to add our plugin there once we have defined it` is incorrect. Now Plugins list is in Plugins.hs.

cgeorgii:
Hey, I'm taking a stab at this :)

fendor:
Finally fixed! And the tutorial will be typechecked now.

PR #3266, opened 2022-10-08
Update plugin tutorial

dyniec:
Addresses  #2901
Also, it was mentioning freenode, but now it is libera chat

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3266"></a>


pepeiborra (approved):
Thanks!

Issue #3135, opened 2022-09-02
More code actions when a default method implementation doesn't work


akshaymankar:
## Is your enhancement request related to a problem? Please describe.

When writing code like this:

```haskell
instance FromJSON Foo
```

I always get an error about `Foo` not having an instance for `Generic` and HLS sugests that I make the line look like

```haskell
instance Generic Foo => FromJSON Foo
```

This is almost never what anyone wants to do for most type classes. 

## Describe the solution you'd like

I understand that in the general case, this might be a good idea, like when the constraint is on a parameter of the type in question. So this suggestions shouldn't be removed. I think having these two other suggestions would be greatly helpful for me:
1. A suggestion to add placeholders for the missing methods, like the hls-class-plugin does.
2. A suggestion to add another instance for the type instead of just the constraint.

## Describe alternatives you've considered

N/A.

## Additional context

N/A.

wz1000:
This seems like a reasonable feature request. Is this something you're interested in trying to implement?

akshaymankar:
Yes I am interested in implementing it. I am planing to be at Munihac next Month and I am not sure if I will find time to work on this until then. If nobody else has taken stab at this by then I can hack on this there.

akshaymankar:
I see the label `component:hls-class-plugin` on this, I would like to point out that the current code action comes from ghcide, so this would definitely involve some work on ghcide.

PR #3267, opened 2022-10-08
class-plugin: Add code action for implementing instance method when default implementation doesn't work

akshaymankar:
Half of what #3135 proposes

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3267"></a>

### DIFF
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
@@ -5,7 +5,6 @@
 
 module Ide.Plugin.Class.CodeAction where
 
-import           Control.Applicative                  (liftA2)
 import           Control.Lens                         hiding (List, use)
 import           Control.Monad.Extra
 import           Control.Monad.IO.Class               (liftIO)
@@ -89,7 +88,7 @@ codeAction recorder state plId (CodeActionParams _ _ docId _ context) = pluginRe
         List diags = context ^. J.diagnostics
 
         ghcDiags = filter (\d -> d ^. J.source == Just "typecheck") diags
-        methodDiags = filter (\d -> isClassMethodWarning (d ^. J.message)) ghcDiags
+        methodDiags = filter (\d -> isClassMethodWarning (d ^. J.message) || isInstanceMissingError (d ^. J.message)) ghcDiags
 
         mkActions
             :: NormalizedFilePath
@@ -111,11 +110,17 @@ codeAction recorder state plId (CodeActionParams _ _ docId _ context) = pluginRe
                 $ use GetInstanceBindTypeSigs docPath
             implemented <- findImplementedMethods ast instancePosition
             logWith recorder Info (LogImplementedMethods cls implemented)
+            let groups =
+                    -- If the mininal def contains nothing, the plugin offers to
+                    -- add stub implementations of all methods with default
+                    -- implementations
+                    case classMinimalDef cls of
+                        And [] -> classOpItemsGroups range sigs cls
+                        Or []  -> classOpItemsGroups range sigs cls
+                        minDef -> minDefToMethodGroups range sigs minDef
             pure
                 $ concatMap mkAction
-                $ fmap (filter (\(bind, _) -> bind `notElem` implemented))
-                $ minDefToMethodGroups range sigs
-                $ classMinimalDef cls
+                $ fmap (filter (\(bind, _) -> bind `notElem` implemented)) groups
             where
                 range = diag ^. J.range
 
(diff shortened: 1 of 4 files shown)


July541:
Sorry for the delay, I'll review this on weekend.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -207,6 +213,9 @@ isClassNodeIdentifier ident = (isNothing . identType) ident && Use `Set.member`
 isClassMethodWarning :: T.Text -> Bool
 isClassMethodWarning = T.isPrefixOf "• No explicit implementation for"
 
+isInstanceMissingError :: T.Text -> Bool
+isInstanceMissingError = T.isPrefixOf "• No instance for"
```
Ailrun, line 217: What if this error comes from something not in an instance definition?
If I guess correctly, this will give error responses, which is not so desirable.
I think it's probably better to have a more specific way to detect this "invalid default impl" case.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -111,11 +111,17 @@ codeAction recorder state plId (CodeActionParams _ _ docId _ context) = pluginRe
                 $ use GetInstanceBindTypeSigs docPath
             implemented <- findImplementedMethods ast instancePosition
             logWith recorder Info (LogImplementedMethods cls implemented)
+            let groups =
+                    -- If the mininal def contains nothing, perhaps there are
+                    -- default implementations of methods, so the LSP offers to
+                    -- add stub implementations of all methods.
```
Ailrun, line 117: As we have `cls`, we don't need to guess whether there are default method impls or not. We can use `classOpItems` to get the information.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
akshaymankar: Hmm, yes this produces logs like `Warning | class: InternalError: No Identifier found`. I am not sure what is a better way to detect this error, here are two errors:

```
• No instance for (Test1 X)
    arising from a use of ‘DefaultImplementation.$dmf’
• In the expression: DefaultImplementation.$dmf @(X)
  In an equation for ‘f’: f = DefaultImplementation.$dmf @(X)
  In the instance declaration for ‘Test X’
```

```
• No instance for (Show a) arising from a use of ‘show’
  Possible fix:
    add (Show a) to the context of
      the type signature for:
        foo :: forall a. a -> String
• In the expression: show
  In an equation for ‘foo’: foo = show
```

generated from file:

```
{-# LANGUAGE DefaultSignatures #-}
module DefaultImplementation where

class Test1 a where

class Test a where
  f :: a -> a
  default f :: Test1 a => a -> a
  f = id

data X = X | Y


foo :: a -> String
foo = show

instance Test X where
```

Do you think trying to see if there is no "Possible fix" would be better? Or should the plugin just start expecting this other case and not throw errors when it doesn't find any class identifier?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
akshaymankar: Thanks! I was not aware of `classOpItems` :+1:

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
Ailrun: How about looking for the "In instance" part? That sounds more promising. And surely it'd be better if it doesn't report HLS errors for non-default instance resolution GHC errors.

fendor:
@akshaymankar Anything blocking this? If not, I am going to give it a quick review and then merge it ;D

akshaymankar:
> @akshaymankar Anything blocking this? If not, I am going to give it a quick review and then merge it ;D

I don't think so. Ship it if it looks good! :shipit:

fendor (approved):
LGTM, let's get it merged!

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -215,11 +225,19 @@ isInstanceValBind _                          = False
...
+mkGroups range sigs name =
+    [[ (T.pack . occNameString . occName $ name, bindRendered sig)
+     | sig <- sigs
+     , inRange range (getSrcSpan $ bindName sig)
+     , printOutputable name == T.drop (T.length bindingPrefix) (printOutputable (bindName sig))
+     ]]
+
+classOpItemsGroups :: Range -> [InstanceBindTypeSig] -> Class -> [[(T.Text, T.Text)]]
+classOpItemsGroups range sigs cls =
+    concatMap (mkGroups range sigs . getName . fst) $ classOpItems cls
```
Ailrun, line 243: These are not just ones with default impl, but all methods, aren't they?

Ailrun (changes requested):

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
Ailrun: BTW, how does this work for something like
```hs
data X = X | Y

class Test a where
  f :: a -> String

instance Test X where
  f = show
```
?

akshaymankar:
I forgot about this one, I'll rework it and send another PR.

PR #3271, opened 2022-10-08
Use an importance score to order the suggested import code action

### LINKED ISSUE #3234: Better ordering for import code actions

Suppose I use `Vector.toList` without importing it.  HLS highlights the error, and provides suggestions for where I might want to import `Vector.toList` from.  Unfortunately, the list of possible import locations is extremely long, and therefore so difficult to use that I might as well scroll up and type the import declaration myself.

Instead of listing *all* modules that export a `toList`, in alphabetical order, HLS would ideally notice that I'm attempting to qualify the import as `Vector.toList`, and prioritize modules with `Vector` in their name.  I suppose it should also recognize the common patterns of using prefixes or first letters of different module components (e.g., `Seq` for `Data.Sequence` or `LB` for `Data.ByteString.Lazy`.

In the greatest generality, I think what I'd imagine here is a heuristic scoring function that determines how likely it is that I want to import from any given module, and then ordering the suggestions by that.  Over time, this logic could be extended, for instance to prioritize modules that are already use in the project, or that belong to the same packages that are already used in this module, etc., and thus to function even for unqualified imports.  Inevitably, I suppose, someone will eventually try to replace the heuristic with a machine learning ranking model... but at least just matching the proposed qualifier for qualified imports looks like an obvious win.

One disadvantage is that alphabetical order makes things easier to find, if people do actually scroll through the entire list of suggestions.  Ways to mitigate this might include:

1. Only picking particularly good matches to promote to the top, but leaving most of the list in alphabetical order.
2. Only applying this heuristic when the list is beyond a certain length.  For example, more than a dozen or two elements makes the list unusable in VSCode because it doesn't actually scroll at all, but instead extends beyond the bottom of the screen where most of it is unusable.

ChristophHochrainer:
* fixes #3234
* creates an importance score function for import suggestions based on similarity and module length
* wraps the computed score and the previous result tuple (`CodeAction`, `NewImport`) into an  `ImportSuggestion`
* orders the `ImportSuggestion` first based on the score in descending order and second based on the import module names in alphabetic order

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3271"></a>

Before the PR changes:

image

After the PR changes:

image

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
@@ -38,7 +38,6 @@ import           Data.Ord                                          (comparing)
 import qualified Data.Set                                          as S
 import qualified Data.Text                                         as T
 import qualified Data.Text.Utf16.Rope                              as Rope
-import           Data.Tuple.Extra                                  (fst3)
 import           Development.IDE.Core.Rules
 import           Development.IDE.Core.RuleTypes
 import           Development.IDE.Core.Service
@@ -87,6 +86,7 @@ import           Language.LSP.Types                                (ApplyWorkspa
                                                                     uriToFilePath)
 import           Language.LSP.VFS                                  (VirtualFile,
                                                                     _file_text)
+import qualified Text.Fuzzy.Parallel                               as TFP
 import           Text.Regex.TDFA                                   (mrAfter,
                                                                     (=~), (=~~))
 #if MIN_VERSION_ghc(9,2,0)
@@ -99,7 +99,6 @@ import           GHC                                               (AddEpAnn (Ad
                                                                     EpaLocation (..),
                                                                     LEpaComment,
                                                                     LocatedA)
-
 #else
 import           Language.Haskell.GHC.ExactPrint.Types             (Annotation (annsDP),
                                                                     DeltaPos,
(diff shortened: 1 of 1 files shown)


fendor:
Hi! Please include a test, or an example how the behaviour is now improved!

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1495,35 +1494,63 @@ suggestNewImport packageExportsMap ps fileContents Diagnostic{_message}
...
     where
+        -- The importance score takes 2 metrics into account. The first being the similarity using
+        -- the Text.Fuzzy.Parallel.match function. The second is a factor of the relation between
+        -- the modules prefix import suggestion and the unknown identifier names.
+        importanceScore
+          | Just q <- qual
+          = let
+              similarityScore = fromIntegral $ unpackMatchScore (TFP.match (T.toLower q) (T.toLower m)) :: Double
+              maxLength = fromIntegral (max (T.length q) (T.length m)) :: Double
+              minLength = fromIntegral (min (T.length q) (T.length m)) :: Double
```
pepeiborra, line 1532:
```suggestion
              (maxLength, minLength) = case (T.length a, T.length b) of
                 (la, lb) 
                   | la >= lb -> (fromIntegral la, fromIntegral lb) 
                   | otherwise -> (fromIntegral lb, fromIntegral la)
```

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1495,35 +1494,63 @@ suggestNewImport packageExportsMap ps fileContents Diagnostic{_message}
...
+          where
+            unpackMatchScore pScore
+              | Just score <- pScore = score
+              | otherwise = 0
         m = moduleNameText identInfo
 
+-- | Implements a lexicographic order for import suggestions.
+-- First compares the importance score in DESCENDING order.
+-- If the scores are equal it compares the import names alphabetical order.
+data ImportSuggestion = ImportSuggestion (Int, CodeActionKind, NewImport)
```
pepeiborra, line 1546: Either make `ImportSuggestion` a new type or, preferably, unpack the tuple and mark the fields strict

```suggestion
data ImportSuggestion = ImportSuggestion !Int !CodeActionKind !NewImport
```

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1495,35 +1494,63 @@ suggestNewImport packageExportsMap ps fileContents Diagnostic{_message}
   , Just (range, indent) <- newImportInsertRange ps fileContents
   , extendImportSuggestions <- matchRegexUnifySpaces msg
     "Perhaps you want to add ‘[^’]*’ to the import list in the import of ‘([^’]*)’"
-  = sortOn fst3 [(imp, kind, TextEdit range (imp <> "\n" <> T.replicate indent " "))
-    | (kind, unNewImport -> imp) <- constructNewImportSuggestions packageExportsMap (qual <|> qual', thingMissing) extendImportSuggestions
-    ]
+  = let suggestions = nubSort
+          [ s | s <- constructNewImportSuggestions packageExportsMap (qual <|> qual', thingMissing) extendImportSuggestions ] in
```
pepeiborra, line 1498:
```suggestion
          (constructNewImportSuggestions packageExportsMap (qual <|> qual', thingMissing) extendImportSuggestions) in
```

fendor (approved):
LGTM, awesome!

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1521,35 +1520,65 @@ suggestNewImport packageExportsMap ps fileContents Diagnostic{_message}
...
+              | otherwise = 0
         m = moduleNameText identInfo
 
+-- | Implements a lexicographic order for import suggestions.
+-- First compares the importance score in DESCENDING order.
+-- If the scores are equal it compares the import names alphabetical order.
+data ImportSuggestion = ImportSuggestion !Int !CodeActionKind !NewImport
+  deriving ( Eq )
+
+instance Ord ImportSuggestion where
```
michaelpj, line 1577: I think this `Ord` instance is not lawful because it doesn't compare the `CodeActionKind`, which is taken into account by the `Eq` instance.

michaelpj (commented):
Were we going to get some tests?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
ChristophHochrainer: Thanks for catching this! I will fix this ASAP

PR #3276, opened 2022-10-09
Make eval plugin work at ghc 9.4

dyniec:
This is work in progress, but it works(given issues below) :)
It will  not compile for older ghc version, and so far it doesn't handle unit ids at all. 
It is hardcoded for package "blah" version 0.1.0.0 now. Proper solutions for those issues will show up in separate commits

It will need rebase when  #3249 lands.

Leaving it as draft until I get it to a better state

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3276"></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -177,7 +177,7 @@ jobs:
         name: Test hls-pragmas-plugin
         run: cabal test hls-pragmas-plugin --test-options="$TEST_OPTS" || cabal test hls-pragmas-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-pragmas-plugin --test-options="$TEST_OPTS"
 
-      - if: matrix.test && matrix.ghc != '9.4.2'
+      - if: matrix.test
         name: Test hls-eval-plugin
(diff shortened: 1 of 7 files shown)


dyniec:
Fixed hardcoded package name and adaded ifdefs guarding differences between 9.2 and 9.4. 
When I manage to confirm it works on earlier versions (8.8 and maybe 8.6?) I will either mark it as ready to review or patch this as well

michaelpj:
> 8.8 and maybe 8.6?

We just dropped support for 8.6 and 8.8 is going to be dropped soon, so don't bother. Does 8.10 work?

pepeiborra:
Don't forget enabling the testsuite in the `test.yml` file.

dyniec:
Ghc 8.10.7 seems to still work fine. Rebased on top of main commit, and enabled tests. I updated some tests for ghc94, but I have 6 left failing due to a difference in error format. I will try to figure that out over the weekend
Examples:
```
    expected: Just "-- Variable not in scope: noFunctionWithThisName"
     but got: Just "-- /home/dyniec/projects/haskell-language-server/plugins/hls-eval-plugin/test/testdata/T8.hs:2:2: error:\n--     Variable not in scope: noFunctionWithThisName"
```
and 
```
    --- Not in scope: type variable `a'
    +-- <interactive>:1:1: error: Not in scope: type variable `a'
```

PR #3281, opened 2022-10-11
Cleanup GHC macros (because min version is 8.8.4)

andys8:
With https://github.com/haskell/haskell-language-server/pull/3101 the minimum supported GHC version is `8.8.4`. This allows us to drop (or modify all conditional pragmas) targetting lower ghc versions or remove them since we already know GHC is min 8.8.4.

## Changes

I grepped through the codebase with  `"#if.*MIN_VERSION_ghc\(8,8"` and `"__GLASGOW_HASKELL__.*808"` and code that can now be removed.

## Review

* Merge after https://github.com/haskell/haskell-language-server/pull/3101 (merged :heavy_check_mark: )
* Context: https://github.com/haskell/haskell-language-server/pull/3101#issuecomment-1251708828

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3281"></a>


ghcide/src/Development/IDE/GHC/Compat/Core.hs
```diff
@@ -931,49 +920,22 @@ pattern L l a <- GHC.L (getLoc -> l) a
...
-         , Module.ml_obj_file = Module.addBootSuffix (Module.ml_obj_file locn)
-         }
-#else
 addBootSuffixLocnOut = Module.addBootSuffixLocnOut
-#endif
-
 
 dataConExTyCoVars :: DataCon -> [TyCoVar]
-#if __GLASGOW_HASKELL__ >= 808
 dataConExTyCoVars = DataCon.dataConExTyCoVars
```
michaelpj, line 938: I think some things like this could probably be inlined now?

ghcide/src/Development/IDE/GHC/Compat/Core.hs
```diff
@@ -1031,25 +993,15 @@ noExtField = GHC.noExt
 #endif
 
 ml_hie_file :: GHC.ModLocation -> FilePath
-#if !MIN_VERSION_ghc(8,8,0)
-ml_hie_file ml
-  | "boot" `isSuffixOf ` Module.ml_hi_file ml = Module.ml_hi_file ml -<.> ".hie-boot"
-  | otherwise  = Module.ml_hi_file ml -<.> ".hie"
-#else
 ml_hie_file = Module.ml_hie_file
```
michaelpj, line 996: ditto

ghcide/src/Development/IDE/GHC/Compat/Core.hs
```diff
@@ -1031,25 +993,15 @@ noExtField = GHC.noExt
...
-#endif
 
 #if !MIN_VERSION_ghc(9,0,0)
 pattern NotBoot, IsBoot :: IsBootInterface
 pattern NotBoot = False
 pattern IsBoot = True
 #endif
 
-#if MIN_VERSION_ghc(8,8,0)
 type PlainGhcException = Plain.PlainGhcException
```
michaelpj, line 1004: ditto

michaelpj (approved):
beautiful!

July541:
Should we drop related hie-compat also?

andys8:
> We're going to drop 8.8 also, do you want to wait for that before tidying up? I guess this is good regardless!

Yeah, I was asking about [dropping 8.8 support](https://github.com/haskell/haskell-language-server/pull/3101#issuecomment-1254762784) but since 8.6 and 8.8 are split in several PRs, I would also split tidying up the macros. First getting rid of 8.6, then getting rid of 8.8.

ghcide/src/Development/IDE/GHC/Compat/Core.hs
andys8: Thanks for these hints :+1:

isovector (approved):

plugins/hls-eval-plugin/test/Main.hs
```diff
@@ -139,15 +139,6 @@ tests =
   , goldenWithEval "Don't panic on {-# UNPACK #-} pragma" "TUNPACK" "hs"
   , goldenWithEval "Can handle eval inside nested comment properly" "TNested" "hs"
   , goldenWithEval "Test on last line insert results correctly" "TLastLine" "hs"
-  , testGroup "with preprocessors"
```
michaelpj, line 142: hang on, what happened here? Aren't these _broken_ in 8.6? So we should instead just be able to enable them?

andys8: Oooohh, that would obviously be better, yeah -> d9a2c395e69ef59db70e42b1eb55f4d9fc2730a3

plugins/hls-eval-plugin/test/Main.hs
```diff
@@ -140,13 +140,8 @@ tests =
   , goldenWithEval "Can handle eval inside nested comment properly" "TNested" "hs"
   , goldenWithEval "Test on last line insert results correctly" "TLastLine" "hs"
   , testGroup "with preprocessors"
-    [ knownBrokenInEnv [HostOS Windows, GhcVer GHC86]
```
michaelpj, line 143: You were a little too zealous - we still need to say it's broken on Windows.

andys8: Assumed it was `and` not `or` 😅 Good that there is CI for all cases.

PR #3284, opened 2022-10-11
Register Fourmolu plugin properties

georgefst:
One effect of this is that `haskell-language-server vscode-extension-schema` now generates the schema for this plugin. I'm unsure if it has any other effect, as this is all quite under-documented (see #2827).


michaelpj (approved):

michaelpj:
TBH I also don't know how this works otherwise I'd have documented it by now :)

PR #3300, opened 2022-10-23
Make a test more reliable

pepeiborra:
"iface-error-test-1" fails occasionally in CI. The problem is that progress reporting and diagnostics arrive in non-deterministic ordering: a test that waits for both is creating a race condition.

Ideally we would solve the non-determinism problem for all tests, but I'm not sure how to do that. In the meantime, we should review all tests in the ghcide test suite. When fixing a test, prefer diagnostics over progress reporting since the former are more reliable.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3300"></a>


pepeiborra:
This needs #3301 to work properly

fendor (approved):
LGTM, thank you!

pepeiborra:
I'm planning to add a comment and land this but it might take me a few weeks

PR #3301, opened 2022-10-23
Record diagnostics source rule when testing

pepeiborra:
I have seen unexpected duplicate diagnostics leading to a failure in the ghcide testsuite: what seems to happen here is that both `Typecheck` and `GetModIfaceFromDisk` contribute the same diagnostics. But this should never happen, because we never run both these rules for the same file:
* `GetModIfaceFromDisk` is only run for non FOIs via `GetModIface`
* `Typecheck` is only run for FOIs, either via `GetModIface` or on demand by code actions 

This PR adds more testing-only logging to help diagnostics when this happens again next time

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3301"></a>

### DIFF
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -1233,8 +1233,18 @@ updateFileDiagnostics recorder fp ver k ShakeExtras{diagnostics, hiddenDiagnosti
                  return action
     where
         diagsFromRule :: Diagnostic -> Diagnostic
-        diagsFromRule c
-            | coerce ideTesting = c{_code = Just $ InR $ T.pack(show k)}
+        diagsFromRule c@Diagnostic{_range}
+            | coerce ideTesting = c
+                {_relatedInformation =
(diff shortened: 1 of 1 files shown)


fendor (approved):
Nice addition

michaelpj (approved):
Slightly an abuse of the `source` field but I guess it's fine. We could put it in the `data` field? which is entirely implementation-defined. Then we could consider even doing it unconditionally, or having some richer-structured "provenance" structure.

pepeiborra:
> Slightly an abuse of the `source` field but I guess it's fine. We could put it in the `data` field? which is entirely implementation-defined. Then we could consider even doing it unconditionally, or having some richer-structured "provenance" structure.

Do you mean the `code` field? I like the idea, but it doesn't look like diagnostics have a `data` field

michaelpj:
No, `code` is for the error code, which we might actually get from GHC before long. Diagnostics do have a data field, but not in the version of the lsp library we're using, I guess.

Issue #3210, opened 2022-09-22
Use AnnotatedTextEdits throughout HLS


michaelpj:
`AnnotatedTextEdit`s are a feature of the LSP spec that allow more structured edits to be returned. In particular:
- Multiple edits can be associated with a single _annotation_
- The annotation can have a name and a description
- The annotation can also indicate that the edit requires confirmation from the user

I haven't tested out what this looks like in the client, but in principle this should allow much nicer communication of what's going on with edits. For example, when adding imports for completions, we can annotate the edits for modifying imports to say what we're doing. The confirmation feature (if it actually works) seems nice for questionable or wide-ranging edits, like for renaming across modules.

We should try using them! 

A good place to start would be here: https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs#L202, which sends a message to the user to tell them what we did. Hopefully an `AnnotatedTextEdit` would do that job better.

nilsalex:
I would be interested in working on this!

michaelpj:
See https://github.com/microsoft/vscode/issues/151240

michaelpj:
We will need to not send these if clients don't understand them. I had an idea for that. We can always translate an `AnnotatedTextEdit` into a `TextEdit`. So just before sending the response, we can check if the client supports what we're sending and downgrade it if it doesn't support that. I guess we can do this in `combineResponses` although that's somewhat gross.

We could potentially do this with something like this:
```
class Conform (m :: Method t k) where
  type Caps :: *
  caps :: Lens' ClientCapabilities Caps
  conform :: Caps -> ResponseResult m -> ResponseResult m
```
i.e. for a given method, given the capabilities relevant to that method, try to make the result we produced acceptable to the client by downgrading things.

This could probably live in `lsp`... and here's the issue I wrote, even: https://github.com/haskell/lsp/issues/367

nilsalex:
I played around with it a bit @michaelpj, switching to AnnotatedTextEdits for import list extensions (the place you suggested above).

Without requiring confirmation, I could not see any visual difference, neither in VS Code, nor with coc.nvim. With confirmation required, VS Code shows a list of changes before optionally applying them. I'll push my experiments to my fork.

michaelpj:
That's a shame. I wish there was some guide in the spec showing what editors support what features, and what they actually do...

PR #3304, opened 2022-10-24
New plugin: Explicit record fields

ozkutuk:
#### What?
This is a plugin to expand record wildcards, explicitly listing all fields as field puns. [Here](https://asciinema.org/a/gQJW6rlqeEmbQnJ4ASvBw6KNL) is a little demo. It works in both record construction and pattern binding scenarios, and it works as you would expect regardless of whether there are explicitly provided fields or puns in addition to the wildcard.

#### Known issues
One of the shortcomings of the current approach is that all fields of the record are expanded, whether they are actually used or not. This results in warnings of unused bindings, if the corresponding warning flag is enabled. I am looking for ways to expand only used fields, but I still think this is useful as it is, hence the PR :slightly_smiling_face:

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -38,12 +38,13 @@ import           Development.IDE.GHC.Compat      (HasSrcSpan (..),
                                                   HsRecFields (..), HscEnv (..),
                                                   LPat, Outputable, SrcSpan,
                                                   pm_mod_summary, unLoc)
-import           Development.IDE.GHC.Compat.Core (GenLocated (..), GhcPass (..),
+import           Development.IDE.GHC.Compat.Core (Extension (NamedFieldPuns),
+                                                  GhcPass (..),
                                                   HsExpr (RecordCon, rcon_flds),
-                                                  HsRecField' (..), LHsExpr,
-                                                  ModSummary (..), Pass (..),
-                                                  Pat (..), extensionFlags,
-                                                  hs_valds)
+                                                  LHsExpr, ModSummary (..),
+                                                  Pass (..), Pat (..),
+                                                  extensionFlags, hfbPun,
+                                                  hs_valds, mapLoc)
 import           Development.IDE.GHC.Compat.Util (toList)
 import           Development.IDE.GHC.Util        (printOutputable)
 import           Development.IDE.Graph           (RuleResult)
@@ -53,7 +54,6 @@ import           Development.IDE.Spans.Pragmas   (NextPragmaInfo (..),
                                                   insertNewPragma)
 import           Development.IDE.Types.Logger    (cmapWithPrio)
 import           GHC.Generics                    (Generic)
-import           GHC.LanguageExtensions.Type     (Extension (..))
 import           Ide.PluginUtils                 (getNormalizedFilePath,
                                                   handleMaybeM, pluginResponse,
                                                   subRange)
@@ -79,14 +79,9 @@ import qualified Language.LSP.Types.Lens         as L
 -- the records that originally had wildcards with dots, even after they
 -- are removed by the renamer pass. Here `rec_dotdot` is set to
 -- `Nothing` so that fields are printed without such post-processing.
-preprocessRecord :: HsRecFields p arg -> HsRecFields p arg
+preprocessRecord :: HsRecFields (GhcPass c) arg -> HsRecFields (GhcPass c) arg
 preprocessRecord flds = flds { rec_dotdot = Nothing , rec_flds = rec_flds' }
   where
-    -- TODO(ozkutuk): HsRecField' is renamed to HsFieldBind in GHC 9.4
-    -- Add it as a pattern synonym to the ghcide compat module, so it would
-    -- work on all HLS builds.
-    -- https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5757
-
     no_pun_count = maybe (length (rec_flds flds)) unLoc (rec_dotdot flds)
     -- Field binds of the explicit form (e.g. `{ a = a' }`) should be
     -- left as is, hence the split.
@@ -94,16 +89,16 @@ preprocessRecord flds = flds { rec_dotdot = Nothing , rec_flds = rec_flds' }
     -- `hsRecPun` is set to `True` in order to pretty-print the fields as field
     -- puns (since there is similar mechanism in the `Outputable` instance as
     -- explained above).
-    puns' = map (\(L ss fld) -> L ss (fld { hsRecPun = True })) puns
+    puns' = map (mapLoc (\fld -> fld { hfbPun = True })) puns
     rec_flds' = no_puns <> puns'
 
-showRecordPat :: Outputable (Pat p) => Pat p -> Maybe Text
+showRecordPat :: Outputable (Pat (GhcPass c)) => Pat (GhcPass c) -> Maybe Text
 showRecordPat pat@(ConPat _ _ (RecCon flds)) =
   Just $ printOutputable $
     pat { pat_args = RecCon (preprocessRecord flds) }
 showRecordPat _ = Nothing
 
-showRecordCon :: Outputable (HsExpr p) => HsExpr p -> Maybe Text
+showRecordCon :: Outputable (HsExpr (GhcPass c)) => HsExpr (GhcPass c) -> Maybe Text
 showRecordCon expr@(RecordCon _ _ flds) =
   Just $ printOutputable $
     expr { rcon_flds = preprocessRecord flds }
@@ -239,14 +234,14 @@ codeActionProvider ideState pId (CodeActionParams _ _ docId range _) = pluginRes
         mkTextEdit :: RenderedRecordInfo -> Maybe TextEdit
         mkTextEdit (RenderedRecordInfo ss r) = TextEdit <$> srcSpanToRange ss <*> pure r
 
-        -- TODO(ozkutuk): `RecordPuns` extension is renamed to `NamedFieldPuns`
-        -- in GHC 9.4, so I probably need to add this to the compat module as
-        -- well.
+        -- NOTE(ozkutuk): `RecordPuns` extension is renamed to `NamedFieldPuns`
+        -- in GHC 9.4, but we still want to insert `NamedFieldPuns` in pre-9.4
+        -- GHC as well, hence the replacement.
         -- https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6156
         pragmaEdit :: Maybe TextEdit
-        pragmaEdit = if RecordPuns `elem` exts
+        pragmaEdit = if NamedFieldPuns `elem` exts
                        then Nothing
-                       else Just $ patchExtName $ insertNewPragma pragma RecordPuns
+                       else Just $ patchExtName $ insertNewPragma pragma NamedFieldPuns
           where
             patchExtName = L.newText %~ T.replace "Record" "NamedField"
 
@@ -257,7 +252,7 @@ codeActionProvider ideState pId (CodeActionParams _ _ docId range _) = pluginRes
 
 mkCodeActionTitle :: [Extension] -> Text
 mkCodeActionTitle exts =
-  if RecordPuns `elem` exts
+  if NamedFieldPuns `elem` exts
     then title
     else title <> " (needs extension: NamedFieldPuns)"
     where
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -13,7 +13,7 @@ module Ide.Plugin.ExplicitFields
   ) where
 
 import           Control.Arrow                   ((&&&))
-import           Control.Lens                    ((^.))
+import           Control.Lens                    ((%~), (^.))
 import           Control.Monad.IO.Class          (MonadIO, liftIO)
 import           Control.Monad.Trans.Except      (ExceptT)
 import           Data.Data                       (Data)
@@ -276,7 +276,9 @@ codeActionProvider recorder ideState pId (CodeActionParams _ _ docId range _) = 
         pragmaEdit :: Maybe TextEdit
         pragmaEdit = if RecordPuns `elem` exts
                        then Nothing
-                       else Just $ insertNewPragma pragma RecordPuns
+                       else Just $ patchExtName $ insertNewPragma pragma RecordPuns
+          where
+            patchExtName = L.newText %~ T.replace "Record" "NamedField"
 
     mkWorkspaceEdit :: NormalizedFilePath -> [TextEdit] -> WorkspaceEdit
     mkWorkspaceEdit nfp edits = WorkspaceEdit changes Nothing Nothing
@@ -287,7 +289,7 @@ mkCodeActionTitle :: [Extension] -> Text
 mkCodeActionTitle exts =
   if RecordPuns `elem` exts
     then title
-    else title <> " (needs extension: " <> (T.pack $ show RecordPuns) <> ")"
+    else title <> " (needs extension: NamedFieldPuns)"
     where
       title = "Expand record wildcard"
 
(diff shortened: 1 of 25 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+                                                  Method (..), SMethod (..),
+                                                  TextEdit (..),
+                                                  WorkspaceEdit (WorkspaceEdit),
+                                                  fromNormalizedUri,
+                                                  normalizedFilePathToUri,
+                                                  type (|?) (InR))
+import qualified Language.LSP.Types.Lens         as L
+
+
+data Log = LogShake Shake.Log
```
michaelpj, line 78: I immediately wonder: do we not want to log anything ourselves?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+      , _command = Nothing
+      , _xdata = Nothing
+      }
+      where
+        edits = catMaybes [ mkTextEdit rec , pragmaEdit ]
+
+        mkTextEdit :: RenderedRecordInfo -> Maybe TextEdit
+        mkTextEdit (RenderedRecordInfo ss r) = TextEdit <$> srcSpanToRange ss <*> pure r
+
+        -- NOTE(ozkutuk): `RecordPuns` extension is renamed to `NamedFieldPuns`
```
michaelpj, line 116: This seems like arguably it should be logic inside `insertNewPragma` or similar? Presumably anyone else who might try to insert this pragma also needs to be aware of this, which suggests pushing it upstream.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+    mkWorkspaceEdit :: NormalizedFilePath -> [TextEdit] -> WorkspaceEdit
+    mkWorkspaceEdit nfp edits = WorkspaceEdit changes Nothing Nothing
+      where
+        changes = Just $ HashMap.singleton (fromNormalizedUri (normalizedFilePathToUri nfp)) (List edits)
+
+    mkCodeActionTitle :: [Extension] -> Text
+    mkCodeActionTitle exts =
+      if NamedFieldPuns `elem` exts
+        then title
+        else title <> " (needs extension: NamedFieldPuns)"
```
michaelpj, line 136: Cute, but I think this is inconsistent UX: we never normally say in the code action title if we're going to insert pragmas. I think the right way to do this consistently is `AnnotatedTextEdit`s : https://github.com/haskell/haskell-language-server/issues/3210

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+
+collectRecordsRule :: Recorder (WithPriority Log) -> Rules ()
+collectRecordsRule recorder = define (cmapWithPrio LogShake recorder) $ \CollectRecords nfp -> do
+  tmr <- use TypeCheck nfp
+  let exts = getEnabledExtensions <$> tmr
+      recs = getRecords <$> tmr
+      renderedRecs = mapMaybe renderRecordInfo <$> recs
+  pure ([], CRR <$> renderedRecs <*> exts)
+
+getEnabledExtensions :: TcModuleResult -> [GhcExtension]
```
michaelpj, line 148: Hmm, I also wonder if this function is not already written somewhere?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+  collectRecords valBinds
+
+data CollectRecords = CollectRecords
+                    deriving (Eq, Show, Generic)
+
+instance Hashable CollectRecords
+instance NFData CollectRecords
+
+data CollectRecordsResult = CRR
+  { recordInfos       :: ![RenderedRecordInfo]
```
michaelpj, line 162: Making a field of type `[a]` strict typically does very little: all it does is force the outermost cons constructor, while the rest of the spine of the list will not be forced. If you want the container to be strict, do this and use a `Vector` or similar.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+  deriving (Generic)
+
+instance NFData CollectRecordsResult
+
+instance Show CollectRecordsResult where
+  show _ = "<CollectRecordsResult>"
+
+type instance RuleResult CollectRecords = CollectRecordsResult
+
+newtype GhcExtension = GhcExtension { unExt :: Extension }
```
michaelpj, line 174: ... why? Maybe I'll find out by inferring backwards, but you know already, so write it down in the haddock!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+instance NFData RenderedRecordInfo
+
+renderRecordInfo :: RecordInfo -> Maybe RenderedRecordInfo
+renderRecordInfo (RecordInfoPat ss pat) = RenderedRecordInfo ss <$> showRecordPat pat
+renderRecordInfo (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordCon expr
+
+-- `Outputable` instance of `HsRecFields` does smart things to print
+-- the records that originally had wildcards with dots, even after they
+-- are removed by the renamer pass. Here `rec_dotdot` is set to
+-- `Nothing` so that fields are printed without such post-processing.
```
michaelpj, line 198: why is that bad? maybe give an example!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+collectRecords' ideState =
+  handleMaybeM "Unable to TypeCheck"
+    . liftIO
+    . runAction "ExplicitFields" ideState
+    . use CollectRecords
+
+collectRecordsInRange :: MonadIO m => Range -> IdeState -> NormalizedFilePath -> ExceptT String m CollectRecordsResult
+collectRecordsInRange range ideState nfp = do
+  CRR renderedRecs exts <- collectRecords' ideState nfp
+  pure $ CRR (filter inRange renderedRecs) exts
```
michaelpj, line 253: Just thinking through how this works: we collect all the record stuff in the module and cache it with a rule, good. But then we do a linear amount of work every time we get a code action request filtering them to see if they're in range. Maybe that's fine, but I could imagine this blowing up in modules with a lot of record stuff.

A cheap improvement would be to do some kind of binary search for the range filtering. Have a look at the code-ranges plugin, which I think does something similar.

michaelpj (approved):
Nice! Generally very clean. I could have even more comments and logging, but I'm a bit of a fanatic in that direction ;)

We would also need to update `features.md` and the plugin support page.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I newtype'd it so that I can provide an `NFData` instance (which is required for types returned from rules) without creating an orphan instance. Do you think there is a better way to do this? If this is sensible, I can put this rationale in a comment.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I made it strict out of habit, I don't think we necessarily need to make this container strict. I will remove the strictness annotation to avoid future confusion.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I have adapted this from the alternate number format plugin, as seen [here](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs#L124-L130). I do agree that `AnnotatedTextEdit`s seem to be a nicer way to do this though. I will look into it to see if we can make use of it here.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: What would be a useful thing to log in your opinion? I have been logging the list of records collected from a given source file for debugging during development, but later removed once I got the plugin working correctly. Do you think it would be a useful thing to provide still?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I don't think there is one, but this code is duplicated among several plugins so I think it a good idea to write one. Where would be a good place to add this definition?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: I usually just reach for `StrictData` for that use case. Here it stands out, which made me think it might have a specific intention behind it!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: No pressure to use it in this PR - AFAIK we use it nowhere yet. But I'd be delighted if someone looked into it!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: Well, that's what debug logging is for. I can't remember if we have a level below debug at the minute, I'm not sure we do.

I think it's helpful to imagine some user scenarios and consider where logging might help. Probably we'd want to make it debug logging still since otherwise it'll be very noisy. Some ideas:
- I thought I should be able to use the action here but I can't? Might be helpful to see "Found X record usages. No record usages in range for this request" or something.
- I used the action but it did the wrong thing somehow. This is probably covered by the requests (and hence the edit) being logged already.
- ???

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: Not sure. We don't have a great home for it at the minute, admittedly. There might be somewhere in ghcide.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: No, that's fine, it just needs a comment explaining that.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I did some testing by converting this to use `AnnotatedTextEdit`. It looks like it is a pretty new addition to spec as the editors I tried didn't properly support them. Neovim LSP didn't care, and just handled it like a usual `TextEdit` without the `ChangeAnnotation` niceties. VSCode didn't display the code action at all.

The spec states that servers shouldn't send `AnnotatedTextEdit`s if the client doesn't support them. Therefore wherever we use them, we need to query the client capabilities and fallback to plain `TextEdit`s if needed. Adding some convenience functions to take care of these kinds of details sounds like a good idea, but seems out-of-scope for this PR, so I am leaving this as is.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -0,0 +1,266 @@
...
+
+  where
+    inRange :: RenderedRecordInfo -> Bool
+    inRange (RenderedRecordInfo ss _) = maybe False (subRange range) (srcSpanToRange ss)
+
+-- Copied from hls-alternate-number-format-plugin
```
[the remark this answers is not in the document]
ozkutuk: I have moved the definition of `getFirstPragma` to `ghcide/src/Development/IDE/Spans/Pragmas.hs` and replaced the duplications among plugins to use this function.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: Hmm, that's disappointing, I would have expected vscode to do something :thinking:

ozkutuk: related: https://github.com/microsoft/vscode/issues/151240

pepeiborra (approved):
Very nice. There are a few things missing:
1. You forgot to add the test suite to the Github CI, 
2. The plugin should be listed somewhere under `docs`

Other than that, this looks fine to me

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I have made some changes to the filtering logic: Now the rule builds and caches an `IntervalMap` rather than a list, so the filtering should be more efficient. Can you take a look to see if the changes are reasonable?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I have added the list of collected records with their source spans to the debug log. I think this should cover the potential scenarios of "it didn't do what I expected it to".

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: Looks like `hw-prim` package (transitively depended on by `hw-fingertree`) doesn't yet work with GHC 9.4 though.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I have moved it to `ghcide/src/Development/IDE/GHC/Util.hs`, and replaced ad-hoc definitions among different plugins to use this one.

michaelpj:
The windows failures are in your testsuite and seem genuine, maybe some path issues.

ozkutuk:
> The windows failures are in your testsuite and seem genuine, maybe some path issues.

@michaelpj Do you have any idea how I might tackle this? The error message mentions a `clang` related problem, and I honestly have no idea what it might be related to. Unfortunately I don't have a windows box to test this myself. Has something like this happened with other plugins before?

michaelpj:
Sorry, I spoke too soon. I thought it was a problem with your tests looking up paths, but indeed it's some clang thing, so probably indeed random toolchain issues :( I'll rerun again...

michaelpj:
Hmmm... I notice that the plugin name is long and so is the name of the test component, such that the path it can't find is extremely long. I know we've had path length issues before, and I notice that some of the other plugins with long names have short names for their test components.

What happens if you rename your test component to something short like "tests"?

michaelpj:
hooray, thanks for sticking with it!

michaelpj:
Ah, one more thing: can you add yourself to CODEOWNERS for the plugin, please?

Issue #2779, opened 2022-03-14
Difficulties with integration with source plugins


edsko:
Consider this minimal source plugin which whenever it encounters the expression `False` checks if `ScopedTypeVariables` are enabled, and if not, throws an error:

```haskell
{-# LANGUAGE CPP #-}

module SayYes (plugin) where

import Control.Monad
import Data.Data
import Data.Generics
import Language.Haskell.TH (Extension(..))

#if __GLASGOW_HASKELL__ < 900
import Bag
import ErrUtils
import GHC
import GhcPlugins
#else
import GHC.Data.Bag
import GHC.Hs
import GHC.Plugins
import GHC.Utils.Error
#endif

plugin :: Plugin
plugin = defaultPlugin { parsedResultAction = \_ _ -> action }

action :: HsParsedModule -> Hsc HsParsedModule
action parsed@HsParsedModule{hpm_module = m} =
    (\m' -> parsed {hpm_module = m'}) <$>
      gmapM (everywhereM (mkM sayYes)) m

sayYes :: HsExpr GhcPs -> Hsc (HsExpr GhcPs)
sayYes e = do
   case e of
     HsVar _ (L l x) | occNameString (rdrNameOcc x) == "False" -> do
       dynFlags <- getDynFlags
       unless (ScopedTypeVariables `xopt` dynFlags) $
         issueError l $ text "Everyone needs scoped type variables"
       return $ e
     _otherwise  ->
       return $ e

issueError :: SrcSpan -> SDoc -> Hsc a
issueError l errMsg = do
    dynFlags <- getDynFlags
    throwOneError $
      mkErrMsg dynFlags l neverQualify errMsg

issueWarning :: SrcSpan -> SDoc -> Hsc ()
issueWarning l errMsg = do
    dynFlags <- getDynFlags
    liftIO $ printOrThrowWarnings dynFlags . listToBag . (:[]) $
      mkWarnMsg dynFlags l neverQualify errMsg
```

Example use case:

```haskell
-- {-# LANGUAGE ScopedTypeVariables #-}

{-# OPTIONS_GHC -fplugin=SayYes #-}

module Main where

main :: IO ()
main = print False
```

If we compile this on the command line (say, with `cabal`), we get

```
app/Main.hs:8:14: error: Everyone needs scoped type variables
  |
8 | main = print False
  |              ^^^^^
```

unless we enable the pragma. All good. 

If we open a project with HLS and open `Main.hs`, there are two orthogonal problems:

[...]

PR #3309, opened 2022-10-30
Restore ability to run source plugins

### LINKED ISSUE #3299: HLS ignores GHC plugins
### Your environment

Ubuntu (under Windows 11 WSL)
ghc 9.4.2 via ghcup

Reproduction repo with minimal example:
https://github.com/JakobBruenker/plugin-test

VS Code, but running haskell-language-server outside of it produces the same error

1.8.0.0 via ghcup

Have you configured HLS in any way (especially: a hie.yaml file)?
No (though an hie.yaml file generated via implicit-hie seemed to make no difference)

### Steps to reproduce

Clone the reproduction repo linked above and run `haskell-language-server` inside `just-some-project`.

NB: Because of #3297, this does not work if you run HLS in the top-level directory `plugin-test`.

### Expected behaviour

There should be no errors, since the plugin replaces `puStrLn` with `putStrLn`.

### Actual behaviour

HSL shows the error
```
  Variable not in scope: puStrLn :: String -> IO ()
  Suggested fix: Perhaps use ‘putStrLn’ (imported from Prelude)
```
Even though the plugin changes the name to `putStrLn`, which means `cabal run` works just fine.

Strangely, I seem to remember doing something similar 6 months ago and it worked, but now I can't get it to work with 9.2 either, but it may have been a different kind of plugin.

JakobBruenker:
Since ghc 9.0, plugins are stored in the HscEnv, not in the DynFlags. This caused HLS not to run source plugins anymore. This commit fixes that.

Fixes #3299.

It also fixes one of the two problems mentioned in #2779, which is that the plugin wasn't running in HLS. It does not fix the problem where HLS shows the wrong location for the error message described in that issue.

If I should add a test it would be helpful to know which test-data directory it should go into, and which files I have to look at to make the test run.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3309"></a>


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -255,7 +255,7 @@ getParsedModuleRule recorder =
   define (cmapWithPrio LogShake recorder) $ \GetParsedModule file -> do
     ModSummaryResult{msrModSummary = ms'} <- use_ GetModSummary file
     sess <- use_ GhcSession file
-    let hsc = hscEnv sess
+    (ms', hsc) <- liftIO $ initPlugins (hscEnv sess) ms'
```
pepeiborra, line 258: It occurs to me that in order to always initialise plugins consistently, the initialisation should be done in the `ModSummary` rule, the  initialized `HscEnv` should be stored in a field of `ModSummaryResult`, and all rules downstream should use it instead of asking for a `GhcSession`. 

This refactoring (or something of the sort) will prevent a bunch of future bugs.

pepeiborra:
Thank you! This is great stuff.

> If I should add a test it would be helpful to know which test-data directory it should go into, and which files I have to look at to make the test run.

The right place for a test would be the ghcide test suite. You can take inspiration on the existing plugin tests, as well to reenable the ones that are disabled for GHC 9.x

pepeiborra (changes requested):
Changes requested: we need tests to avoid breaking this stuff in the future

fendor:
You can enable the test here: https://github.com/haskell/haskell-language-server/blob/caa5ce55c49049632c62018fd7eac1b8dedd76bd/ghcide/test/exe/Main.hs#L1190

ghcide/src/Development/IDE/GHC/Compat/Plugins.hs
```diff
@@ -67,6 +68,12 @@ initializePlugins env = do
     pure $ hscSetFlags newDf env
 #endif
 
+-- Plugins aren't stored in ModSummary anymore since GHC 9.0, but this
```
fendor, line 71: Use Haddock comments, this way we can see the documentation just by hovering over the identifier:

```suggestion
-- | Plugins aren't stored in ModSummary anymore since GHC 9.0, but this
```

JakobBruenker: I also just noticed that it was changed in 9.2, not 9.0, so I'll have to update the comment

JakobBruenker:
Before I address your comments: I noticed that the ability for parser plugins to change error messages, introduced in 9.4, wasn't supported by HLS. [This new commit](https://github.com/haskell/haskell-language-server/pull/3309/commits/0e35e5297277ad9ec3bff63ad8d7f0076e57d916) fixes that. I haven't had time yet to test it with every supported GHC version, however.

I'm still seeing an `expectJust` crash ([here](https://hackage.haskell.org/package/ghc-9.4.2/docs/src/GHC.Linker.Loader.html#local-6989586621681514211)) when running HLS on one of my tests cases, so I'll have to figure out why that is happening still.

m4dc4p:
I ran cabal test (using latest master) and they all passed. Can we get this merged? I'd really like to have this functionality back.

fendor:
@JakobBruenker Do you think you have time to push this over the finish line for #3409?

ghcide/src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -357,6 +357,12 @@ data ModSummaryResult = ModSummaryResult
   { msrModSummary  :: !ModSummary
   , msrImports     :: [LImportDecl GhcPs]
   , msrFingerprint :: !Fingerprint
+  , msrHscEnv      :: !HscEnv
+  -- ^ HscEnv for this particular ModSummary.
+  -- Contains initialised plugins, parsed options, etc...
+  --
+  -- Implicit assumption: DynFlags in 'msrModSummary' are the same as
+  -- the DynFlags in 'msrHscEnv'.
```
fendor, line 365: This is inaccurate, later in the pipeline, dozens of functions modify the 'DynFlags' from 'msrModSummary'. This comment will be changed to basically telling users to prefer the DynFlags from msrModSummary instead of from this HscEnv. Does this seem sensible?

wz1000 (approved):

JakobBruenker:
@fendor sorry, didn't see your comment yesterday. Sorry to have this let sit idle, I'm happy to see it done!

fendor:
@JakobBruenker Dont be sorry, you did a valuable contribution, and I am genuinely thankful for that! I didn't give you any time to react, happy that you are not mad at me for taking over!

Issue #3280, opened 2022-10-10
Don't show hovers for typeclass dictionaries (or other generated names)


drone-rites:
## Your environment 

Which OS do you use?
macOS

Which version of GHC do you use and how did you install it?
9.2.2 via Stack project

How is your project built (alternative: link to the project)?
Stack. 
The project is available here: https://github.com/kudu-dynamics/blaze

Which LSP client (editor/plugin) do you use?
emacs+lsp-mode

Which version of HLS do you use and how did you install it?
1.8.0.0 from ghcup

Have you configured HLS in any way (especially: a `hie.yaml` file)?
Yes, see the project link above for details.

## What's wrong?

In certain Haskell modules, the hovertip may end up with thousands of lines of entries which reference Core variables (e.g., $dEq). I've created a patch that updates the `AtPoint` module in HLS to filter out variables by name when they have prefixes that look like Core variables (e.g., "$d" and "$c").

The issue seems to only happens in modules with both typeclass and type definitions.  In such a module, when the cursor is on a comment above a type definition which includes a deriving clause, HLS ends up thinking there are thousands of lines to include in the hovertip. The rendering of such a large hovertip buffer can cause emacs to freeze for an extended period of time. In practice, I've noticed ~30 seconds on my several years old MBP.

Should HLS be presenting Core variables to the user through hovertips? Would a patch that removes Core variables be considered for fixing this issue? I ask because I have written such a patch in order to prevent HLS from freezing up emacs. However, I believe there may be a deeper issue: why is an `HieAST` for a comment associated with thousands of Core variables?

### Debug information

A smaller reproducible case is available here: https://github.com/drone-rites/debug-hls

Move the cursor on a comment before one of the `data` definitions in the `Main` module and note that there are 126 lines of hover output produced with references to non-unique names of Core variables used in the implementation of type classes (e.g., `$dEq_1234` becomes `$dEq`).

In the `blaze-analysis` project referenced earlier, the number of lines of hovertip documentation is in the thousands. To reproduce, place the cursor over a comment above any of the type definitions in the `Blaze.Types.Pil.Checker` module.

michaelpj:
So at the moment we show hovers for all the variables at a location, which includes typeclass dictionaries. We probably don't want to do that (nobody wants to see the type signatures of the dictionaries!). I think it would be reasonable to filter out everything generated if we can figure out how to identify them.

drone-rites:
[...]

PR #3316, opened 2022-11-02
Filter generated Core variable names from hovertip documentation

drone-rites:
This addresses https://github.com/haskell/haskell-language-server/issues/3280 by checking if names associated with a Haskell source file location follow particular prefix patterns used by generated Core variables. The two prefixes used are: `$d` and `$c`. If a name is prefixed with either of those strings then it is removed from the list of names used to generate the hovertip documentation content.

This change has been tested with the cases mentioned in the associated issue.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3316"></a>

### DIFF
--- a/ghcide/src/Development/IDE/GHC/CoreFile.hs
+++ b/ghcide/src/Development/IDE/GHC/CoreFile.hs
@@ -10,7 +10,8 @@ module Development.IDE.GHC.CoreFile
   , typecheckCoreFile
   , readBinCoreFile
   , writeBinCoreFile
-  , getImplicitBinds) where
+  , getImplicitBinds
+  , occNamePrefixes) where
 
 import           Control.Monad
 import           Control.Monad.IO.Class
@@ -18,6 +19,7 @@ import           Data.Foldable
 import           Data.IORef
 import           Data.List                       (isPrefixOf)
 import           Data.Maybe
+import qualified Data.Text as T
 import           GHC.Fingerprint
 
 import           Development.IDE.GHC.Compat
@@ -228,3 +230,45 @@ tc_iface_bindings (TopIfaceNonRec v e) = do
 tc_iface_bindings (TopIfaceRec vs) = do
   vs' <- traverse (\(v, e) -> (,) <$> pure v <*> tcIfaceExpr e) vs
   pure $ Rec vs'
+
+-- | Prefixes that can occur in a GHC OccName
+occNamePrefixes :: [T.Text]
+occNamePrefixes =
+  [
+    -- long ones
+    "$con2tag_"
+  , "$tag2con_"
+  , "$maxtag_"
+
+  -- four chars
+  , "$sel:"
+  , "$tc'"
+
+  -- three chars
+  , "$dm"
+  , "$co"
+  , "$tc"
+  , "$cp"
+  , "$fx"
+
+  -- two chars
+  , "$W"
+  , "$w"
+  , "$m"
+  , "$b"
+  , "$c"
+  , "$d"
+  , "$i"
+  , "$s"
+  , "$f"
+  , "$r"
+  , "C:"
+  , "N:"
+  , "D:"
+  , "$p"
+  , "$L"
+  , "$f"
+  , "$t"
+  , "$c"
+  , "$m"
+  ]
--- a/ghcide/src/Development/IDE/Spans/AtPoint.hs
+++ b/ghcide/src/Development/IDE/Spans/AtPoint.hs
@@ -227,10 +227,17 @@ atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ p
         wrapHaskell x = "\n```haskell\n"<>x<>"\n```\n"
         info = nodeInfoH kind ast
         names = M.assocs $ nodeIdentifiers info
+        isInternal :: (Identifier, IdentifierDetails a) -> Bool
+        isInternal (Right n, _) =
+          let name = printOutputable n
+              prefix = T.take 2 name
+          in elem prefix ["$d", "$c"]
+        isInternal (Left _, _) = False
+        filteredNames = filter (not . isInternal) names
         types = nodeType info
 
         prettyNames :: [T.Text]
-        prettyNames = map prettyName names
+        prettyNames = map prettyName filteredNames
         prettyName (Right n, dets) = T.unlines $
           wrapHaskell (printOutputable n <> maybe "" (" :: " <>) ((prettyType <$> identType dets) <|> maybeKind))
           : maybeToList (pretty (definedAt n) (prettyPackageName n))
--- a/ghcide/src/Development/IDE/Spans/AtPoint.hs
+++ b/ghcide/src/Development/IDE/Spans/AtPoint.hs
@@ -227,10 +227,14 @@ atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ p
         wrapHaskell x = "\n```haskell\n"<>x<>"\n```\n"
         info = nodeInfoH kind ast
         names = M.assocs $ nodeIdentifiers info
-        -- Check if a name matches a pattern for a generated Core variable.
+        -- Check for evidence bindings
         isInternal :: (Identifier, IdentifierDetails a) -> Bool
         isInternal (Right _, dets) =
+#if MIN_VERSION_ghc(9,0,1)
           any isEvidenceContext $ identInfo dets
+#else
+          False
+#endif
         isInternal (Left _, _) = False
         filteredNames = filter (not . isInternal) names
         types = nodeType info
--- a/ghcide/src/Development/IDE/Spans/AtPoint.hs
+++ b/ghcide/src/Development/IDE/Spans/AtPoint.hs
@@ -34,6 +34,7 @@ import           Development.IDE.Core.PositionMapping
 import           Development.IDE.Core.RuleTypes
 import           Development.IDE.GHC.Compat
 import qualified Development.IDE.GHC.Compat.Util      as Util
+import           Development.IDE.GHC.CoreFile         (occNamePrefixes)
 import           Development.IDE.GHC.Util             (printOutputable)
 import           Development.IDE.Spans.Common
 import           Development.IDE.Types.Options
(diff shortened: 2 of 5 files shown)


ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -225,10 +225,17 @@ atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ p
         wrapHaskell x = "\n```haskell\n"<>x<>"\n```\n"
         info = nodeInfoH kind ast
         names = M.assocs $ nodeIdentifiers info
+        isInternal :: (Identifier, IdentifierDetails a) -> Bool
+        isInternal (Right n, _) =
+          let name = printOutputable n
+              prefix = T.take 2 name
+          in elem prefix ["$d", "$c"]
```
fendor, line 232: Would you mind adding a comment, what we consider to be an internal name?

santiweight:
If it's not too inconvenient: is it possible to have a test for this? Something with Generics is a great candidate, because those are the most problematic in my experience

fendor (changes requested):
Looks like a good start, would it be possible to add the two cases you mention in the associated issues as test-cases? Otherwise, regressions or changes in how the internal names look like will go through unchecked.

ghcide/src/Development/IDE/Spans/AtPoint.hs
drone-rites: Will do. I had done this locally right before pushing, but forgot to commit.

drone-rites: Good thought, there is a list of GHC OccName prefixes in `Development.IDE.Plugin.Completions.Logic`.

drone-rites: Re: tests, is there a good example test or test module I can use as a reference? It looks like there aren't existing tests for the `AtPoint` module. Should `ghcide` tests be placed within the top-level `test/` directory or the `ghcide/test/` directory?

I think we'll want to move the `isInternal` function into the `Development.IDE.Plugin.Completions.Logic` module, and perhaps give it a slightly better name. We could also extract the retrieval of names from the `atPoint` function, make it a top-level function (`getNamesAtPoint`?), and then test that directly rather than test the result of `atPoint`.

I'm new to HLS dev, so I'll proceed with whatever the experienced members recommend.

pepeiborra: Hey @drone-rites sorry for the late response. The ghcide test suite lives all in `ghcide/test/exe/Main.hs` and is almost 100% composed of integration tests, you will find almost no unit tests at all. The best place to test this change is:

https://github.com/haskell/haskell-language-server/blob/6dd5a3b29f857a73f2c526e93003b26928c3d71d/ghcide/test/exe/Main.hs#L996-L1090

drone-rites: Hi @pepeiborra, thanks for the pointer; perfect timing as I was just refactoring the updates today.

drone-rites: Added a single, minimal test that captures the case I was originally observing. Type class dictionary variables associated with a type definition were showing up in the hover info when cursor was placed on a comment above the type definition.

pepeiborra (approved):
This looks good to me, thank you for the test. I only have one question: can someone remind me why aren't we just filtering all identifiers starting with `$`, other than `$` itself?

fendor (approved):
LGTM, amazing and necessary improvement :)

drone-rites:
> This looks good to me, thank you for the test. I only have one question: can someone remind me why aren't we just filtering all identifiers starting with `$`, other than `$` itself?

I started with a predicate like that locally, but then moved to an explicit short list of `$d` and `$c` after recalling there are other valid Haskell operators that match—e.g., `$!`, `$$`, and `$-$`. That short list was expanded once we discovered there was a list of Core variable prefixes already in ghcide.

An alternative predicate that should work is: the identifier has length of at least two characters, starts with "$", and contains an alphanumeric. Maybe there is a specification that describes the naming scheme of generated Core symbols and we could later update this check to follow it.

michaelpj:
It seems that `SrcSpan`s can be `UnhelpfulSpan`s (https://hackage.haskell.org/package/ghc-9.4.2/docs/GHC-Types-SrcLoc.html#t:SrcSpan) which can indicate that they are generated. There's even a function for it: https://hackage.haskell.org/package/ghc-9.4.2/docs/GHC-Types-SrcLoc.html#v:isGeneratedSrcSpan 

Could we replace the name-based logic with some logic that instead uses the `SrcSpan` of the `Name`?

drone-rites:
> It seems that `SrcSpan`s can be `UnhelpfulSpan`s (https://hackage.haskell.org/package/ghc-9.4.2/docs/GHC-Types-SrcLoc.html#t:SrcSpan) which can indicate that they are generated. There's even a function for it: https://hackage.haskell.org/package/ghc-9.4.2/docs/GHC-Types-SrcLoc.html#v:isGeneratedSrcSpan
> 
> Could we replace the name-based logic with some logic that instead uses the `SrcSpan` of the `Name`?

Unfortunately, it doesn't look like access to `UnhelpfulSpan`s are provided through an `HieAST a`. The `nodeSpan` field in an `HieAST a` is a `Span` type which is just an alias for `RealSrcSpan`. That said, the generated variables are showing up when the point is on a non-whitespace character of a comment, which seems like it would be covered by a `RealSrcSpan`. We could dive deeper and figure out why generated variables are being linked with comment spans.

michaelpj:
That does seem odd indeed. A question for @wz1000 I guess.

michaelpj:
Sorry, I thought this was merged! I think this was basically good to go, although I would appreciate it if @wz1000 could take a look and think about the issue with detecting generated code.

wz1000:
Sorry, I didn't look at this earlier, but I'm not sure if looking at occname prefixes is the best way to go about this.

I think this line from #1983 might do what you need to out evidence bindings (`$dFoo` and `$fFoo`:

```haskell
        prettyName (Right n, dets)
#if MIN_VERSION_ghc(9,0,1)
          | any isEvidenceUse (identInfo dets) = maybe "" (printOutputable . renderEvidenceTree) (getEvidenceTree _rf n) <> "\n"
```

https://github.com/haskell/haskell-language-server/pull/1983/files#diff-91f773bae93bbf86035573f4dbffa3035c1689b1bb7e17bde5937bb1271c0131R251

I'm not completely confident that it will remove all the generated names, but you could also have better luck filtering them out if you dropped the generated code like so: https://github.com/haskell/haskell-language-server/blob/2b6f603b6bb88b81ef3392a51d8f774661fe8db5/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs#L234-L244

Maybe these two fixes won't be enough and we would still need to filter out some names based on the prefix (like `$sel` in the presence of Duplicated record fields), so I don't know the best way forward.

michaelpj:
@wz1000 what do you think about the idea of preserving the `UnhelpfulGenerated` `SrcSpan`s through HIE files so we can pick them up again later?

drone-rites:
@wz1000 thanks, I was hoping there'd be a more direct approach. I will try out your suggestions.

drone-rites:
I tested updating `prettyLine` to use `isEvidenceUse` like so:

```
        prettyName (Right n, dets)
#if MIN_VERSION_ghc(9,0,1)
          | any isEvidenceUse (identInfo dets) = "wat...\n"
          | otherwise
#endif
           = T.unlines $
          wrapHaskell (printOutputable n <> maybe "" (" :: " <>) ((prettyType <$> identType dets) <|> maybeKind))
          : maybeToList (pretty (definedAt n) (prettyPackageName n))
          ++ catMaybes [ T.unlines . spanDocToMarkdown <$> lookupNameEnv dm n
                       ]
          where maybeKind = fmap printOutputable $ safeTyThingType =<< lookupNameEnv km n
                pretty Nothing Nothing = Nothing
                pretty (Just define) Nothing = Just $ define <> "\n"
                pretty Nothing (Just pkgName) = Just $ pkgName <> "\n"
                pretty (Just define) (Just pkgName) = Just $ define <> " " <> pkgName <> "\n"
```
It failed to replace the entries for the evidence bindings with the test string `"wat...\n"`.

I also tried using `removeGenerated` which I modified to operate directly on `HieASTs a` and used like so:

```atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ pointCommand (removeGenerated hf) pos hoverInfo```

It also failed to remove the entries.

@wz1000 any other suggestions I should try? Or did I misunderstand your earlier suggestions? Thanks!

**Comment by wz1000** [comment]
Ah, try `isEvidenceContext` which should catch more cases.

**Comment by drone-rites** [comment]
That works, thanks! Sorry for the noise on the PR, somewhere along the lines I merged master into the branch instead of rebasing.

**Review by michaelpj — APPROVED** [review]
Lovely!

**Comment by michaelpj** [comment]
It doesn't seem to build however, it complains that `isEvidenceContext` is not in scope?

File: ghcide/src/Development/IDE/GHC/CoreFile.hs [hunk 2]
```diff
@@ -228,3 +230,45 @@ tc_iface_bindings (TopIfaceNonRec v e) = do
 tc_iface_bindings (TopIfaceRec vs) = do
   vs' <- traverse (\(v, e) -> (,) <$> pure v <*> tcIfaceExpr e) vs
   pure $ Rec vs'
+
+-- | Prefixes that can occur in a GHC OccName
+occNamePrefixes :: [T.Text]
+occNamePrefixes =
```
wz1000, line 236: Is this needed now?

drone-rites: This is a refactor to allow the `prefixes` from the `Logic` module available to other plugins. Thanks to your suggestion, we now don't need to access to those prefixes from the `AtPoint` module now but other modules might need access to the prefixes in the future. We can revert the refactor, or if it's considered an improvement we can keep it. I'll defer to you and others with more experience on the codebase.

drone-rites:
@michaelpj there was a missing preprocessor check. This commit should fix that: https://github.com/haskell/haskell-language-server/pull/3316/commits/30c42bb6d41a0e151690ab9ccc84a0b9b2a247a2

michaelpj:
Thanks!

PR #3317, opened 2022-11-03
Update refactor/splice/hlint/fourmolu/retrie/gadt plugin for GHC 9.4

9999years:
WIP tracking work on GHC 9.4 support for HLS plugins, particularly interested in `hls-refactor-plugin`.

We will need `fourmolu` to cut a new release before CI can start building this PR successfully. https://github.com/fourmolu/fourmolu/pull/242#issuecomment-1314754505


santiweight:
Let me know if you need any help with ls-refactor-plugin!

flake.nix
```diff
@@ -190,6 +193,47 @@
...
+              fourmoluHEAD =
+                appendPatches
+                (doJailbreak (if final.system == "aarch64-darwin"
+                              then overrideCabal hsuper.fourmolu_0_8_2_0 (_: { enableSeparateBinOutput = false; })
+                              else hsuper.fourmolu_0_8_2_0))
+                [
+                  # The GHC 9.4 pull request builds upon these unpublished changes,
+                  # which is why we include them.
+                  #
+                  # Also, there is no PR for these changes.  The fourmolu maintainers
```
brandonchinn178, line 207: this isn't true, for example the first commit is this pr: https://github.com/fourmolu/fourmolu/pull/229

9999years: OK, but this is just for testing. We'll have to release a new Fourmolu before this PR can build outside of Nix.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -192,7 +192,15 @@ typecheckModule (IdeDefer defer) hsc tc_helpers pm = do
               Left errs -> return (map snd diags ++ errs, Nothing)
               Right tcm -> return (map snd diags, Just $ tcm{tmrDeferredError = deferredError})
     where
+#if MIN_VERSION_ghc(9,4,0)
+        -- Unfortunately, on GHC 9.4, with DynFlags corresponding -fdefer-type-errors,
+        -- the error messages are significantly different from the ones without that, and
+        -- thus the code action triggers (such as in hls-refactor-plugin) are not fired properly.
+        -- So we disable the error demotion.
+        demoteIfDefer = id
```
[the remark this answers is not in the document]
wavewave: cc: @santiweight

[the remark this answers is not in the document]
pepeiborra: Disabling deferred type errors will have an impact on many other IDE features, including diagnostics and code navigation. For this reason, I think the only alternatives here are:

* update the regexes to account for the changes in 9.4 (incrementally, in other PRs if necessary)
* migrate the suggestion providers to use typed diagnostics

pepeiborra:
retrie 1.2.1 is now available in Hackage

ghcide/src/Development/IDE/Core/Compile.hs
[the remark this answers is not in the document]
wavewave: Thanks for the feedback! 
I will revert this disabling and mark the failed tests due to this with comments. The fix for the failed tests will not be small as previously the necessary error source location info  was obtained from the error message which is gone with the deferred type error flag, so let us address that in a separate PR. After all, typed diagnostics is the best direction, so we had better move on in that direction.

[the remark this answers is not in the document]
wavewave: reverted the change and mark the affected test as broken.

pepeiborra:
Please let me know if there are any further CI problems blocking this effort

wavewave:
@pepeiborra 
Thank you! I am now fixing the upstream apply-refact problem first for correct GHC 9.2 and 9.0 support for the release.
I will let you know if any CI issue happens in this build.

wavewave:
@pepeiborra : what is this CI issue?
https://github.com/haskell/haskell-language-server/actions/runs/3698751170/jobs/6265340276#step:11:26

wavewave:
oh. that was an actual test error (thought it was different CI-specific issue as 9.2.4 was marked successful). never mind. will fix that.

wavewave:
The currently failing wrapper test should be removed from haskell-language-server.
```
   stack with existing cabal build artifact: FAIL (0.65s)
      test/wrapper/Main.hs:43:
      Wrapper reported wrong project type: Cradle {cradleRootDir = "/tmp/hls-test-508e1b444a12818e", cradleOptsProg = CradleAction: Stack}
      Use -p '/stack with existing cabal build artifact/' to rerun this test only.
```
This is actually depends on which version of implicit-hie.
with implicit-hie-0.1.3.0 and implicit-hie-cradle-0.5.0.0, it finds Stack
```
$ ghci
ghci> import Hie.Implicit.Cradle 
ghci> loadImplicitHieCradle "/Users/ianwookim/repo/srcc/haskell-language-server/tmp/stack-with-dist-newstyle/"
Cradle {cradleRootDir = "/Users/ianwookim/repo/srcc/haskell-language-server/tmp/stack-with-dist-newstyle", cradleOptsProg = CradleAction: Stack}
```

with implicit-hie-0.1.2.7 and implicit-hie-cradle-0.5.0.0, it finds Cabal
```
$ ghci -p
ghci> import Hie.Implicit.Cradle 
ghci> loadImplicitHieCradle "/Users/ianwookim/repo/srcc/haskell-language-server/tmp/stack-with-dist-newstyle/"
Cradle {cradleRootDir = "/Users/ianwookim/repo/srcc/haskell-language-server/tmp/stack-with-dist-newstyle", cradleOptsProg = CradleAction: Cabal}
```
This is out of the scope of haskell-language-server and if relevant, the test should exist on implicit-hie side. It was very difficult for me to figure out what was going on, and will make future developers suffer again.

pepeiborra:
Thanks for the investigation.

I agree that the test is out of scope or borderline, but it does have value: it highlighted a change in behaviour that will probably impact some of our users.

@fendor what's the right thing to do here?

fendor:
> @fendor what's the right thing to do here?

In my opinion, not upgrading `implicit-hie`. This sounds like a regression in implicit-hie, and looking at issues such as https://github.com/Avi-D-coder/implicit-hie/issues/50, I don't think we should upgrade the `implicit-hie` version. 

Unfortunately, we don't directly depend on `implicit-hie` but on `implicit-hie-cradle`, which accepts either version.

wavewave:
@pepeiborra @fendor Thank you for the suggestions. I will downgrade it to the old version. and please don't feel sorry! I just wanted to call out such a buried difficulty as new contributors (and experienced contributors as well) may encounter if a problem is not localized in this mega-repo.
This issue is simply improved by moving the test into `implicit-hie`, as the test actually does not involve anything specific to haskell-language-server.

wavewave:
it seems like `cabal test func-test` is flaky. each time i run, i have different results (though certainly some tests are consistently broken)

cabal.project
```diff
@@ -65,7 +65,10 @@ constraints:
   -- https://github.com/ndmitchell/hlint/issues/1376
   hlint +ghc-lib,
   ghc-lib-parser-ex -auto,
-  stylish-haskell +ghc-lib
+  stylish-haskell +ghc-lib,
+  -- implicit-hie 0.1.3.0 introduced an unexpected behavioral change.
+  -- https://github.com/Avi-D-coder/implicit-hie/issues/50
+  implicit-hie < 0.1.3
```
fendor: I think this needs to go into `ghcide.cabal`, otherwise the Hackage release will still permit this version

wavewave: Unfortunately, we cannot specify it there. The `ghcide` library depends on `implicit-hie-cradle` only, not `implicit-hie` directly. Though we have the target `ghcide:ghcide-tests` that depends on `implicit-hie:gen-hie`, putting a constraint there does not forbid installing `implicit-hie 0.1.3.0` in my experiment, so I couldn't find a better way than this yet.

fendor: doesn't adding the depedency `implicit-hie` to ghcide.cabal work, even if the dependency is unused?

michaelpj: It probably does. But also we need to try and resolve this upstream. @Avi-D-coder do you know what's up? Should we be deprecating the broken versions so it doesn't get picked by cabal ever?

wavewave: ah. yeah. I refrain myself from adding an unused dep. saw the comment in #3404 now. will wait for merging that fix onto master first as discussed.

Avi-D-coder: I have just set `[0.1.3.0](https://hackage.haskell.org/package/implicit-hie-0.1.3.0)` as deprecated.
I have not had the time to finish implementing https://github.com/Avi-D-coder/implicit-hie/issues/50#issuecomment-1316100308.

pepeiborra:
Is this ready for review? CI is green

cabal.project
```diff
@@ -81,6 +82,13 @@ source-repository-package
   -- https://github.com/tibbe/ekg-json/pull/12
 -- END DELETE
 
+if (impl(ghc >= 9.2))
+  source-repository-package
+    type:git
+    location: https://github.com/wavewave/apply-refact
+    tag: 2f346adb54ae3d911c9055b48fb62eb4ceab583a
+    -- https://github.com/wavewave/apply-refact/pull/1
```
pepeiborra, line 90: Is there a plan for upstreaming this to https://github.com/mpickering/apply-refact ?

wavewave: yup. I contacted the maintainers and kept discussing about this directly last few days. Will remove this as the new version is released. thanks.

wavewave:
I think that other than `apply-refact` replaced by 0.11, and then the PR is reviewable.
@9999years : would you change the title more specifically like "refactor/splice/hline/fourmolu plugin on GHC 9.4"  and make it reviewable? Thanks!

wavewave:
apply-refact is updated to the hackage version :-)

konn (approved):
LGTM as for Splice Plugin. Thanks! 👍

9999years:
Sorry I don't know how the GitHub reviewers mechanic really works I guess

fendor (approved):
Awesome, thank you very much! 
I skimmed the changes for sanity but not correctness, so take this approval with a grain of salt :see_no_evil:

konn (approved):
LGTM again, as the CI has passed. Thanks!

wz1000:
I've added 9.4 support for the retrie and gadt plugins to this PR.

wz1000 (approved):

PR #3323, opened 2022-11-07
Refactor overlay composition

Gabriella439:
… so that it's easier to add and remove overlays by creating a list of overlays to compose.

This should be a behavior-preserving change.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3323"></a>


guibou (approved):
I tested by running a few shells and build and it works for me without any noticable behavior change.

Thank you.

PR #3338, opened 2022-11-15
Draft: revive Wingman

santiweight:
It's looking pretty good. Still working on it. I expect to have a final version working this weekend, depending on whether I feel like working on this or the extract method code action...

Tests are at 14/77. ~10 of these failures are non-serious (such as fresh variables being slightly different). 4 of these failures are real, and I haven't investigated. I suspect I will be ignoring them for the sake of this MR

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3338"></a>


cabal.project
```diff
@@ -9,7 +9,7 @@ packages:
          ./hls-plugin-api
          ./hls-test-utils
          ./plugins/hls-cabal-fmt-plugin
-         ./plugins/hls-tactics-plugin
+         ./plugins/hls-refine-destruct-plugin
```
michaelpj, line 12: I suggest we don't do this - we do publish these to hackage, so renaming them is just a bit unnecessarily annoying.

michaelpj (commented):
Thanks for taking this on! Was there anyone else keen on maintaining wingman who we could pull in for review?

ChrisPenner:
Very excited this is being added, thanks for taking it on.

Mostly I use the pattern match expansion action 🙌🏼

michaelpj:
I'm closing this, but we should remember that it exists for reference if someone wants to come back to this.

michaelpj:
@bzm3r do you mean you want to pick up the work? I think there's quite a lot to do, not just maintenance :sweat_smile: You might be able to discuss with @santiweight what needs doing

artimath:
@bzm3r are you still working on this?

I am rather keen to dive in face-first, and rather sure that I have very little understanding of the scope of work entailed.

PR #3343, opened 2022-11-18
Add `RangeMap` for unified "in-range" filtering

ozkutuk:
This PR makes `RangeMap` a part of the plugin API, first introduced as part of `hls-explicit-record-fields-plugin`. `RangeMap` is a data structure that allows for fast queries of range intersections. Details on usage can be found in the provided documentation, and the performance characteristics compared to the traditional list-based approach can be found in the newly-added benchmark.

Browsing through existing plugins, I couldn't find as much opportunities to use this as I hoped to (without significantly modifying plugin source), since to make the best use of this, a plugin needs to:

1. do range filtering
2. utilize custom Shake rules (for the reason specified in the documentation)

However, I still think this lays down a nice foundation for future plugins, if they were to need such a functionality.

---

Screenshot from the output of: `cabal bench hls-plugin-api -O2 --benchmark-options="-o benchmark.html"`

benchmark

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3343"></a>


hls-plugin-api/hls-plugin-api.cabal
```diff
@@ -57,6 +58,7 @@ library
     , transformers
     , unordered-containers
     , megaparsec > 9
+    , hw-fingertree
```
pepeiborra, line 61: I'm not keen on adding new dependencies to `his-plugin-api`. Other alternatives worth considering:

- Extract `RangeMap` to a standalone package 
- Leave it in its current home - plugins can depend on each other just fine

ozkutuk: Although I think the functionality is pretty small to warrant a standalone package, I think I can do that. I am not too keen on the second option. Even if the plugins _can_ depend on each other, I think having this as part of a specific plugin signals the idea that this is something specific to that plugin. I think that may discourage plugin authors to use it (it certainly would discourage me).

What is the downside of adding new dependencies to `hls-plugin-api`?

michaelpj: > What is the downside of adding new dependencies to hls-plugin-api?

When we update HLS for a new version of GHC, `hls-plugin-api` is one of the packages that needs to get updated if we are to do _anything_. We are frequently blocked by upstream dependencies not working with the new version of GHC; every new dependency that we add to key packages is another opportunity for that to happen...

michaelpj: That said, `hw-fingertree` depends only on `base`, which is promising.

ozkutuk: I see. Indeed, I had to add GHC 9.4 support to `hw-fingertree` myself, so I can definitely see that happening :sweat_smile: Then I will extract this into a separate small package.

hls-plugin-api/src/Ide/Plugin/RangeMap.hs
```diff
@@ -0,0 +1,44 @@
...
+  where
+    toIntervalMap :: Ord v => [(IM.Interval v, a)] -> IM.IntervalMap v a
+    toIntervalMap = foldl' (\m (i, v) -> IM.insert i v m) IM.empty
+
+-- | Filter a 'RangeMap' by a given 'Range'.
+filterByRange :: Range -> RangeMap a -> [a]
+filterByRange range = map snd . IM.dominators (rangeToInterval range) . unRangeMap
+
+rangeToInterval :: Range -> IM.Interval Position
+rangeToInterval (Range s e) = IM.Interval s e
```
michaelpj, line 44: This is, strictly, wrong! LSP `Range`s have _exclusive_ upper bounds, whereas the intervals here are supposed to be closed (i.e. inclusive at both ends). I think converting a `Range` to a closed interval may be difficult if you need to drop down a line (i.e. if the end of the `Range` is `(x,0)`), because you'd need to know where the end of the previous line is. Not sure what to do there or whether this is a problem but it definitely needs a comment.

hls-plugin-api/src/Ide/Plugin/RangeMap.hs
```diff
@@ -0,0 +1,44 @@
...
+import qualified HaskellWorks.Data.IntervalMap.FingerTree as IM
+import           Language.LSP.Types                       (Position,
+                                                           Range (Range))
+
+-- | A map from code ranges to values.
+newtype RangeMap a = RangeMap
+  { unRangeMap :: IM.IntervalMap Position a
+    -- ^ 'IM.Interval' of 'Position' corresponds to a 'Range'
+  }
+  deriving newtype (NFData)
```
michaelpj, line 27: might as well add `Show, Eq, Ord`?

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -52,7 +54,7 @@ instance NFData CollectLiterals
 type instance RuleResult CollectLiterals = CollectLiteralsResult
 
 data CollectLiteralsResult = CLR
-    { literals          :: [Literal]
+    { literals          :: RangeMap Literal
```
michaelpj, line 57: nice

plugins/hls-alternate-number-format-plugin/test/Main.hs
```diff
@@ -29,7 +29,7 @@ test :: TestTree
 test = testGroup "alternateNumberFormat" [
     codeActionHex "TIntDtoH" 3 13
     , codeActionOctal "TIntDtoO" 3 13
-    , codeActionBinary "TIntDtoB" 4 12
+    , codeActionBinary "TIntDtoB" 4 13
```
michaelpj, line 32: what happened here?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -154,7 +142,8 @@ collectRecordsRule recorder = define (cmapWithPrio LogShake recorder) $ \Collect
       recs = concat $ maybeToList (getRecords <$> tmr)
   logWith recorder Debug (LogCollectedRecords recs)
   let renderedRecs = traverse renderRecordInfo recs
-      recMap = buildIntervalMap <$> renderedRecs
+      -- All spans are supposed to be `RealSrcSpan`, hence the use of `fromJust`.
```
michaelpj, line 145: there is no `fromJust`? Which we should never use anyway.

michaelpj (approved):
I like it!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: Ah, right. There used to be a `fromJust` but I was dissatisfied with that so removed it along the way. Seems like I forgot to remove the comment...

hls-plugin-api/src/Ide/Plugin/RangeMap.hs
ozkutuk: I only added what was currently necessary at the use-site, but adding those as well seems like a good idea :+1:

plugins/hls-alternate-number-format-plugin/test/Main.hs
drsooch: Shouldn't cause any worries, it's just the column number to run the code action. I think the column number is arbitrarily in the middle of a number. Looks like it just matches the other test cases.

hls-plugin-api/hls-plugin-api.cabal
ozkutuk: > Personally I suggest extracting it to a _module_ in `hls-plugin-api` or something.

I am a bit confused by this. Do you suggest something other than what currently is present in the PR? Because it currently _is_ a module within `hls-plugin-api` (unless there is a separate concept of _module_ which I'm not aware).

> what if we add a cabal flag to `hls-plugin-api` `use-fingertree`

Sounds good to me. What do you think @pepeiborra?

hls-plugin-api/src/Ide/Plugin/RangeMap.hs
ozkutuk: Well, `IntervalMap` apparently [doesn't have](https://hackage.haskell.org/package/hw-fingertree-0.1.2.1/docs/HaskellWorks-Data-IntervalMap-FingerTree.html#t:IntervalMap) those instances. Perhaps it would make sense to provide the `Functor` instance though?

michaelpj: Huh. I think maybe I just scrolled to `Interval` by accident :sweat_smile: `Traversable` and `Foldable` and `Monoid` are probably all useful too.

hls-plugin-api/hls-plugin-api.cabal
michaelpj: Ah, I think I misunderstood. I thought we were talking about extracting the _fingertree_ code from `hw-fingertree` so we controlled it, rather than extracting `RangeMap`. Ignore me, then.

> what if we add a cabal flag to hls-plugin-api use-fingertree

... but I still think this is the cleanest solution.

ozkutuk: I have put the fingertree implementation behind a flag. All related tests (tests of `RangeMap` itself and the plugins that make use of it) pass with both implementations. Can you take another look to see if it reflects what you had in mind? @michaelpj

hls-plugin-api/src/Ide/Plugin/RangeMap.hs
ozkutuk: I couldn't say something about it right away because the test data I generated at the time didn't have any multiline ranges. However I extended the generators now, and the tests still pass (that is, `RangeMap` behaves identical to the list-based approach with multiline ranges as well). I think the reason for this is, even if `rangeToInterval` isn't a correct 1:1 conversion by itself, it is used for both the construction of the `RangeMap` and during the actual filtering (`filterByRange`), so it still behaves identical to the list approach. You can say it is a "two wrongs make a right" situation :smile:

michaelpj: I think regardless this needs a comment! Future people will thank us! Even if it's just "this is wrong but it empirically works and I'm not sure why"

hls-plugin-api/hls-plugin-api.cabal
```diff
@@ -20,6 +20,11 @@ flag pedantic
   default:     False
   manual:      True
 
+flag use-fingertree
```
michaelpj, line 23:
```suggestion
-- This flag can be used to avoid the dependency on hw-fingertree.
-- We can set this temporarily if we have problems building hw-fingertree
-- for a new version of GHC.
flag use-fingertree
```

michaelpj (approved):
I think this just needs a couple of comments and it's good to go!

PR #3347, opened 2022-11-19
Introduce common code for Recorders in Plugin Tests

fendor:
Plugin Tests need a recorder. Most plugin test-suite just set it to `mempty`, which, I think, will make it more difficult to debug test-failures.

Some plugin test-suites roll their own plugin initialisation code. This is fine, but makes the logging behaviour non-uniform.

Introduce the `PluginTestDescriptor` type definition.
Utility functions to create `PluginTestDescriptor`'s `mkPluginTestDescriptor`  and `mkPluginTestDescriptor'` which will help migrating plugins to the newer co-log style logger. 

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3347"></a>

### DIFF
--- a/hls-test-utils/src/Test/Hls.hs
+++ b/hls-test-utils/src/Test/Hls.hs
@@ -20,16 +20,18 @@ module Test.Hls
     goldenWithHaskellDocFormatter,
     goldenWithCabalDocFormatter,
     def,
-    pluginTestRecorder,
+    -- * Running HLS for integration tests
     runSessionWithServer,
-    runSessionWithServerAndRecorder,
+    runSessionWithServerAndCaps,
     runSessionWithServerFormatter,
     runSessionWithCabalServerFormatter,
     runSessionWithServer',
-    waitForProgressDone,
-    waitForAllProgressDone,
+    -- * Helpful re-exports
     PluginDescriptor,
     IdeState,
+    -- * Assertion helper functions
+    waitForProgressDone,
+    waitForAllProgressDone,
     waitForBuildQueue,
     waitForTypecheck,
     waitForAction,
@@ -37,6 +39,11 @@ module Test.Hls
     getLastBuildKeys,
     waitForKickDone,
     waitForKickStart,
+    -- * Plugin descriptor helper functions for tests
+    PluginTestDescriptor,
+    pluginTestRecorder,
+    mkPluginTestDescriptor,
+    mkPluginTestDescriptor',
     -- * Re-export logger types
     -- Avoids slightly annoying ghcide imports when they are unnecessary.
     WithPriority(..),
@@ -50,6 +57,7 @@ import           Control.Concurrent.Async        (async, cancel, wait)
 import           Control.Concurrent.Extra
 import           Control.Exception.Base
 import           Control.Monad                   (guard, unless, void)
+import           Control.Monad.Extra             (forM)
 import           Control.Monad.IO.Class
 import           Data.Aeson                      (Result (Success),
                                                   Value (Null), fromJSON,
@@ -69,7 +77,7 @@ import qualified Development.IDE.Main            as IDEMain
 import           Development.IDE.Plugin.Test     (TestRequest (GetBuildKeysBuilt, WaitForIdeRule, WaitForShakeQueue),
                                                   WaitForIdeRuleResult (ideResultSuccess))
 import qualified Development.IDE.Plugin.Test     as Test
-import           Development.IDE.Types.Logger    (Logger (Logger),
+import           Development.IDE.Types.Logger    (Doc, Logger (Logger),
                                                   Pretty (pretty),
                                                   Priority (Debug),
                                                   Recorder (Recorder, logger_),
@@ -106,8 +114,6 @@ import           Test.Tasty.Golden
 import           Test.Tasty.HUnit
 import           Test.Tasty.Ingredients.Rerun
 import           Test.Tasty.Runners              (NumThreads (..))
-import Control.Monad.Extra (forM)
-import Development.IDE.Types.Logger (Doc)
 
 newtype Log = LogIDEMain IDEMain.Log
 
@@ -126,7 +132,8 @@ goldenGitDiff :: TestName -> FilePath -> IO ByteString -> TestTree
 goldenGitDiff name = goldenVsStringDiff name gitDiff
 
 goldenWithHaskellDoc
-  :: PluginDescriptor IdeState
+  :: Pretty b
+  => PluginTestDescriptor b
   -> TestName
   -> FilePath
   -> FilePath
@@ -137,7 +144,8 @@ goldenWithHaskellDoc
 goldenWithHaskellDoc = goldenWithDoc "haskell"
 
 goldenWithCabalDoc
-  :: PluginDescriptor IdeState
+  :: Pretty b
+  => PluginTestDescriptor b
   -> TestName
   -> FilePath
   -> FilePath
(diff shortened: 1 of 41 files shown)


hls-test-utils/src/Test/Hls.hs
```diff
@@ -134,10 +144,34 @@ goldenWithHaskellDoc plugin title testDataDir path desc ext act =
...
+  -> TestTree
+goldenWithHaskellDocAndRecorder plugin title testDataDir path desc ext act =
+  goldenGitDiff title (testDataDir </> path <.> desc <.> ext)
+  $ runSessionWithServerAndRecorder plugin testDataDir
+  $ TL.encodeUtf8 . TL.fromStrict
+  <$> do
+    doc <- openDoc (path <.> ext) "haskell"
+    void waitForBuildQueue
+    act doc
+    documentContents doc
```
pepeiborra, line 165: This completely subsumes `goldenWithHaskellDoc`. Do we need a new function? I would vote for refactoring `goldenWithHaskellDoc` to provide a recorder.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -134,10 +144,34 @@ goldenWithHaskellDoc plugin title testDataDir path desc ext act =
...
 
 runSessionWithServer :: PluginDescriptor IdeState -> FilePath -> Session a -> IO a
 runSessionWithServer plugin = runSessionWithServer' [plugin] def def fullCaps
 
+runSessionWithServerAndRecorder :: Pretty b => (Recorder (WithPriority b) -> PluginDescriptor IdeState) -> FilePath -> Session a -> IO a
+runSessionWithServerAndRecorder pluginF fp act = do
+  recorder <- pluginTestRecorder
+  runSessionWithServer' [pluginF recorder] def def fullCaps fp act
```
pepeiborra, line 173: Same about this: I would refactor `runSessionWithServer` to provide a recorder

hls-test-utils/src/Test/Hls.hs
```diff
@@ -211,6 +245,46 @@ keepCurrentDirectory = bracket getCurrentDirectory setCurrentDirectory . const
...
+initialiseTestRecorder :: Pretty a => [String] -> IO (Recorder (WithPriority a), WithPriority (Doc ann) -> IO ())
+initialiseTestRecorder envVars = do
+    docWithPriorityRecorder <- makeDefaultStderrRecorder Nothing Debug
+    -- There are potentially multiple environment variables that enable this logger
+    definedEnvVars <- forM envVars (\var -> fromMaybe "0" <$> lookupEnv var)
+    let logStdErr = any (/= "0") definedEnvVars
+
+        docWithFilteredPriorityRecorder =
+          if logStdErr then mempty
+          else cfilter (\WithPriority{ priority } -> priority >= Debug) docWithPriorityRecorder
```
pepeiborra, line 282: maybe I'm confused but I would say that the then-else branches should be the other way around

hls-test-utils/src/Test/Hls.hs
```diff
@@ -211,6 +245,46 @@ keepCurrentDirectory = bracket getCurrentDirectory setCurrentDirectory . const
...
+-- to write all logs to stderr.
+pluginTestRecorder :: Pretty a => IO (Recorder (WithPriority a))
+pluginTestRecorder = do
+  (recorder, _) <- initialiseTestRecorder ["HLS_TEST_PLUGIN_LOG_STDERR"]
+  pure recorder
+
+-- | Generic recorder initialisation for plugins and the HLS server for test-cases.
+--
+-- The created recorder writes to stderr if any of the given environment variables
+-- have been set to a value different to @0@.
```
pepeiborra, line 267: Why not write to stderr unconditionally, as the default recorder for HLS does?

hls-test-utils/src/Test/Hls.hs
fendor: Yeah you are right, messed up the condition inversion

hls-test-utils/src/Test/Hls.hs
fendor: Lots of noise, makes it hard to tell what testcase failed, imo. However, thats also fine by me

pepeiborra: But this is already the status quo, and `lsp-test` silences it anyway, so I'm not sure what the difference is

hls-test-utils/src/Test/Hls.hs
fendor: Exactly that was the question, then all plugins that don't use the new recorder infrastructure will have to be updated, too. I don't see a problem with that if you give a thumbs up to that idea.

hls-test-utils/src/Test/Hls.hs
fendor: No it is not, the status quo is, we hide the stderr unless `LSP_TEST_LOG_STDERR=1` is given, take a look at this example run with the loggers for HLS and the testsuite writing to stderr vs not writing to stderr: https://paste.tomsmeding.com/0UbJEowE

michaelpj:
> This requires a new sway of test-functions, as long as we have plugins that use the old hslogger functions.

Argh, I forgot we hadn't purged all of that yet >:(

hls-test-utils/src/Test/Hls.hs
michaelpj, line 282: Might we want an env var for the priority?

michaelpj (approved):
LGTM!

michaelpj:
I think "always have a recorder" is a good default tbh, so :+1: for refactoring things to make that the main way.

hls-test-utils/src/Test/Hls.hs
fendor: Yeah, but ideally in a follow up PR, this PR just exploded in size :)

hls-test-utils/src/Test/Hls.hs
fendor: Modified as requested

hls-test-utils/src/Test/Hls.hs
fendor: Modified as requested

fendor:
Ready for review, each commit modifies a single plugin so code owners can review exactly the commit they need to.

konn (approved):
Thank you for great refactoring! LGTM 👍

ozkutuk (approved):
LGTM!

plugins/hls-haddock-comments-plugin/src/Ide/Plugin/HaddockComments.hs
```diff
@@ -7,7 +7,7 @@
 {-# LANGUAGE RecordWildCards           #-}
 {-# LANGUAGE ViewPatterns              #-}
 
-module Ide.Plugin.HaddockComments (descriptor) where
+module Ide.Plugin.HaddockComments (descriptor, E.Log) where
```
kokobd, line 10: Is this new import item used somewhere?

fendor: Yes, here: https://github.com/haskell/haskell-language-server/pull/3347/files#diff-52fd7f78c5c4a67b8059732e6f4d8216411b28c7d890cfccc01180c1332740a2R21

VeryMilkyJoe (approved):

kokobd (approved):
LGTM!

georgefst (approved):

pepeiborra (approved):

PR #3351, opened 2022-11-21
Update cabal.project to enable build with GHC-9.4.3

mouse07410:
Change `index-state` to today's date to make it possible to pull `ghc-exactprint-1.6.1` instead of 1.6.0, which breaks build with GHC-9.4.3.

In addition to above, may need to pass to Cabal something like `--constraint ghc-exactprint ==1.6.1`

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -50,7 +50,7 @@ package *
 
 write-ghc-environment-files: never
 
-index-state: 2022-10-07T12:19:15Z
+index-state: 2022-11-20T20:19:15Z
 
 constraints:
   -- For GHC 9.4, older versions of entropy fail to build on Windows


fendor:
Hi, thank you for the contribution, please add GHC 9.4.3 to CI so we don't accidentally break it again.For example, here: https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/test.yml#L60

michaelpj:
Probably drop `9.4.1` also.

mouse07410:
Any idea what it means if a test was "Canceled"? Who or what "canceled" it, and why?

michaelpj:
You may need to look at `test.yml` some more: various tests are disabled on various GHC versions, and some of them may need to be disabled for the new versions. You're getting a genuine failure in the brittany tests.

mouse07410:
Ok, what do we do now?

haskell-language-server.cabal
```diff
@@ -497,6 +497,9 @@ executable haskell-language-server
 
   default-language: Haskell2010
   default-extensions: DataKinds, TypeOperators
+  
+  if impl(ghc >= 9.4)
+    build-depends: ghc-exactprint >= 1.6.1
```
michaelpj, line 502: This doesn't seem correct. Surely the dependency is from another package, and that's what should be updated?

mouse07410: OK, removed. 

Somewhere, something was pulling `ghc-exactprint-1.6.0`. I checked `ghcide`, and it was not it. I checked `hls-refactor-plugin`, and it had ` , ghc-exactprint < 1 || >= 1.4` - which should pull 1.6.1, but in some cases in my experience didn't (no idea why).

mouse07410: Results show that the "wrong" works, and the "right" fails the tests. I'm leaving PR-that-works here.

mouse07410:
Guys, whoever's responsible for CircleCI integration - it seems to fail to authenticate. Please feel free to address this problem.

mouse07410:
From CircleCI:
```
Could not find a usable config.yml, you may have revoked the CircleCI OAuth app.
Please sign out of CircleCI and log back in with your VCS before triggering a new pipeline.
```

Since I don't have an account with CircleCI, I cannot "log back in".

Issue #2971, opened 2022-06-18
Support Wingman on ghc 9.2


santiweight:
Wingman does not currently compile with 9.2. I've investigated a little and I think I've figured out one of the main barriers for the migration.

In <9.x, the staticPlugins field was in DynFlags. Something like:

```haskell
data DynFlags = {
  ...,
  staticPlugins :: [StaticPlugin]
  }
```

In 9.2+, plugins are now contained in the Plugins datatype:

```haskell
data Plugins = Plugins
  { staticPlugins :: ![StaticPlugin]
  , loadedPlugins :: ![LoadedPlugin]
  , loadedPluginDeps :: !([Linkable], PkgsLoaded)
  }
```

which is now found in HscEnv:

```haskell
data HscEnv
  = HscEnv {
        hsc_dflags :: DynFlags,
        , hsc_plugins :: !Plugins
  }
```

Afaict, this means that the PluginDescriptor type should really have a HscEnvModifications field, or alternatively _also_ include a GhcPluginsModifications field.

I figure this is quite a hefty and widespread change, since you would have to update all tutorials and current HLS plugins; so maybe someone knows of a better way to modify the HscEnv?

Paging @isovector and @anka-213 since they might have good feedback

santiweight:
Linking to #2179

michaelpj:
Well, I think `pluginModifyDynFlags` was added exclusively for wingman and I don't see any other users, so I think you can do what you want with it. That said, it seems subtle, looking at the original discussion: https://github.com/haskell/haskell-language-server/pull/1814

ProofOfKeags:
Please <3. How pervasive a change would it be? Is it doable by a first time contributor?

santiweight:
I investigated some more yesterday (I would be a first time contributor; though I've poked around the internals before).

The biggest barrier has ended up being the GHC exactprint API changes. I was able to modify the plugins logic to compile, but then ran into a bunch of errors relating to the new exactprint AST in GHC. This is definitely surmountable, but it is hard for me to tell immediately what parts of the AST isovector is using, and for what. I was working through whack-a-moling the exactprint errors when I stopped. I'll be coming back to it soon...

For example, I can't tell which pass Wingman lives in. Much of the logic is polymorphic in its GHC pass but that no longer works (I think) because you can only annotate certain passes of the GHC with certain annotations. So such polymorphism might be less possible iiuc.

@ProofOfKeags if you want to work on this together let me know :)

[...]

PR #3352, opened 2022-11-21
wingman: move wingman to new directory

santiweight:
This implements step 1 of the plan outlined in https://github.com/haskell/haskell-language-server/issues/2971#issuecomment-1321296785


santiweight:
@michaelpj You had some comments about conditional compilation in https://github.com/haskell/haskell-language-server/issues/2971#issuecomment-1164170146 so I'm paging you :)

michaelpj (approved):

isovector (approved):
LGTM. Thanks for pushing this through, Santi!

santiweight:
How come I can only restart one job at a time? I've had each Windows job be cancelled, which means I have to restart ~4/5 jobs each one-by-one with an hour between each job 😆

michaelpj:
There should be a way to restart all failed jobs, I don't know why you can't do it?

santiweight:
I had an option to restart all tests, but not all jobs. Perhaps I just didn't see the button :/

PR #3360, opened 2022-11-22
Cleanup Development.IDE.CodeAction

santiweight:
Extract a couple of modules from the massive CodeAction file...

Intended for commit-by-commit review

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3360"></a>


michaelpj (approved):
LGTM. You're getting some hlint failures for forbidden functions since `CodeAction.hs` is whitelisted for historical reasons. If you feel like fixing those that would be great, otherwise you can add the new files to the whitelist in `hlint.yaml`.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -20,18 +20,24 @@ suggestFillHole Diagnostic{_range=_range,..}
...
       addBackticks text = "`" <> text <> "`"
       addParens text = "(" <> text <> ")"
       proposeHoleFit holeName parenthise isInfixHole name =
-        let isInfixOperator = T.head name == '('
-            name' = getOperatorNotation isInfixHole isInfixOperator name in
-          ( "replace " <> holeName <> " with " <> name
-          , TextEdit _range (if parenthise then addParens name' else name')
-          )
+        case T.uncons name of
+          Nothing -> error "impossible: empty name provided by ghc"
```
michaelpj, line 28: To be clear, these are an improvement in that they give clearer errors, but they will still _crash HLS unrecoverably_. We really, really want to avoid `error` and friends also. But fixing them properly would require more refactoring.

PR #3363, opened 2022-11-23
Wingman copy old to new

santiweight:
This implements step 2 of the plan outlined in https://github.com/haskell/haskell-language-server/issues/2971#issuecomment-1321296785

I've chosen to implement this in multiple MRs because it just gets the ball rolling. I have other MRs I am working on in parallel, and so implementing small chunks that I can get in parallel works better for me rn...

------------

This MR does something really boring: it copy and pastes the code in the `old` directory into the `new` directory. Note that the new code doesn't actually build with 9.2.1+. I am doing this MR (instead of merging later) because it has to be done either way and I don't want to battle all of CI in one big change.

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3363"></a>

Sanity check (which returns an empty diff): `diff -r old new`


isovector (approved):

michaelpj:
Not sure what's up with the benchmarking job :thinking:

pepeiborra:
> Not sure what's up with the benchmarking job 🤔

Maybe `hs-boot` files need to be added to `extra-src-files`

PR #3364, opened 2022-11-24
Fix build and tests with GHC-9.4.3.

mouse07410:
Fixes build with GHC-9.4.3. 
Includes workaround for apparently-broken `implicit-hie` 0.1.3.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,8 +57,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        ghc: [ "9.4.2"
-             , "9.4.1"
+        ghc: [ "9.4.3"
+             , "9.4.2"
              , "9.2.4"
              , "9.2.3"
              , "9.0.2"
@@ -69,6 +69,9 @@ jobs:
             ]
         include:
            # only test supported ghc major versions
+           - os: ubuntu-latest
+             ghc: '9.4.3'
+             test: true
            - os: ubuntu-latest
              ghc: '9.4.2'
              test: true
@@ -150,15 +153,15 @@ jobs:
           HLS_WRAPPER_TEST_EXE: hls-wrapper
         run: cabal test wrapper-test --test-options="$TEST_OPTS --rerun-log-file .tasty-rerun-log-wrapper"
 
-      - if: matrix.test && matrix.ghc != '9.2.4' && matrix.ghc != '9.4.2'
+      - if: matrix.test && matrix.ghc != '9.2.4' && matrix.ghc != '9.4.2' && matrix.ghc != '9.4.3'
         name: Test hls-brittany-plugin
         run: cabal test hls-brittany-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-brittany-plugin --test-options="$TEST_OPTS"
 
-      - if: matrix.test && matrix.ghc != '9.4.2'
+      - if: matrix.test && matrix.ghc != '9.4.2' && matrix.ghc != '9.4.3'
         name: Test hls-refactor-plugin
         run: cabal test hls-refactor-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-refactor-plugin --test-options="$TEST_OPTS"
 
-      - if: matrix.test && matrix.ghc != '9.4.2'
+      - if: matrix.test && matrix.ghc != '9.4.2' && matrix.ghc != '9.4.3'
         name: Test hls-floskell-plugin
         run: cabal test hls-floskell-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-floskell-plugin --test-options="$TEST_OPTS"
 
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -69,6 +69,9 @@ jobs:
             ]
         include:
            # only test supported ghc major versions
+           - os: ubuntu-latest
+             ghc: '9.4.3'
+             test: true
            - os: ubuntu-latest
              ghc: '9.4.2'
              test: true
```
pepeiborra, line 77: I think we only run the test suite once per major version, here 9.4.3 would replace 9.4.2. 
```suggestion
           - os: ubuntu-latest
             ghc: '9.4.3'
             test: true
```

.github/workflows/test.yml
```diff
@@ -150,15 +153,15 @@ jobs:
           HLS_WRAPPER_TEST_EXE: hls-wrapper
         run: cabal test wrapper-test --test-options="$TEST_OPTS --rerun-log-file .tasty-rerun-log-wrapper"
 
-      - if: matrix.test && matrix.ghc != '9.2.4' && matrix.ghc != '9.4.2'
+      - if: matrix.test && matrix.ghc != '9.2.4' && matrix.ghc != '9.4.2' && matrix.ghc != '9.4.3'
```
pepeiborra, line 156: We don't need to filter 9.4.2 out since there will be no test run for it. Please apply to all other lines in the PR.
```suggestion
      - if: matrix.test && matrix.ghc != '9.2.4' && matrix.ghc != '9.4.3'
```

haskell-language-server.cabal
```diff
@@ -497,6 +497,9 @@ executable haskell-language-server
 
   default-language: Haskell2010
   default-extensions: DataKinds, TypeOperators
+  
+  if impl(ghc >= 9.4)
+    build-depends: ghc-exactprint >= 1.6.1
```
pepeiborra, line 502: why is this needed?

.github/workflows/test.yml
mouse07410: Thank you - done.

Do you think the same should be some with GHC-9.2.5? Drop 9.2.4 and replace it with 9.2.5?

haskell-language-server.cabal
mouse07410: Ah, a good question. Without this seemingly unnecessary constraint, Cabal somehow manages to pull `ghc-exactprint-1.6.0` - presumably because there's a dependency of dependency that's asking for it. And CI tests kept failing.

With this addition, all the relevant tests pass.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
```diff
@@ -45,6 +45,11 @@ library
     , process-extras  >= 0.7.1
     , text
 
+  if impl(ghc >= 9.2)
+    build-depends: fourmolu ^>= 0.8.2
+  if impl(ghc >= 9.4)
+    build-depends: fourmolu ^>= 0.9
+
```
georgefst, line 52: I'm a bit confused by exactly why this is done this way (same as https://github.com/haskell/haskell-language-server/pull/3364#discussion_r1031792168, basically). But if it passes CI I'm fine with it.

I'll revisit all these bounds next time I come along to bump the version.

georgefst (approved):

haskell-language-server.cabal
michaelpj: I really think this needs to go in the right place. It shouldn't be _that_ hard to figure out. At the _very_ least it needs a clear comment explaining this.

plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal
mouse07410: > I'm a bit confused by exactly why this is done this way

Mainly, because I don't know any better. It seemed to be a crude but correct way to get tests pass.

I'll be happy if somebody with more knowledge fixes this the right way.

mouse07410:
I don't know what to do with this test failure (GHC-9.4.3, ubuntu-latest):
```
doc
      local define:                                                                   OK (1.50s)
      local empty doc:                                                                OK (1.52s)
      local single line doc without '\n':                                             FAIL (expected: Completion doc doesn't support ghc9) (1.51s)
        test/exe/Main.hs:2044:
        expected: ["*Defined at line 3, column 1 in this module*\n* * *\ndocdoc\n"]
         but got: ["*Defined at line 3, column 1 in this module*\n"] (expected failure)
      local multi line doc with '\n':                                                 FAIL (expected: Completion doc doesn't support ghc9) (1.51s)
        test/exe/Main.hs:[204](https://github.com/haskell/haskell-language-server/actions/runs/3545320619/jobs/5953373235#step:8:205)4:
        expected: ["*Defined at line 4, column 1 in this module*\n* * *\n abcabc\n"]
         but got: ["*Defined at line 4, column 1 in this module*\n"] (expected failure)
      local multi line doc without '\n':                                              FAIL (expected: Completion doc doesn't support ghc9) (1.51s)
```

fendor:
The testcases you mention are expected failures and require no action.
However,
```
   addDependentFile
    file-changed
      test:                                                                           FAIL
        Last message received:
        {
            "jsonrpc": "2.0",
            "method": "window/logMessage",
            "params": {
                "message": "Live bytes: 29.27MB Heap size: 331.35MB",
                "type": 3
            }
        }
        Use -p '/file-changed.test/' to rerun this test only.
```

might be a genuine bug.

mouse07410:
Very likely so. But I don't see how this PR could be related, or how to fix it. Or even whether the fix should belong to this PR.

fendor:
Well, if the PR that adds support for GHC 9.4.3, and the test-suite for GHC 9.4.3 breaks, then it is related because GHC changed something, somehow and now some assumption we previously had does not hold anymore. In other words, HLS is *not* compatible with GHC 9.4.3 out of the box.

A fix may very well belong into this PR, and the fix is likely to be rather minor, but potentially very hard to track down.
It is in theory possible to disable this specific test if you have a good reason, but I think fixing it is the right course of action.

mouse07410:
> .  .  .  HLS is not compatible with GHC 9.4.3 out of the box.

I call it "limited support". ;-)

> It is in theory possible to disable this specific test if you have a good reason, but I think fixing it is the right course of action.

My only/main reason to ignore this test is that I don't have the slightest clue how to go about fixing, let alone actually fixing it. 

The change to make HLS build and pass the majority of the tests seems fairly small and simple. The choice appears to be between providing some support for GHC-9.4.3 now, and waiting for an unknown length of time until somebody with enough knowledge figures out what's wrong and makes all the tests pass. 

Also, what should we do with GHC-9.2.5? Stackage released LTS for it, but there's no official HLS support yet.

mouse07410:
Is there any way to make tests continue with other GHC versions rather than cancel, if one fails?

fendor:
> Is there any way to make tests continue with other GHC versions rather than cancel, if one fails?

No, this is by design to spare some CI resources.

mouse07410:
> > Is there any way to make tests continue with other GHC versions rather than cancel, if one fails?
> 
> No, this is by design to spare some CI resources.

I understand that. But here it looks like we need to know which GHC versions do and which do not fail the tests, with the given changes.

haskell-language-server.cabal
fendor: It is not necessary, afaict, this worked for me out of the box:
```
> cabal update
> cabal build -w ghc-9.4.3 exe:haskell-language-server
```

mouse07410: Interesting - perhaps, `index-state` changed since...? Should I update `index-state` in this PR too?

Also, have you tried just `cabal build -w ghc-9.4.3` (to build whatever's there)?

fendor: Since you updated the index-state in cabal.project, you locally need to make sure your global index-state is newer than the one in cabal.project (which you added). So just running `cabal update`  in the project or outside of it should do the trick.

Yeah, works just the same. It'd be weird if cabal didn't choose the most recent version of 'ghc-exactprint'.

pepeiborra:
@mouse07410 you could mark the test as an expected failure in ghc 9.4.5. This will allow CI to continue past the error.

mouse07410:
> @mouse07410 you could mark the test as an expected failure in ghc 9.4.5. This will allow CI to continue past the error.

Please excuse my ignorance. How to do that?

fendor:
Use the `knownBrokenFor` function in front of the test-case, e.g. `knownBrokenFor [GHC GHC94]  "Not supported on GHC 9.4.3" $ testSession ...`. Another example at https://github.com/haskell/haskell-language-server/blob/master/ghcide/test/exe/Main.hs#L1989

mouse07410:
> Use the `knownBrokenFor` function in front of the test-case, e.g. `knownBrokenFor [GHC GHC94] "Not supported on GHC 9.4.3" $ testSession ...`. Another example at https://github.com/haskell/haskell-language-server/blob/master/ghcide/test/exe/Main.hs#L1989

I see, thanks. Frankly, I think it is beyond my competency to start messing with the code. Perhaps, one of the maintainers would care to tweak these?

wavewave:
The currently broken master CI makes us hard to work on. (We are currently trying to land some changes for GHC 9.4 such as https://github.com/haskell/haskell-language-server/pull/3317 but could not make progress related to CI failures)
Is there any blocker for this to be merged and make CI succeed again? Thanks.
cc: @pepeiborra @michaelpj

pepeiborra:
> The currently broken master CI makes us hard to work on. (We are currently trying to land some changes for GHC 9.4 such as #3317 but could not make progress related to CI failures)
> Is there any blocker for this to be merged and make CI succeed again? Thanks.
> cc: @pepeiborra @michaelpj

@wavewave what is broken in master CI and what parts of this PR would help unblock you?

9999years:
> > The currently broken master CI makes us hard to work on. (We are currently trying to land some changes for GHC 9.4 such as #3317 but could not make progress related to CI failures)
> > Is there any blocker for this to be merged and make CI succeed again? Thanks.
> > cc: @pepeiborra @michaelpj
> 
> @wavewave what is broken in master CI and what parts of this PR would help unblock you?

C'mon, just look at the most recent `master` commit. It tells you exactly which checks failed:

https://github.com/haskell/haskell-language-server/commit/924b93247ec467dbd6839b37d124bdaa137700a0

You can see that most of the checks in https://github.com/haskell/haskell-language-server/pull/3317 are failing. We need this PR to fix those checks. I believe this PR will be necessary but perhaps not sufficient to fix those checks.

pepeiborra:
> C'mon, just look at the most recent `master` commit. It tells you exactly which checks failed:
> 
> [924b932](https://github.com/haskell/haskell-language-server/commit/924b93247ec467dbd6839b37d124bdaa137700a0)

Can you be more explicit? I don't see any checks failing other than the Nix ones.

> You can see that most of the checks in #3317 are failing. We need this PR to fix those checks. I believe this PR will be necessary but perhaps not sufficient to fix those checks.

I see that there is a build error in a dependency:

```
Building library for apply-refact-0.10.0.0..
[508](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:509)
[1 of 5] Compiling Refact.Compat    ( src/Refact/Compat.hs, dist/build/Refact/Compat.o, dist/build/Refact/Compat.dyn_o )
[509](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:510)

[510](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:511)
src/Refact/Compat.hs:183:15: error:
[511](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:512)
Error:     Not in scope: type constructor or class ‘ErrorMessages’
[512](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:513)
    |
[513](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:514)
183 | type Errors = ErrorMessages
[514](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:515)
    |               ^^^^^^^^^^^^^
[515](https://github.com/haskell/haskell-language-server/actions/runs/3681170875/jobs/6227583497#step:4:516)
Error: Process completed with exit code 1.
```

Can you clarify how this PR fixes that?

wavewave:
@pepeiborra 
I probably do not follow the current status about CI correctly. My question was at more high level, and the CI badge on master commits is shown red cross mark, and the PR title says "it fixes build", so was asking whether we can merge and fix the CI. Thanks for trying to clarify the actual issue further. 
The above issue you linked is another upstream issue for `apply-refact`, which should be solved by updating hackage with GHC-9.4 compatible version (the master of `apply-refact` is already updated) (note that this couldn't be solved locally here by `source-repository-package` since the newer version of `apply-refact` is breaking GHC 9.2 build.

pepeiborra:
Superseded by #3396

michaelpj:
It is certainly a problem that our CI checks are flaky, yeah. But I don't believe that master or anything else is actually broken.

Issue #3318, opened 2022-11-04
CI: Caching seems not to do anything


fendor:
CI runs are incredibly slow and seem to have no caching whatsoever:

![image](https://user-images.githubusercontent.com/8463814/199980678-50e799c9-883b-4012-8471-693d01b16360.png)

What's going on? At least dependencies should *always* be cached.

andys8:
Haven't looked into it, but want to share that there is a quite newish page that shows caching details for GitHub.

https://github.com/haskell/haskell-language-server/actions/caches

Also this subpage contains this warning - which might or might not be relevant for this issue.

![Screenshot_2022-11-05-18-27-19-92_4d38fce200f96aeac5e860e739312e76](https://user-images.githubusercontent.com/13085980/200133230-51020700-6c06-49e6-838f-ee3b8fb1d433.jpg)

fendor:
-_- Cant we just buy more cache? Or we have to smarter when it comes to artefact sharing.

hasufell:
10GB is nowhere near enough. 

We should use cabal-cache: https://hackage.haskell.org/package/cabal-cache

I'm currently trying to implemented it for ghcup CI as well. Is uses S3 and supports multicloud. We can just buy backblaze bucket with the opencollective money.

PR #3366, opened 2022-11-26
Bump gha versions in setup-build/action.yml

fendor:
CI is insanely slow. Only *some* caching of build dependencies.
Caching should take care of it, why are the caches not hit in some cases?

This PR is meant to investigate #3318 

Moreover, dependabot seems not to work for our custom defined github action?

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/3366"></a>

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -26,6 +26,8 @@ jobs:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/setup-build
       with:
+        # select a stable GHC version
+        ghc: 9.2.5
         os: ${{ runner.os }}
         shorten-hls: false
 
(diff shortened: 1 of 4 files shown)


fendor:
Afaict, this should speed up the CI for GHC 9.4 by caching

michaelpj (approved):

.github/workflows/pre-commit.yml
```diff
@@ -26,6 +26,8 @@ jobs:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/setup-build
       with:
+        # select a stable GHC version
+        ghc: 9.2.5
```
fendor, line 30: This makes me mad, this should be an error, since it is a required input. But apparently, validation only happens for externally defined actions?  I dunno.

PR #3373, opened 2022-12-02
nix: no more yolo on ghc 9.4; test suites generally work!

lf-:
I tried using HLS again in one of my projects and found that the nix packaging
isn't possible to use with other projects due to overriding mkDerivation
and causing world-rebuilds. (work vendored most of the HLS flake
because of this but I want to actually fix it upstream)

It turns out that the GHC 9.4 set is stable enough to remove the yolo anyway.

I've also added an overlay attribute that uses the nixpkgs set without
the *other* overrides to mkDerivation, for use as an overlay in other flakes.

### DIFF
--- a/configuration-ghc-94.nix
+++ b/configuration-ghc-94.nix
@@ -18,8 +18,9 @@ let
       # https://github.com/nikita-volkov/ptr-poker/issues/11
       ptr-poker = hself.callCabal2nix "ptr-poker" inputs.ptr-poker { };
 
+      # Freezes in test for some reason.
       ghc-exactprint =
-        hself.callCabal2nix "ghc-exactprint" inputs.ghc-exactprint-160 { };
+        dontCheck (hself.callCabal2nix "ghc-exactprint" inputs.ghc-exactprint-160 { });
       # Hlint is still broken
       hlint = doJailbreak (hself.callCabal2nix "hlint" inputs.hlint { });
 
@@ -28,7 +29,9 @@ let
       # Re-generate HLS drv excluding some plugins
       haskell-language-server =
         hself.callCabal2nixWithOptions "haskell-language-server" ./.
-        (pkgs.lib.concatStringsSep " " [ "-fpedantic" "-f-hlint" ]) { };
+        # Pedantic cannot be used due to -Werror=unused-top-binds
+        # Check must be disabled due to some missing required files
+        (pkgs.lib.concatStringsSep " " [ "--no-check" "-f-pedantic" "-f-hlint" ]) { };
     });
 in {
   inherit disabledPlugins;
--- a/configuration-ghc-94.nix
+++ b/configuration-ghc-94.nix
@@ -13,12 +13,6 @@ let
     with pkgs.haskell.lib;
     {
       hlsDisabledPlugins = disabledPlugins;
-      # YOLO
-      mkDerivation = args:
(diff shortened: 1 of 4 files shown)


lf-:
```
hls-cabal-plugin> src/Ide/Plugin/Cabal/Parse.hs:21:1: error:
hls-cabal-plugin>     Ambiguous module name ‘Distribution.Types.Version’:
hls-cabal-plugin>       it was found in multiple packages:
hls-cabal-plugin>       Cabal-3.6.3.0 Cabal-syntax-3.8.1.0
```

how does that even *happen*?!

lf-:
Ohhhh. Gross.

```
  # 2022-10-27: implicit-hie 0.1.3.0 needs a newer version of Cabal-syntax.
  implicit-hie = super.implicit-hie.override {
    Cabal-syntax = self.Cabal-syntax_3_8_1_0;
  };
```

lf-:
Yeah the build was jacked before I touched it, that explains a lot 😇

Explanation of the eldritch nix horrors that have been invoked here: what happens when you override boot packages is "you can't". So you *can't* override cabal to be newer without forcing the package set to be inconsistent and have multiple versions of the same package wind up in a package database and cause build failures. That is just what happens if you override a boot package. Nix/cabal goes all zalgo-text on you :(

fendor (approved):
Changes to the cabal plugin look good to me!

michaelpj (approved):

michaelpj:
Hmm, the Nix CI jobs are still failing, would you expect them to succeed?

guibou (approved):
Great.

Note for the future: it is possible to change the YOLO thing so the `jailbreak` and `doCheck` are only added if the package is `broken`. This improves a lot the caching.

lf-:
> Hmm, the Nix CI jobs are still failing, would you expect them to succeed?

nope, they were broken before i changed anything

lf-:
I believe that this doesn't make the CI state any worse, as far as I can tell. So it should be mergeable, with further CI fixes as future work.

lf-:
Aah, no, the benchmark build gets silly bounds problems since cabal-syntax and cabal have to be the same version, and cabal syntax 3.8 allows building on old ghc even though this causes this problem.

lf-:
I am not sure how to inject an *extra* constraint into the Cabal 3.6 cabal file. So I'm not sure how to fix this benchmark. Does anyone have any ideas?

cydparser:
@lf- Try removing Cabal from build-depends.

lf-:
yep you're right @cydparser. Going to see if this fixes the older CI as well.

michaelpj:
Let's see if we can get https://github.com/haskell/haskell-language-server/pull/3383 in, then?

cydparser:
@lf-, Cabal-syntax was removed as a dependency in the mainline branch. Try removing commit d5afadaa4effa0b5a36f42bdc847b4c77e78867e and rebasing.

lf-:
I think yes because it would utilize nixpkgs binary caches better to not disable all test suites. I'm not sure if I'll get around to rebase super soon.

cydparser:
@fendor, these changes should already be in master: https://github.com/haskell/haskell-language-server/pull/3503/commits

michaelpj:
I'm assuming this is stale, please reopen if not.

PR #3380, opened 2022-12-04
Less allow-newer

July541:
Try some clean up:)


July541:
Close due to can't repro on my m1 mac with ghc-9.2.

PR #3386, opened 2022-12-07
[hls-explicit-record-fields-plugin] Expand used fields only

ozkutuk:
This is a follow-up PR to #3304, which makes record wildcard expansion code action to only expand fields that are used.

### How it works?
Since the plugin already operates on `RenamedSource`, we have access to all the `Name`s that GHC has resolved for us. We utilize this information by first collecting all the `Name`s within the source file by a `syb` traversal, and then for every field of the record to be expanded we check if the `Name` of the field exists within this collected list of `Name`s (except for the field itself).

(@santiweight: you might be interested in this approach)

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -18,6 +18,7 @@ module Ide.Plugin.ExplicitFields
 import           Control.Lens                    ((^.))
 import           Control.Monad.IO.Class          (MonadIO, liftIO)
 import           Control.Monad.Trans.Except      (ExceptT)
+import           Data.Functor                    ((<&>))
 import           Data.Generics                   (GenericQ, everything, extQ,
                                                   mkQ)
 import qualified Data.HashMap.Strict             as HashMap
@@ -46,7 +47,8 @@ import           Development.IDE.GHC.Compat.Core (Extension (NamedFieldPuns),
                                                   hfbPun, hfbRHS, hs_valds,
                                                   lookupUFM, mapConPatDetail,
                                                   mapLoc, pattern RealSrcSpan,
-                                                  plusUFM_C, unitUFM)
+                                                  plusUFM_C, ufmToIntMap,
+                                                  unitUFM)
 import           Development.IDE.GHC.Util        (getExtensions,
                                                   printOutputable)
 import           Development.IDE.Graph           (RuleResult)
@@ -93,7 +95,7 @@ instance Pretty Log where
 descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
 descriptor recorder plId = (defaultPluginDescriptor plId)
   { pluginHandlers = mkPluginHandler STextDocumentCodeAction codeActionProvider
-  , pluginRules = collectRecordsRule recorder
+  , pluginRules = collectRecordsRule recorder *> collectNamesRule
   }
 
 codeActionProvider :: PluginMethodHandler IdeState 'TextDocumentCodeAction
@@ -148,11 +150,13 @@ collectRecordsRule recorder = define (cmapWithPrio LogShake recorder) $ \Collect
       let exts = getEnabledExtensions tmr
           recs = getRecords tmr
       logWith recorder Debug (LogCollectedRecords recs)
-      let names = getNames tmr
-          renderedRecs = traverse (renderRecordInfo names) recs
-          recMap = RangeMap.fromList (realSrcSpanToRange . renderedSrcSpan) <$> renderedRecs
-      logWith recorder Debug (LogRenderedRecords (concat renderedRecs))
-      pure ([], CRR <$> recMap <*> Just exts)
+      use CollectNames nfp >>= \case
+        Nothing -> pure ([], Nothing)
+        Just (CNR names) -> do
+          let renderedRecs = traverse (renderRecordInfo names) recs
+              recMap = RangeMap.fromList (realSrcSpanToRange . renderedSrcSpan) <$> renderedRecs
+          logWith recorder Debug (LogRenderedRecords (concat renderedRecs))
+          pure ([], CRR <$> recMap <*> Just exts)
 
   where
     getEnabledExtensions :: TcModuleResult -> [GhcExtension]
@@ -162,10 +166,16 @@ getRecords :: TcModuleResult -> [RecordInfo]
 getRecords (tmrRenamed -> (hs_valds -> valBinds,_,_,_)) =
   collectRecords valBinds
 
+collectNamesRule :: Rules ()
+collectNamesRule = define mempty $ \CollectNames nfp ->
+  use TypeCheck nfp <&> \case
+    Nothing  -> ([], Nothing)
+    Just tmr -> ([], Just (CNR (getNames tmr)))
+
 -- | Collects all 'Name's of a given source file, to be used
 -- in the variable usage analysis.
-getNames :: TcModuleResult -> UniqFM Name [LocatedN Name]
-getNames (tmrRenamed -> (group,_,_,_)) = collectNames group
+getNames :: TcModuleResult -> NameMap
+getNames (tmrRenamed -> (group,_,_,_)) = NameMap (collectNames group)
 
 data CollectRecords = CollectRecords
                     deriving (Eq, Show, Generic)
@@ -186,12 +196,35 @@ instance Show CollectRecordsResult where
 
 type instance RuleResult CollectRecords = CollectRecordsResult
 
+data CollectNames = CollectNames
+                  deriving (Eq, Show, Generic)
+
+instance Hashable CollectNames
+instance NFData CollectNames
+
+data CollectNamesResult = CNR NameMap
+  deriving (Generic)
+
+instance NFData CollectNamesResult
+
+instance Show CollectNamesResult where
+  show _ = "<CollectNamesResult>"
+
+type instance RuleResult CollectNames = CollectNamesResult
+
 -- `Extension` is wrapped so that we can provide an `NFData` instance
 -- (without resorting to creating an orphan instance).
 newtype GhcExtension = GhcExtension { unExt :: Extension }
 
 instance NFData GhcExtension where
   rnf x = x `seq` ()
+
+-- As with `GhcExtension`, this newtype exists mostly to attach
+-- an `NFData` instance to `UniqFM`.
+newtype NameMap = NameMap (UniqFM Name [LocatedN Name])
+
+instance NFData NameMap where
+  rnf (NameMap (ufmToIntMap -> m)) = rnf m
 
 data RecordInfo
   = RecordInfoPat RealSrcSpan (Pat (GhcPass 'Renamed))
@@ -212,7 +245,7 @@ instance Pretty RenderedRecordInfo where
 
 instance NFData RenderedRecordInfo
 
-renderRecordInfo :: UniqFM Name [LocatedN Name] -> RecordInfo -> Maybe RenderedRecordInfo
+renderRecordInfo :: NameMap -> RecordInfo -> Maybe RenderedRecordInfo
 renderRecordInfo names (RecordInfoPat ss pat) = RenderedRecordInfo ss <$> showRecordPat names pat
 renderRecordInfo _ (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordCon expr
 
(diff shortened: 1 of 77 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -137,23 +147,42 @@ codeActionProvider ideState pId (CodeActionParams _ _ docId range _) = pluginRes
...
-  tmr <- use TypeCheck nfp
-  let exts = getEnabledExtensions <$> tmr
-      recs = concat $ maybeToList (getRecords <$> tmr)
-  logWith recorder Debug (LogCollectedRecords recs)
-  let renderedRecs = traverse renderRecordInfo recs
-      recMap = RangeMap.fromList (realSrcSpanToRange . renderedSrcSpan) <$> renderedRecs
-  logWith recorder Debug (LogRenderedRecords (concat renderedRecs))
-  pure ([], CRR <$> recMap <*> exts)
+collectRecordsRule recorder = define (cmapWithPrio LogShake recorder) $ \CollectRecords nfp ->
+  justOrFail "Unable to TypeCheck" (use TypeCheck nfp) $ \tmr -> do
```
ozkutuk, line 151: Is using `fail` this way within a plugin rule acceptable? I didn't want to propagate the `Maybe` all the way to the end, especially when the plugin can't really do anything if it can't get a hold of the `TcModuleResult`. I am open to other suggestions regarding this.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -137,23 +147,42 @@ codeActionProvider ideState pId (CodeActionParams _ _ docId range _) = pluginRes
...
+getNames (tmrRenamed -> (group,_,_,_)) = collectNames group
+
+newtype UniqueKey = UniqueKey Unique
+                  deriving newtype Eq
+
+getUniqueKey :: Name -> UniqueKey
+getUniqueKey = UniqueKey . nameUnique
+
+instance Ord UniqueKey where
+  compare (UniqueKey u1) (UniqueKey u2) = getUnique u1 `nonDetCmpUnique` getUnique u2
```
ozkutuk, line 184: Based on what I understood from the notes in GHC source (namely, `No Ord for Unique` and `Unique Determinism`) it should be safe to use `nonDetCmpUnique` here to be able to use `Unique`s as map keys. Still it would be good to get an approval regarding this.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -137,23 +147,42 @@ codeActionProvider ideState pId (CodeActionParams _ _ docId range _) = pluginRes
...
     getEnabledExtensions = map GhcExtension . getExtensions . tmrParsed
 
+    justOrFail :: MonadFail m => String -> m (Maybe a) -> (a -> m b) -> m b
+    justOrFail = flip . maybeM . fail
+
 getRecords :: TcModuleResult -> [RecordInfo]
 getRecords (tmrRenamed -> (hs_valds -> valBinds,_,_,_)) =
   collectRecords valBinds
 
+-- | Collects all 'Name's of a given source file, to be used
```
michaelpj, line 172: Is this something useful for other plugins? It could potentially be in its own rule.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: ... and whatever reasoning you arrive at should go into the code as a comment

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -199,9 +228,41 @@ instance Pretty RenderedRecordInfo where
...
-renderRecordInfo :: RecordInfo -> Maybe RenderedRecordInfo
-renderRecordInfo (RecordInfoPat ss pat) = RenderedRecordInfo ss <$> showRecordPat pat
-renderRecordInfo (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordCon expr
+renderRecordInfo :: Map UniqueKey [LocatedN Name] -> RecordInfo -> Maybe RenderedRecordInfo
+renderRecordInfo names (RecordInfoPat ss pat) = RenderedRecordInfo ss <$> showRecordPat names pat
+renderRecordInfo _ (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordCon expr
+
+-- | Checks if a 'Name' is referenced in a given list of names. The 'Eq'
+-- instance of 'Name's makes use of their unique identifiers, hence any
+-- to 'Name' referring to the same entity is considered equal. In order
```
michaelpj, line 237:
```suggestion
-- 'Name' referring to the same entity is considered equal. In order
```

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -199,9 +228,41 @@ instance Pretty RenderedRecordInfo where
...
 instance NFData RenderedRecordInfo
 
-renderRecordInfo :: RecordInfo -> Maybe RenderedRecordInfo
-renderRecordInfo (RecordInfoPat ss pat) = RenderedRecordInfo ss <$> showRecordPat pat
-renderRecordInfo (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordCon expr
+renderRecordInfo :: Map UniqueKey [LocatedN Name] -> RecordInfo -> Maybe RenderedRecordInfo
+renderRecordInfo names (RecordInfoPat ss pat) = RenderedRecordInfo ss <$> showRecordPat names pat
+renderRecordInfo _ (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordCon expr
+
+-- | Checks if a 'Name' is referenced in a given list of names. The 'Eq'
```
michaelpj, line 235: None of the arguments are a list of names?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -199,9 +228,41 @@ instance Pretty RenderedRecordInfo where
...
+-- instance of 'Name's makes use of their unique identifiers, hence any
+-- to 'Name' referring to the same entity is considered equal. In order
+-- to ensure that no false-positive is reported (in the case where the
+-- 'name' itself is part of the given list), the inequality of source
+-- locations is also checked.
```
michaelpj, line 240: Perhaps an example? I couldn't quite figure out why this matters, and tbh it seems rather weird to have a match on the unique but not on the srcspan?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -212,8 +273,13 @@ renderRecordInfo (RecordInfoCon ss expr) = RenderedRecordInfo ss <$> showRecordC
 -- as we want to print the records in their fully expanded form.
 -- Here `rec_dotdot` is set to `Nothing` so that fields are printed without
 -- such post-processing.
```
michaelpj, line 275: I think this comment could do with some tweaks. It suggests that this is primarily about printing the records which makes it sounds like you always print them as they are, but now you're adding some additional logic to print something else depending on the fields in use.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -137,23 +147,42 @@ codeActionProvider ideState pId (CodeActionParams _ _ docId range _) = pluginRes
...
 
+    justOrFail :: MonadFail m => String -> m (Maybe a) -> (a -> m b) -> m b
+    justOrFail = flip . maybeM . fail
+
 getRecords :: TcModuleResult -> [RecordInfo]
 getRecords (tmrRenamed -> (hs_valds -> valBinds,_,_,_)) =
   collectRecords valBinds
 
+-- | Collects all 'Name's of a given source file, to be used
+-- in the variable usage analysis.
```
michaelpj, line 173: More specifically, it collects all the names and a backwards lookup to the original AST node that the name was harvested from?

michaelpj: And it seems like that last bit is mostly because of the srcspan check you do later, which I still don't really understand the need for. Is there a test case that fails without it?

michaelpj (approved):
Nice!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: The `fail` used here makes use of `MonadFail Action`, which amounts to `MonadFail IO`. Since `fail @IO` just raises a user exception, this is a no-go then. I'll try to come up with an alternative solution.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: It is easier to explain over an example:
```haskell
module Test where

f :: Int -> Int
f foo = foo + 3 * foo + let foo = 5 in foo + 1
```
In this code segment, there are 2 **unique** names (excluding the names not related to `foo`), namely the `foo` that is the function argument and the `foo` which is let-bound. However, there are 5 names in total (again, that occurs as `foo` in the source). GHC lists all these as follows:
```
({ Test.hs:4:3-5 } {Name: foo_atU})
({ Test.hs:4:9-11 } {Name: foo_atU})
({ Test.hs:4:19-21 } {Name: foo_atU})
({ Test.hs:4:29-31 } {Name: foo_awb})
({ Test.hs:4:40-42 } {name: foo_awb})
```
Indeed, this is what we expected. There are 3 references to the `foo` that is the argument to the function, **one of which is the argument itself.** If we are only concerned with whether this argument is used, we shouldn't consider this first `Name`, and check if there are other `Name`s with the same `Unique` (in this example, `foo_atU`). The source span inequality check ensures that we in fact exclude this first `Name` that corresponds to the argument in our considerations.

---

Now, it is obvious that such filtering is necessary. However, considering the fact that we build a map keyed with the `Unique`s of `Name`s, we can simplify our filtering, and probably can just check if the list of `Name`s corresponding to a `Unique` has length greater than 1 (since we know the name at the binding-site is always present). The initial implementation worked with just a big list of `Name`s rather than building a map, and this check involving source spans was carried over from that implementation.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: I'm not sure: we might catch it somewhere? I think it's probably bad. Maybe @pepeiborra knows? If it's not okay, we can potentially delete the `MonadFail Action` instance.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: Hmmmm. Maybe we're not quite stating the problem right? From your example I would say that the problem is that you are looking for all _uses_ of `foo`, but your naive traversal also turns up _definitions_ of `foo`. I'm kind of surprised that GHC doesn't have some tools for finding uses... but maybe it has a way to tell whether something is a definition? I'm not entirely convinced that your solutions as described will work. e.g. for "list of names greater than 1", consider
```
haha :: Int -> Int -> Int
haha foo 1 = 2
haha foo _ = 3
```
... although maybe that will have different uniques for the different `foo`s? Unclear. That makes me wonder if there's something more robust to be had in GHC. Maybe worth asking the GHC devs.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I replaced the fail usage with more straightforward `Maybe` handling, but still it would be good to know if using the `MonadFail Action` instance is okay.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: While reading more of the GHC notes, I realized GHC provides a map type exactly for this purpose: creating maps indexed with uniques. I have opted to used that instead of rolling my own solution.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: Turns out that it's fine and exceptions will get turned into a diagnostic. Perhaps not a very good one, though. So up to you what you want to do here.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I have asked in GHC chatroom whether GHC provides anything to find the _uses_ of a `Name`, but it doesn't seem to be the case. In the example you provided it isn't an issue because as you pointed out GHC indeed assigns different uniques to those `foo`s. I don't expect there to be an edge-case regarding this approach, since GHC does all the heavy-lifting regarding the name resolution. We merely make use of a single assumption: that given a list of names having same uniques, only a single of them is the defining one. I would be really surprised if there was a case where this does not hold.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: I have revised this and a bunch of other comments, can you take another look to see if it looks better?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: I think you're probably right, it just feels like a thing that might have nasty edge cases lurking. But if GHC doesn't give us anything (surprisingly?) then this seems good to me!

michaelpj (approved):
great stuff!

PR #3390, opened 2022-12-09
Unload once per linkable instead of once per splice

wz1000:
We need to unload old linkables before we can load in new linkables. However,
the `unload` function in the GHC API takes a list of linkables to **keep** (i.e.
not unload). Earlier we `unload`ed right before loading in new linkables, which
is effectively once per splice. This can be slow as `unload` needs to walk over
the list of all loaded linkables, for each splice.

Instead, now we unload old linkables right after we generate a new linkable and
just before returning it to be loaded. This has a substantial effect on recompile
times as the number of loaded modules and splices increases.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -160,8 +160,7 @@ computePackageDeps env pkg = do
 
 data TypecheckHelpers
   = TypecheckHelpers
-  { getLinkablesToKeep :: !(IO (ModuleEnv UTCTime))
-  , getLinkables       :: !([NormalizedFilePath] -> IO [LinkableResult])
+  { getLinkables       :: !([NormalizedFilePath] -> IO [LinkableResult])
   }
 
 typecheckModule :: IdeDefer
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -160,8 +160,7 @@ computePackageDeps env pkg = do
 
 data TypecheckHelpers
   = TypecheckHelpers
-  { getLinkablesToKeep :: !(IO (ModuleEnv UTCTime))
-  , getLinkables       :: !([NormalizedFilePath] -> IO [LinkableResult])
+  { getLinkables       :: !([NormalizedFilePath] -> IO [LinkableResult])
```
michaelpj, line 163: haddock?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1119,7 +1118,10 @@ getLinkableRule recorder =
         -- Record the linkable so we know not to unload it
         whenJust (hm_linkable =<< hmi) $ \(LM time mod _) -> do
             compiledLinkables <- getCompiledLinkables <$> getIdeGlobalAction
-            liftIO $ void $ modifyVar' compiledLinkables $ \old -> extendModuleEnv old mod time
+            liftIO $ modifyVar compiledLinkables $ \old -> do
+              let !to_keep = extendModuleEnv old mod time
```
michaelpj, line 1122: Surely _some_ of the explanation you put in the PR description should be written in the code for future people?

michaelpj (commented):
Just my usual plea for documentation.

ghcide/src/Development/IDE/Core/Rules.hs
wz1000: Added.

ghcide/src/Development/IDE/Core/Compile.hs
wz1000: added.

ghcide/src/Development/IDE/Core/Rules.hs
wz1000: Sorry, forgot to include the file. I've updated it now.

michaelpj (approved):

PR #3391, opened 2022-12-09
Eval plugin for GHC 9.4

wz1000:
Supersedes #3249 and depends on #3390

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -148,7 +148,7 @@ jobs:
         name: Test hls-pragmas-plugin
         run: cabal test hls-pragmas-plugin --test-options="$TEST_OPTS" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test hls-pragmas-plugin --test-options="$TEST_OPTS"
 
-      - if: matrix.test && matrix.ghc != '9.4.4'
+      - if: matrix.test
         name: Test hls-eval-plugin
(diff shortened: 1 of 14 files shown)


michaelpj:
Looks like Pepe's question never got answered (https://github.com/haskell/haskell-language-server/pull/3249#discussion_r990634700), but I'm okay with this anyway. Has conflicts now.

michaelpj:
What's the status of this?

wz1000:
This is ready for review

@pepeiborra `setupDynFlagsForGHCiLike` doesn't do much as setting the optimisation level to 0 sets the same set of flags.

pepeiborra:
> The plugin was implemented by calling "load" which circumvents all of HLSs caching mechanisms for interface files and linkables. Instead we should work like the other typechecking functions which get all the stuff we need using HLS rules and setup the HscEnv with all the state in the right places.

Is there any way to check this in a test? Or a benchmark? Seems like an important property.

pepeiborra (approved):

wz1000:
> > The plugin was implemented by calling "load" which circumvents all of HLSs caching mechanisms for interface files and linkables. Instead we should work like the other typechecking functions which get all the stuff we need using HLS rules and setup the HscEnv with all the state in the right places.
> 
> Is there any way to check this in a test? Or a benchmark? Seems like an important property.

I don't thin there is an easy way to do this. We can blacklist the `load` family of functions using hlint.

michaelpj:
Blacklisting the functions sounds good!

wz1000:
I have added an hlint rule and also fixed a regression that caused the eval plugin not to work due to https://github.com/haskell/haskell-language-server/pull/3414 landing. See the comment in the last commit for details.

PR #3393, opened 2022-12-12
Haskell to Latex Plugin - WIP

bolt12:
This is a WIP of a Haskell to Latex Plugin. This plugin will generate the latex equivalent of a Haskell function and display it on hover. It can also give you the latex on demand via code action.

This plugin was inspired by latexify for Python: https://github.com/google/latexify_py

Since Haskell's AST is very large and complex, I decided to just support the simpler constructs so as to not deal with any indentation pretty printing. Suggestions and improvements are always welcome 😃 

This plugin works by parsing and type-checking the buffer file and looking for the function binding closer to the cursor. After that it traverses the data structure and generates the latex code. When it is finished, it will write the latex to a temporary file and run a script to compile the latex and get a png out of the resulting pdf. It displays the png in the hover message and accordingly cleans all leftover files.

It depends on:

- `imagemagick`;
- `texlive combined scheme-full` (in order to get the `standalone` format).

Currently the plugin is a bit flaky as it is not always capable of generating the latex image, but a few hover tries should do the trick. Another bug is that it seems I haven't managed to fetch only the bindings for the function which I am hovering, so if there are other functions with the same name in the file (no matter what scope) the plugin is going to display it as well.

Here are some examples of what this plugin is capable of (sorry for the low res):

20221212_18h24m40s_grim

20221212_18h27m22s_grim

20221212_18h29m21s_grim

20221212_18h31m24s_grim

20221212_18h34m52s_grim

### DIFF
--- a/plugins/hls-latex-plugin/src/Ide/Plugin/Latex.hs
+++ b/plugins/hls-latex-plugin/src/Ide/Plugin/Latex.hs
@@ -1,0 +1,558 @@
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE DeriveAnyClass        #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PackageImports        #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE StandaloneDeriving    #-}
+{-# LANGUAGE TypeApplications      #-}
+
+module Ide.Plugin.Latex where
+
+import           Control.Applicative                                 (liftA2,
+                                                                      liftA3)
+import           Control.Exception                                   (SomeException (SomeException),
+                                                                      catch,
+                                                                      try)
+import           Control.Monad                                       (foldM_,
+                                                                      (>=>))
+import           Control.Monad.IO.Class                              (liftIO)
+import           Data.Aeson                                          (FromJSON,
+                                                                      ToJSON,
+                                                                      Value (..),
+                                                                      toJSON)
+import           Data.Foldable                                       (toList)
+import           Data.List                                           (intercalate,
+                                                                      intersperse)
+import           Data.Maybe                                          (catMaybes,
+                                                                      mapMaybe)
+import qualified Data.Text                                           as Text
+import qualified Data.Text.IO                                        as Text
+import qualified Debug.Trace                                         as Debug
+import           Development.IDE                                     hiding
+                                                                     (PFailed,
+                                                                      POk,
+                                                                      ParseResult,
+                                                                      align,
+                                                                      braces,
+                                                                      indent,
+                                                                      runParser,
+                                                                      space)
+import           Development.IDE.Core.Compile                        (lookupName)
+import qualified Development.IDE.GHC.Compat                          as GHC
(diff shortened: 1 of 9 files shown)


michaelpj:
Cool!

My _personal_ feeling is that this is a bit niche to make it into HLS mainline, but if you're keen to maintain it that's not necessarily a problem. I wish we had a nice way for people to mix in their own, separately maintained plugins, but we don't at the moment...

bolt12:
Thanks for your reply!

> I wish we had a nice way for people to mix in their own, separately maintained plugins, but we don't at the moment...

That would certainly be a cool thing. I think for now a good compromise if you think this is not fit for HLS mainline would be to leave this as a PR or just a branch and maybe have a section in README for external plugins from the community. I would be available to keep the branch up to date and rebased but as soon as me or someone would show a lack of care that plugin could be marked as deprecated in the README.

plugins/hls-latex-plugin/src/Ide/Plugin/Latex.hs
```diff
@@ -0,0 +1,558 @@
...
+descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
+descriptor recorder plId =
+  (defaultPluginDescriptor plId)
+    { pluginHandlers = mkPluginHandler STextDocumentCodeAction codeActionProvider
+                    <> mkPluginHandler STextDocumentHover renderLatexOnHover
+    , pluginCommands = [PluginCommand "sendLatexNotification"
+                                      "Show Latex for this function"
+                                      sendLatexNotificationCommand
+                       ]
+    }
```
pepeiborra, line 169: If the plugin is flaky you should make it an "opt-in" plugin, disabled by default:

```suggestion
    { pluginHandlers = mkPluginHandler STextDocumentCodeAction codeActionProvider
                    <> mkPluginHandler STextDocumentHover renderLatexOnHover
    , pluginCommands = [PluginCommand "sendLatexNotification"
                                      "Show Latex for this function"
                                      sendLatexNotificationCommand
                       ]
     , configInitialGenericConfig = def {
        plcGlobalOn = False
      }
    }
```
Then you can enable it by setting the LSP property below in your editor config:
```
    "haskell.plugin.latex.globalOn": true,
```

When we make an HLS release, @fendor or someone else will also release the Haskell VSCode extension and your plugin will show up in the configuration pages, being easier to discover and enable

michaelpj:
I didn't mean to sound too negative - if other people are keen and you're happy to maintain it, then we can have it in and optional.

bolt12:
The plugin already uses ghc-lib-parser so it should be easy to maintain different GHC versions. And it is simple enough (at least for now) to require minimal maintenance which I am willing to do. I'll polish this a little bit more and make the required changes!

plugins/hls-latex-plugin/src/Ide/Plugin/Latex.hs
bolt12: I was also kind of hoping someone more knowledgeable than me could look into the code and see if my cursor logic to get the function and whatnot could be improved, since I kind of just went for a trial and error and settle with the first thing that worked

michaelpj: Sadly we're a bit low on review cycles and it's a decent amount of code...

bolt12: That's okay, I understand, but I'd still appreciate if someone could go through the code and offer feedback. I have time, I'll just be randomly bumping this every once in a while 😛

PR #3394, opened 2022-12-12
hls-class-plugin - Add placeholders for all class methods

### LINKED ISSUE #3017: Add placeholders for all class instance methods
In #2920, we've got all class instance methods and dropped them immediately. So it's not hard to pick them up and supply `add placeholder for xxx` even if they are not the minimal requirements.

The main point is how to display these code actions properly because there may be so many items. Reference: https://github.com/haskell/haskell-language-server/pull/2920#discussion_r890959437

I'm willing to offer help if someone plans to pick this:)

batkot:
Added code action for creating placeholders for all not implemented class methods. Action is skipped when only required methods are missing to avoid duplication. 

Currently it's available only when minimal class definition is not met, wasn't sure how to locate class instances :thinking: 

missing-placeholders

Close #3017


Ailrun (commented):
I don't have enough time at least for now, sorry for the delay. This is some minor comment. Hopefully I could have some timespan soon for a fuller review.

michaelpj:
Maybe @July541 could take a look?

July541:
I can do a review this weekend if @Ailrun is busy still.

plugins/hls-class-plugin/test/testdata/AllMethodsRequired.hs
```diff
@@ -0,0 +1,9 @@
+module AllMethodsRequired where
```
July541, line 1: A corresponding expected file is required, or the test will succeed forever.

plugins/hls-class-plugin/test/testdata/MinimalDefinitionMeet.hs
```diff
@@ -0,0 +1,6 @@
+module MinimalDefinitionMeet where
```
July541, line 1: expected file.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -113,30 +115,31 @@ codeAction recorder state plId (CodeActionParams _ _ docId _ context) = pluginRe
             logWith recorder Info (LogImplementedMethods cls implemented)
             pure
                 $ concatMap mkAction
-                $ fmap (filter (\(bind, _) -> bind `notElem` implemented))
-                $ minDefToMethodGroups range sigs
-                $ classMinimalDef cls
+                $ nubOrdOn snd
```
July541, line 118: Is this for `Action is skipped when only required methods are missing to avoid duplication`?

Can you add a test for this?

July541 (changes requested):
Many thanks for your improvement! please add tests for it robust.

July541:
> Currently it's available only when minimal class definition is not met, wasn't sure how to locate class instances 🤔

class-plugin relies on ghc diagnostic to produce code actions, you can't have code actions for all instances under the current mechanism.

https://github.com/haskell/haskell-language-server/blob/dfd8f0b402a47588f8245b384dbd18a741dbf4f1/plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs#L207-L208

plugins/hls-class-plugin/test/testdata/AllMethodsRequired.hs
batkot: Those files (`AllMethodsRequired.hs` and `MinimalDefinitionMeet.hs` ) are used to only to [check which code actions are available](https://github.com/batkot/haskell-language-server/blob/a704dfe21329d6d555988e4121f4db2fd8f464cd/plugins/hls-class-plugin/test/Main.hs#L72-L76). There are no golden tests for them, so we don't need any  `*.expected.hs` files.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
batkot: I did add test for that :wink: Maybe naming is a bit unfortunate? :thinking: 

[`Add placeholders for all missing methods is unavailable when all methods are required`](https://github.com/batkot/haskell-language-server/blob/a704dfe21329d6d555988e4121f4db2fd8f464cd/plugins/hls-class-plugin/test/Main.hs#L73-L76)

batkot:
Thanks for review, @July541  :+1: 

> class-plugin relies on ghc diagnostic to produce code actions, you can't have code actions for all instances under the current mechanism.

Yeah I noticed that, I was just wondering if there's a nicer way to do that :thinking: But I don't know GHC codebase that well (not to say at all) :sweat_smile:

July541 (approved):
Everything looks fine, thank you for your work!

July541:
> I was just wondering if there's a nicer way to do that

I think we can ignore diagnostics and emit the code actions directly.

PR #3397, opened 2022-12-14
Correct list of GHC versions in caching.yml to match test.yml

fendor:
Small addition to #3396, exclude list wasn't completely up-to-date

Fixes Caching CI run https://github.com/haskell/haskell-language-server/actions/runs/3693755201/jobs/6254126700


michaelpj:
Something like this? https://github.com/orgs/community/discussions/26284

michaelpj:
Seems kind of hacky though.

Issue #3362, opened 2022-11-22
Respect `ImportQualifiedPost`?


kindaro:
## Is your enhancement request related to a problem? Please describe.

I often ask HLS to import stuff for me. For example, if I write `Text.toLower`, I can then ask HLS to `import qualified Data.Text as Text` — fantastic! However, I often have the extension `ImportQualifiedPost` on, so my imports should look like `import Data.Text qualified as Text`. I have to make this change by hand.

## Describe the solution you'd like

I want HLS to see that `ImportQualifiedPost` is on, and write qualified imports appropriately.

## Describe alternatives you've considered

I can format with `ormolu` or `fourmolu` — it automatically respects `ImportQualifiedPost`. But this is another distraction, another step that needs to be consciously and appropriately executed. Only a few of the projects I work on are thoroughly automatically formatted, and I cannot blindly format a file because this would incur an outsized diff. Therefore, I should have to tell the formatter which lines to format. This is too hard — I should be better off repairing the problematic import line by hand.

PR #3399, opened 2022-12-14
Add qualified imports in postfix position when `ImportQualifiedPost` and `WarnPrePositiveQualifiedModule` are set

3kyro:
With `ImportQualifiedPost` language extension and the `-warnPrePositiveQualifiedModule` enabled, the default behavior of the import-export plugin to add qualified imports in prefix position produces a warning. This PR changes the position of the `qualified`  descriptor to postfix when the above two conditions are met. This is a partial fix for #3362.


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1421,8 +1426,8 @@ suggestNewOrExtendImportForClassMethod packageExportsMap ps fileContents Diagnos
             | otherwise -> []
 #endif
 
-suggestNewImport :: ExportsMap -> Annotated ParsedSource -> T.Text -> Diagnostic -> [(T.Text, CodeActionKind, TextEdit)]
-suggestNewImport packageExportsMap ps fileContents Diagnostic{_message}
+suggestNewImport :: DynFlags -> ExportsMap -> Annotated ParsedSource -> T.Text -> Diagnostic -> [(T.Text, CodeActionKind, TextEdit)]
```
fendor, line 1429: I, personally, dislike passing around `DynFlags` since they are GHC internals subject to a lot of change. Also, contain a plethora of information unimportant to us.

Are you opposed replacing this with a type, for example:
```haskell
data QualifiedImportStyle = QualifiedImportPostfix | QualifiedImportPrefix
```

and calling functions such as `suggestNewImport` with the `QualifiedImportStyle`?
Name subject to bikeshedding.

3kyro: Sorry for the delay. I've added `QualifiedImportStyle` but kept passing `DynFlags` to `suggestNewImport` as the other options (add `QualifiedImportStyle` to `CodeActionArgs` or somehow compute it inside `iePluginDescriptor` and pass it along) don't seem ideal. I think using the `ToCodeAction` instance for `DynFlags` here is enough?

fendor (changes requested):
LGTM, look great! Tell me if I am wrong, but it seems there is no test-case for the postfix position import?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
fendor: Fair enough, SGTM!

3kyro:
@fendor thanks,  I couldn't find a way to set ` -Wprepositive-qualified-module` in the test environment and thought I might as well add an extra regression test. But it is obviously possible with `GHC_OPTIONS` 🤦🏽. I've added it now.

fendor (approved):
LGTM, nice contribution!

PR #3400, opened 2022-12-14
Fix fourmolu with -f-fixity-th in nix env

wavewave:
on current master, `.#haskell-language-server-925-dev-nix` shell is broken. 
```
% nix develop .#haskell-language-server-925-dev-nix
warning: Using saved setting for 'extra-substituters = https://haskell-language-server.cachix.org' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = haskell-language-server.cachix.org-1:juFfHrwkOxqIOZShtC4YC1uT1bBcq2RSvC7OMKx0Nz8=' from ~/.local/share/nix/trusted-settings.json.
[1/44/249 built, 3 copied (1489.5 MiB), 127.6 MiB DL] building HsYAML-0.2.1.1 (buildPhase): [ 4 of 14] Compiling Data.YAML.Token  ( src/Derror: interrupted by the user
ianwookim@IANWOOs-MacBook-Pro haskell-language-server % nix develop -j auto .#haskell-language-server-925-dev-nix
warning: Using saved setting for 'extra-substituters = https://haskell-language-server.cachix.org' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = haskell-language-server.cachix.org-1:juFfHrwkOxqIOZShtC4YC1uT1bBcq2RSvC7OMKx0Nz8=' from ~/.local/share/nix/trusted-settings.json.
[5/191/205 built] building ghc-lib-parser-9.2.4.20220729 (buildPhase): [330 of 333] Compiling GHC.Tc.Erro[1/202/205 built] building fourmolu-0.9.0.0 (buildPhase): [ 5 of 55] Compiling Ormolu.Fixity    ( serror: builder for '/nix/store/36r7hcfph5f8xn2vcs0sn5s2rba5vjly-fourmolu-0.9.0.0.drv' failed with exit code 1;
       last 10 log lines:
       > [54 of 55] Compiling Ormolu.Utils.Fixity ( src/Ormolu/Utils/Fixity.hs, dist/build/Ormolu/Utils/Fixity.o, dist/build/Ormolu/Utils/Fixity.dyn_o )
       > [55 of 55] Compiling Ormolu           ( src/Ormolu.hs, dist/build/Ormolu.o, dist/build/Ormolu.dyn_o )
       > Preprocessing executable 'fourmolu' for fourmolu-0.9.0.0..
[...]

PR #3383, opened 2022-12-06
Remove Cabal dependency from hls-cabal-plugin

cydparser:
Only Cabal-syntax is needed. I was unable to build using Nix without this change due to errors like the following:

```
src/Ide/Plugin/Cabal/Diagnostics.hs:17:1: error:
    Ambiguous module name ‘Distribution.Fields’:
      it was found in multiple packages:
      Cabal-3.6.3.0 Cabal-syntax-3.8.1.0
```

PR #3404, opened 2022-12-16
Fix nix build CI 

wavewave:
flakes don't need the special installer anymore
https://github.com/cachix/install-nix-action/commit/46187b354da8be43b69ee117d29793fa8b0ee3ea

This fixes the nix build CI, which has been broken for a while with this permission error:

 
Screen Shot 2022-12-15 at 11 29 50 PM

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -65,7 +65,10 @@ constraints:
   -- https://github.com/ndmitchell/hlint/issues/1376
   hlint +ghc-lib,
   ghc-lib-parser-ex -auto,
-  stylish-haskell +ghc-lib
+  stylish-haskell +ghc-lib,
+  -- implicit-hie 0.1.3.0 introduced an unexpected behavioral change.
+  -- https://github.com/Avi-D-coder/implicit-hie/issues/50
+  implicit-hie < 0.1.3
 
 -- This is benign and won't affect our ability to release to Hackage,
 -- because we only depend on `ekg-json` when a non-default flag
--- a/cabal.project
+++ b/cabal.project
@@ -65,10 +65,7 @@ constraints:
   -- https://github.com/ndmitchell/hlint/issues/1376
   hlint +ghc-lib,
   ghc-lib-parser-ex -auto,
-  stylish-haskell +ghc-lib,
-  -- implicit-hie 0.1.3.0 introduced an unexpected behavioral change.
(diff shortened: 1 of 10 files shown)


wavewave:
This branch merged #3400 and has a fix for hls-cabal-plugin (in an opposite way to #3383, by removing Cabal-syntax, not cabal, for making stackage test pass)
I think whether to choose Cabal or Cabal-syntax can be determined once we have the CI on master green.

wavewave:
also cherry-picked the commit "implicit-hie < 0.1.3" due to subtle cradle detection behavior change in implicit-hie, resulting in test failure, which was found in #3317

michaelpj:
If you're superseding https://github.com/haskell/haskell-language-server/pull/3400 then please close that one.

Happy to merge, but I do want a comment about the fourmolu issue.

michaelpj (approved):

cabal.project
```diff
@@ -65,7 +65,10 @@ constraints:
   -- https://github.com/ndmitchell/hlint/issues/1376
   hlint +ghc-lib,
   ghc-lib-parser-ex -auto,
-  stylish-haskell +ghc-lib
+  stylish-haskell +ghc-lib,
+  -- implicit-hie 0.1.3.0 introduced an unexpected behavioral change.
+  -- https://github.com/Avi-D-coder/implicit-hie/issues/50
+  implicit-hie < 0.1.3
```
fendor, line 71: See https://github.com/haskell/haskell-language-server/pull/3317#discussion_r1050818694, this applies here, too. @michaelpj what do you think? Fix in a follow-up PR? Must not be forgotten for #3317

michaelpj: Yeah, good point. I was thinking it was okay to be temporary, but indeed it can mess up our ability to release. Better to put a fake dependency in `ghcide` if we have to.

wavewave: added implicit-hie to ghcide.

wavewave:
hmm. tests on GHC 8.10.7 failed after putting implicit-hie to ghcide dependency. Is it due to some flakiness?

michaelpj:
Yeah, almost certainly. I've rerun the jobs.

wavewave:
oh thank you!

wavewave:
It was really due to the flaky tests. Now it's back to green. :-)

fendor (approved):
LGTM, thanks!

PR #3407, opened 2022-12-16
Document exception handling in Actions/Rules

michaelpj:
I also took the liberty of removing the `MonadFail` instance for `Rules`, since failing in `Rules` seems much worse as it won't automatically be caught. There was only one use of that, which now has an appropriately threatening comment.


hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -54,8 +54,13 @@ unwrapDynamic x = fromMaybe (error msg) $ fromDynamic x
 
 type TheRules = Map.HashMap TypeRep Dynamic
 
+-- | A computation that defines all the rules that form part of the computation graph.
+--
+-- 'Rules' has access to 'IO' through 'MonadIO'. Use of 'IO' is at your own risk: if
+-- you write 'Rules' that thow exceptions, then you need to make sure to handle them
```
fendor, line 60:
```suggestion
-- you write 'Rules' that throw exceptions, then you need to make sure to handle them
```

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -67,6 +72,12 @@ data SRules = SRules {
 ---------------------------------------------------------------------
 -- ACTIONS
 
+-- | An action representing something that can be run as part of a 'Rule'.
+-- 
+-- 'Action's can be pure functions but also have access to 'IO' via 'MonadIO' and 'MonadUnliftIO.
+-- It should be assumed that actions can throw exceptions, these can be caught with
```
fendor, line 78: Sorry for nitpicking, but who assumes these actions can throw exceptions? The user, or the Rules engine? Or both? E.g., are exceptions always caught or may they bubble up and eventually crash HLS?

fendor (approved):
LGTM, one comment

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
michaelpj: It's unclear! I'm writing this for the "user of `hls-graph`", which in this case is HLS itself, the point being that `hls-graph` does not do anything about the exceptions, but the user can. Now HLS _does_ use `actionCatch` to catch such exceptions and turn them into diagnostics, but it's a bit of a layering violation to write that here. Perhaps I should say it anyway?

Conceivably some other user of `hls-graph` could choose to just let them bubble up and crash the system...

fendor: Ah I see, then slightly adapting the beginning of the sentence seems fine by me.

```suggestion
-- It should be assumed that actions throw exceptions, these can be caught with
```

Maybe this is better?

pepeiborra (approved):
The best source of documentation for 'Action', 'Rules' and the exception semantics is Shake itself, which has pretty good docs.

michaelpj:
I think one relevant difference to Shake broadly assumes that it's the top-level process, and so e.g. failing during `Rules` should just lead to a quick user-visible failure. I think in our case that's much less clear - crashing HLS is bad and won't generally give a good error, so it's justifiable to make it harder to throw in `Rules`.

fendor (approved):

PR #3413, opened 2022-12-20
Various strictness improvements

wz1000:
Various strictness annotations that marginally improve heap usage.


ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -1455,18 +1455,6 @@ loadInterface session ms linkableNeeded RecompilationInfo{..} = do
 
     case (mb_checked_iface, recomp_iface_reqd) of
       (Just iface, UpToDate) -> do
-         -- If we have an old value, just return it
```
wz1000, line 1458: It turns out that reusing the same `ModDetails` as we had last time can keep the previous `HscEnv` and corresponding `HomePackageTable` data structures alive. Solution is to delete this code and always regenerate the `ModDetails` from the interface.

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -45,7 +46,7 @@ data GetInstanceBindTypeSigs = GetInstanceBindTypeSigs
 
 data InstanceBindTypeSig = InstanceBindTypeSig
     { bindName     :: Name
-    , bindRendered :: T.Text
+    , bindRendered :: !T.Text
```
fendor, line 49: Any reason this needs to be strict while the other values arent?

wz1000: It showed up in a profile, the others are likely to already be evaluated I think.

fendor: Then a comment would be nice.
Also, would a guideline be wise that generally says everything in records needs a bang, and only in certain exceptional cases you don't need them?

fendor (approved):
LGTM

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
michaelpj: Yeah, it seems clearer to bang everything in situations like this? Unsure.

ghcide/src/Development/IDE/Core/Compile.hs
mpickering: Yes!! Very important to do this (and cheap)

michaelpj: Do we need a cautionary note? should we somehow get rid of `old_value` entirely? or is it okay to use it so long as we don't retain anything from it?

wz1000: We need it for the old `ModIface` (which is safe to use).

PR #3414, opened 2022-12-20
Clear the mi_globals field when generating an iface

wz1000:
GHC populates the mi_global field in interfaces for GHCi if we are using the bytecode
interpreter.
However, this field is expensive in terms of heap usage, and we don't use it in HLS
anywhere. So we zero it out.
The field is not serialized or deserialised from disk, so we don't need to remove it
while reading an iface from disk, only if we just generated an iface in memory


fendor (approved):
LGTM!

michaelpj (approved):

PR #3415, opened 2022-12-20
GHC 9.4: Compute the source hash before the preprocessor

wz1000:
Compute the source hash before the preprocessor because that is
how GHC does it.

This reduces spurious rebuilds of modules using CPP on GHC 9.4


fendor (approved):
LGTM!

PR #3417, opened 2022-12-21
9.4 support for rename plugin

wz1000:


michaelpj:
Seems like this PR is making some substantive changes, worth comments?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -74,7 +78,10 @@ renameProvider state pluginId (RenameParams (TextDocumentIdentifier uri) pos _pr
            See the `IndirectPuns` test for an example. -}
         indirectOldNames <- concat . filter ((>1) . Prelude.length) <$>
             mapM (uncurry (getNamesAtPos state) . locToFilePos) directRefs
-        let oldNames = indirectOldNames ++ directOldNames
+        let oldNames = (filter matchesDirect indirectOldNames) ++ directOldNames
+            matchesDirect n = occNameFS (nameOccName n) `elem` directFS
+              where
+                directFS = map (occNameFS. nameOccName) directOldNames
```
OliverMadine, line 84: Why is this necessary? Is there a case that we were missing?

The only case I am aware of for indirect names is with record field puns, but these should always "match" the direct names since they are, well, puns.

Looks good otherwise - many thanks

wz1000: If there are implicit dictionary names (eg `$fShow` for code like `show 1`) then we don't want to attempt to rename those.

wz1000:
> Seems like this PR is making some substantive changes, worth comments?

I've added some comments.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
mpickering: You should mention that you need to filter stuff out since GHC-9.4 because now there are implicit dictionary names like `$fShow` which we don't want to attempt to rename.

mpickering:
I discussed this with Zubin and asked him to write a comment explaining what was going on with the `matchesDirect` filtering otherwise looks sensible to me.

mpickering (approved):

PR #3418, opened 2022-12-21
Add a mergify action to update PRs

michaelpj:
After a bit of digging I've worked out what's up with mergify. It's
still working fine, but it's not updating PRs very often now. The reason
for that is that it updates PRs that are in the queue, but to get
_into_ the queue you have to pass CI, which is hard ATM because of our
flaky CI. So we can add a separate update rule that just updates the
labelled PRs, so that they are at least kept up-to-date with master.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -13,3 +13,15 @@ pull_request_rules:
       queue:
         method: squash
         name: default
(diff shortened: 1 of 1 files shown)


fendor (approved):
LGTM, thank you!

PR #3420, opened 2022-12-21
Enable plugin tests in CI for GHC 9.4

wz1000:
We neglected to enable testing in CI of all the newly patched plugins for GHC 9.4


pepeiborra (approved):
Thanks for noticing this!

michaelpj:
Oof, good catch!

mpickering:
Good work Zubin.

konn (approved):
Good catch, thanks! 👍

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -1669,8 +1669,10 @@ suggestImportTests = testGroup "suggest import actions"
     , test True []          "f = empty"                   []                "import Control.Applicative (empty)"
     , test True []          "f = empty"                   []                "import Control.Applicative"
     , test True []          "f = (&)"                     []                "import Data.Function ((&))"
-    , test True []          "f = NE.nonEmpty"             []                "import qualified Data.List.NonEmpty as NE"
-    , test True []          "f = Data.List.NonEmpty.nonEmpty" []            "import qualified Data.List.NonEmpty"
+    , ignoreForGHC94 "On GHC 9.4 the error message doesn't contain the qualified module name"
+      $ test True []          "f = NE.nonEmpty"             []                "import qualified Data.List.NonEmpty as NE"
+    , ignoreForGHC94 "On GHC 9.4 the error message doesn't contain the qualified module name"
+      $ test True []          "f = Data.List.NonEmpty.nonEmpty" []            "import qualified Data.List.NonEmpty"
```
fendor, line 1675: This seems like rather important regressions, we ought to create an issue for that

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -1679,14 +1681,17 @@ suggestImportTests = testGroup "suggest import actions"
     , test True []          "f = (.|.)"                   []                "import Data.Bits (Bits((.|.)))"
     , test True []          "f = (.|.)"                   []                "import Data.Bits ((.|.))"
     , test True []          "f :: a ~~ b"                 []                "import Data.Type.Equality ((~~))"
-    , test True
+    , ignoreForGHC94 "On GHC 9.4 the error message doesn't contain the qualified module name"
+      $ test True
```
fendor, line 1685: Same here and all the subsequent tests

fendor (approved):
LGTM

plugins/hls-refactor-plugin/test/Main.hs
wz1000: I opened https://gitlab.haskell.org/ghc/ghc/-/issues/22130 a while ago but I didn't realise that it also affected import suggestions. Unfortunately it seems difficult to fix this in HLS and we have to wait for a fix in GHC.

wz1000: I've updated the message to point to this ticket.

PR #3424, opened 2022-12-27
Add CI for 9.4.4

wz1000:
Also drop 9.4.3, 9.4.2 and 9.2.3

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,11 +60,9 @@ jobs:
       # **don't forget**
       # to update the ghc versions in 'caching.yml'.
       matrix:
-        ghc: [ "9.4.3"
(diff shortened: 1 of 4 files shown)


pepeiborra (approved):

.github/workflows/test.yml
```diff
@@ -98,7 +96,7 @@ jobs:
              ghc: '8.10.7'
              test: true
            - os: windows-latest
-             ghc: '9.4.2'
+             ghc: '9.4.4'
```
fendor, line 99: This won't work, unless https://github.com/haskell/actions/issues/129 has been resolved.

michaelpj:
Looks like the upstream issue got fixed, might work now if we bump the setup action version.

PR #3426, opened 2022-12-28
ekg-json seems unused

Bodigrim:


Bodigrim:
Ah, it's a dependency of `ghcide`, not `haskell-language-server`...

PR #3428, opened 2022-12-29
Update ghc-version-support.md

noughtmare:
I've updated these according to the release notes of 1.9.0.0. Although I don't know about the full vs basic support status.


michaelpj:
Thanks! I think we're close to full support but not quite there yet.

michaelpj (approved):

PR #3431, opened 2022-12-30
various plugins: Add testdata to sdist

maralorn:
Missing testdata breaks tests when running them from the hackage sdist. This affects nixpkgs because we run tests by default.


pepeiborra (approved):
Thanks!

fendor (approved):
LGTM, thank you :see_no_evil:

PR #3434, opened 2023-01-02
Drop support for GHC 8.10

michaelpj:
Per our support policy, we should now deprecated 8.10. We might not want
to do this immediately, but this PR implements it.

TODO:
- Make sure CI is still all happy
- Simplify a lot of bindings that can now just be re-exports


hasufell:
I'm against it.

The 2022 Haskell survey results show that almost half of users are still on 8.10.x: https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4

It seems quite radical to drop it.

fendor:
Assuming 1.9.0.0 is fairly stable, GHC 8.10.7 users may use that for the foreseeable future, or are we expecting even more groundbreaking performance improvements for HLS and GHC 8.10.7?

michaelpj:
I misremembered. Our policy says that we support three major GHC versions back from a version that has full support from HLS and is in a stackage LTS. The latest stackage LTS is _9.2_ not 9.4. So 8.10 is still in our support window until that happens.

I'll leave this open for now since I did most of the work already.

hasufell:
> Assuming 1.9.0.0 is fairly stable, GHC 8.10.7 users may use that for the foreseeable future, or are we expecting even more groundbreaking performance improvements for HLS and GHC 8.10.7?

The policy is a policy, not a rule. And it doesn't make sense in this case unless we want to annoy half of our userbase.

michaelpj:
The policy is also a _justification_. Dropping versions will always annoy people, part of the reason for having a policy is so there's something we can point at and say, "look, we have to do something, here's this reasonable policy that says why we're doing it, do you disagree with that?".

hasufell:
Sure, I'll send a PR to update the policy then.

michaelpj:
No, the policy is fine, I misremembered and this PR is premature, see my earlier comment. We should not drop 8.10 according to the policy.

hasufell:
> No, the policy is fine

As one of the initiators of the policy, I disagree.

Kleidukos:
Please for the love of God I urge you not to deprecate 8.10, its latest minor (8.10.7) is one of the most used GHC version, it has wonderful support in the ecosystem while still being quite reliable in terms of code generation.

michaelpj:
Just to reiterate:
1. The current policy is _not_ to drop it, we're not doing it until there's a stackage LTS with 9.4 at least.
2. Dropping support for 8.10 just means that 8.10 users won't get _new_ HLS updates. The existing versions will work fine forever.

Kleidukos:
> Dropping support for 8.10 just means that 8.10 users won't get new HLS updates. The existing versions will work fine forever.

Please correct me if I'm wrong but does that mean that there won't be patch releases for the last HLS to support 8.10.7?

hasufell:
> Please correct me if I'm wrong but does that mean that there won't be patch releases for the last HLS to support 8.10.7?

Yeah I think there never really have been long-running branches and my intuition says that would be more work than keeping that CPP code in `master`.

andys8:
This is maybe a bit out of scope, but for context: There might be some large libraries that aren't ready / released for ghc 9.x. If I'm not mistaken `amazonka` is [one of them](https://github.com/brendanhay/amazonka/discussions/716). It might be possible to work around it this issue by using unreleased code, but there is no official release yet. As shown in the survey, there might be a good amount of production projects that will keep using ghc 8.10.7 for now.

michaelpj:
I want to revisit this. I think it is a good time. Stackage LTS is on 9.4. 8.10.7 was released in 2021. The compiler has not received any bug fixes since then, I think it is not so bad that 8.10-compatible HLS will now no longer receive bug fixes. Other than that the existing releases will continue to work perfectly well.

michaelpj:
We're going to do a release next week, though, this can wait until after that.

michaelpj:
Okay, it's green now. I can't make the pre-commit check green because stylish-haskell chokes on a file and I can't figure out why. But I've updated the formatting as much as possible.

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs
joyfulmantis, line 1: For line 84: stylish-haskell doesn't like function body and arguments to be separated like that. I think you might be able to move just the "=" over to the function body section and get it to work. Otherwise, you'll need to duplicate the function twice.

fendor (approved):
LGTM, thanks!

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs
michaelpj: Thanks, I'll fix this separately so I can get this in for now.

juhp:
Can haskell-language-server-wrapper handle different versions of HLS?  (My guess would be no?)

fendor:
It only knows about the HLS versions on PATH, iirc, so kind of?

PR #3438, opened 2023-01-04
Refine GHC deprecation policy

hasufell:
https://github.com/haskell/haskell-language-server/pull/3434

### DIFF
--- a/docs/support/ghc-version-support.md
+++ b/docs/support/ghc-version-support.md
@@ -83,10 +83,13 @@ This is the static part of the policy that can be checked by a machine.
 
 A major GHC version is a "legacy" version if it is 3 or more major versions behind the latest GHC version that is
 
-1. Fully supported by HLS
-2. Used in a Stackage LTS snapshot
+1. Fully supported by HLS (support status "full support")
+2. Used in a Stackage LTS snapshot or the GHCup recommended version, whichever is lower.
+    * `min(ghcup recommended version, stackage lts snapshot)`
 
 For example, if 9.2 is the latest major version fully supported by HLS and used in a Stackage LTS, then the 8.8 major version and older will be legacy.
+
+Another example, if there is a Stackage LTS for GHC 9.6 and recommended version
 
 HLS will support all non-legacy major versions of GHC.
 
@@ -107,6 +110,7 @@ To establish and apply the policy we take the following ecosystem factors into a
 
 - Support status of HLS
 - The most recent [stackage](https://www.stackage.org/) LTS snapshot
+- The GHC version recommended by GHCup
 - The GHC versions used in the most popular [linux distributions](https://repology.org/project/ghc/versions)
 - The reliability of different ghc versions on the major operating systems (Linux, Windows, MacOS)
 - The [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)
--- a/docs/support/ghc-version-support.md
+++ b/docs/support/ghc-version-support.md
@@ -81,15 +81,17 @@ This is the static part of the policy that can be checked by a machine.
 
 #### Major versions
 
-A major GHC version is a "legacy" version if it is a major versions behind the latest GHC version that is
-used in a Stackage LTS snapshot or the GHCup recommended version, whichever is lower.
-* `min(ghcup recommended version, stackage lts snapshot)`
+HLS will support major versions of GHC until they are older than _both_
 
-For example, if 9.2 is the latest major version fully supported by HLS and used in a Stackage LTS, then the 9.1 major version and older will be legacy.
+1. The major version of GHC used in the current Stackage LTS; and
+2. The major version of GHC recommended by GHCup
 
-Another example, if there is a Stackage LTS for GHC 9.6 and recommended version
+For example, if 
 
-HLS will support all non-legacy major versions of GHC.
+1. Stackage LTS uses GHC 9.2; and
+2. GHCUp recommends GHC 9.4
+
+then HLS will support back to GHC 9.2.
 
 #### Minor versions
 
@@ -129,10 +131,6 @@ When we decide to keep on an old version, we should track it as follows:
     - e.g. if Haskell Survey results show that 25% or more of users are still on the GHC version in question, then dropping should be avoided
 3. if dropping is still undesired, but maintenance burden is also high, then set out a call-for-help and contact HF for additional funding to support this GHC version
 4. if no help or funding was received within 2 releases (say, e.g. 3-6 months), then drop the version regardless
-
-### Announcements
-
-We will warn users about the upcoming deprecation of a GHC version in the notes of the release *prior* to the deprecation itself.
 
 ### Why deprecate older versions of GHC?
 
--- a/docs/support/ghc-version-support.md
+++ b/docs/support/ghc-version-support.md
@@ -111,7 +111,7 @@ To establish and apply the policy we also take into account:
 - The reliability of different ghc versions on the major operating systems (Linux, Windows, MacOS)
 - The [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)
 
-### Supporting a GHC version beyond its shelf life
+### Supporting a GHC version beyond normal deprecation time
 
 In cases where the base policy demands a deprecation, but ecosystem factors
 suggest that it's still widely used (e.g. last [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)), the deprecation should be suspended for the next release and the situation be re-evaluated for the release after that.
--- a/docs/support/ghc-version-support.md
+++ b/docs/support/ghc-version-support.md
@@ -84,7 +84,7 @@ This is the static part of the policy that can be checked by a machine.
 A major GHC version is a "legacy" version if it is 3 or more major versions behind the latest GHC version that is
 
 1. Fully supported by HLS
-2. Used in the a Stackage LTS
+2. Used in a Stackage LTS snapshot
 
 For example, if 9.2 is the latest major version fully supported by HLS and used in a Stackage LTS, then the 8.8 major version and older will be legacy.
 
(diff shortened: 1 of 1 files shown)


hasufell:
The existing policy already demands to consider ecosystem factors, so adding the Haskell Survey as an input seems like a no-brainer (because it's one of the largest indicators for ecosystem questions).

Additionally, it seems it wasn't very clear that there are two different parts of the policy: one that can be statically checked and one that can't. And that both parts need to be considered. As such, I tried to make that clearer with sectioning them.

fendor (approved):
This addition to the deprecation policy seems sensible to me.

hasufell:
@michaelpj any opinions?

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
 HLS will support all non-legacy major versions of GHC.
 
-### Minor versions
+#### Minor versions
 
 For the latest supported major GHC version we will support at least 2 minor versions.
 
 For the rest of the supported major GHC versions, we will support at least the latest minor version in Stackage LTS (so 1 minor version).
 
+### Extended policy
```
michaelpj, line 89: I think the base/extended distinction is confusing and you could just drop these headers. It's fairly clear that the later sections talk about exceptions to the rules.

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+
+This is the part of the policy that needs evaluation by a human and possibly followed
+by a discussion.
+
+#### Ecosystem factors
+
+To establish and apply the policy we also take into account:
+
+- Completeness: support includes all plugins and features
```
michaelpj, line 98: This is already included above

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+- The most recent [stackage](https://www.stackage.org/) LTS snapshot
```
michaelpj, line 99: ditto

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+- The GHC versions used in the most popular [linux distributions](https://repology.org/project/ghc/versions)
+- The reliability of different ghc versions on the major operating systems (Linux, Windows, MacOS)
+- The [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)
```
michaelpj, line 102: I think these three points are the interesting ones. I'd move them into the next section where they're used.

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+
+### Supporting a GHC version beyond its shelf life
+
+In cases where the base policy demands a deprecation, but ecosystem factors
+suggest that it's still widely used (e.g. last [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)), the deprecation should be suspended for the next release and the situation be re-evaluated for the release after that.
```
michaelpj, line 107:
```suggestion
Sometimes the policy suggests that we should deprecate a version of GHC, but doing so would be particularly harmful for our users because it remains very widely used. In cases like this we can decide to keep supporting that GHC version, but this decision should be revisited each time we do a release.

To determine whether we should keep supporting a GHC version we can use 
- The GHC versions used in the most popular [linux distributions](https://repology.org/project/ghc/versions)
- The [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)
- Anecdotal evidence about he reliability of different GHC versions on the major operating systems (Linux, Windows, MacOS), e.g. noting the presence of serious bugs
```

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+
+To establish and apply the policy we also take into account:
+
+- Completeness: support includes all plugins and features
+- The most recent [stackage](https://www.stackage.org/) LTS snapshot
+- The GHC versions used in the most popular [linux distributions](https://repology.org/project/ghc/versions)
+- The reliability of different ghc versions on the major operating systems (Linux, Windows, MacOS)
+- The [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)
+
+### Supporting a GHC version beyond its shelf life
```
michaelpj, line 104:
```suggestion
### Supporting a GHC version beyond normal deprecation time
```

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+
+In cases where the base policy demands a deprecation, but ecosystem factors
+suggest that it's still widely used (e.g. last [Haskell Survey results](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s2q4)), the deprecation should be suspended for the next release and the situation be re-evaluated for the release after that.
+
+In case of maintenance burden concerns, the following process should be followed:
```
michaelpj, line 109:
```suggestion
When we decide to keep on an old version, we should track it as follows:
```

docs/support/ghc-version-support.md
```diff
@@ -76,12 +80,43 @@ For example, if 9.2 is the latest major version fully supported by HLS and used
...
+
+1. open a ticket on HLS issue tracker wrt discussing to deprecate said GHC version
+    - explain the reason the GHC version wasn't deprecated (context)
+    - explain the maintenance burden it causes (reason)
+    - evaluate whether it impacts the next HLS release (impact)
+2. discuss whether ecosystem factors changed
+    - e.g. if Haskell Survey results show that 25% or more of users are still on the GHC version in question, then dropping should be avoided
+3. if dropping is still undesired, but maintenance burden is also high, then set out a call-for-help and contact HF for additional funding to support this GHC version
+4. if no help or funding was received within 2 releases (say, e.g. 3-6 months), then drop the version regardless
```
michaelpj, line 118: I'm fine with this but we could also leave it out and just experiment with various ways of asking for help before we write it down

michaelpj (approved):
Fine, just some prose suggestions.

docs/support/ghc-version-support.md
hasufell: True

docs/support/ghc-version-support.md
hasufell: Hm, I think we're having a slightly different perception. 

I'm not seeing it as an exception rule that can be applied if someone speaks up about it.

Someone actually updating the supported GHC versions is **required** to check/consided both.

michaelpj: Okay, yes, I agree with that. Then maybe we just want to add after line 81: "HLS may continue to support legacy GHC versions, see the section below on 'Supporting legacy GHC versions'" or something?

docs/support/ghc-version-support.md
hasufell: I don't understand this comment

hasufell: Included where?

Do you mean 

> Support status (see the support policy below for more details):
> - "initial support": support for this GHC version is underway, but it is not ready to be released yet
> - "basic support": this GHC version is currently actively supported, and all [tier 1 plugins](./plugin-support.md) work
> - "full support": this GHC version is currently actively supported, and most [tier 2 plugins](./plugin-support.md) work
> - "deprecated": this GHC version was supported in the past, but is now deprecated

?

docs/support/ghc-version-support.md
hasufell: I think this is mixing up things. Ecosystem factors should be its own section and be referenced from this section explaining the **process** (not criteria) of supporting an older GHC version.

hasufell:
@michaelpj I think I'm not quite getting what you're proposing. I feel the sections are fine. The wording can definitely improved, but it's a running document.

fendor:
@michaelpj is there still some open discussion here?

fendor:
@michaelpj I think the wording and policy is fine and we should merge this.

Would you mind reading through one more time?

docs/support/ghc-version-support.md
fendor: I agree and think this separation helps the reader understanding what is going on.

docs/support/ghc-version-support.md
fendor: I rephrased this to make sure it is clear we are referring to the HLS support status.

Also, I made clear that we aim to list *all* factors.

docs/support/ghc-version-support.md
```diff
@@ -75,23 +75,59 @@ Major versions of GHC which are not supported by HLS on master are extremely unl
...
 ## GHC version deprecation policy
 
-### Major versions
+### Base policy
+
+This is the static part of the policy that can be checked by a machine.
+
+#### Major versions
 
 A major GHC version is a "legacy" version if it is 3 or more major versions behind the latest GHC version that is
```
soulomoon, line 84: are we like switching to the “oldest popular version” as min(stackage_lts, ghcup_recommended). As in https://discourse.haskell.org/t/call-for-opinions-shortening-the-hls-support-window/9641/16?u=soulomoon

fendor: Yes, we are! Didn't have time to add this section yet :)

fendor: Updated!

soulomoon (commented):
I reword it a bit more. But I think we should still wait for @michaelpj to confirm

michaelpj:
I reworded it some more. I also removed the section about announcements, since we haven't been doing that, and I think it seems in practice to not be that big a deal.

michaelpj (approved):

Issue #3425, opened 2022-12-28
hls-fourmolu-plugin-1.1.1.0 is incompatible with ghcide-1.8.0.0


lehins:
Building `hls-fourmolu-plugin-1.1.1.0` results in a build error, when it is built with `haskell-language-server-1.8.0.0`:

```
Preprocessing library for hls-fourmolu-plugin-1.1.1.0..
Building library for hls-fourmolu-plugin-1.1.1.0..
[1 of 2] Compiling Ide.Plugin.Fourmolu.Shim ( src/Ide/Plugin/Fourmolu/Shim.hs, dist/build/Ide/Plugin/Fourmolu/Shim.o, dist/build/Ide/Plugin/Fourmolu/Shim.dyn_o )
[2 of 2] Compiling Ide.Plugin.Fourmolu ( src/Ide/Plugin/Fourmolu.hs, dist/build/Ide/Plugin/Fourmolu.o, dist/build/Ide/Plugin/Fourmolu.dyn_o )

src/Ide/Plugin/Fourmolu.hs:62:56: error:
    Variable not in scope:
      usePropertyAction
        :: t0
           -> PluginId
           -> Properties '[ 'PropertyKey "external" 'TBoolean]
           -> Action Bool
   |
62 |     useCLI <- liftIO $ runAction "Fourmolu" ideState $ usePropertyAction #external plId properties
   |                                                        ^^^^^^^^^^^^^^^^^
````

Which makes sense, since `usePropertyAction` was not exported until `1.8.0.2`, which is not on hackage:
https://github.com/haskell/haskell-language-server/blob/ddb21a0c8d4e657c4b81ce250239bccf28fc9524/ghcide/src/Development/IDE.hs#L19-L20

I'd recommend creating a revision on hackage that fixes the lower bound: `ghcide >= 1.9`

georgefst:
Good catch.

> I'd recommend creating a revision on hackage that fixes the lower bound: `ghcide >= 1.9`

Any reason not to make it `ghcide >= 1.8.0.2`? It's more correct, and a user could always supply 1.8.0.2 locally.

michaelpj:
Fixed

PR #3443, opened 2023-01-07
Correct Fourmolu plugin's lower bound on Ghcide

georgefst:
See #3425.

I'll make a Hackage revision once this is merged.


michaelpj (approved):
Thanks!

PR #3444, opened 2023-01-07
Add Inline code action to Retrie plugin

pepeiborra:
This adds a new code action to hls-retrie-plugin that allows to inline a definition "on the spot". There are some caveats:

1. The code action only appears on identifiers for which we can locate the source code
2. Inlining only works up to retrie capabilities. Currently retrie cannot inline functions that use pattern matching, for instance, and there are other bugs.

Another addition is a test suite for the plugin. Since hls-retrie-plugin depends on hls-refactor-plugin, the test suite needs to tell `goldenWithHaskellDoc` to run both plugins. I had to refactor a few things in order to make this possible, hence the large changeset. 

https://user-images.githubusercontent.com/26626/211166901-a6af7c51-f3b0-4d5e-9488-7ea90ff312c5.mov


pepeiborra:
This is ready for review: a look of @fendor at the test suite changes should be all is needed

fendor (approved):
Changes to the test-suite look good to me!

PR #3446, opened 2023-01-09
Support stylish-haskell and ormolu in 9.4

michaelpj:
Recent releases of both of these make them 9.4-compatible.

I think this qualifies 9.4 to be fully-supported by HLS, also.


michaelpj:
Apparently floskell works too!

Ailrun (approved):
Awesome, thank you for the update!

July541:
Just for a reminder, hls-evel-plugin still doesn't support ghc-9.4.

michaelpj:
Yeah, we just need https://github.com/haskell/haskell-language-server/pull/3391

Issue #411, opened 2020-09-18
Coherence of command-line formatters versus formatter plugins


michaelpj:
As many people do, I have a gate in my CI that ensures that code passes a formatter before it is merged. That formatter (in this case `stylish-haskell`) is run as an executable.

If I also use HLS, I can format my code with `stylish-haskell` via the formatter plugin. But this will use the version of `stylish-haskell` that HLS was compiled with, not the one that I have (elsewhere) specified for use in CI.

This makes the in-editor formatting useless to me, as I will typically have to reformat with the executable version before my change will pass CI.

I can see two ways to keep things consistent.

1. HLS follows environment

Ultimately I think this would mean that HLS would have to call the executable. I understand that this is undesirable since the current model allows us to use the nicer Haskell interfaces to tools and test them easily.

2. Environment follows HLS

Perhaps we could have some way to format "as-if" using one of the HLS formatting plugins. For example:
- Ship the formatter executables with HLS, so users can call those directly.
- Expose the formatting capabilites through HLS itself as part of the CLI, e.g. `haskell-language-server --format --formatter ormolu Foo.hs`.

michaelpj:
Oh, another irritation of HLS directly picking a version of the formatter is that then the "canonical" formatting of my codebase changes with the version of HLS. I can't pin `stylish-haskell` independently of HLS, as I would be able to if it called it as an external command.

anka-213:
I wonder if it would be possible to take this one step further and make more functionality available via the command line. For example, if I want to apply some fix everywhere in my code, it is often more convenient to automate from the command line than to try to coerce a GUI into doing it (or doing it everywhere manually). But maybe this is a completely separate issue.

alanz:
This could be ameliorated with the coming plugin architecture, which is specifically aimed at allowing a given user to configure exactly what plugins they want to use.

In time this should lead to a formatter publishing the appropriate plugin descriptor for use in HLS, then this becomes a simple matter of using the same version in both places.

michaelpj:
How would the plugin API help here? Would the formatter provider have to provide N different plugins for N different versions? Or would they be loaded dynamically somehow so the user can choose which one? It seems difficult to make this work in the model where HLS statically depends on a particular formatter (at a particular version).

alanz:
[...]

PR #3449, opened 2023-01-09
Make the Ormolu plugin respect `.ormolu` fixity files when Ormolu ≥0.5.3.0

### LINKED ISSUE #3448: `.ormolu` files are ignored when formatting with Ormolu
### Your environment

**Which OS do you use?**
NixOS
**Which version of GHC do you use and how did you install it?**
9.0.2 from Nix (branch `nixos-22.11`).
**How is your project built (alternative: link to the project)?**
With Cabal (the project is private).

**Which LSP client (editor/plugin) do you use?**
emacs+lsp-mode
**Which version of HLS do you use and how did you install it?**
`1.8.0.0` from Nix (branch `nixos-22.11`).
**Have you configured HLS in any way (especially: a `hie.yaml` file)?**
Yes, with `hie.yaml`. I think the configuration is correct because everything seems to work (type info, lenses, code actions, completion, go to definition, find references, etc.).

### Steps to reproduce

1. Create a file `.ormolu` [as per Ormolu's documentation](https://github.com/tweag/ormolu#language-extensions-dependencies-and-fixities) containing:
```
infixl 7 .=?

```
2. Have a Haskell file containing:
```hs
(.=?) :: String -> String -> String
s1 .=? s2 = s1 <> s2

infixl 7 .=?

foo :: String
foo = "a" .=? "b"
    <> "c" .=? "d" <> "e" .=? "f"

```
3. Format from the command line:
```
$ ormolu path/to/my/file.hs
```
The format is as expected:
```hs
foo :: String
foo =
  "a" .=? "b"
    <> "c" .=? "d"
    <> "e" .=? "f"
```
4. Format from the editor (e.g. in Emacs: `M-x lsp-format-buffer`), the format is different (does not seem to take into account `.ormolu` file):
```hs
foo :: String
foo =
  "a"
    .=? "b"
    <> "c"
    .=? "d"
    <> "e"
    .=? "f"

```

### Expected behaviour

Formatting behaves exactly the same from the CLI and from the LSP.

### Actual behaviour

LSP behavior is different.

My assumption is that LSP does not format the file, but formats using `stdin`. In that case, as per Ormolu documentation, the LSP should pass `--stdin-input-file <file name>` option to Ormolu to help it detect the `.ormolu` file.

This is obviously just a guess :sweat_smile: Maybe I'm completely wrong.

amesgen:
Closes #3448; includes a regression test


georgefst (approved):
Looks good.

As for the reasons this is currently a draft, is there any particular conceptual reason why it would be difficult to better expose the functionality from upstream? Or is it just something that requires some boring grunt work?

amesgen:
> As for the reasons this is currently a draft, is there any particular conceptual reason why it would be difficult to better expose the functionality from upstream? Or is it just something that requires some boring grunt work?

It is just an API design question, e.g. involving how "coarsely-grained" mirroring the behavior of the executable via the Haskell API should be (see https://github.com/tweag/ormolu/pull/969 and in particular https://github.com/tweag/ormolu/pull/970#pullrequestreview-1252364825).

amesgen:
Relevant upstream PR: https://github.com/tweag/ormolu/pull/976

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
```diff
@@ -44,13 +47,26 @@ provider ideState typ contents fp _ = withIndefiniteProgress title Cancellable $
...
-    fmt cont conf =
+    fmt :: T.Text -> Config RegionIndices -> IO (Either SomeException T.Text)
+    fmt cont conf = flip catches handlers $ do
+      let fp' = fromNormalizedFilePath fp
 #if MIN_VERSION_ormolu(0,5,3)
-      try @OrmoluException $ ormolu conf (fromNormalizedFilePath fp) cont
+      cabalInfo <- getCabalInfoForSourceFile fp' <&> \case
+        CabalNotFound                -> Nothing
+        CabalDidNotMention cabalInfo -> Just cabalInfo
+        CabalFound cabalInfo         -> Just cabalInfo
```
fendor, line 57: Since this function is not using the information from hie-bios, we are destined to have unexpected behaviour with this at some point in the future when ghcide and ormolu don't agree on the .cabal file. This seems to be loading information from .cabal files, without knowing which .cabal file is the correct one?
I am uncertain whether this is really better, but we could ask for the cradle options for this file in particular and find the "more correct" .cabal file if it is a stack or cabal project. Do you know what the cabal file used for?

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
```diff
@@ -59,7 +75,7 @@ provider ideState typ contents fp _ = withIndefiniteProgress title Cancellable $
  where
    title = T.pack $ "Formatting " <> takeFileName (fromNormalizedFilePath fp)
 
-   ret :: Either OrmoluException T.Text -> Either ResponseError (List TextEdit)
+   ret :: Either SomeException T.Text -> Either ResponseError (List TextEdit)
```
fendor, line 78: Why `SomeException` now? Doesn't this risk of swallowing async exceptions? IIRC, ghcide relies heavily on being able to interrupt actions, unless there is a good reason, we shouldn't catch `SomeException` since code will change and then eventually not do what we expect.

fendor (changes requested):
Thank you, awesome!
Just want to discuss the `SomeException` change, otherwise good to go!

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
amesgen: > Why `SomeException` now? Doesn't this risk of swallowing async exceptions?

It does not, as we are only catching `OrmoluException` and `IOException`:
https://github.com/haskell/haskell-language-server/blob/e9f9a49014276eb2dceab0cb91faccf4d717f5e8/plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs#L66-L69

Previously, only `OrmoluException` was caught, but `getCabalInfoForSourceFile`/`getCabalInfoForSourceFile`, which are added in this PR, can alos throw `IOException`.

fendor: Ah, my mistake. This is unfortunate, wish we had some union type here :(

fendor (approved):

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
amesgen: The idea is that the Ormolu invoked via HLS should have equal behavior with the Ormolu executable (this currently being not the case is the point of #3448). For how the information from the `.cabal` file is used by Ormolu, see this section in its README: https://github.com/tweag/ormolu#language-extensions-dependencies-and-fixities

More generally, this touches #411; i.e. in the distant future, it could be the case that all formatters are invoked via HLS, then using hie-bios etc.

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
amesgen: Yeah, I can change it to `Either (Either OrmoluException IOException) T.Text` if you prefer, but it indeed seems more clumsy :crying_cat_face:

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
fendor: You are right, this is more about #411

plugins/hls-ormolu-plugin/src/Ide/Plugin/Ormolu.hs
fendor: I don't think I do :D

Issue #3450, opened 2023-01-12
Recompilation avoidance in GHC 9.4 conflicts with recompilation avoidance in HLS


wz1000:
We added very fine grained recompilation avoidance to HLS in #2316. With that change, a module using TH is recompiled if the hash of the Core of any of the modules it uses in a splice changes.

However, we still use GHC's recompilation avoidance by way of the `checkOldIface` function, which handles all the other recompilation avoidance cases which don't involve Template Haskell.

GHC 9.4 also added a similar finer grained recompilation avoidance scheme, in that it will recompile a module using TH if the hash of the **source** of any modules it uses in a splice changes. This is weaker than what we have in HLS, and as such we may end up recompiling when we didn't need to before.

Possible solution: Filter out the `UsageHomeModuleInterface` entries added to usages by GHCs recompilation checker.

michaelpj:
Should we be trying to improve GHC's recompilation avoidance so it's as good as ours and we can just rely on it?

mouse07410:
> Should we be trying to improve GHC's recompilation avoidance so it's as good as ours and we can just rely on it?

Yes - because right now HLS is almost unusable with GHC-9.4.4 and VS Code: whenever I try to apply a Quick Fix hint, the hint is **not** applied and HLS errors out on this very issue.

michaelpj:
I think you're commenting on the wrong issue. This has nothing to do with code actions.

mouse07410:
> I think you're commenting on the wrong issue. This has nothing to do with code actions.

Possible - but that's where  viscose plugin threw me upon hitting that problem. I _think_ the problem was not code actions per se, but the internal error triggered by attempt to apply the action, presumably because it got entangled with "whether to recompile" thing. But I don't know. In fact, if you're sure this is barking up the wrong tree - please feel free to delete my (irrelevant) comments in this issue.

fendor:
Fixed by #3452 afaict

PR #3452, opened 2023-01-13
Two recompilation avoidance related bug fixes

wz1000:
1. Recompilation avoidance regresses in GHC 9.4 due to interactions between GHC and HLS's implementations.
   Avoid this by filtering out the information that causes the conflict
   See #3450.

2. The recompilation avoidance info GHC stores in interfaces can blow up to be
   extremely large when deserialised from disk. See https://gitlab.haskell.org/ghc/ghc/-/issues/22744
   Deduplicate these filepaths.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -70,7 +70,7 @@ import           Data.Time                         (UTCTime (..))
 import           Data.Tuple.Extra                  (dupe)
 import           Data.Unique                       as Unique
 import           Debug.Trace
-import           Development.IDE.Core.FileStore    (resetInterfaceStore)
+import           Development.IDE.Core.FileStore    (resetInterfaceStore, shareFilePath)
 import           Development.IDE.Core.Preprocessor
 import           Development.IDE.Core.RuleTypes
 import           Development.IDE.Core.Shake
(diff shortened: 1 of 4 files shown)


ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -435,6 +436,42 @@ tcRnModule hsc_env tc_helpers pmod = do
...
+#endif
+
+-- | Mitigation for https://gitlab.haskell.org/ghc/ghc/-/issues/22744
+shareUsages :: ModIface -> ModIface
+shareUsages iface = iface {mi_usages = usages}
+  where usages = map go (mi_usages iface)
+        go usg = usg {usg_file_path = fp}
+          where !fp = shareFilePath (usg_file_path usg)
+
+filePathMap :: IORef (HashMap.HashMap FilePath FilePath)
```
fendor, line 462: Correct me if I am wrong, but I thought we want to prefer `Var` over `IORef` because it has better guarantees? E.g., it was easier for us to avoid space leaks?

fendor (commented):
Can't really judge the change itself, but one comment

ghcide/src/Development/IDE/Core/Compile.hs
wz1000: I don't think it really matters here, except that `Var` generally has slightly more overhead.

michaelpj:
Do we definitely want to include the mitigation for issue 2? would it be better to just fix in GHC and let it be in the next minor 9.4 release?

wz1000:
> Do we definitely want to include the mitigation for issue 2? would it be better to just fix in GHC and let it be in the next minor 9.4 release?

It is unlikely to be fixed in a minor GHC release because the simplest way to perform the fix involves an API change.

michaelpj:
That's a shame :(

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -435,6 +436,42 @@ tcRnModule hsc_env tc_helpers pmod = do
...
+filePathMap = unsafePerformIO $ newIORef HashMap.empty
+{-# NOINLINE filePathMap #-}
+
+shareFilePath :: FilePath -> FilePath
```
michaelpj, line 466: This and `shareUsages` are actually called in `IO`, so we could just have them be in `IO` and have slightly less unsafety.

michaelpj (approved):

ghcide/src/Development/IDE/Core/Compile.hs
pepeiborra, line 466: `shareFilePath` doesn't really belong here. Maybe move it to `D.I.C.FileStore` to make it easier to find?

pepeiborra (approved):

guibou:
I've tried this MR and I can confirm that HLS is way faster and seems to not rebuild multiple files at each changes.

However, some files are failing with:

```
No match in record selector usg_file_path
```

Because `usg_file_path` appears in this MR, looks like it is related.

wz1000:
> However, some files are failing with:
> 
> ```
> No match in record selector usg_file_path
> ```
> 
> Because `usg_file_path` appears in this MR, looks like it is related.
> 
> And any change in any file which appears in my workspace error log with such error does not seem to be taken into account anymore in files which are depending on them.

Thanks, I've fixed this.

guibou:
> Thanks, I've fixed this.

Thank you. I've tested and it is fast like it had never been before AND it does not fail like described above.

I'm pushing the fix in production tomorrow, I'll tell you if developers are happy or grumpy ;)

ghcide/src/Development/IDE/Core/Compile.hs
wz1000: I moved `shareFilePath` to a different module, and I think it makes sense for it to be pure so that it can be used in other places in the future.

Issue #3421, opened 2022-12-22
Calling Fourmolu via HLS infers wrong fixities


RobinStoll:
When using Fourmolu as a formatter via HLS, it infers different fixities of binary operators than when calling Fourmolu directly via the command line. This happens regardless of whether HLS uses its built-in Fourmolu or calls an external one.

It happened with the following operators:
- `^.` from [`lens`](https://hackage.haskell.org/package/lens-5.2/docs/Control-Lens-Getter.html#v:-94-.)
- `##` from [`map-syntax`](https://hackage.haskell.org/package/map-syntax-0.3/docs/Data-Map-Syntax.html#v:-35--35-)
- `.:` from [`digestive-functors`](https://hackage.haskell.org/package/digestive-functors-0.8.4.2/docs/Text-Digestive-Form.html#v:.:)

It is relatively easy to circumvent by specifying these fixities explicitly in `fourmolu.yaml`, but it is still a little annoying. It also took me a while to figure out why it was behaving differently.

### Your environment

Which OS do you use?
openSUSE
Which LSP client (editor/plugin) do you use?
VS Code & Haskell extension
Which version of HLS do you use and how did you install it?
1.8.0.0 via ghcup via VS Code Haskell extension
Have you configured HLS in any way (especially: a `hie.yaml` file)?
Only via VS Code

georgefst:
See #3454.

Getting these working for non-CLI mode may require some upstream refactoring to expose the right functionality, similarly to https://github.com/haskell/haskell-language-server/pull/3449.

PR #3454, opened 2023-01-17
Use Hackage fixities in Fourmolu CLI mode

georgefst:
See https://github.com/haskell/haskell-language-server/issues/3421.

### DIFF
--- a/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
+++ b/plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments           #-}
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE DisambiguateRecordFields #-}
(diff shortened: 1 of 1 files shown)


plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -56,35 +58,46 @@ properties =
...
+                []
+                ( map (T.unpack . T.intercalate "-" . reverse . drop' 2 . reverse . T.splitOn "-" . printOutputable)
+                    . mapMaybe getUnit . packageFlags
+                )
+                dynFlags
+          where
+            getUnit = \case
+                ExposePackage _ (UnitIdArg (RealUnit (Definite (UnitId s)))) _ -> Just s
+                _ -> Nothing
+            drop' n xs = drop (n `min` (Prelude.length xs - 1)) xs
```
georgefst, line 74: This is a horrible hack, but I don't have any better ideas. I wouldn't be surprised if there's no clean way to go from GHC `package-id` strings to Hackage library names.

PR #3456, opened 2023-01-17
Fix ghc min bounds for hie-compat

pepeiborra:
ghc 8.8 is no longer supported


fendor (approved):
LGTM

pepeiborra:
I'll make a Hackage revision too

Issue #3422, opened 2022-12-23
Closure property for multiple home units is not enforced nor checked


mpickering:
The blog post [[The Interface for Multiple Home Units](https://well-typed.com/blog/2022/01/multiple-home-units)](https://well-typed.com/blog/2022/01/multiple-home-units/#closure-property-for-home-units) explains the closure property needed for setting up a multiple component session.

HLS does not ensure this holds which leads to broken sessions when loading multiple components. For example.. random overlapping instance errors, panics in checkFamConst etc etc. It is **critical** that any session which HLS constructs obeys the closure property.

Stated simply the closure property is:

> Any dependency which is not a home unit must not (transitively) depend on a home unit.

And in terms of implementation there is a function in GHC called `checkHomeUnitsClosed` which implements this check and reports the reason why it's not closed. 

My proposal is

* If a new unit is added and the closure property doesn't hold, then the error is reported prominently to the user and the new component is ignored completely by the IDE.
* If a user then subsequently adds the units needed for closure then the session can resume in a multi-component manner.

My experience is that multi-component works without problem as long as the closure property. I spent the last week using the support whilst developing cabal-install, which has 5 components.

In the future:

There should be a better interface for initialising a session which contains multiple components rather than having to create the session incrementally one component at a time.

michaelpj:
Users don't "add units", so I'm not sure that the workflow you lay out can map into HLS. I guess the closest thing that would correspond to is:

- If a user opens a file in a unit such that adding it would break the closure property, then we give them an error and ignore the file.
- If a user later opens a different files (leading to more units being added) such that the original file is now fine, then when they revisit it we should try to load it again and succeed.

This is a slightly odd situation anyway: it supposes that we have two units A and B such that A depends on B _and the user can open files from both_. In that scenario why would we not immediately load B as a home unit as soon as the user opens a file in A?

The problematic case is I guess where we have A -> B -> C where A and C are local but B is not. That does seem tricky and we'll probably have to give an error. But if all the units are local I don't see why we can't just load them all as home units to begin with?

pepeiborra:
I made a similar proposal a while ago: https://github.com/haskell/hie-bios/issues/269

The problem back then is that the user doesn't really know which dependency is breaking the closure property. It sounds like `checkHomeUnitsClosed` addresses this, which is awesome. 

[...]

PR #3462, opened 2023-01-23
Add support for multi unit argument syntax

wz1000:
Implements the argument parsing logic for multiple home units found in GHC 9.4: https://downloads.haskell.org/ghc/9.4.4/docs/users_guide/using.html#multiple-home-units

We now support arguments of the form
```
-unit @unitA -unit @unitB
```

where the response files `unitA` and `unitB` contain the actual list of arguments for that unit:

```
-this-unit-id a-0.1.0.0
-i
-isrc
A1
A2
```

Also refactor the session loader and simplify it.

Also adds error messages on GHC 9.4 if the units are not closed (#3422).

Depends on https://github.com/haskell/hie-bios/pull/387


mpickering:
What happens if the closure property is violated? That seemed the main point of design on the issue?

ghcide/test/data/multi-unit/cabal.project
```diff
@@ -0,0 +1,2 @@
+packages: a b c
+multi-repl: True
```
pepeiborra, line 2: Wat?

fendor: Probably the new cabal feature that uses multiple home-units when loading them into a repl?

wz1000: Yes, @fendor is right, I generated the `-unit` files by using a special branch of cabal with the new feature and then editing it a bit.

This cabal.project isn't necessary or even useful in this patch so I will remove it.

wz1000:
> What happens if the closure property is violated?

We get the same error we would get from GHC, which is not ideal and could probably be improved.

I checked if recovery is possible once more components are loaded but it seems it is not, so that needs a bit more work.

ghcide/test/data/multi-unit/cabal.project
fendor: I am interested to know *what* branch and what @mpickering is planning there, and whether we still need the work from https://github.com/haskell/cabal/pull/7500, or whether the changes done in that fork render `cabal status` command irrelevant?
Maybe we can briefly discuss that at the next meeting.

wz1000: Here is the branch, it is still a WIP: https://github.com/mpickering/cabal/tree/wip/no-configure

mpickering:
> > What happens if the closure property is violated?
> 
> We get the same error we would get from GHC, which is not ideal and could probably be improved.
> 
> I checked if recovery is possible once more components are loaded but it seems it is not, so that needs a bit more work.

I think that things should carry on working for the unit you have currently loaded but just ignore any files which would be in the new component (until the require components are loaded in).

mpickering:
I tested again and the closure check seems to be working properly now. 

But I think the question I raised about what should happen if the closure check is violated is still unresolved.

mpickering:
Branch appears to no longer build due to a version mismatch of hie-bios. What's the plan @wz1000 ?

fendor:
I think, the plan looks roughly as follows:
* For cabal versions that support `multi-repl`, we remember the filepaths that caused components to be loaded.
  On subsequent calls to `runCradle`, we pass in all previous filepaths, s.t. hie-bios invokes `cabal repl fp1 fp2 fp3` and cabal then calculates the dependency closure. That works fine, afaict, except for potential issues with the maximal cli parameter length. This branch is currently using an extension to hie-bios allowing to pass in multiple filepaths, but it is neither merged nor has a PR yet.
* For other cabal versions, we might also be able to hack something together. After the closure check is violated, we know exactly which unit-ids are missing. So, we can look them up in `plan.json`, then look up `build-info.json` files and report these back. This will not work for certain project setups, e.g. the dependency chain of hie-compat -> hiedb -> ghcide, since hiedb is not part of this project, we can't find the compilation options, afaict.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -550,22 +546,6 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
                                        rawComponentFP
                                        rawComponentCOptions
                                        rawComponentDependencyInfo
-              -- Make a new HscEnv, we have to recompile everything from
-              -- scratch again (for now)
-              -- It's important to keep the same NameCache though for reasons
-              -- that I do not fully understand
-              logWith recorder Info $ LogMakingNewHscEnv inplace
```
fendor, line 557: Can we keep this log?

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -483,29 +498,11 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
     -- combined with the components in the old HscEnv into a new HscEnv
     -- which contains the union.
     let packageSetup :: (Maybe FilePath, NormalizedFilePath, ComponentOptions, FilePath)
-                     -> IO (HscEnv, ComponentInfo, [ComponentInfo])
+                     -> IO ([ComponentInfo], [ComponentInfo])
         packageSetup (hieYaml, cfp, opts, libDir) = do
           -- Parse DynFlags for the newly discovered component
           hscEnv <- emptyHscEnv ideNc libDir
```
fendor, line 504: Out of curiosity, do we still need to reuse the NameCache in all situations?
In theory, I don't think so for mhu. Maybe we should keep the comment about the NameCache for GHC 9.2 and 9.4?

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -483,29 +498,11 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
     -- combined with the components in the old HscEnv into a new HscEnv
     -- which contains the union.
     let packageSetup :: (Maybe FilePath, NormalizedFilePath, ComponentOptions, FilePath)
-                     -> IO (HscEnv, ComponentInfo, [ComponentInfo])
+                     -> IO ([ComponentInfo], [ComponentInfo])
```
fendor, line 501: The comment above may need some updating now, I think.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -574,49 +558,29 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
               -- . The information for the new component which caused this cache miss
               -- . The modified information (without -inplace flags) for
               --   existing packages
-              pure (Map.insert hieYaml (newHscEnv, NE.toList new_deps) m, (newHscEnv, NE.head new_deps', NE.tail new_deps'))
+              let (new,old) = NE.splitAt (NE.length new_deps) all_deps'
+              pure (Map.insert hieYaml (NE.toList all_deps) m, (new,old))
```
fendor, line 558: The comment above is outdated now, I think.

Also, is `all_deps` correct here? Will we not remember too many ComponentInfos if we lazily load all components of a project?
I understand, we don't *load* all of them since we remove duplicates later on in `newComponentCache`, but it seems weird to me.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -810,56 +774,132 @@ setNameCache :: IORef NameCache -> HscEnv -> HscEnv
...
+pprHomeUnitGraph unitEnv = Compat.vcat (map (\(k, v) -> pprHomeUnitEnv k v) $ Map.assocs $ unitEnv_graph unitEnv)
+
+pprHomeUnitEnv :: UnitId -> HomeUnitEnv -> Compat.SDoc
+pprHomeUnitEnv uid env =
+  Compat.ppr uid Compat.<+> Compat.text "(flags:" Compat.<+> Compat.ppr (homeUnitId_ $ homeUnitEnv_dflags env) Compat.<+> Compat.text "," Compat.<+> Compat.ppr (fmap homeUnitId $ homeUnitEnv_home_unit env) Compat.<+> Compat.text ")" Compat.<+> Compat.text "->"
+  Compat.$$ Compat.nest 4 (pprHPT $ homeUnitEnv_hpt env)
+
+
 -- | Create a mapping from FilePaths to HscEnvEqs
 newComponentCache
```
fendor: Outdated Docs, right? Either way, it would be nice if you can expand a little bit on the purpose of this function.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -810,56 +774,132 @@ setNameCache :: IORef NameCache -> HscEnv -> HscEnv
 #endif
 setNameCache nc hsc = hsc { hsc_NC = nc }
 
+pprHomeUnitGraph :: HomeUnitGraph -> Compat.SDoc
+pprHomeUnitGraph unitEnv = Compat.vcat (map (\(k, v) -> pprHomeUnitEnv k v) $ Map.assocs $ unitEnv_graph unitEnv)
+
+pprHomeUnitEnv :: UnitId -> HomeUnitEnv -> Compat.SDoc
+pprHomeUnitEnv uid env =
+  Compat.ppr uid Compat.<+> Compat.text "(flags:" Compat.<+> Compat.ppr (homeUnitId_ $ homeUnitEnv_dflags env) Compat.<+> Compat.text "," Compat.<+> Compat.ppr (fmap homeUnitId $ homeUnitEnv_home_unit env) Compat.<+> Compat.text ")" Compat.<+> Compat.text "->"
+  Compat.$$ Compat.nest 4 (pprHPT $ homeUnitEnv_hpt env)
```
fendor: Nit: This belongs to `Compat.Outputable.hs` imo.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -767,7 +731,7 @@ emptyHscEnv :: NameCache -> FilePath -> IO HscEnv
 emptyHscEnv :: IORef NameCache -> FilePath -> IO HscEnv
 #endif
 emptyHscEnv nc libDir = do
-    env <- runGhc (Just libDir) getSession
+    env <- runGhc (Just libDir) $ getSessionDynFlags >>= setSessionDynFlags >> getSession
```
fendor, line 731: This should be explained a little bit. Doesn't this implicitly call `initUnits` for GHC 9.0 or some other side effects?
What's this supposed to do?

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -817,49 +781,94 @@ newComponentCache
...
+        when (os == "linux") $ do
+          initObjLinker hscEnv'
+          res <- loadDLL hscEnv' "libm.so.6"
+          case res of
+            Nothing  -> pure ()
+            Just err -> logWith recorder Error $ LogDLLLoadError err
+
+        forM (Map.elems cis) $ \ci -> do
+          let df = componentDynFlags ci
+          let newFunc = maybe newHscEnvEqPreserveImportPaths newHscEnvEq cradlePath
```
fendor, line 839: Nit: A more descriptive name may be `createHscEnvEq`.

fendor (changes requested):
Looks good to me!
Quite a lot of changes, I have no objections, just a couple of questions and some nits.
Comments starting with `Nit:` can be ignored if you don't care about them.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -420,6 +434,7 @@ loadSession recorder = loadSessionWithOptions recorder def
 
 loadSessionWithOptions :: Recorder (WithPriority Log) -> SessionLoadingOptions -> FilePath -> IO (Action IdeGhcSession)
 loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
+  cradle_files <- newIORef []
```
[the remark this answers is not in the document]
wz1000: I don't think we handle deleted files very well currently, see `extendKnownFiles` which always grows (so we will always try to typecheck files that have been deleted).

We should try to address this eventually.

ghcide/session-loader/Development/IDE/Session.hs
wz1000: I think we should still reuse the NameCache.

ghcide/session-loader/Development/IDE/Session.hs
fendor: My mistake, I reviewed a commit by accident.

ghcide/session-loader/Development/IDE/Session.hs
wz1000: I've updated the comment.

We remember RawComponentInfos corresponding to each file that caused a component to be loaded. If the options for each component list out all the targets in that component, then this `[RawComponentInfos]`(all_deps) will correspond exactly to the home units in the project.

The only we we have duplication is if the options for a component don't list all the targets. We will add a "special target" for the file that caused the component to be loaded in the first place. However, when a different file loaded, since the targets were not explicitly listed, we will create a new `RawComponentInfo` for that file, so we will have multiple `RawComponentInfos` for the same component, but loaded against different files. The duplicate merging logic in `newComponentCache` takes care of that.

ghcide/session-loader/Development/IDE/Session.hs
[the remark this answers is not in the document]
wz1000: This is complicated, let's do this in another PR which explicitly adds a test for deleting files.

fendor (approved):
LGTM, let's go!

ghcide/session-loader/Development/IDE/Session.hs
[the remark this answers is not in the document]
fendor: Yes, this should happen in another PR, I was just thinking loudly.

PR #3469, opened 2023-01-27
Use apply-refact 0.12.0

### LINKED ISSUE #3241: HLS suggests quick fixes but cannot apply them
### Your environment

* Arch Linux
* Emacs 28.2 + Eglot (also tested: lsp-mode, VSCode OSS with Haskell LSP plugin 2.2.1)
* GHC 9.2.4
* HLS 1.8.0.0
* Cabal 3.6.2.0
* Stack 2.9.1

### Steps to reproduce

Write any code that HLS will suggest a quick fix for, such as: f x = id x.

### Expected behaviour

HLS should suggest a quick fix, then apply the change after selecting the quick fix.

### Actual behaviour

HLS offers the appropriate code actions, but quick fixes throw errors instead of applying the changes to the code. Other code actions, such as "Ignore hint ... in this module" still work. Other HLS features also work.
I have tested this with both Emacs LSP packages as well as VSCode, with identical results in all three cases.

### Debug information

Error condition, please check your setup and/or the [issue tracker](https://github.com/haskell/haskell-language-server/issues): ReactorMessageActionException ExitFailure 1

fendor:
Fixes #3241 by upgrading to apply-refact 0.12.0 for compatible GHC versions. Other GHC versions are not affected.

### DIFF
--- a/plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
+++ b/plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
@@ -61,10 +61,7 @@ library
     , unordered-containers
     , ghc-lib-parser
     , ghc-lib-parser-ex
-  if impl(ghc >= 9.2)
-    build-depends: apply-refact ^>= 0.11.0.0
-  else
-    build-depends: apply-refact ^>= 0.9.0.0
+    , apply-refact
 
   cpp-options:   -DHLINT_ON_GHC_LIB
   ghc-options:
--- a/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
+++ b/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
@@ -565,15 +565,7 @@ applyHint recorder ide nfp mhint =
     oldContent <- maybe (liftIO $ fmap T.decodeUtf8 (BS.readFile fp)) return mbOldContent
     modsum <- liftIO $ runAction' $ use_ GetModSummary nfp
     let dflags = ms_hspp_opts $ msrModSummary modsum
-    -- Setting a environment variable with the libdir used by ghc-exactprint.
-    -- It is a workaround for an error caused by the use of a hardcoded at compile time libdir
-    -- in ghc-exactprint that makes dependent executables non portables.
-    -- See https://github.com/alanz/ghc-exactprint/issues/96.
-    -- WARNING: this code is not thread safe, so if you try to apply several async refactorings
-    -- it could fail. That case is not very likely so we assume the risk.
-    let withRuntimeLibdir :: IO a -> IO a
-        withRuntimeLibdir = bracket_ (setEnv key $ topDir dflags) (unsetEnv key)
-            where key = "GHC_EXACTPRINT_GHC_LIBDIR"
+
     -- set Nothing as "position" for "applyRefactorings" because
     -- applyRefactorings expects the provided position to be _within_ the scope
     -- of each refactoring it will apply.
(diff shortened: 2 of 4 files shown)


plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -609,7 +601,7 @@ applyHint recorder ide nfp mhint =
                 (anns', modu') <-
                     ExceptT $ mapM (uncurry Refact.applyFixities)
                             $ postParseTransform (Right (anns, [], dflags, modu)) rigidLayout
-                liftIO $ (Right <$> withRuntimeLibdir (Refact.applyRefactorings' position commands anns' modu'))
+                liftIO $ (Right <$> Refact.applyRefactorings' position commands anns' modu')
```
fendor, line 604: This is curious, this code-path is definitely not exercised for GHC >= 9.2, since the API in apply-refact has changed in 0.11.0. And in apply-refact <= 0.10.0, I don't think we need the hack `withRuntimeLibdir` because the libdir is only needed for parsing and we pass in a module already.

fendor:
PR breaks because of https://github.com/tweag/ormolu/issues/981

hasufell (approved):

michaelpj (approved):

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
```diff
@@ -62,7 +62,7 @@ library
     , ghc-lib-parser
     , ghc-lib-parser-ex
   if impl(ghc >= 9.2)
-    build-depends: apply-refact ^>= 0.11.0.0
```
pepeiborra, line 65: Why isn't the version selection handled by Cabal? Just set proper bounds for the `ghc` dependency  in the `apply-refact` Cabal descriptor and remove all these conditionals from here

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -641,3 +633,57 @@ bimapExceptT f g (ExceptT m) = ExceptT (fmap h m) where
...
+  -- to right. An extension (e.g., @StarIsType@) may be overridden later (e.g., by @NoStarIsType@).
+  --
+  -- These are in addition to the @LANGUAGE@ pragmas in the target file. When they conflict
+  -- with the @LANGUAGE@ pragmas, pragmas win.
+  [String] ->
+  IO String
+applyRefactorings =
+#if MIN_VERSION_apply_refact(0,12,0)
+  Refact.applyRefactorings
+#elif MIN_VERSION_apply_refact(0,10,0) && !MIN_VERSION_apply_refact(0,11,0)
```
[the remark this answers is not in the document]
fendor: ok, removing it!

plugins/hls-hlint-plugin/hls-hlint-plugin.cabal
fendor: Like this? Now we just explicitly forbid 0.11.0

pepeiborra: No, simply: `build-depends: apply-refact`. Then cabal will select the version that works with the `ghc` package that is built-in

fendor: That would in theory permit version 0.11.0, I want to explicitly forbid this. Or do you think that's pointless?

pepeiborra: I think it's pointless, as it won't be used in any scenarios. If you still have concerns, mark it as deprecated in Hackage

georgefst (approved):

PR #3481, opened 2023-02-08
Experiment with loading matrix values from a file

michaelpj:
The goal here is to avoid duplication of e.g. which versions of GHC we
support.

This is an experiment at the moment.


.github/workflows/bench.yml
```diff
@@ -46,8 +46,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ['8.10.7', '9.2.5']
-        os: [ubuntu-latest]
+        ghc:
+          - '8.10.7'
+          - '9.2.5'
```
michaelpj, line 51: I'm unsure what to do here. 
- Benchmark on all GHC versions, why not?
- Benchmark on GHC versions determined by some rule, e.g. the latest two versions?
- Benchmark on some specific named versions

cc @pepeiborra

michaelpj:
Seems to work, and ensures that the jobs can't get out of sync any more!

.github/workflows/bench.yml
pepeiborra: We don't do automated regression analysis on benchmark results, so it doesn't really matter much. Benchmark on each ghc major version latest minor would be reasonable

michaelpj: Okay, I'll add a note but I don't want to get into JSON processing _right_ now :D

.github/workflows/test.yml
```diff
@@ -20,7 +20,12 @@ jobs:
     outputs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
       should_skip_ghcide: ${{ steps.skip_ghcide_check.outputs.should_skip }}
+      ghcs: ${{ steps.ghcs.outputs.ghcs }}
     steps:
+      # Need the repo checked out in order to read the file
+      - uses: actions/checkout@v3
+      - id: ghcs
+        run: echo "ghcs=$(cat ./.github/workflows/supported-ghc-versions.json)" >> $GITHUB_OUTPUT
```
fendor, line 28: Can we maybe extract these steps to https://github.com/haskell/haskell-language-server/blob/mpj/dynamic-matrices/.github/actions/setup-build/action.yml if everybody needs it?

fendor (approved):

.github/workflows/test.yml
michaelpj: Not quite, I think, because it has to be in a separate job. But we could make another reusable workflow for it.

michaelpj: Although tbh the reusable part is essentially this one line, so I'm not sure if it's worth it :thinking:

PR #3485, opened 2023-02-10
Expose `makeDefaultHandleRecorder`

ners:
I am trying to run HLS/Ghcide without giving it stdin/stdout/stderr. We can currently redirect stdout and stderr by passing appropriate values in `Arguments.argsHandleIn` and `argsHandleOut`, but we can't redirect the recorder anywhere but stderr.


fendor (approved):
LGTM

PR #3492, opened 2023-02-14
Try turning the test job into a matrix build

michaelpj:
This should reduce the latency of CI, since it will do all the tests for
all the packages in parallel. It also makes it easier and faster to
retry if the tests for one package fail.


.github/actions/setup-build/action.yml
```diff
@@ -57,11 +57,9 @@ runs:
     - if: inputs.shorten-hls == 'true'
       name: Workaround shorten binary names
       run: |
-        sed -i.bak -e 's/haskell-language-server/hls/g' \
-                    -e 's/haskell_language_server/hls/g' \
+        sed -i.bak -e 's/executable haskell-language-server/executable hls/g' \
```
michaelpj, line 60: This now _just_ shortens the executable names, _not_ the package name. That makes it easier to refer to the package.

PR #3501, opened 2023-02-21
Fix lower ghcide bounds of rename and fourmolu plugins

pepeiborra:
These plugins depend on the `usePropertyAction` api that was introduced first in ghcide 1.9


pepeiborra:
I have published Hackage revisions already

georgefst (approved):
Thanks!

OliverMadine (approved):

PR #3503, opened 2023-02-23
 Update flake to GHC 9.2.6 and 9.4.4

cydparser:
This branch includes #3373, which was approved, but needed to be rebased. Only the last two commits need to be reviewed.


fendor (approved):
Ormolu changes look good to me

PR #3517, opened 2023-03-12
check in `.envrc`

peterbecich:
Some projects check in `.envrc`, i.e.: https://github.com/input-output-hk/haskell.nix/blob/master/.envrc

Can we consider this?

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ test/testdata/**/hie.yaml
 
 # direnv
 /.direnv/
-/.envrc
 
 # bench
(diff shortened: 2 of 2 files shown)


fendor:
Hi!

I am not so surprised that a nix-first project has its '.envrc' checked in :sweat_smile: 

May I ask, why you think this is nice-to-have? Personally, I don't use the flakes and have my own expression to load dependencies. I add my own .envrc.

peterbecich:
@fendor @michaelpj good points, thanks

@fendor to answer your question, the idea of a standard dev shell for all project developers appeals to me, so long as it does not clash with local overrides. Support for `.envrc.local` or similar in the works: https://github.com/direnv/direnv/issues/556#issuecomment-1465329883

PR #3524, opened 2023-03-14
Use hie-bios 0.12

wz1000:

PR #3526, opened 2023-03-15
Remove HsLogger

fendor:
Building on top of #3524 

Last dependency on hslogger is in implicit-hie-cradle which seemingly doesn't use it, see https://github.com/Avi-D-coder/implicit-hie-cradle/pull/14


michaelpj (approved):
hooray

michaelpj:
Genuine failure with `-Werror` due to a missing import.

PR #3531, opened 2023-03-18
Don't show lenses for TH generated instances

### LINKED ISSUE #3319: Lenses for type signatures for class methods appear even for generated classes
It's common to generate classes with TH. At the moment the class plugin will generate type signature lenses for the generated class methods, which is obviously nonsense.

image

July541:
Close #3319.

Long time no see!

Coming back with a minor improvement that cut off lenses for TH-generated instances.

@michaelpj You may be interested in this :)


fendor (approved):
LGTM, awesome, was just thinking yesterday that we shouldnt show these instances!

July541:
> LGTM, awesome, was just thinking yesterday that we shouldnt show these instances!

The server kicked me off because of inactivation, but anyway, users will have a clear screen after the next release.

PR #3537, opened 2023-03-24
Prepare 1.10.0.0

wz1000:


fendor (approved):
Changes to cabal plugins look good to me.

konn (approved):
LGTM as for the Splice Plugin :+1:

isovector (approved):

July541 (approved):
call hierarchy and gadt plugins look good to me. Thanks for your work!

maralorn:
Really don’t want to hurry you, but I would like to know whether there is a timeline on this and if there are any obstacles that need to be resolved.
Our nixpkgs update is currently stuck on the half release.
We can work around this, but that doesn’t make sense if you finish the release in the next few days anyway.

Issue #3541, opened 2022-04-19
Flickering code lenses on typing


Esgariot:
https://user-images.githubusercontent.com/6072243/163984639-37625683-8efa-440d-9570-bb175aae8a70.mov

I've tried taking a look at `ghcide`, `haskell-language-server` and `lsp` repos and codebases to see which one might be responsible for dropping code lenses so quickly after typing, but didn't manage to find any clues. 

Notice how "imports" lenses stay there, while signature suggestion ones are immediately dropped on typing. Maybe leaving "stale" lenses until new ones arrive would be a way to do it? 

I suspect this is a known and old problem and everyone just disables that particular plugin and goes on. I've decided to create issue here, even though there's no code responsible for code lenses in this repo, because the issue is specific to vscode [(and their "lenses above declarations" stand)](https://github.com/Microsoft/vscode/issues/22127#issuecomment-285061033). If someone could describe some of the code parts or modules responsible for general code lens sending/updating or provide a guideline on "what" "where" and "why" I'd be happy to try to investigate.

### Your environment

Which OS do you use: MacOS

### Steps to reproduce

1. Have "insert signature" lens enabled
2. Write a snippet that would use that lens
3. Wait for lens to appear
4. Modify document

### Expected behaviour

Vertical spacing allocated by code lenses stays there, enabling typing without redrawing file.

### Actual behaviour

Editor lines are moved up and down after typing anything, as presented in attached screencast.

### Debug info

<details>
<summary>
Debug output:
</summary>

```
No 'hie.yaml' found. Try to discover the project type!
Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 1.6.1.0, Git revision f4022c5bb8530cd306c53b941878244bf27a5d41 (dirty) (51 commits) aarch64 ghc-8.10.7
Current directory: /Users/esgariot/Development/adventofcode2021/1
Operating system: darwin
Arguments: ["--debug","."]
Cradle directory: /Users/esgariot/Development/adventofcode2021/1
Cradle type: Default

Tool versions found on the $PATH
cabal:		3.6.2.0
stack:		Not found
ghc:		8.10.7


Consulting the cradle to get project GHC version...
Project GHC version: 8.10.7
haskell-language-server exe candidates: ["haskell-language-server-8.10.7","haskell-language-server"]
Launching haskell-language-server exe at:/Users/esgariot/.local/bin/haskell-language-server-8.10.7
[...]

Issue #3536, opened 2023-03-21
Implement `codeLens/resolve` for some code lenses


michaelpj:
Same as https://github.com/haskell/haskell-language-server/issues/3534 but for code lenses. Again, ideally this would come with generic support to make it easy to add to new code lens providers.

July541:
> Use use to compute the edit in codeLens/resolve

We should `useWithStale` in `resolve` also, because `useWithStale` will return the fresh result if available, and fall back to staled if unavailable.

Particularly, I think we should use `useWithStale` in the whole pipeline of code lens generation.

michaelpj:
> We should useWithStale in resolve also, because useWithStale will return the fresh result if available, and fall back to staled if unavailable.

I think we actively don't want stale results though, right? Suppose I have e.g. a hotkey to trigger the type signature code lens, and I quickly edit a function to add a new argument and then hit the hotkey. I really _don't_ want to get the type signature without the new argument - I'd much rather wait.

wz1000:
There might be a bit of confusion here, which is understandable given that there are three rather poorly named functions here

- `use`/`use_`: these will either give you an up to date result or fail.
- `useWithStale`: this will give you an up to date result, or if that fails, give you stale info. However, it will always trigger rebuilds of values and block until that operation completes.
- `useWithStaleFast`: this will never block and give you the latest stale value.

July541:
> Suppose I have e.g. a hotkey to trigger the type signature code lens, and I quickly edit a function to add a new argument and then hit the hotkey. I really don't want to get the type signature without the new argument - I'd much rather wait.

In this situation, you'll get a correct edit if your code parsed succeed, or an invalid one if parsing failed.

In some situations, the staled result still makes sense.

<img width="271" alt="image" src="https://user-images.githubusercontent.com/16057632/232293022-1d8c18d2-cb2f-4ab1-bba3-9650ca571578.png">

This staled lens is correct even if my code can't parse temporarily.

That means we have to trade off the value of the staled result.

---

As for `useWithStaleFast`, I think we never want this in lens?

michaelpj:
Ah I see. Then I guess `useWithStale` almost certainly is right. And then maybe we should be using `useWithStaleFast` more often?

July541:
> And then maybe we should be using useWithStaleFast more often?

No, `useWithStaleFast` should only be used with `IdeAction`.
[...]

PR #3545, opened 2023-04-01
Keep instance lenses stable even if parsed results are unavailable

July541:
Not function signature lenses only(#3541), instance signatures have the same issue.

This pr uses staled results to prevent flicker.

I take hls-class-plugin considered first just because I'm more familiar with this than familiar with function type lenses😉

stale


plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeLens.hs
```diff
@@ -26,17 +27,17 @@ import qualified Language.LSP.Types.Lens         as J
 codeLens :: PluginMethodHandler IdeState TextDocumentCodeLens
 codeLens state plId CodeLensParams{..} = pluginResponse $ do
     nfp <- getNormalizedFilePath uri
-    tmr <- handleMaybeM "Unable to typecheck"
+    (tmr, _) <- handleMaybeM "Unable to typecheck"
         $ liftIO
         $ runAction "classplugin.TypeCheck" state
-        $ use TypeCheck nfp
+        $ useWithStale TypeCheck nfp
```
michaelpj, line 33:
```suggestion
        -- Using stale results means that we can almost always return a value. In practice
        -- this means the lenses don't 'flicker'
        $ useWithStale TypeCheck nfp
```
worth a comment! This is a real decision!

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeLens.hs
```diff
@@ -26,17 +27,17 @@ import qualified Language.LSP.Types.Lens         as J
...
 
     -- All instance binds
-    InstanceBindTypeSigsResult allBinds <-
+    (InstanceBindTypeSigsResult allBinds, mp) <-
         handleMaybeM "Unable to get InstanceBindTypeSigsResult"
         $ liftIO
         $ runAction "classplugin.GetInstanceBindTypeSigs" state
-        $ use GetInstanceBindTypeSigs nfp
+        $ useWithStale GetInstanceBindTypeSigs nfp
```
michaelpj, line 40: and here

plugins/hls-class-plugin/src/Ide/Plugin/Class/Utils.hs
```diff
@@ -59,10 +59,10 @@ insertPragmaIfNotPresent state nfp pragma = do
     (_, fileContents) <- liftIO
         $ runAction "classplugin.insertPragmaIfNotPresent.GetFileContents" state
         $ getFileContents nfp
-    pm <- handleMaybeM "Unable to GetParsedModuleWithComments"
+    (pm, _) <- handleMaybeM "Unable to GetParsedModuleWithComments"
         $ liftIO
         $ runAction "classplugin.insertPragmaIfNotPresent.GetParsedModuleWithComments" state
-        $ use GetParsedModuleWithComments nfp
+        $ useWithStale GetParsedModuleWithComments nfp
```
michaelpj, line 65: and here

michaelpj: Interestingly, this seems like somewhere we _shouldn't_ do it. We really want the up-to-date module to compute an edit! But because we return the edit as part of the code lens, we have to. So the thing to do in the long run would be to implment `codeLens/resolve` so only the resolution method would need this. Worth a comment!

Also relevant for https://github.com/haskell/haskell-language-server/issues/3536

michaelpj (approved):
Seems like there's a general principle: code lenses should always use `useWithStale`, since otherwise they flicker. Doesn't really matter for other stuff so much. It would be nice to write that down somewhere, not sure where.

July541:
Should be pointing out that we have three lens behaviors now:
1. import lens
2. top-level signature lens
3. class instance lens

import lens has been using `useWithStale`, and I'm applying the substitution for the other two.

That means we'd have similar comments for every lens provider :(

michaelpj:
> That means we'd have similar comments for every lens provider :(

I think it's an important comment in all those places, unfortunately. Unless we want to adopt the GHC Note convention and just write a Note once and reference it from all of them.

July541:
>  So the thing to do in the long run would be to implment codeLens/resolve so only the resolution method would need this. Worth a comment!

@michaelpj I'm a little confused, do you mean we compute lenses by fresh type checked result, and resolve it with the staled result?

And, what's the meaning of `resolve`, does it mean `apply the lens` or `refresh the lens`, or anything else?

michaelpj:
> I'm a little confused, do you mean we compute lenses by fresh type checked result, and resolve it with the staled result?

Other way around :sweat_smile: I think that's reasonable? It means that when you click on the lens it might be a bit out of date, but then you'll get the right then when its inserted.

> And, what's the meaning of resolve, does it mean apply the lens or refresh the lens, or anything else?

The `codeLens/resolve` method lets you fill in some information in the lens result later. In particular, you can fill in the edit later! So you can quickly/inaccurately produce the initial lens results, and then when someone clicks on it fill in the full correct edit.

But this is future work anyway!

michaelpj (approved):

PR #3558, opened 2023-04-07
Keep type lenses stable

July541:
I got the same effect without using `PositionMapping` compared to instance lenses, but don't know why.

Can anyone explain this?

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/TypeLenses.hs
+++ b/ghcide/src/Development/IDE/Plugin/TypeLenses.hs
@@ -165,7 +165,8 @@ commandHandler _ideState wedit = do
 --------------------------------------------------------------------------------
 
 suggestSignature :: Bool -> Maybe HscEnv -> Maybe GlobalBindingTypeSigsResult -> Maybe TcModuleResult -> Maybe Bindings -> Diagnostic -> [(T.Text, [TextEdit])]
-suggestSignature isQuickFix env mGblSigs mTmr mBindings diag = suggestGlobalSignature isQuickFix mGblSigs diag <> suggestLocalSignature isQuickFix env mTmr mBindings diag
+suggestSignature isQuickFix env mGblSigs mTmr mBindings diag =
+  suggestGlobalSignature isQuickFix mGblSigs diag <> suggestLocalSignature isQuickFix env mTmr mBindings diag
 
 suggestGlobalSignature :: Bool -> Maybe GlobalBindingTypeSigsResult -> Diagnostic -> [(T.Text, [TextEdit])]
 suggestGlobalSignature isQuickFix mGblSigs Diagnostic{_message, _range}
--- a/ghcide/src/Development/IDE/Plugin/TypeLenses.hs
+++ b/ghcide/src/Development/IDE/Plugin/TypeLenses.hs
@@ -13,66 +13,69 @@ module Development.IDE.Plugin.TypeLenses (
   Log(..)
   ) where
 
-import           Control.Concurrent.STM.Stats        (atomically)
-import           Control.DeepSeq                     (rwhnf)
-import           Control.Monad                       (mzero)
-import           Control.Monad.Extra                 (whenMaybe)
-import           Control.Monad.IO.Class              (MonadIO (liftIO))
-import           Data.Aeson.Types                    (Value (..), toJSON)
-import qualified Data.Aeson.Types                    as A
-import qualified Data.HashMap.Strict                 as Map
-import           Data.List                           (find)
-import           Data.Maybe                          (catMaybes)
-import qualified Data.Text                           as T
-import           Development.IDE                     (GhcSession (..),
-                                                      HscEnvEq (hscEnv),
-                                                      RuleResult, Rules, define,
-                                                      srcSpanToRange,
-                                                      usePropertyAction,
-                                                      useWithStale)
-import           Development.IDE.Core.Compile        (TcModuleResult (..))
-import           Development.IDE.Core.Rules          (IdeState, runAction)
-import           Development.IDE.Core.RuleTypes      (GetBindings (GetBindings),
-                                                      TypeCheck (TypeCheck))
-import           Development.IDE.Core.Service        (getDiagnostics)
-import           Development.IDE.Core.Shake          (getHiddenDiagnostics, use)
-import qualified Development.IDE.Core.Shake          as Shake
+import           Control.Concurrent.STM.Stats         (atomically)
+import           Control.DeepSeq                      (rwhnf)
+import           Control.Monad                        (mzero)
+import           Control.Monad.Extra                  (whenMaybe)
+import           Control.Monad.IO.Class               (MonadIO (liftIO))
+import           Data.Aeson.Types                     (Value (..), toJSON)
+import qualified Data.Aeson.Types                     as A
+import qualified Data.HashMap.Strict                  as Map
+import           Data.List                            (find)
+import           Data.Maybe                           (catMaybes, fromMaybe,
+                                                       mapMaybe)
+import qualified Data.Text                            as T
+import           Development.IDE                      (GhcSession (..),
+                                                       HscEnvEq (hscEnv),
+                                                       RuleResult, Rules,
+                                                       define, srcSpanToRange,
+                                                       usePropertyAction,
+                                                       useWithStale)
+import           Development.IDE.Core.Compile         (TcModuleResult (..))
+import           Development.IDE.Core.PositionMapping (PositionMapping,
+                                                       toCurrentRange)
+import           Development.IDE.Core.Rules           (IdeState, runAction)
+import           Development.IDE.Core.RuleTypes       (GetBindings (GetBindings),
+                                                       TypeCheck (TypeCheck))
+import           Development.IDE.Core.Service         (getDiagnostics)
+import           Development.IDE.Core.Shake           (getHiddenDiagnostics,
+                                                       use)
+import qualified Development.IDE.Core.Shake           as Shake
 import           Development.IDE.GHC.Compat
-import           Development.IDE.GHC.Util            (printName)
+import           Development.IDE.GHC.Util             (printName)
 import           Development.IDE.Graph.Classes
-import           Development.IDE.Spans.LocalBindings (Bindings, getFuzzyScope)
-import           Development.IDE.Types.Location      (Position (Position, _character, _line),
-                                                      Range (Range, _end, _start),
-                                                      toNormalizedFilePath',
-                                                      uriToFilePath')
-import           Development.IDE.Types.Logger        (Pretty (pretty), Recorder,
-                                                      WithPriority,
-                                                      cmapWithPrio)
-import           GHC.Generics                        (Generic)
-import           Ide.Plugin.Config                   (Config)
+import           Development.IDE.Spans.LocalBindings  (Bindings, getFuzzyScope)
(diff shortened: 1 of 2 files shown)


michaelpj:
I think you should still need the position mapping otherwise the edit might go in the wrong place? It doesn't look like there's a `usesWithStale` that just gives you _one_ `PositionMapping` for the lot, which feels like an oversight (probably what you want if you're getting multiple stale pieces of info). I guess otherwise you could try to use the `PositionMapping` for the bit you get positions from?

July541:
Introducing `useWithStale` need some refactoring. Taking this opportunity, I'm wondering if it is worth extracting type lens code to a new plugin? 

cc @michaelpj @pepeiborra

michaelpj:
No particular opinion from me, I can't remember if we have any clear policy about where we're going with the `ghcide` plugins.

July541:
Replace `use` with `useWithStale` without more design.

I'll open a new pr to refactor these messed code.

July541:
@michaelpj Would you mind doing a review, I don't want mix bug fixing and refactoring

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -99,46 +103,55 @@ properties = emptyProperties
...
+        generateLensForGlobal mp sig@GlobalBindingTypeSig{..} = do
+            range <- toCurrentRange mp =<< srcSpanToRange (gbSrcSpan sig)
+            tedit <- gblBindingTypeSigToEdit sig (Just gblSigsMp)
             let wedit = toWorkSpaceEdit [tedit]
             pure $ generateLens pId range (T.pack gbRendered) wedit
-          gblSigs' = maybe [] (\(GlobalBindingTypeSigsResult x) -> x) gblSigs
-          generateLensFromDiags f =
-            sequence
-              [ pure $ generateLens pId _range title edit
+        generateLensFromDiags mp f =
```
michaelpj, line 138: Should this be using a mapping at all? The functions for getting the diagnostics don't give us a mapping, instead you're using `bindingsMp`, is that correct? It does seem like the functions for getting diagnostics should give us a mapping if they can be stale?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -99,46 +103,55 @@ properties = emptyProperties
...
+    (gblSigs@(GlobalBindingTypeSigsResult gblSigs'), gblSigsMp) <-
+      handleMaybeM "Unable to GetGlobalBindingTypeSigs"
+      $ liftIO
+      $ runAction "codeLens.GetGlobalBindingTypeSigs" ideState (useWithStale GetGlobalBindingTypeSigs nfp)
+
+    diag <- liftIO $ atomically $ getDiagnostics ideState
+    hDiag <- liftIO $ atomically $ getHiddenDiagnostics ideState
+
+    let toWorkSpaceEdit tedit = WorkspaceEdit (Just $ Map.singleton uri $ List tedit) Nothing Nothing
+        generateLensForGlobal mp sig@GlobalBindingTypeSig{..} = do
```
michaelpj, line 133: This is always called with `globalSigsMp`, and that's in scope at the definition, so maybe just use it directly and don't have the parameter?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -187,19 +212,21 @@ suggestLocalSignature isQuickFix mEnv mTmr mBindings Diagnostic{_message, _range
     , startOfLine <- Position (_line _start) startCharacter
     , beforeLine <- Range startOfLine startOfLine
     , title <- if isQuickFix then "add signature: " <> signature else signature
-    , action <- TextEdit beforeLine $ signature <> "\n" <> T.replicate (fromIntegral startCharacter) " " =
+    , range' <- fromMaybe beforeLine (flip toCurrentRange beforeLine =<< mmp)
```
michaelpj, line 215:
```suggestion
    , Just range' <- flip toCurrentRange beforeLine =<< mmp
```
?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -187,19 +212,21 @@ suggestLocalSignature isQuickFix mEnv mTmr mBindings Diagnostic{_message, _range
...
-gblBindingTypeSigToEdit :: GlobalBindingTypeSig -> Maybe TextEdit
-gblBindingTypeSigToEdit GlobalBindingTypeSig{..}
+gblBindingTypeSigToEdit :: GlobalBindingTypeSig -> Maybe PositionMapping -> Maybe TextEdit
+gblBindingTypeSigToEdit GlobalBindingTypeSig{..} mmp
   | Just Range{..} <- srcSpanToRange $ getSrcSpan gbName
     , startOfLine <- Position (_line _start) 0
-    , beforeLine <- Range startOfLine startOfLine =
-    Just $ TextEdit beforeLine $ T.pack gbRendered <> "\n"
+    , beforeLine <- Range startOfLine startOfLine
+    , range' <- fromMaybe beforeLine (flip toCurrentRange beforeLine =<< mmp)
```
michaelpj, line 228: ditto

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
michaelpj: Or rather, the logic seems slightly off. You want to return `beforeLine` if `mmp` is `Nothing`, but fail the guard if `toCurrentRange` is `Nothing`, I think.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
michaelpj: Oh this is confusing. It takes a _function_ and applies that, and _that_ will have had a position mapping applied to it, which had better be the same as this one :exploding_head:

michaelpj (approved):
A few nits but generally looks fine.

I generally feel like we should have a `usesWithStale` that returns a _single_ position mapping for all the values, so you can request many things and get a single coherent stale mapping. The current setup probably works but feels risky to me.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
July541: `gbRendered` is the parameter, I marked this explicitly.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
July541: Yes, I want to return `beforeLine` if `mmp` is Nothing or `toCurrentRange` returns Nothing, the guard seems won't fail.🤔

July541:
Stale lens for diagnostics doesn't make sense, because we always have fresh diagnostics even if current module parsed failed( the diagnostic is `parse failed`).

So there is nothing to do if we generate lenses from diagnostics.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
July541: See https://github.com/haskell/haskell-language-server/pull/3558#issuecomment-1518637683

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
michaelpj: I think we shouldn't return if `toCurrentRange` returns `Nothing`. That means we have a mapping but we can't get it to work. Rather I think the guard should just fail! Same as here: https://github.com/haskell/haskell-language-server/pull/3570#discussion_r1174427561

So it's something like:
```
Just range' <- case mmp of { Nothing -> Just beforeLine; Just mp -> toCurrentRange mp beforeLine; }
```

michaelpj:
> Stale lens for diagnostics doesn't make sense, because we always have fresh diagnostics even if current module parsed failed

But don't we mix together things we got from the diagnostics with things we got from the bindings? So the locations could be all messed up there? I haven't quite understood what's going on there.

July541:
> But don't we mix together things we got from the diagnostics with things we got from the bindings? So the locations could be all messed up there? I haven't quite understood what's going on there.

We don't. The secret is `getDiagnostics` will never fail, which means the lenses from diagnostic will flicker but others from parsed module won't.

We have 3 modes for type lenses, only lenses from `generateLensForGlobal` should keep stable, `suggestLocalSignature` relies on diagnostics. Different modes need different methods to generate lenses, so they won't be mixed.
https://github.com/haskell/haskell-language-server/blob/6640ebf33eeb8a62ccd094e2f47f69106acfd200/ghcide/src/Development/IDE/Plugin/TypeLenses.hs#L135-L140

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -99,46 +102,54 @@ properties = emptyProperties
...
+    pure $ List $ case mode of
         Always ->
-          pure (catMaybes $ generateLensForGlobal <$> gblSigs')
-            <> generateLensFromDiags (suggestLocalSignature False env tmr bindings) -- we still need diagnostics for local bindings
-        Exported -> pure $ catMaybes $ generateLensForGlobal <$> filter gbExported gblSigs'
-        Diagnostics -> generateLensFromDiags $ suggestSignature False env gblSigs tmr bindings
-    Nothing -> pure []
+          mapMaybe (generateLensForGlobal gblSigsMp) gblSigs'
+            <> generateLensFromDiags
+                (suggestLocalSignature False (Just env) (Just tmr) (Just bindings)) -- we still need diagnostics for local bindings
```
michaelpj, line 149: Here it looks like after generating stuff from the diagnostics, we're going to apply this function to the diagnostics. But this function takes `bindings`, which might need position mapping. Maybe I'm totally wrong here, it just seems suspicious.

July541: `suggestLocalSignature` relies on diagnostic, if diagnostics don't have the local signature warning, the `bindings` is useless, and if diagnostic has, that means we parsed success, and the `bindings` is fresh.
https://github.com/haskell/haskell-language-server/blob/6f5a73507f8d9266a486feaf8695c052362b9b95/ghcide/src/Development/IDE/Plugin/TypeLenses.hs#L171-L176

michaelpj: Okay. Worth a comment!

July541: Done

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -212,8 +212,9 @@ gblBindingTypeSigToEdit GlobalBindingTypeSig{..} mmp
   | Just Range{..} <- srcSpanToRange $ getSrcSpan gbName
     , startOfLine <- Position (_line _start) 0
     , beforeLine <- Range startOfLine startOfLine
-    , Just mp <- mmp
-    , Just range <- toCurrentRange mp beforeLine
+    -- If `mmp` is `Nothing`, return the original range, it used by lenses from diagnostic,
+    -- otherwise we apply `toCurrentRange`, and the guard should fail if `toCurrentRange` failed.
+    , Just range <- maybe (Just beforeLine) (flip toCurrentRange beforeLine) mmp
```
July541, line 217: I mixed the scene if `mmp` is Nothing, thanks ci halt this 😅

PR #3565, opened 2023-04-19
Allow Fourmolu 0.12

georgefst:


July541:
https://github.com/fourmolu/fourmolu/issues/309

Since fourmolu-0.11.0.0 lack of some configuration, is it possible to remove this version from the candidate list?

July541:
> 0.11 isn't actually buggy

0.11 introduced a big format diff without any space for user to configure, it doesn't obey the fourmolu's principle: `allows arbitrary configuration`

I pushed https://github.com/haskell/haskell-language-server/commit/1a8d8a28b6710d711688c134709bf58ee3795171 just make CI to green without any version opinion😅

> Besides, with index-state bumped, Cabal should always pick up 0.12 when possible.

Agree, keeping v0.11 doesn't matter for me.

georgefst:
> 0.11 introduced a big format diff without any space for user to configure, it doesn't obey the fourmolu's principle: `allows arbitrary configuration`

Fair enough, but it's never going to be possible for that to be literally true!

> I pushed https://github.com/haskell/haskell-language-server/commit/1a8d8a28b6710d711688c134709bf58ee3795171 just make CI to green without any version opinionsweat_smile

Ah, okay, I don't really understand why it made a difference, but it doesn't matter now.

> Agree, keeping v0.11 doesn't matter for me.

So, just to be clear, you're happy with this PR as-is?

July541:
> So, just to be clear, you're happy with this PR as-is?

Sure, can't wait to use Fourmolu v0.12!

July541 (approved):

Issue #3498, opened 2023-02-17
Is "version" in lsp response used to assure file consistency (e.g. for renames)?


codygman:
The context is mostly from https://github.com/joaotavora/eglot/issues/1051#issuecomment-1252791891 which says:

> Probably sending the :version isn't even necessary and Eglot will not perform the check if it isn't present. But if you are the server are going to send one, Eglot assumes its purpose is to assure file consistency and will refuse to perform the edit.

In response to an issue where the user tries to execute a rename and nothing seems to change in emacs with eglot.

michaelpj:
This is unclear. It's handled in `lsp` mostly, to be honest I'm not 100% sure how it works, so someone would need to take a look.

jameschensmith:
I believe the issue is at [`Ide.PluginUtils#L173`](https://github.com/haskell/haskell-language-server/blob/399d1a24fc578f67e47e2277429f64afb7a28f90/hls-plugin-api/src/Ide/PluginUtils.hs#L173). This is my first time going through Haskell code, but by going through `Ide.Plugin.Rename` I was able to arrive at `Ide.PluginUtils`. It looks like version is either always 0, or it defaults to 0 if there is no version passed by the client. Hope this helps!

July541:
@jameschensmith Thanks for pointing it out! I agree with you, and there is an ongoing [pr](https://github.com/haskell/haskell-language-server/pull/3566) for both this and #3547.

maralorn:
This should be fixed by the just merged #3643, maybe you can test this with master or the next hls release, when that happens.

michaelpj:
I'm going to tentatively mark this as fixed, please reopen if it recurs after the next release.

PR #3566, opened 2023-04-19
Update version while editing to conform lsp spec

### LINKED ISSUE #3547: Some LSP Code actions (Hints) are not applied.
### Your environment

Which OS do you use?
MacOS aarch64

9.2.7 from ghcup

stack

[Helix](https://github.com/helix-editor/helix/)

1.10.0.0 from ghcup

No

### Steps to reproduce

1. `hx file.hs`              # open file.hs with helix editor
2. add `func = \a -> a`
3. LSP hint will appear: "Redundant lambda (...)"
4. `space + a` to apply code action `Apply Redundant lambda`

### Expected behaviour

Code action is executed. 

### Actual behaviour
Error is shown in the editor: `Outdated workspace edit for path/to/file.hs` 
(This seems to only apply to "code hints".)

### Debug information

After asking in the helix repo (https://github.com/helix-editor/helix/issues/6543), I received this response:

> I can reproduce this but this seems like an upstream issue.
> 
> The LS standard contains the notion of a document version. Whenever a document is edited a new unique version is created. Any change is associated with a unique id and if that version doesn't match the client version then the edit is invalid and ignored by the client (as happens here).
> 
> Looking at the logs it seems that the haskell language server sends the correct version for the most part but for some events the version is just set to `Number(0)`. Perhaphs they meant to send `null` (in which case the version is ignored)? Either way there is nothing we can do about this on our side.
> 
> It seems that this currently works in vscode because vscode isn't as strict as it could/should be about enforcing these versions currently. This is a known issue with the vscode language client. See for example [microsoft/vscode-languageserver-node#752 (comment)](https://github.com/microsoft/vscode-languageserver-node/issues/752#issuecomment-795566339).

The log: 
~/.cache/helix/helix.log
```
2023-04-01T16:01:07.494 helix_term::commands::lsp [ERROR] outdated workspace edit for "/Users/roger/Documents/Dev/haskell-proj/schedule-maker/src/Lib.hs", expected 3548 but got 0
2023-04-01T16:01:07.494 helix_view::editor [ERROR] editor error: outdated workspace edit for "/Users/roger/Documents/Dev/haskell-proj/schedule-maker/src/Lib.hs"
```

July541:
Fix #3547 and #3498.

Tests is broken due to `lsp-test`'s limitation.

### DIFF
--- a/hls-plugin-api/src/Ide/PluginUtils.hs
+++ b/hls-plugin-api/src/Ide/PluginUtils.hs
@@ -98,7 +98,7 @@ data WithDeletions = IncludeDeletions | SkipDeletions
   deriving Eq
 
 -- | Generate a 'WorkspaceEdit' value from a pair of source Text
-diffText :: ClientCapabilities -> (Uri,T.Text) -> T.Text -> WithDeletions -> WorkspaceEdit
+diffText :: ClientCapabilities -> (Uri,T.Text) -> T.Text -> WithDeletions -> TextDocumentVersion -> WorkspaceEdit
 diffText clientCaps old new withDeletions =
   let
     supports = clientSupportsDocumentChanges clientCaps
@@ -161,8 +161,8 @@ diffTextEdit fText f2Text withDeletions = J.List r
 
 
 -- | A pure version of 'diffText' for testing
-diffText' :: Bool -> (Uri,T.Text) -> T.Text -> WithDeletions -> WorkspaceEdit
-diffText' supports (f,fText) f2Text withDeletions  =
+diffText' :: Bool -> (Uri,T.Text) -> T.Text -> WithDeletions -> TextDocumentVersion -> WorkspaceEdit
+diffText' supports (f,fText) f2Text withDeletions version =
   if supports
     then WorkspaceEdit Nothing (Just docChanges) Nothing
     else WorkspaceEdit (Just h) Nothing Nothing
@@ -170,7 +170,7 @@ diffText' supports (f,fText) f2Text withDeletions  =
     diff = diffTextEdit fText f2Text withDeletions
     h = H.singleton f diff
     docChanges = J.List [InL docEdit]
-    docEdit = J.TextDocumentEdit (J.VersionedTextDocumentIdentifier f (Just 0)) $ fmap InL diff
+    docEdit = J.TextDocumentEdit (J.VersionedTextDocumentIdentifier f version) $ fmap InL diff
 
 -- ---------------------------------------------------------------------
 
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
@@ -1,10 +1,10 @@
+
 {-# LANGUAGE DeriveAnyClass   #-}
 {-# LANGUAGE DeriveGeneric    #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase       #-}
 {-# LANGUAGE TypeFamilies     #-}
 {-# LANGUAGE ViewPatterns     #-}
-{-# LANGUAGE BangPatterns     #-}
 
 module Ide.Plugin.Class.Types where
 
@@ -21,6 +21,7 @@ import           Development.IDE.Graph.Classes
 import           GHC.Generics
 import           Ide.Plugin.Class.Utils
 import           Ide.Types
+import           Language.LSP.Types            (TextDocumentVersion)
 
 typeLensCommandId :: CommandId
 typeLensCommandId = "classplugin.typelens"
@@ -38,6 +39,7 @@ data AddMinimalMethodsParams = AddMinimalMethodsParams
     , methodGroup :: List (T.Text, T.Text)
     -- ^ (name text, signature text)
     , withSig     :: Bool
+    , textVersion :: TextDocumentVersion
     }
     deriving (Show, Eq, Generic, ToJSON, FromJSON)
 
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
@@ -1,4 +1,3 @@
-
 {-# LANGUAGE DeriveAnyClass   #-}
 {-# LANGUAGE DeriveGeneric    #-}
 {-# LANGUAGE FlexibleContexts #-}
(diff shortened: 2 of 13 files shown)


michaelpj:
I confess I'm very hazy about how file versions are supposed to work at all, and the LSP spec doesn't really tell you. But shouldn't there be somewhere in the code where you bump the version?

July541:
> I confess I'm very hazy about how file versions are supposed to work at all, and the LSP spec doesn't really tell you. But shouldn't there be somewhere in the code where you bump the version?

I've consulted with the client developer, see details here https://github.com/helix-editor/helix/issues/6543#issuecomment-1513286206

michaelpj:
Okay, sounds good. Maybe we should write something about how to use it? Or audit the other uses now you know what we should be doing?

July541:
I think we can add version checking in `lsp-test` for every editing after https://github.com/haskell/lsp/issues/474 addressed.

July541:
Ideally, I want we can track version info in the plugin descriptor, then we can remove some parameters passing, but it's not easy.

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -161,16 +161,16 @@ diffTextEdit fText f2Text withDeletions = J.List r
 
 
 -- | A pure version of 'diffText' for testing
-diffText' :: Bool -> (Uri,T.Text) -> T.Text -> WithDeletions -> WorkspaceEdit
-diffText' supports (f,fText) f2Text withDeletions  =
+diffText' :: Bool -> (Uri,T.Text) -> T.Text -> WithDeletions -> TextDocumentVersion -> WorkspaceEdit
```
michaelpj, line 164:
```suggestion
diffText' :: Bool -> (VersionedTextDocmentIdentifier, T.Text) -> T.Text -> WithDeletions -> TextDocumentVersion -> WorkspaceEdit
```
what about something like this? Since really the idea is that we want to use the same identifier as the one we were given - at the moment we're taking the parts and then putting it back together, we could just take the identifier.

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -38,6 +38,7 @@ data AddMinimalMethodsParams = AddMinimalMethodsParams
     , methodGroup :: List (T.Text, T.Text)
     -- ^ (name text, signature text)
     , withSig     :: Bool
+    , textVersion :: TextDocumentVersion
```
michaelpj, line 41: similarly, replace the `uri` field with a `VerisonedTextDocumentIdentifier`

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -451,8 +455,8 @@ codeActionProvider ideState pluginId (CodeActionParams _ _ documentId _ context)
 
 -- | Convert a hlint diagnostic into an apply and an ignore code action
 -- if applicable
-diagnosticToCodeActions :: DynFlags -> T.Text -> PluginId -> TextDocumentIdentifier -> LSP.Diagnostic -> [LSP.CodeAction]
-diagnosticToCodeActions dynFlags fileContents pluginId documentId diagnostic
+diagnosticToCodeActions :: DynFlags -> T.Text -> PluginId -> TextDocumentIdentifier -> TextDocumentVersion -> LSP.Diagnostic -> [LSP.CodeAction]
```
michaelpj, line 458:
```suggestion
diagnosticToCodeActions :: DynFlags -> T.Text -> PluginId -> VersionedTextDocumentIdentifier -> LSP.Diagnostic -> [LSP.CodeAction]
```

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -511,13 +515,13 @@ mkSuppressHintTextEdits dynFlags fileContents hint =
     combinedTextEdit : lineSplitTextEditList
 -- ---------------------------------------------------------------------
 
-applyAllCmd :: Recorder (WithPriority Log) -> CommandFunction IdeState Uri
-applyAllCmd recorder ide uri = do
+applyAllCmd :: Recorder (WithPriority Log) -> CommandFunction IdeState (Uri, TextDocumentVersion)
```
michaelpj, line 518:
```suggestion
applyAllCmd :: Recorder (WithPriority Log) -> CommandFunction IdeState VersionedTextDocumentIdentifier
```

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -528,10 +532,11 @@ applyAllCmd recorder ide uri = do
...
 
 data ApplyOneParams = AOP
-  { file      :: Uri
-  , start_pos :: Position
+  { file        :: Uri
+  , start_pos   :: Position
   -- | There can be more than one hint suggested at the same position, so HintTitle is used to distinguish between them.
-  , hintTitle :: HintTitle
+  , hintTitle   :: HintTitle
+  , textVersion :: TextDocumentVersion
```
michaelpj, line 539: etc.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -542,22 +547,22 @@ data OneHint = OneHint
...
     logWith recorder Debug $ LogApplying file res
     case res of
       Left err -> pure $ Left (responseError (T.pack $ "hlint:applyOne: " ++ show err))
       Right fs -> do
         _ <- sendRequest SWorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing fs) (\_ -> pure ())
         pure $ Right Null
 
-applyHint :: Recorder (WithPriority Log) -> IdeState -> NormalizedFilePath -> Maybe OneHint -> IO (Either String WorkspaceEdit)
-applyHint recorder ide nfp mhint =
+applyHint :: Recorder (WithPriority Log) -> IdeState -> NormalizedFilePath -> Maybe OneHint -> TextDocumentVersion -> IO (Either String WorkspaceEdit)
```
michaelpj, line 564: bit akward her because we've got a NFP here, but maybe we could get away with passing the VTDI instead?

michaelpj (approved):
Okay, I think I'm convinced that this is the right thing to do. It would be nice to write down the logic somewhere, I'm not sure where. Alternatively, if we go the route of passing the VTDI everywhere instead of reconstructing it, we could add a hlint rule that warns if people use the constructor?

maralorn:
I can report that the PR in this form fixes code actions in helix for me.

@July541 Just so you know, I am trying to work on this here at zurihac. Will report back once I have made progress.

maralorn:
Thank you, @July541 for doing most of the work on this. With #3643 brought over the finish line, this PR can be closed.

konn (approved):
LGTM foSplice Plugin 👍

PR #3570, opened 2023-04-22
Keep stale lenses for module name

### LINKED ISSUE #1769: Module name codelens suggestion always present (with flickering)
The codelens hint at the top of the editor always suggests a module name if the existing module name doesn't match the dir/filename structure. However, the hint collapses and is displayed again while typing, making the editor scroll up and down the screen.

Is there a way to turn module name hints off? It is possible to disable all codelens hints via the settings, but I'd like to preserve the other hints.

Here is a screen recording demonstrating the jumping around:

demo-hint-jump

July541:
Fix #1769

### DIFF
--- a/plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
+++ b/plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
@@ -17,46 +17,53 @@ module Ide.Plugin.ModuleName (
     Log,
 ) where
 
-import           Control.Monad                (forM_, void)
-import           Control.Monad.IO.Class       (liftIO)
-import           Control.Monad.Trans.Class    (lift)
+import           Control.Monad                        (forM_, void)
+import           Control.Monad.IO.Class               (liftIO)
+import           Control.Monad.Trans.Class            (lift)
 import           Control.Monad.Trans.Maybe
-import           Data.Aeson                   (Value (Null), toJSON)
-import           Data.Char                    (isLower)
-import qualified Data.HashMap.Strict          as HashMap
-import           Data.List                    (intercalate, isPrefixOf,
-                                               minimumBy)
-import qualified Data.List.NonEmpty           as NE
-import           Data.Maybe                   (maybeToList)
-import           Data.Ord                     (comparing)
-import           Data.String                  (IsString)
-import qualified Data.Text                    as T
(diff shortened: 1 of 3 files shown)


plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
```diff
@@ -160,9 +167,11 @@ pathModuleNames recorder state normFilePath filePath
 -- | The module name, as stated in the module
 codeModuleName :: IdeState -> NormalizedFilePath -> IO (Maybe (Range, T.Text))
 codeModuleName state nfp = runMaybeT $ do
-  pm <- MaybeT . runAction "ModuleName.GetParsedModule" state $ use GetParsedModule nfp
+  (pm, mp) <- MaybeT . runAction "ModuleName.GetParsedModule" state $ useWithStale GetParsedModule nfp
   L (locA -> (RealSrcSpan l _)) m <- MaybeT . pure . hsmodName . unLoc $ pm_parsed_source pm
-  pure (realSrcSpanToRange l, T.pack $ moduleNameString m)
+  let range = realSrcSpanToRange l
+  let range' = fromMaybe range (toCurrentRange mp range)
```
michaelpj, line 173: This seems wrong, shouldn't we fail or give up somehow if we can't map the range to the current thing, not use the probably-wrong range we have?

michaelpj (approved):

plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
July541: Turns out I can't have a counterexample that `toCurrentRange` can fail, keep it unless we know how it fails.

michaelpj: Seems risky to me :shrug: Why does it return a `Maybe` if it can't happen? And if it does happen we'll just silently do the wrong thing.

July541: Well, it makes sense, I'll let them be failed if `toCurrentRange` returns `Nothing`. But I'm still curious about how it fails 😅

michaelpj: If you find out, please add some Haddock! :upside_down_face:

PR #3579, opened 2023-05-06
Consider `HsExpanded` expressions during SYB traversal

### LINKED ISSUE #3574: Multiple code action for (at least) explicit record fields
Any time that expansion is used, for example, because of a dot record, the renamed ast includes an `HsExpanded` expression that contains both the original and generated expressions. This is problematic because an ordinary syb traversal of the ast usually traverses both sets of expressions, causing duplicate returns. In explicit record fields, this causes duplicate codeActions, for example.
### Steps to reproduce
This small modification of one of explicit record fields tests was able to trigger the issue

```
{-# LANGUAGE Haskell2010 #-}
{-# Language OverloadedRecordDot #-}
module Construction where

data MyRec = MyRec
  { foo :: Int
  , bar :: Int
  , baz :: Char
  }

convertMe :: () -> Int
convertMe _ =
  let foo = 3
      bar = 5
      baz = 'a'
  in MyRec {..}.foo
```
### Expected behavior
In this case, instead of the expected one codeAction offering to rewrite the wildcard when clicking upon MyRec, there were two completely identical ones.

If using `everything`, this can be fixed with using `everythingBut`, manually starting traversal on one of the branches when `HsExpanded` is found, and at the same time returning `True`, which keeps the first everything from traversing any more on that branch.
For example, this is how I did it on  the overloaded-record-dot-plugin https://github.com/haskell/haskell-language-server/blob/37463321288aeb4d8676e2f218f20fcc75d65127/plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs#L197-L208

I, unfortunately, opened a pull request on the main branch of my fork of hls, so until that closes, I won't be able to submit a fix for explicit record fields. This bug may also affect other plugins, but probably only if they used the Renamed or TypeChecked ast.

ozkutuk:
Fixes #3574. Implements the proposed solution and adds the given reproduction steps as a test case.

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE PatternSynonyms       #-}
+{-# LANGUAGE TupleSections         #-}
 {-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE TypeOperators         #-}
 {-# LANGUAGE ViewPatterns          #-}
@@ -18,12 +19,13 @@ module Ide.Plugin.ExplicitFields
 import           Control.Lens                    ((^.))
 import           Control.Monad.IO.Class          (MonadIO, liftIO)
 import           Control.Monad.Trans.Except      (ExceptT)
+import           Data.Bifunctor                  (first)
 import           Data.Functor                    ((<&>))
-import           Data.Generics                   (GenericQ, everything, extQ,
-                                                  mkQ)
+import           Data.Generics                   (GenericQ, everything,
+                                                  everythingBut, extQ, mkQ)
 import qualified Data.HashMap.Strict             as HashMap
-import           Data.Maybe                      (isJust, listToMaybe,
-                                                  maybeToList, fromMaybe)
+import           Data.Maybe                      (fromMaybe, isJust,
+                                                  listToMaybe, maybeToList)
 import           Data.Text                       (Text)
 import           Development.IDE                 (IdeState, NormalizedFilePath,
                                                   Pretty (..), Recorder (..),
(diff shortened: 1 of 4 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -347,14 +354,19 @@ collectRecords = everything (<>) (maybeToList . (Nothing `mkQ` getRecPatterns `e
 collectNames :: GenericQ (UniqFM Name [Name])
 collectNames = everything (plusUFM_C (<>)) (emptyUFM `mkQ` (\x -> unitUFM x [x]))
 
-getRecCons :: LHsExpr (GhcPass 'Renamed) -> Maybe RecordInfo
+getRecCons :: LHsExpr (GhcPass 'Renamed) -> (Maybe RecordInfo, Bool)
```
joyfulmantis, line 357: The problem with returning a Maybe instead of a List here is that you can only return one match from within the HsExpanded expression, even though there could be multiple matches. This example showcases this. 

```
{-# LANGUAGE Haskell2010 #-}
{-# LANGUAGE OverloadedRecordDot #-}
module Construction where

data MyRec = MyRec
  { foo :: Int
  , bar :: Int
  , baz :: Char
  }

data YourRec = YourRec
  { foo2 :: MyRec
  , bar2 :: Int
  , baz2 :: Char
  }

myRecExample = MyRec {..}
  where
    foo = 5
    bar = 6
    baz = 'a'

yourRecExample = YourRec {..}
    where
        foo2 = myRecExample
        bar2 = 5
        baz2 = 'a'

convertMe :: () -> Int
convertMe _ =
  (let MyRec{..} = myRecExample
       YourRec{..} = yourRecExample
    in foo2).foo
```
In this case, your code, would only offer to rewrite the first matched wildcard instead of both. 
This example is very constructed,  however, HsExpansion could possibly be used for more cases in the future, which would cause these sorts of edge cases to also appear more often.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -347,14 +354,19 @@ collectRecords = everything (<>) (maybeToList . (Nothing `mkQ` getRecPatterns `e
...
+  | isJust (rec_dotdot flds) = (mkRecInfo e, False)
   where
     mkRecInfo :: LHsExpr (GhcPass 'Renamed) -> Maybe RecordInfo
     mkRecInfo expr = listToMaybe
       [ RecordInfoCon realSpan' (unLoc expr) | RealSrcSpan realSpan' _ <- [ getLoc expr ]]
-getRecCons _ = Nothing
+getRecCons _ = (Nothing, False)
 
 getRecPatterns :: LPat (GhcPass 'Renamed) -> Maybe RecordInfo
 getRecPatterns conPat@(conPatDetails . unLoc -> Just (RecCon flds))
```
joyfulmantis, line 372: ConPat can also occur inside an HsExpanded expression. For example, the constructed example above uses ConPat (because the wildcards occur in the let binding)

joyfulmantis (changes requested):
Thanks for getting the patch out so quickly!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
joyfulmantis: Actually thinking about it though, you shouldn't need to change this, because `getRecCon` should already, by calling `collectRecords`, return both `RecordCon` and `ConPat` matches. May want to point that out in the comments though.

joyfulmantis:
Closed. Completed in #3750 .

PR #3512, opened 2023-03-03
CI: GHC 9.2.7 Support

andys8:
Draft for now to check with CI what needs to be changed.

Solves #3511

PR #3580, opened 2023-05-07
CI: Simplify to read CI output and don't fail fast

fendor:
Dont rerun CI with various `lsp-test` flags to increase the output verbosity.
In my opinion, it makes it unnecessarily hard to find the failing test-case in CI, and I don't think anyone is reading these logs to actually *debug* test-cases.

Also, we don't want to fail fast.
We used to fail fast, to avoid caches of failing PRs to overpopulate the CI
caches, evicting known good build caches.
However, PRs do not cache anything any more, and can only use the caches from the current master.
See 'caching.yml' for our caching set up.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -50,12 +50,6 @@ runs:
         echo "CABAL_PKGS_DIR=~/.cabal/packages" >> $GITHUB_ENV
       shell: bash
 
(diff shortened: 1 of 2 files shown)


fendor:
@andys8 With this PR, it should be trivial to get #3512 in.

michaelpj (approved):

michaelpj:
I agree with the reasoning :+1:

Issue #3581, opened 2023-05-07
CI: Dont block CI on flaky tests


fendor:
## Is your enhancement request related to a problem? Please describe.

CI is currently getting stuck a lot. This is terrible for contributors, even simple, small changes can take weeks to merge because CI is taking multiple runs to finally succeed. The main reason for these CI issues are flaky tests, i.e. tests that sometimes fail and sometimes succeed. Contributors have no way of fixing these failing tests, waiting for a contributor who has the permission to restart failing CI jobs.

I think this drags down morale and exhausts possible contributors.

## Describe the solution you'd like

As long as we cannot avoid having flaky tests, these flaky tests should not block CI.
Thus, I propose to run the tests in CI, but no longer fail a CI job if a flaky test fails.
We introduce a new combinator 

```haskell
-- | Mark test as flaky for a specific target
knownFlakyFor :: BrokenTarget -> String -> TestTree -> TestTree
knownFlakyFor ...

-- | Mark test as flaky for all GHC versions
knownFlaky :: String -> TestTree -> TestTree
knownFlaky ...
```

which works like the existing `knownBrokenFor` and `ignoreFor` functions and ties into the same infrastructure, by allowing us to specify on which platform the tests are flaky although, it stands to reason, they are not flaky because of GHC but LSP things.

I want to mark all tests that are known flaky with these combinators, together with possible reasons.

CI should no longer retry a single test suite, but whenever a test is demonstrable flaky, then it should be marked as such.

## Describe alternatives you've considered

Fix all the flaky tests, by either improving lsp-test to be easier to use correctly, migrating integration-tests to unit-tests or fixing the test-cases by introducing some kind of sync points.

I also don't think we run the risk of marking a test as flaky and forgetting about them instead of fixing them. Currently, if a test is flaky, we just restart the CI, usually not even looking at what test failed in particular (correct me if I am wrong).
Testsuite maintenance and bitrot is an important topic anyway, and we need to review the tests that are marked as broken as well, so we just document the status quo in the code, in my opinion.

PR #3582, opened 2023-05-07
Add testutils to mark tests as flaky

fendor:
Issue: #3581 

This PR does three things:
* Don't rerun CI test failures
* Introduce combinators to mark tests as flaky
* Mark all tests as flaky that fail on CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,14 +118,14 @@ jobs:
 
       - if: matrix.test
         name: Test hls-plugin-api
-        run: cabal test hls-plugin-api --test-options="$TEST_OPTS" || cabal test hls-plugin-api --test-options="$TEST_OPTS"
+        run: cabal test hls-plugin-api --test-options="$TEST_OPTS"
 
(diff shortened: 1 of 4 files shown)


fendor:
> Should we have a corresponding ticket for every flaky test?

It would be great, but looking at this CI result, I expect to have 40-50 flaky tests.
Having an issue for each of them looks like overkill to me.

fendor:
Due to the extreme flakiness of lsp-tests, I'm back paddling my extreme approach of marking everything that fails once as flaky. That's a bit disappointing, but I think we'd have to essentially mark every test as flaky eventually, since it is that difficult to write a completely deterministic test with lsp-test (not necessarily the fault of lsp-test).
Thus, we run the test suite twice to wheat out the most flaky behaviour. If a test fails twice in a row, it is marked as flaky. It looks like this will allow us to have fewer flaky tests, thus creating one issue for each seems feasible again.

fendor:
Closing as this seems to be useless for now and I don't intend to get this ready for merge any time soon.

PR #3599, opened 2023-05-17
Bump all packages to version 2.0.0.0 and prepare the release

wz1000:
Bump all packages to version 2.0.0.0.
    
This simplifies the release process as we can bump all package versions in
lockstep prior to a release instead of having to decide version bumps on a case
to case basis while adhering to PVP.

We don't test arbitrary configurations mixing versions of plugins/ghcide which
you might get if you install HLS from hackage. With this in mind, it makes sense
to pin the versions of all packages in this repository

Exceptions include:
- hie-compat (it is used by hiedb and is mostly separate from the other packages - it changes about once a GHC release)
- shake-bench (not meant to be distributed for users, for internal testing only

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -212,7 +212,7 @@ jobs:
 
   build-mac-x86_64:
     name: Build binary (Mac x86_64)
-    runs-on: macOS-10.15
+    runs-on: macOS-11
     env:
       MACOSX_DEPLOYMENT_TARGET: 10.13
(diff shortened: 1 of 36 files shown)


fendor (approved):
LGTM!

July541 (approved):
Thanks, reasonable to me!

konn (approved):
LGTM, thanks!

michaelpj (approved):

georgefst (approved):
Looks good, though I would be tempted to wait a day or two to see if there's a possibility of a new HLint release which we could include: https://github.com/ndmitchell/hlint/issues/1477#issuecomment-1551544708.

PR #3608, opened 2023-05-27
Prepare 2.0.0.1 release (for GHC 9.6.2 and 9.2.8)

### LINKED ISSUE #3350: Plugin for record-selector to record dot syntax
If the user's editor looks like this:

```
   bar = unFoo foo
             ^
```

We could provide a code action that produces:

```
   bar = foo.unFoo
```

We could also provide this functionality file-wide, either as a an action on the module or in the file explorer.

hasufell:
## Changelog 2.0.0.1

- Add overloaded record dot plugin intial version (closes #3350) (#3560)
- Binaries for GHC 9.2.8 and GHC 9.6.2

----

@michaelpj @fendor @wz1000 @mpickering 

----

## Release checklist

- [x] check ghcup supports new GHC releases if any
- [x] set the supported GHCs in workflow file
- [x] check all plugins still work
- [x] bump package versions
- [x] generate and update changelog
- [x] create release branch
    - https://github.com/haskell/haskell-language-server/tree/wip/2.0.0.1
- [x] create release tag
    - https://github.com/haskell/haskell-language-server/tree/2.0.0.1
- [x] trigger release pipeline (creates a draft)
- [x] run `scripts/release/download-gh-artifacts` to add FreeBSD bindist and signatures
- [x] upload bindists to downloads.haskell.org
- [x] create PR to ghcup-metadata
- [x] get sign-off on release
- [x] publish release on github
- [x] upload hackage packages
- [x] update https://github.com/haskell/ghcup-metadata/blob/develop/hls-metadata-0.0.1.json
- [x] update https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html#current-ghc-version-support-status
- [ ] post release on discourse and reddit

### DIFF
--- a/.github/scripts/bindist.sh
+++ b/.github/scripts/bindist.sh
@@ -25,7 +25,6 @@ case "${TARBALL_EXT}" in
 		: "${GHCS:="$(cd "$CI_PROJECT_DIR/out/${ARTIFACT}" && rm -f ./*.json && for ghc in * ; do printf "%s\n" "$ghc" ; done | sort -r | tr '\n' ' ')"}"
 		emake --version
 		emake GHCUP=ghcup ARTIFACT="${ARTIFACT}" GHCS="${GHCS}" bindist
-		emake GHCUP=ghcup                        GHCS="${GHCS}" clean-ghcs
 		emake GHCUP=ghcup ARTIFACT="${ARTIFACT}"                bindist-tar
         ;;
     *)
--- a/.github/scripts/bindist.sh
+++ b/.github/scripts/bindist.sh
@@ -25,8 +25,8 @@ case "${TARBALL_EXT}" in
 		: "${GHCS:="$(cd "$CI_PROJECT_DIR/out/${ARTIFACT}" && rm -f ./*.json && for ghc in * ; do printf "%s\n" "$ghc" ; done | sort -r | tr '\n' ' ')"}"
 		emake --version
 		emake GHCUP=ghcup ARTIFACT="${ARTIFACT}" GHCS="${GHCS}" bindist
-		emake GHCUP=ghcup ARTIFACT="${ARTIFACT}"                bindist-tar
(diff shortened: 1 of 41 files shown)


hasufell:
Tag build: https://github.com/haskell/haskell-language-server/actions/runs/5096814660

hasufell:
There's a problem with the release tar jobs running out of space... it seems we need to improve the space management somehow, otherwise we'll be constrained to 6 GHCs per tarball or so.

wz1000:
I don't think we need 9.2.7 either.

wz1000 (approved):

hasufell:
> I don't think we need 9.2.7 either.

I think we do. It's the last 'recommended' release, which means lots of users will still be on it and have a saturated cabal cache.

But if dropping it makes CI pass, ok.

We might need some tweaks to the Makefile to delete GHC versions directly after they are used.

My laptop is in repair, so I can't continue.

wz1000:
> I think we do. It's the last 'recommended' release, which means lots of users will still be on it and have a saturated cabal cache.

Those users can continue using 2.0.0.0

hasufell:
> > I think we do. It's the last 'recommended' release, which means lots of users will still be on it and have a saturated cabal cache.
> 
> Those users can continue using 2.0.0.0

That's not really good user experience. Only our vscode plugin is smart enough to downgrade automatically. 

GHCup doesn't have a simple way to show a matrix. Most users will bump HLS to recommended/latest and still have projects using 9.2.7, because they don't want to rebuild all dependencies just yet.

aryairani:
Out of disk space in CI

hasufell:
Space issues have been fixed: https://github.com/haskell/haskell-language-server/actions/runs/5370295916

I'll finalize the release shortly.

hasufell:
https://github.com/haskell/ghcup-metadata/pull/96

Issue #2931, opened 2022-05-26
SQLite error: database is locked


JoeMShanahan:
I am using HLS on a large project with VSCode, and I am being spammed with the following the error:

    Error condition, please check your setup and/or the issue tracker: 
    HieDb writer thread SQLite error:
      SQLite3 returned ErrorBusy while attempting to perform step: database is locked

I've tried restarting HLS and destroying any stack cache (didn't really expect that one to work!). Happy to provide anything else if there's not enough to go on here. I am thinking there's a manual way to clear the lock, but I don't know what that is!

eddiemundo:
You can try removing `.cache/ghcide`.

aryairani:
I run into this problem too. Restarting VSCode fixes it temporarily, but I wonder if there's a better solution?

eddiemundo:
According to https://www.sqlite.org/rescode.html#locked it says if not in shared memory mode then this error only occurs if incompatible operations happen on the same connection. So because shared memory is off (I think), then it must be because of incompatible operations.

The only operations I see that might do this after HLS has started up is when indexing hie files (`addRefsFromLoaded` from hiedb), and when `setSomethingModified` (`deleteMissingRealFiles` from hiedb) which happens when closing files/workspace config changes. All writes to the hiedb are queued up, but the reads aren't, so my theory is that a read, and a delete on the same table or row is happening simultaneously resulting in a locked error.

Another possibility is writes are happening simultaneously at startup somewhere, but there are only a few places that can happen, writer thread startup, and `runWithDb`.

So if my theory is actually right then I guess one way of fixing it is to retry not only on SQL_BUSY but SQL_LOCKED as well.

July541:
hls-call-hierarchy-plugin had a hiedb write before hls-1.7.0.0, but I never saw errors like this while I'm debugging that, just write here for reference.

PaulJohnson:
I believe this is the bug being discussed in https://github.com/haskell/vscode-haskell/issues/300 and https://github.com/haskell/vscode-haskell/issues/480

Briefly, it seems the VSCode Haskell extension sometimes starts up more than one HLS. That would explain why you get database lock errors. You will probably see duplicate info on popups driven by HLS as well.

lylek:
This has been a major problem for me lately in my daily work. Just within the past month, it has started happening almost constantly, and it's very distracting. Would love to see some resolution on this.

PR #3610, opened 2023-05-28
fix: remove the `Indexing` progress message when exeption in withHieDb

guibou:
If any exception was happening in `withHieDb`, the `post` action was never called, leading to a stale `Indexing` message in the client.

By using `bracket`, we ensure that the `post` message is called in the event of an exception, cleaning the confusing message in the client.

It does not change the well behaving of the indexing (i.e. if it was crashing, it is still crashing), but at least user won't be confused by a stale `Indexing` message.

See https://github.com/haskell/haskell-language-server/issues/2931 for an example when an exception can happen during the indexing.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -901,9 +901,10 @@ indexHieFile se mod_summary srcPath !hash hf = do
             -- If the hash in the pending list doesn't match the current hash, then skip
             Just pendingHash -> pendingHash /= hash
         unless newerScheduled $ do
-          pre optProgressStyle
-          withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
-          post
+          -- Using bracket, so even if an exception happen during withHieDb call,
+          -- the `post` (which clean the progress indicator) will still be called.
+          bracket_ (pre optProgressStyle) post $
+            withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
   where
     mod_location    = ms_location mod_summary
     targetPath      = Compat.ml_hie_file mod_location


guibou:
I tried this code for one week at work (super large haskell codebase (hls takes 3 to 5 minutes to start), 2k+ modules, 20 devs) and the stuck `Indexing` message does not appears anymore.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -902,8 +902,8 @@ indexHieFile se mod_summary srcPath !hash hf = do
             Just pendingHash -> pendingHash /= hash
         unless newerScheduled $ do
           pre optProgressStyle
-          withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
-          post
+          bracket_ (pre optProgressStyle) post $
+            withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
```
fendor, line 906: Is the first line still necessary?

fendor (changes requested):
LGTM! 

What do you think about adding your reasoning to the code here?

wz1000 (approved):
This seems like a good idea

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -902,8 +902,8 @@ indexHieFile se mod_summary srcPath !hash hf = do
             Just pendingHash -> pendingHash /= hash
         unless newerScheduled $ do
           pre optProgressStyle
```
wz1000, line 904: This should be removed?

ghcide/src/Development/IDE/Core/Compile.hs
guibou: OMG, that's a mistake. Sorry. Yes. I'll remove it.

ghcide/src/Development/IDE/Core/Compile.hs
guibou: Yes. Sorry about that.

guibou: Fixed. I'm so sorry (i've done some cleanup before sending the MR and as usual, I broke stuff)

guibou:
> LGTM!
> 
> What do you think about adding your reasoning to the code here?

I've added a comment.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -901,9 +901,10 @@ indexHieFile se mod_summary srcPath !hash hf = do
             -- If the hash in the pending list doesn't match the current hash, then skip
             Just pendingHash -> pendingHash /= hash
         unless newerScheduled $ do
-          pre optProgressStyle
-          withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
-          post
+          -- Using bracket, so even if an exception happen during withHieDb call,
+          -- the `post` (which clean the progress indicator) will still be called.
+          bracket_ (pre optProgressStyle) post $
+            withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
```
wz1000, line 907: thinking about it more, I guess we should report to the user that an exception happened instead of silently ignoring it?

guibou: Well, `bracket` does not ignore the exception. Right now it is propagated up, but before it ensures that `post` is called.

Actually, the behavior I had in my editor (neovim) was:

- Before: An exception is displayed in the editor, and then `Indexing` is stuck for an eternity
- Now: The exception is still displayed, but `Indexing` is not stuck anymore.

wz1000: OK, that makes sense. But it still seems wrong in `post` to just increment the counter and pretend nothing happened.

Perhaps we should also keep a record of the failures and report them separately when progress ends?

Ideally we would also retry on the failures, but that gets thorny very quickly. We don't want to keep retrying indefinitely! How have you observed this failing? Is it always just multiple HLS processes? We could add a lock to the `~/.cache/ghcide/...` directory to prevent starting multiple HLS processes on the same project.

I don't expect these to be resolved in this PR, but we should think more about all these interactions.

guibou:
```
HieDb writer thread SQLite error:\n  SQLite3 returned ErrorConstraint while attempting to perform step: UNIQUE constraint failed: defs.hieFile, defs.occ
```

Here is the exception which was returned up to my editor.

fendor (approved):
LGTM :)

ghcide/src/Development/IDE/Core/Compile.hs
guibou: I've opened https://github.com/haskell/haskell-language-server/issues/3622, thank you for remaining me to do it.

Issue #3619, opened 2023-05-31
Editor Configuration Documentation improvement


nlander:
@hasufell @wz1000 As a follow up to discussion on the monthly hls contributers meeting today, I am attempting to lay out some of the issues I have had setting up hls for neovim and some thoughts I have on how the docs could be improved. Appologies if this is a bit longwinded; I am trying to be thorough in documenting my experience.

One idea that stands out to me is that the docs seem to assume that users know how to configure their editor of choice, which I don't think is always a reasonable assumption. In my case, I have previously only used neovim with an extremely minimal configuration and no plugins. I know nothing about lua and how to use it to configure neovim for more complex plugin configuration. As vim/neovim is an editor that can be associated with minimalism, I could imagine that I am not the only person who could find themselves wanting to use hls with neovim and not knowing how to use lua. I could also imagine someone new to haskell trying out more than one new thing at once, and not knowing anything about their editor of choice for the moment. In general I think it would be nice if for as many editors as possible we can include a minimal viable complete configuration for the editor, assuming that the user is coming in with no previous configuration. This should include anything that needs to be installed for the configuration to work (like lua, luarocks, dependency plugins, etc.).

For me specifically, here are the pain points I encountered while getting neovim set up, and the questions I still have. I chose to try [haskell-tools.nvim](https://github.com/MrcJkb/haskell-tools.nvim) first.

- I needed to upgrade my version of neovim, as the apt package is pretty far out of date. I ended up using the [neovim](https://github.com/richin13/asdf-neovim) plugin for [asdf](https://asdf-vm.com/) for this, which worked fine, except that for some reason `~/.asdf/shims` didn't get added to my path when I installed asdf (this likely is related to the fact that I use fish shell).
[...]

PR #3620, opened 2023-06-01
WIP Add detailed neovim setup docs

nlander:
This adds detailed neovim setup documentation as described in #3619.

Any comments from hls users that also use neovim would be appreciated!

I am still unclear if the configuration I have in `~/.config/nvim/ftplugin/haskell.lua` is correct, complete and not redundant.

It would also be nice if instead of stopping at the installation stage if we could include a short description of how to use some of the most common hls features with the default keybindings. I'm still experimenting with using hls in neovim, so any help on this part would be appreciated.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -262,7 +262,7 @@ See [the Sublime Text LSP documentation](https://lsp.sublimetext.io) for informa
 
 ### [Neovim](https://neovim.io)
 
-Neovim provides a [native LSP implementation with a Lua framework](https://neovim.io/doc/user/lsp).
+Neovim provides a [native LSP implementation with a Lua framework](https://neovim.io/doc/user/lsp). If you are unfamiliar with how to use lua to configure neovim, check out [our detailed neovim instructions](./configuration/neovim.md)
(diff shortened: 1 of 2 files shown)


Anrock:
Not sure if using asdf (this is literally the first time I heard about it) instead of native package manager to install neovim is good idea. Neovim pretty much universally available on all distros with newest versions sometimes requiring some additional steps like setting up a ppa for .deb based stuff. IMO it would be better to just leave a link to [official doc](https://github.com/neovim/neovim#install-from-package) instead of introducing some additional tool.

Same goes for installing lua with asdf. Why use asdf to install lua if lua is widely spread and probably every existing distro has it packaged. Why install it at all if neovim has builtin lua interpreter? Why install packages via luarocks if common way to install plugins is to use some neovim plugin manager like lazy, paq or whatever is popular this month.

Copypasting install instructions from haskell-tools, ghcup and other things is bad imo because as soon as they change something in setup procedure this doc will become obsolete.

Overall I'd say this doc is bad: it's brittle because instructions are copypasted instead of linking and following described procedure will leave user with uncommon neovim setup and bunch of stuff that isn't managed by system package manager.

PR #3621, opened 2023-06-01
HLS for the new generated LSP 2 release

joyfulmantis:
I am going through the codebase to update it to @michaelpj 's upcoming LSP 2 release. So far the only part that compiles is hls-plugin-api. I am currently working on ghcide, which should be the last hard part of this change. The new LSP release is generated, which means there are a lot of changes, hence this PR will probably end up touching pretty much every part of the project.

If you would like to start reviewing this work, you should start with hls-plugin-api. Again it is the only part that compiles, and other parts will be subject to significant change.


joyfulmantis:
Currently, everything should compile. However, there are still a bunch of test failures. Most of the failures are caused by strangeness with TextDocumentContentChangeEvent, for some reason it seems like lsp is thinking partial changes are whole (even though the vfs seems to have the properly edited file)

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -303,7 +303,7 @@ type WithProgressFunc = forall a.
 type WithIndefiniteProgressFunc = forall a.
     T.Text -> LSP.ProgressCancellable -> IO a -> IO a
 
-type GetStalePersistent = NormalizedFilePath -> IdeAction (Maybe (Dynamic,PositionDelta,TextDocumentVersion))
+type GetStalePersistent = NormalizedFilePath -> IdeAction (Maybe (Dynamic,PositionDelta,Maybe Int32))
```
michaelpj, line 306: Hmm, maybe we should define this synonym again...

ghcide/src/Development/IDE/LSP/HoverDefinition.hs
```diff
@@ -16,44 +16,46 @@ module Development.IDE.LSP.HoverDefinition
...
-gotoDefinition :: IdeState -> TextDocumentPositionParams -> LSP.LspM c (Either ResponseError (ResponseResult TextDocumentDefinition))
-hover          :: IdeState -> TextDocumentPositionParams -> LSP.LspM c (Either ResponseError (Maybe Hover))
-gotoTypeDefinition :: IdeState -> TextDocumentPositionParams -> LSP.LspM c (Either ResponseError (ResponseResult TextDocumentTypeDefinition))
-documentHighlight :: IdeState -> TextDocumentPositionParams -> LSP.LspM c (Either ResponseError (List DocumentHighlight))
-gotoDefinition = request "Definition" getDefinition (InR $ InL $ List []) (InR . InL . List)
-gotoTypeDefinition = request "TypeDefinition" getTypeDefinition (InR $ InL $ List []) (InR . InL . List)
-hover          = request "Hover"      getAtPoint     Nothing      foundHover
-documentHighlight = request "DocumentHighlight" highlightAtPoint (List []) List
+gotoDefinition :: IdeState -> TextDocumentPositionParams -> LSP.LspM c (Either ResponseError (MessageResult Method_TextDocumentDefinition))
+hover          :: IdeState -> TextDocumentPositionParams -> LSP.LspM c (Either ResponseError (Hover |? Null))
```
michaelpj, line 32: Similarly, I'd be tempted to tidy these up and put some adaptation code in the handler. Not sure.

michaelpj: But probably not in this PR to keep things simple.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -263,27 +263,30 @@ untilMVar mvar io = void $
     waitAnyCancel =<< traverse async [ io , readMVar mvar ]
 
 cancelHandler :: (SomeLspId -> IO ()) -> LSP.Handlers (ServerM c)
-cancelHandler cancelRequest = LSP.notificationHandler SCancelRequest $ \NotificationMessage{_params=CancelParams{_id}} ->
-  liftIO $ cancelRequest (SomeLspId _id)
+cancelHandler cancelRequest = LSP.notificationHandler SMethod_CancelRequest $ \TNotificationMessage{_params=CancelParams{_id}} ->
+  liftIO $ cancelRequest (SomeLspId (toLspId _id))
+  where toLspId :: (Int32 |? T.Text) -> LspId a
```
michaelpj, line 268: Should be in `lsp`, perhaps? We define `LspId` there...

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -258,7 +258,9 @@ defDocumentSymbol l = DocumentSymbol { .. } where
   _detail         = Nothing
   _deprecated     = Nothing
   _name           = ""
-  _kind           = SkUnknown 0
+  -- This used to be SkUnknown 0, which is invalid, as SymbolKinds start at 1,
```
michaelpj, line 261: Yeah, this is kind of gross, but it's clearly meant to be an arbitrary choice, so this seems okay.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -278,29 +283,29 @@ showForSnippet x = printOutputable x
 
 mkModCompl :: T.Text -> CompletionItem
 mkModCompl label =
-  CompletionItem label (Just CiModule) Nothing Nothing
+  CompletionItem label Nothing (Just CompletionItemKind_Module) Nothing Nothing
```
michaelpj, line 286: :vomiting_face: maybe we could make these use record syntax...

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -157,46 +157,47 @@ executeCommandHandlers recorder ecs = requestHandler SWorkspaceExecuteCommand ex
       _                    -> Nothing
 
     -- The parameters to the HLS command are always the first element
-
+    execCmd :: IdeState -> ExecuteCommandParams -> LSP.LspT Config IO (Either ResponseError (A.Value |? Null))
```
michaelpj, line 160: :muscle:

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -391,13 +392,15 @@ toUri = fromNormalizedUri . filePathToUri' . toNormalizedFilePath'
...
   where
     kind
-      | isVarOcc defNameOcc = SkVariable
-      | isDataOcc defNameOcc = SkConstructor
-      | isTcOcc defNameOcc = SkStruct
-      | otherwise = SkUnknown 1
+      | isVarOcc defNameOcc = SymbolKind_Variable
+      | isDataOcc defNameOcc = SymbolKind_Constructor
+      | isTcOcc defNameOcc = SymbolKind_Struct
+        -- This used to be (SkUnknown 1), buth there is no SymbolKind_Unknown.
```
michaelpj, line 401: the function returns a Maybe, can we make it return Nothing in that case? maybe we just don't want to, and we'd rather return something?

michaelpj: Although again I guess not in this PR

ghcide/test/exe/Main.hs
```diff
@@ -175,14 +170,14 @@ instance Pretty Log where
 -- | Wait for the next progress begin step
 waitForProgressBegin :: Session ()
 waitForProgressBegin = skipManyTill anyMessage $ satisfyMaybe $ \case
-  FromServerMess SProgress (NotificationMessage _ _ (ProgressParams _ (Begin _))) -> Just ()
+  FromServerMess  SMethod_Progress  (TNotificationMessage _ _ (ProgressParams _ v)) | not $ Lens.isn't _workDoneProgressBegin v-> Just ()
```
michaelpj, line 173: still vaguely sus about this. Or maybe we need to export a more useful helper from lsp?

ghcide/test/src/Development/IDE/Test.hs
```diff
@@ -252,9 +251,10 @@ isReferenceReady p = void $ referenceReady (equalFilePath p)
 
 referenceReady :: (FilePath -> Bool) -> Session FilePath
 referenceReady pred = satisfyMaybe $ \case
-  FromServerMess (SCustomMethod "ghcide/reference/ready") (NotMess NotificationMessage{_params})
+  FromServerMess cm@(SMethod_CustomMethod _) (NotMess TNotificationMessage{_params})
     | A.Success fp <- A.fromJSON _params
     , pred fp
+    , someMethodToMethodString (SomeMethod cm) == "ghcide/reference/ready"
```
michaelpj, line 257: I think we should also be able to get it directly by using the `KnownSymbol` constraint?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -431,304 +414,332 @@ instance PluginRequestMethod TextDocumentCodeAction where
...
         -- This is somewhat vague, but due to the hierarchical nature of action kinds, we
         -- should check whether the requested kind is a *prefix* of the action kind.
         -- That means, for example, we will return actions with kinds `quickfix.import` and
         -- `quickfix.somethingElse` if the requested kind is `quickfix`.
-        , Just caKind <- ca ^. kind = any (\k -> k `codeActionKindSubsumes` caKind) allowed
+        , Just caKind <- ca ^. L.kind = any (\k -> k `codeActionKindSubsumes` caKind) allowed
         | otherwise = False
 
-instance PluginMethod Request TextDocumentDefinition where
+      -- Copied form lsp-types 1.6 to get compilation working. May make more
```
michaelpj, line 433: yes, we should do this, TODO MPJ

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -431,304 +414,332 @@ instance PluginRequestMethod TextDocumentCodeAction where
...
     where
-      r = listToMaybe $ mapMaybe (^. range) hs
-      h = case foldMap (^. contents) hs of
-            HoverContentsMS (List []) -> Nothing
-            hh                        -> Just $ Hover hh r
-
-instance PluginRequestMethod TextDocumentDocumentSymbol where
-  combineResponses _ _ (ClientCapabilities _ tdc _ _ _) params xs = res
+      r = listToMaybe $ mapMaybe (^. L.range) hs
+      -- We are only taking MarkupContent here, because MarkedStrings have been
```
michaelpj, line 571: it would be nice to support both, but not a big deal

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -431,304 +414,332 @@ instance PluginRequestMethod TextDocumentCodeAction where
...
+      dsOrSi :: [Either [SymbolInformation] [DocumentSymbol]]
+      dsOrSi =  toEither <$> mapMaybe nullToMaybe' (toList xs)
+      res :: [SymbolInformation] |? ([DocumentSymbol] |? Null)
       res
-        | supportsHierarchy = InL $ sconcat $ fmap (either id (fmap siToDs)) dsOrSi
-        | otherwise = InR $ sconcat $ fmap (either (List . concatMap dsToSi) id) dsOrSi
-      siToDs (SymbolInformation name kind _tags dep (Location _uri range) cont)
+        | supportsHierarchy = InR $ InL $ concatMap (either (fmap siToDs) id) dsOrSi
+        | otherwise = InL $ concatMap (either id ( concatMap dsToSi)) dsOrSi
+      -- Is this actually a good conversion? It's what there was before, but some
```
michaelpj, line 587: no, these are all kind of dodgy, and I'd like to move functions for a lot of these conversions to lsp and test them

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -956,7 +967,7 @@ mkLspCommand :: PluginId -> CommandId -> T.Text -> Maybe [Value] -> Command
 mkLspCommand plid cn title args' = Command title cmdId args
   where
     cmdId = mkLspCmdId pROCESS_ID plid cn
-    args = List <$> args'
+    args = args'
```
michaelpj, line 970: inline

plugins/hls-call-hierarchy-plugin/src/Ide/Plugin/CallHierarchy/Internal.hs
```diff
@@ -115,7 +116,9 @@ construct nfp hf (ident, contexts, ssp)
         renderSpan _           = ssp
 
         -- https://github.com/haskell/lsp/blob/e11b7c09658610f6d815d04db08a64e7cf6b4467/lsp-types/src/Language/LSP/Types/DocumentSymbol.hs#L97
-        skUnknown = SkUnknown 27 -- 27 is the first unused number while ToJSON
+        -- There is no longer an unknown symbol, thus using SymbolKind_TypeParameter
```
michaelpj, line 119: Given it's a call hierarchy I'd be inclined to use "function" for an unknown item in the hierachy?

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -18,7 +18,7 @@ import           Control.Monad.IO.Class               (MonadIO, liftIO)
 import           Control.Monad.Trans.Except           (ExceptT)
```
michaelpj, line 18: missing a cabal file edit?

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -98,23 +99,4 @@ rules recorder plId = do
...
-            { _range = realSrcSpanToRange observationSrcSpan,
-              _severity = Just LSP.DsHint,
-              _code = Just (LSP.InR $ unId (inspectionId inspection)),
-              _source = Just "stan",
-              _message = message,
-              _relatedInformation = Nothing,
-              _tags = Nothing
-            }
-          )
+              [ "
```
michaelpj, line 102: ?

michaelpj (commented):
Looks great, mostly just nitpicks!

joyfulmantis:
For why the tests are failing: I believe most of the errors are because TextDocumentContentChangeEvent isn't being processed properly. I've looked at the changeDoc (from lsp: lsp-test/src/Language/LSP/Test.hs:474) execution path, and the server side where we process the TextDocumentDidChange notification (from hls: ghcide/src/Development/IDE/LSP/Notifications.hs:70), and the diff for those sections between the new and old hls&lsp and I didn't see anything incorrect.
For the first failing test, I rewrote it as a TextDocumentContentChangeEvent without a range, and that caused the test to succeed. I'll probably separate that out into one test to make it easier to test by itself. 
I've looked over the debug log and the messages between the server and the client and don't see anything obviously wrong.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
joyfulmantis: Thanks, don't know how I missed that

ghcide/src/Development/IDE/Core/PositionMapping.hs
```diff
@@ -120,10 +123,12 @@ mkDelta cs = foldl' applyChange idDelta cs
 addDelta :: PositionDelta -> PositionMapping -> PositionMapping
 addDelta delta (PositionMapping pm) = PositionMapping (composeDelta delta pm)
 
+-- TODO: We currently ignore the right hand side (if there is only text), as
```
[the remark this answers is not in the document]
joyfulmantis: Diffing between the changes, it seems like I am keeping the original functionality. The idea of a TextDocumentContentChangeEvent without a range is to replace the whole document's text, so updating the PositionMappings would be useless. I would imagine? that they are invalidated somewhere else if this type of TextDocumentContentChangeEvent was recieved, but that may just be wishful thinking.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
joyfulmantis: Originally it was a candidate for a lsp package function, however it's only ever used this once, so I figured it's best just to leave it here.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
joyfulmantis: Yeah actually, this would be a nice place to create an instance for Default, and here just add the specific fields we are using.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: It compiles, so I guess some package is reexporting it for us? Tbh I kind of hate this rexporting business, makes it really unclear what package is providing what.

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: The reason why I didn't incude it in the helper instances and functions pr is because it is only ever used this once.

ghcide-bench/src/Experiments.hs
```diff
@@ -708,8 +708,8 @@ searchSymbol doc@TextDocumentIdentifier{_uri} fileContents pos = do
       checkDefinitions pos = do
         defs <- getDefinitions doc pos
         case defs of
-            (InL [Location uri _]) -> return $ uri /= _uri
-            _                      -> return False
+            (InL (Definition (InR [Location uri _]))) -> return $ uri /= _uri
+            _                                         -> return False
```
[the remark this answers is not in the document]
joyfulmantis: For here and now, yes. This sends a TextDocmuentDefinition request, and we handle it in (ghcide/src/Development/IDE/LSP/HoverDefinition.hs:31). It's a litle fragile handling it like this, but it should also be immediently obvious when the situation changes.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
michaelpj: Yeah maybe, although maybe `emptyCompletionItem :: Text -> CompletionItem` that at least sets the non-optional label field is better.

Mostly I just think these massive positional constructor applications are hard to understand or update.

hls-plugin-api/src/Ide/Types.hs
michaelpj: yeah, but logically it's part of LSP and something other people might want

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
michaelpj: Sorry, I didn't mean on this line, I just meant that in general there were no edits to the overloaded-record-dot plugin whereas there are to all the others

joyfulmantis: Oh right, I completely forgot about bumping the version for this one.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
joyfulmantis: Yes absolutely, I think this area that has the highest chance of containing a bug with the new hls and lsp is the constructor applications. Fields were switched with the new generated version, and all it needs is 2 fields with the same type to switch to get a bunch of errors that could be in untouched code.

joyfulmantis:
@guibou with commit [8cb740f](https://github.com/haskell/haskell-language-server/pull/3621/commits/8cb740fc32f8c07e03e1f0806b85b8304b13751f), the new hls should build on nix now. Would love for you to check it though, as not sure it's done the right way.

joyfulmantis:
@everyone This builds and all the tests pass. Would love some more eyes on the changes though, to make sure I didn't let anything nasty slip through.

michaelpj (approved):
TBH, this is a tough one to review. I read it all, but I'm not sure others will have the time to. I think I'm inclined to merge it and let people refer back if they want or if there are problems.

PR #3629, opened 2023-06-08
hls-notes-plugin: Initial implementation

jvanbruegge:
A plugin to expand hls' jump-to-definition to work for GHC-style notes.

TODO:
- [x] Support notes in the same module
- [x] Support notes in other modules
- [x] Search for notes at index time, not in the handler
- [x] Be more liberal with the format of note references/definitions

### DIFF
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -1182,6 +1182,7 @@ installSigUsr1Handler h = void $ installHandler sigUSR1 (Catch h) Nothing
 #endif
 
 {- Note [Resolve in PluginHandlers]
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   Resolve methods have a few guarantees that need to be made by HLS,
   specifically they need to only be called once, as neither their errors nor
   their responses can be easily combined. Whereas commands, which similarly have
--- a/plugins/hls-notes-plugin/README.md
+++ b/plugins/hls-notes-plugin/README.md
@@ -11,6 +11,8 @@ module Main where
 main :: IO
 main = do
     doSomething -- We need this here, see Note [Do Something] in Foo
+                -- Using at-signs around the note works as well:
+                -- see @Note [Do Something]@ in Foo
 ```
 
 Foo.hs
--- a/plugins/hls-notes-plugin/README.md
+++ b/plugins/hls-notes-plugin/README.md
@@ -16,7 +16,7 @@ main = do
 ```
 
 Foo.hs
-```
+```haskell
 module Foo where
 
 doSomething :: IO ()
@@ -29,4 +29,4 @@ Some very important explanation
 -}
 ```
 
-Using "Go-to-definition on the Note reference in `Main.hs` will jump to the beginning of the note in `Foo.hs`.
+Using "Go-to-definition" on the Note reference in `Main.hs` will jump to the beginning of the note in `Foo.hs`.
--- a/plugins/hls-notes-plugin/README.md
+++ b/plugins/hls-notes-plugin/README.md
@@ -1,0 +1,30 @@
+# Note plugin
+
+The [Note convention](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/coding-style#2-using-notes) is a nice way to hoist and share big chunks of documentation out of the body of functions. This is done by referencing a long form note from within the function. This plugin extends goto-definition to jump from the reference to the note.
+
+# Example
+
+Main.hs
+```haskell
+module Main where
+
+main :: IO
+main = do
+    doSomething -- We need this here, see Note [Do Something] in Foo
+```
+
+Foo.hs
+```
+module Foo where
+
+doSomething :: IO ()
+doSomething = undefined
+
+{-
+Note [Do Something]
+~~~~~~~~~~~~~~~~~~~
+Some very important explanation
+-}
+```
+
+Using "Go-to-definition on the Note reference in `Main.hs` will jump to the beginning of the note in `Foo.hs`.
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -137,7 +137,7 @@ findNotesInFile file recorder = do
 noteRefRegex, noteRegex :: Regex
 (noteRefRegex, noteRegex) =
     ( mkReg ("note \\[(.+)\\]" :: String)
-    , mkReg ("note \\[([[:print:]]+)\\][[:blank:]]*[[:space:]][[:blank:]]*(--)?[[:blank:]]*~~~" :: String)
+    , mkReg ("note \\[([[:print:]]+)\\][[:blank:]]*\r?\n[[:blank:]]*(--)?[[:blank:]]*~~~" :: String)
     )
     where
         mkReg = makeRegexOpts (defaultCompOpt { caseSensitive = False }) defaultExecOpt
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -128,7 +128,7 @@ findNotesInFile file recorder = do
             ((x@(name, (char, _)):xs, m), (n, nc, c)) -> \char' ->
                 let !c' = c + 1
                     (!n', !nc') = if char' == '\n' then (n + 1, c') else (n, nc)
-                    p = if char == c then
+                    p@(!_, !_) = if char == c then
                             (xs, HM.insert name (Position (uint n') (uint (char - nc'))) m)
                         else (x:xs, m)
                 in (p, (n', nc', c'))
@@ -137,7 +137,7 @@ findNotesInFile file recorder = do
 noteRefRegex, noteRegex :: Regex
 (noteRefRegex, noteRegex) =
     ( mkReg ("note \\[(.+)\\]" :: String)
-    , mkReg ("note \\[([[:print:]]+)\\][[:blank:]]*[[:space:]][[:space:]]?~~~" :: String)
+    , mkReg ("note \\[([[:print:]]+)\\][[:blank:]]*[[:space:]](--)?[[:blank:]]*~~~" :: String)
     )
     where
         mkReg = makeRegexOpts (defaultCompOpt { caseSensitive = False }) defaultExecOpt
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -87,7 +87,7 @@ jumpToNote state _ param
         note <- err "No note at this position" $ listToMaybe $
             mapMaybe (atPos $ fromIntegral c) $ matchAllText noteRefRegex line
         notes <- runActionE "notes.definedNotes" state $ useE MkGetNotes nfp
-        (noteFp, pos) <- err "Note not found" (HM.lookup note notes)
+        (noteFp, pos) <- err ("Note definition (a comment of the form `{- Note [" <> note <> "]\\n~~~ ... -}`) not found") (HM.lookup note notes)
         pure $ InL (Definition (InL
                 (Location (fromNormalizedUri $ normalizedFilePathToUri noteFp) (Range pos pos))
             ))
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -137,7 +137,7 @@ findNotesInFile file recorder = do
 noteRefRegex, noteRegex :: Regex
 (noteRefRegex, noteRegex) =
     ( mkReg ("note \\[(.+)\\]" :: String)
-    , mkReg ("note \\[([[:print:]]+)\\][[:blank:]]*[[:space:]](--)?[[:blank:]]*~~~" :: String)
+    , mkReg ("note \\[([[:print:]]+)\\][[:blank:]]*[[:space:]][[:blank:]]*(--)?[[:blank:]]*~~~" :: String)
     )
     where
         mkReg = makeRegexOpts (defaultCompOpt { caseSensitive = False }) defaultExecOpt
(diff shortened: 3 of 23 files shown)


plugins/hls-notes-plugin/src/Ide/Plugin/Notes/Internal.hs
```diff
@@ -0,0 +1,89 @@
...
+                                             matchAllText)
+import           Text.Regex.TDFA.Text       (Regex)
+
+jumpToNote :: PluginMethodHandler IdeState TextDocumentDefinition
+jumpToNote state _ param = pluginResponse $ do
+    let uriOrig = param ^. (L.textDocument . L.uri)
+    nfp <- getNormalizedFilePath uriOrig
+
+    file <- liftIO $ runAction "Notes.fileContents" state (getSourceFileSource nfp)
+    let lines = BS.split '\n' file
```
michaelpj, line 35: There's a bunch of "linear in the length of the file" stuff in here. That seems unavoidable, since you're searching files for notes, but we _definitely_ don't want to do this in a handler like this otherwise things will get quite bad on big files.

The thing to do is to define a rule that finds all the note definitions in the file and stores them in a structure that then permits quick lookup. That way we can run and cache it separately from the handler, and the handler can be fast. You'll also definitely need something like this to get definitions from other files. 

You can find lots of examples of this pattern in other plugins.

plugins/hls-notes-plugin/src/Ide/Plugin/Notes/Internal.hs
```diff
@@ -0,0 +1,89 @@
...
+    (name, mod) <- case getNote (param ^. L.position) lines of
+        Just x  -> pure x
+        Nothing -> throwPluginError "No note under cursor"
+
+    let (pos, uri) = case mod of
+            Just _  -> error "not implemented"
```
michaelpj, line 41:
```suggestion
            Just _  -> throwPluginError "not implemented"
```
just never :)

plugins/hls-notes-plugin/src/Ide/Plugin/Notes/Internal.hs
```diff
@@ -0,0 +1,89 @@
...
+    line <- decode <$> safeIndex bss l
+    let matches = A.elems <$> matchAllText noteRefRegex line
+        c' = fromInteger (fromIntegral c)
+    listToMaybe $ mapMaybe (\case
+        ((_, (s, len)):(name, _):_:(mod, _):_) -> if s <= c' && c' <= s + len then
+            Just (name, if mod == "" then Nothing else Just mod)
+            else Nothing
+        _ -> Nothing) matches
+
+noteRefRegex, noteRegex :: Regex
```
michaelpj, line 78: I'd like to see some tests for these functions - tricky regexes are always annoying to reason through and easy to accidentally break!

plugins/hls-notes-plugin/src/Ide/Plugin/Notes/Internal.hs
```diff
@@ -0,0 +1,89 @@
...
+(noteRefRegex, noteRegex) =
+    ( mkReg ("see note \\[(.+)\\]( in (([A-Za-z0-9]+\\.)*[A-Za-z0-9]+))?" :: String)
+    , mkReg ("Note \\[(.+)\\][[:space:]]*~+" :: String)
+    )
+    where
+        mkReg = makeRegexOpts (defaultCompOpt { caseSensitive = False }) defaultExecOpt
+
+safeIndex :: [a] -> UInt -> Maybe a
```
michaelpj, line 86: https://hackage.haskell.org/package/lens-5.2.2/docs/Control-Lens-Fold.html#v:-94--63- perhaps

plugins/hls-notes-plugin/src/Ide/Plugin/Notes/Internal.hs
```diff
@@ -0,0 +1,89 @@
...
+                                             defaultCompOpt, makeRegexOpts,
+                                             matchAllText)
+import           Text.Regex.TDFA.Text       (Regex)
+
+jumpToNote :: PluginMethodHandler IdeState TextDocumentDefinition
+jumpToNote state _ param = pluginResponse $ do
+    let uriOrig = param ^. (L.textDocument . L.uri)
+    nfp <- getNormalizedFilePath uriOrig
+
+    file <- liftIO $ runAction "Notes.fileContents" state (getSourceFileSource nfp)
```
michaelpj, line 34: The VFS internally stores things as a rope, which is a line-oriented data structure. You _might_ want to try and get that instead, it would at least let you avoid doing the line splitting yourself.

michaelpj: Also I think at least you want `getFileContents`, which gives you `Text` rather than a `ByteString`, which is probably what you should be operating on here!

michaelpj (commented):
Looks cool! Maybe we can even do completions, although that would be ambitious :)

jvanbruegge:
Hi, I am trying to finish this up, but I have trouble making progress. The current approach does not work because it only takes into account the files that are open in the editor, which is really useless for jump-to-definition. So at ZuriHac after talking to Michael we decided to search for notes in all files on startup and save their locations.

I've been trying to do that now, but I can't get HLS to give me all the files in the project. I found `GetKnownFiles` and tried to use that from within a `pluginRules`, but that always returns an empty list if I try to run my unit test. Does anyone have an idea what I am doing wrong? I pushed my WIP code in case that helps

wz1000:
Your unit test needs to list all the modules in the arguments section of the `hie.yaml` file for `GetKnownFiles` to work.

jvanbruegge:
Thanks, I will try that later

jvanbruegge:
Ok, I think this is done now. The nix build failures do not seem related. I tried this on the GHC codebase and it does work:
Demo video

jvanbruegge:
@michaelpj is there anything left to do here?

jvanbruegge:
I've rebased on the current master now, can this be merged now? @michaelpj

jvanbruegge:
The test failures are unrelated, the nix job fails because of fourmoulu, and the other test failures are twice in ghcide and once in hls-splice-plugin

fendor:
I restarted the testsuites, sorry for the delay in responses!

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -568,11 +568,27 @@ instance PluginRequestMethod Method_CodeActionResolve where
     -- See Note [Resolve in PluginHandlers].
     combineResponses _ _ _ _ (x :| _) = x
 
+combineDefinitions :: NonEmpty (Definition |? ([DefinitionLink] |? Null)) -> (Definition |? ([DefinitionLink] |? Null))
```
fendor, line 571: Is there specific reasoning for how this is defined? If yes, a comment would be nice.

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -0,0 +1,125 @@
...
+
+instance Pretty Log where
+    pretty = \case
+        LogShake l -> pretty l
+        LogNotesFound file notes ->
+            "Found notes in " <> pretty (show file) <> ": ["
+            <> pretty (intercalate ", " (fmap (\(s, p) -> "\"" <> s <> "\" at " <> T.pack (show p)) notes)) <> "]"
+
+descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
+descriptor recorder plId = (defaultPluginDescriptor plId)
```
fendor, line 58: Could we have some high-level documentation on how this plugin works? 
Something of the like "whenever a session is initialised, we run on all files of the component an index step for the notes available. On Goto Definition request, we look up the position"

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -0,0 +1,125 @@
...
+        (noteFp, pos) <- err "Note not found" (HM.lookup note notes)
+        pure $ InL (Definition (InL
+                (Location (fromNormalizedUri $ normalizedFilePathToUri noteFp) (Range pos pos))
+            ))
+    where
+        uriOrig = toNormalizedUri $ param ^. (L.textDocument . L.uri)
+        err s = maybe (throwError $ PluginInternalError s) pure
+        atPos c arr = case arr A.! 0 of
+            (_, (c', len)) -> if c' <= c && c <= c' + len
+                then Just (fst (arr A.! 1)) else Nothing
```
fendor, line 93: This seems a little fishy to me, could you elaborate?

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -0,0 +1,125 @@
...
+jumpToNote _ _ _ = throwError $ PluginInternalError "conversion to normalized file path failed"
+
+findNotesInFile :: NormalizedFilePath -> Recorder (WithPriority Log) -> Action (Maybe (HM.HashMap Text Position))
+findNotesInFile file recorder = do
+    contentOpt <- (snd =<<) <$> use GetFileContents file
+    content <- case contentOpt of
+        Just x  -> pure x
+        Nothing -> liftIO $ readFileUtf8 $ fromNormalizedFilePath file
```
fendor, line 101: This needs a comment, please!

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -0,0 +1,125 @@
...
+        toPositions matches = snd . fst . T.foldl' (\case
+            (([], m), _) -> const (([], m), (0, 0, 0))
+            ((x@(name, (char, _)):xs, m), (n, nc, c)) -> \char' ->
+                let !c' = c + 1
+                    (!n', !nc') = if char' == '\n' then (n + 1, c') else (n, nc)
+                    p = if char == c then
+                            (xs, HM.insert name (Position (uint n') (uint (char - nc'))) m)
+                        else (x:xs, m)
+                in (p, (n', nc', c'))
+            ) ((matches, HM.empty), (0, 0, 0))
```
fendor, line 117: Comment that explains *what* it is doing.

plugins/hls-notes-plugin/README.md
```diff
@@ -0,0 +1,30 @@
...
+# Example
+
+Main.hs
+```haskell
+module Main where
+
+main :: IO
+main = do
+    doSomething -- We need this here, see Note [Do Something] in Foo
+```
```
fendor, line 14: Seems, both `@Note [Do Something]@` and `Note [Do Something]` are supported, right?

The documentation likely should reflect this!

fendor (changes requested):
This looks really good to me! I only have requests for more documentation which it is currently lacking and some organisational stuff.

Please add this plugin to the CODEOWNERS file and add yourself as a maintainer.
Additionally, you should add the feature list to the documentation, https://github.com/haskell/haskell-language-server/blob/master/docs/features.md#jump-to-definition and add it is a new plugin to the plugin table https://github.com/haskell/haskell-language-server/blob/master/docs/support/plugin-support.md. If you plan to regularly maintain, add tier 2, otherwise tier 3, please.

hls-plugin-api/src/Ide/Types.hs
jvanbruegge: I removed this now as this was already done by https://github.com/haskell/haskell-language-server/pull/3846

jvanbruegge:
@fendor I had time now to fix this up. I added all the comments you requested and added myself to the codeowners file.

jvanbruegge:
I am confused on why the circleCI is failing. Is says that the import of `lift` and `liftIO` are unused, but they are used?

jhrcek:
hls tells me that liftIO is actually re-exported by Control.Monad.Except.
Screenshot from 2024-02-27 20-54-15

Either make export list of this module explicit or remove the unused imports.
The reason why it's only failing in stack job is that this job enables `pedantic: true` in stack.yaml

jvanbruegge:
Ahh, yes. Fixed that now

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -0,0 +1,142 @@
...
+        pure $ Just $ HM.unions definedNotes
+
+jumpToNote :: PluginMethodHandler IdeState Method_TextDocumentDefinition
+jumpToNote state _ param
+    | Just nfp <- uriToNormalizedFilePath uriOrig
+    = do
+        let Position l c = param ^. L.position
+        contents <- fmap _file_text . err "Error getting file contents"
+            =<< lift (LSP.getVirtualFile uriOrig)
+        line <- err "Line not found in file" (Rope.lines contents ^? ix (fromIntegral l))
```
jhrcek, line 85: Rather than getting all rope lines it might be faster to use 
Rope.splitAt twice as it's done e.g. here:
https://github.com/haskell/haskell-language-server/blob/50923e5c790e9c55c2b9b1bfcc8e78f7169f8ea3/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs#L907-L908

jhrcek:
I just tried the plugin on hls codebase and it works in some cases, but is broken in others.
Check for example the `ghcide/src/Development/IDE/Core/Compile.hs` module in hls codebase - even though there's a note "Note [Recompilation avoidance in the presence of TH]"

it gives error whenever I try to go to its definition within the same file (see gif illustrating the issue).
Could you please check what's going on?

NotesNotWorking

jvanbruegge:
I've addressed the review comments, also made the note format more liberal and improved the error message

jvanbruegge:
Also added a commit now that should make all notes in the hls codebase compliant with the plugin. Given that I made the format more liberal this only boiled down to adding underscores to a bunch of notes.

jhrcek (approved):
Thank you!
Looks good to me :+1:

jvanbruegge:
I've also rebased onto the current master now

jvanbruegge:
I've added a `waitForBuildQueue` in the tests now and run it with only 5% CPUQuota to simulate a weaker/overloaded machine. If the tests still fail I am out of ideas

jvanbruegge:
Ok, I found the issue on windows: Line endings. The regex only allowed a single newline character, but on windows it's `\r\n`. So the regex now uses `\r?\n`. The one failing test is a flaky test in `hs-eval-plugin` could someone restart that job?

jhrcek:
Restarted

jvanbruegge:
Great, CI is passing now. @fendor I have done the requested changes earlier already.

jvanbruegge:
I'll go ahead and close this. My motivation to rebase this again because it has been sitting so long is pretty much zero.

PR #3631, opened 2023-06-09
Support fourmolu 0.13.0.0

brandonchinn178:


michaelpj (approved):
CI passes, seems good

georgefst:
Thanks @brandonchinn178! I'm skeptical that we actually still support as far back as 0.3, but we can look in to that next time.

joyfulmantis:
Just changing the version is not enough, fourmolu 0.13 no longer exports FixityMap, so the shim needs to be updated too.

PR #3633, opened 2023-06-10
Simplify selection of GHCs to build on

michaelpj:
The haskell setup action supports using the major version, and then it
will select the latest minor version. That should be fine for us - we
don't have to change anything to get new minor versions, and it's not
too much of a problem to get new minor versions implicitly, since they
should be very unlikely to break things.


fendor:
Then the minor version we install depends on the setup action version?

michaelpj:
Yep. Which I think is fine? We'll still discover any issues deterministically when that gets updated.

fendor (approved):
LGTM after a discussion with @michaelpj

PR #3640, opened 2023-06-10
Split pragmas plugin by providers + decrease disable-warning priority

### LINKED ISSUE #3636: codeAction "Disable <foo-bar> warnings" is the first suggestion
As suggested by @Kleidukos @ ZuriHac 2023: The "Disable <xyz> warnings codeAction should be the last in the list.
Usually, it is not what you want to select.

Change the order of codeActions so that the "Disable this warning" action appears last in the list.

mrcjkb:
Fixes #3636.

### DIFF
--- a/plugins/hls-pragmas-plugin/test/Main.hs
+++ b/plugins/hls-pragmas-plugin/test/Main.hs
@@ -15,8 +15,14 @@ import           Test.Hls.Util           (onlyWorkForGhcVersions)
 main :: IO ()
 main = defaultTestRunner tests
 
-pragmasPlugin :: PluginTestDescriptor ()
-pragmasPlugin = mkPluginTestDescriptor' descriptor "pragmas"
+pragmasSuggestPlugin :: PluginTestDescriptor ()
+pragmasSuggestPlugin = mkPluginTestDescriptor' suggestPragmaDescriptor "pragmas"
+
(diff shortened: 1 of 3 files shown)


plugins/hls-pragmas-plugin/test/Main.hs
```diff
@@ -15,8 +15,13 @@ import           Test.Hls.Util           (onlyWorkForGhcVersions)
 main :: IO ()
 main = defaultTestRunner tests
 
+-- FIXME: Combine these plugins into a single one
+
 pragmasPlugin :: PluginTestDescriptor ()
```
fendor, line 20: I dont think this fixme is necessary :)

mrcjkb: Not sure why CI is succeeding.
The pragma plugin test is failing locally for me, because the "disable warning" code actions are missing.

I don't think combining the plugins is the solution though - I should probably add separate tests for the "disable warnings" descriptor.

mrcjkb:
The current test suite is fixed.

I guess it would be nice to have a test that verifies that the "disable this warning" action appears last, but that would be outside the scope of the `hls-pragmas-plugin` test, since this would have to be tested with multiple plugins.

fendor (approved):

PR #3642, opened 2023-06-10
Add an assist for importing record fields when using OverloadedRecordDot 

simmsb:
When using OverloadedRecordDot to access fields of a record, if the field being accessed is not imported then a constraint error will be emitted.

In ghc 9.6 the error message carries a suggestion to import the field (and so the assist will be generated), however in ghc 9.2 this suggestion is not present and so we don't get any assist.

image


fendor (changes requested):
LGTM, could we have a test for GHC 9.6 and above?

simmsb:
> LGTM, could we have a test for GHC 9.6 and above?

Yup, I'll give it a check tomorrow

michaelpj:
stylish-haskell failed in pre-commit, weirdly?

joyfulmantis:
> stylish-haskell failed in pre-commit, weirdly?

Probably due to cpp stuff. Some of the cpp tags especially for ghc >=9.5 poison stylish-haskell. I think it's usually possible to format cpp that stylish-haskell can accept usually at the cost of brevity, but once there is a poisonous cpp in a file you edit you can either change it yourself and add a bunch of lines to your diff, or commit with --no-verify

fendor (approved):
thank you! LGTM

PR #3643, opened 2023-06-10
Correctly pass VersionedTextDocumentIdentifier through hls

maralorn:
This is an attempt at a continuation of #3566 because I don’t have rights to push to that branch.

Sadly, while I can’t reproduce the concrete error message we are trying to fix here. There are still a lot of files in which hlint fixes don’t get applied for me.

### DIFF
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
@@ -6,9 +6,9 @@
 {-# LANGUAGE OverloadedLabels    #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
-{-# LANGUAGE RecordWildCards     #-}
 
 module Ide.Plugin.Rename (descriptor, E.Log) where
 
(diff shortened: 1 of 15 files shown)


maralorn:
Okay, I can confirm, that the failing edits happen in files with CPP and the error log clearly shows that it is because of the CPP.

maralorn:
I still need to fix compilation with wingman.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -92,8 +94,10 @@ renameProvider state pluginId (RenameParams (TextDocumentIdentifier uri) pos _pr
         -- Perform rename
         let newName = mkTcOcc $ T.unpack newNameText
             filesRefs = collectWith locToUri refs
-            getFileEdit = flip $ getSrcEdit state . replaceRefs newName
-        fileEdits <- mapM (uncurry getFileEdit) filesRefs
+            getFileEdit (uri, locations) = do
+              verTxtDocId <- lift $ getVersionedTextDoc (TextDocumentIdentifier uri)
```
michaelpj, line 98: I think this may well fix the rename plugin to work properly if one of the files that we rename in has been changed in the editor!

plugins/hls-tactics-plugin/old/src/Wingman/AbstractLSP/Types.hs
```diff
@@ -124,6 +124,7 @@ data FileContext = FileContext
   , fc_range    :: Maybe (Tracked 'Current Range)
     -- ^ For code actions, this is 'Just'. For code lenses, you'll get
     -- a 'Nothing' in the request, and a 'Just' in the response.
+  , textVersion     :: TextDocumentVersion
```
michaelpj, line 127: looks like there's a URI field too, so can we just make it a VTDI?

plugins/hls-tactics-plugin/old/src/Wingman/LanguageServer.hs
```diff
@@ -622,12 +622,13 @@ mkWorkspaceEdits
     :: DynFlags
     -> ClientCapabilities
     -> Uri
+    -> TextDocumentVersion
```
michaelpj, line 625: VTDI?

michaelpj (approved):
Looks great!

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
maralorn: Yeah, I had the feeling that I am fixing a bug here.

maralorn:
@michaelpj async reminder to add the "merge me" label or give me rights to do it.

maralorn:
I think there was a transient test error on windows. Can we maybe restart it?

PR #3646, opened 2023-06-11
add outOfScopeOn completion option

oberblastmeister:
This allows someone to disable out of scope completions to improve performance and accuracy.


fendor:
What this aims to do essentially is to only have suggestions for things that are actually in scope. So, no completions from third-party dependencies if the module hasn't been explicitly imported. This, seemingly, makes the IDE experience better for expert users.

The idea is that this is for expert users only, by default, we still suggest items that are not imported right now, but might fit.

michaelpj:
Generally I think we're trying to avoid this kind of configurability? Can we not do something like just putting them at the bottom of the list? Or adding some text to the label text that makes it clear they'll require an import?

fendor:
That is indeed also true.

This is purely about performance I think, so no, just listing them at the bottom is likely not enough.

PR #3658, opened 2023-06-13
Resolve 1: Support for resolve in overloaded-record-dot

joyfulmantis:
In order to better understand how resolve should be implemented, I have started to implement codeAction resolve in overloaded-record-dot. This branch is based off of the new-lsp-release.


plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -135,56 +147,100 @@ instance Pretty RecordSelectorExpr where
...
 codeActionProvider ideState pId (CodeActionParams _ _ caDocId caRange _) =
     pluginResponse $ do
         nfp <- getNormalizedFilePath (caDocId ^. L.uri)
         pragma <- getFirstPragma pId ideState nfp
-        CRSR crsMap exts <- collectRecSelResult ideState nfp
-        let pragmaEdit =
+        caps <- lift getClientCapabilities
+        CRSR crsMap crsDetails exts <- collectRecSelResult ideState nfp
+        let supportsResolve :: Maybe Bool
+            supportsResolve = caps ^? L.textDocument . _Just . L.codeAction . _Just . L.dataSupport . _Just
```
michaelpj, line 197: This is checking for `dataSupport`, which we do need, but we also need `resolveSupport.properties` to contain "edit".

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -135,56 +147,100 @@ instance Pretty RecordSelectorExpr where
...
                 then Nothing
                 else Just $ insertNewPragma pragma OverloadedRecordDot
-            edits crs = convertRecordSelectors crs : maybeToList pragmaEdit
-            changes crs =
-                Just $ Map.singleton (fromNormalizedUri
+            edits (Just crs) = convertRecordSelectors crs : maybeToList pragmaEdit
+            edits _ = []
+            changes crsM crsD =
+                case supportsResolve of
+                    Just False -> Just $ WorkspaceEdit
```
michaelpj, line 206: duplicate code with the resolve handler. This bit is where I feel like we could structure things nicely: we could have one "continuation" that we use in both the resolve and no-resolve cases.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -201,11 +257,15 @@ collectRecSelsRule recorder = define (cmapWithPrio LogShake recorder) $
                 -- the OverloadedRecordDot pragma
                 exts = getEnabledExtensions tmr
                 recSels = mapMaybe (rewriteRange pm) (getRecordSelectors tmr)
+            uniques <- liftIO $ replicateM (length recSels) (hashUnique <$> newUnique)
```
michaelpj, line 260: Is `newUnique` threadsafe? We run handlers in individual threads, so there could well be multiple instances of this running at once.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
michaelpj: There are also some bits that should definitely be in the "continuation" since we don't need them if we're resolving later, like `getFirstPragma`.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -135,56 +147,100 @@ instance Pretty RecordSelectorExpr where
...
+    mkCodeActionTitle exts crsM crsD =
         if OverloadedRecordDot `elem` exts
             then title
             else title <> " (needs extension: OverloadedRecordDot)"
         where
-            title = "Convert `" <> name <> "` to record dot syntax"
-            name = printOutputable se
+            title = "Convert `" <> name (IntMap.lookup crsM crsD) <> "` to record dot syntax"
+            name (Just (RecordSelectorExpr _ se _)) = printOutputable se
+            name _                                  = ""
```
michaelpj, line 243: Maybe just return a `Maybe` and do a `mapMaybe` on line 232?

plugins/hls-overloaded-record-dot-plugin/test/Main.hs
```diff
@@ -28,13 +42,36 @@ test = testGroup "overloaded-record-dot"
     mkTest "MultilineCase" "MultilineCase" "name" 10 7 12 15,
     mkTest "Multiline" "Multiline" "name" 10 7 11 15,
     mkTest "MultilineExpanded" "MultilineExpanded" "owner" 28 8 28 19
-  ]
+  ],
+  testGroup
+  "with Resolve"
+  [ mkResolveTest "Simple" "Simple" "name" 10 7 10 15,
```
michaelpj, line 48: I think the idea is that we use the same expected files for both kinds of test since they should have the same result?

plugins/hls-overloaded-record-dot-plugin/test/Main.hs
```diff
@@ -28,13 +42,36 @@ test = testGroup "overloaded-record-dot"
     mkTest "MultilineCase" "MultilineCase" "name" 10 7 12 15,
     mkTest "Multiline" "Multiline" "name" 10 7 11 15,
     mkTest "MultilineExpanded" "MultilineExpanded" "owner" 28 8 28 19
-  ]
+  ],
+  testGroup
+  "with Resolve"
```
michaelpj, line 47: I'd be tempted to reverse the nesting, i.e. define the test case as a group and then do a resolve/non-resolve test inside it. Should let you avoid the duplication of the test specs.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
michaelpj: sketching, could imagine a type signature like this maybe?
```
mkCodeActionHandlerWithResolve ::
  (CodeActionParams -> LspM [CodeAction])
  -> (CodeActionParams -> CodeAction -> LspM CodeAction) -- fill in 'edit'
  -> (CodeActionParams -> CodeAction -> LspM CodeAction) -- fill in some other field if we decide we want to support that
 -> (PluginMethodHandler ideState, PluginMethodHandler ideState)
```

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: Yes https://stackoverflow.com/questions/70515566/when-is-data-unique-thread-safe

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -105,7 +116,8 @@ instance Hashable CollectRecordSelectors
 instance NFData CollectRecordSelectors
 
 data CollectRecordSelectorsResult = CRSR
-    { recordInfos       :: RangeMap RecordSelectorExpr
+    { records           :: RangeMap Int
```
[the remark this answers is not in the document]
joyfulmantis: Yes, haddock is incoming. The int is a unique id, to allow us to offer up a codeAction first, and then resolve the edit with the id.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -135,56 +147,100 @@ instance Pretty RecordSelectorExpr where
 instance NFData RecordSelectorExpr where
   rnf = rwhnf
 
+data ORDResolveData = ORDRD {
```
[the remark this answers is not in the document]
joyfulmantis: At least in this case, the CodeActionParams doesn't make sense. The problem with the codeActionParams is we are going to need to do processing anyways to know whether we can provide the codeAction, and right now it's a title too, so it makes sense to just process once instead of once to present and once to execute

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: Theoretically yes, however, I think currently edit is the only widely supported resolvable field, and everyone that supports something else also supports edit

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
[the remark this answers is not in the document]
michaelpj: Right. I think it definitely makes sense that we'll want a "stateful version" of resolve-based handlers. Maybe we'll also want a stateless one... I think there are some plugins that don't define any of their own rules so don't even have anywhere to put state. But perhaps easier to do a few and then refactor afterwards.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
michaelpj: We should still do it right :) And we definitely need to check that `resolveSupport` is there. A client could have data support but not resolve support! (although that's weird)

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
michaelpj: we might have to worry about contention on the ioref if we end up using this everywhere, but I expect that won't manifest until we're using it a lot. We should remember to try and check when we're nearly done

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -992,3 +1001,55 @@ getProcessID = fromIntegral <$> P.getProcessID
...
+  -> (ideState -> PluginId -> CodeAction -> LspM Config (Either ResponseError CodeAction))
+  -> PluginHandlers ideState
+mkCodeActionHandlerWithResolve codeActionMethod codeResolveMethod =
+    let newCodeActionMethod ideState pid params = runExceptT $
+            do codeActionReturn <- ExceptT $ codeActionMethod ideState pid params
+               caps <- lift getClientCapabilities
+               case codeActionReturn of
+                r@(InR _) -> pure r
+                (InL ls) -> do
+                    let mapper :: (Command |? CodeAction) -> ExceptT ResponseError (LspT Config IO) (Command |? CodeAction)
```
michaelpj, line 1021: put in the where for decluttering?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -992,3 +1001,55 @@ getProcessID = fromIntegral <$> P.getProcessID
...
+            && case caps ^? L.textDocument . _Just . L.codeAction . _Just . L.resolveSupport . _Just of
+                Just row -> "edit" `elem` row .! #properties
+                _        -> False
+          dropData :: CodeAction -> CodeAction
+          dropData ca = ca & L.data_ .~ Nothing
+
+-- |Allow plugins to "own" resolve data, allowing only them to be queried for
+-- the resolve action. This design has added flexibility at the cost of nested
+-- Value types
+data OwnedResolveData = ORD {
```
michaelpj, line 1041: Another option would be something like:
```
data OwnedData a = OD { owner :: PluginId, value : a }
```
and then you implicitly require to/from JSON instances when you use it.

We could then make things a bit more generic, have `pluginOwnsData` or whatever.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -992,3 +1001,55 @@ getProcessID = fromIntegral <$> P.getProcessID
...
+  owner :: PluginId
+, value :: Value
+} deriving (Generic, Show)
+instance ToJSON OwnedResolveData
+instance FromJSON OwnedResolveData
+
+pluginResolverResponsible :: Maybe Value -> PluginDescriptor c -> Bool
+pluginResolverResponsible (Just val) pluginDesc =
```
michaelpj, line 1049: Ah I see, you need to be able to generically deserialize it here, annoying. I guess you could just fish out the `owner` field directly, but that's also awkward.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -404,7 +412,9 @@ instance PluginMethod Request Method_TextDocumentCodeAction where
       uri = msgParams ^. L.textDocument . L.uri
 
 instance PluginMethod Request Method_CodeActionResolve where
-  pluginEnabled _ msgParams pluginDesc config = pluginEnabledConfig plcCodeActionsOn (configForPlugin config pluginDesc)
+  pluginEnabled _ msgParams pluginDesc config =
+    pluginResolverResponsible (msgParams ^. L.data_) pluginDesc
```
michaelpj, line 416: An issue with this is that it assumes that all the resolve implementations are using `OwnedResolveData`, which they might not be.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -992,3 +1001,55 @@ getProcessID = fromIntegral <$> P.getProcessID
 
 installSigUsr1Handler h = void $ installHandler sigUSR1 (Catch h) Nothing
 #endif
+
+-- |When provided with both a codeAction provider and an affiliated codeAction
+-- resolve provider, this function creates a handler that automatically uses
+-- your resolve provider to fill out you original codeAction if the client doesn't
+-- have codeAction resolve support. This means you don't have to check whether
+-- the client supports resolve and act accordingly in your own providers.
+mkCodeActionHandlerWithResolve
```
michaelpj, line 1010: I wonder if we could also fold the handling of the `OwnedResolveData` into this, so that plugins don't have to see it. We could wrap the main handler to take out whatever's in `data`, wrap it in ORD, and send it; and then the second handler could strip it off agian.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -992,3 +1001,55 @@ getProcessID = fromIntegral <$> P.getProcessID
...
+, value :: Value
+} deriving (Generic, Show)
+instance ToJSON OwnedResolveData
+instance FromJSON OwnedResolveData
+
+pluginResolverResponsible :: Maybe Value -> PluginDescriptor c -> Bool
+pluginResolverResponsible (Just val) pluginDesc =
+    case fromJSON val of
+       (Success (ORD o _)) -> pluginId pluginDesc == o
+       _                   -> True -- We want to fail open in case our resolver is not using the ORD type
```
michaelpj, line 1052: Ah good, this seems sensible indeed.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -158,89 +170,65 @@ descriptor :: Recorder (WithPriority Log) -> PluginId
...
+    Just $ WorkspaceEdit
+    { _changes = Just (Map.singleton uri (edits (IntMap.lookup crsM crsD)))
+    , _documentChanges = Nothing
+    , _changeAnnotations = Nothing}
+    where pragmaEdit =
+            if OverloadedRecordDot `elem` exts
+            then Nothing
+            else Just $ insertNewPragma pragma OverloadedRecordDot
+          edits (Just crs) = convertRecordSelectors crs : maybeToList pragmaEdit
+          edits _          = []
```
michaelpj, line 231: This case shouldn't happen and is a bug, right? Can we arrange to log or fail here?

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -158,89 +170,65 @@ descriptor :: Recorder (WithPriority Log) -> PluginId
...
         where
-            title = "Convert `" <> name (IntMap.lookup crsM crsD) <> "` to record dot syntax"
-            name (Just (RecordSelectorExpr _ se _)) = printOutputable se
-            name _                                  = ""
+            title = "Convert `" <> printOutputable se <> "` to record dot syntax"
+
+mkWorkspaceEdit:: Uri -> Int -> IntMap.IntMap RecordSelectorExpr -> [Extension] -> NextPragmaInfo-> Maybe WorkspaceEdit
+mkWorkspaceEdit uri crsM crsD exts pragma =
+    Just $ WorkspaceEdit
+    { _changes = Just (Map.singleton uri (edits (IntMap.lookup crsM crsD)))
```
michaelpj, line 223: I'd probably handle this by pushing the lookup outside of this function and get the caller to deal with the failure.

michaelpj (commented):
Nice, I like where this is going.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: Yes, this is important because we need to respond with a specific error. Completely overlooked it here though.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -105,7 +117,14 @@ instance Hashable CollectRecordSelectors
 instance NFData CollectRecordSelectors
 
 data CollectRecordSelectorsResult = CRSR
-    { recordInfos       :: RangeMap RecordSelectorExpr
+    { -- |We store everything in here that we need to create the unresolved
+      -- codeAction, the range, an uniquely identifiable int, and the expression
+      -- for the selector expression that we use to generate the name
+      records           :: RangeMap (Int, HsExpr (GhcPass 'Renamed))
```
michaelpj, line 123: Why are we not using the actual `Unique` as the key?

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -105,7 +117,14 @@ instance Hashable CollectRecordSelectors
...
+      -- |This is for when we need to fully generate a textEdit. It contains the
+      -- whole expression we are interested in indexed to the unique id we got
+      -- from the previous field
+    , recordInfos       :: IntMap.IntMap RecordSelectorExpr
```
michaelpj, line 127: The logic here is a bit funny to me: for the `HsExpr` we put it in the result of the `RangeMap`, but the `RecordSelectorExpr` is in a separate map! I would expect something like:
```
records :: RangeMap Int
recordInfo :: IntMap (HsExpr, RecordSelectorExpr)
```

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -135,56 +154,83 @@ instance Pretty RecordSelectorExpr where
...
     { pluginHandlers =
-        mkPluginHandler SMethod_TextDocumentCodeAction codeActionProvider
+      mkCodeActionHandlerWithResolve codeActionProvider resolveProvider
     , pluginRules = collectRecSelsRule recorder
     }
 
+resolveProvider :: PluginMethodHandler IdeState 'Method_CodeActionResolve
+resolveProvider ideState pId ca@(CodeAction _ _ _ _ _ _ _ (Just resData)) =
+    pluginResponse $ do
+        case fromJSON resData of
```
michaelpj, line 179: Just flagging that this seems like it's going to be a pretty much universal pattern for resolve handlers: the first thing they're going to do is decode their data from the data field. So we might want to make it part of the generic machinery.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
```diff
@@ -135,56 +154,83 @@ instance Pretty RecordSelectorExpr where
...
+            Success (ORDRD uri int) -> do
+                nfp <- getNormalizedFilePath uri
+                CRSR _ crsDetails exts <- collectRecSelResult ideState nfp
+                pragma <- getFirstPragma pId ideState nfp
+                case IntMap.lookup int crsDetails of
+                    Just rse -> pure $ ca {_edit = mkWorkspaceEdit uri rse exts pragma}
+                    -- We need to throw a content modified error here, but we need fendor's
+                    -- plugin error response pr to make it convenient to use here.
+                    _        -> throwE "Content Modified Error"
```
michaelpj, line 188: Looking at the spec, I'm not sure this is the correct meaning of `ContentModified`. I don't think there's actually anything _wrong_ with the client sending us a document change notification in between asking for code actions and resolving them, even if we can't handle that. Perhaps just `RequestFailed`?

michaelpj (approved):

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: It was mentioned in a GitHub issue that this was the way to do it. https://github.com/microsoft/language-server-protocol/issues/1738

michaelpj: Okay, let's link to that when we do this!

michaelpj: Bonus points: make a PR upstream to clarify in the spec that this is what you should do!

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: That is what I originally had, but then I had to query the map, later on, to extract the selector expression because we need that for the name. Now I put the selector expression, which is the part I need for the name, in the range map, and then with one lookup to the range map I can build the first code action, and then I only need to do one lookup to the int map when I want to actually resolve the code action.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
joyfulmantis: Because we need to serialize the Unique when we put it in the data field, and it doesn't have json instances. I could make orphans for it, but that didn't seem worth it when it's just a wrapper around an integer itself.

michaelpj: Hmm, that's a bit awkward. Worth a comment explaining that, since in theory `Data.Unique` suggests you should only really use it as a hash key.

plugins/hls-overloaded-record-dot-plugin/src/Ide/Plugin/OverloadedRecordDot.hs
michaelpj: Okay, maybe just try to explain that a bit more in the comment? Basically the first bit is what we need for the initial code action, and the second bit is what we need for resolve.

PR #3659, opened 2023-06-14
Draft: Introduce hierarchical error types and error handling for plugins

fendor:
Essentially, infuse the existing `pluginResponse` infrastructure with steroids:

* Provide a simple interface for performing the work you usually want to do
* Allow fine-grained control how errors are finally reported to the client
* Hierarchical error types, somewhat analogous to the logger

### DIFF
--- a/ghcide/src/Development/IDE/Core/PluginUtils.hs
+++ b/ghcide/src/Development/IDE/Core/PluginUtils.hs
@@ -1,0 +1,108 @@
+{-# LANGUAGE GADTs #-}
+module Development.IDE.Core.PluginUtils where
+
+import           Control.Monad.Extra
+import           Control.Monad.IO.Class
+import           Control.Monad.Reader                 (runReaderT)
+import           Control.Monad.Trans.Except
+import           Control.Monad.Trans.Maybe
+import           Data.Either.Extra                    (maybeToEither)
+import           Data.Functor.Identity
+import qualified Data.Text                            as T
+import           Development.IDE.Core.PositionMapping
+import           Development.IDE.Core.Shake           (IdeAction, IdeRule,
+                                                       IdeState (shakeExtras),
+                                                       mkDelayedAction,
+                                                       shakeEnqueue)
+import qualified Development.IDE.Core.Shake           as Shake
+import           Development.IDE.GHC.Orphans          ()
+import           Development.IDE.Graph                hiding (ShakeValue)
+import           Development.IDE.Types.Location       (NormalizedFilePath)
+import qualified Development.IDE.Types.Location       as Location
+import qualified Development.IDE.Types.Logger         as Logger
+import qualified Ide.PluginUtils                      as PluginUtils
+import qualified Language.LSP.Types                   as LSP
+import           Prettyprinter
+import           Prettyprinter.Render.Text            (renderStrict)
+
+-- ----------------------------------------------------------------------------
+-- Plugin Error wrapping
+-- ----------------------------------------------------------------------------
+
+data GhcidePluginError
+    = forall a . Show a => FastRuleNotReady a
+    | forall a . Show a => RuleFailed a
+    | CoreError PluginUtils.PluginError
+
+instance Pretty GhcidePluginError where
+  pretty = \case
+    FastRuleNotReady rule -> "FastRuleNotReady:" <+> viaShow rule
+    RuleFailed rule       -> "RuleFailed:" <+> viaShow rule
+    CoreError perror      -> pretty $ PluginUtils.prettyPluginError perror
+
+pluginResponse ::
+    Monad m =>
+    ExceptT GhcidePluginError m a ->
+    m (Either LSP.ResponseError a)
+pluginResponse = PluginUtils.pluginResponse' handlePluginError
+
+withPluginError :: Functor m => ExceptT PluginUtils.PluginError m a -> ExceptT GhcidePluginError m a
+withPluginError = PluginUtils.withError CoreError
+
+mkPluginErrorMessage :: T.Text -> GhcidePluginError
+mkPluginErrorMessage = CoreError . PluginUtils.mkPluginErrorMessage
+
+handlePluginError :: GhcidePluginError -> LSP.ResponseError
+handlePluginError msg = PluginUtils.mkSimpleResponseError $ renderStrict simpleDoc
+    where
+        simpleDoc = layoutPretty defaultLayoutOptions $ pretty msg
+
+-- ----------------------------------------------------------------------------
+-- Action wrappers
+-- ----------------------------------------------------------------------------
+
+runAction :: MonadIO m => String -> IdeState -> ExceptT e Action a -> ExceptT e m a
+runAction herald ide act =
+  PluginUtils.hoistExceptT . ExceptT $
+    join $ shakeEnqueue (shakeExtras ide) (mkDelayedAction herald Logger.Debug $ runExceptT act)
+
+-- | Request a Rule result, it not available return the last computed result which may be stale.
+--   Errors out if none available.
+useWithStale_ ::(IdeRule k v)
+    => k -> NormalizedFilePath -> ExceptT e Action (v, PositionMapping)
+useWithStale_ key file = ExceptT $ fmap Right $ Shake.useWithStale_ key file
+
+useWithStale :: IdeRule k v
+    => k -> NormalizedFilePath -> ExceptT GhcidePluginError Action (v, PositionMapping)
+useWithStale key file = maybeToExceptT (FastRuleNotReady key) $ useWithStaleMaybeT key file
+
+-- | useE is useful to implement functions that aren’t rules but need shortcircuiting
+-- e.g. getDefinition.
+use :: IdeRule k v => k -> NormalizedFilePath -> ExceptT GhcidePluginError Action v
+use k = maybeToExceptT (RuleFailed k) . MaybeT . Shake.use k
+
+useWithStaleMaybeT :: IdeRule k v
+    => k -> NormalizedFilePath -> MaybeT Action (v, PositionMapping)
+useWithStaleMaybeT key file = MaybeT $ runIdentity <$> Shake.usesWithStale key (Identity file)
+
+-- ----------------------------------------------------------------------------
+-- IdeAction wrappers
+-- ----------------------------------------------------------------------------
+
+runIdeAction :: MonadIO m => String -> Shake.ShakeExtras -> ExceptT e IdeAction a -> ExceptT e m a
+runIdeAction _herald s i = ExceptT $ liftIO $ runReaderT (Shake.runIdeActionT $ runExceptT i) s
+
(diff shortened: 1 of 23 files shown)


ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,46 @@
...
+
+useWithStale :: IdeRule k v
+    => k -> NormalizedFilePath -> MaybeT Action (v, PositionMapping)
+useWithStale key file = MaybeT $ runIdentity <$> Shake.usesWithStale key (Identity file)
+
+useWithStaleE :: IdeRule k v
+    => k -> NormalizedFilePath -> ExceptT PluginError Action (v, PositionMapping)
+useWithStaleE key file = maybeToExceptT PluginTemporarilyUnresponsive $ useWithStale key file
+
+-- | useE is useful to implement functions that aren’t rules but need shortcircuiting
```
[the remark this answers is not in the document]
fendor: Absolutely true! It is a work in progress, trying to pool all the functions we need to improve the plugin error reporting and logging. Names are not final, yet, but I am happy for suggestions!

joyfulmantis:
Yeah, sorry I only saw the DRAFT marking after commenting 😅

fendor ***@***.***> 於 2023年6月14日 週三 17:15 寫道：

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,108 @@
...
+-- IdeAction wrappers
+-- ----------------------------------------------------------------------------
+
+runIdeAction :: MonadIO m => String -> Shake.ShakeExtras -> ExceptT e IdeAction a -> ExceptT e m a
+runIdeAction _herald s i = ExceptT $ liftIO $ runReaderT (Shake.runIdeActionT $ runExceptT i) s
+
+-- | useE is useful to implement functions that aren’t rules but need shortcircuiting
+-- e.g. getDefinition.
+useWithStaleFast :: IdeRule k v => k -> NormalizedFilePath -> ExceptT GhcidePluginError IdeAction (v, PositionMapping)
+useWithStaleFast k = maybeToExceptT (RuleFailed k) . MaybeT . Shake.useWithStaleFast k
```
michaelpj, line 98: I'm not totally sure why this gets a different error than `useWithStale` :thinking: More generally I think it would be good for the errors to match the function names. We can rename the functions... but let's keep the two consistent.

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,108 @@
...
+-- Action wrappers
+-- ----------------------------------------------------------------------------
+
+runAction :: MonadIO m => String -> IdeState -> ExceptT e Action a -> ExceptT e m a
+runAction herald ide act =
+  PluginUtils.hoistExceptT . ExceptT $
+    join $ shakeEnqueue (shakeExtras ide) (mkDelayedAction herald Logger.Debug $ runExceptT act)
+
+-- | Request a Rule result, it not available return the last computed result which may be stale.
+--   Errors out if none available.
```
michaelpj, line 70: Can I request that we not use "errors out", it's very ambiguous when we have both `ExceptT` and exceptions in the picture. Maybe just "throws an exception"?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -263,27 +277,64 @@ allLspCmdIds pid commands = concatMap go commands
...
 
 handleMaybeM :: Monad m => e -> m (Maybe b) -> ExceptT e m b
 handleMaybeM msg act = maybeM (throwE msg) return $ lift act
 
-pluginResponse :: Monad m => ExceptT String m a -> m (Either ResponseError a)
-pluginResponse =
-  fmap (first (\msg -> ResponseError InternalError (fromString msg) Nothing))
-    . runExceptT
+withError :: Functor m => (e' -> e) -> ExceptT e' m a -> ExceptT e m a
+withError = withExceptT
```
michaelpj, line 334: Again, I think I'm broadly in favour of getting people to use "normal" `ExceptT` functions as much as possible.

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,108 @@
...
+useWithStale_ ::(IdeRule k v)
+    => k -> NormalizedFilePath -> ExceptT e Action (v, PositionMapping)
+useWithStale_ key file = ExceptT $ fmap Right $ Shake.useWithStale_ key file
+
+useWithStale :: IdeRule k v
```
michaelpj, line 75: Okay, so serious question: since we're in `Action` here, and it's safe to throw exceptions, is there any reason not to throw these errors as exceptions too? 

That would give us an (I think) sensible heuristic:
- Functions that run in `Action` throw exceptions
- Functions that run outside `Action` use `ExceptT`

michaelpj: That also suggests that the functions here could just become "stricter" versions of `useWithStale` and friends, which throw `RuleNotReady` instead of `RuleFailed` or whatever. Then this module mostly goes away and we instead add a few functions to `Shake.hs`

ghcide/src/Development/IDE/Spans/Pragmas.hs
```diff
@@ -51,13 +52,13 @@ insertNewPragma (NextPragmaInfo nextPragmaLine _) newPragma =  LSP.TextEdit prag
         pragmaInsertPosition = LSP.Position (fromIntegral nextPragmaLine) 0
         pragmaInsertRange = LSP.Range pragmaInsertPosition pragmaInsertPosition
 
-getFirstPragma :: MonadIO m => PluginId -> IdeState -> NormalizedFilePath -> ExceptT String m NextPragmaInfo
-getFirstPragma (PluginId pId) state nfp = handleMaybeM "Could not get NextPragmaInfo" $ do
-  ghcSession <- liftIO $ runAction (T.unpack pId <> ".GhcSession") state $ useWithStale GhcSession nfp
-  (_, fileContents) <- liftIO $ runAction (T.unpack pId <> ".GetFileContents") state $ getFileContents nfp
+getFirstPragma :: MonadIO m => PluginId -> IdeState -> NormalizedFilePath -> ExceptT GhcidePluginError m NextPragmaInfo
```
michaelpj, line 55: If we did what I suggested, this would "just" be in `MonadIO`. I think that's okay: if it's in `MonadIO` it can throw!

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -263,27 +277,64 @@ allLspCmdIds pid commands = concatMap go commands
...
-    where
-        errMsg = T.unpack $ "Failed converting " <> getUri uri <> " to NormalizedFilePath"
 
 -- ---------------------------------------------------------------------
-throwPluginError :: Monad m => String -> ExceptT String m b
-throwPluginError = throwE
+
+type PluginHandler e m a = ExceptT e m a
+
+pluginResponse :: Monad m => ExceptT PluginError m a -> m (Either ResponseError a)
```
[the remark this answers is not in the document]
michaelpj: Hmm, we can't make it completely generic as it is now because of the attempt to fold the `getNormalizedFilePath` functions in...

michaelpj (commented):
I like the direction. At a high level my suggestions are:
- Stuff inside `Action` already throws a bunch. Let's just use exceptions there, and have "strict" versions of the `use` functions that can throw if the thing isn't there.
- Using `ExceptT` for the "main line" of plugin code seems nice. It would be nice if we could make the error type there fully generic so plugins can provide their own. I'm not sure how to do that because of the normalized file path stuff. The only suggestion I have is that we could also use exceptions there, and provide a throwing variant that runs in `MonadIO`. It's not like anyone wants to handle the failures there anyway!

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -263,27 +277,64 @@ allLspCmdIds pid commands = concatMap go commands
 
 -- ---------------------------------------------------------------------
 
-getNormalizedFilePath :: Monad m => Uri -> ExceptT String m NormalizedFilePath
-getNormalizedFilePath uri = handleMaybe errMsg
+getNormalizedFilePath :: Monad m => Uri -> ExceptT PluginError m NormalizedFilePath
```
michaelpj, line 280: So if we did what I'm suggesting, we'd also provide
```
getNormalizedFilePathIO :: MonadIO m => Uri -> m NormalizedFilePath
```

michaelpj:
I realise that what I'm proposing is actually more radical than I maybe thought. It pretty much amounts to:
- All plugin helper functions provide a variant that runs in `MonadIO` and throws on error. Plugins all run in `MonadIO` and have exception catching at the top level, so why not?
- Plugins should just be expected to return `ExceptT ResponseError m (Result...)`. If they want to use `ExceptT` with their own error type and use standard `ExceptT` to translate it to a `ResponseError`, that's pretty much completely up to them! At most we maybe could provide `toResponseError :: Pretty e => e -> ResponseError`

Perhaps this is a bit heretical, I'm basically proposing that we embrace exceptions in plugins. But maybe it would be fine?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -103,28 +104,22 @@ properties = emptyProperties
...
-              $ runAction "codeLens.TypeCheck" ideState (useWithStale TypeCheck nfp)
-              )
-    bindings <- fst <$> (
-                handleMaybeM "Unable to GetBindings"
-                $ liftIO
-                $ runAction "codeLens.GetBindings" ideState (useWithStale GetBindings nfp)
-                )
+    nfp <- PluginUtils.withPluginError $ getNormalizedFilePath uri
+    env <- hscEnv . fst <$>
+      PluginUtils.runAction "codeLens.GhcSession" ideState
```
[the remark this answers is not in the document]
fendor: I did not want to impose Plugin changes for no reason, to keep it easier to review. 
Also, I think some people prefer this style, because otherwise you have to return a big tuple and pattern match on it. I don't think there is any performance impact.

fendor:
I like the idea of embracing exceptions, catching them and providing sensible log messages.
I am going to move the API towards what you are suggesting and let's how the final code looks like.

joyfulmantis:
It would be nice if there was a PluginError type that wrapped ResponseErrors. The reason is that resolve methods will need to throw a ContentModified ResponseError if they are unable to resolve the provided request, and it would be nice to be able to use the available ExceptT pluginResponse infrastructure while at the same time being able to throw specific response errors.

joyfulmantis:
If you end up going ahead with this, I think it might be a good idea to first merge the general infrastructure as a parallel system to what we have now. Then anyone interested in migrating the plugins, can do that piece by piece. Finally once everything is migrated, the old stuff can be removed.

joyfulmantis:
Closing this as #3717 was merged

PR #3661, opened 2023-06-14
Revert "Support fourmolu 0.13.0.0"

joyfulmantis:
Reverts haskell/haskell-language-server#3631 . When fourmolu actually ends up being built with 0.13.0.0 it fails to compile, as the API has changed.


joyfulmantis:
@brandonchinn178 If you still want to support fourmolu 0.13 make a new pull request that incorporates adaptions to the breaking changes from fourmolu 0.13. One of the breaking changes is it no longer exports FixityMap, but there could be others. I would recommend you force cabal to build with 0.13, by temporarily removing the other version options. That way you can make sure it compiles with 0.13.

brandonchinn178:
Yes, I'll fix, sorry about that. Is there a reason CI doesn't build these files?

PR #3662, opened 2023-06-15
Support fourmolu 0.13

brandonchinn178:
Includes the revert from #3661. Feel free to merge into one PR. cc @joyfulmantis 

Tested with the following script:
```bash
(
    set -e
    function test_fourmolu() {
        local constraint="fourmolu$1"
        echo "#### $constraint"
        cabal test \
            -w "$2" \
            --constraint "$constraint" \
            --index-state HEAD \
            hls-fourmolu-plugin
    }
    test_fourmolu '==0.13.0.0' 'ghc-9.6.1'
    test_fourmolu '<0.13' 'ghc-9.6.1'
    test_fourmolu '<0.8.1' 'ghc-9.2.7'
    test_fourmolu '<0.7' 'ghc-9.2.7'
)
```
---
Note: I noticed the tests sometimes flakily fail with
```
Could not access 'plugins/hls-fourmolu-plugin/test/testdata/test/testdata/Fourmolu2.formatted.hs'
```
Not sure what's happening, and why the test directory is repeated.


plugins/hls-fourmolu-plugin/test/Main.hs
```diff
@@ -26,7 +26,7 @@ tests =
         (if cli then "cli" else "lib")
         [ goldenWithFourmolu cli "formats correctly" "Fourmolu" "formatted" $ \doc -> do
             formatDoc doc (FormattingOptions 4 True Nothing Nothing Nothing)
-        , goldenWithFourmolu cli "formats imports correctly" "Fourmolu" "formatted" $ \doc -> do
+        , goldenWithFourmolu cli "formats imports correctly" "Fourmolu2" "formatted" $ \doc -> do
```
brandonchinn178, line 29: This seems to have been testing the wrong file for quite a while:
https://github.com/haskell/haskell-language-server/pull/1836/commits/f6535554815a980d936ad2e1822c8bde4155d9a9#diff-5e783ba6072f1ca27c7171a3ece5043ea46f212e57f68272b52ecc36a8152d59L26-L36

PR #3665, opened 2023-06-16
Add arguments to direct logs to various locations

michaelpj:
This adds arguments to HLS to allow the user to select whether to send logs to any or all of:
- a file
- stderr
- the client

Importantly, we can toggle off the default stderr logging, so the vscode extension can turn it off to avoid the double logging that arises from logging to both the client and stderr.

I've set the default to _not_ log to the client. This is a change of behaviour (today we log to the client by default), but I think it gives the best experience by default, since most clients do show stderr output somewhere, and then we probably want to make a case-by-case decision on whether to use the client logging instead.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -1,22 +1,26 @@
 -- Copyright (c) 2019 The DAML Authors. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
-{-# LANGUAGE NamedFieldPuns    #-}
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 module Main(main) where
 
-import           Control.Arrow                 ((&&&))
+import           Control.Exception             (displayException)
 import           Control.Monad.IO.Class        (liftIO)
+import           Data.Bifunctor                (first)
 import           Data.Function                 ((&))
+import           Data.Functor                  ((<&>))
+import           Data.Maybe                    (catMaybes)
 import           Data.Text                     (Text)
-import qualified Development.IDE.Main          as GhcideMain
 import           Development.IDE.Types.Logger  (Doc, Priority (Error, Info),
+                                                Recorder,
                                                 WithPriority (WithPriority, priority),
                                                 cfilter, cmapWithPrio,
                                                 defaultLayoutOptions,
-                                                layoutPretty,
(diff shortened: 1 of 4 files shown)


exe/Main.hs
```diff
@@ -58,28 +79,35 @@ main = do
...
-                & cmapWithPrio fst
-            , lspMessageRecorder
-                & cfilter (\WithPriority{ priority } -> priority >= Error)
-                & cmapWithPrio (renderDoc . fst)
-            , lspLogRecorder
-                & cfilter (\WithPriority{ priority } -> priority >= minPriority)
-                & cmapWithPrio (renderStrict . layoutPretty defaultLayoutOptions . fst)
-                -- do not log heap stats to the LSP log as they interfere with the
-                -- ability of lsp-test to detect a stuck server in tests and benchmarks
-                & if argsTesting then cfilter (not . heapStats . snd . payload) else id
```
michaelpj, line 81: We don't need this any more since we're not logging to the client by default.

fendor (approved):
LGTM, loving that change

fendor:
Ah, some tests actually use the logs

fendor (approved):
LGTM :)

src/Ide/Arguments.hs
```diff
@@ -138,12 +140,40 @@ arguments plugins = GhcideArguments
           <> help "Sets the log level to Debug, alias for '--log-level Debug'"
           )
         )
-      <*> optional (strOption
-           (long "logfile"
-          <> short 'l'
```
alanz, line 143: I would suggest keeping the short form for at least one release cycle.
I just got bitten by this, and had to re-configure my emacs to use it 

```elisp
  :custom
  (lsp-haskell-server-args `("-d" "--log-file" ,lsp-haskell-server-log-file))
```

michaelpj: Ah sorry, that was an oversight, I meant to leave the old options in! I left the old spelling but missed the short option.

michaelpj: Although I don't think leaving it in for a release cycle helps all that much: it just means that people will be surprised by it one cycle later. It's only useful if you're switching between and old and new version.

michaelpj: (... then why am I leaving the old spelling in? I guess we could just never remove it)

alanz: The one release cycle thing is so you could update clients to use the new one, but if people still ran an older server it would still work. Then in time get rid of it.
To me, keeping the short form is the best option

michaelpj: https://github.com/haskell/haskell-language-server/pull/3672

PR #3674, opened 2023-06-26
Fixes pragma plugin offering incorrect code actions #3673

### LINKED ISSUE #3673: pragma plugin offering incorrect code actions
Currently, the pragma plugin pattern matches on the occurrence of language extensions in the diagnostics to create an "Add <LanguageExtension>" code action. This section that matches on the diagnostic needs to be pickier, as currently, it incorrectly produces code actions for when hlint suggests removing the pragma.

joyfulmantis:
Fixes #3673

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -34,6 +34,8 @@ module Development.IDE.Core.Compile
   , ml_core_file
   , coreFileToLinkable
   , TypecheckHelpers(..)
+  , sourceTypecheck
+  , sourceParser
   ) where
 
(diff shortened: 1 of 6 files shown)


fendor (approved):
LGTM, thanks!

could we maybe create a constant in ghcide for these strings? (Assuming they come from ghcide)

joyfulmantis:
> LGTM, thanks!
> 
> could we maybe create a constant in ghcide for these strings? (Assuming they come from ghcide)

Yes, they all come from gchide, strewn throughout the code. A constant would certainly be wise. I'll add it to this pr before merging.

fendor (approved):
LGTM, thank you!

PR #3676, opened 2023-06-27
Remove extra call to newHscEnvEqWithImportPaths

nlander:
Submitting this at the request of @wz1000 
Instead of calling `newHscEnvEqWithImportPaths` again in `ghcSessionDepsDefinition`, we only generate a new `Unique` for the `HscEnvEq` and update the `envUnique` field.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -100,6 +100,7 @@ import qualified Data.Text.Encoding                           as T
 import           Data.Time                                    (UTCTime (..))
 import           Data.Tuple.Extra
 import           Data.Typeable                                (cast)
+import qualified Data.Unique                                  as Unique
 import           Development.IDE.Core.Compile
 import           Development.IDE.Core.FileExists hiding (LogShake, Log)
 import           Development.IDE.Core.FileStore               (getFileContents,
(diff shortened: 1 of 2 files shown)


nlander:
We also update the `hscEnv` field.

fendor (approved):
LGTM!

michaelpj:
Worth a comment explaining what's going on?

wz1000:
`newHscEnvEqWithImportPaths` recreates the `ExportsMap` for package dependencies on every call, we probably only want to do that once per session instead of once per call to `ghcSessionDepsDefinition` (which is called for every single file we need to compile).

michaelpj:
put it in the code!

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -799,7 +799,12 @@ ghcSessionDepsDefinition fullModSummary GhcSessionDepsConfig{..} env file = do
 #endif
             session' <- liftIO $ mergeEnvs hsc moduleNode inLoadOrder depSessions
 
-            Just <$> liftIO (newHscEnvEqWithImportPaths (envImportPaths env) session' [])
+            -- Here we avoid a call to to `newHscEnvEqWithImportPaths`, which creates a new
```
nlander, line 802: @michaelpj Does this sufficiently explain what's going on here? Is there anything we should add?

michaelpj: lgtm

michaelpj (approved):

ghcide/src/Development/IDE/Core/Rules.hs
michaelpj: I would say ask yourself:
1. If I was reading this for the first time, would I know what is going on and why?
2. Has this captured the knowledge that I gained in order to make the fix?

PR #3678, opened 2023-06-27
Generic support for resolve in hls packages

joyfulmantis:
This is part of the larger resolve support pr group, and all of the other resolve pr's depend upon this one


hls-plugin-api/src/Ide/Types.hs
```diff
@@ -983,3 +1012,107 @@ getProcessID = fromIntegral <$> P.getProcessID
 
 installSigUsr1Handler h = void $ installHandler sigUSR1 (Catch h) Nothing
 #endif
+
+-- |When provided with both a codeAction provider and an affiliated codeAction
```
michaelpj, line 1016: Not for this PR, but this module is pretty big and could probably be split up.

michaelpj (approved):

PR #3679, opened 2023-06-27
Resolve 2: Support for resolve in hls-hlint-plugin

joyfulmantis:
Commands and resolve act very similarly, and since hlint already heavily uses commands, I am choosing to first move everything to use commands, and then offer resolve if it's available, falling back to commands for clients that don't support it.


michaelpj (approved):

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1051,30 +1054,48 @@ mkCodeActionHandlerWithResolve codeActionMethod codeResolveMethod =
...
+                         | otherwise -> pure $ InL $ moveDataToCommand <$> ls
         newCodeResolveMethod ideState pid params =
             codeResolveMethod ideState pid (unwrapCodeActionResolveData params)
-    in mkPluginHandler SMethod_TextDocumentCodeAction newCodeActionMethod
-    <> mkPluginHandler SMethod_CodeActionResolve newCodeResolveMethod
-  where dropData :: Command |? CodeAction -> Command |? CodeAction
-        dropData ca = ca & _R . L.data_ .~ Nothing
-        dropCommands :: Command |? CodeAction -> Command |? CodeAction
-        dropCommands ca = ca & _R . L.command .~ Nothing
+    in ([PluginCommand "codeActionResolve" "Executes resolve for code action" (executeResolveCmd plId codeResolveMethod)],
```
michaelpj, line 1077: Hmm, I'm not sure if this will work? This seems to use the same command ID for all the resolve commands. I'm not sure what will happen then! Could we disambiguate the command ID using the plugin ID?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1051,30 +1054,48 @@ mkCodeActionHandlerWithResolve codeActionMethod codeResolveMethod =
...
+          let dat = toJSON <$> ca ^? _R -- We need to take the whole codeAction
+              -- And put it in the argument for the Command, that way we can later
+              -- pas it to the resolve handler (which expects a whole code action)
+              cmd = mkLspCommand plId (CommandId "codeActionResolve") "Execute Code Action" (pure <$> dat)
+          in ca
+              & _R . L.data_ .~ Nothing -- Set the data field to nothing
+              & _R . L.command ?~ cmd -- And set the command to our previously created command
+        executeResolveCmd :: PluginId -> PluginMethodHandler ideState Method_CodeActionResolve -> CommandFunction ideState CodeAction
+        executeResolveCmd pluginId resolveProvider ideState ca =  do
+          withIndefiniteProgress "Executing code action..." Cancellable $ do
```
michaelpj, line 1091: "Applying edits for code action"?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1051,30 +1054,48 @@ mkCodeActionHandlerWithResolve codeActionMethod codeResolveMethod =
...
+            resolveResult <- resolveProvider ideState pluginId ca
+            case resolveResult of
+              Right CodeAction {_edit = Just wedits } -> do
```
michaelpj, line 1094: What happens if it has the `command` set? It probably shouldn't, right? But maybe we should fail or at least log there.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1051,30 +1054,48 @@ mkCodeActionHandlerWithResolve codeActionMethod codeResolveMethod =
...
+                  _ <- sendRequest SMethod_WorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing wedits) (\_ -> pure ())
```
michaelpj, line 1095: It seems like we probably should care a little about the response? e.g. we probably want to do something if we fail, like at least log

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1051,30 +1054,48 @@ mkCodeActionHandlerWithResolve codeActionMethod codeResolveMethod =
...
                 r@(InR Null) -> pure r
                 (InL ls) | -- If the client supports resolve, we will wrap the resolve data in a owned
                            -- resolve data type to allow the server to know who to send the resolve request to
-                           -- and dump the command fields.
                            supportsCodeActionResolve caps ->
-                            pure $ InL (dropCommands . wrapCodeActionResolveData pid <$> ls)
-                           -- If they do not we will drop the data field.
-                         | otherwise -> pure $ InL $ dropData <$> ls
+                            pure $ InL (wrapCodeActionResolveData pid <$> ls)
+                           -- If they do not we will drop the data field, in addition we will populate the command
```
michaelpj, line 1072: I think it would be nice to have a longer comment or note explaining the overall flow here, perhaps something like:
```
{- Note [Code action resolve fallback to commands]
To make supporting code action resolve easy for plugins, we want to let them
provide one implementation that can be used both when clients support
resolve, and when they don't.

The way we do this is to have them always implement a resolve handler.
Then, if the client doesn't support resolve, we instead install the resolve
handler as a _command_ handler, passing the code action literal itself
as the command argument. This allows the command handler to have
the same interface as the resolve handler!
-}
```

hls-plugin-api/src/Ide/Types.hs
michaelpj: Also, if it's cheap it would be nice to check that all the other fields of the code action are the same as the input one, i.e. the resolve handler only gave us an edit. In the more generalized case, we should check that the only fields that differ are the ones that the client says it can resolve. Then we can log if the handler does something wrong there.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -519,28 +508,32 @@ mkSuppressHintTextEdits dynFlags fileContents hint =
...
 
 -- ---------------------------------------------------------------------
-
-data ApplyOneParams = AOP
-  { verTxtDocId :: VersionedTextDocumentIdentifier
-  , start_pos   :: Position
-  -- | There can be more than one hint suggested at the same position, so HintTitle is used to distinguish between them.
-  , hintTitle   :: HintTitle
-  } deriving (Eq,Show,Generic,FromJSON,ToJSON)
+data HlintResolveCommands = AA { verTxtDocId :: VersionedTextDocumentIdentifier}
```
michaelpj, line 528: I think it's now not obvious what these constructors correspond to, i.e. the string "apply all" has gone from the file!

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -519,28 +508,32 @@ mkSuppressHintTextEdits dynFlags fileContents hint =
...
+                          | AO { verTxtDocId :: VersionedTextDocumentIdentifier
+                               , start_pos   :: Position
```
michaelpj, line 530: Could just put a `OneHint` in here? Or even merge AO and AA, since the function that consumes them is the same, and have a `Maybe (OneHint)` field?

michaelpj (commented):
Sorry I didn't get to this sooner! I had a few thoughts. Generally I really like the approach with the command!

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: I believe it already "magically" does this for us

michaelpj: Ah yes, you're right. Could you add some haddock maybe?

hls-plugin-api/src/Ide/Types.hs
michaelpj: Well, I guess also the command and data fields should be unset?

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: That's quite hard here, as we can't use a recorder (That's defined in the ghcide package, so cyclic dependency), and we only have a callback to work with.

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: I think, according to the spec, we can only change the set of resolvable fields, so even though it would make sense to dump the data field, it still needs to stay the same as when we received it. (We also don't have the client capabilities so can't do anything with that). Finally in this case the only thing we can execute on is the returned WorkspaceEdit. We can return an error if any other field is changed, but not sure about the performance penalty of doing that.

hls-plugin-api/src/Ide/Types.hs
michaelpj: Ugh. Okay, I guess we can't do much here, but that's rather unsatisfying. Maybe we need to move some of the recorder stuff into `hls-plugin-api` as the "lowest" package.

michaelpj: I would really like to be able to do some logging in the generic functions you're defining, that way we can get somewhat-decent logging for lots of plugins by defining it in one place.

hls-plugin-api/src/Ide/Types.hs
michaelpj: Can't we pass in the client caps here?

I don't know if it's worth doing the checking of field setting, it just seems like a way to sanity check plugins, which seems somewhat useful.

joyfulmantis: Okay, I guess we can't pass in the client capabilities here, but we should be able to pull them from the monad itself, since the plugin handler is effectively in the Lsp monad already

michaelpj: Ah I see, this is exactly the sort of thing we could do if we had the client caps when making the handlers!

PR #3680, opened 2023-06-28
Hls 2.0.0.1 forward port

hasufell:


RELEASING.md
```diff
@@ -0,0 +1,31 @@
+# Releasing
+
+## Release checklist
```
fendor, line 3: I love this addition!

This has a big overlap with https://github.com/haskell/haskell-language-server/blob/master/docs/contributing/releases.md, we should probably merge these documents.

hasufell: Possibly... but the other document is old and I've never looked at it.

My suggestion is to have another contributor do a release while I'm available and ask questions when anything in the checklist is unclear... and then we can add more information as we go.

michaelpj: An acceptable merge operation is "discard the old one" :)

fendor: The old list is at least partially up-to-date and contains a lot of detailed information, I would not just delete it.

hasufell:
who keeps pushing to master?

michaelpj:
It's called PR's getting merged ;)

Issue #496, opened 2020-10-13
Add a code action provider for the eval plugin (existing code lens)


michaelpj:
At the moment it's only a code lens, but it could be a code action too.

In `lsp-mode` this means it shows up when you explicitly show code lenses, but not in the sideline that shows code actions. So I actually thought it was broken (although arguably this just shows that I don't know `lsp-mode` well enough :sweat_smile:).

jmorag:
https://github.com/emacs-lsp/lsp-haskell/issues/89

This also means that the eval plugin doesn't work with emacs lsp-mode

michaelpj:
It does work, it just only works as a code lens. So you need to run `lsp-lens-show` to make the lenses show up. That's the way `lsp-mode` does code lenses vs code actions. Maybe we should open an issue there about improving the discoverability of code lenses.

jmorag:
When I click the code lens, it doesn't insert the value of the evaluation, it just crashes with 
```
(error "BadDependency "GhcSessionDeps"")
LSP :: Please open an issue in lsp-mode for implementing `21444:eval:evalCommand'.
```
@michaelpj does it insert the evaluation for you?

michaelpj:
Yes, it works fine. I think you have a server issue that's unrelated to this one, I suggest opening another issue with the reproduction instructions.

jneira:
There is other issue asking to have code lenses as code actions: #550

michaelpj:
I think it would be nice if all code lens providers also provided equivalent code actions.

jneira:
we could do it generically in hls-plugin-api, i hope

LeuschkeTressa:
This would be a great thing to have.

I have chosen to disable code lenses since I find it clutters the code (granularity of settings for enabling code lenses in various contexts is limited, so electing to turn them off completely for the moment), however, this currently means I can't use the eval plugin.

PR #3682, opened 2023-06-30
Resolve for explicit-imports

joyfulmantis:
I am pretty much rewriting the interface with hls for this plugin, as it was pretty poorly written (functions inside both an io and maybe monad, that used neither of them). In the process besides adding support for resolve for the code lens and code actions, I am also adding specific code actions for each import (before there was just one code action for explicit imports for everything)

### DIFF
--- a/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
+++ b/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
@@ -110,7 +110,7 @@ runImportCommand recorder ideState eird@(ResolveOne _ _) = pluginResponse $ do
           pure ()
         logErrors (Right _) = pure ()
 runImportCommand _ _ (ResolveAll _) = do
-  pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexepected argument for command handler: ResolveAll" Nothing
+  pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for command handler: ResolveAll" Nothing
 
 -- | For every implicit import statement, return a code lens of the corresponding explicit import
 -- Example - for the module below:
@@ -126,12 +126,11 @@ lensProvider :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState Meth
 lensProvider _  state _ CodeLensParams {_textDocument = TextDocumentIdentifier {_uri}}
   = pluginResponse $ do
     nfp <- getNormalizedFilePath _uri
-    mbMinImports <- liftIO $ runAction "MinimalImports" state $ useWithStale MinimalImports nfp
+    mbMinImports <- liftIO $ runAction "MinimalImports" state $ use MinimalImports nfp
     case mbMinImports of
-      Just (MinimalImportsResult minImports _ _, posMapping) -> do
-        let lens = [ generateLens _uri curRange int
-                    | (range, int) <- minImports
-                    , Just curRange <- [toCurrentRange posMapping range]]
+      Just (MinimalImportsResult minImports _ _) -> do
+        let lens = [ generateLens _uri range int
+                    | (range, int) <- minImports]
         pure $ InL lens
       _ ->
         pure $ InL []
@@ -158,11 +157,11 @@ lensResolveProvider _ ideState plId cl@(CodeLens {_data_ = Just data_@(A.fromJSO
               _arguments = Just [data_]
           in mkLspCommand pId importCommandId title _arguments
 lensResolveProvider _  _ _ (CodeLens {_data_ = Just (A.fromJSON -> A.Success (ResolveAll _))}) = do
-   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexepected argument for lens resolve handler: ResolveAll" Nothing
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for lens resolve handler: ResolveAll" Nothing
 lensResolveProvider _  _ _ (CodeLens {_data_ = Just (A.fromJSON @EIResolveData -> (A.Error (T.pack -> str)))}) =
   pure $ Left $ ResponseError (InR ErrorCodes_ParseError) str Nothing
 lensResolveProvider _  _ _ (CodeLens {_data_ = _}) = do
-   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexepected argument for lens resolve handler: (Probably Nothing)" Nothing
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for lens resolve handler: (Probably Nothing)" Nothing
 
 -- | If there are any implicit imports, provide both one code action per import
 --   to make that specific import explicit, and one code action to turn them all
@@ -201,7 +200,7 @@ codeActionResolveProvider _ ideState _ ca@(CodeAction{_data_= Just (A.fromJSON -
 codeActionResolveProvider _ _ _ (CodeAction{_data_= Just (A.fromJSON @EIResolveData -> A.Error (T.pack -> str))}) =
     pure $ Left $ ResponseError (InR ErrorCodes_ParseError) str Nothing
 codeActionResolveProvider _  _ _ (CodeAction {_data_ = _}) = do
-   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexepected argument for code action resolve handler: (Probably Nothing)" Nothing
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for code action resolve handler: (Probably Nothing)" Nothing
 --------------------------------------------------------------------------------
 
 resolveWTextEdit :: IdeState -> EIResolveData -> ExceptT String (LspT Config IO) WorkspaceEdit
@@ -244,10 +243,10 @@ data MinimalImportsResult = MinimalImportsResult
     forLens        :: [(Range, Int)]
     -- |For the code actions we have the same data as for the code lenses, but
     -- we store it in a RangeMap, because that allows us to filter on a specific
-    -- range with better performance, and code actions are almost allways only
+    -- range with better performance, and code actions are almost always only
     -- requested for a specific range
   , forCodeActions :: RM.RangeMap (Range, Int)
-    -- |For resolve we have an intMap where for every previously provied unique id
+    -- |For resolve we have an intMap where for every previously provided unique id
     -- we provide a textEdit to allow our code actions or code lens to be resolved
   , forResolve     :: IM.IntMap TextEdit }
 
@@ -272,9 +271,9 @@ exportedModuleStrings _ = []
 minimalImportsRule :: Recorder (WithPriority Log) -> (ModuleName -> Bool) -> Rules ()
 minimalImportsRule recorder pred = define (cmapWithPrio LogShake recorder) $ \MinimalImports nfp -> do
   -- Get the typechecking artifacts from the module
-  Just tmr <- use TypeCheck nfp
+  Just (tmr, tmrpm) <- useWithStale TypeCheck nfp
   -- We also need a GHC session with all the dependencies
-  Just hsc <- use GhcSessionDeps nfp
+  Just (hsc, _) <- useWithStale GhcSessionDeps nfp
   -- Use the GHC api to extract the "minimal" imports
   Just (imports, mbMinImports) <- liftIO $ extractMinimalImports hsc tmr
   let importsMap =
@@ -283,14 +282,16 @@ minimalImportsRule recorder pred = define (cmapWithPrio LogShake recorder) $ \Mi
             | L (locA -> RealSrcSpan l _) i <- mbMinImports
           ]
       res =
-        [ (realSrcSpanToRange l, minImport)
-          | i@(L _ imp) <- imports
-          , not (isQualifiedImport imp)
-          , not (isExplicitImport imp)
-          , let L _ mn = ideclName imp
-          , pred mn
-          , RealSrcSpan l _ <- [getLoc i]
-          , Just minImport <- [Map.lookup (realSrcSpanStart l) importsMap]
+        [ (newRange, minImport)
+          | imp@(L _ impDecl) <- imports
+          , not (isQualifiedImport impDecl)
+          , not (isExplicitImport impDecl)
+          , let L _ moduleName = ideclName impDecl
+          , pred moduleName
+          , RealSrcSpan location _ <- [getLoc imp]
+          , let range = realSrcSpanToRange location
+          , Just minImport <- [Map.lookup (realSrcSpanStart location) importsMap]
+          , Just newRange <- [toCurrentRange tmrpm range]
         ]
   uniqueAndRangeAndText <- liftIO $ for res $ \rt -> do
                                 u <- U.hashUnique <$> U.newUnique
@@ -385,10 +386,6 @@ abbreviateImportTitle input =
   in title
 
 --------------------------------------------------------------------------------
-
--- | A helper to run ide actions
-runIde :: IdeState -> Action a -> IO a
-runIde = runAction "importLens"
 
 within :: Range -> SrcSpan -> Bool
 within (Range start end) span =
--- a/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
+++ b/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
@@ -160,8 +160,8 @@ lensResolveProvider _  _ _ (CodeLens {_data_ = Just (A.fromJSON -> A.Success (Re
    pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for lens resolve handler: ResolveAll" Nothing
 lensResolveProvider _  _ _ (CodeLens {_data_ = Just (A.fromJSON @EIResolveData -> (A.Error (T.pack -> str)))}) =
   pure $ Left $ ResponseError (InR ErrorCodes_ParseError) str Nothing
-lensResolveProvider _  _ _ (CodeLens {_data_ = _}) = do
-   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for lens resolve handler: (Probably Nothing)" Nothing
+lensResolveProvider _  _ _ (CodeLens {_data_ = v}) = do
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) ("Unexpected argument for lens resolve handler: " <> (T.pack $ show v)) Nothing
 
 -- | If there are any implicit imports, provide both one code action per import
 --   to make that specific import explicit, and one code action to turn them all
@@ -199,8 +199,8 @@ codeActionResolveProvider _ ideState _ ca@(CodeAction{_data_= Just (A.fromJSON -
     pure $ ca & L.edit ?~ wedit
 codeActionResolveProvider _ _ _ (CodeAction{_data_= Just (A.fromJSON @EIResolveData -> A.Error (T.pack -> str))}) =
     pure $ Left $ ResponseError (InR ErrorCodes_ParseError) str Nothing
-codeActionResolveProvider _  _ _ (CodeAction {_data_ = _}) = do
-   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for code action resolve handler: (Probably Nothing)" Nothing
+codeActionResolveProvider _  _ _ (CodeAction {_data_ = v}) = do
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) ("Unexpected argument for code action resolve handler: " <> (T.pack $ show v)) Nothing
 --------------------------------------------------------------------------------
 
 resolveWTextEdit :: IdeState -> EIResolveData -> ExceptT String (LspT Config IO) WorkspaceEdit
--- a/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
+++ b/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE NamedFieldPuns     #-}
 {-# LANGUAGE OverloadedStrings  #-}
 {-# LANGUAGE RecordWildCards    #-}
+{-# LANGUAGE TypeApplications   #-}
 {-# LANGUAGE TypeFamilies       #-}
 {-# LANGUAGE ViewPatterns       #-}
 
@@ -17,25 +18,25 @@ module Ide.Plugin.ExplicitImports
   , abbreviateImportTitle
   , Log(..)
   ) where
+
 import           Control.DeepSeq
 import           Control.Lens                         ((&), (?~))
-import           Control.Monad                        (replicateM)
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Class            (lift)
-import           Control.Monad.Trans.Except           (throwE)
-import           Data.Aeson                           (Result (Success),
-                                                       ToJSON (toJSON),
-                                                       Value (Null), fromJSON)
+import           Control.Monad.Trans.Except           (ExceptT)
+import           Control.Monad.Trans.Maybe
+import qualified Data.Aeson                           as A (Result (..),
+                                                            ToJSON (toJSON),
+                                                            Value (Null),
+                                                            fromJSON)
 import           Data.Aeson.Types                     (FromJSON)
-import qualified Data.HashMap.Strict                  as HashMap
 import qualified Data.IntMap                          as IM (IntMap, elems,
                                                              fromList, (!?))
 import           Data.IORef                           (readIORef)
 import qualified Data.Map.Strict                      as Map
-import           Data.Maybe                           (catMaybes, fromMaybe,
-                                                       isJust)
 import           Data.String                          (fromString)
 import qualified Data.Text                            as T
+import           Data.Traversable                     (for)
 import qualified Data.Unique                          as U (hashUnique,
                                                             newUnique)
 import           Development.IDE                      hiding (pluginHandlers,
@@ -44,17 +45,14 @@ import           Development.IDE.Core.PositionMapping
 import qualified Development.IDE.Core.Shake           as Shake
 import           Development.IDE.GHC.Compat
 import           Development.IDE.Graph.Classes
-import           Development.IDE.Types.Logger         as Logger (Pretty (pretty))
 import           GHC.Generics                         (Generic)
 import           Ide.Plugin.RangeMap                  (filterByRange)
 import qualified Ide.Plugin.RangeMap                  as RM (RangeMap, fromList)
 import           Ide.PluginUtils                      (getNormalizedFilePath,
                                                        handleMaybe,
                                                        handleMaybeM,
-                                                       mkLspCommand,
                                                        pluginResponse)
 import           Ide.Types
-import           Language.LSP.Protocol.Lens           (HasDocumentChanges (documentChanges))
 import qualified Language.LSP.Protocol.Lens           as L
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types          hiding (Null)
@@ -63,13 +61,15 @@ import           Language.LSP.Server
 importCommandId :: CommandId
 importCommandId = "ImportLensCommand"
 
-newtype Log
+data Log
   = LogShake Shake.Log
+  | LogResponseError ResponseError
   deriving Show
 
 instance Pretty Log where
   pretty = \case
     LogShake log -> pretty log
+    LogResponseError msg -> pretty (show msg)
 
 -- | The "main" function of a plugin
 descriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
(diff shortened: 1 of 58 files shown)


michaelpj:
> In the process besides adding support for resolve for the code lens and code actions, I am also adding specific code actions for each import

This is nice, I think I even have an issue for this. In particular, it seems easier for people to write client-side stuff to trigger specific code actions, so this will make it easier for people to define a hotkey for "make current import (only) explicit".

michaelpj:
https://github.com/haskell/haskell-language-server/issues/496

joyfulmantis (commented):
The plugin "works" now and there are specific tests for it. While I am pretty happy with the general structure, I will want to go over and see if there is anything I want to refactor or clean up still though

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -71,33 +87,36 @@ descriptorForModules recorder pred plId =
...
       -- This plugin defines a new rule
-      pluginRules = minimalImportsRule recorder,
-      pluginHandlers = mconcat
-        [ -- This plugin provides code lenses
-          mkPluginHandler SMethod_TextDocumentCodeLens $ lensProvider pred
+      pluginRules = minimalImportsRule recorder pred,
+      pluginHandlers =
+         -- This plugin provides code lenses
+           mkPluginHandler SMethod_TextDocumentCodeLens (lensProvider pred)
+        <> mkPluginHandler SMethod_CodeLensResolve lensResolveProvider
```
michaelpj, line 96: Do we want to design a similar paired handler function for lenses like we have for code actions?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -71,33 +87,36 @@ descriptorForModules recorder pred plId =
...
 -- | The actual command handler
-runImportCommand :: CommandFunction IdeState ImportCommandParams
-runImportCommand _state (ImportCommandParams edit) = do
-  -- This command simply triggers a workspace edit!
-  _ <- sendRequest SMethod_WorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing edit) (\_ -> pure ())
-  return (Right Null)
+runImportCommand :: CommandFunction IdeState EIResolveData
+runImportCommand ideState eird = pluginResponse $ do
+  case eird of
+    ResolveOne uri int -> do
```
michaelpj, line 106: Missing case for `ResolveAll`?

michaelpj: Also, since you've named the fields, this can be a nice place to use `NamedFieldPuns`, i.e.
```
ResolveOne{uri, int} -> ...
```
just makes sure you can't mix up the positional args

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,49 +240,70 @@ instance NFData MinimalImports
...
+  , forCodeActions :: RM.RangeMap (Range, Int)
+  , forResolve     :: IM.IntMap (Range, T.Text) }
 
 instance Show MinimalImportsResult where show _ = "<minimalImportsResult>"
 
 instance NFData MinimalImportsResult where rnf = rwhnf
 
+data EIResolveData = ResolveOne
+                      { uri :: Uri
+                      , int :: Int }
```
michaelpj, line 254: better field name? :joy: how about "importId" or something?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -117,65 +136,99 @@ lensProvider
...
       return $ Right $ InL []
+  where generateLens :: Uri  -> Range -> Int -> CodeLens
+        generateLens uri range int =
+          CodeLens { _data_ = Just $ toJSON $ ResolveOne uri int
+                   , _range = range
+                   , _command = Nothing }
+
+lensResolveProvider :: PluginMethodHandler IdeState Method_CodeLensResolve
+lensResolveProvider ideState plId cl@(CodeLens {_data_ = Just data_}) = pluginResponse $ do
+  case fromJSON data_ of
```
michaelpj, line 160: This applies in a few other places where you've done something like this, but if you like you can use `ViewPatterns` for something like this, e.g.
```
CodeLens {_data = Just (fromJSON -> Success (ResolveOne uri int)) } = ...
```
It's good for cases where you don't want to do different handling if the different parts fail. Arguably we might want to do that here: we might want to log differently if the code lens is missing a data field versus if the field is present but doesn't have the form we expect.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -117,65 +136,99 @@ lensProvider
...
+    Success (ResolveOne uri int) -> do
+      nfp <- getNormalizedFilePath uri
+      (MinimalImportsResult _ _ resolveMinImp) <-
+        handleMaybeM "Unable to run Minimal Imports"
+        $ liftIO
+        $ runAction "MinimalImports" ideState $ use MinimalImports nfp
+      target <- handleMaybe "Unable to resolve lens" $ resolveMinImp IM.!? int
+      let updatedCodeLens = cl & L.command ?~  mkCommand plId uri target data_
+      return updatedCodeLens
+    _ -> throwE "unable to parse data_ field of CodeLens"
```
michaelpj, line 170: Error handling things: this doesn't tell us what it _was_, and also discards the error from aeson that might be helpful. Hopefully we will already have logged the CodeAction itself, so we'll be able to see the data field there, but it's usually helpful to repeat things. And we definitely want to reproduce the aeson error.

This would be easier with structured errors, but even just printing them is an improvement over getting no info!

(This applies in some of your other code too, I think)

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -117,65 +136,99 @@ lensProvider
...
+lensResolveProvider ideState plId cl@(CodeLens {_data_ = Just data_}) = pluginResponse $ do
+  case fromJSON data_ of
+    Success (ResolveOne uri int) -> do
+      nfp <- getNormalizedFilePath uri
+      (MinimalImportsResult _ _ resolveMinImp) <-
+        handleMaybeM "Unable to run Minimal Imports"
+        $ liftIO
+        $ runAction "MinimalImports" ideState $ use MinimalImports nfp
+      target <- handleMaybe "Unable to resolve lens" $ resolveMinImp IM.!? int
+      let updatedCodeLens = cl & L.command ?~  mkCommand plId uri target data_
```
michaelpj, line 168: Interesting, so we really split things into three here:
- First we work out where the lenses are
- Then, in resolve we work out the title
- Then, in the command handler we work out the edit

And all of them actually use `MinimalImports`. I wonder if we need to shuffle the work around a bit - if we end up doing most of the work in `MinimalImports`, then maybe we're not saving much with the resolve provider. Not sure.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -117,65 +136,99 @@ lensProvider
...
+
+lensResolveProvider :: PluginMethodHandler IdeState Method_CodeLensResolve
+lensResolveProvider ideState plId cl@(CodeLens {_data_ = Just data_}) = pluginResponse $ do
+  case fromJSON data_ of
+    Success (ResolveOne uri int) -> do
+      nfp <- getNormalizedFilePath uri
+      (MinimalImportsResult _ _ resolveMinImp) <-
+        handleMaybeM "Unable to run Minimal Imports"
+        $ liftIO
+        $ runAction "MinimalImports" ideState $ use MinimalImports nfp
```
michaelpj, line 166: This is the only one that uses `use` rather than `useWithStale`, why?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -117,65 +136,99 @@ lensProvider
...
+      let updatedCodeLens = cl & L.command ?~  mkCommand plId uri target data_
+      return updatedCodeLens
+    _ -> throwE "unable to parse data_ field of CodeLens"
+  where mkCommand ::  PluginId -> Uri -> (Range, T.Text) -> Value -> Command
+        mkCommand pId uri (_, text) data_ =
+          let title = abbreviateImportTitle text
+              _arguments = Just [data_]
+          in mkLspCommand pId importCommandId title _arguments
 
 -- | If there are any implicit imports, provide one code action to turn them all
```
michaelpj, line 177: out of date

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -117,65 +136,99 @@ lensProvider
...
+codeActionResolveProvider ideState plId ca@(CodeAction{_data_= Just value}) =
+  pluginResponse $ do
+    case fromJSON value of
+      Success (ResolveOne uri int) -> do
+        nfp <- getNormalizedFilePath uri
+        (MinimalImportsResult _ _ resolveMinImp) <-
+          handleMaybeM "Unable to run Minimal Imports"
+          $ liftIO
+          $ runAction "MinimalImports" ideState $ use MinimalImports nfp
+        (range, text) <- handleMaybe "Unable to resolve lens" $ resolveMinImp IM.!? int
```
michaelpj, line 216:
```suggestion
        (range, text) <- handleMaybe "Unable to resolve code action" $ resolveMinImp IM.!? int
```

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,49 +240,70 @@ instance NFData MinimalImports
 
 type instance RuleResult MinimalImports = MinimalImportsResult
 
-newtype MinimalImportsResult = MinimalImportsResult
-  {getMinimalImportsResult :: [(LImportDecl GhcRn, Maybe T.Text)]}
+data MinimalImportsResult = MinimalImportsResult
+  { forLens        :: [(Range, Int)]
+  , forCodeActions :: RM.RangeMap (Range, Int)
```
michaelpj, line 245: Why do we need the previous field, can't we just enumerate the `(Range, Int)` pairs from the `RangeMap`?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,49 +240,70 @@ instance NFData MinimalImports
...
+  , forResolve     :: IM.IntMap (Range, T.Text) }
```
michaelpj, line 246: `(Range, Text)` is basically just a `TextEdit`, let's use that instead!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: ... also why does this have a range in the output? is it the same range as in the input?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,49 +240,70 @@ instance NFData MinimalImports
...
+minimalImportsRule recorder pred = define (cmapWithPrio LogShake recorder) $ \MinimalImports nfp -> do
   -- Get the typechecking artifacts from the module
-  tmr <- use TypeCheck nfp
+  Just tmr <- use TypeCheck nfp
   -- We also need a GHC session with all the dependencies
-  hsc <- use GhcSessionDeps nfp
+  Just hsc <- use GhcSessionDeps nfp
   -- Use the GHC api to extract the "minimal" imports
-  (imports, mbMinImports) <- liftIO $ extractMinimalImports hsc tmr
+  (imports, Just mbMinImports) <- liftIO $ extractMinimalImports (Just hsc) (Just tmr)
```
michaelpj, line 273: This is a bit weird, you extract it from the Maybe and then put it right back?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,49 +240,70 @@ instance NFData MinimalImports
...
+          , let L _ mn = ideclName imp
+          , pred mn
+          , let RealSrcSpan l _ = getLoc i
+          , Just minImport <- [Map.lookup (realSrcSpanStart l) importsMap]
         ]
-  return ([], MinimalImportsResult res <$ mbMinImports)
-  where
-    isImplicitPrelude :: (Outputable a) => a -> Bool
+  uniques <- liftIO $ replicateM (length res) (U.hashUnique <$> U.newUnique)
+  let uniqueAndRangeAndText = zip uniques res
```
michaelpj, line 292: I sometimes write this sort of thing like this:
```
uniqueAndRangeAndText = for res $ \r ->
  u <- U.hashUnique <$> U.newUnique
  pure (r, u)
```
I just find it a bit easier often sine you have a variable for the thing and a natural place for a do block

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -294,21 +357,7 @@ maxColumns = 120
 
 -- | Given an import declaration, generate a code lens unless it has an
```
michaelpj, line 358: dead

michaelpj (commented):
Needs a bit more work but generally looking plausible. I like how we get to unify all the logic into the rule and the (sort of) command handler.

It _almost_ feels like there's a pattern here you could use to unify all of the things. Something like:
- You have a rule that generates a bunch of possible actions in a file, with ranges.
- You have a command handler/resolve function handler that applies the action
- You have a code lens handler that generates lenses for some subset of the actions
- You have a code lens resolve handler that calls the command handler and uses some logic to set the title appropriately
- You have a code action handler that generates code actions for some subset of the actions.
- You have a code action resolve handler that calls the command handler/fills in the edit, depending.

I think it probably would only work for a restricted set of things: you really need your "logic handler" to just produce an edit, so you can use that same edit in all scenarios (`applyEdit`, filling it in on a code action, etc.).

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: The command actually is responding only to the lens, so shouldn't handle the ResolveAll case. In this case I'm actually reusing the data function, but it might be better to have seperate types.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: Since the client never advertises a code lens resolve capability, we have to assume that all clients support it, and can't change our behaviour depending on whether they support it.

michaelpj: seems busted, I opened https://github.com/microsoft/language-server-protocol/issues/1761

michaelpj: I think it's very likely that almost no clients support it so this may just break for many clients.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: I think it seems reasonable to share the code for running the action if possible? Having one function that handles either case and then we just call it seems good.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: Yeah, refine-imports also relies on extractMinimalImports, so I originally wanted to move the failure to the action monad, and then later delayed changing the extractMinimalImports to work with the new types to avoid changing refine-imports too

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: From what I've researched it's not possible to easily extract the Range key from the RangeMap. Also the range in the output is the range of the textedit, but the range of the input is (in this case) what the client sends over, which is not identical

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: There's not that much we can delay. I guess we can delay printing the expression till when we get the title, which currently isn't done, mainy because that's what the old code did, and I havent changed it. The key thing is that for most plugins the tricky part is finding out where we need to provide it, and that always needs to be done as the first step, regardless of resolve. Especially for codeLens, there is no way to avoid computing the whole documents codeLens locations every edit.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: Okay good point, there is no reason to use resolve from codeActions then, but no reason to implement resolve for no reason.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: I think that actually not having a capability for it must mean that all clients (need to?) support it. It's also a realtively important resolve type, because you send all your codelens every edit. Vscode works with resolve, and I know the rust server uses it too.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -71,33 +88,36 @@ descriptorForModules recorder pred plId =
...
+  case eird of
+    ResolveOne uri int -> do
+      nfp <- getNormalizedFilePath uri
+      (MinimalImportsResult _ _ resolveMinImp) <-
+        handleMaybeM "Unable to run Minimal Imports"
+        $ liftIO
+        $ runAction "MinimalImports" ideState $ use MinimalImports nfp
+      (range, text) <- handleMaybe "Unable to resolve lens" $ resolveMinImp IM.!? int
+      let edit = mkWorkspaceEdit uri range text
+      _ <- lift $ sendRequest SMethod_WorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing edit) (\_ -> pure ())
```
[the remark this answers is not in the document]
joyfulmantis: So I guess the problem is we don't get anything useful back from sendRequest, just a LspId, so there's no way to like parse that and detect failure.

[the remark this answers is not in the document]
michaelpj: Right, but we also have the callback argument. We can't do all that much useful there, but at least we could log.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: Could still be useful to have the paired function since maybe we can simplify some stuff that's always the same like the data handling.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -109,75 +122,111 @@ runImportCommand _state (ImportCommandParams edit) = do
...
+        mkCommand pId TextEdit{_newText} =
+          let title = abbreviateImportTitle _newText
+              _arguments = Just [data_]
+          in mkLspCommand pId importCommandId title _arguments
+lensResolveProvider _  _ _ (CodeLens {_data_ = Just (A.fromJSON -> A.Success (ResolveAll _))}) = do
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for lens resolve handler: ResolveAll" Nothing
+lensResolveProvider _  _ _ (CodeLens {_data_ = Just (A.fromJSON @EIResolveData -> (A.Error (T.pack -> str)))}) =
+  pure $ Left $ ResponseError (InR ErrorCodes_ParseError) str Nothing
+lensResolveProvider _  _ _ (CodeLens {_data_ = _}) = do
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for lens resolve handler: (Probably Nothing)" Nothing
```
michaelpj, line 164: I'd just shove the whole data `Value` in the error tbh. If this ever happens you can't have too much info

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -109,75 +122,111 @@ runImportCommand _state (ImportCommandParams edit) = do
...
+
+codeActionResolveProvider :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState 'Method_CodeActionResolve
+codeActionResolveProvider _ ideState _ ca@(CodeAction{_data_= Just (A.fromJSON -> A.Success rd)}) =
+  pluginResponse $ do
+    wedit <- resolveWTextEdit ideState rd
+    pure $ ca & L.edit ?~ wedit
+codeActionResolveProvider _ _ _ (CodeAction{_data_= Just (A.fromJSON @EIResolveData -> A.Error (T.pack -> str))}) =
+    pure $ Left $ ResponseError (InR ErrorCodes_ParseError) str Nothing
+codeActionResolveProvider _  _ _ (CodeAction {_data_ = _}) = do
+   pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for code action resolve handler: (Probably Nothing)" Nothing
```
michaelpj, line 203: ditto

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -109,75 +122,111 @@ runImportCommand _state (ImportCommandParams edit) = do
...
 --------------------------------------------------------------------------------
 
+resolveWTextEdit :: IdeState -> EIResolveData -> ExceptT String (LspT Config IO) WorkspaceEdit
+resolveWTextEdit ideState (ResolveOne uri int) = do
```
michaelpj, line 207: Nit: you could deduplicate a bit by doing:
```
resolveWTextEdit ideState action = do
  nfp <- ..
  (MinimalImports ...
  case action of
     ResolveOne ...
```

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -109,75 +122,111 @@ runImportCommand _state (ImportCommandParams edit) = do
...
+resolveWTextEdit ideState (ResolveAll uri) = do
+  nfp <- getNormalizedFilePath uri
+  (MinimalImportsResult{forResolve}) <-
+    handleMaybeM "Unable to run Minimal Imports"
+    $ liftIO
+    $ runAction "MinimalImports" ideState $ use MinimalImports nfp
+  let edits = IM.elems forResolve
+  pure $ mkWorkspaceEdit uri edits
+
+mkWorkspaceEdit :: Uri -> [TextEdit] -> WorkspaceEdit
```
michaelpj, line 224: ... and if you did the deduplication you could probably inline this

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,76 +236,98 @@ instance NFData MinimalImports
...
-minimalImportsRule :: Recorder (WithPriority Log) -> Rules ()
-minimalImportsRule recorder = define (cmapWithPrio LogShake recorder) $ \MinimalImports nfp -> do
+minimalImportsRule :: Recorder (WithPriority Log) -> (ModuleName -> Bool) -> Rules ()
+minimalImportsRule recorder modFilter = defineNoDiagnostics (cmapWithPrio LogShake recorder) $ \MinimalImports nfp -> runMaybeT $ do
   -- Get the typechecking artifacts from the module
-  tmr <- use TypeCheck nfp
+  (tmr, tmrpm) <- MaybeT $ useWithStale TypeCheck nfp
   -- We also need a GHC session with all the dependencies
-  hsc <- use GhcSessionDeps nfp
+  (hsc, _) <- MaybeT $ useWithStale GhcSessionDeps nfp
```
michaelpj, line 276: ugh, this is a bit awkward because we now have two position mappings. I _think_ this is actually fine because we're never using any positions from this result. But it's worth a comment!

plugins/hls-explicit-imports-plugin/test/Main.hs
```diff
@@ -65,12 +70,51 @@ main = defaultTestRunner $
...
+  actions <- getCodeActions doc (pointRange l c)
+  case find ((== Just "Make all imports explicit") . caTitle) actions of
+    Just (InR x) -> executeCodeAction x
+    _            -> liftIO $ assertFailure "Unable to find CodeAction"
+
+codeActionBreakFile :: FilePath -> Int -> Int -> TestTree
+codeActionBreakFile fp l c = goldenWithExplicitImports " code action" fp codeActionNoResolveCaps $ \doc -> do
+  _ <- waitForDiagnostics
+  changeDoc doc [edit]
   actions <- getCodeActions doc (pointRange l c)
```
michaelpj, line 84: This test doesn't look quite right to me: shouldn't we be getting the code actions, picking one, _then_ editing the file, _then_ resolving the action we picked before and seeing it fail?

plugins/hls-explicit-imports-plugin/test/Main.hs
```diff
@@ -65,12 +70,51 @@ main = defaultTestRunner $
...
+    _            -> liftIO $ assertFailure "Unable to find CodeAction"
+
+codeActionOnlyResolveGoldenTest :: FilePath -> Int -> Int -> TestTree
+codeActionOnlyResolveGoldenTest fp l c = goldenWithExplicitImports " code action resolve" fp codeActionResolveCaps $ \doc -> do
+  actions <- getCodeActions doc (pointRange l c)
+  Just (InR x) <- pure $ find ((== Just "Make this import explicit") . caTitle) actions
+  resolved <- resolveCodeAction x
+  executeCodeAction resolved
+
+resolveCodeAction :: CodeAction -> Session CodeAction
```
michaelpj, line 113: I think these helpers could go into `lsp-test`? Or there's `hls-test-utils`

plugins/hls-explicit-imports-plugin/test/Main.hs
```diff
@@ -65,12 +70,51 @@ main = defaultTestRunner $
...
+resolveCodeAction ca = do
+  resolveResponse <- request SMethod_CodeActionResolve ca
+  Right resolved <- pure $ resolveResponse ^. L.result
```
michaelpj, line 116: ... but if we do that it might be nice to get a better error here, I assume we'll get something crap from the MonadFail instance

plugins/hls-explicit-imports-plugin/test/Main.hs
```diff
@@ -79,18 +123,16 @@ caTitle _                         = Nothing
 -- code lens tests
 
 codeLensGoldenTest :: FilePath -> Int -> TestTree
```
michaelpj, line 125: maybe these things could also go in `hls-test-utils`? Seems like you've been repeating something similar a few times. Good to deduplicate the test stuff too!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -67,37 +83,34 @@ descriptorForModules
...
           -- This plugin provides code actions
-        , mkPluginHandler SMethod_TextDocumentCodeAction $ codeActionProvider pred
-        ]
-    }
-
--- | The command descriptor
-importLensCommand :: PluginCommand IdeState
-importLensCommand =
-  PluginCommand importCommandId "Explicit import command" runImportCommand
+        <> mkCodeActionHandlerWithResolve (codeActionProvider recorder) (codeActionResolveProvider recorder)
```
michaelpj, line 98: Ah, so I think the thing I'm confused about is why we don't use `mkCodeActionHandlerWithResolveAndCommand` here. Then if we made `runImportCommand` handle everything then we could just use that logic for everything?

michaelpj (approved):
Generally looking good!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: I think hsc (GhcSessionDeps) is only used once to filter out exported modules, most of the heavy lifting is done by the (tmr)  (TypeCheck), so it shouldn't be a problem.

plugins/hls-explicit-imports-plugin/test/Main.hs
joyfulmantis: So this test causes a parsing error in the file and then gets a code action and executes it. Because we are using "useWithStale" we can still operate on a file that has a parse fail.

michaelpj: Ah I see, that's slightly different then. I think it might also be interesting to put an edit in between getting the actions and resolving one. Potentially one that doesn't cause a parse error, so it just mixes up the uniques, which should cause a failure still.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: Ah right, I see. Since we always need the handler for the lenses.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -187,76 +236,98 @@ instance NFData MinimalImports
...
+  TcModuleResult ->
+  IO (Maybe ([LImportDecl GhcRn], [LImportDecl GhcRn]))
+extractMinimalImports hsc TcModuleResult {..} = runMaybeT $ do
   -- extract the original imports and the typechecking environment
   let tcEnv = tmrTypechecked
       (_, imports, _, _) = tmrRenamed
       ParsedModule {pm_parsed_source = L loc _} = tmrParsed
       emss = exportedModuleStrings tmrParsed
-      span = fromMaybe (error "expected real") $ realSpan loc
+  Just srcSpan <- pure $ realSpan loc
```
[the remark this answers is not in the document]
joyfulmantis: Some version of transformers have a hoistMaybe I believe, but a failable pattern match in maybeT has the same effect. (And if I wanted to use hoistMaybe I would need to update constraints)

Issue #3513, opened 2023-03-05
Incorrect flags when loading multiple executable components on GHC 9.4


Vic-M:
Hello, I'm not sure if this is a configuration issue with my hie.yaml, but I'm unable to get HLS to work with two Main.hs files at the same time. There's no issues with compiling both of them with cabal though.

### Your environment
**Which OS do you use?**
Windows WSL2 running CentOS Stream 9

**Which version of GHC do you use and how did you install it?**
GHC 9.4.4 from ghcup

**How is your project built (alternative: link to the project)?**
https://gitlab.com/vmatassa/hls-two-main-example

**Which LSP client (editor/plugin) do you use?**
vscode

**Which version of HLS do you use and how did you install it?**
HLS 1.9.1.0 from ghcup

### Steps to reproduce

1. Open projectA/app/Main.hs
2. Open projectB/app/Main.hs right after

### Actual behaviour

HLS is unable to recognize libraries that are exclusively in projectB cabal's file.

```
Could not load module ‘Data.Aeson’
It is a member of the hidden package ‘aeson-2.1.2.1’.
Perhaps you need to add ‘aeson’ to the build-depends in your .cabal file.
It is a member of the hidden package ‘aeson-2.1.1.0’.
Perhaps you need to add ‘aeson’ to the build-depends in your .cabal file.
It is a member of the hidden package ‘aeson-2.1.1.0’.
Perhaps you need to add ‘aeson’ to the build-depends in your .cabal file.
```

![image](https://user-images.githubusercontent.com/32042587/222935491-354c5ac8-9e2b-4389-9f5c-30ef69dd80fd.png)

### Expected behaviour

HLS should be able to find installed libraries from all cabal files in the project.

fendor:
Hi, thank you for your bug report!

I can reproduce on windows.
Maybe we use relativ filepaths somewhere...

L0neGamer:
My company upgraded to GHC 9.4.5 last thursday and we're getting similar issues with our HLS instances. We have multiple packages in the same project, and the ones marked as missing include both our own packages and ones like `shakespeare-2.1.0`, `directory-1.3.7.1`, and others.

Do the maintainers need any investigation on symptom-observers' behalf?

georgefst:
[...]

Issue #1822, opened 2021-05-11
False positive: Main.hs (executable) does not recognise function defined in Lib.hs (library)


mgpwilson:
Hello, I am having trouble with erroneous error messages from the linter.

The steps I have taken are:
1. Created a new Haskell Stack project with `stack new <project-name>`
2. Used `stack setup`
3. Created a function in Lib.hs and then used that function in Main.hs

Below are screenshots of the error I see.

![image](https://user-images.githubusercontent.com/9961083/117804358-9791b000-b24f-11eb-9704-b77b10a83863.png)
![image](https://user-images.githubusercontent.com/9961083/117804387-9fe9eb00-b24f-11eb-9ce4-697e452046a4.png)

Whenever I use `stack run`, the program compiles and runs as expected.

googleson78:
I think this is also related to https://github.com/haskell/haskell-language-server/issues/735  (same root issue?)

jneira:
> 
> 
> I think this is also related to #735 (same root issues?)

Yeah, thanks for linking the previous issue, not totally sure though

mgpwilson:
Hi there,

Adding a `hie.yaml` file with the contents
```
cradle:
  stack:
```
seemed to fix the issue for me.

googleson78:
> > I think this is also related to #735 (same root issues?)
> 
> Yeah, thanks for linking the previous issue, not totally sure though

I was linking it because I constantly get this "staleness" issue - if I add a new export to something in `lib`, hls won't detect in in another component, until I manually rebuild `lib`. And it seems related to me, because they're both related to intercomponent dependencies of some sort.

jneira:
> 
> 
> Hi there,
> 
> Adding a `hie.yaml` file with the contents
> 
> ```
> cradle:
>   stack:
> ```
> 
> seemed to fix the issue for me.

Nice to see you can workaround the issue, have you got the opportunity to check if removing `hie.yaml` the issue is reproduced again?

mgpwilson:
Hi @jneira

When I remove the `hie.yaml` file I get the following error:

![image](https://user-images.githubusercontent.com/9961083/117974304-5370df00-b325-11eb-9ea3-2296399a67e0.png)

When I reload developer window, that error is gone and the former error is reproduced:

![image](https://user-images.githubusercontent.com/9961083/117974576-a2b70f80-b325-11eb-9a99-5119e51c2060.png)

[...]

PR #3686, opened 2023-07-05
Some fixes for multi component stuff

wz1000:
1. Only bring units actually depended on into scope on 9.4+
2. Cabal uses `main` as the unit id of all executable packages. This confused multi component sessions.
   Solution: include the hash of the options in the unit id when the unit id is called "main".

Fixes https://github.com/haskell/haskell-language-server/issues/3513

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -484,7 +484,25 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
         packageSetup (hieYaml, cfp, opts, libDir) = do
           -- Parse DynFlags for the newly discovered component
           hscEnv <- emptyHscEnv ideNc libDir
-          (df, targets) <- evalGhcEnv hscEnv $ setOptions opts (hsc_dflags hscEnv)
+          (df', targets) <- evalGhcEnv hscEnv $ setOptions opts (hsc_dflags hscEnv)
+          let df =
+#if MIN_VERSION_ghc(9,3,0)
+                case unitIdString (homeUnitId_ df') of
+                     -- cabal uses main for the unit id of all executable packages
+                     -- This makes multi-component sessions confused about what
+                     -- options to use for that component.
+                     -- Solution: hash the options and use that as part of the unit id
+                     -- This works because there won't be any dependencies on the
+                     -- executable unit.
+                     "main" ->
+                       let hash = B.unpack $ B16.encode $ H.finalize $ H.updates H.init (map B.pack $ componentOptions opts)
+                           hashed_uid = Compat.toUnitId (Compat.stringToUnit ("main-"++hash))
+                       in setHomeUnitId_ hashed_uid df'
+                     _ -> df'
+#else
+                df'
+#endif
+
           let deps = componentDependencies opts ++ maybeToList hieYaml
           dep_info <- getDependencyInfo deps
           -- Now lookup to see whether we are combining with an existing HscEnv
@@ -499,6 +517,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
                   -- We will modify the unitId and DynFlags used for
                   -- compilation but these are the true source of
                   -- information.
+                  
                   new_deps = RawComponentInfo (homeUnitId_ df) df targets cfp opts dep_info
                                 : maybe [] snd oldDeps
                   -- Get all the unit-ids for things in this component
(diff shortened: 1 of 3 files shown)


ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -484,7 +484,24 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
+          (df', targets) <- evalGhcEnv hscEnv $ setOptions opts (hsc_dflags hscEnv)
+          let df =
+#if MIN_VERSION_ghc(9,3,0)
+                case unitIdString (homeUnitId_ df) of
+                     -- cabal uses main for the unit id of all executable packages
+                     -- This makes multi-component sessions confused about what
+                     -- options to use for that component.
+                     -- Solution: hash the options and use that as part of the unit id
+                     -- This works because there won't be any dependencies on the
+                     -- executable unit.
```
fendor, line 496: I think https://github.com/haskell/cabal/pull/8726 fixes this, right? If so, I think we should include that here.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -484,7 +484,24 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
+                     "main" ->
+                       let hash = B.unpack $ B16.encode $ H.finalize $ H.updates H.init (map B.pack $ componentOptions opts)
+                           hashed_uid = Compat.toUnitId (Compat.stringToUnit ("main-"++hash))
+                       in setHomeUnitId_ hashed_uid df'
```
fendor, line 500: Can we use this logic to improve Stack support? E.g. stack repl doesn't always include a `this-unit-id` flag, can we just then insert a unit-id like this here? See https://github.com/haskell/haskell-language-server/issues/1822#issuecomment-895063073 for context.

ghcide/src/Development/IDE/Import/FindImports.hs
```diff
@@ -147,13 +148,26 @@ locateModule env comp_info exts targetFor modName mbPkgName isSource = do
...
   where
     dflags = hsc_dflags env
     import_paths = mapMaybe (mkImportDirs env) comp_info
+    other_imports =
+#if MIN_VERSION_ghc(9,4,0)
+      -- On 9.4+ instead of bringing all the units into scope, only bring into scope the units
+      -- this one depends on
+      map (\uid -> (uid, importPaths (homeUnitEnv_dflags (ue_findHomeUnitEnv uid ue)))) hpt_deps
+    ue = hsc_unit_env env
+    units = homeUnitEnv_units $ ue_findHomeUnitEnv (homeUnitId_ dflags) ue
```
fendor, line 164: This will throw if the unit id can't be found, right? Not much we can do, though.

fendor (approved):
Minor comments, looks good to me!

ghcide/src/Development/IDE/Import/FindImports.hs
wz1000: the home unit id should always be set in ghcide.

michaelpj:
> Cabal uses main as the unit id of all executable packages. This confused multi component sessions.

Is this a cabal bug? Presumably this means that e.g. a multi-repl with multiple executables will go wrong also?

michaelpj:
Ah I see, Fendor says it might be fixed.

ghcide/src/Development/IDE/Import/FindImports.hs
fendor: Yeah, I agree

ghcide/session-loader/Development/IDE/Session.hs
wz1000: I think this will just work in that case, because the default unit id is `main`.

fendor: Awesome!

michaelpj:
Test failures seem maybe genuine?

wz1000:
I don't think so, different tests fail every time

michaelpj:
IDK, it looks like it's almost always the `simple-multi-test` tests failing, which I think are the multi-component tests, so seems pretty plausible to me that this has made things worse?

wz1000:
Yeah, you are probably right, I see consistent failures in the 9.4+ jobs, I was looking at the <9.2 jobs and saw random failures earlier. I will investigate.

wz1000:
Turned out to be a missing fallthrough case that snuck in while cherry picking the commit :facepalm:

georgefst:
I've had my eye on this PR for a while since #3513 has been really annoying me, and for the record, it now works perfectly for my projects, where previous iterations before today would hang at initialisation. Thanks for the fix!

PR #3689, opened 2023-07-06
Enable RTS options to be specified at runtime

Kleidukos:
This allows downstream consumers to set the `GHCRTS` environment variable to limit memory consumption, without having to recompile HLS from source.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -50,6 +50,9 @@ tests: True
 package mfsolve
   tests: False
 
+package haskell-language-server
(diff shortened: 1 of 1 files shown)


hasufell:
@wz1000 any concerns?

hasufell:
oh... @Kleidukos did it not work in the official binaries?

Kleidukos:
Well I actually managed to miss that one, apologies.

PR #3691, opened 2023-07-07
Show package name and its version while hovering on import statements

July541:
HieFile doesn't have package information for import statements and we *can't* get package information from a simple `ModuleName` in a purity environment.

This pr still uses `GHC.findImportedModule` to find package info from ghc env, and `IO` only affect while hovering on import statements, so they are not expensive I think.


ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -213,21 +214,33 @@ atPoint
   -> DocAndKindMap
   -> HscEnv
   -> Position
-  -> Maybe (Maybe Range, [T.Text])
-atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ pointCommand hf pos hoverInfo
+  -> IO (Maybe (Maybe Range, [T.Text]))
+atPoint IdeOptions{} (HAR _ hf _ _ (kind :: HieKind hietype)) (DKMap dm km) env pos =
```
fendor, line 218:
```suggestion
atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos =
```

Or is the type annotation important?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -251,21 +265,48 @@ atPoint IdeOptions{} (HAR _ hf _ _ kind) (DKMap dm km) env pos = listToMaybe $ p
...
+          pkgTxt <- packageNameWithVersion m env
+          pure $ "*(" <> pkgTxt <> ")*"
+
+        -- Return the module text itself and
+        -- the package(with version) this `ModuleName` belongs to.
+        packageNameForImportStatement :: ModuleName -> IO T.Text
+        packageNameForImportStatement mod = do
+          mpkg <- findImportedModule env mod :: IO (Maybe Module)
+          let moduleName = printOutputable mod
+          case join $ traverse (flip packageNameWithVersion env) mpkg of
```
fendor, line 282:
```suggestion
          case mpkg >>= flip packageNameWithVersion env of
```

perhaps?

fendor (approved):
LGTM, some nitpicks

ghcide/src/Development/IDE/Spans/AtPoint.hs
July541: Yes, because `HAR` has an existential type, following signatures wont work without this, .

https://github.com/haskell/haskell-language-server/blob/47cf162ec7c7ff6a0960ccddee85f3bc31d2b268/ghcide/src/Development/IDE/Core/RuleTypes.hs#L209-L221

July541:
I'll merge this tomorrow if no more review here.

July541:
Nix build ran out of disks, can someone take a look?

fendor:
@July541 Failing nix CI is not blocking, so don't worry about it. It will likely get fixed during the next hackathon :)

Just rebase the PR, everything else, should be fine, imo.

PR #3696, opened 2023-07-08
Catch exceptions in commands and use lsp null

joyfulmantis:


ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -197,7 +211,9 @@ executeCommandHandlers recorder ecs = requestHandler SMethod_WorkspaceExecuteCom
           Nothing -> logAndReturnError recorder p (InR ErrorCodes_InvalidRequest) (commandDoesntExist com p xs)
           Just (PluginCommand _ _ f) -> case A.fromJSON arg of
             A.Error err -> logAndReturnError recorder p (InR ErrorCodes_InvalidParams) (failedToParseArgs com p err arg)
-            A.Success a -> fmap InL <$> f ide a
+            A.Success a ->
+                f ide a `catchAny`
+                (\e -> logAndReturnError' recorder (InR ErrorCodes_InternalError) (ExceptionInPlugin p (Some SMethod_WorkspaceApplyEdit) e))
```
fendor, line 216: Does this catch async exceptions?

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -261,22 +276,25 @@ extensibleNotificationPlugins recorder xs = mempty { P.pluginHandlers = handlers
           Just fs -> do
             -- We run the notifications in order, so the core ghcide provider
             -- (which restarts the shake process) hopefully comes last
-            mapM_ (\(pid,_,f) -> otTracedProvider pid (fromString $ show m) $ f ide vfs params) fs
+            mapM_ (\(pid,_,f) -> otTracedProvider pid (fromString $ show m) $ f ide vfs params
+                                    `catchAny`
+                                    (\e -> logWith recorder Warning (ExceptionInPlugin pid (Some m) e))) fs
```
fendor, line 281: Same question here, is this the `catchAny` from safe-exceptions?

joyfulmantis: This is the catchAny from UnliftIO.Exception, and it doesn't catch async exceptions. It is the same one used for catching exceptions in normal plugin handlers, I am just adding it to the command and notification handlers here.

ghcide/src/Development/IDE/Plugin/HLS.hs
joyfulmantis: answered in the other comment

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -261,22 +276,25 @@ extensibleNotificationPlugins recorder xs = mempty { P.pluginHandlers = handlers
...
+
 
 -- ---------------------------------------------------------------------
 
 runConcurrently
   :: MonadUnliftIO m
-  => (SomeException -> PluginId -> T.Text)
-  -> String -- ^ label
+  => (PluginId -> SMethod method -> SomeException -> T.Text)
+  -> SMethod method -- ^ label
```
michaelpj, line 289: wrong comment

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -177,9 +191,9 @@ executeCommandHandlers recorder ecs = requestHandler SMethod_WorkspaceExecuteCom
                 -- If we have a command, continue to execute it
                 Just (Command _ innerCmdId innerArgs)
                     -> execCmd ide (ExecuteCommandParams Nothing innerCmdId innerArgs)
-                Nothing -> return $ Right $ InL A.Null
+                Nothing -> return $ Right $ InR Null
```
michaelpj, line 194: This is fine but I it's literally the same behaviour. The upstream spec is weird, I made an issue: https://github.com/microsoft/language-server-protocol/issues/1766

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -873,7 +873,7 @@ data PluginCommand ideState = forall a. (FromJSON a) =>
 type CommandFunction ideState a
   = ideState
   -> a
-  -> LspM Config (Either ResponseError Value)
+  -> LspM Config (Either ResponseError (Value |? Null))
```
michaelpj, line 876: tbh, since they're represented the same, I'm not sure this really adds much except making things more complex for the users

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -261,22 +276,25 @@ extensibleNotificationPlugins recorder xs = mempty { P.pluginHandlers = handlers
           Just fs -> do
             -- We run the notifications in order, so the core ghcide provider
             -- (which restarts the shake process) hopefully comes last
```
michaelpj, line 278: :see_no_evil:

michaelpj (approved):
Can we add a Note explaining our exception handling strategy? Then we can add to it as we go. Something like:

```
{- Note [Exception handling in plugins]
Plugins run in LspM, and so have access to IO. This means they are likely to throw exceptions, even if only by accident or through calling libraries that throw exceptions.
Ultimately, we're running a bunch of less-trusted IO code, so we should be robust to
it throwing.

We don't want these to bring down HLS. So we catch and log exceptions wherever we run
a handler defined in a plugin.

The flip side of this is that it's okay for plugins to throw exceptions as a way of signalling
failure!
-}

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: I would say less complex. LSP Types are used everywhere, and now command users don't need to use Aeson if they don't use it anywhere else (Pretty much every command returns null).

joyfulmantis: Good point. Is there any reason why we can't use Aeson.Null for Null in lsp-types, perhaps by reexporting it instead of declaring our own? that would remove a source of name conflicts.

joyfulmantis: Okay right, because it's only a constructor, not a type in its own right.

PR #3704, opened 2023-07-11
WIP Goto dependency definition

nlander:
For users using cabal HEAD, enables indexing of dependency HIE files so that the source can be written out on calls to gotoDefinition.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -569,10 +569,28 @@ reportImportCyclesRule recorder =
 
 getHieAstsRule :: Recorder (WithPriority Log) -> Rules ()
 getHieAstsRule recorder =
-    define (cmapWithPrio LogShake recorder) $ \GetHieAst f -> do
-      tmr <- use_ TypeCheck f
-      hsc <- hscEnv <$> use_ GhcSessionDeps f
-      getHieAstRuleDefinition f hsc tmr
+    define (cmapWithPrio LogShake recorder) $ \GetHieAst f ->
+        case getSourceFileOrigin f of
+            FromProject -> do
+                tmr <- use_ TypeCheck f
+                hsc <- hscEnv <$> use_ GhcSessionDeps f
+                getHieAstRuleDefinition f hsc tmr
+            FromDependency -> do
+                se <- getShakeExtras
+                mHieFile <- liftIO
+                    $ runIdeAction "GetHieAst" se
+                    $ runMaybeT
+                    $ readHieFileForSrcFromDisk recorder f
+                pure ([], makeHieAstResult <$> mHieFile)
+    where
+        makeHieAstResult :: Compat.HieFile -> HieAstResult
+        makeHieAstResult hieFile =
+            HAR
+                (Compat.hie_module hieFile)
+                (Compat.hie_asts hieFile)
+                mempty
+                mempty
+                (HieFromDisk hieFile)
 
 persistentHieFileRule :: Recorder (WithPriority Log) -> Rules ()
 persistentHieFileRule recorder = addPersistentRule GetHieAst $ \file -> runMaybeT $ do
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -593,10 +593,13 @@ getHieAstsRule recorder =
         makeHieAstResult hieFile =
             HAR
                 (Compat.hie_module hieFile)
-                (Compat.hie_asts hieFile)
-                mempty
+                hieAsts
+                (Compat.generateReferencesMap $ M.elems $ getAsts hieAsts)
                 mempty
                 (HieFromDisk hieFile)
+            where
+                hieAsts :: HieASTs TypeIndex
+                hieAsts = Compat.hie_asts hieFile
 
 persistentHieFileRule :: Recorder (WithPriority Log) -> Rules ()
 persistentHieFileRule recorder = addPersistentRule GetHieAst $ \file -> runMaybeT $ do
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -569,19 +569,20 @@ reportImportCyclesRule recorder =
 
 getHieAstsRule :: Recorder (WithPriority Log) -> Rules ()
 getHieAstsRule recorder =
-    define (cmapWithPrio LogShake recorder) $ \GetHieAst f ->
-        case getSourceFileOrigin f of
-            FromProject -> do
-                tmr <- use_ TypeCheck f
-                hsc <- hscEnv <$> use_ GhcSessionDeps f
-                getHieAstRuleDefinition f hsc tmr
-            FromDependency -> do
+    define (cmapWithPrio LogShake recorder) $ \GetHieAst f -> do
+        isFoi <- use_ IsFileOfInterest f
+        case isFoi of
+            IsFOI ReadOnly -> do
                 se <- getShakeExtras
                 mHieFile <- liftIO
                     $ runIdeAction "GetHieAst" se
                     $ runMaybeT
                     $ readHieFileForSrcFromDisk recorder f
                 pure ([], makeHieAstResult <$> mHieFile)
+            _ -> do
+                tmr <- use_ TypeCheck f
+                hsc <- hscEnv <$> use_ GhcSessionDeps f
+                getHieAstRuleDefinition f isFoi hsc tmr
     where
         makeHieAstResult :: Compat.HieFile -> HieAstResult
         makeHieAstResult hieFile =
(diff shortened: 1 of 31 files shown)


michaelpj:
Tests?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -570,9 +576,27 @@ reportImportCyclesRule recorder =
...
+                tmr <- use_ TypeCheck f
+                hsc <- hscEnv <$> use_ GhcSessionDeps f
+                getHieAstRuleDefinition f hsc tmr
+    where
+        makeHieAstResult :: Compat.HieFile -> HieAstResult
+        makeHieAstResult hieFile =
+            HAR
+                (Compat.hie_module hieFile)
+                (Compat.hie_asts hieFile)
+                mempty
```
wz1000, line 597: we can fill in the `RefMap` using `Compat.generateReferencesMap`

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -112,10 +123,102 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
...
+            readMVar completionToken
+        indexPackageHieFiles :: Package -> [Module] -> IO ()
+        indexPackageHieFiles (Package package) modules = do
+            let pkgLibDir :: FilePath
+                pkgLibDir = case unitLibraryDirs package of
+                  [] -> ""
+                  (libraryDir : _) -> libraryDir
+                hieDir :: FilePath
+                hieDir = pkgLibDir </> "extra-compilation-artifacts"
+            modIfaces <- mapMaybeM loadModIface modules
```
wz1000, line 152: why do we need to load `.hi` files here?

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -112,10 +123,102 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
...
+                allModules :: [(ModuleName, Maybe Module)]
+                allModules = unitExposedModules package
+                    ++ zip (unitHiddenModules package) (repeat Nothing)
+                makeModule :: (ModuleName, Maybe Module)
+                           -> Module
+                makeModule (moduleName, Nothing) =
+                    mkModule (unitInfoId package) moduleName
+                -- When module is re-exported from another package,
+                -- the origin module is represented by value in Just
+                makeModule (_, Just otherPackageMod) = otherPackageMod
```
wz1000, line 212: modules from other packages won't have a `.hie` file in the `extra-compilation` directory of the package under consideration, so we can just skip them from the list.

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -112,10 +123,102 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
...
+                hieDir :: FilePath
+                hieDir = pkgLibDir </> "extra-compilation-artifacts"
+            modIfaces <- mapMaybeM loadModIface modules
+            traverse_ (indexModuleHieFile hieDir) modIfaces
+        indexModuleHieFile :: FilePath -> ModIface -> IO ()
+        indexModuleHieFile hieDir modIface = do
+            let hiePath :: NormalizedFilePath
+                hiePath = toNormalizedFilePath' $
+                  hieDir </> toFilePath (moduleName $ mi_module modIface) ++ ".hie"
+            hieCheck <- checkHieFile recorder se "newHscEnvEqWithImportPaths" hiePath
```
wz1000, line 159: we might want to be less granular with the `HieAlreadyIndexed` check - each package can contain hundreds/thousands of modules, and we don't want to check all these modules on every startup.

It's also wasteful because once we successfully index a whole unit/package we can assume that it doesn't change. So maybe we need to only check if a package has been indexed rather than checking for each file in the package.

This might require another table in `hiedb` to keep track of the packages we have indexed.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -106,52 +106,55 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
     mode <- liftIO $ runAction "codeLens.config" ideState $ usePropertyAction #mode pId properties
     nfp <- getNormalizedFilePath uri
-    env <- hscEnv . fst
-            <$> (handleMaybeM "Unable to get GhcSession"
-                $ liftIO
-                $ runAction "codeLens.GhcSession" ideState (useWithStale GhcSession nfp)
-                )
-    tmr <- fst <$> (
-                handleMaybeM "Unable to TypeCheck"
+    case Shake.getSourceFileOrigin nfp of
```
wz1000, line 109: We have `pluginFileType` in `PluginDescriptor` which specifies which filetypes plugins want to handle. We should extend this concept to extend to specifying whether plugins want to handle `FromDependency` files.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1125,6 +1128,14 @@ defineEarlyCutOffNoFile recorder f = defineEarlyCutoff recorder $ RuleNoDiagnost
     if file == emptyFilePath then do (hash, res) <- f k; return (Just hash, Just res) else
         fail $ "Rule " ++ show k ++ " should always be called with the empty string for a file"
 
+data SourceFileOrigin = FromProject | FromDependency
+
+getSourceFileOrigin :: NormalizedFilePath -> SourceFileOrigin
+getSourceFileOrigin f =
+    case isInfixOf ".hls/dependencies" (show f) of
```
wz1000, line 1135: `isInfixOf` and `show` are the wrong functions to use here.

You probably want to inspect the path using functions from `System.FilePath` and use `fromNormalizedFilePath` to convert `NormalizedFilePath` to `FilePath`.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1125,6 +1128,17 @@ defineEarlyCutOffNoFile recorder f = defineEarlyCutoff recorder $ RuleNoDiagnost
     if file == emptyFilePath then do (hash, res) <- f k; return (Just hash, Just res) else
         fail $ "Rule " ++ show k ++ " should always be called with the empty string for a file"
 
+data SourceFileOrigin = FromProject | FromDependency
+
+getSourceFileOrigin :: NormalizedFilePath -> SourceFileOrigin
+getSourceFileOrigin f =
+    case [".hls", "dependencies"] `isInfixOf` (splitDirectories file) of
```
nlander, line 1135: @wz1000  What do you think of this? It still uses `isInfixOf` but not show. I don't see any predicate in `System.FilePath` that is like `isInfixOf` but for directories being contained in a `FilePath`.

wz1000: This seems alright to me

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -59,21 +73,21 @@ updateHscEnvEq oldHscEnvEq newHscEnv = do
...
-    newHscEnvEqWithImportPaths (Just $ Set.fromList importPathsCanon) hscEnv deps
+    newHscEnvEqWithImportPaths (Just $ Set.fromList importPathsCanon) recorder se hscEnv deps
 
-newHscEnvEqWithImportPaths :: Maybe (Set FilePath) -> HscEnv -> [(UnitId, DynFlags)] -> IO HscEnvEq
-newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
+newHscEnvEqWithImportPaths :: Maybe (Set FilePath) -> Recorder (WithPriority Log) -> ShakeExtras -> HscEnv -> [(UnitId, DynFlags)] -> IO HscEnvEq
+newHscEnvEqWithImportPaths envImportPaths recorder se hscEnv deps = do
 
-    let dflags = hsc_dflags hscEnv
+    indexDependencyHieFiles
```
wz1000, line 90: We probably want to do this asynchronously.

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -112,10 +120,88 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
...
+            dotHlsDirExists <- maybe (pure False) doesDirectoryExist mHlsDir
+            unless dotHlsDirExists deleteMissingDependencySources
+            void $ Map.traverseWithKey indexPackageHieFiles packagesWithModules
+        mHlsDir :: Maybe FilePath
+        mHlsDir = do
+            projectDir <- resRootPath =<< lspEnv se
+            pure $ projectDir </> ".hls"
+        deleteMissingDependencySources :: IO ()
+        deleteMissingDependencySources = do
+            completionToken <- newEmptyMVar
```
wz1000, line 135: I don't think we need to wait for this to complete, we can just queue it up

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -112,10 +120,88 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
...
+                  (libraryDir : _) -> libraryDir
+                hieDir :: FilePath
+                hieDir = pkgLibDir </> "extra-compilation-artifacts"
+                unit :: Unit
+                unit = RealUnit $ Definite $ unitId package
+            moduleRows <- withHieDb se $ \db ->
+                lookupPackage db unit
+            case moduleRows of
+                [] -> traverse_ (indexModuleHieFile hieDir) modules
+                _  -> return ()
```
wz1000, line 156: comments explaining that we have already indexed the package

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -112,10 +120,88 @@ newHscEnvEqWithImportPaths envImportPaths hscEnv deps = do
           (evaluate . force . Just $ listVisibleModuleNames hscEnv)
 
     return HscEnvEq{..}
+    where
```
wz1000, line 123: I think we probably want to move this code to another module.

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -377,14 +384,24 @@ nameToLocation withHieDb lookupModule name = runMaybeT $
...
   file <- case modInfoSrcFile info of
-    Just src -> pure $ toUri src
-    Nothing -> lookupModule (defSrc row) (modInfoName info) (modInfoUnit info) (modInfoIsBoot info)
+    Just src -> do
+      -- Checking that the file exists covers the case where a
+      -- dependency file in .hls is in the database but got deleted
+      -- for any reason.
+      -- See the function `lookupMod` in Development.IDE.Core.Actions
+      -- for where dependency files get created and indexed in hiedb.
+      fileExists <- liftIO $ doesFileExist src
```
wz1000, line 400: I don't think we need to do this, as long as we check for .hls on startup.

joyfulmantis:
Closed as per request by @nlander on Matrix

PR #3706, opened 2023-07-11
Update stack stuff

July541:
Update `stack.yaml` to make it satisfy our policy.


michaelpj (approved):
I'll need to change the required jobs, but I'll do that once this is green.

Issue #3695, opened 2023-07-08
Improve the error message for unknown modules


fendor:
## Context

When users add a new module to a cabal project, they are usually greeted with an uninformative `Multi Cradle: No prefixes matched` or something similar. 

We can identify three different cases and different behaviour:

### 1. Adding an other-module `Test.hs` to an executable component 

In an example project with only a `Main.hs` module, adding an other-module `Test.hs` results in the following error message:
```
Multi Cradle: No prefixes matched
pwd: d:\Privat\Documents\programming\haskell\example
filepath: D:\Privat\Documents\programming\haskell\example\app\Test.hs
prefixes:
("app/Main.hs",Cabal {component = Just "example:exe:example"})
```

The cause is that `implicit-hie` generates roughly this hie.yaml file
```yaml
cradle:
  cabal:
    - path: ./app/Main.hs
      component: "example:exe:example"
```
Which essentially means, only map `./app/Main.hs` to the component `example:exe:example`. Clearly `./app/Test.hs` is not listed here, resulting in the sub-par error message from `hie-bios`.

Note, `implicit-hie` doesn't handle overlapping source directories correctly in most cases, see https://github.com/haskell/haskell-language-server/issues/3606 for example.

### 2. Adding an exposed-module to a library component

This generally works ok because `implicit-hie` generates `hie.yaml`s of the form:

```yaml
cradle:
  cabal:
    - path: ./src
      component: "example:lib"
```

Essentially, maps for each source directory to the library component. So why does this work even if a hypothetical `Lib.hs` is not part of the library component? `hie-bios` doesn't care about that and just gives us the compilation options for the component `example.:lib`. Then haskell-language-server just takes the options, creates a unit, and adds `Lib.hs` to it: https://github.com/haskell/haskell-language-server/blob/feb596592de95f09cf4ee885f3e74178161919f1/ghcide/session-loader/Development/IDE/Session.hs#L841
Read the comment right above it for the justification.

That's why adding modules to a library works "okayish" most of the time.

### 3. Adding a new module to a component with a simple 'hie.yaml'

However, this all falls apart when we handwrite the recommended hie.yaml file:

```yaml
cradle:
  cabal:
```
This basically runs `cabal repl src/Lib.hs` to find the compilation options. However, then `cabal` complains even in the case where it previously was working fine!

```
[...]

Issue #3693, opened 2023-07-07
Forgetting to add modules to cabal leads to confusing "Multi Cradle: No prefixes matched" error


seanhess:
## Is your enhancement request related to a problem? Please describe.

I just came back to Haskell after a few years. I'm trying the latest cabal instead of stack this time around, and running into all sorts of confusing problems. [See this thread for details](https://discourse.haskell.org/t/cabal-quickstart-with-hls-pains/6867). 

In short, if you create a new project with cabal, then add a module, you get the "multi-cradle no prefixes matched" error, which doesn't mean anything to me. As a stack/hpack user, I had forgotten that cabal required this, and had no idea what the error message meant. I suspect that any new Haskellers would be equally confused. 

In Visual Studio Code it highlights the first line of the second modules and ONLY shows you the following:

```
Multi Cradle: No prefixes matched
pwd: /Users/sean/Downloads/error-example
filepath: /Users/sean/Downloads/error-example/app/Test.hs
prefixes:
("app/Main.hs",Cabal {component = Just "error-example:exe:error-example"})
cradle
```

## Describe the solution you'd like

I would like HLS to give a clear error message saying that I may have forgotten to add the file to the cabal file. 

Note: I have a week off and I'm willing to fix this. I just want to make sure it's something that will be merged, and know about any caveats or potential issues before starting.

## Steps to Reproduce

1. Create a new project with default options
```
$ cabal --version
cabal-install version 3.10.1.0
compiled using version 3.10.1.0 of the Cabal librar

$ mkdir test && cd test
$ cabal init      # all default options
```

2. Add app/Test.hs with the following:

```
module Test where

test :: IO ()
test = putStrLn "TEST"
```

3. Import Test into Main

module Main where

```
import Test

main :: IO ()
main = do
        putStrLn "Hello, Haskell!"
        Test.test
```

4. Run HLS:

```
$ haskell-language-server-wrapper-2.0.0.1
...
2023-07-07T17:16:41.846129Z | Info | Cradle path: app/Test.hs
2023-07-07T17:16:41.846221Z | Warning | No [cradle](https://github.com/mpickering/hie-bios#hie-bios) found for app/Test.hs.
Proceeding with [implicit cradle](https://hackage.haskell.org/package/implicit-hie).
You should ignore this message, unless you see a 'Multi Cradle: No prefixes matched' error.
2023-07-07T17:16:41.869410Z | Info | invoking build tool to determine build flags (this may take some time depending on the cache)
[...]

PR #3711, opened 2023-07-11
User-friendly error message when modules aren't listed

seanhess:
Addresses #3695. When a user forgets to add their module to other-modules or exposed-modules, they would get a confusing "Multi Cradle: no prefixes matched" error. 

This PR replaces the old error message with the following:
```
  Loading the module 'app/Test.hs' failed. It seems that it is not listed in your .cabal file!
  Perhaps you need to add `Test` to other-modules or exposed-modules
  For more information, visit:
  https://cabal.readthedocs.io/en/3.4/developing-packages.html#modules-included-in-the-package
```

This is my first PR for HSL, please let me know which conventions I've missed.


fendor:
This looks like the perfect first step!

It would be great if we can include the cabal file in question in the error message. For cabal cradles, [`cradleErrorDependencies`](https://hackage.haskell.org/package/hie-bios-0.12.0/docs/HIE-Bios.html#v:cradleErrorDependencies) should contain the .cabal filepath if we were able to find it.

Further, I, personally, think we should show this error message only for cabal cradles (and perhaps stack cradles), so the error message still needs to be guarded. Look at the call site of `renderCradleError`, can we thread the CradleType or Cradle through such that we can call [`isCabalCradle`](https://hackage.haskell.org/package/hie-bios-0.12.0/docs/HIE-Bios-Cradle.html#v:isCabalCradle)?

As a word of warning, our tests are unfortunately rather flaky. Sometimes a test just fails. In that case, don't panic, we rerun the testsuite.

seanhess:
It looks like parsing any information from the prefixes is going to be too complicated to be worth it. Here are all the possibilities from hie-bios:

```
instance Show (CradleType a) where
    show (Cabal comp) = "Cabal {component = " ++ show (cabalComponent comp) ++ "}"
    show (CabalMulti d a) = "CabalMulti {defaultCabal = " ++ show d ++ ", subCabalComponents = " ++ show a ++ "}"
    show (Stack comp) = "Stack {component = " ++ show (stackComponent comp) ++ ", stackYaml = " ++ show (stackYaml comp) ++ "}"
    show (StackMulti d a) = "StackMulti {defaultStack = " ++ show d ++ ", subStackComponents = "  ++ show a ++ "}"
    show Bios { call, depsCall } = "Bios {call = " ++ show call ++ ", depsCall = " ++ show depsCall ++ "}"
    show (Direct args) = "Direct {arguments = " ++ show args ++ "}"
    show None = "None"
    show (Multi a) = "Multi " ++ show a
    show (Other _ val) = "Other {originalYamlValue = " ++ show val ++ "}"
```

We can show them the output of each prefix directly, something like:

```
  Loading the module 'app/Test.hs' failed. It seems that it is not listed in your .cabal file!
  Perhaps you need to add `Test` to other-modules or exposed-modules
  For more information, visit:
  https://cabal.readthedocs.io/en/3.4/developing-packages.html#modules-included-in-the-package

  app/Main.hs - Cabal {component = Just "hls-file-error:exe:hls-file-error"})
  ... any other prefixes here ...
```

Which, while it isn't the most user-friendly, at least surfaces all the relevant information.

fendor:
Oof, that is unintended behaviour that we have never realised before! We have to fix that in hie-bios!

You can do that if you'd like! 

The code that eventually produces the error message is this: https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L107

It is a bit tricky to follow, but essentially, we take a [`CabalMulti`](https://github.com/haskell/hie-bios#cabal) and wrap it in a MultiCradle again. However, we know at Cradle.hs#L107 already that it will be a cabal cradle.

After L107, override the `runCradle` action of the produced Cradle and look at how the `cabalAction` finds the cradle dependencies: https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L702C26-L702C26. I think, we will be able to get a better error message that way.

seanhess:
I'd be happy to take a look at it, if you can help me figure out how to test hie-bios locally. I got it compiling. Is there a way to run hie-bios on my error test project and get it to generate this error? My naive attempts didn't work. 

Even better, I would like to get haskell-language-server to compile against a local hie-bios, but when I tried this on the latest master for both, they didn't compile. Any guidance? I can post the error here if that would help

fendor:
>  Is there a way to run hie-bios on my error test project and get it to generate this error? My naive attempts didn't work.

Install the `hie-bios` binary and run `hie-bios debug ./app/A.hs`. This should show some logs and produce the final error message.

> Even better, I would like to get haskell-language-server to compile against a local hie-bios, but when I tried this on the latest master for both, they didn't compile. Any guidance? I can post the error here if that would help

It is possible, that we changed something, and now it doesn't compile any more... I haven't looked into it yet, I am afraid. 

Perhaps you want to `git checkout` the latest release of hie-bios and work off that? There haven't been many changes, so if you get something working, it is trivial to get it merged upstream.

seanhess:
I think I'm in over my head. Here's the issue I'm running in to:

```
src/Hie/Implicit/Cradle.hs:67:39: error: [GHC-83865]
    • Couldn't match type ‘Maybe FilePath -> CabalType’
                     with ‘CabalType’
      Expected: Name -> Component -> (FilePath, CabalType)
        Actual: Name
                -> Component -> (FilePath, Maybe FilePath -> CabalType)
    • In the second argument of ‘build’, namely ‘cabalComponent'’
      In the expression:
        build (CabalMulti mempty) cabalComponent' cabalPkgs
      In an equation for ‘cabal’:
          cabal = build (CabalMulti mempty) cabalComponent' cabalPkgs
   |
67 |     cabal = build (CabalMulti mempty) cabalComponent' cabalPkgs
   |                                       ^^^^^^^^^^^^^^^
```

[I found this related comment in flake.nix](https://github.com/haskell/haskell-language-server/blob/master/flake.nix#L77). I have VERY limited exposure to nix, but I think I'm getting the error for the same reason mentioned in the comment:

```
    # smunix: github:haskell/hie-bios defines
    #   'CabalType :: Maybe String -> Maybe FilePath -> CabalType'
    # while the original githcom:Avi-D-coder/hie-bios still has this:
    #   'CabalType :: Maybe String -> CabalType'
    # We need a patched version of implicit-hie-cradle that works with hls, so I've created
    # the repository below. Obviously, this is not sustainable as it adds more technical debt.
    # We need a better strategy to streamline changes required by HLS from other hie-bios related
    # packages.
    # See details here: https://github.com/Avi-D-coder/implicit-hie-cradle/compare/master...smunix:implicit-hie-cradle:smunix-patch-hls-1?expand=1
```

fendor:
Yeah, that was something we changed. So, essentially, clone implicit-hie pass in `Nothing` for this new parameter and point the cabal.project in hls to your patched implicit-hie version. You can refer to relative paths in `cabal.project`, so for example: `packages: ../hie-bios ../implicit-hie ...`, where your forks for `hie-bios` and `implicit-hie` are in the same directory as `hls`.

However, checking out an older hie-bios (the latest released tag) is likely quicker.

seanhess:
@fendor I think I'm not going to have time to address the hie-bios changes ATM. Would you consider this PR without them? What else should change? Final error output is:

```
  Loading the module 'app/Test.hs' failed. It seems that it is not listed in your .cabal file!
  Perhaps you need to add `Test` to other-modules or exposed-modules
  For more information, visit:
  https://cabal.readthedocs.io/en/3.4/developing-packages.html#modules-included-in-the-package

  app/Main.hs - Cabal {component = Just "hls-file-error:exe:hls-file-error"})
  ...
```

fendor:
@seanhess Yes, we will take the PR as is. Our HSoC student @VeryMilkyJoe can take it over to push it over the finish line, maybe try to fix the hie-bios issue in the process as well.

Regarding the error message, I would ask you to write down what kind of information you would have needed to understand/be able to resolve the issue you encountered in #3693? Is the given information enough?

seanhess:
Hey @fendor, anything short of cracking open hie-bios I can do to help get this merged?

fendor:
> Hey fendor, anything short of cracking open hie-bios I can do to help get this merged?

I have it on my to-do list for the next weeks to improve the hie-bios error messages, and also get this merged.
@VeryMilkyJoe will also need it for their work for the hls-cabal-plugin. 

From my perspective, you don't need to do anything, it is a matter of plugging everything together, which requires some work and coordination.

fendor (approved):
LGTM! First step to help new users. Many more improvements are yet to follow.

Issue #3473, opened 2023-01-31
The code action to import a qualified identifier does not produce a qualified import anymore


Kleidukos:
### Your environment

Which OS do you use? Fedora 36

Which version of GHC do you use and how did you install it? GHC 9.4.4 via ghcup

How is your project built (alternative: link to the project)? cabal

Which LSP client (editor/plugin) do you use? nvim_lsp

Which version of HLS do you use and how did you install it? latest master via ghcup

Have you configured HLS in any way (especially: a `hie.yaml` file)? no

### Steps to reproduce

<details>
<summary>Reproducer code</summary>

```haskell
module Reproducer where

operation :: [a] -> [b]
operation = List.map
```
</details>

<details>
<summary>screenshot of the suggestions</summary>

![Screenshot_20230131_155908](https://user-images.githubusercontent.com/29253044/215796110-a243cb23-72d3-45d9-89fa-15083b943a9f.png)
</details>

### Expected behaviour

HLS used to be able to suggest qualified imports for qualified identifiers

### Actual behaviour

HLS is not able to suggest qualified imports for qualified identifiers anymore

### Debug information

No idea what to include, please advise.

wz1000:
In particular https://gitlab.haskell.org/ghc/ghc/-/issues/22130

Kleidukos:
Okay! So I have two questions:

1. Is it possible to convey this information in the UI between the time where this is discovered and fixed?

2. Shouldn't https://haskell-language-server.readthedocs.io/en/latest/support/plugin-support.html#current-plugin-support-tiers be heavily updated? 
One of the reasons that made me switch to GHC 9.4 for everyday development was seeing that only eval, stan, haddock and tactics plugins were unsupported, on the `latest` version of the documentation.

michaelpj:
> Is it possible to convey this information in the UI between the time where this is discovered and fixed?

Not AFAIK. How would you convey it?

> Shouldn't https://haskell-language-server.readthedocs.io/en/latest/support/plugin-support.html#current-plugin-support-tiers be heavily updated?

We don't distinguish between "working at all for a GHC version" and "working but with caveats". We could do, but we can barely keep up with the information as it is.

Kleidukos:
@michaelpj now I'm confused, isn't the whole purpose of Tier 2 to signify "it works but with caveats"??

michaelpj:
[...]

PR #3712, opened 2023-07-12
Best-effort support of Qualified Imports in GHC 9.4

konn:
As discussed in #3473, we cannot provide `import qualified` feature in GHC 9.4 due to the [regression in GHC](https://gitlab.haskell.org/ghc/ghc/-/issues/22130), at least based on the textual information given by error messages.

On the other hand, we can still determine qualified module name from the information of error-range and textual source code.
As I am heavily using `Import Qualified` feature everyday life, so I try to follow this path to tentatively support it in GHC 9.4.
Even though there can be some corner cases that the method implemented in this PR cannot handle correctly, but even a partial support of this could be still helpful.

# TODOs

- [x] Check if this works with multibyte symbols
- ~~(Optional) revive `adding new import item to existing list` along this line~~ This cannot be handled as GHC 9.4 won't give us suggested imports.


konn:
It seems that [regression #22130 in GHC](https://gitlab.haskell.org/ghc/ghc/-/issues/22130) removes import suggestion for qualified symbols. This cannot be fixed even with our approach, so I decided to concentrate on reviving `add qualified import` feature in this PR.

konn:
I think this PR is ready for review now.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1473,7 +1476,7 @@ suggestNewOrExtendImportForClassMethod packageExportsMap ps fileContents Diagnos
         where moduleText = moduleNameText identInfo
 
 suggestNewImport :: DynFlags -> ExportsMap -> Annotated ParsedSource -> T.Text -> Diagnostic -> [(T.Text, CodeActionKind, TextEdit)]
-suggestNewImport df packageExportsMap ps fileContents Diagnostic{_message}
+suggestNewImport df packageExportsMap ps fileContents Diagnostic{_message, ..}
```
fendor, line 1479: I think it would be better if we decide on one style or the other within the same function. Personally, I don't mind either style.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1482,17 +1485,62 @@ suggestNewImport df packageExportsMap ps fileContents Diagnostic{_message}
...
-          (constructNewImportSuggestions packageExportsMap (qual <|> qual', thingMissing) extendImportSuggestions qis) in
+          (constructNewImportSuggestions packageExportsMap (qual <|> qual' <|> qualGHC94, missing) extendImportSuggestions qis) in
     map (\(ImportSuggestion _ kind (unNewImport -> imp)) -> (imp, kind, TextEdit range (imp <> "\n" <> T.replicate indent " "))) suggestions
   where
     L _ HsModule {..} = astA ps
 suggestNewImport _ _ _ _ _ = []
 
+-- tentative workaround for detecting qualification in GHC 9.4
+-- FIXME: We can delete this after dropping the support for GHC 9.4
+extractQualifiedModuleNameFromMissingName :: T.Text -> Maybe T.Text
```
fendor, line 1513: Please, add a small haddock comment.
I had good experience with code examples like this: https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L748
Gives you an intuition what the function does without having to read it.

Also, perhaps the link to the issue would be great.

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -1670,10 +1670,11 @@ suggestImportTests = testGroup "suggest import actions"
...
     , test True []          "f = (&)"                     []                "import Data.Function ((&))"
-    , ignoreForGHC94 "On GHC 9.4 the error message doesn't contain the qualified module name: https://gitlab.haskell.org/ghc/ghc/-/issues/20472"
-      $ test True []          "f = NE.nonEmpty"             []                "import qualified Data.List.NonEmpty as NE"
-    , ignoreForGHC94 "On GHC 9.4 the error message doesn't contain the qualified module name: https://gitlab.haskell.org/ghc/ghc/-/issues/20472"
-      $ test True []          "f = Data.List.NonEmpty.nonEmpty" []            "import qualified Data.List.NonEmpty"
+    , test True []          "f = NE.nonEmpty"             []                "import qualified Data.List.NonEmpty as NE"
+    , test True []          "f = (NE.:|)"             []                "import qualified Data.List.NonEmpty as NE"
+    , test True []          "f = (Data.List.NonEmpty.:|)" []            "import qualified Data.List.NonEmpty"
+    , test True []          "f = (B..|.)"             []                "import qualified Data.Bits as B"
+    , test True []          "f = (Data.Bits..|.)" []            "import qualified Data.Bits"
```
fendor, line 1677: While we are here, can you make sure the indentation looks correct again?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -223,11 +224,27 @@ usePropertyLsp kn pId p = do
 
 -- ---------------------------------------------------------------------
 
-extractRange :: Range -> T.Text -> T.Text
-extractRange (Range (Position sl _) (Position el _)) s = newS
+-- | Extracts lines in the text overlapping the given range.
+--
+-- See also: 'extractTextInRange'
+extractOverlappingLinesWithRange :: Range -> T.Text -> T.Text
+extractOverlappingLinesWithRange (Range (Position sl _) (Position el _)) s = newS
```
fendor, line 231: I am not sure if I like the name, but I can't think of anything else. @michaelpj any ideas?

fendor (approved):
LGTM! Thank you very much, this is a great addition!

Small documentation things, otherwise, gtg.

hls-plugin-api/src/Ide/PluginUtils.hs
michaelpj: If it were me, I would be tempted to instead provide `expandToLines :: Range -> Range` and make `extractTextInRange` exact.

plugins/hls-refactor-plugin/hls-refactor-plugin.cabal
```diff
@@ -86,6 +86,9 @@ library
     , lens
     , data-default
     , time
+    -- FIXME: Only needed to workaround for qualified imports in GHC 9.4
```
michaelpj, line 89: Do you want to guard the inclusion of these with a cabal conditional to make it explicit?

hls-plugin-api/src/Ide/PluginUtils.hs
fendor: Perhaps `extractLinesInRange`?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1482,17 +1485,62 @@ suggestNewImport df packageExportsMap ps fileContents Diagnostic{_message}
         >>= (findImportDeclByModuleName hsmodImports . T.unpack)
         >>= ideclAs . unLoc
         <&> T.pack . moduleNameString . unLoc
+  , -- tentative workaround for detecting qualification in GHC 9.4
+    -- FIXME: We can delete this after dropping the support for GHC 9.4
+    qualGHC94 <-
+        guard (ghcVersion == GHC94)
```
michaelpj, line 1491: alternatively we could use some CPP. that would let us only have the dependencies conditionally

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
michaelpj: :100: for examples

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
fendor: Why CPP when this works just as well? CPP is always akward, to format, to read... I think this is great the way it is.

plugins/hls-refactor-plugin/hls-refactor-plugin.cabal
fendor: I think that's not worth the trouble. What are a couple of not 100% necessary dependencies that we likely depend on already, in comparison to a cabal flag?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
michaelpj: The only reason would be to make the extra dependencies conditional, but I agree it's probably not worth it.

hls-plugin-api/src/Ide/PluginUtils.hs
fendor: Migrating to Rope would be nice, since we would presumably speed up the formatting code actions for selections :)

michaelpj: It's not necessarily just fine. This is in a code action, and I don't think we have resolve support here. So this may make the code action request grindingly slow on large files on 9.4 if there is an out-of-scope identifier. Maybe we're okay with that but it's potentially a real regression. I'll make a ticket anyway.

michaelpj: https://github.com/haskell/haskell-language-server/issues/3720

fendor: Fair enough.

konn:
@fendor @michaelpj Thank you for your reviews and sorry for my late response! I had been a bit busy this week. I'm starting responding to the reviews from now.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1482,17 +1485,62 @@ suggestNewImport df packageExportsMap ps fileContents Diagnostic{_message}
...
   , Just (range, indent) <- newImportInsertRange ps fileContents
   , extendImportSuggestions <- matchRegexUnifySpaces msg
     "Perhaps you want to add ‘[^’]*’ to the import list in the import of ‘([^’]*)’"
   = let qis = qualifiedImportStyle df
+        -- FIXME: we can use thingMissing once the support for GHC 9.4 is dropped.
+        missing
+            | GHC94 <- ghcVersion
+            , isNothing (qual <|> qual')
+            , Just q <- qualGHC94 =
+                mapNotInScope ((q <> ".") <>) thingMissing
```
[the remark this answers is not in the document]
konn: We need it because the remaining logic assumes `missing` (formerly `thingMissing`) to be qualified name.
I added a note in a8ecc929e39d4a5b66161ce50e340759e8ef39fa.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
konn: Added in 02cc20110d91e71e55c4a036202daf4f32f8aa69.
Eval plugin didn't work for some reasons, so I evaluated example lines with GHCi manually.

hls-plugin-api/src/Ide/PluginUtils.hs
konn: I followed the suggestion from @michaelpj in b2c5f78513b65b1f0efe2936f1a950b993870450.
Migration to `Rope` (#3720) sounds promising. It seems like a change affecting wider range than the initial purpose of this PR, so I would rather leave it to the future refactoring (by someone else, or myself if the time allows :grin:).

plugins/hls-refactor-plugin/hls-refactor-plugin.cabal
konn: Given that `regex-applicative` is stable and somewhat lightweight dependency, I would leave the dependency as-is.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
konn: Good catch! I changed it to use RecordWildCards alone in d0e5765af1803fe8fd4fb3ee687336aac6d83fa0.

plugins/hls-refactor-plugin/test/Main.hs
konn: Addressed in 9e1a23097581e05afff2daed660f9f8ccbcb1b5b.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
konn: As I decided to keep dependency for `regex-applicative` as-is, I also left the import unchanged.
I don't like conditional import so much as it also makes `add import` code action behaving awkwardly; e.g. adding import inside CPP conditional block in some cases even when meant as an unconditional import!

konn: Oops, this thread was meant for logic not import - sorry for my misreading. But again, I don't like CPP so much when we can use source-level conditional instead with a little overhead. In this case I expect (but not verified) compilers to optimise away this conditional.

konn:
I think I addressed all the reviews so far. Thank you for your feedbacks! I mark some threads that as resolved with some changes, but please feel free to unresolve threads if you are unsatisfied with changes.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1740,6 +1815,11 @@ data NotInScope
     | NotInScopeThing T.Text
     deriving Show
 
+mapNotInScope :: (T.Text -> T.Text) -> NotInScope -> NotInScope
+mapNotInScope f (NotInScopeDataConstructor d) = NotInScopeDataConstructor (f d)
+mapNotInScope f (NotInScopeTypeConstructorOrClass d) = NotInScopeTypeConstructorOrClass (f d)
+mapNotInScope f (NotInScopeThing d) = NotInScopeThing (f d)
```
fendor, line 1821: Is this function used for anything else? If not, we can likely turn it into a local function (e.g. in the where block) and remove the parameter f with the thing it is supposed to do (adding a ".")

fendor (approved):
LGTM, thank you very much!

Two small comments

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
konn: Good point! I made it local, changed to more speic name `qualified`, and take qualifier `q` instead of general mapping function `f`.
Addressed in e0eb90a159d0ede5c8b6aab106a79d2d994f1b7f.

hls-plugin-api/src/Ide/PluginUtils.hs
michaelpj: Yes, definitely out of scope for this PR!

konn:
Test failure with GHC _9.2_ on Win seems like a genuine regression:

~~~
Test suite func-test: RUNNING...
haskell-language-server
  code actions
    redundant import code actions
      remove solitary redundant imports: FAIL (6.48s)
        test\functional\FunctionalCodeAction.hs:259:
        expected: Nothing
         but got: Just (Command {_title = "Execute Code Action", _command = "4884:hlint:codeActionResolve", _arguments = Just [Object (fromList [("data",Object (fromList [("_uri",String "file:///D:/a/haskell-language-server/haskell-language-server/test/testdata/redundantImportTest/src/CodeActionRedundant.hs"),("_value",Object (fromList [("ignoreHintTitle",String "Use module export list"),("tag",String "IgnoreHint"),("verTxtDocId",Object (fromList [("uri",String "file:///D:/a/haskell-language-server/haskell-language-server/test/testdata/redundantImportTest/src/CodeActionRedundant.hs"),("version",Number 0.0)]))]))])),("diagnostics",Array [Object (fromList [("code",String "Use module export list"),("message",String "Use module export list\nFound:\n  module CodeActionRedundant where\nWhy not:\n  module CodeActionRedundant (\n        module CodeActionRedundant\n    ) where\nan explicit list is usually better\n"),("range",Object (fromList [("end",Object (fromList [("character",Number 0.0),("line",Number 0.0)])),("start",Object (fromList [("character",Number 0.0),("line",Number 0.0)]))])),("severity",Number 3.0),("source",String "hlint")])]),("isPreferred",Bool False),("kind",String "quickfix"),("title",String "Ignore hint \"Use module export list\" in this module")])]})
~~~

https://github.com/haskell/haskell-language-server/blob/8c22b84f897ef6d0c35bc6b53eb7911c478b5473/test/functional/FunctionalCodeAction.hs#L259

https://github.com/haskell/haskell-language-server/actions/runs/5629658777/job/15256519280?pr=3712#step:10:211

It seems rather strange as all other versions of GHC on Win passes tests. 
Does anybody guess the root cause, or is it just a flaky test?

fendor:
I presume flaky error, I rerun the failing CI run.

konn:
@fendor Now it seems that all `Testing`s are green even with GHC 9.4 on Win. Thanks!

It smees Nix build on ubuntu also failed, but it seems due to the storage shortage. So I suppose everything is OK, right?

fendor:
@konn CI and I are happy, but you'll need to rebase the PR 😅

PR #3713, opened 2023-07-12
Actually force usages

wz1000:
In 00f4e61da6c07f9a632e58dd57c73a79a1640fa3 we attempted to share the filepaths in GHC's usage field.
Unfortunately the `mi_usages` field is lazy and so these turned out to not actually be shared because
the thunk responsible for sharing was never evaluated.

Add the right strictness annotations to ensure these are actually shared


michaelpj:
Seems to not build on all GHCs

michaelpj (approved):

michaelpj:
Still doesn't build.

PR #3714, opened 2023-07-12
Move Recorder to hls-plugin-api

joyfulmantis:
Moves the Recorder to hls-plugin-api, and implements logging in the Ide.Plugin.Resolve module.


michaelpj:
As I said on IRC, I think a bunch of the functions in `hls-plugin-api` where we want to log are already in `LspM` and so `MonadIO`. For `combineResponses` yeah, we might want to use something else.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -61,12 +61,14 @@ importCommandId = "ImportLensCommand"
...
   | LogWAEResponseError ResponseError
-  deriving Show
+  | forall a. (Pretty a) => LogResolve a
+
 
 instance Pretty Log where
   pretty = \case
     LogShake logMsg -> pretty logMsg
     LogWAEResponseError rspErr -> "RequestWorkspaceApplyEdit Failed with " <+> viaShow rspErr
+    LogResolve msg -> "ResolveHelpers:" <+> pretty msg
```
michaelpj, line 71: in situations like this we usually don't add a herald at the front, like with `LogShake`. Otherwise it gets very busy

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -162,6 +163,7 @@ instance Pretty Log where
     LogGeneratedIdeas fp ideas -> "Generated hlint ideas for for" <+> viaShow fp <> ":" <+> viaShow ideas
     LogUsingExtensions fp exts -> "Using extensions for " <+> viaShow fp <> ":" <+> pretty exts
     LogGetIdeas fp -> "Getting hlint ideas for " <+> viaShow fp
+    LogResolve msg -> "ResolveHelpers:" <+> pretty msg
```
michaelpj, line 166: ditto

michaelpj: and in other places but I'll shut up now

hls-plugin-api/src/Ide/Plugin/Resolve.hs
```diff
@@ -29,44 +31,61 @@ import           Language.LSP.Server           (LspM, LspT,
...
-              A.Error err -> throwE $ parseError (Just value) (T.pack err)
-              A.Success innerValueDecoded -> do
-                resolveResult <- ExceptT $ codeResolveMethod ideState pid codeAction uri innerValueDecoded
-                case resolveResult of
-                  CodeAction {_edit = Just _ } -> do
-                      pure $ InR $ dropData resolveResult
-                  _ -> throwE $ invalidParamsError "Returned CodeAction has no data field"
-        resolveCodeAction _ _ _ (InR CodeAction{_data_=Nothing}) = throwE $ invalidParamsError "CodeAction has no data field"
+          case A.fromJSON value of
+            A.Error err -> logAndThrow recorder LogParseError $ parseError (Just value) (T.pack err)
```
michaelpj, line 80: I'm unsure if this is the best place to log. _Hopefully_ caught exceptions should be logged in general, so I think it's probably overkill to always log before throwing, then I think we'll get things twice?

hls-plugin-api/src/Ide/Plugin/Resolve.hs
```diff
@@ -29,44 +31,61 @@ import           Language.LSP.Server           (LspM, LspT,
...
-        dropData :: CodeAction -> CodeAction
+mkCodeActionHandlerWithResolve recorder codeActionMethod codeResolveMethod =
+  let newCodeActionMethod ideState pid params = runExceptT $
+        do codeActionReturn <- ExceptT $ codeActionMethod ideState pid params
+           caps <- lift getClientCapabilities
+           case codeActionReturn of
+             r@(InR Null) -> pure r
+             (InL ls) | -- If the client supports resolve, we will wrap the resolve data in a owned
+                        -- resolve data type to allow the server to know who to send the resolve request to
+                        supportsCodeActionResolve caps -> pure $ InL ls
```
michaelpj, line 69: Somewhere around here we might want to log at the lowest level to record what decision we're making about whether or not to try and use resolve?

hls-plugin-api/src/Ide/Plugin/Resolve.hs
```diff
@@ -113,27 +133,26 @@ mkCodeActionWithResolveAndCommand plId codeActionMethod codeResolveMethod =
...
-                    resolveResult <- resolveProvider ideState plId ca uri innerValueDecoded
+                    resolveResult <- ExceptT $ resolveProvider ideState plId ca uri innerValueDecoded
                     case resolveResult of
-                      Right ca2@CodeAction {_edit = Just wedits } | diffCodeActions ca ca2 == ["edit"] -> do
-                          _ <- sendRequest SMethod_WorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing wedits) (\_ -> pure ())
-                          pure $ Right $ InR Null
-                      Right ca2@CodeAction {_edit = Just _ }  ->
-                        pure $ Left $
+                      ca2@CodeAction {_edit = Just wedits } | diffCodeActions ca ca2 == ["edit"] -> do
+                          _ <- ExceptT $ Right <$> sendRequest SMethod_WorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing wedits) (\_ -> pure ())
```
michaelpj, line 147: log in case `applyEdit` failed?

michaelpj (approved):
I think we just need to think about what logging policy we want here. As I said in the comment, I _think_ we shouldn't need to additionally log if we are throwing, since I would expect those all to be recorded generically. So then the question is when _do_ we want to log? I think one answer is "if we make a decision that might not be obvious to the user", or if something interesting or unexpected happens and we're _not_ throwing.

hls-plugin-api/src/Ide/Plugin/Resolve.hs
joyfulmantis: Right

PR #3717, opened 2023-07-14
Better plugin error infrastructure

joyfulmantis:
This PR is based on fendors, he did a lot of the initial work originally in PR #3659. I finished up what he started, and then tried a slightly different approach, where I deemphasize hierarchical errors and instead try to define the common "errors" that occur in plugins centrally in the hls-plugin-api.
What I have also done, is have all method handlers now return PluginError instead of ResponseError,  this allows us to define specific error types that correspond to concrete error responses we want to send to the client and specific ways we want to log the error. 
This also allows us to keep all the ExceptT infrastructure the fendor has written, give much more accurate error response codes to the server, and log non-serious errors such as rule failures with info or even debug, while still being able to log more serious errors such as uncaught exceptions or bugs in the client or plugin with error.


joyfulmantis:
So I am pretty happy with the state of this PR, as I think it gives us some pretty concrete benefits, namely:
- ExceptT infrastructure for plugin handlers
-  Much more accurate error response codes to the server (that are controlled in one place)
-  Ability to log each error type differently

The changes in the plugins here are relatively superficial, only converting the most obvious failures to the right PluginError types, but without going through and thoroughly understanding every plugin there's not much more we can do. What we can say is that we have no regression in handling errors, as in the worst case we return a InternalError response code -- which is previously what we did for everything.

Hierarchical errors are still supported, and an example of them being used is in plugins/hls-gadt-plugin/src/Ide/Plugin/GADT.hs . In general, if you are only going to throw a custom error once, the only benefit is you get to move writing your error messages to a separate section of your file, but I could imagine that if you wanted to throw your own custom error multiple times in your codebase it could become useful. Again all custom errors need to eventually be converted to the standard PluginError type which reduces its usefulness.

HLS will now respond to rule and dependency failures with ContentModified, which is an incorrect use of the error code. However, we have our hands tied, as the method we would want, SeverCancelled, is only supported in 3.17, (and only with a per method capabilities flag! (that is currently only supported by SemanticCapabilities and Diagnostics!!)). Even though technically ContentModefied is the incorrect response here, it elicits the client response we want with VsCode, and should be with other lsp clients too. 

Most method responses allow us to return Null, so the other option for solving this quandary would be to create a type family that provided a method-appropriate response for these sorts of situations. The type family would still have to fall back to throwing some sort of error when encountering a method response that doesn't allow for null responses.

michaelpj:
What happens if exceptions get thrown in Action now?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -43,17 +43,13 @@ import           Language.LSP.Server
...
-            $ runAction "classplugin.addMethodPlaceholders.GetParsedModule" state
-            $ use GetParsedModule nfp
-        (hsc_dflags . hscEnv -> df) <- handleMaybeM "Unable to GhcSessionDeps"
-            $ liftIO
-            $ runAction "classplugin.addMethodPlaceholders.GhcSessionDeps" state
-            $ use GhcSessionDeps nfp
-        (old, new) <- handleMaybeM "Unable to makeEditText"
+    runExceptT $ do
+        nfp <- getNormalizedFilePathE (verTxtDocId ^. L.uri)
+        pm <- runActionE "classplugin.addMethodPlaceholders.GetParsedModule" state
```
michaelpj, line 48: these weird strings that we pass in are also a bit suboptimal. not sure what to do about it, just noticing them today

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
```diff
@@ -154,7 +116,7 @@ getSelectionRanges file positions = do
              in fromMaybe defaultSelectionRange . findPosition pos $ codeRange
 
     -- 'positionMapping' should be applied to the output ranges before returning them
-    maybeToExceptT SelectionRangeOutputPositionMappingFailure . MaybeT . pure $
+    maybeToExceptT (PluginDependencyFailed "toCurrentSelectionRange") . MaybeT . pure $
```
michaelpj, line 119: wrong error kind?

plugins/hls-code-range-plugin/test/Main.hs
```diff
@@ -48,6 +48,7 @@ selectionRangeGoldenTest testName positions = goldenGitDiff testName (testDataDi
         let res = resp ^. result
         pure $ fmap (showSelectionRangesForTest . absorbNull) res
     case res of
+        Left (ResponseError (InL LSPErrorCodes_ContentModified) _ _) -> pure ""
```
michaelpj, line 51: seems suspicious????

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Util.hs
```diff
@@ -66,13 +68,15 @@ logLevel = Debug -- Info
...
 isLiterate x = takeExtension x `elem` [".lhs", ".lhs-boot"]
 
-response' :: ExceptT String (LspM c) WorkspaceEdit -> LspM c (Either ResponseError (Value |? Null))
+response' :: ExceptT PluginError (LspM c) WorkspaceEdit -> LspM c (Either PluginError (Value |? Null))
 response' act = do
     res <- runExceptT act
-             `catchAny` showErr
+             `catchAny` \e -> do
+                res <- showErr e
+                pure . Left . PluginInternalError $ fromString res
```
michaelpj, line 76: ditto

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Util.hs
```diff
@@ -66,13 +68,15 @@ logLevel = Debug -- Info
...
     case res of
```
michaelpj, line 77: this is literally the identity now?

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Util.hs
```diff
@@ -81,28 +85,28 @@ gStrictTry :: (MonadIO m, MonadCatch m) => m b -> m (Either String b)
...
     catch
         (op >>= fmap Right . gevaluate)
-        showErr
+        (fmap Left . showErr)
 
 gevaluate :: MonadIO m => a -> m a
 gevaluate = liftIO . evaluate
 
-showErr :: Monad m => SomeException -> m (Either String b)
+showErr :: Monad m => SomeException -> m String
```
michaelpj, line 93: :+1:

plugins/hls-explicit-fixity-plugin/src/Ide/Plugin/ExplicitFixity.hs
```diff
@@ -45,14 +43,14 @@ descriptor recorder pluginId = (defaultPluginDescriptor pluginId)
...
-hover state _ (HoverParams (TextDocumentIdentifier uri) pos _) = pluginResponse $ do
-    nfp <- getNormalizedFilePath uri
-    handleMaybeM "ExplicitFixity: Unable to get fixity" $ liftIO $ runIdeAction "ExplicitFixity" (shakeExtras state) $ runMaybeT $ do
-      (FixityMap fixmap, _) <- useE GetFixity nfp
-      (HAR{hieAst}, mapping) <- useE GetHieAst nfp
+hover state _ (HoverParams (TextDocumentIdentifier uri) pos _) = runExceptT $ do
+    nfp <- getNormalizedFilePathE uri
+    runIdeActionE "ExplicitFixity" (shakeExtras state) $ do
+      (FixityMap fixmap, _) <-  useWithStaleFastE GetFixity nfp
+      (HAR{hieAst}, mapping) <- useWithStaleFastE GetHieAst nfp
```
michaelpj, line 50: :+1:

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -99,10 +100,10 @@ provider recorder plId ideState typ contents fp fo = withIndefiniteProgress titl
                         pure . Right $ InL $ makeDiffTextEdit contents out
                     ExitFailure n -> do
                         logWith recorder Info $ StdErr err
-                        pure . Left . responseError $ "Fourmolu failed with exit code " <> T.pack (show n)
+                        pure . Left . PluginInternalError $ "Fourmolu failed with exit code " <> T.pack (show n)
```
michaelpj, line 103: another good place for `throwError` IMO

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -134,13 +135,12 @@ provider recorder plId ideState typ contents fp fo = withIndefiniteProgress titl
                                 { _type_ = MessageType_Error
                                 , _message = errorMessage
                                 }
-                        return . Left $ responseError errorMessage
+                        return . Left $ PluginInternalError errorMessage
```
michaelpj, line 138: etc.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -512,8 +513,8 @@ ignoreHint _recorder ideState nfp verTxtDocId ignoreHintTitle = do
                   (Just (M.singleton (verTxtDocId ^. LSP.uri) textEdits))
                   Nothing
                   Nothing
-        pure $ Right workspaceEdit
-    Nothing -> pure $ Left  "Unable to get fileContents"
+        pure workspaceEdit
+    Nothing -> throwE $ PluginInternalError "Unable to get fileContents"
```
michaelpj, line 517: Ah, I think all `throwE`s should be `throwError`s. Generally it's pretty good to use the `MonadError` stuff, it's very standard and people will recognize it.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -776,7 +776,7 @@ instance Semigroup (PluginNotificationHandlers a) where
 instance Monoid (PluginNotificationHandlers a) where
   mempty = PluginNotificationHandlers mempty
 
-type PluginMethodHandler a m = a -> PluginId -> MessageParams m -> LspM Config (Either ResponseError (MessageResult m))
+type PluginMethodHandler a m = a -> PluginId -> MessageParams m -> LspM Config (Either PluginError (MessageResult m))
```
michaelpj, line 779: this could definitely end in `ExceptT PluginError (LspM Config) (MessageResult m)`, I think

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -751,7 +751,7 @@ instance GCompare IdeNotification where
 
 -- | Combine handlers for the
 newtype PluginHandler a (m :: Method ClientToServer Request)
-  = PluginHandler (PluginId -> a -> MessageParams m -> LspM Config (NonEmpty (Either ResponseError (MessageResult m))))
```
michaelpj, line 754: this is awkward because of the `NonEmpty` :thinking:

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -776,7 +776,7 @@ instance Semigroup (PluginNotificationHandlers a) where
 instance Monoid (PluginNotificationHandlers a) where
   mempty = PluginNotificationHandlers mempty
 
-type PluginMethodHandler a m = a -> PluginId -> MessageParams m -> LspM Config (Either ResponseError (MessageResult m))
+type PluginMethodHandler a m = a -> PluginId -> MessageParams m -> LspM Config (Either PluginError (MessageResult m))
 
 type PluginNotificationMethodHandler a m = a -> VFS -> PluginId -> MessageParams m -> LspM Config ()
```
michaelpj, line 781: at the plugin level, I feel like this should be able to return an error also! We won't send it back as a responseError, but we might still want to log it etc.! And surely the notification handlers would also like to use the helpers we're defining that work in `ExceptT`

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -924,10 +924,10 @@ mkResolveHandler m f = mkPluginHandler m $ \ideState plId params -> do
             let newParams = params & L.data_ ?~ value
             in f ideState plId newParams uri decodedValue
           Error err ->
-            pure $ Left $ ResponseError (InR ErrorCodes_ParseError) (parseError value err) Nothing
-      else pure $ Left $ ResponseError (InR ErrorCodes_InternalError) invalidRequest Nothing
-    (Just (Error err)) -> pure $ Left $ ResponseError (InR ErrorCodes_ParseError) (parseError (params ^. L.data_) err) Nothing
-    _ -> pure $ Left $ ResponseError (InR ErrorCodes_InternalError) invalidRequest Nothing
+            pure $ Left $ PluginParseError (parseError value err)
+      else pure $ Left $ PluginInternalError invalidRequest
```
michaelpj, line 928: invalid request code?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -924,10 +924,10 @@ mkResolveHandler m f = mkPluginHandler m $ \ideState plId params -> do
...
+    (Just (Error err)) -> pure $ Left $ PluginParseError (parseError (params ^. L.data_) err)
+    _ -> pure $ Left $ PluginInternalError invalidRequest
```
michaelpj, line 930: invalid request code?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+import           Control.Monad.Trans.Maybe   (MaybeT, mapMaybeT)
+import qualified Data.Text                   as T
+import           Language.LSP.Protocol.Types
+import           Prettyprinter
+
+-- ----------------------------------------------------------------------------
+-- Plugin Error wrapping
+-- ----------------------------------------------------------------------------
+
+pluginResponseM :: Monad m => (t -> m (Either a b)) -> ExceptT t m b -> m (Either a b)
```
michaelpj, line 31: I'm sceptical of this function: I think it would be just as clear to inline it! Because it's just `runExceptT . handleError handler`, right? Generally I think we shouldn't be afraid of letting people work with `ExceptT`/`MonadError` directly, they're very familiar interfaces.

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+pluginResponseM handler act =
+    runExceptT act >>= \case
+        Right r  -> pure $ Right r
+        Left err -> handler err
+
+-- See Note [PluginError]
+data PluginError
```
michaelpj, line 38: Worth having at least some haddock

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+    -- (otherwise they wouldn't be sending you requests with invalid
+    -- parameters).
+    --
+    -- This error will be logged individually with Warning, and will be
+    -- converted into a InvalidParams response code. It takes medium precedence
+    --  (2)in being returned as a response to the client.
+  | PluginInvalidParams T.Text
+    -- |PluginInvalidRequest should be used if the request is invalid. This
+    -- error means that there is a bug in the client's code (otherwise they
+    -- wouldn't be sending you an invalid request).
```
michaelpj, line 57: what's the difference to the previous one?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+    -- This error will be logged individually with Warning, and will be
+    -- converted into a InvalidRequest response code. It takes medium precedence
+    --  (2) in being returned as a response to the client.
+  | PluginInvalidRequest T.Text
+    -- |PluginParseError should be used sparingly for parse errors. Prefer to
+    -- use PluginInternalError for bugs in the plugin's code, or
+    -- PluginInvalidRequest/PluginInvalidParams for invalid requests from the
+    -- client.
+    --
+    -- This error will be logged individually with Warning, and will be
```
michaelpj, line 68: seems like it should be higher? this is a bug in _someone_'s code, right?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+    -- converted into a ParseError response code. It takes a medium precedence
+    --  (2) in being returned as a response to the client.
+  | PluginParseError T.Text
+    -- |PluginDependencyFailed should be thrown when a function that your plugin
```
michaelpj, line 72:
```suggestion
    -- |PluginDependencyFailed should be thrown when a rule that your plugin
```
I think?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+    -- depends on fails. This should only be used when the function fails
+    -- because the files the user is working on is in an invalid state.
+    --
+    -- This error takes the name of the function that failed. Prefer to catch
```
michaelpj, line 76:
```suggestion
    -- This error takes the name of the rule that failed. Prefer to catch
```

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+    -- this error as close to the source as possible.
+    --
+    -- This error will be logged together with other errors of the same type
+    -- with Info, and will be converted into a ContentModified response. It
+    -- takes a low precedence (3) in being returned as a response to the client.
+  | PluginDependencyFailed T.Text
+    -- |PluginRequestRefused allows your handler to inspect a request before
+    -- rejecting it. In effect it allows your plugin to act make a secondary
+    -- `pluginEnabled` decision after receiving the request. This should only be
```
michaelpj, line 85: dumb question - should we just replace the use of `pluginEnabled` with this?

hls-plugin-api/src/Ide/Plugin/Error.hs
michaelpj: wait, you also have PluginRuleFailed. I'm quite confused tbh

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+
+pluginResponseM :: Monad m => (t -> m (Either a b)) -> ExceptT t m b -> m (Either a b)
+pluginResponseM handler act =
+    runExceptT act >>= \case
+        Right r  -> pure $ Right r
+        Left err -> handler err
+
+-- See Note [PluginError]
+data PluginError
+  = -- |PluginInternalError should be used if something has gone horribly wrong.
```
michaelpj, line 39: as in, a bug in our code?

I feel like I'm missing an error for "something went wrong, but just in a way that can sometimes happen and it's okay"?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+      PluginInternalError msg -> "Internal Error:" <+> pretty msg
+      PluginStaleResolve -> "Stale Resolve"
+      PluginRuleFailed rule       -> "Rule Failed:" <+> pretty rule
+      PluginInvalidParams text -> "Invalid Params:" <+> pretty text
+      PluginParseError text -> "Parse Error:" <+> pretty text
+      PluginInvalidRequest text -> "Invalid Request:" <+> pretty text
+      PluginDependencyFailed text -> "Dependency Failed:" <+> pretty text
+      PluginRequestRefused -> "Request Refused"
+
+handleMaybe :: Monad m => e -> Maybe b -> ExceptT e m b
```
michaelpj, line 121: I have mixed feelings about these. They are quite common, but also I think people can write them inline. This one in particular really just saves people a case expression, and I would usually rather read the case expression!

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+handleMaybe msg = maybe (throwE msg) return
+
+handleMaybeM :: Monad m => e -> m (Maybe b) -> ExceptT e m b
```
michaelpj, line 124: this one is maybe worth it

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+handleMaybeM msg act = maybeM (throwE msg) return $ lift act
+
+hoistExceptT :: MonadIO m => ExceptT e IO a -> ExceptT e m a
```
michaelpj, line 127: I'd just write it inline, it's trivial

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+hoistExceptT = mapExceptT liftIO
+
+hoistMaybeT :: MonadIO m => MaybeT IO a -> MaybeT m a
```
michaelpj, line 130: ditto

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+hoistMaybeT = mapMaybeT liftIO
+
+getNormalizedFilePathE :: Monad m => Uri -> ExceptT PluginError m NormalizedFilePath
+getNormalizedFilePathE uri = handleMaybe (PluginInvalidParams (T.pack $ "uriToNormalizedFile failed. Uri:" <>  show uri))
+        $ uriToNormalizedFilePath
+        $ toNormalizedUri uri
+
+-- ---------------------------------------------------------------------
+{- Note [PluginError]
+-- Each PluginError corresponds to either a specific ResponseError we want to
```
michaelpj, line 140: The point of doing a multi-line comment is so you don't need the `--` at the start of each line :joy:

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,163 @@
...
+hoistMaybeT :: MonadIO m => MaybeT IO a -> MaybeT m a
+hoistMaybeT = mapMaybeT liftIO
+
+getNormalizedFilePathE :: Monad m => Uri -> ExceptT PluginError m NormalizedFilePath
+getNormalizedFilePathE uri = handleMaybe (PluginInvalidParams (T.pack $ "uriToNormalizedFile failed. Uri:" <>  show uri))
+        $ uriToNormalizedFilePath
+        $ toNormalizedUri uri
+
+-- ---------------------------------------------------------------------
+{- Note [PluginError]
```
michaelpj, line 139: In this case I think this would actually be pretty appropriate as the Haddock!

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,132 @@
+{-# LANGUAGE GADTs #-}
+module Development.IDE.Core.PluginUtils where
```
michaelpj, line 2: should also be in `hls-plugin-api`?

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,132 @@
...
+-- ----------------------------------------------------------------------------
+
+-- |ExceptT version of `runAction`, takes a ExceptT Action
+runActionE :: MonadIO m => String -> IdeState -> ExceptT e Action a -> ExceptT e m a
+runActionE herald ide act =
+  hoistExceptT . ExceptT $
+    join $ shakeEnqueue (shakeExtras ide) (mkDelayedAction herald Logger.Debug $ runExceptT act)
+
+-- |MaybeT version of `runAction`, takes a MaybeT Action
+runActionMT :: MonadIO m => String -> IdeState -> MaybeT Action a -> MaybeT m a
```
michaelpj, line 36: I'd be tempted to just bin the `MT` variants. If we provide people with error-producing ones, they can always throw away the errors. And I think they're not used that much?

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,132 @@
...
+-- ----------------------------------------------------------------------------
+-- IdeAction wrappers
+-- ----------------------------------------------------------------------------
+
+-- |ExceptT version of `runIdeAction`, takes a ExceptT IdeAction
+runIdeActionE :: MonadIO m => String -> Shake.ShakeExtras -> ExceptT e IdeAction a -> ExceptT e m a
+runIdeActionE _herald s i = ExceptT $ liftIO $ runReaderT (Shake.runIdeActionT $ runExceptT i) s
+
+-- |MaybeT version of `runIdeAction`, takes a MaybeT IdeAction
+runIdeActionMT :: MonadIO m => String -> Shake.ShakeExtras -> MaybeT IdeAction a -> MaybeT m a
```
michaelpj, line 69: I really need to remind myself what the difference between IdeAction and Action is :/

michaelpj (commented):
The plugin code looks _much_ nicer. 

High-level concerns:
- We still have exceptions in the mix. What's happening with them?
- Generally I think it's good to avoid _too_ many layers of helper functions that don't do much, especially when we're working with something as familiar as `ExceptT`
- I think the processing of the errors could be a little less specific.

For the last point, I think we _might_ want to switch to having a more "open" error type, e..g something like `PluginError e` with a constructor that can hold the `e`. That would let plugins throw their own specific errors. Then we'd need to abstract over what we need from plugin errors, which I think is pretty much just two functions, one to ErrorCode and one to Severity. So it would be nice to avoid doing too much pattern matching on the specific set of errors, perhaps.

joyfulmantis:
> What happens if exceptions get thrown in Action now?

I haven't changed anything in Action or Rules in this PR. I would like to dump the MonadFail and underscore functions from Rule and move most plugin's rule definitions to MaybeT but It's distinct (and possibly controversial) enough to warrant a separate PR.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -288,22 +314,110 @@ runConcurrently
...
+    (PluginId plId, err@(PluginInternalError _)) ->
+        ResponseError (InR ErrorCodes_InternalError) (plId <> ": " <> tPretty err) Nothing
+    (PluginId plId, err@(PluginInvalidParams _)) ->
+        ResponseError (InR ErrorCodes_InvalidParams) (plId <> ": " <> tPretty err) Nothing
+    (PluginId plId, err@(PluginInvalidRequest _)) ->
+        ResponseError (InR ErrorCodes_InvalidRequest) (plId <> ": " <> tPretty err) Nothing
+    (PluginId plId, err@(PluginParseError _)) ->
+        ResponseError (InR ErrorCodes_ParseError) (plId <> ": " <> tPretty err) Nothing
+    (PluginId plId, err@(PluginDependencyFailed _)) ->
+        ResponseError (InL LSPErrorCodes_ContentModified) (plId <> ": " <> tPretty err) Nothing
```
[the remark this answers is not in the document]
joyfulmantis: Yes, I mentioned my rationale for this decision in the last two paragraphs here https://github.com/haskell/haskell-language-server/pull/3717#issuecomment-1647784191

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -288,22 +314,110 @@ runConcurrently
...
+                        | (plId, PluginRequestRefused) <- errs]
+          case nonEmpty rrErrs of
+            Nothing -> pure ()
+            Just xs -> logWith recorder Info $ LogRequestRefused xs
+        logRuleFailed errs = do
+          let rfErrs = [(plId, rule)
+                        | (plId, PluginRuleFailed rule) <- errs]
+          case groupBySnd rfErrs of
+            Nothing -> pure ()
+            Just xs -> logWith recorder Info $ LogRuleFailed xs
```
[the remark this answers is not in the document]
joyfulmantis: We are doing some rewriting of the log message to allow the error messages to be combined nicely. We can log everything individually, but this way we get to cut down on the log messages (by a bit)

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -6,48 +6,48 @@ module Ide.Plugin.ChangeTypeSignature (descriptor
...
-import           Text.Regex.TDFA                ((=~))
+import           Text.Regex.TDFA                  ((=~))
 
 descriptor :: PluginId -> PluginDescriptor IdeState
 descriptor plId = (defaultPluginDescriptor plId) { pluginHandlers = mkPluginHandler SMethod_TextDocumentCodeAction (codeActionHandler plId) }
 
 codeActionHandler :: PluginId -> PluginMethodHandler IdeState 'Method_TextDocumentCodeAction
-codeActionHandler plId ideState _ CodeActionParams {_textDocument = TextDocumentIdentifier uri, _context = CodeActionContext diags _ _} = pluginResponse $ do
-      nfp <- getNormalizedFilePath uri
+codeActionHandler plId ideState _ CodeActionParams {_textDocument = TextDocumentIdentifier uri, _context = CodeActionContext diags _ _} = runExceptT $ do
```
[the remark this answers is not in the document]
joyfulmantis: Not everyone uses ExceptT, and PluginHandlers are used widely, for example also by FormattingHandler, so I think it makes sense to leave it as Either and let the plugins make the choice whether to use it or not.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
joyfulmantis: I believe they are for shake logs, but it certainly would be nice to get rid of them

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: They correlate to the ErrorCodes of the same name. I would say InvalidParams is specifically for parameters, and that InvalidRequest is broader. Mostly, I defined both because both were used in the code base.

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: Before everything was logged with Warning, and then with this update I upped the InternalError severity to Error.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -236,23 +256,29 @@ extensiblePlugins recorder xs = mempty { P.pluginHandlers = handlers }
...
+                case nonEmpty filteredErrs of
+                  Nothing -> liftIO $ noPluginEnabled m fs'
+                  Just xs -> pure $ Left $ combineErrors xs
               Just xs -> do
-                caps <- LSP.getClientCapabilities
                 pure $ Right $ combineResponses m config caps params xs
+    noPluginEnabled :: SMethod m -> [(PluginId, b, a)] -> IO (Either ResponseError c)
+    noPluginEnabled m fs' = do
+      logWith recorder Warning (LogNoPluginForMethod $ Some m)
+      let err = ResponseError (InR ErrorCodes_InvalidRequest) msg Nothing
```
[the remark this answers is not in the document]
michaelpj: Sure, I just think it's a good time to audit some of these error codes!

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: I put this Error in specifically for the scenario where someone wants to have more than one resolve hander for the same method with different data types. This error allows us to refuse to answer the ones that don't decode to the type provided by the handler signature.

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: Yes, I feel if the function you need fails if the user's state is invalid, that's where the dependency failed error comes in. If it's not because of the user's invalid state, then arguably it's a bug, and should never happen, (and can be logged with the PluginInternalError).

ghcide/src/Development/IDE/Core/PluginUtils.hs
joyfulmantis: mostly we are rewrapping around functions in gchide, and because hls-plugin-api can't import ghcide, it makes sense to put these stuff here (not sure if there's a better name for this module though)

ghcide/src/Development/IDE/Core/PluginUtils.hs
joyfulmantis: The MT functions are pretty much not used at this moment, but I hope to later do a PR for the rules, where I dump the underscored functions and convert everything to use MaybeT. I think that these functions would make sense then.

ghcide/src/Development/IDE/Plugin/HLS.hs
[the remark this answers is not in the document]
michaelpj: Ah I see. I guess I wonder if we'd be better off with InternalError or something.

ghcide/src/Development/IDE/Plugin/HLS.hs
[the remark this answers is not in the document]
michaelpj: I think at this point I'd mildly prefer more logs over more code :)

ghcide/src/Development/IDE/Plugin/Test.hs
```diff
@@ -85,12 +87,12 @@ plugin = (defaultPluginDescriptor "test") {
         = testRequestHandler ide customReq
         | otherwise
         = return $ Left
-        $ ResponseError (InR ErrorCodes_InvalidRequest) "Cannot parse request" Nothing
+        $ PluginInvalidRequest "Cannot parse request"
```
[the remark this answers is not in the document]
michaelpj: Is that not the case here? I guess this is test code so it doesn't really matter though!

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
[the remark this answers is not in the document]
michaelpj: Well, it's also easy to go from an Either to ExceptT. So I guess it's a question of whether we have more `runExceptT`s this way than we would have `ExceptT`s the other way!

hls-plugin-api/src/Ide/Plugin/Error.hs
michaelpj: Yeah, I'm just arguing that ParseError also seems like it should be Error to me

hls-plugin-api/src/Ide/Plugin/Error.hs
michaelpj: Right, but we _could_ use it to do the stuff that we now do with `pluginEnabled`, which would give us more options perhaps?

ghcide/src/Development/IDE/Core/PluginUtils.hs
michaelpj: Oh right, because of the shake stuff, got it

ghcide/src/Development/IDE/Core/PluginUtils.hs
michaelpj: Aha, that makes sense

michaelpj:
> I would like to dump the MonadFail and underscore functions from Rule and move most plugin's rule definitions to MaybeT

I do think we may be forced to accept the presence of exceptions in Action, but I'm not sure.

ghcide/src/Development/IDE/Core/PluginUtils.hs
fendor: IdeAction are read-only while Action can produce results, iirc.

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
joyfulmantis: This is actually a IO do block, so throwError won't work with PluginError here (And if it did we would end up throwing an IO exception...)

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
joyfulmantis: This is also not inside an ExceptT block, so throwError won't work. (unless we convert it to use ExceptT that is...)

plugins/hls-code-range-plugin/test/Main.hs
joyfulmantis: The test was assuming that if there is no response they should get back an empty result, but we now handle empty responses with errors instead.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
joyfulmantis: We use PluginDependencyFailed for all failed dependency besides rule failures. In this case it's the function toCurrentSelectionRange (which relies on PositionMapping)

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1003,19 +1003,15 @@ mkFormattingHandlers f = mkPluginHandler SMethod_TextDocumentFormatting ( provid
                   SMethod_TextDocumentRangeFormatting -> FormatRange (params ^. L.range)
                   _ -> Prelude.error "mkFormattingHandlers: impossible"
             f ide typ (virtualFileText vf) nfp opts
-          Nothing -> pure $ Left $ responseError $ T.pack $ "Formatter plugin: could not get file contents for " ++ show uri
+          Nothing -> pure $ Left $ PluginInvalidParams $ T.pack $ "Formatter plugin: could not get file contents for " ++ show uri
 
-      | otherwise = pure $ Left $ responseError $ T.pack $ "Formatter plugin: uriToFilePath failed for: " ++ show uri
+      | otherwise = pure $ Left $ PluginInvalidParams $ T.pack $ "Formatter plugin: uriToFilePath failed for: " ++ show uri
```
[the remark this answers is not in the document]
joyfulmantis: Yes, it's also what was used before

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: The only plugins that have notification handlers are the cabal plugin and ghcide. They do all of their work in a liftIO block, and don't use any of the wrapped methods we have provided. They also all do their own logging already. Not sure how much we would gain by allowing them to return an error.

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: We should never be getting a request that doesn't have a data field, so arguably this signifies a bug in our code.

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: I haven't looked at it extensively, but I imagine that we would lose some nice things we get by having pluginEnabled per method now.

plugins/hls-code-range-plugin/test/Main.hs
michaelpj: Hmmm, is that right? You don't mean empty result as in "the plugin correctly computed the results and it happened to be empty" right? because that probably should be returned as an empty result.

hls-plugin-api/src/Ide/Types.hs
michaelpj: Hmm okay, I guess they're just a bit weird. But it might be nice to make them less weird later...

hls-plugin-api/src/Ide/Types.hs
michaelpj: I guess that's true, since it's just been passed to us. Worth a comment maybe

hls-plugin-api/src/Ide/Types.hs
michaelpj: Worth a comment

hls-plugin-api/src/Ide/Plugin/Error.hs
michaelpj: Yeah, looking at the JSON-RPC spec again, its says that ParseError is if the JSON itself is invalid! So we shouldn't be using it at all. Probably we should indeed be using the LSP-specific codes like `RequestFailed` and so on.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -288,22 +314,110 @@ runConcurrently
...
   -> a
   -> b
-  -> m (NonEmpty(NonEmpty (Either ResponseError d)))
+  -> m (NonEmpty(NonEmpty (Either PluginError d)))
 runConcurrently msg method fs a b = forConcurrently fs $ \(pid,f) -> otTracedProvider pid (fromString (show method)) $ do
   f a b  -- See Note [Exception handling in plugins]
-     `catchAny` (\e -> pure $ pure $ Left $ ResponseError (InR ErrorCodes_InternalError) (msg pid method e) Nothing)
+     `catchAny` (\e -> pure $ pure $ Left $ PluginInternalError (msg pid method e))
+
+combineErrors :: NonEmpty (PluginId, PluginError) -> ResponseError
```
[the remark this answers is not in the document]
joyfulmantis: It could be. I considered moving it, but left it here as didn't feel the change was worthwhile enough.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
joyfulmantis: I think there should be a distinction from your plugin didn't work because the user's code is in an invalid state, and your code didn't work because there is a bug in it.

michaelpj: Okay, but that's a third thing! I think by that you mean the code buffer is in a weird state, right? So that could manifest as:
- Some rule that you need fails (`BadDependency` or whatever)
- The rules all succeed but we can't process the result for some reason. Position mapping failing seems like a good example.

I totally agree the latter is a valid category, I just think it's weird to call it "DependencyFailed"! Maybe something like:
- `InternalError` -> a bug in our code, you should file an issue
- `DependencyFailed` -> a rule dependency failed, a "normal" thing that happens often but means we can't process the request
- `HandlerFailed` -> the handler failed, but in a "normal" way that doesn't indicate a bug

ghcide/src/Development/IDE/Plugin/HLS.hs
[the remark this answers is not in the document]
joyfulmantis: So, thinking about this, one of the reasons we don't want to do MethodNotFound is it means we will get a popup every time that happens (which right now is fairly often) InvalidParams or InvalidRequest don't have that problem.

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: So the advantage of leaving InvalidParams as Warning, and InternalError as Error, is it means we can implement what you mentioned before, sorting the errors by severity and choosing InternalError over InvalidParams, which we arguably want to do.

plugins/hls-code-range-plugin/test/Main.hs
joyfulmantis: Nope, a lot of plugins used to (and some still do!) handle failures by just returning an empty result. The reason they did this was to avoid having popups when the data they wanted wasn't ready. For here the specific PR is https://github.com/haskell/haskell-language-server/pull/3229 . This is also the reason we return rule and dependency failures with ContentModified because we don't want to show popups unless there's actually an error.

ghcide/src/Development/IDE/Plugin/HLS.hs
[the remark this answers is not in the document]
joyfulmantis: So, I think the problem with InternalError is we will have a client popup whereas with the InvalidRequest/InvalidParams we won't

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
fendor: Shake logs and I think they might be used by opentelemetry to provide string representations. I don't think we should get rid of them, unless we derive the label via TH or something like that.

plugins/hls-code-range-plugin/test/Main.hs
fendor: I still don't understand, why do we produce a `LSPErrorCodes_ContentModified` for the empty result in this plugin? Is it because the only reason why we might produce an empty result is because we have modified the content before resolving `SMethod_TextDocumentSelectionRange `?

fendor (approved):
The PR looks great to me already!

Lots of code clean-ups as well, thank you very much!

fendor:
> I do think we may be forced to accept the presence of exceptions in Action, but I'm not sure.

I think Actions use AsyncExceptions for cutting off stale request, which is likely unavoidable. I don't think we should/have to remove all exceptions from `Action`, but removing them from the plugin code is sensible code.

plugins/hls-code-range-plugin/src/Ide/Plugin/CodeRange.hs
joyfulmantis: what about PluginInvalidUserState ? That at least clarifies in the name that the failure can only be about the user's state vs PluginHandlerFailed being more generic?

ghcide/src/Development/IDE/Plugin/HLS.hs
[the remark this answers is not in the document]
joyfulmantis: So I guess the error code doesn't influence whether or not it is shown by the client, so MethodNotFound is more correct.

ghcide/src/Development/IDE/Plugin/HLS.hs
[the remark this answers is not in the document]
joyfulmantis: I'm wrong about when errors are shown (I guess they never are?) So moving these to RequestFailed as mentioned on IRC

plugins/hls-code-range-plugin/test/Main.hs
joyfulmantis: Previously it was because that's how we treated all rule failures.. now we will respond with a RequestFailed response, so we need to change that here to consider RequestFailed as nothing.

joyfulmantis:
> > I do think we may be forced to accept the presence of exceptions in Action, but I'm not sure.
> 
> I think Actions use AsyncExceptions for cutting off stale request, which is likely unavoidable. I don't think we should/have to remove all exceptions from `Action`, but removing them from the plugin code is sensible code.

Do you know if the underscore functions and returning Nothing in the rules are functionally equivalent? At least if that's the case we can dump the underscore functions and convert rule users of them to MaybeT

fendor:
> Do you know if the underscore functions and returning Nothing in the rules are functionally equivalent?

Sorry, I do not

ghcide/test/exe/ExceptionTests.hs
```diff
@@ -0,0 +1,212 @@
...
+                    , Notifications.descriptor (cmapWithPrio LogNotifications recorder) "ghcide-core"]
+
+          testIde recorder (testingLite recorder logger plugins) $ do
+              doc <- createDoc "A.hs" "haskell" "module A where"
+              waitForProgressDone
+              (view L.result -> lens) <- request SMethod_TextDocumentCodeLens (CodeLensParams Nothing Nothing doc)
+              case lens of
+                Right (InL []) ->
+                  pure ()
+                _ -> liftIO $ assertFailure $ "We should have had an empty list" <> show lens]
```
michaelpj, line 108: I'm not sure I quite understand the logic here, worth a comment. I think the idea is that we want to have _something_ to check that we made it this far and didn't die?

ghcide/test/exe/ExceptionTests.hs
```diff
@@ -0,0 +1,212 @@
...
+
+   , testGroup "Testing PluginError order..."
+      [ testCase "InternalError over InvalidParams" $ do
```
michaelpj, line 111: I think you could abstract this a bit into something that just takes the two errors to throw, perhaps?

ghcide/test/exe/ExceptionTests.hs
```diff
@@ -0,0 +1,212 @@
...
+            waitForProgressDone
+            (view L.result -> lens) <- request SMethod_TextDocumentCodeLens (CodeLensParams Nothing Nothing doc)
+            case lens of
+              Left (ResponseError {_code = InL LSPErrorCodes_RequestFailed, _message}) ->
+                  liftIO $ assertBool "We caught an error, but it wasn't ours!"
+                          (T.isInfixOf "error test" _message && T.isInfixOf (coerce pluginId) _message)
+              _ -> liftIO $ assertFailure $ show lens
+     ]]
+
+testIde :: Recorder (WithPriority Log) -> IDE.Arguments -> Session () -> IO ()
```
michaelpj, line 179: Are we duplicating stuff from `func-tests` or something? I'm surprised the functions we need aren't already defined?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -961,13 +961,23 @@ useWithStale :: IdeRule k v
     => k -> NormalizedFilePath -> Action (Maybe (v, PositionMapping))
 useWithStale key file = runIdentity <$> usesWithStale key (Identity file)
 
--- | Request a Rule result, it not available return the last computed result which may be stale.
---   Errors out if none available.
+-- |Request a Rule result, it not available return the last computed result
+--  which may be stale.
+--
+-- Throws an `BadDependency` IO exception which is caught by the rule system if
```
michaelpj, line 967: It's just an exception right, not an `IOException`?

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -212,8 +218,8 @@ executeCommandHandlers recorder ecs = requestHandler SMethod_WorkspaceExecuteCom
           Nothing -> logAndReturnError recorder p (InR ErrorCodes_InvalidRequest) (commandDoesntExist com p xs)
           Just (PluginCommand _ _ f) -> case A.fromJSON arg of
             A.Error err -> logAndReturnError recorder p (InR ErrorCodes_InvalidParams) (failedToParseArgs com p err arg)
-            A.Success a ->
-                f ide a `catchAny` -- See Note [Exception handling in plugins]
+            A.Success a -> do
+              (first (toResponseError . (p,)) <$> runExceptT (f ide a)) `catchAny` -- See Note [Exception handling in plugins]
```
michaelpj, line 222: this is pretty hard to read. I'd probably try to make it a bit more "point-ful", maybe use the do block more.

hls-plugin-api/src/Ide/Plugin/Error.hs
michaelpj: > If it's not because of the user's invalid state, then arguably it's a bug, and should never happen

I feel like this is a pretty strong statement and I don't know that I would want to commit to it. I think I'd feel a bit more comfortable with a generic "error but not a bug" error, especially since I think being more prescriptive here isn't really buying us anything.

It also seems like "is this a bug in our code or not" is the dimension that we actually care about?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,141 @@
...
+      PluginInvalidUserState text -> "Invalid User State:" <+> pretty text
+      PluginRequestRefused msg    -> "Request Refused: "   <+> pretty msg
+
+toErrorCode :: PluginError -> (LSPErrorCodes |? ErrorCodes)
+toErrorCode (PluginInternalError _)    = InR ErrorCodes_InternalError
+toErrorCode (PluginInvalidParams _)    = InR ErrorCodes_InvalidParams
+toErrorCode (PluginInvalidUserState _) = InL LSPErrorCodes_RequestFailed
+-- PluginRequestRefused should never be a argument to `toResponseError`, as
+-- it should be dealt with in `extensiblePlugins`, but this is here to make
+-- this function complete
```
michaelpj, line 119: ... but also I think this is pretty logical

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -0,0 +1,141 @@
...
+toErrorCode (PluginRequestRefused _)   = InR ErrorCodes_MethodNotFound
+toErrorCode (PluginRuleFailed _)       = InL LSPErrorCodes_RequestFailed
+toErrorCode PluginStaleResolve         = InL LSPErrorCodes_ContentModified
+
+toPriority :: PluginError -> Priority
```
michaelpj, line 124: Haddock! good to explain when this should be used

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -0,0 +1,140 @@
...
+import qualified Ide.Logger                           as Logger
+import           Ide.Plugin.Error
+import qualified Language.LSP.Protocol.Types          as LSP
+
+-- ----------------------------------------------------------------------------
+-- Action wrappers
+-- ----------------------------------------------------------------------------
+
+-- |ExceptT version of `runAction`, takes a ExceptT Action
+runActionE :: MonadIO m => String -> IdeState -> ExceptT e Action a -> ExceptT e m a
```
michaelpj, line 30: We _could_ consider generalizing these functions to `MonadError` rather than `ExceptT`. Not sure that buys us much, I probably would instinctively do it, but I'm used to programming in a very `mtl`-y way!

michaelpj (approved):
This is looking pretty good now!

ghcide/test/exe/ExceptionTests.hs
joyfulmantis: Yes notification handlers don't return any response, so we can only test that the server is still alive

ghcide/test/exe/ExceptionTests.hs
joyfulmantis: This was shared with unit tests, so I moved the function from both there and unit tests and moved it to testutils

hls-plugin-api/src/Ide/Plugin/Error.hs
joyfulmantis: Theoretically, we could implement PluginError as a type class, and allow plugins to easily generate their own errors. However, I feel we are gaining here by restricting the possible set of errors. As we saw previously, everyone was either just throwing InternalError for everything, or were throwing errors that were incorrect (InvalidReqest for example). If there is a legitimate case in some plugin for a different error, then it can still easily be added here, but I feel it will keep the codebase saner and more coherent to restrict the errors here (And to keep them all in one place). 

However, I did change the wording on the PluginInternalError haddock to be more balanced. I feel if people would like to throw errors more than not, we can also downgrade PluginInternalError to Warning so as not to show a popup upon failure.

PR #3718, opened 2023-07-15
use versioned hackage libraries

smunix:
`ormolu_0_5_3_3` has been removed from nixpkgs. To prevent future breaks, let's pull specific versions from hackage when they are available.

### DIFF
--- a/configuration-ghc-92.nix
+++ b/configuration-ghc-92.nix
@@ -19,7 +19,7 @@ let
           doCheck = false;
         });
     } // (builtins.mapAttrs (_: drv: disableLibraryProfiling drv) {
-      apply-refact = hsuper.apply-refact_0_13_0_0;
+      apply-refact = hself.callHackage "apply-refact" "0.13.0.0" { };
 
       # ptr-poker breaks on MacOS without SSE2 optimizations
       # https://github.com/nikita-volkov/ptr-poker/issues/11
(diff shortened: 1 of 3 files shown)


fendor (approved):
LGTM!

fendor:
Are the changes supposed to fix nix CI?

smunix:
> Are the changes supposed to fix nix CI?

No, they just prevent breaks on our next update to nixpkgs. Ormolu version has been upgraded on Nix

cydparser:
There is a [comment in flake.nix](https://github.com/haskell/haskell-language-server/blob/c501f386b55be0f3ad0a8238dd9d135b7b26fe45/flake.nix#L141) stating not to use `callHackage`. @berberman, is that comment still relevant?

cydparser:
@smunix, I've fixed CI in https://github.com/haskell/haskell-language-server/pull/3724.

berberman:
> There is a [comment in flake.nix](https://github.com/haskell/haskell-language-server/blob/c501f386b55be0f3ad0a8238dd9d135b7b26fe45/flake.nix#L141) stating not to use `callHackage`. @berberman, is that comment still relevant?

`callHackage` works only if the corresponding version is in https://github.com/commercialhaskell/all-cabal-hashes, so previously we used `callCabal2nix` with `fetchurl` for very new packages. Given that now we have those package sources in flake inputs, I think we can delete this comment now.

For this PR, it should be fine to use `callHackage`.

smunix:
`haskell-language-server-96` requires `ormolu 0.7`, and `nix` can't build that version off of hackage. I'm closing this PR for now

PR #3725, opened 2023-07-21
Fix other file goto definition

nlander:
In my work on gotoDefinition for dependencies I found a bug that predates my code. In the first few seconds of startup for big projects, gotoDefinition uses the persistent rule for GetHieAst and gets its PositionMapping from that. This PositionMapping gets used on any location a definition is found, even if that definition is in a different file. This doesn't usually cause a problem, except when the definition in the other file is found on a line that is greater than the number of lines in the file we are coming from. In this case, the PositionMapping will end up nullifying the result returned by AtPoint.gotoDefinition.

If you want to see this bug in action, try opening ghcide/src/Development/IDE/LSP/Outline.hs and try to immediately do gotoDefinition on `extract_cons` on line 115.

This code adds some tests for gotoDefinition and fixes the problem by checking if the file where the definition is found is the same as the file we are calling gotoDefinition from. If not, we get the PositionMapping for the file we are going to and use that instead.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Actions.hs
+++ b/ghcide/src/Development/IDE/Core/Actions.hs
@@ -13,6 +13,7 @@ module Development.IDE.Core.Actions
 , lookupMod
 ) where
 
+import           Control.Monad.Extra                  (mapMaybeM)
 import           Control.Monad.Reader
 import           Control.Monad.Trans.Maybe
 import qualified Data.HashMap.Strict                  as HM
@@ -31,7 +32,9 @@ import           Development.IDE.Types.HscEnvEq       (hscEnv)
 import           Development.IDE.Types.Location
 import qualified HieDb
 import           Language.LSP.Protocol.Types          (DocumentHighlight (..),
-                                                       SymbolInformation (..))
+                                                       SymbolInformation (..),
+                                                       normalizedFilePathToUri,
+                                                       uriToNormalizedFilePath)
 
 
 -- | Eventually this will lookup/generate URIs for files in dependencies, but not in the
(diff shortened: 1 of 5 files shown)


fendor (approved):
Nice catch, LGTM!

michaelpj (approved):
Great investigation!

michaelpj:
I wonder if we have this problem anywhere else? I could imagine that e.g. `DocumentSymbols` and so on might produce locations in other files, which could have the same issues.

michaelpj:
Wait, it would have to be workspace symbols, or maybe call hierarchy.

michaelpj:
e.g. https://github.com/haskell/haskell-language-server/blob/f20562f9b9112e29c94eb9d2e5006cfe85e67679/ghcide/src/Development/IDE/Core/Actions.hs#L121

@wz1000 I'm guessing that these hiedb operations will return the positions in the file from some past point in time? Or will it do the right thing if the target file is open in the editor and has been altered by the user?

wz1000:
> e.g.
> 
> https://github.com/haskell/haskell-language-server/blob/f20562f9b9112e29c94eb9d2e5006cfe85e67679/ghcide/src/Development/IDE/Core/Actions.hs#L121
> 
> @wz1000 I'm guessing that these hiedb operations will return the positions in the file from some past point in time? Or will it do the right thing if the target file is open in the editor and has been altered by the user?

we index files asynchronously whenever they are changed, so if that is completed it will do the right thing (as long as the file compiles).

nlander:
Thanks all! Anything else that needs to happen before it gets the mergeme label?

michaelpj:
> so if that is completed it will do the right thing

And if it isn't it won't! It seems like in principle we probably ought to be computing position mappings for things from hiedb also, tbh.

PR #3729, opened 2023-07-25
Add resolve support in refine imports by merging it with explicit imports

joyfulmantis:


plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -110,28 +110,31 @@ runImportCommand recorder ideState eird@(ResolveOne _ _) = pluginResponse $ do
           logWith recorder Error (LogWAEResponseError re)
           pure ()
         logErrors (Right _) = pure ()
-runImportCommand _ _ (ResolveAll _) = do
-  pure $ Left $ ResponseError (InR ErrorCodes_InvalidParams) "Unexpected argument for command handler: ResolveAll" Nothing
+runImportCommand _ _ rd = do
```
michaelpj, line 113: It's weird that we don't have `RefineOne`, especially since we have the per-import lens, but I guess we didn't before either.

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -110,28 +110,31 @@ runImportCommand recorder ideState eird@(ResolveOne _ _) = pluginResponse $ do
...
+-- explicit. For example, for the module below:
 -- > import Data.List
--- >
 -- > f = intercalate " " . sortBy length
---
 -- the provider should produce one code lens associated to the import statement:
---
 -- > import Data.List (intercalate, sortBy)
+--
+-- | The second one allows us to import functions directly from the original
```
michaelpj, line 123:
```suggestion
-- The second one allows us to import functions directly from the original
```

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -110,28 +110,31 @@ runImportCommand recorder ideState eird@(ResolveOne _ _) = pluginResponse $ do
...
-    mbMinImports <- liftIO $ runAction "MinimalImports" state $ use MinimalImports nfp
+    mbMinImports <- liftIO $ runAction "ImportActions" state $ useWithStale ImportActions nfp
     case mbMinImports of
-      Just (MinimalImportsResult{forLens}) -> do
-        let lens = [ generateLens _uri range int
-                    | (range, int) <- forLens]
+      Just (ImportActionsResult{forLens}, pm) -> do
+        let lens = [ generateLens _uri newRange int
+                    | (range, int) <- forLens
+                    , Just newRange <- [toCurrentRange pm range]]
```
michaelpj, line 137: :+1:

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -141,42 +144,52 @@ lensProvider _  state _ CodeLensParams {_textDocument = TextDocumentIdentifier {
...
-        $ runAction "MinimalImports" ideState $ use MinimalImports nfp
-    let relevantCodeActions = filterByRange range forCodeActions
+        $ runAction "ImportActions" ideState $ useWithStale ImportActions nfp
+    newRange <- handleMaybe "Unable to get new range"
+        $ toCurrentRange pm range
+    let relevantCodeActions = filterByRange newRange forCodeActions
         allExplicit =
-          [InR $ mkCodeAction "Make all imports explicit" (Just $ A.toJSON $ ResolveAll _uri)
+          [InR $ mkCodeAction "Make all imports explicit" (Just $ A.toJSON $ ExplicitAll _uri)
+          | not $ null relevantCodeActions ]
```
michaelpj, line 184: This seems wrong: it should be `not $ null $ explicitImportCodActions` or something. We don't want to show this when there are only refine import actions!

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -141,42 +144,52 @@ lensProvider _  state _ CodeLensParams {_textDocument = TextDocumentIdentifier {
...
+        allRefine =
+          [InR $ mkCodeAction "Refine all imports" (Just $ A.toJSON $ RefineAll _uri)
           | not $ null relevantCodeActions ]
```
michaelpj, line 187: vice versa here

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -141,42 +144,52 @@ lensProvider _  state _ CodeLensParams {_textDocument = TextDocumentIdentifier {
...
-        toCodeAction uri (_, int) =
+        toCodeAction uri (ImportAction _ int ExplicitImport) =
           mkCodeAction "Make this import explicit" (Just $ A.toJSON $ ResolveOne uri int)
-    pure $ InL ((InR . toCodeAction _uri <$> relevantCodeActions) <> allExplicit)
+        toCodeAction uri (ImportAction _  int RefineImport) =
+          mkCodeAction "Refine this import" (Just $ A.toJSON $ ResolveOne uri int)
```
michaelpj, line 191: This seems wrong, it's also producing ResolveOne, shouldn't it be RefineOne or sometihng?

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -188,47 +201,58 @@ codeActionProvider _ ideState _pId (CodeActionParams _ _ TextDocumentIdentifier
...
+codeActionResolveProvider :: Recorder (WithPriority Log) -> ResolveFunction IdeState IAResolveData 'Method_CodeActionResolve
 codeActionResolveProvider _ ideState _ ca _ rd =
   pluginResponse $ do
     wedit <- resolveWTextEdit ideState rd
     pure $ ca & L.edit ?~ wedit
 --------------------------------------------------------------------------------
 
-resolveWTextEdit :: IdeState -> EIResolveData -> ExceptT String (LspT Config IO) WorkspaceEdit
+resolveWTextEdit :: IdeState -> IAResolveData -> ExceptT String (LspT Config IO) WorkspaceEdit
 resolveWTextEdit ideState (ResolveOne uri int) = do
```
michaelpj, line 212: TBH this `ResolveOne` name is weird. Shouldn't it be `ExplicitOne`?

michaelpj: Ugh, I figured it out. We are resolving individual edits _generically_, using the same machinery for both the explicit import edits and the refine import edits.

That makes me think that we could take this faster. What if we had
```
data ResolveData = ResolveData { uri :: Uri, importEdits :: [Int] }
```
Now we can have a totally generic resolve handler: get _all_ the referenced edits, and stick them together into a workspace edit, return. It's just up to the original code action provider to select the edits it will want eventually.

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
```diff
@@ -188,47 +201,58 @@ codeActionProvider _ ideState _pId (CodeActionParams _ _ TextDocumentIdentifier
...
-      WorkspaceEdit {_changes = Just $ Map.fromList [(uri, edits)]
+    $ runAction "ImportActions" ideState $ useWithStale ImportActions nfp
+  let edits = [ ie | ie@ImportEdit{ieResType = ExplicitImport} <- IM.elems forResolve]
+  pure $ mkWorkspaceEdit uri edits pm
+resolveWTextEdit ideState (RefineAll uri) = do
+  nfp <- getNormalizedFilePath uri
+  (ImportActionsResult{forResolve}, pm) <-
+    handleMaybeM "Unable to run Import Actions"
+    $ liftIO
+    $ runAction "ImportActions" ideState $ useWithStale ImportActions nfp
```
michaelpj, line 233: I forget, are we generally using `useWithStale` in the resolve handlers, or is it better to use `use`?

michaelpj (commented):
haven't looked at the tests yet but I assume they're the union of the old tests!

plugins/hls-explicit-imports-plugin/test/testdata/OnlyThis.expected.hs
```diff
@@ -1,7 +0,0 @@
-module OnlyThis where
```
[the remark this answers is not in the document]
joyfulmantis: Not removed, just renamed. Previously there were name clashes (because most of refine imports was just copied from explicit imports), so I prefixed the specific action before the testdata files.

plugins/hls-import-actions-plugin/src/Ide/Plugin/ImportActions.hs
joyfulmantis: We need to know the type when we offer code actions or resolve because we have slightly different behavior between the two (titles are different for example).

joyfulmantis:
> Looks like some actual failures in the test suite??

Yeah. There was a module name qualifier in a testdata dependent file I forgot to change. Funnily enough, GHC compile rules changed with 9.4+, because they had no problem with the test even though one of the dependent files shouldn't have been able to typecheck...

PR #3737, opened 2023-07-31
Various PluginError PR suggestions I missed earlier

joyfulmantis:


ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -80,10 +81,17 @@ prettyResponseError err = errorCode <> ":" <+> errorBody
         errorCode = pretty $ show $ err ^. L.code
         errorBody = pretty $ err ^. L.message
 
-pluginNotEnabled :: SMethod m -> [(PluginId, b, a)] -> Text
+pluginNotEnabled :: SMethod m -> [PluginId] -> Text
```
michaelpj, line 84: should this just be inlined into the function below now?

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -219,8 +220,15 @@ executeCommandHandlers recorder ecs = requestHandler SMethod_WorkspaceExecuteCom
           Just (PluginCommand _ _ f) -> case A.fromJSON arg of
             A.Error err -> logAndReturnError recorder p (InR ErrorCodes_InvalidParams) (failedToParseArgs com p err arg)
             A.Success a -> do
-              (first (toResponseError . (p,)) <$> runExceptT (f ide a)) `catchAny` -- See Note [Exception handling in plugins]
-                (\e -> logAndReturnError' recorder (InR ErrorCodes_InternalError) (ExceptionInPlugin p (Some SMethod_WorkspaceApplyEdit) e))
+              res <- runExceptT (f ide a) `catchAny` -- See Note [Exception handling in plugins]
+                (\e -> pure $ Left $ PluginInternalError (exceptionInPlugin p SMethod_WorkspaceExecuteCommand e))
```
michaelpj, line 224: Why this method? And it was different before??

michaelpj (approved):

ghcide/src/Development/IDE/Plugin/HLS.hs
joyfulmantis: Because that's it's actually method, see line 171. WorkspaceApplyEdit before was an error, either autocomplete typo, brain fog, or sloppy mistake on my part.

ghcide/src/Development/IDE/Plugin/HLS.hs
joyfulmantis: Yes, good catch!

PR #3743, opened 2023-08-04
Support resolve in type lenses

joyfulmantis:
I was hoping to merge all signature related stuff into one plugin, as it gives refactoring flexibility in changing rules and some functions. Unfortunately, the plugin that currently includes those code actions is hls-refactor, which is very creatively engineered. Hence, I had to give up on that effort.
I have so far come across two bugs in type-lenses HEAD, one where long signatures are formatted incorrectly, which is fixed here by putting the signature into a single line. The other, where local bindings mostly don't work is still unresolved. I hope to solve that and fix it here before merging, but currently still unsure what's causing it.


ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,81 +117,154 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
+-- and applies it.
 commandHandler :: CommandFunction IdeState WorkspaceEdit
 commandHandler _ideState wedit = do
   _ <- lift $ LSP.sendRequest SMethod_WorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing wedit) (\_ -> pure ())
   pure $ InR Null
 
 --------------------------------------------------------------------------------
-
-suggestSignature :: Bool -> Maybe HscEnv -> Maybe GlobalBindingTypeSigsResult -> Maybe TcModuleResult -> Maybe Bindings -> Diagnostic -> [(T.Text, [TextEdit])]
+-- To give one an idea about how creative hls-refactor plugin is, the end type
```
michaelpj, line 223: I don't quite understand this coment

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,81 +117,154 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
+-- here can be changed within certain parameters, and even though it is used by
+-- the hls-refactor-plugin, the hls-refactor-plugin itself won't need adaptions
+suggestSignature :: Bool -> Maybe HscEnv -> Maybe GlobalBindingTypeSigsResult -> Maybe TcModuleResult -> Maybe Bindings -> Diagnostic -> [(T.Text, TextEdit)]
 suggestSignature isQuickFix env mGblSigs mTmr mBindings diag =
-  suggestGlobalSignature isQuickFix mGblSigs diag <> suggestLocalSignature isQuickFix env mTmr mBindings diag
-
-suggestGlobalSignature :: Bool -> Maybe GlobalBindingTypeSigsResult -> Diagnostic -> [(T.Text, [TextEdit])]
+  maybeToList (suggestGlobalSignature isQuickFix mGblSigs diag) <> maybeToList (suggestLocalSignature isQuickFix env mTmr mBindings diag)
+
+-- Both the suggestGlobalSignature and suggestLocalSignature functions have been
```
michaelpj, line 230: This is a comment about a change - better to describe the current state

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,81 +117,154 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
 suggestGlobalSignature isQuickFix mGblSigs Diagnostic{_message, _range}
-  | _message
-      =~ ("(Top-level binding|Pattern synonym) with no type signature" :: T.Text)
-    , Just (GlobalBindingTypeSigsResult sigs) <- mGblSigs
-    , Just sig <- find (\x -> sameThing (gbSrcSpan x) _range) sigs
+  | _message =~ ("(Top-level binding|Pattern synonym) with no type signature" :: T.Text) =
+    suggestGlobalSignature' isQuickFix mGblSigs Nothing _range
+  | otherwise = Nothing
+
+-- In addition, for suggestGlobalSignature, we added the option of having a
```
michaelpj, line 241: ditto

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,81 +117,154 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
+-- PositionMapping. In this case if there is no PositionMapping provided, it will
+-- ignore it. However if a PositionMapping is supplied, it will assume that the
+-- range provided is already converted with the PositionMapping, and will attempt
+-- to convert it back before attempting to find the signature from the rule.
```
michaelpj, line 245: This feels a little funny - if the Range needs to be predictably converted in this way, why can't it be the caller's responsibility?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -212,9 +293,18 @@ gblBindingTypeSigToEdit GlobalBindingTypeSig{..} mmp
     -- If `mmp` is `Nothing`, return the original range, it used by lenses from diagnostic,
     -- otherwise we apply `toCurrentRange`, and the guard should fail if `toCurrentRange` failed.
     , Just range <- maybe (Just beforeLine) (flip toCurrentRange beforeLine) mmp
-    = Just $ TextEdit range $ T.pack gbRendered <> "\n"
+    -- We need to flatten the signature, as otherwise long signatures are
+    -- rendered on multiple lines with invalid formatting.
```
michaelpj, line 297: What specifically goes wrong? Should we instead fix the formatting? Do multi-line lenses even work?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -212,9 +293,18 @@ gblBindingTypeSigToEdit GlobalBindingTypeSig{..} mmp
...
+    , renderedFlat <- intercalate " " $ lines gbRendered
+    = Just $ TextEdit range $ T.pack renderedFlat <> "\n"
   | otherwise = Nothing
 
+-- |What we need to resolve our lenses, the type of binding it is, and if it's
+-- a local binding, it's identifier and range.
+data TypeLensesResolveData = TypeLensesResolveLocal {identifier :: T.Text, range :: Range}
```
michaelpj, line 304: How is this range different from the one in the lens?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,81 +117,154 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
+        -- bindings.
+        -- TODO: We need the identifier, but not sure we need the _range.
+        -- One I get it to reliably work I can find out.
+        generateLensFromLocalDiags diags =
+          [ CodeLens _range Nothing (Just $ toJSON $ TypeLensesResolveLocal identifier _range)
+             | (dFile, _, Diagnostic{_range, _message}) <- diags
+            , dFile == nfp
+            , Just (_ :: T.Text, _ :: T.Text, _ :: T.Text, [identifier]) <-
+                [(T.unwords . T.words $ _message)
+                  =~~ ("Polymorphic local binding with no type signature: (.*) ::" :: T.Text)]]
```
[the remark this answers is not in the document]
michaelpj: Generally it feels like maybe this function should pair up next to the function that fills in the details, which I guess is `suggestLocalSignature`? Not sure

michaelpj (approved):

michaelpj:
Not sure why the CI isn't running. Maybe fix the conflicts and push again.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
joyfulmantis: The lenses display was never multiline, just the textedit was. Not sure why, possibly because the lens display stuff already squishes it into a line. You should be able to see the problem if you remove any long signature from a function, and use type lenses to get it back. The lines following the first line aren't indented, which is illegal syntax.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,81 +117,154 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
+    let -- The first option is to generate the lens from diagnostics about local
+        -- bindings.
+        -- TODO: We need the identifier, but not sure we need the _range.
+        -- One I get it to reliably work I can find out.
+        generateLensFromLocalDiags diags =
+          [ CodeLens _range Nothing (Just $ toJSON $ TypeLensesResolveLocal identifier _range)
+             | (dFile, _, Diagnostic{_range, _message}) <- diags
+            , dFile == nfp
+            , Just (_ :: T.Text, _ :: T.Text, _ :: T.Text, [identifier]) <-
+                [(T.unwords . T.words $ _message)
```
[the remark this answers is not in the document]
joyfulmantis: Yeah, I copied that from the old code, but looking at it, it's obviously shouldn't be there.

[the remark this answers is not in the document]
joyfulmantis: No longer applicable as I removed all local signature code

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
fendor: Can confirm, that behaviour has bitten me many times before.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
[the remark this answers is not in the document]
joyfulmantis: No longer applicable as I removed all local signature code

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -109,97 +115,103 @@ codeLensProvider :: PluginMethodHandler IdeState Method_TextDocumentCodeLens
...
+isGlobalDiagnostic :: Diagnostic -> Bool
+isGlobalDiagnostic Diagnostic{_message} = _message =~ ("(Top-level binding|Pattern synonym) with no type signature" :: T.Text)
 
-suggestGlobalSignature :: Bool -> Maybe GlobalBindingTypeSigsResult -> Diagnostic -> [(T.Text, [TextEdit])]
-suggestGlobalSignature isQuickFix mGblSigs Diagnostic{_message, _range}
-  | _message
-      =~ ("(Top-level binding|Pattern synonym) with no type signature" :: T.Text)
-    , Just (GlobalBindingTypeSigsResult sigs) <- mGblSigs
-    , Just sig <- find (\x -> sameThing (gbSrcSpan x) _range) sigs
+-- If a PositionMapping is supplied, this function will call
```
michaelpj, line 204: I think I made a comment about this and it got lost - why can't it be the caller's responsibility to map the Range?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
michaelpj: Can we at least add a test for this?

michaelpj (approved):
Looking good, just a couple nits

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
joyfulmantis: I marked your previous comment as resolved, because I did move the range position mapping stuff to the caller. However the position mapping is still sent to the function that generates lenses from the global signature

Issue #3660, opened 2023-06-14
More plugin introspection information


michaelpj:
At the moment we know a little bit about plugins, but mostly their names. It would be nice if plugins could provide us with some extra information: a description, perhaps some optional version information about what key libraries it was built against (e.g. ormolu).

We could then at least log this information when we startup.

michaelpj:
This would help the user in e.g. https://github.com/haskell/haskell-language-server/issues/3618 to know what's going on.

michaelpj:
It would also be nice to know whether a plugin is enabled or not. We could imagine having something like:
```
#if stylish-haskell
   StylishHaskell.plugin
#else
   disabledPlugin "stylish-haskell" "Say something about why here?"
#endif
```

Then `--list-plugins` could also show that plugins are disabled.

michaelpj:
I think that this could let us push much more conditional logic into the plugin packages, which might be nice.

For example, we could have something like this in `hls-stylish-haskell-plugin.cabal`:

```
flag hls-stylish-haskell-enabled

library
  ...
  if flag(hls-stylish-haskell-enabled)
     build-depends: stylish-haskell
     cpp-options: -Dhls_stylishHaskell
     exposed-modules: StylishHaskell4Real
```
and then in `StylishHaskell.hs`:
```
#if hls_stylishHaskell
import StylishHaskell4Real as Real
#endif 

descriptor :: PluginDescriptor
descriptor = PluginDescriptor {
  description = "Plugin that provides formatting with stylish-haskell"
#if hls_stylishHaskell
  , pluginHandlers = Real.pluginHandlers
  , disabled = Nothing
#else
  , pluginHandlers = mempty
  , disabled = Disabled "Plugin is disabled"
#endif
}
```

Then in the main HLS executable we can unconditionally depend on the stylish-haskell plugin, and we'll either get the real plugin or a placeholder "disabled" plugin, depending.

michaelpj:
Brainstorming:
- Human-readable description (could have a lot more detail)
- Build information (versions of key libraries using `VERSION_pkgname`)
- List of handlers that it provides (we could even annotate handlers with more information about what they're for)
- CLI options it provides (do we even use this at the moment?)

Some of this we already know, but might need more annotation in order to let us show it to the user in a readable way.

July541:
Before making a decision here, should we add a log while formatters are activated to describe its version?

PR #3744, opened 2023-08-05
Log fourmolu and ormolu version that hls using

July541:
Since fourmolu exported its version from fourmolu-0.12, we can utilize this to log version info.

And this pr also adds the missing fourmolu-0.13 dependency...


plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -75,6 +81,11 @@ provider recorder plId ideState typ contents fp fo = ExceptT $ withIndefinitePro
             (pure . Left . PluginInternalError . T.pack . show)
              $ runExceptT $ cliHandler fileOpts
         else do
+#if MIN_VERSION_fourmolu(0,12,0)
+            logWith recorder Info $ LogBuiltinVersion $ Just Fourmolu.version
```
michaelpj, line 85: Is this better than `read #VERSION_fourmolu`? That should work all the time, and for all packages, right?

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
```diff
@@ -170,6 +186,8 @@ instance Pretty LogEvent where
         NoConfigPath ps -> "No " <> pretty configFileName <> " found in any of:"
             <> line <> indent 2 (vsep (map (pretty . show) ps))
         StdErr t -> "Fourmolu stderr:" <> line <> indent 2 (pretty t)
+        LogBuiltinVersion v -> "Using builtin fourmolu" <> pretty (maybe "" showVersion v)
```
michaelpj, line 189: I think "builtin" is a bit ambiguous. Maybe "compiled in"?

michaelpj (commented):
At the moment this will be fairly noisy: it's logged at Info and it's going to be logged every time there's a formatting request. It seems to me like it would be better to do it once when the plugin is set up, if we can? That might require doing a "version probe" of the CLI tool at startup, but that seems okay?

plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs
July541: Thanks! I never know this before!😅

July541:
> That might require doing a "version probe" of the CLI tool at startup, but that seems okay?

I think this is vary from #3660, I just want to add some tracing to keep which formatter is using, hence we can locate the problem like https://github.com/haskell/vscode-haskell/issues/920 easily.

How about downgrading this log to debug level?

michaelpj (approved):
I'm fine with this. I'd love to do it more systematically, but this is still an improvement!

July541:
Also log ormolu version.

fendor (changes requested):
Please update the test cases as well! 

In particular, since the plugins now use the logger infrastructure, change `mkPluginTestDescriptor'` to `mkPluginTestDescriptor` for both plugins.

fendor (approved):
LGTM

PR #3745, opened 2023-08-06
Sending `workspace/configuration` request while handling `Initilized`

July541:
We *can* rely on `workspace/configuration` to keep track of the configuration after we resolve the following problems.

- [ ] How to upgrade the config in the `Initialized` handler, seems we can't modify the `IdeState` in place.
- [ ] We have dozens of configurations, do these config field names all available for different clients like vscode and emacs?
- [ ] How to test this pr? Ideally I think we can add a test in vscode-haskell.

I'll try to investigate these questions, and also very welcome any suggestions or guidance in case I have to browse through lsp protocol and some of their implementations :)

### DIFF
--- a/ghcide/src/Development/IDE/LSP/Notifications.hs
+++ b/ghcide/src/Development/IDE/LSP/Notifications.hs
@@ -40,6 +40,7 @@ import           Development.IDE.Types.Location
 import           Development.IDE.Types.Shake           (toKey)
 import           Ide.Logger
 import           Ide.Types
+import           Language.LSP.Server                   (sendRequest)
 import           Numeric.Natural
 
 data Log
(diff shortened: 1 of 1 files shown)


michaelpj:
We should have handling for this in the part of `ServerDefinition` that deals with `didChangeConfiguration` today. So we should just be able to port that over to here. I do think that `lsp` should handle this generically, but if we can patch this in here that's a good short term fix!

ghcide/src/Development/IDE/LSP/Notifications.hs
```diff
@@ -127,6 +128,12 @@ descriptor recorder plId = (defaultPluginDescriptor plId) { pluginNotificationHa
         setSomethingModified (VFSModified vfs) ide [toKey GetClientSettings emptyFilePath] "config change"
 
   , mkPluginNotificationHandler LSP.SMethod_Initialized $ \ide _ _ _ -> do
+
+      let params = ConfigurationParams [ConfigurationItem Nothing (Just "haskell.formattingProvider")]
+      void $ sendRequest LSP.SMethod_WorkspaceConfiguration params $ \res -> do
+        liftIO $ logWarning (ideLogger ide) $ "zltest: " <> (Text.pack $ show res)
+        pure ()
+
```
wz1000, line 136: you probably want to block waiting for the response to SMethod_WorkspaceConfiguration here?

michaelpj: I think we don't need to. We have a default config, so we should always be able to operate and just update the config dynamically when we get the response, as we would if we got `didChangeConfiguration`.

michaelpj:
I've made a decent amount of progress on this. I have it fixed generically in `lsp`, I'm just making `lsp-test` handle it properly, and I think I've also got HLS working on top of that. I'm going away for the weekend but I think I should be able to finish it off early next week.

July541:
@michaelpj  I have some hours available this Sunday, so I'll continue your work if you don't have enough time.

PR #3751, opened 2023-08-10
Pedantic ghcide

joyfulmantis:
This is taking much more time than I had anticipated, so I'm going get this merged in, and then if I have time later, finish the rest then
Currently the following warnings are fixed

- [x] redundant-imports
- [x] name-shadowing
- [x]  unused-top-binds
- [x]  unused-local-binds
- [x]  orphans
- [x]  unused-matches
- [x] simplifiable-class-constraints

These warnings don't make sense to fix
- unrecognised-pragmas
- unused-packages 

These warnings are currently left to fix (some of these also might not make sense to fix)
-  dodgy-imports
-  missing-signatures
-  duplicate-exports
-  dodgy-exports
-  incomplete-patterns
-  ambiguous-fields
-  overlapping-patterns
- incomplete-record-updates

I have also drastically simplified the CPP in the imports section. Whereas before there were several layers of nested if-else CPP flags spread throughout the imports, now there are very clear single-layer if clauses at the end of the standard imports that are ordered by the ghc version they are for.


ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -79,8 +70,14 @@ import qualified Language.LSP.Protocol.Types                  as LSP
 import qualified Language.LSP.Server                          as LSP
 import           Language.LSP.VFS
 import           System.FilePath
+import           System.IO.Error
 import           System.IO.Unsafe
 
+#ifdef mingw32_HOST_OS
+import qualified System.Directory                             as Dir
```
michaelpj, line 77: looks dead?

ghcide/src/Development/IDE/Core/UseStale.hs
```diff
@@ -82,8 +82,8 @@ dualPositionMap (PositionMap (P.PositionMapping (P.PositionDelta from to))) =
 -- a 'PositionMapping' that will fast-forward values to the current age.
 data TrackedStale a where
   TrackedStale
-      :: Tracked (Stale s) a
-      -> PositionMap (Stale s) Current
+      :: Tracked ('Stale s) a
```
michaelpj, line 85: Can we just not include this warning? I think `-Wno-unticked-promoted-whatever` is now generally considered to be a bad idea.

ghcide/src/Development/IDE/LSP/HoverDefinition.hs
```diff
@@ -31,9 +31,9 @@ import qualified Language.LSP.Server            as LSP
 
 import qualified Data.Text                      as T
 
-gotoDefinition :: IdeState -> TextDocumentPositionParams -> ExceptT PluginError (LSP.LspM c) (MessageResult Method_TextDocumentDefinition)
+gotoDefinition :: IdeState -> TextDocumentPositionParams -> ExceptT PluginError (LSP.LspM c) (MessageResult 'Method_TextDocumentDefinition)
```
michaelpj, line 34: yeah, this stuff is just worse IMO

michaelpj (approved):
Great stuff, the CPP is _way_ better. It seems to me that you've come up with a policy for how to write the big blocks of CPP in imports - perhaps you could write it down in a Note somewhere and refer to it from the files where we do this a lot? That would help us keep to it in future.

ghcide/src/Development/IDE/Core/UseStale.hs
fendor: I agree, unticked is the way to go nowadays.

ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
```diff
@@ -220,17 +224,21 @@ type ErrMsg  = MsgEnvelope DecoratedSDoc
...
+mkPrintUnqualifiedDefault :: HscEnv -> GlobalRdrEnv -> PrintUnqualified
+mkPrintUnqualifiedDefault env =
   -- GHC 9.2 version
   -- mkPrintUnqualified :: UnitEnv -> GlobalRdrEnv -> PrintUnqualified
   mkPrintUnqualified (hsc_unit_env env)
 #else
+mkPrintUnqualifiedDefault :: HscEnv -> GlobalRdrEnv -> PrintUnqualified
+mkPrintUnqualifiedDefault env =
   HscTypes.mkPrintUnqualified (hsc_dflags env)
 #endif
```
fendor, line 243: This seems like there is no benefit at all here?

Last time I checked, the guidance on CPP in code is, minimal region possible. This change allows that the three type signatures are out of sync.

fendor (approved):
Looking pretty good, just one thing I don't think is an improvement

ghcide/src/Development/IDE/GHC/Compat/Outputable.hs
joyfulmantis: This change was to avoid a stylish-haskell error. However, duplicating the signature was unnecessary, and I have removed that from the CPP now.

ghcide/src/Development/IDE/Core/FileStore.hs
joyfulmantis: good catch!

ghcide/ghcide.cabal
```diff
@@ -221,14 +226,27 @@ library
...
+                     -Wwarn=dodgy-imports
+                     -Wwarn=missing-signatures
+                     -Wwarn=duplicate-exports
+                     -Wwarn=dodgy-exports
+                     -Wwarn=incomplete-patterns
+                     -Wwarn=overlapping-patterns
+                     -Wwarn=incomplete-record-updates
+    
+    if impl(ghc >= 9.2) && flag(pedantic)
+        ghc-options: -Wwarn=ambiguous-fields
```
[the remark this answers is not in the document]
joyfulmantis: Yes, ambiguous-fields is only understood by GHC >= 9.2, so we have to gate it here. Added a comment for clarification.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -111,6 +110,10 @@ import           HieDb.Utils
 import qualified System.Random                        as Random
 import           System.Random                        (RandomGen)
 
+#if !MIN_VERSION_ghc(9,4,0)
```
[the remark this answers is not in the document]
joyfulmantis: It was necessary to clean up redundant imports (it's not used GHC >= 9.4 (Or it's possibly rexported by someone else for those versions?!))

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -96,56 +94,66 @@ import           Development.IDE.Types.Location
...
+#if !MIN_VERSION_ghc(9,0,1)
+import           HscTypes
+import           TcSplice
+#endif
 
 #if MIN_VERSION_ghc(9,0,1)
 import           GHC.Tc.Gen.Splice
+#endif
+
+#if MIN_VERSION_ghc(9,0,1) && !MIN_VERSION_ghc(9,2,1)
```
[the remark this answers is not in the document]
joyfulmantis: I believe `else` CPP is not worth it in the imports section. As this statement uses an `&&` symbol, it would need to nested CPP, and then you're pretty much back to where we started.

PR #3769, opened 2023-08-21
Support for resolve for class-plugin lenses

joyfulmantis:
This PR only changes the lenses in the class plugin to rely on resolve. I'm unsure of the benefit of doing the same for the code actions. The resolve was implemented by moving the logic from the code lens method to a separate rule, and then relying on that for the code lens method, the code lens resolve method, and the command.


July541 (approved):
Not familiar with the resolve, but vote 👍 since you are working on this continuously.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeLens.hs
```diff
@@ -25,118 +25,73 @@ import           Language.LSP.Protocol.Message
...
-                    ]
-        targetSigs = matchBind bindInfos allBinds
-        makeLens (range, title) =
-            generateLens plId range title
-                $ workspaceEdit pragmaInsertion
-                $ makeEdit range title mp
-        codeLens = makeLens <$> mapMaybe getRangeWithSig targetSigs
-
-    pure $ InL codeLens
+    (InstanceBindLensResult (InstanceBindLens{lensRendered}), _)
```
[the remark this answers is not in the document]
joyfulmantis: All three methods: the code lens, the code lens resolve, and the command handler rely on the same rule. However, we extract different data from the rule. For the first code lens handler, we only return the range of each code lens and a unique ID that allows us to resolve it later. To resolve the lens we match the unique ID the client sends us to the text edit of the lens (this allows us to extract the title). Finally, for the command, we actually generate a correct textEdit with position mapping properly adjusted and use that.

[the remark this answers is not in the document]
July541: My point is about the action, we are calling the same `classplugin.GetInstanceBindLens` action for different functions, not sure if we can omit some calls, but overall this is good enough now.

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -81,30 +115,101 @@ data BindInfo = BindInfo
...
+    where
+        rdrEnv gblEnv= tcg_rdr_env gblEnv
+        showDoc hsc gblEnv ty = showSDocForUser' hsc (mkPrintUnqualifiedDefault hsc (rdrEnv gblEnv)) (pprSigmaType ty)
+        bindToSig hsc gblEnv id = do
+            let name = idName id
+            whenMaybe (isBindingName name) $ do
+                env <- tcInitTidyEnv
+                let (_, ty) = tidyOpenType env (idType id)
+                pure $ InstanceBindTypeSig name
+                        (prettyBindingNameString (printOutputable name) <> " :: " <> T.pack (showDoc hsc gblEnv ty))
```
michaelpj, line 214: This field of the rule result is strict, which means we're probably doing all this formatting work up-front, which I suspect we don't in fact want? Indeed, I think that's some of the primary stuff we want to delay until the resolve happens.

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -81,30 +115,101 @@ data BindInfo = BindInfo
...
-    define (cmapWithPrio LogShake recorder) $ \GetInstanceBindTypeSigs nfp -> do
-        tmr <- use TypeCheck nfp
-        hsc <- use GhcSession nfp
-        result <- liftIO $ instanceBindType (hscEnv <$> hsc) (tmrTypechecked <$> tmr)
-        pure ([], result)
+getInstanceBindLensRule :: Recorder (WithPriority Log) -> Rules ()
+getInstanceBindLensRule recorder = do
+    defineNoDiagnostics (cmapWithPrio LogShake recorder) $ \GetInstanceBindLens nfp -> runMaybeT $ do
+        tmr@(tmrRenamed ->  (hs_tyclds -> tycls, _, _, _)) <- useMT TypeCheck nfp
+        (InstanceBindTypeSigsResult allBinds) <- useMT GetInstanceBindTypeSigs nfp
```
michaelpj, line 122: This is the only place this rule is called... so is there any reason not to just inline the logic here???

michaelpj (commented):
Couple of high-level worries

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
joyfulmantis: It is used also for code actions, hence the need to create a seperate rule for lens specific stuff.

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -71,17 +72,32 @@ instance NFData InstanceBindTypeSigsResult where
...
 type instance RuleResult GetInstanceBindTypeSigs = InstanceBindTypeSigsResult
 
+-- |The necessary data to execute our code lens
 data InstanceBindLensCommand = InstanceBindLensCommand
-    { commandUri :: Uri
-    , commandUid :: Int}
+    { -- |The URI needed to run actions in the command
+      commandUri  :: Uri
+      -- |The specific TextEdit we want to apply. This does not include the
+      -- pragma edit which is computed in the rule
```
joyfulmantis, line 80: /s/rule/command

PR #3771, opened 2023-08-22
Add support for external Ormolu

sir4ur0n:
Related to https://github.com/haskell/haskell-language-server/issues/411

This PR adds support for a new LSP flag `haskell.plugin.ormolu.config.external`, boolean, default `false`. If set to `true`, it uses the `ormolu` binary from the path rather than the built-in one.
This is similar to `haskell.plugin.fourmolu.config.external`.

This PR takes inspiration from:
* https://github.com/haskell/haskell-language-server/pull/2763
* the current state of the file `plugins/hls-fourmolu-plugin/src/Ide/Plugin/Fourmolu.hs`

This PR also documents the flag `haskell.plugin.fourmolu.config.external`.

Example HLS debug logs when formatting:
```
2023-08-22T11:48:55.091644Z | Debug | ormolu: Using external ormolu-0.5.3.0
2023-08-22T11:48:55.091712Z | Debug | ormolu: Running: `ormolu -o-XMonomorphismRestriction -o-XDeepSubsumption -o-XRelaxedPolyRec -o-XForeignFunctionInterface -o-XImplicitPrelude -o-XOverloadedStrings -o-XDisambiguateRecordFields -o-XRecordWildCards -o-XNamedFieldPuns -o-XDoAndIfThenElse -o-XEmptyDataDecls -o-XPatternGuards -o-XDatatypeContexts -o-XTraditionalRecordSyntax -o-XLambdaCase -o-XStarIsType -o-XCUSKs -o-XFieldSelectors --stdin-input-file /home/sir4ur0n/code/haskell-language-server/plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs` in directory /home/sir4ur0n/code/haskell-language-server/plugins/hls-alternate-number-format-plugin/src/Ide/Plugin
```

I will try to open PRs to the various Haskell LSP Clients to also support those 2 flags (Ormolu and Fourmolu) once this PR is merged.

PR #3772, opened 2023-08-23
Rework "Configuration" and "Manually testing HLS" documentations

sir4ur0n:
While working on HLS in the last couple days for https://github.com/haskell/haskell-language-server/pull/3771 I found it difficult to understand how to:
* configure HLS
* manually test HLS when modifying the code

This PR tries to improve the current documentation.

In my opinion this makes the documentation clearer:
* explicit separation between "implicit vs generated vs manual" `hie.yaml` configuration
* remove parts of the "Using HLS on HLS code" section that were not specific to the HLS code at all (it could be applied to any code base) and reuse the "Configuration" documentation
* explicit separation of the 4 steps to use a custom HLS on any code base. Some steps depend on the build tool (Stack vs Cabal) while others depend on the editor, so I think this separation is more intuitive and exhaustive

Please feel free to make any additional suggestion


docs/configuration.md
```diff
@@ -90,83 +90,97 @@ Settings like this are typically be provided by the language-specific LSP client
...
-**For a full explanation of how `hie-bios` determines the project build configuration, and how to configure it manually, refer to the [hie-bios README](https://github.com/mpickering/hie-bios/blob/master/README.md).**
+There are several ways to provide this configuration to `hie-bios`, detailed below.
 
-At the moment, `haskell-language-server` has support to automatically detect your project build configuration to handle most use cases.
+### Implicit configuration
+If no `hie.yaml` file is present, `haskell-language-server` automatically detects your `hie-bios` configuration.
+**For most cases, this works just fine, and is the recommended way.**
 
-*So using a explicit `hie.yaml` file will not likely fix your ide setup*. It will do it almost only if you see an error like `Multi Cradle: No prefixes matched`
+See [the hie-bios documentation](https://github.com/haskell/hie-bios#implicit-configuration) for more information on implicit configuration.
```
michaelpj, line 102: I'm not sure this is quite accurate. I believe we in fact will use `implicit-hie` in this instance, not the `hie-bios` implicit auto-configuration...

docs/configuration.md
```diff
@@ -90,83 +90,97 @@ Settings like this are typically be provided by the language-specific LSP client
...
+In that case, you can automatically generate a `hie.yaml` file, using [implicit-hie](https://github.com/Avi-D-coder/implicit-hie):
+
+```shell
+gen-hie > hie.yaml  # In the root directory of your project
+```
+
+### Explicit, manual configuration
+Maybe using the generated `hie.yaml` file does not suit you.
+E.g., it still does not work, or you want to fine-tune the configuration.
+We recommend to start from the `hie.yaml` file generated by `implicit-hie` (see previous section) and modify it to suit your needs.
```
michaelpj, line 116: I think this recommendation is no longer quite right for cabal projects at least. For cabal projects you can often use the "empty" cabal cradle, which often even works better than the one generated by `implicit-hie`. However for stack I believe the `implicit-hie` one is better.

This situation is a mess, unfortunately.

docs/configuration.md
```diff
@@ -90,83 +90,97 @@ Settings like this are typically be provided by the language-specific LSP client
...
-A `hie.yaml` file **explicitly** describes how to setup the environment to compile the various parts of your project.
 For that you need to know what *components* your project has, and the path associated with each one.
 So you will need some knowledge about
 [stack](https://docs.haskellstack.org/en/stable/build_command/#components) or [cabal](https://cabal.readthedocs.io/en/latest/cabal-commands.html?#cabal-v2-build) components.
 
-You also can use [implicit-hie](https://github.com/Avi-D-coder/implicit-hie) to automatically generate `hie.yaml` files for
-the most common stack and cabal configurations
+**For a full explanation of how to configure it manually, refer to the [hie-bios documentation](https://github.com/mpickering/hie-bios/blob/master/README.md).**
+
+#### Examples of explicit `hie-yaml` configurations
```
michaelpj, line 124: we should include the "empty cabal cradle" which mostly works for everything now:

```
cradle:
  cabal:
```

(you might ask "why doesn't implicit-hie generate this?" which is a good question...)

docs/configuration.md
```diff
@@ -196,6 +210,12 @@ dependencies:
   - someDep
 ```
 
+### Common problems
```
michaelpj, line 213: I think this should go in the Troubleshooting page, perhaps?

docs/contributing/contributing.md
```diff
@@ -108,79 +108,73 @@ $ cabal run haskell-language-server:func-test -- -p "hlint enables"
...
-
-that configuration should help.
-
-3. Inspect & tune configuration explicitly.
-
-[Configuring project build](../configuration.md#configuring-your-project-build) applies to HLS project source code loading just as to any other.
-
-Note: HLS may implicitly detect codebase as a Stack project (see [hie-bios implicit configuration documentation](https://github.com/haskell/hie-bios/blob/master/README.md#implicit-configuration)). To use Cabal, try creating an `hie.yaml` file:
+Note: HLS implicitly detects the HLS codebase as a Stack project (since there is a `stack.yaml` file).
+If you want HLS to use Cabal, create this `hie.yaml` file at the root of the project:
```
michaelpj, line 114: :+1:

michaelpj (approved):
Thanks, this is great! sorry that the hie-bios configuration reality is a bit of a mess...

docs/configuration.md
sir4ur0n: Done in 739e58feeaaa80c1314686bd3f269986da2c5cdb

docs/configuration.md
sir4ur0n: I reworked and simplified a bit this section in 63ba56a7d4bbcb0c9cfcaa08ac2968d6636cd1c1

Basically rather than trying to re-explain the hie-bios documentation, we directly link to the relevant documentation

docs/configuration.md
sir4ur0n: Agreed, I added it in 63ba56a7d4bbcb0c9cfcaa08ac2968d6636cd1c1

docs/configuration.md
sir4ur0n: Good idea, done in 386f87cf367fe7d3ddae23fedd5c418f7b60d98b

docs/contributing/contributing.md
```diff
@@ -108,79 +108,73 @@ $ cabal run haskell-language-server:func-test -- -p "hlint enables"
...
+Note that unless you change the GHC version or the HLS version between builds, the path should remain the same, this is why you need to set it only once.
 
-  In this example output, the path would be `/home/user/.cabal/bin/haskell-language-server`.
+#### Using Cabal
+Run:
+```shell
+$ cabal build exe:haskell-language-server && cabal list-bin exe:haskell-language-server
+[..]
+<some long path>/haskell-language-server
+```
```
fendor, line 138: I know I am a little bit late to the party, but I feel like using the `cabal build` binary for testing can be slightly unexpected.
When you switch branches and run `cabal build` just to make sure it compiles, or to speed up HLS to load everything, then you suddenly update your test binary.
Such behaviour might be unexpected to non-expert users, perhaps even expert users. I, for one, definitely prefer `cabal install` since I have a couple of local HLS installations lying around.

That's why we previously recommended using `cabal install`, updating your tested HLS binary is a very conscious decision. Additionally, it allows you to set program-suffixes and other custom locations.

sir4ur0n: Arguably the opposite argument is as inconvenient for others 😅
1. `cabal install` will install in a global location outside of the project, so it makes it harder to keep things isolated
2. It might replace the stable binary they had installed in the past, and they might get confused about how to get back to the stable binary
3. With the `cabal build` approach, you can "undo" all the steps in the documentation and land back on your original setup. With the `cabal install` approach, "undoing" the steps may not be enough to get back to the original setup (because you also need to know how to get back the original binary, which you may have installed long ago, and you did not know which version you had).

I guess different people have different priorities 😅 And it's completely ok by the way!
In my opinion the new version of the documentation is more welcoming towards non-experts (ability to undo), and while some experts may get surprised by the scenario you mentioned, I think they would be able to identify and work around.

In any case it might also be beneficial to add this caveat to the documentation, if you suspect many will get confused by it

fendor: Fair enough, the arguments are also convincing. Ideally, we'd have a command `cabal build --copy-to-local-bindir`, that copies the binary to `dist-newstyle/bin/` or something similar. 

What do you think about instructions such as `cabal install --installdir dist-newstyle/bin/`? Arguably, we shouldn't touch `dist-newstyle`, but I think that's relatively safe.

michaelpj: FWIW I use `cabal build` and it's fine, but it seems reasonable to include both as alternatives?

PR #3773, opened 2023-08-23
Adapt to lsp changes for workspace/configuration

michaelpj:
DRAFT: depends on unreleased `lsp` code, I expect I still messed some stuff up with the tests

This has a few substantive changes and a lot of messing with tests.

- We now tell `lsp` our config section, and parse just that section.
- We move the logic for updating the shake build rules for client config from a `workspace/didChangeConfiguration` handler to the new `lsp` callback, which will ensure it gets called in all circumstances that can be relevant.

The test changes are more annoying:
- We ignore config and logging messages by default now, so we have to stop doing that when we care about it.
- Many tests didn't really need to _change_ the config, but rather just to set it once at the beginning. I adjusted a lot of test functions to allow passing in the initial config for this reason.


michaelpj:
Stack builds will fail since I haven't put the git dep in there. Shouldn't be a problem once we're using a released `lsp`.

July541:
It looks like you are just going to satisfy the lsp changes, LGTM.

fendor (approved):
LGTM

PR #3776, opened 2023-08-25
hls-cabal-fmt-plugin: Use the file contents of the LSP request

fendor:
Reported in https://github.com/haskell/vscode-haskell/issues/945

The fix is basic: don't read from disk.


michaelpj (approved):
oof yes

fendor:
@michaelpj @joyfulmantis We have to disable a feature: https://github.com/haskell/haskell-language-server/pull/3776/files#diff-4d5e5fc912b0611ade80726df74fd36b044251414b17842d367fc3935ddd68e3R46

See https://github.com/phadej/cabal-fmt/pull/82 for details

PR #3779, opened 2023-08-28
Add cradle dependencies to session loading errors

VeryMilkyJoe:
Extracted part of #3778, should be merged first.

Improve error message and information in cradle error related to unknown module.

When a cradle error is found to be related to an unknown module, we edit the error by adding information (the module's path, the most likely cabal file to add the module to) such that the cabal plugin can offer code actions for adding the module to a cabal file.


fendor:
Related to https://github.com/haskell/cabal/issues/9171 which provides the initial terrible error message.

michaelpj:
We're packing extra information into the diagnostic, but we're using the diagnostic text for that. There are at least two other places we could put that: `data`, or `relatedInformation`. `relatedInformation` actually seems useful: even without the code action, indicating the location of the relevant cabal file is directly useful to the user.

We'll still have to parse it out later, but at least it'll be slightly more structured.

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
```diff
@@ -0,0 +1,86 @@
...
+      multiCradleErrMessage <$> parseMultiCradleErr ms
+
+-- | Information included in Multi Cradle error messages
+data MultiCradleErr = MultiCradleErr
+  { mcPwd      :: FilePath
+  , mcFilePath :: FilePath
+  , mcPrefixes :: [(FilePath, String)]
+  } deriving (Show)
+
+-- | Attempt to parse a multi-cradle message
```
michaelpj, line 44: I have to say, it's pretty non-obvious to me what this is doing. I think it would be worth adding a test, it almost always is for these string-parsing things IMO, otherwise it's very easy for them to stop working accidentally.

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
```diff
@@ -0,0 +1,86 @@
...
+
+{- | Takes a cradle error, the corresponding cradle and the file path where
+  the cradle error occurred (of the file we attempted to load).
+  Depicts the cradle error in a user-friendly way.
+-}
+renderCradleError :: CradleError -> Cradle a -> NormalizedFilePath -> FileDiagnostic
+renderCradleError (CradleError deps _ec ms) cradle nfp
+  | HieBios.isCabalCradle cradle && any (isInfixOf "Error: cabal: Failed extracting script block:") ms =
+      let (fp, showDiag, diag) = ideErrorWithSource (Just "cradle") (Just DiagnosticSeverity_Error) nfp $ T.unlines $ map T.pack userFriendlyMessage in
+        (fp, showDiag, diag{_data_ = Just (Aeson.Array $ Vector.fromList $ map (Aeson.String . T.pack) absDeps)})
```
michaelpj, line 24: Ah, you are putting the dependencies in `Data`. That seems okay

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
```diff
@@ -0,0 +1,86 @@
...
+
+-- | Information included in Multi Cradle error messages
+data MultiCradleErr = MultiCradleErr
+  { mcPwd      :: FilePath
+  , mcFilePath :: FilePath
+  , mcPrefixes :: [(FilePath, String)]
+  } deriving (Show)
+
+-- | Attempt to parse a multi-cradle message
+parseMultiCradleErr :: [String] -> Maybe MultiCradleErr
```
michaelpj, line 45: Remind me again why we aren't just doing structured errors from `hie-bios`? :sweat_smile:

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
michaelpj, line 24: I do think it's worth defining a datatype that you use here with derived to/fromJSON instances, even if it's ultimately just a list of strings. Just makes it harder to mess up.

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
fendor: No reason :) Ill put it on my long term list of things to do after the release is finished.

test/functional/FunctionalBadProject.hs
```diff
@@ -2,37 +2,28 @@
...
 
--- ---------------------------------------------------------------------
--- TODO: Currently this can not succeed, since such an error is thrown in "runActionWithContext" which
--- can produce diagnostics at the moment. Needs more investigation
--- TODO: @fendor: Add issue link here
---
 tests :: TestTree
-tests = testGroup "behaviour on malformed projects" [
-    testCase "no test executed" $ True @?= True
+tests = testGroup "behaviour on malformed projects"
```
michaelpj, line 13: This is much more sophisticated than what I was thinking (although I imagine we'll want it anyway). I was just thinking of a unit test for the string matching function!

VeryMilkyJoe: Does that mean we don't want it?

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
```diff
@@ -0,0 +1,86 @@
...
+    [ "Loading the module '" <> moduleFileName <> "' failed. It may not be listed in your .cabal file!"
+    , "Perhaps you need to add `"<> moduleName <> "` to other-modules or exposed-modules."
+    , "For more information, visit: https://cabal.readthedocs.io/en/3.4/developing-packages.html#modules-included-in-the-package"
+    , ""
+    ] <> map prefix (mcPrefixes e)
+  where
+    localFilePath f = dropWhile (==pathSeparator) $ dropPrefix (mcPwd e) f
+    moduleFileName = localFilePath $ mcFilePath e
+    moduleName = intercalate "." $ map dropExtension $ dropWhile isSourceFolder $ splitDirectories moduleFileName
+    isSourceFolder p = all isLower $ take 1 p
```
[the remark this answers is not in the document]
VeryMilkyJoe: Yes, this seems weird. Let's remove it.

test/functional/FunctionalBadProject.hs
michaelpj: IDK, potentially good to have both?

fendor: I think the tests are there and we shouldn't have them forever, any way. Ideally cabal gives us better error messages and hie-bios doesn't demand the parsing. Should be good enough, imo.

michaelpj (approved):

fendor (approved):

PR #3775, opened 2023-08-25
Prepare release 2.2.0.0

fendor:
## Release 2.2.0.0

* Binaries for GHC 9.4.7
* Forward compatibility with latest VSCode client changes

### Pull Requests

- Adapt to lsp changes for workspace/configuration
  ([#3773](https://github.com/haskell/haskell-language-server/pull/3773)) by @michaelpj
- Rework "Configuration" and "Manually testing HLS" documentations
  ([#3772](https://github.com/haskell/haskell-language-server/pull/3772)) by @sir4ur0n
- Fix `main-is` completion suggestions not being relative to `hs-source-dirs`
  ([#3766](https://github.com/haskell/haskell-language-server/pull/3766)) by @VeryMilkyJoe
- Remove suggestion of stanzas inside of stanza context
  ([#3761](https://github.com/haskell/haskell-language-server/pull/3761)) by @VeryMilkyJoe
- Pedantic ghcide
  ([#3751](https://github.com/haskell/haskell-language-server/pull/3751)) by @joyfulmantis
- Fix #3574 and support resolve in explicit records
  ([#3750](https://github.com/haskell/haskell-language-server/pull/3750)) by @joyfulmantis

## Release checklist

- [ ] check ghcup supports new GHC releases if any
- [ ] set the supported GHCs in workflow file `.github/workflows/release.yaml`
  - There is currently a list of GHC versions for each major platform. Search for `ghc: [` to find all lists.
  - Look for `TODO:` to find locations that require extra care for GHC versions.
- [ ] check all plugins still work if release includes code changes
- [ ] bump package versions in all `*.cabal` files (same version as hls)
  - HLS uses lockstep versioning. The core packages and all plugins use the same version number, and only support exactly this version.
    - Exceptions:
      - `hie-compat` requires no automatic version bump.
      - `shake-bench` is an internal testing tool, not exposed to the outside world. Thus, no version bump required for releases.
  - For updating cabal files, the following script can be used:
    - ```sh
      # Update all `version:` fields
      sed -ri "s/^version:( +)2.2.0.0/version:\12.3.0.0/" **/*.cabal
      # Update all constraints expected to be in the form `== <version>`.
      # We usually don't force an exact version, so this is relatively unambiguous.
      # We could introduce some more ad-hoc parsing, if there is still ambiguity.
      sed -ri "s/== 2.1.0.0/== 2.2.0.0/" **/*.cabal
      ```
    - It still requires manual verification and review
- [ ] generate and update changelog
[...]

PR #3781, opened 2023-08-29
Enhance/releasing checklist

fendor:
After #3775 is merged


docs/support/ghc-version-support.md
```diff
@@ -19,6 +19,7 @@ Support status (see the support policy below for more details):
 |--------------|--------------------------------------------------------------------------------------|-----------------------------------------------------------------------------|
 | 9.6.2        | [latest](https://github.com/haskell/haskell-language-server/releases/latest)         | basic support                                                               |
 | 9.6.1        | [2.0.0.0](https://github.com/haskell/haskell-language-server/releases/tag/2.0.0.0)   | basic support                                                               |
+| 9.4.7        | [latest](https://github.com/haskell/haskell-language-server/releases/latest)         | full support                                                                |
 | 9.4.6        | [latest](https://github.com/haskell/haskell-language-server/releases/latest)         | full support                                                                |
 | 9.4.5        | [latest](https://github.com/haskell/haskell-language-server/releases/latest)         | full support                                                                |
```
michaelpj, line 24: I think the entries for 9.4.5 and 9.4.6 are wrong also? They should say the last supported version was something previous.

michaelpj (approved):

docs/support/ghc-version-support.md
fendor: We ship binaries for `9.4.6` and `9.4.5` with the `2.2.0.0` release as well, so the table accurate afaict

michaelpj: Hmm, I missed that. I thought the policy was to only ship binaries for the last minor release of each major release?

fendor: Uhm, yeah, that's true, too.

Essentially, I thought it would be better to ship binaries for these two older minor releases as well, so people don't have to upgrade immediately, don't invalidate their entire cabal store for a minor release, but still get an updated HLS version. There was no additional overhead to shipping the other binaries, iirc.

michaelpj: No worries, you know me, I just like there to be a policy so I know what we're doing :D

PR #3784, opened 2023-08-31
Do not suggest bogus module names

Bodigrim:
In one of my projects HLS sometimes suggests bogus module names such as `src.Foo.Bar` or `test.Foo.Quux`. It's not quite clear what's exactly going on: it's a complex project with a custom `shake`-based build system and intricate environment, so I cannot distill a reproducer to share. It might very well be a deeper issue (with dysfunctional GHC session?), but what I propose here is a simple sanity check: under no circumstances we should suggest malformed module names with components starting from lower-case characters.


fendor (approved):
LGTM.

We now have a number of locations where we check whether a string could be a valid module name. We likely should unify them.

PR #3787, opened 2023-09-03
Clean up Release CI

fendor:
Drops GHC 8.10.7 from the release CI.

Unifies the download directory for release binaries and the upload directory for uploading the binaries to downloads.haskell.org.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,9 +21,6 @@ build_task:
     GITHUB_WORKSPACE: ${CIRRUS_WORKING_DIR}
     CABAL_CACHE_NONFATAL: "yes"
   matrix:
-    - name: build-ghc-8.10.7
-      env:
-        GHC_VERSION: 8.10.7
     - name: build-ghc-9.0.2
       env:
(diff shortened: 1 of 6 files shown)


.cirrus.yml
```diff
@@ -46,7 +43,6 @@ build_task:
 bindist_task:
   name: bindist
   depends_on:
-    - build-ghc-8.10.7
     - build-ghc-9.0.2
     - build-ghc-9.2.5
     - build-ghc-9.2.7
```
michaelpj, line 48: suspicious? should just be 9.2.8?

michaelpj (approved):

.cirrus.yml
fendor: In theory, yes, but since we don't have any experience with freebsd, I, personally, don't know which ghc versions have a custom build available.

Additionally, Cirrus CI seems to discontinue their free freebsd CI, which will likely cause us to drop it in the next release. https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/

PR #3804, opened 2023-09-12
Reduce Nix support

### LINKED ISSUE #3800: Reduce the Nix setup in this repository
The current Nix setup is approximately always broken. This isn't surprising, since we tend to rely on lots of packages that have just been released, which is hard for the current Nix setup to support.

The only thing that I think we can offer if we're not able to build all the dependencies is a simple shell with a compiler and dev tools. And I think that's fine - personally I already only use the "simple" dev shell, because it's the only one that reliably works.

So I propose cutting it down to just the simple dev shells.

cc some people who might care @guibou @cydparser @Gabriella439

michaelpj:
This removes the support for:
- Dev shells with pre-built packages
- Building binaries

These have been almost continually broken, and nobody is really maintaining them. Better to just do the simple thing we can do reliably, which is to provide dev shells.

Closes #3800


.github/workflows/nix.yml
```diff
@@ -35,17 +34,14 @@ jobs:
...
-          cancel_others: false
-          paths: '[ "**.nix" ]'
 
   # Enter the development shell and run `cabal build`
   develop:
     if: needs.pre_job.outputs.should_skip_develop != 'true'
     needs: pre_job
     runs-on: ${{ matrix.os }}
+    env:
+      HAS_TOKEN: ${{ secrets.HLS_CACHIX_AUTH_TOKEN != '' }}
```
cydparser, line 44: `HAS_TOKEN` appears unused.

cydparser (approved):

fendor (approved):
Awesome simplification!

Issue #3770, opened 2023-08-21
Refine imports should not refine qualified imports


Ptival:
I appreciate the "refine imports" feature, but I'd love being able to disable it for qualified imports.

In general, it'd be nice to be able to control it (disable it entirely, disable it for qualified imports).

EDIT: see lower down for the diagnosis of the bug

michaelpj:
Just to be completely clear about the problem, the issue is that you have an import like this:
```
import Foo qualified as F
```
and it suggests something like
```
import Foo qualified as F (things, actually, used)
```
?

If so, I think that's just plain questionable and we should not do it, who ever wants that? Generally we try and avoid configuration knobs where we can, so if we can just remove the problematic behaviour that's best.

Ptival:
That's exactly the problem, yes.
In practice I used qualified precisely because the import list would be gigantic, so indeed the suggestion is completely useless to me.

In general, for these questions that are at the frontier between the VSCode plug-in and HLS, I'm never too sure who's responsible for the problem, so sorry if this should rather belong on the Haskell VSCode extension issue tracker.

Also, if anyone has a vague idea of where this problem may be coming from, I feel like I may have enough knowledge to be able to fix it myself (at least propose a patch). I'm just not too familiar with the code base. I'm going to assume that it's likely [in this file](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs), so I'll have a quick look to see if I can spot where to fix this.

Ptival:
My impression was that at some point the "refine imports" plugin may have been renamed "make imports explicit" plugin, since I don't find any trace of the former in the repo. I could be wrong.

Should be pretty easy to make a test, I may get to it only this weekend though.

Ptival:
Ah, I looked into this a little more, and I was missing an important part.

Those suggestions only happen when the plugin is also trying to refine the module being imported. So for instance if I have:

```
`- Super.hs
`- Super
  `- SubA.hs
  `- SubB.hs
```

where `Super.hs` is just a "re-exporting" module for things in `Super.SubA` and `Super.SubB`, and I do:

```
import Super qualified as Super

foo = Super.somethingFromSubA
```

then it will suggest to refine the import into `import Super.SubA qualified as Super (somethingFromSubA)`.

So this brings up two questions:
1. Should the suggestion either be replaced, or accompanied, with `import Super.SubA qualified as Super`?
[...]

PR #3806, opened 2023-09-14
Change refine imports behaviour for qualified imports

joyfulmantis:
This PR addresses issue #3770, specifically, it implements [option 2](https://github.com/haskell/haskell-language-server/issues/3770#issuecomment-1716058934) "If an import is qualified, refine it and keep the same qualification, but don't give it an explicit import list."


plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -476,16 +476,20 @@ filterByImport _ _ = Nothing
 
 constructImport :: LImportDecl GhcRn -> (ModuleName, [AvailInfo]) -> LImportDecl GhcRn
 #if MIN_VERSION_ghc(9,5,0)
-constructImport (L lim imd@ImportDecl {ideclName = L _ _, ideclImportList = Just (hiding, L _ names)})
+constructImport (L lim imd@ImportDecl {ideclName = L _ _, ideclQualified = qualified, ideclImportList = Just (hiding, L _ names)})
```
michaelpj, line 479: I'm confused - it looks like this case only triggers in the case that the original import *had* an import list! Otherwise `ideclImportList` would be `Nothing` and so we would just go into the fallthrough.

The original bug report suggested that we were converting imports without import lists to imports with import lists, but I can't see how that would happen from this logic. Does the test definitely fail without these changes?

michaelpj: Because as it stands, it looks to me like it will take something like this:
```
import RefineA as A (x, y)
```
and turn it into
```
import RefineB as A (x)
import RefineC as A (y)
```
which actually seems fine.

michaelpj: (It would also be nice to have a tests for the case of qualified-but-with-import-list also, even if it's weird)

michaelpj (commented):
I'm confused about whether the bug even happens!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: It's because on line 343 we actually calculate the refined imports from the generated minimal imports list, not the actual imports list. This is all legacy stuff, but I could see why they would want to do that if they assumed people would always want an explicit import list for refined imports (It certainly helps to know why we are importing some module)

joyfulmantis: > Because as it stands, it looks to me like it will take something like this:
> 
> ```
> import RefineA as A (x, y)
> ```
> 
> and turn it into
> 
> ```
> import RefineB as A (x)
> import RefineC as A (y)
> ```
> 
> which actually seems fine.

That also means that in this case we will end up dumping the explicit import list.

michaelpj: From what you're saying, it sounds like the minimal imports code action will also put import lists on qualified imports? That _also_ seems questionable to me - in that situation you can see where something is coming from by looking at the qualifier of the usage, e.g.

```
import Foo qualified as F

F.whatever
```
turning it into 
```
import Foo qualified as F (whatever)

F.whatever
```
Seems worse IMO.

I guess an alternative question is: what if we processed the non-minimal imports in refine imports? Then it seems like we would handle the second case I suggested, since it would be looking at the original imports? It seems reasonable to me for these two things to be orthogonal. WDYT?

joyfulmantis: For minimal imports, we specifically filter out qualified imports (line 324 in this PR). I moved some of the logic around, so now refine imports can work on the original import. For some reason, the edge case is still causing problems (There is a test case for it now, and you should be able to see it failing). Not sure why though.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -494,5 +497,7 @@ constructImport (L lim imd@ImportDecl{ideclName = L _ _, ideclHiding = Just (hid
           containsAvail name avail =
             any (\an -> printOutputable an == (printOutputable . ieName . unLoc $ name))
               $ availNamesWithSelectors avail
-
-constructImport lim _ = lim
+          noExplicitList mList = case mList of
+                                      Nothing      -> True
+                                      Just (_, xs) -> not $ null xs
```
michaelpj, line 502: This is a super weird case. This is handling `import Foo qualified ()`, right? I guess we might as well handle it, but I think it would be also reasonable to treat that as an explicitly specified import list, even if it's empty.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -474,18 +473,22 @@ filterByImport (L _ ImportDecl{ideclHiding = Just (_, L _ names)})
...
 #else
-constructImport (L lim imd@ImportDecl{ideclName = L _ _, ideclHiding = Just (hiding, L _ names)})
+constructImport ImportDecl{ideclQualified = qualified, ideclHiding = origHiding} imd@ImportDecl{ideclHiding = Just (hiding, L _ names)}
 #endif
-  (newModuleName, avails) = L lim imd
+  (newModuleName, avails) = imd
     { ideclName = noLocA newModuleName
 #if MIN_VERSION_ghc(9,5,0)
-    , ideclImportList = Just (hiding, noLocA newNames)
+    , ideclImportList = if noExplicitList origHiding && qualified /= NotQualified
```
michaelpj, line 485: TBH, at this point why not just put an import list in exactly if the original import had one? Then we avoid changing the status as much as possible. i.e. this condition could just be `isNothing origHiding`?

michaelpj (approved):
I like the direction of  travel, just one more potential simplification.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
joyfulmantis: Yeah, I was overthinking things. We only need isNothing and that fixes the test case too.

PR #3813, opened 2023-09-19
Revert "Bump actions/checkout from 3 to 4"

wz1000:
This caused release CI to fail, see https://github.com/haskell/haskell-language-server/actions/runs/6230276461

actions/checkout@v4 uses node 20 which requires GLIBC that is too new for our distributions


fendor (approved):
LGTM, although I think only `release.yaml` strictly *requires* the downgrade

PR #3830, opened 2023-10-06
Set the root for tests to the test directory

### LINKED ISSUE #3074: Fails to load with local tarballs in `cabal.project`
### Your environment

Which OS do you use?
Debian 11.4

GHC 9.0.2 from GHCup 0.1.18.0

With Cabal

VSCode 1.69.2 with extension haskell.haskell 2.2.0

Version 1.7.0.0 from GHCup

Simple Cabal cradle in `hie.yaml`, but the problem happens even without it

### Steps to reproduce

Put a local tarball as a package in the `cabal.project` file. For example:

```
-- cabal.project
packages:
  .
  vendor/flow-2.0.0.0.tar.gz
```

### Expected behaviour

The project should load, including the vendored tarball as a dependency. 

### Actual behaviour

The project fails to load.

### Debug information

From the command line:

```
haskell-language-server-wrapper: .../vendor/vendor/flow-2.0.0.0.tar.gz: getDirectoryContents:openDirStream: inappropriate type (Not a directory)
```

From the extension:

```
[Error - 12:36:46 PM] setInitialDynFlags: .../vendor/flow-2.0.0.0.tar.gz: getDirectoryContents:openDirStream: inappropriate type (Not a directory)
```

Link to documentation about tarballs in `cabal.project`: https://cabal.readthedocs.io/en/3.6/cabal-project.html#cfg-field-packages

fendor:
On startup, HLS performs a sanity check to validate the GHC version for the project is the same as the GHC version used to build HLS. Due to some oversights, the tests always use HLS's cabal package to perform this check.
This is (almost) valid for the tests but adds overhead for finding the GHC version for each integration test in HLS.
Improves the startup time for each integration test that depends on hls-test-utils. Thus, ghcide test suite is unaffected.

Fix `findCradle` invocations to point to a file in the root directory. `findCradle` from hie-bios doesn't work correctly for directories and requires a file, even if that file doesn't exist.

Incidentally, I think this fixes #3074, too.

We should fix `hie-bios`'s `findCradle` to also work for directories. If this PR works as hoped, I'll update `hie-bios`.


ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -266,7 +266,7 @@ loadWithImplicitCradle mHieYaml rootDir = do
 
 getInitialGhcLibDirDefault :: Recorder (WithPriority Log) -> FilePath -> IO (Maybe LibDir)
 getInitialGhcLibDirDefault recorder rootDir = do
-  hieYaml <- findCradle def rootDir
+  hieYaml <- findCradle def (rootDir </> "a")
```
michaelpj, line 269: these changes are debugging code I assume?

michaelpj (approved):
I'm amazed anything worked without this. Don't a bunch of the tests have their own `hie.yaml`s? So doesn't this mean we were ignoring them all because we were using the top-level cradle?

ghcide/session-loader/Development/IDE/Session.hs
fendor: not quite, we do have these kind of magic constants all over the place:
* https://github.com/haskell/haskell-language-server/blob/master/exe/Wrapper.hs#L224
* https://github.com/haskell/haskell-language-server/blob/master/src/Ide/Main.hs#L108

The reason is `findCradle` only works for Filepaths, not directories. For finding the cradle type, we don't need a real Filepath, which is why we have these constants.

However, this seems to be a missing feature in `hie-bios`'s API, not a feature.
So, if it speeds up CI, then I would merge this, update hie-bios's API, and then undo these magic filepaths.
But I am unfortunately not sure, they truly speed up CI.

fendor:
> So doesn't this mean we were ignoring them all because we were using the top-level cradle?

No, I think, only to make sure HLS is compiled against the correct GHC version.

In CI, this was always true, and for running the tests, the proper `hie.yaml` files were used.

ghcide/session-loader/Development/IDE/Session.hs
michaelpj: Oh I see. Well then they need comments!

michaelpj:
> For running the tests, the proper hie.yaml files are used.

But when we start the server in the test utils we were setting the wrong root? I don't see how that wouldn't end up running in the wrong place entirely?

fendor:
I am not entirely, as in "I haven't checked", but a Cradle produces the cradle root directory and HLS should be able to handle this easily. The only reason why it might not work correctly is that it might infer the wrong GHC version for a project, e.g., when it is a stack project, ...
This is a problem in practice, but not in CI as we make sure that all cradles (not sure if we still test stack cradles) use the same GHC version.

fendor:
Hie-bios PR https://github.com/haskell/hie-bios/pull/415

PR #3836, opened 2023-10-10
Simplify cabal.project

michaelpj:
- entropy has had bounds revised in
- basement/hw-prim bounds empirically unnecessariy
- hyphenation was for wingman
- bound hlint to a version that uses ghc-lib by default
- stylish-haskell has never had a ghc-lib flag
- none of the versions of fourmolu that we use have a fixity-th flag
- remove some other things that lack justification speculatively
- make shake-bench buildable on later GHCs

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -34,47 +34,26 @@ packages:
          ./plugins/hls-refactor-plugin
          ./plugins/hls-overloaded-record-dot-plugin
 
--- Standard location for temporary packages needed for particular environments
--- For example it is used in the project gitlab mirror to help in the MAcOS M1 build script
--- See https://github.com/haskell/haskell-language-server/blob/master/.gitlab-ci.yml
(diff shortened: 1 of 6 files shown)


cabal.project
```diff
@@ -34,47 +34,26 @@ packages:
...
-optional-packages: vendored/*/*.cabal
+index-state: 2023-10-06T06:12:29Z
 
 tests: True
+test-show-details: direct
 
--- mfsolve has duplicate instances in its test suite
--- See: https://github.com/kuribas/mfsolve/issues/8
-package mfsolve
-  tests: False
```
fendor, line 47: Is this not needed anymore?

fendor (approved):
LGTM, only one question

cabal.project
michaelpj: I can't see how it ever was - unless we had a srp for `mfsolve` at some point.

Issue #3634, opened 2023-06-10
Plugins cannot define jump-to-definition handlers (responses are not combined)


jvanbruegge:
As can be seen here https://github.com/haskell/haskell-language-server/blob/master/hls-plugin-api/src/Ide/Types.hs#L555 HLS always returns the first response for jump-to-definition, which happens to be ghcide. Multiple responses could be combined into one, by turning `LocationLink` into a `Location`

michaelpj:
In fact, I think `Location` can be turned into `LocationLink`, which is the richer type:
- `uri` -> `targetUri`
- `range` -> `targetRange`
- `range` -> `targetSelectionRange`
- don't provide `originSelectionRange`

And the other direction:
- `targetRange` -> `range`
- `targetUri` -> `uri`

So I think what we would want to do is check the client capability for `LocationLink` support, and depending on whether it's set translate `Location`s to `LocationLink`s or vice versa.

michaelpj:
... and then just concatenate the results

joyfulmantis:
Currently the type the plugins provide for is the same as the ones lsp sends on to the client. I was thinking it would be nice to have a type the plugins need to provide that is separate from the actual method type the client accepts, specifically by being more specific. In this case only accepting locationlink for example, and then refactor the combine responses to convert that type into the client accepted type (ideally by querying capabilities) before combining. Separate types would allow us to enforce that the plugin is only returning our preferred type.

michaelpj:
Yeah, I agree that that would be nice, might be a bit of a pain to implement. But we could almost certainly provide types that are e.g. easier to combine.

jvanbruegge:
This was fixed by #3846

PR #3846, opened 2023-10-16
Merge definitions from all plugins for Document(Type)Definition message

JiriLojda:
This PR fixes issue #3634 by merging the definition responses from all plugins into one definition response. It (only when necessary) transforms `Location` to `LocationLink` as suggested by @michaelpj in the issue.
I didn't refactor the type the plugins return as suggested by @joyfulmantis, because I think it will be better to address it in a separate PR and have a straightforward fix merged before that.


hls-plugin-api/src/Ide/Types.hs
```diff
@@ -693,6 +693,24 @@ nullToMaybe' :: (a |? (b |? Null)) -> Maybe (a |? b)
 nullToMaybe' (InL x)       = Just $ InL x
 nullToMaybe' (InR (InL x)) = Just $ InR x
 nullToMaybe' (InR (InR _)) = Nothing
+
+type Definitions = (Definition |? ([DefinitionLink] |? Null))
+
+mergeDefinitions :: Definitions -> Definitions -> Definitions
+mergeDefinitions definitions1 definitions2 = case (definitions1, definitions2) of
```
fendor, line 700: Some haddock explaining what this does, especially what the merging rules are would be great!

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -693,6 +693,24 @@ nullToMaybe' :: (a |? (b |? Null)) -> Maybe (a |? b)
...
+    (InL def1, InL def2) -> InR $ InL (defToLinks def1 ++ defToLinks def2)
+    (InL def1, InR (InL links)) -> InR $ InL (defToLinks def1 ++ links)
+    (InR (InL links), InL def2) -> InR $ InL (links ++ defToLinks def2)
+    (InR (InL links1), InR (InL links2)) -> InR $ InL (links1 ++ links2)
+    where
+        defToLinks :: Definition -> [DefinitionLink]
+        defToLinks (Definition (InL location)) = [DefinitionLink $ locationToLocationLink location]
+        defToLinks (Definition (InR locations)) = map (DefinitionLink . locationToLocationLink) locations
+
+        locationToLocationLink :: Location -> LocationLink
```
fendor, line 712: Nitpick

How about:

```suggestion
        locationToDefinitionLink :: Location -> DefinitionLink
```

Since you use `locationToLocationLink` exclusively to create a `DefinitionLink`.

fendor (approved):
LGTM, thank you for the addition!

One nitpick and a little bit more documentation, looks good to me as is otherwise :)

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -693,6 +693,24 @@ nullToMaybe' :: (a |? (b |? Null)) -> Maybe (a |? b)
...
+        locationToLocationLink Location{_uri, _range} = LocationLink{_originSelectionRange = Just _range, _targetUri = _uri, _targetRange = _range, _targetSelectionRange = _range}
```
michaelpj, line 713: This is wrong, `originSelectiomRange` is quite different and we can't come up with a definition for it, so we should put `Nothing`.

hls-plugin-api/src/Ide/Types.hs
michaelpj, line 700: There's are client capabilities controlling whether location links are supported. Ideally we should check those and if they're not sent we should instead downgrade the `LocationLink`s.

michaelpj (changes requested):
I know we don't have tests for any of the other response combination, but if you felt motivated it would be great to start some!

JiriLojda:
Thank you both for the review. I will address the comments as soon as possible, most likely this week.

hls-plugin-api/src/Ide/Types.hs
JiriLojda: I created a new `downgradeLinks` function instead and used it to downgrade all `LocationLink`s in the implementation of `combineResponses` when required by the missing capability.

JiriLojda:
Oh, it seems I was too eager with the new extensions. I thought I could use GHC 9.6 (probably since it is in the snapshot in the default `stack.yml`). I'll fix that.

michaelpj (approved):
Code looks good, some of this stuff probably belongs in `lsp`, but we can move that later if we want.

Haven't reviewed the tests yet, but happy to merge regardless.

michaelpj:
Build failures are genuine.

JiriLojda:
I suspect that thanks to the CI cache `cabal` doesn't know about the `focus@1.0.3.2` which is necessary for the build to work on `ghc@9.8` (the cause of the previous build failures). I will find a way to clear the cache later.

By the way, do you generally use the `cabal-fmt`? It is configured to run with `stylish-haskell`, but all the cabal files I format are completely changed. I commit it anyway because I suppose this is the correct format that we should follow, but I wonder why it wasn't formatted before and also worry that I will cause someone ugly merge conflicts with the changes. :)

michaelpj:
CI issues should be worked out, can you fix the conflicts and try again?

michaelpj:
Thanks!

JiriLojda:
Great, thank you for your help!

PR #3848, opened 2023-10-17
Rename only if the current module compiles (#3799)

### LINKED ISSUE #3799: Renaming in presence of type errors
### Your environment

Which OS do you use?
Fedora
9.6.2 from ghcup

vim+coc.nvim

no

### Steps to reproduce

Use the following file
```hs
module X where

data X a where
  X :: X Int
  Y :: X String
  Z :: X Int

a :: X Int
a = X

-- this doesn't type check on purpose:
instance Eq X where
  (==) = \_ _ -> True
```
Try to rename the `X` data type to `X'`.

### Expected behaviour

Either nothing should be changed or everything even in presence of type errors.

### Actual behaviour

```hs
module X where

data X a where
  X :: X Int
  Y :: X String
  Z :: X Int

a :: X Int
a = X

-- this doesn't type check on purpose:
instance Eq X' where
  (==) = \_ _ -> True
```
i.e. non of the `X` occurrences in the defintiion of `X` or `a` changed, only its occurrence in the `Eq` instance.

sgillespie:
This PR resolves #3799 by using `useE` rather than `useWithStaleE`. I did not implement the suggestion to have two modi, for local renames only and for cross module renaming, rather I just fixed the issue raised in the issue.


fendor (changes requested):
The changes look good to me!

Can we add a test case for this scenario?
E.g. start with a well-typed module, wait for some diagnostics, modify it such that compilation fails, wait for diagnostics, fix the compile error and request to rename an identifier.
I think, the test case should be flaky (e.g. sometimes fail) without this patch and always succeed with this patch applied.

sgillespie:
I've added a new test for `hls-rename-plugin`, and I've verified it fails for `HEAD^` and succeeds for `HEAD`. I'm not sure about the organization of the new tests (the other tests are straightforward golden tests), so let me know if you want me to change that. I used `hls-hlint-plugin` for inspiration.

fendor (approved):
LGTM, thanks!

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -210,26 +212,26 @@ refsAtName state nfp name = do
...
 
 handleGetHieAst ::
     MonadIO m =>
     IdeState ->
     NormalizedFilePath ->
-    ExceptT PluginError m (HieAstResult, PositionMapping)
+    ExceptT PluginError m HieAstResult
 handleGetHieAst state nfp =
-    fmap (first removeGenerated) $ runActionE "Rename.GetHieAst" state $ useWithStaleE GetHieAst nfp
+    fmap removeGenerated $ runActionE "Rename.GetHieAst" state $ useE GetHieAst nfp
```
michaelpj, line 234:
```suggestion
    -- We explicitly do not want to allow a stale version here - we only want to rename if
    -- the module compiles, otherwise we can't guarantee that we'll rename everything,
    -- which is bad (see https://github.com/haskell/haskell-language-server/issues/3799)
    fmap removeGenerated $ runActionE "Rename.GetHieAst" state $ useE GetHieAst nfp
```

michaelpj (approved):
Please add a comment, otherwise this subtle but important information will be lost.

fendor:
@sgillespie Do you the capacity to push this over the finish line? It looks like, resolving merge conflicts and updating the comment should be enough :)

sgillespie:
@fendor I'm sorry I missed everything way back in Nov-Jan. In any case, I've updated from master and added the comment.

fendor (approved):
LGTM, thanks for being patient with us!

PR #3854, opened 2023-10-26
Fix #3847

### LINKED ISSUE #3847: Last line removed unexpectly when doing whole file range formatting with stylish-haskell.
### Your environment

Which OS do you use?
Ubuntu on WSL2 on Windows
9.6.3 from ghcup
Neovim + lspconfig
2.4.0.0 from ghcup

### Steps to reproduce
1. Config stylish-haskell as format provider.
2. Select all lines and do a range formatting, you'll see the last line get removed.

### Expected behaviour
The last line should not be removed.

### Actual behaviour
The last line get removed unexpectly.

### Debug information
After some investigation(took a look at messages passed between client & server and hls's source code), I'm quite sure it is caused by `Ide.PluginUtils.extractTextInRange`
```haskell
-- | Extracts exact matching text in the range.
extractTextInRange :: Range -> T.Text -> T.Text
extractTextInRange (Range (Position sl sc) (Position el ec)) s = newS
  where
    focusLines = take (fromIntegral $ el - sl + 1) $ drop (fromIntegral sl) $ T.lines s
    -- NOTE: We have to trim the last line first to handle the single-line case
    newS =
      focusLines
        & _last %~ T.take (fromIntegral ec)
        & _head %~ T.drop (fromIntegral sc)
        -- NOTE: We cannot use unlines here, because we don't want to add trailing newline!
        & T.intercalate "\n"
```
Now, let me explain.

Suppose we have codes like:
```haskell
module Main where

main :: IO ()
main = do
  putStrLn "Hello, World"
```
When I select all lines and call format in neovim, it sends a range formatting message with the range (0, 0) ~ (4, 25), and after range normalization, `extractTextInRange` received the range (0, 0) ~ (5, 0) along with the whole file content.

And then, `T.lines` returned a `List` with 5 `Text` inside an further been dropped by 0 and taken by 6. It's taken by 6 but there are only 5 `Text` available. And when creating `newS` the last `Text` is removed by `(T.take (fromIntegral 0))`.

Note that this bug is only triggered by a whole file range formatting, in the example above, if I select only the top 4 lines, `T.lines` will be dropped by 0 and taken by 5, and there are 5 `Text` available so no trouble at all.

BurningLutz:


fendor (changes requested):
Hi, thank you for the fix!

Could you include a unit test for `extractTextInRange` that demonstrates the issue?

BurningLutz:
@fendor Sure! I'll do it this weekend.

fendor (approved):
Thanks, this looks good to me!

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -236,7 +236,13 @@ usePropertyLsp kn pId p = do
 extractTextInRange :: Range -> T.Text -> T.Text
 extractTextInRange (Range (Position sl sc) (Position el ec)) s = newS
   where
-    focusLines = take (fromIntegral $ el - sl + 1) $ drop (fromIntegral sl) $ T.lines s
+    -- NOTE: Always append an empty line to the end to ensure there are
```
michaelpj, line 239: This is a little short - it's okay but it would be helpful to say _why_ we don't have enough lines otherwise. You did all that reasoning in the ticket but now we're losing it all!

michaelpj (approved):

hls-plugin-api/src/Ide/PluginUtils.hs
BurningLutz: @michaelpj 
Oh yes, I thought the ticket could be enough. Should I simply give a link to that ticket or briefly describe the issue in the NOTE?

michaelpj: Well, how would someone find the ticket? The way I think about it is: what would *you* want to read if you were encountering this code for the first time and you had no idea about the history?

BurningLutz:
As @michaelpj suggested, I described the issue briefly in the comment, and gave a link to the ticket for details.

michaelpj:
Great, thanks!

PR #3855, opened 2023-10-29
Load default operator fixities in Fourmolu plugin non-CLI mode

### LINKED ISSUE #3788: Fourmolu plugin formats code differently compared to separate installation of fourmolu
### Your environment

Which OS do you use?
MacOS

9.6.2 from GHCUp

VSCode and vscode-haskell

 2.2.0.0 from GHCUp

Whether you have `fourmolu.yaml` or not doesn't change the behaviour.

### Steps to reproduce

Have the HLS configured to use the built-in installation of fourmolu and run the formatting command.

```hs
foo :: Maybe Int -> Int
foo = undefined

thing :: Int
thing =
    foo $ Just $
            case undefined of
                True -> 1
                False -> 2
```

### Expected behaviour

The could should be formatted with the following output like a separate installation of `fourmolu-0.13.1.0` does.

```hs
foo :: Maybe Int -> Int
foo = undefined

thing :: Int
thing =
    foo $
        Just $
            case undefined of
                True -> 1
                False -> 2
```

### Actual behaviour

Running VSCode's formatting command while having HLS use the built-in fourmolu will format the code to the following form:
```hs
foo :: Maybe Int -> Int
foo = undefined

thing :: Int
thing =
    foo
        $ Just
        $ case undefined of
            True -> 1
            False -> 2
```

### Debug information

vscode-haskell outputs the following log entry:
```
2023-09-03T12:47:51.063236Z | Debug | fourmolu: Using compiled in fourmolu-0.13.1.0
```

georgefst:
Fixes #3788. I'm not particularly familiar with this part of the API, but this seems to do the job. It's similar to what the Ormolu plugin already does.


fendor (approved):
LGTM, can we have a test that shows the changed behaviour?

fendor (approved):
Awesome, thanks!

PR #3860, opened 2023-11-05
 Add option for setting manual path to Fourmolu binary 

georgefst:
I should probably look in to making this work more like `serverExecutablePath`, where it's empty by default, and allows expands `~`. But I'm not sure where that logic lives - I see no results in the whole codebase when searching for "serverExecutablePath".


fendor:
`serverExecutablePath` is a feature implemented by vscode-haskell, not HLS! And the support for `~` is a simple search and replace operation. https://github.com/haskell/vscode-haskell/blob/master/src/utils.ts#L181 and https://github.com/haskell/vscode-haskell/blob/master/src/utils.ts#L188

However, the idea is the same, if the config option is either `""` or `null`, then use the built-in formatter, otherwise use the PATH.

michaelpj:
Can we ask the user's shell to expand the path? That's the limiting case of what we want, right?

michaelpj:
Returning to this: I think it's pretty sensible for HLS to expect an actual full path, and for it to be the job of the client to shell-expand paths or whatever. So I think the current implementation of this seems fine. 

Should we update and merge this?

soulomoon:
LGTM, maybe we can update merge this.

fendor (approved):

soulomoon (approved):

Issue #3827, opened 2023-10-03
!!MISSING: command!! is intermittently displayed instead of type class function signatures


konnik:
### Your environment

Which OS do you use?

Macbook Air M2 15" (Ventura 13.6) 
Intel x86_64 (Ubuntu 20.04.6 LTS)

Same issue on both systems.

Which version of GHC do you use and how did you install it?

Installed with `ghcup`:

```
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.6.3

$ cabal --version
cabal-install version 3.10.1.0
compiled using version 3.10.1.0 of the Cabal library 
```

How is your project built (alternative: link to the project)?

https://github.com/konnik/haskell-json-parser

Which LSP client (editor/plugin) do you use?

VSCode + haskell.haskell

Which version of HLS do you use and how did you install it?

```
$ haskell-language-server-wrapper --version
haskell-language-server version: 2.3.0.0 (GHC: 9.2.8) (PATH: /Users/niklas/.ghcup/hls/2.3.0.0/lib/haskell-language-server-2.3.0.0/bin/haskell-language-server-wrapper)
```

Installed by `ghcup`

Have you configured HLS in any way (especially: a `hie.yaml` file)?

no

### Steps to reproduce

Really don't know how to reproduce it. It's very intermittent and resolves itself. 

### Expected behaviour

The code lens (don't know if this is the correct termiology)  in vscode should show the type class signatures on my instance:

```haskell
instance Monad Parser where
    return :: a -> Parser a             // this is shown by vscode, not part of the source
    return = pure
    (>>=) :: Parser a -> (a -> Parser b) -> Parser b                   // this is shown by vscode, not part of the source
    pa >>= fab = Parser $ \input -> do
        (input', a) <- runParser pa input
        runParser (fab a) input'
```

### Actual behaviour

```haskell
instance Monad Parser where
    !!MISSING: command!!            // this is shown by vscode, not part of the source
    return = pure
    !!MISSING: command!!                  // this is shown by vscode, not part of the source
    pa >>= fab = Parser $ \input -> do
        (input', a) <- runParser pa input
        runParser (fab a) input'
```

If i click on the text "!!MISSING: command!!" the error message "command 'missing' not found" is shown in bottom right of vscode

### Debug information

Some things from the logs. Don't know if its relevant:

```
[...]

PR #3864, opened 2023-11-11
Improve no plugin messages

### LINKED ISSUE #3842: Neovim Autocomplete with HLS >= 2.1.0.0 throws error
### Your environment

Which OS do you use?
Arch Linux
9.0.2 ghcup
cabal

Neovim + haskell-tools + nvim_coq
(All latest)
Version 2.1.0.0 (Though I've tried all the way up to 2.4.0.0)
Version 2.0.0.1 does work for me
Installed via ghcup
no

### Steps to reproduce
cabal init
cabal build
nvim app/Main.hs

wait for server to load

type in fm and wait for fmap to show up
tab to the "fmap" that has "LSP" on the side
(see pics below) for more

### Expected behaviour

It should complete or show various info on it
2023-10-13-185958_1054x1055_escrotum

### Actual behaviour

When hovering over on autocomplete of a name, and the name has "LSP" on the side it pops up a window with an error.
This does not occur if on the right side it is a "BUF" based suggestion.
2023-10-13-190118_1057x1058_escrotum

### Debug information

HLS log
2023-10-13T23:42:21.023882Z | Error | LSP: incoming message parse error:
Error in $.params.insertTextMode: parsing Integer failed, expected Number, but encountered Null
when processing
{"method":"completionItem\/resolve","params":{"label":"spark#","detail":"from GHC.Base","sortText":"05","data":null,"insertTextMode":null,"insertTextFormat":1,"documentation":{"kind":"markdown","value":"*Defined in 'GHC.Base'*\n"},"command":{"arguments":[{"doc":"file:\/\/\/home\/maeve\/projects\/haskell\/algorithm-design\/app\/Main.hs","importName":"GHC.Base","newThing":"spark#"}],"command":"404320:ghcide-extend-import-action:extendImport","title":"extend import"},"kind":3},"id":6,"jsonrpc":"2.0"}

I believe it might have something to do with the insertTextMode in plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Completer/Simple.hs

insertTestMode Specification (Might be wrong version for nvim):
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertTextMode

Error seems to refer to the text mode not being an integer or number as expected in the specification above.

joyfulmantis:
This PR rewrites the way we check if a plugin is enabled for a request. This allows us to give more in-depth logging when we don't have any plugins that accept a request.

Things left to do for this PR

- [x] Documentation
- [x] Ensure edge cases work well


michaelpj:
Okay, just getting to this. I'm unsure about the bouncing, tbh. I think it's probably just going to mask failures, which might be bad. What did you think about my suggestion [here](https://github.com/haskell/haskell-language-server/issues/3827#issuecomment-1854287463)? I think we could do something similar for completions: use a key that's something like just the fully qualified name or something. Even if we had to do a mixed strategy (e.g. use the fully qualified name for non-local names, use a unique key for local names), that still means that we'd get resolve hits for e.g. all non-local names even if things are a bit stale.

Could we do the plugin error bit separately while we think about it?

joyfulmantis:
> Okay, just getting to this. I'm unsure about the bouncing, tbh. I think it's probably just going to mask failures, which might be bad. What did you think about my suggestion [here](https://github.com/haskell/haskell-language-server/issues/3827#issuecomment-1854287463)? I think we could do something similar for completions: use a key that's something like just the fully qualified name or something. Even if we had to do a mixed strategy (e.g. use the fully qualified name for non-local names, use a unique key for local names), that still means that we'd get resolve hits for e.g. all non-local names even if things are a bit stale.
> 
> Could we do the plugin error bit separately while we think about it?

Okay sounds good, I'll dump the bounce stuff from this PR then, and we can revisit it later if we still need it.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -72,16 +73,17 @@ instance Pretty Log where
...
   let err = ResponseError (InR ErrorCodes_MethodNotFound) msg Nothing
       msg = pluginNotEnabled m fs'
   return $ Left err
-  where pluginNotEnabled :: SMethod m -> [PluginId] -> Text
+  where pluginNotEnabled :: SMethod m -> [(PluginId, PluginStatus)] -> Text
+        pluginNotEnabled method [] = "No plugin installed for this " <> T.pack (show method) <> " request."
         pluginNotEnabled method availPlugins =
-            "No plugin enabled for " <> T.pack (show method) <> ", potentially available: "
-                <> (T.intercalate ", " $ map (\(PluginId plid) -> plid) availPlugins)
+            "No plugin enabled for this " <> T.pack (show method) <> " request.\n Plugins installed for this method, but not enabled for this request are:\n"
```
michaelpj, line 85: I think this message is still confusing, and I think it's because of the "enabled" terminology. Something is usually enabled/disabled by an external force, and we have a _different_ use of that (which IMO makes more sense) for plugins that are disabled by the config.

So how about we instead say something like
```
No plugin wanted to handle this request.
Plugins which can handle <method> requests but which declined:
- <plugin 1>: disabled by config
... etc...
```
i.e. go with the "declined" metaphor instead. WDYT?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -383,168 +406,201 @@ class HasTracing (MessageParams m) => PluginMethod (k :: MessageKind) (m :: Meth
...
-    -> Bool
+    -> PluginStatus
     -- ^ Is this plugin enabled and allowed to respond to the given request
     -- with the given parameters?
 
   default pluginEnabled :: (L.HasTextDocument (MessageParams m) doc, L.HasUri doc Uri)
-                              => SMethod m -> MessageParams m -> PluginDescriptor c -> Config -> Bool
-  pluginEnabled _ params desc conf = pluginResponsible uri desc && plcGlobalOn (configForPlugin conf desc)
+                              => SMethod m -> MessageParams m -> PluginDescriptor c -> Config -> PluginStatus
+  pluginEnabled _ params desc conf =
```
michaelpj, line 415: I also think we should rename this method...

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -383,168 +406,201 @@ class HasTracing (MessageParams m) => PluginMethod (k :: MessageKind) (m :: Meth
...
     where
         uri = params ^. L.textDocument . L.uri
 
--- ---------------------------------------------------------------------
--- Plugin Requests
--- ---------------------------------------------------------------------
+-- | Only check if the file type is supported
+pluginEnabledOnlyFileType :: (L.HasTextDocument (MessageParams m) doc, L.HasUri doc Uri)
+                              => SMethod m -> MessageParams m -> PluginDescriptor c -> Config -> PluginStatus
+pluginEnabledOnlyFileType  _ msgParams pluginDesc _ = pluginResponsible uri pluginDesc
```
michaelpj, line 423: I think you could just inline `pluginResponsible` into here now?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -383,168 +406,201 @@ class HasTracing (MessageParams m) => PluginMethod (k :: MessageKind) (m :: Meth
...
-    where
-      uri = msgParams ^. L.textDocument . L.uri
+
+  pluginEnabled = pluginEnabledWithFeature plcSymbolsOn
 
 instance PluginMethod Request Method_CompletionItemResolve where
   -- See Note [Resolve in PluginHandlers]
-  pluginEnabled _ msgParams pluginDesc config = pluginResolverResponsible (msgParams ^. L.data_) pluginDesc
-    && pluginEnabledConfig plcCompletionOn (configForPlugin config pluginDesc)
+  pluginEnabled _ (CompletionItem {_data_=Nothing}) (PluginDescriptor {pluginId=PluginId "ghcide-completions-bounce"}) _ = PluginEnabled
```
michaelpj, line 494: This seems odd?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1158,11 +1197,13 @@ installSigUsr1Handler h = void $ installHandler sigUSR1 (Catch h) Nothing
...
-pluginResolverResponsible (Just (fromJSON -> (Success (PluginResolveData o _ _)))) pluginDesc =
-  pluginId pluginDesc == o
+pluginResolverResponsible :: Maybe Value -> PluginDescriptor c -> PluginStatus
+pluginResolverResponsible (Just (fromJSON -> (Success (PluginResolveData o@(PluginId ot) _ _)))) pluginDesc =
+  if pluginId pluginDesc == o
+    then PluginEnabled
+    else PluginDisabled $ NotResolveOwner ot
 -- We want to fail closed
-pluginResolverResponsible _ _ = False
+pluginResolverResponsible _ _ = PluginDisabled $ NotResolveOwner ""
```
michaelpj, line 1206: should have text?

hls-plugin-api/src/Ide/Types.hs
michaelpj: Or alternatively, I think it would be reasonable to have `pluginEnabled` tell us the "global" enabled/disabled status (and hence doesn't look at the current request), and then have `pluginCanRespond` which checks if the plugin is disabled and also anything that depends on the current request.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1028,10 +1059,18 @@ configForPlugin :: Config -> PluginDescriptor c -> PluginConfig
...
+pluginGlobalEnabled :: PluginConfig -> PluginStatus
+pluginGlobalEnabled pc = if plcGlobalOn pc
+                           then PluginEnabled
+                           else PluginDisabled DisabledByConfig
+
 -- | Checks that a given plugin is both enabled and the specific feature is
 -- enabled
-pluginEnabledConfig :: (PluginConfig -> Bool) -> PluginConfig -> Bool
-pluginEnabledConfig f pluginConfig = plcGlobalOn pluginConfig && f pluginConfig
+pluginEnabledConfig :: (PluginConfig -> Bool) -> PluginConfig -> PluginStatus
```
michaelpj, line 1070: It feels like we're moving towards the definition of `pluginEnabled` listing the various predicates that are tried and monoidally combining them together. So it feels a bit off to be doing that partially inside here. I'd be tempted to define

```
pluginEnabledConfig f pluginConfig = if f pluginConfig then PluginEnabled else PluginDisabled DisabledByConfig
```
and then just have
```
pluginGlobalEnabled pluginConfig <> pluginEnabledConfig codeLensesOn pluginConfig
```
or whatever at the use sites.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -80,7 +81,7 @@ rules recorder plId = do
   define (cmapWithPrio LogShake recorder) $
     \GetStanDiagnostics file -> do
       config <- getPluginConfigAction plId
-      if pluginEnabledConfig plcDiagnosticsOn config then do
+      if pluginEnabledConfig plcDiagnosticsOn config  == PluginEnabled then do
```
michaelpj, line 84: This sort of thing would look nicer if we split the "plugin is globally enabled" method from the "plugin wants to respond to this request" method. Then it would be sensible to generally call the "globally enabled" method from rules, since it of course doesn't look at the current request. And then we wouldn't be at risk of this getting out of sync.

michaelpj (approved):
I like the general direction, a few thoughts

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: Yes, thanks for catching this. It is leftover from the bounce stuff. Cleaned it up!

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: It seems like we don't need to provide a way to other parts of the code to determine whether a plugin is enabled or not here. That actually should just use plcGlobalOn. In addition, determining whether a feature is enabled should use the specific field for that. Hence I have changed the two instances in the plugins that relied on it, and have also renamed it to the more correct "handlesRequest"

michaelpj: So it seems like the specific case we have is for diagnostics produced by a plugin, where we need to check that the plugin is enabled and that diagnostics are on. Maybe we should have a specific function for that? But what's there is also fine.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -65,23 +67,29 @@ instance Pretty Log where
...
   let err = ResponseError (InR ErrorCodes_MethodNotFound) msg Nothing
-      msg = pluginNotEnabled m fs'
+      msg = noPluginHandlesMsg m fs'
   return $ Left err
-  where pluginNotEnabled :: SMethod m -> [PluginId] -> Text
-        pluginNotEnabled method availPlugins =
-            "No plugin enabled for " <> T.pack (show method) <> ", potentially available: "
-                <> (T.intercalate ", " $ map (\(PluginId plid) -> plid) availPlugins)
+  where noPluginHandlesMsg :: SMethod m -> [(PluginId, HandleRequestResult)] -> Text
+        noPluginHandlesMsg method [] = "No plugin installed for this " <> T.pack (show method) <> " request."
```
michaelpj, line 84: this should also change?

michaelpj (approved):
This looks great now

ghcide/src/Development/IDE/Plugin/HLS.hs
fendor: I think this is intentional. Since if there is no plugin for this request, there is apparently no plugin installed to handle this request. Slightly different to having plugins installed that could handle the request, but didn't due to file extensions.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -344,34 +336,79 @@ defaultConfigDescriptor :: ConfigDescriptor
...
+
+-- | Checks that a specific feature for a given plugin is enabled in order
+-- to respond to requests
+pluginFeatureEnabled :: (PluginConfig -> Bool) -> PluginDescriptor c -> Config -> HandleRequestResult
+pluginFeatureEnabled f desc conf = if f (configForPlugin conf desc)
+                                      then HandlesRequest
+                                      else DoesNotHandleRequest FeatureDisabled
+
+-- |Determine whether this request should be routed to the plugin. Fails closed
+-- if we can't determine which plugin it should be routed to.
```
fendor, line 359: What does "Fails closed" mean here?

fendor (approved):
LGTM!

hls-plugin-api/src/Ide/Types.hs
joyfulmantis: If we can't determine who this request belongs to, then we don't want any plugin to handle it. I've rephrased it to be clearer

ghcide/src/Development/IDE/Plugin/HLS.hs
joyfulmantis: > I think this is intentional. Since if there is no plugin for this request, there is apparently no plugin installed to handle this request. Slightly different to having plugins installed that could handle the request, but didn't due to file extensions.

This, but after further thought, there is no reason to make the wording here separate.

Issue #3872, opened 2023-11-15
Drop support for GHC 9.0 early


michaelpj:
GHC 9.0 is still in our support window, but:
1. Anecdotally, it's an unpopular release, with most people having gone straight from 8.10 to 9.2
2. It is relatively expensive for us to support, since the exactprint in-tree migration happened in 9.2. So if we drop it we can get rid of a bunch of CPP.

For these reasons I propose dropping GHC 9.0 after the next release.

hasufell:
+1

PR #3875, opened 2023-11-16
Drop support for GHC 9.0

michaelpj:
Fixes https://github.com/haskell/haskell-language-server/issues/3872.

I removed the macros automatically, so let's see if this builds!


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1219,12 +1215,7 @@ uses_th_qq (ms_hspp_opts -> dflags) =
 -- Depends on whether it uses unboxed tuples or sums
 computeLinkableTypeForDynFlags :: DynFlags -> LinkableType
 computeLinkableTypeForDynFlags d
-#if defined(GHC_PATCHED_UNBOXED_BYTECODE) || MIN_VERSION_ghc(9,2,0)
```
wz1000, line 1222: Remove this flag from the cabal file?

ghcide/src/Development/IDE/GHC/Compat/Core.hs
```diff
@@ -703,45 +652,27 @@ instance HasSrcSpan (SrcLoc.GenLocated (SrcSpanAnn' ann) a) where
...
-#if !MIN_VERSION_ghc(9,2,0)
-splitForAllTyCoVars :: Type -> ([TyCoVar], Type)
-splitForAllTyCoVars =
-  splitForAllTys
-#endif
 
 tcSplitForAllTyVars :: Type -> ([TyVar], Type)
 tcSplitForAllTyVars =
-#if MIN_VERSION_ghc(9,2,0)
   TcType.tcSplitForAllTyVars
```
fendor, line 664: Should we perhaps delete this simple function definition?

Its initial purpose is gone.
This raises the question again, whether we can carve out an GHC API out of our usage...

fendor (approved):
LGTM, once CI is green, this looks ready to go to me.

ghcide/src/Development/IDE/GHC/Compat/Core.hs
michaelpj: Yes, I didn't go through doing simplification, which I should...

michaelpj:
Okay, tests hopefully fixed and I did a bit of simplification in ghcide.

PR #3890, opened 2023-12-05
Demote home unit closure errors to warnings.

wz1000:
Users can't really do anything to fix them until cabal 3.12 is released.
Perhaps they could previously get by despite the unsoundness before we started
throwing these errors.

So demote them to warnings to allow HLS to continue to "function" despite them.


fendor (approved):
Looks good to me, but I didn't look at the diff in detail.

wz1000:
> Looks good to me, but I didn't look at the diff in detail.

it's quite small if you ignore whitespace changes.

Issue #1654, opened 2021-04-02
Semantic tokens in HLS


michaelpj:
I've got an open [PR](https://github.com/haskell/lsp/pull/314) for semantic tokens support in `lsp`.

This is for tracking what we'd need to do to use it in HLS.

- Does it have to go through the `hls-plugin-api` model of everything being provided as plugin methods? I'm not sure how well that will work, because a) combining results from multiple plugins won't be nice, since the tokens will all be using relative positions (we could absolutize them, combine them, and then relativize them again, though), and b) if you want to do the delta support then you need state, result ids, all that fun stuff that is probably much easier with a single provider of highlighting.
- I have no idea where this should go generally. In `ghcide`?
- What do we want to highlight?
    - Normal Haskell syntax in doctest comments? https://github.com/haskell/haskell-language-server/issues/1535
    - Type families? https://github.com/haskell/haskell-language-server/discussions/1637 
    - Other kinds of "type-like things", e.g. constraints?
    - Deprecated things?
    - Class methods?
- What types and modifiers do we want to use for highlighting? The ones which are listed in the spec aren't great for Haskelly things, and I don't know how well custom types work in practice.

pepeiborra:
* Entirely up to you, but it seems overkill to use a plugin system when there is one canonical implementation for Haskell. Sure, HLS can actually be used to target languages other than Haskell, but it is always possible to override the LSP handler directly at the lower ghcide plugin level.
* If in `hls-plugin-api`, you would add another provider and then the corresponding translation in `D.IDE.Plugin.HLS`. If in `ghcide`, then it would be registered as an LSP handler directly, as is done for e.g. document symbols (see `D.IDE.LSP.Outline`)
* Template Haskell? 
* What are we using for document symbols?

wz1000:
I'm not sure if an implementation in HLS would be worth it. Syntax highlighting already works quite well for most editors. Implementing it using an IDE sounds like a lot of work for little gain:

- Syntax highlighting has to be very low latency to present a good UX - it is unclear if this will be achievable
- It has to be tolerant of errors - a parse error means HLS has no access to the ParsedModule, so how would we return highlighting results?
- We can use stale ParsedModules, but this seems complicated to get right and seems very dubious to implement
- It will add a lot of traffic to the protocol (communication is all single-threaded) and potentially make everything else less responsive.
[...]

PR #3892, opened 2023-12-06
Implement semantic tokens plugin to support semantic highlighting(textDocument/semanticTokens/full)

soulomoon:
Working out #1654
specification: [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens)

## motivation
It enhance the original syntax highlighting by differing tokens semanticly. 
image

## What is implemented
- [x] The current implementation only [textDocument/semanticTokens/full](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_fullRequest).  
- [future] [textDocument/semanticTokens/full/delta](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_deltaRequest) and [workspace/semanticTokens/refresh](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_refreshRequest) Not planning in this pull request
we need to store last computed tokens for files in Ide state. I will try to implement them in the future after getting a deeper understanding of how hls works.

## Features:

* Supported PluginMethodHandler
  - [x] [textDocument/semanticTokens/full](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_fullRequest).
  - [future] [textDocument/semanticTokens/full/delta](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_deltaRequest) 
  - [future] [workspace/semanticTokens/refresh](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_refreshRequest)

* Supported semantic tokens type:
  - [x] class and class method
  - [x] type family name (data family)
  - [x] data constructor name (not distinguishing record and normal data, and GADT)
  - [x] type constructor name (GADT)
  - [x] record field name
  - [x] type synonym
  - [x] pattern synonym
  - [x] ~~pattern bindings~~ In favor of differing functions and none-functions from its type
  - [x] ~~value bindings~~ In favor of differing functions and none-functions from its type
  - [x] functions
  - [x] none-function variables
  - [x] imported name

* Supported modifiers(planning):
  - [future] declaration (as in class declearations, type definition and type family)
  - [future] definition (as in class instance declaration, left hand side value binding, and type family instance)
  - [future] modification (as in rec field update)

* [x] Support unicode

## Implementation details: 
- [x] Compute visible names from renamedsource
- [x] Compute `NameSemanticMap` for imported and top level name tokens using `HscEnv`(with deps) and type checked result
- [x] Compute current module `NameSemanticMap` using `RefMap a` from the result of `GetHieAst`
- [x] Compute all visible `(Name, Span)` in current module, in turn compute their semantic token using the combination map of the above two `NameSemanticMap`
- [x] use default legends, Candidates map of token type with default token type: [Maps to default token types](https://github.com/soulomoon/haskell-language-server/blob/master/plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs)
- [x] add args support to turn the plugin on and off
- [x] enhence test https://github.com/haskell/haskell-language-server/pull/3892#discussion_r1427844520
- [x] enhence error reporting. https://github.com/haskell/haskell-language-server/pull/3892#discussion_r1427955335
- [future] make use of modifiers
- [future] hadling customize legends using server capabilities (how?)

The focus of this implementation of semantic tokens is to support semantic highlight.
modifiers does not seems to provided as much use in pure language as in impure language.
I decided to ignore the modifier support for now, and schedule to the future.

Ps 
I am using the following configuration to test the stuff out in vscode.

```
    "editor.semanticTokenColorCustomizations": {
        "enabled": true, // enable for all themes
        "rules": {
            "interface": {
                "foreground": "#daf049"
            }
            ,
            "type": {
                "foreground": "#f81eba"
            }
            ,
            "typeParameter": {
                "foreground": "#cd4ca0"
            }
            ,
            "property": {
                "foreground": "#f81e"
            },
            "method": {
                "foreground": "#1ef1f8"
            },
            "class": {
                "foreground": "#ffe100"
            }
        }
    },
```


soulomoon:
Extracting information about imported names might require typecheck file that is not foi(file of interest, not opened in editors). It would consume a tons of resource for some situation.

michaelpj:
This looks like it's getting pretty far. Perhaps it would be useful for some of us to have a look? Maybe @wz1000 could look at the side that's focussed on getting information from GHC and I could take a look at the LSP bits?

soulomoon:
@michaelpj yes, I am grateful that you guys could take a look and help out.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,164 @@
...
+    let names = hieAstSpanNames nameSet ast
+    -- ask hscEnv for none local types, some from typechecked can avoid asking hscEnv
+    km <- liftIO $ foldrM (getType hsc) (tcg_type_env tmrTypechecked) nameSet
+    -- let km = tcg_type_env tmrTypechecked
+    -- global name map from type typecheck and hscEnv
+    let globalAndImportedModuleNameSemanticMap = Map.fromList $ flip mapMaybe (Set.toList nameSet) $ \name -> do
+            ty <- lookupNameEnv km name
+            return (name, tyThingSemantic ty)
+    -- local name map from refMap
+    let localNameSemanticMap = toNameSemanticMap $ Map.filterWithKey (\k _ ->
```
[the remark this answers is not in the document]
michaelpj: It would simplify things a lot to just have one source!

michaelpj (commented):
A few comments for now

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,164 @@
...
+semanticTokensFull state _ param = do
+    -- let dbg = logWith state
+    nfp <-  getNormalizedFilePathE (param ^. L.textDocument . L.uri)
+    -- dbg $ "semanticTokensFull: " <> show nfp
+    -- source :: ByteString <- lift $ getSourceFileSource nfp
+    items <- liftIO
+        $ runAction "SemanticTokens.semanticTokensFull" state
+        $ computeSemanticTokens state nfp
+    case items of
+        Nothing -> pure $ InR Null
```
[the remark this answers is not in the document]
soulomoon: Agree

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,164 @@
...
+    (hscEnv -> hsc, _) <- useWithStaleMT GhcSessionDeps nfp
+    -- because the names get from ast might contain derived name
+    let nameSet = nameGetter tmrRenamed
+    -- only take names we cares about from ast
+    let names = hieAstSpanNames nameSet ast
+    -- ask hscEnv for none local types, some from typechecked can avoid asking hscEnv
+    km <- liftIO $ foldrM (getType hsc) (tcg_type_env tmrTypechecked) nameSet
+    -- let km = tcg_type_env tmrTypechecked
+    -- global name map from type typecheck and hscEnv
+    let globalAndImportedModuleNameSemanticMap = Map.fromList $ flip mapMaybe (Set.toList nameSet) $ \name -> do
```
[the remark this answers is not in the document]
soulomoon: The computation involve up to date names though.🤔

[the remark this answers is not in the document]
soulomoon: I found them, cool I will try to implement one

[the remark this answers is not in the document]
soulomoon: getImportedNameSemanticRule rule implemented.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,167 @@
...
+    -- any pattern bind
+    TPatternBind  -> SemanticTokenTypes_Parameter
+    TClass        -> SemanticTokenTypes_Class
+    TClassMethod  -> SemanticTokenTypes_Method
+    TTypeVariable -> SemanticTokenTypes_TypeParameter
+    -- normal data type is a tagged union type look like enum type
+    -- and a record is a product type like struct
+    -- but we don't distinguish them yet
+    TTypeCon      -> SemanticTokenTypes_Enum
+    TDataCon      -> SemanticTokenTypes_EnumMember
```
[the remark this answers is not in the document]
soulomoon: Cool, weired flavor wins 👋

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,167 @@
...
+semanticTokenAbsoluteActualToken :: SemanticTokenAbsolute -> ActualToken
+semanticTokenAbsoluteActualToken (SemanticTokenAbsolute line startChar len tokenType tokenModifiers) =
+    (line, startChar, len, fromLspTokenType tokenType, 0)
+
+-- mapping from our token type to LSP default token type
+toLspTokenType :: SemanticTokenType -> SemanticTokenTypes
+toLspTokenType tk = case tk of
+    -- TVariable     -> SemanticTokenTypes_Variable
+    -- left hand side of none pattern bind
+    TValBind      -> SemanticTokenTypes_Function
```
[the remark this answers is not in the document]
michaelpj: I think I maybe need an example. Are we talking about something like this?
```
f a b c = ...
```
You want to highlight `f` differently from `a`/`b`/`c`?

I guess my argument is that this will (I think) highlight `let x = 1 in ...` with `x` as a "function", which it isn't. Hence why I'd be tempted to use "variable" for all of those. But using "parameter" for the function arguments seems nice.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
michaelpj: > Type checked module's type env only resolve top level names. Local level is missing.

Hmm, but we could traverse the renamed source to look for local names? Maybe that's not worth it. 

I forget how this goes, surely we do similar things elsewhere (@wz1000 ?)

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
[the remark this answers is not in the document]
soulomoon: may be we can use variable for all val bindings
and leave pattern to be parameters :)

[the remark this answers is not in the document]
soulomoon: also I a have an evil statement to make. 
given the fact that Haskell has no zero arity function. We have correspondence as:
f : f()
f x : f(x)
f x x : f(x,x)

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,167 @@
...
+    -- but we don't distinguish them yet
+    TTypeCon      -> SemanticTokenTypes_Enum
+    TDataCon      -> SemanticTokenTypes_EnumMember
+    TRecField     -> SemanticTokenTypes_Property
+    -- pattern syn is like a limited version of macro of constructing a data type
+    TPatternSyn   -> SemanticTokenTypes_Macro
+    -- saturated type
+    TTypeSyn      -> SemanticTokenTypes_Type
+    -- not sure if this is correct choice
+    TTypeFamily   -> SemanticTokenTypes_Interface
```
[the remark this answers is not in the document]
soulomoon: Yes type level function!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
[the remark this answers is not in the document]
michaelpj: I think it would also be somewhat reasonable to use "function" for any binding, but I think that's arguably somewhat misleading... I'm also not _totally_ sold on using "parameter" for all pattern bindings. Many pattern bindings aren't function parameters! It might be trickier to find the bindings that really are function parameters.

Your screenshot is very helpful, thanks. It seems to me that why that the second picture is helpful is not so much that it distinguishes pattern bindings from other kinds of binding, but that it distinguishes _local_ names from _global_ names. It just happens that pattern bindings are always local and most local variables are pattern bindings. Unfortunately it looks like we don't have a good way to express this in the current standard set of token types. Perhaps we could suggest they add a `local` token modifier or something?

Apart from that I'm a bit torn - generally I think we should do what the spec says is correct and then if that shows up badly it's not our problem. On the other hand, sometimes it makes a big difference to do the "wrong" thing.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
soulomoon: correction for me: not all of local level are missing, just some.

I find out that the `hover` in the code base, use `GetDocMap` to get type information.
In the `mkDocMap`, it is collecting all type inforamtion for names
from `TypeEnv` and `HscEnv`. 
It won't work for some name though. Here is one.

image

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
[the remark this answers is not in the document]
soulomoon: I agree with you for the fact that we won't have a good way of expressing haskell token type to current standard set of token types. It is more of compromise.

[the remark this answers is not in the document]
soulomoon: And also these mapping is temporary, 
I am digging more to seek, if there are a reasonable way of differing function, paramaters, and none function variable.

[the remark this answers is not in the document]
soulomoon: Also it is more of a ghc problem for this. quoted of `Note [FunBind vs PatBind]` in ghc
```
{- Note [FunBind vs PatBind]
   ~~~~~~~~~~~~~~~~~~~~~~~~~
The distinction between FunBind and PatBind is a bit subtle. FunBind covers
patterns which resemble function bindings and simple variable bindings.

    f x = e
    f !x = e
    f = e
    !x = e          -- FunRhs has SrcStrict
    x `f` y = e     -- FunRhs has Infix

The actual patterns and RHSs of a FunBind are encoding in fun_matches.
The m_ctxt field of each Match in fun_matches will be FunRhs and carries
two bits of information about the match,

  * The mc_fixity field on each Match describes the fixity of the
    function binder in that match.  E.g. this is legal:
         f True False  = e1
         True `f` True = e2

  * The mc_strictness field is used /only/ for nullary FunBinds: ones
    with one Match, which has no pats. For these, it describes whether
    the match is decorated with a bang (e.g. `!x = e`).

By contrast, PatBind represents data constructor patterns, as well as a few
other interesting cases. Namely,

    Just x = e
    (x) = e
    x :: Ty = e
-}
```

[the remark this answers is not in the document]
soulomoon: Also I am convinced that variable and function is more of a thing that user cares about than wether it is from value bindings or pattern bindings

soulomoon:
I observed a bizzared behaviour when using possition mappings
in vscode.
in the first 14 space insert, the using possition mapping to ident the code seems completely working,
Screenshot 2023-12-23 at 12 46 06
but after that, the possition starts to be broken(notice the s in the end of spanNames).
Screenshot 2023-12-23 at 12 46 53

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
soulomoon: Implemented

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,179 @@
...
+
+-- * 2. Mapping from GHC type and tyThing to semantic token type.
+
+toTokenType :: Name -> SemanticTokenType
+toTokenType locName = case occNameSpace $ occName locName of
+  x | isDataConNameSpace x -> TDataCon
+  x | isTvNameSpace x      -> TTypeVariable
+  x | isTcClsNameSpace x   -> TTypeCon -- Type constructors and classes in the same name space for now
+  x | isVarNameSpace x     -> TVariable
+  _                        -> TNothing
```
michaelpj, line 70: Looks like we can also distinguish field names this way?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,179 @@
...
+  ATyCon tyCon
+    | isTypeSynonymTyCon tyCon -> TTypeSyn
+    | isTypeFamilyTyCon tyCon -> TTypeFamily
+    | isClassTyCon tyCon -> TClass
+    -- fall back to TTypeCon the result
+    | otherwise -> TTypeCon
+  ACoAxiom _ -> TNothing
+  where
+    isFunVar :: Var -> Bool
+    isFunVar var = isFunType $ varType var
```
michaelpj, line 94: `ghc` defines `isFunType` for you!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,179 @@
...
+
+isFunType :: Type -> Bool
+isFunType a = case a of
+    ForAllTy _ t -> isFunType t
+    _x           -> isFunTy a
+
+typeSemantic :: HieKind hType -> hType -> SemanticTokenType
+typeSemantic kind t = case kind of
+  HieFresh -> if isFunType t then TFunction else TNothing
```
michaelpj, line 103: shouldn't we fall back to variable here? not quite sure what's going on here

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,126 @@
...
+        mergeNameFromSamSpan xs = Map.toList $ Map.fromListWith (<>) xs
+
+        toAbsSemanticToken :: Range -> SemanticTokenType -> Maybe SemanticTokenAbsolute
+        toAbsSemanticToken (Range (Position startLine startColumn) (Position _endLine endColumn)) tokenType =
+            let len = endColumn - startColumn
+            in SemanticTokenAbsolute (fromIntegral startLine) (fromIntegral startColumn)
+                (fromIntegral len) <$> toLspTokenType tokenType <*> return []
+        getSemantic nameMap (span, name) = do
+            let tokenType = lookupNameEnv nameMap name
+            fmap (span,) tokenType
```
michaelpj, line 123:
```suggestion
        getSemantic nameMap (span, name) = do
            tokenType <- lookupNameEnv nameMap name
            pure (span, tokenType)
```

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,126 @@
...
+extractSemanticTokensFromNames :: NameSemanticMap -> [(Range, Name)] -> [SemanticTokenAbsolute]
+extractSemanticTokensFromNames nsm =
+    mapMaybe (uncurry toAbsSemanticToken) . mergeNameFromSamSpan . mapMaybe (getSemantic nsm)
+    where
+        -- merge all tokens with same span
+        -- mergeNameFromSamSpan :: [(Span, SemanticTokenType)] -> [(Span, SemanticTokenType)]
+        mergeNameFromSamSpan xs = Map.toList $ Map.fromListWith (<>) xs
+
+        toAbsSemanticToken :: Range -> SemanticTokenType -> Maybe SemanticTokenAbsolute
+        toAbsSemanticToken (Range (Position startLine startColumn) (Position _endLine endColumn)) tokenType =
```
michaelpj, line 117: Something like this might be a useful helper to add to `lsp` :thinking:

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Types.hs
```diff
@@ -0,0 +1,67 @@
...
+    | TClass -- Type class
+    | TTypeSyn -- Type synonym
+    | TTypeFamily -- type family
+    | TRecField -- from match bind
+    deriving (Eq, Ord, Show, Enum, Bounded)
+
+instance Semigroup SemanticTokenType where
+    -- one in higher enum is more specific
+    a <> b = max a b
+instance Monoid SemanticTokenType where
```
michaelpj, line 38: `Maybe SemanticTokenType` would have the monoid instance you want!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
```diff
@@ -0,0 +1,94 @@
...
+import qualified Data.Map                        as Map
+import           Development.IDE.GHC.Compat
+import           Ide.Plugin.SemanticTokens.Types
+import           Prelude                         hiding (span)
+
+deriving instance Show DeclType
+deriving instance Show BindType
+deriving instance Show RecFieldContext
+
+instance Show ContextInfo where
```
michaelpj, line 19: can we upstream this?

plugins/hls-semantic-tokens-plugin/test/Main.hs
```diff
@@ -0,0 +1,194 @@
...
+                                                     testCase, testGroup,
+                                                     waitForAction, (@?=))
+import qualified Test.Hls.FileSystem                as FS
+import           Test.Hls.Util                      (withCanonicalTempDir)
+
+testDataDir :: FilePath
+testDataDir = "test" </> "testdata"
+
+mkFs :: [FS.FileTree] -> FS.VirtualFileTree
+mkFs = FS.mkVirtualFileTree testDataDir
```
michaelpj, line 62: cc @fendor someone is using it!

plugins/hls-semantic-tokens-plugin/test/testdata/T1.expected
```diff
@@ -0,0 +1,81 @@
+SemanticTokenOriginal {_tokenType = TTypeCon, _loc = Loc {_line = 6, _startChar = 18, _len = 3}, _name = "Set"}
```
michaelpj, line 1: I'm not totally sure why this is so much more useful than printing the `SemanticTokenAbsolute`s? We'd see the "final output" rather than the haskell token types you assign, but maybe that's fine?

michaelpj: Ah, you include the text under the span! That does seem useful!

plugins/hls-semantic-tokens-plugin/test/testdata/T1.expected
```diff
@@ -0,0 +1,81 @@
...
+SemanticTokenOriginal {_tokenType = TVariable, _loc = Loc {_line = 36, _startChar = 4, _len = 1}, _name = "f"}
+SemanticTokenOriginal {_tokenType = TVariable, _loc = Loc {_line = 36, _startChar = 7, _len = 2}, _name = "gg"}
+SemanticTokenOriginal {_tokenType = TVariable, _loc = Loc {_line = 36, _startChar = 11, _len = 2}, _name = "vv"}
+SemanticTokenOriginal {_tokenType = TVariable, _loc = Loc {_line = 37, _startChar = 10, _len = 2}, _name = "gg"}
+SemanticTokenOriginal {_tokenType = TRecField, _loc = Loc {_line = 38, _startChar = 14, _len = 3}, _name = "foo"}
+SemanticTokenOriginal {_tokenType = TFunction, _loc = Loc {_line = 38, _startChar = 18, _len = 1}, _name = "$"}
+SemanticTokenOriginal {_tokenType = TVariable, _loc = Loc {_line = 38, _startChar = 20, _len = 1}, _name = "f"}
+SemanticTokenOriginal {_tokenType = TRecField, _loc = Loc {_line = 38, _startChar = 24, _len = 3}, _name = "foo"}
+SemanticTokenOriginal {_tokenType = TRecField, _loc = Loc {_line = 39, _startChar = 14, _len = 3}, _name = "foo"}
+SemanticTokenOriginal {_tokenType = TFunction, _loc = Loc {_line = 39, _startChar = 18, _len = 1}, _name = "$"}
```
michaelpj, line 71: Hmm, maybe we should be using "operator" for infix things? But then it'll be awkward - what about infix class methods, are they methods or operators? Worth a comment at least.

plugins/hls-semantic-tokens-plugin/test/testdata/T1.expected
```diff
@@ -0,0 +1,81 @@
+SemanticTokenOriginal {_tokenType = TTypeCon, _loc = Loc {_line = 6, _startChar = 18, _len = 3}, _name = "Set"}
+SemanticTokenOriginal {_tokenType = TFunction, _loc = Loc {_line = 6, _startChar = 23, _len = 6}, _name = "insert"}
```
michaelpj, line 2: I think we could give these a `Pretty` instance and pretty print this stuff in a way that's easier to read. Something like:
```
6:23-6:29 TFunction "insert"
```
perhaps?

michaelpj (commented):
This is starting to look pretty good. I still really want @wz1000 to look at the name-analysis stuff because I don't know what the right thing to do there is.

One thing that we discussed is that we might want to have this plugin disabled by default for a release or two until we see how it works out in practice. We're potentially doing a bunch more work and sending a lot of stuff over the wire (until we implement the delta method), and we want to be careful about performance.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,179 @@
...
+  -- and a record is a product type like struct
+  -- but we don't distinguish them yet
+  TTypeCon      -> Just SemanticTokenTypes_Enum
+  TDataCon      -> Just SemanticTokenTypes_EnumMember
+  TRecField     -> Just SemanticTokenTypes_Property
+  -- pattern syn is like a limited version of macro of constructing a data type
+  TPatternSyn   -> Just SemanticTokenTypes_Macro
+  -- saturated type
+  TTypeSyn      -> Just SemanticTokenTypes_Type
+  -- not sure if this is correct choice
```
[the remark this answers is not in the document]
soulomoon: Function Variable and TypeFamily would have the same color, which might not look good. 🤔

plugins/hls-semantic-tokens-plugin/test/testdata/T1.expected
soulomoon: compact the test output would look good surely

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: We need to peels off the forall here. I don't find the ghc versions for it.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,133 @@
...
+        getTypeExclude localMap env n nameMap
+            | Nothing <- lookupNameEnv localMap n =
+                do  tyThing <- lookupImported env n
+                    pure $ maybe nameMap (extendNameEnv nameMap n . tyThingSemantic) tyThing
+            | otherwise = pure nameMap
+        lookupImported :: HscEnv -> Name -> IO (Maybe TyThing)
+        lookupImported env = fmap (fromRight Nothing) . catchSrcErrors (hsc_dflags env) "span" . lookupName env
+
+
+-- | Persistent rule to ensure that semantic tokens doesn't block on startup
```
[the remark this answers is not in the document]
soulomoon: I have been thinking about this too. Pushing the whole things into rule.

plugins/hls-semantic-tokens-plugin/README.md
```diff
@@ -0,0 +1,54 @@
...
+  * [x] record field name
+  * [x] type synonym
+  * [x] pattern synonym
+  * [x] pattern bindings (map to parameters in standard token)
+  * [x] value bindings (map to functions in standard token)
+  * [x] imported name
+
+* Supported modifiers:
+  * [ ] declaration (as in class declarations and type family)
+  * [ ] definition (as in class instance declaration and type family instance)
```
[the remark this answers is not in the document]
soulomoon: added to plan

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,179 @@
...
+lspTokenReverseMap = Map.fromList $ mapMaybe (\x -> fmap (,x) (toLspTokenType x)) $ enumFrom minBound
+
+fromLspTokenType :: SemanticTokenTypes -> SemanticTokenType
+fromLspTokenType tk = fromMaybe TNothing $ Map.lookup tk lspTokenReverseMap
+
+-- * 2. Mapping from GHC type and tyThing to semantic token type.
+
+toTokenType :: Name -> SemanticTokenType
+toTokenType locName = case occNameSpace $ occName locName of
+  x | isDataConNameSpace x -> TDataCon
```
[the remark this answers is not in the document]
soulomoon: The information obtained from namespace is too rough, I am abandoning this function.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
soulomoon: Implemented, seems working fine

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
soulomoon: I am keeping this here for convenient of development, the upstream have [Outputable](https://hackage.haskell.org/package/ghc-9.8.1/docs/src/GHC.Utils.Outputable.html#Outputable) for this.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
michaelpj: Oh I see, you want to consider _polymorphic_ functions here. Worth a comment! You could use `GHC.dropForAlls` to get rid of the leading foralls.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
michaelpj: Ah, I didn't realise this was a GHC type. And I guess the outputable instance isn't useful? We use that for other types here...

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+import           Prelude                              hiding (span)
+
+logWith :: (MonadIO m) => IdeState -> Priority -> String -> m ()
+logWith st prior = liftIO . logPriority (ideLogger st) prior . T.pack
+
+-----------------------
+---- the api
+-----------------------
+
+computeSemanticTokens :: IdeState -> NormalizedFilePath -> ExceptT PluginError Action SemanticTokens
```
michaelpj, line 71: inline?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+--
+-- It then combines this information to compute the semantic tokens for the file.
+--
+getSemanticTokensRule :: Recorder (WithPriority Log) -> Rules ()
+getSemanticTokensRule recorder =
+  define (cmapWithPrio LogShake recorder) $ \GetSemanticTokens nfp -> do
+    (TcModuleResult {..}) <- use_ TypeCheck nfp
+    (hscEnv -> hsc) <- use_ GhcSessionDeps nfp
+    (HAR {..}) <- use_ GetHieAst nfp
+    Just (_, ast) <- return $ listToMaybe $ Map.toList $ getAsts hieAst
```
michaelpj, line 102: partial match? this will throw if it fails

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+    let spanNames = hieAstSpanNames nameSet ast
+    let sMap = plusNameEnv_C (<>) importedNameSemanticMap $ mkLocalNameSemanticFromAst hieKind refMap
+    let rangeTokenType = extractSemanticTokensFromNames sMap spanNames
+    return ([], Just $ RangeHsSemanticTokenTypes rangeTokenType)
+    where
+        -- ignore one already in current module
+        getTypeExclude localMap env n nameMap
+            | Nothing <- lookupNameEnv localMap n =
+                do  tyThing <- lookupImported env n
+                    pure $ maybe nameMap (extendNameEnv nameMap n) (tyThing >>= tyThingSemantic)
```
michaelpj, line 116: Nit: personally I prefer case expressions over `maybe` unless it's partially applied

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+    let nameSet = nameGetter tmrRenamed
+    -- get imported name semantic map
+    importedNameSemanticMap <- liftIO $ foldrM (getTypeExclude (tcg_type_env tmrTypechecked) hsc) emptyNameEnv $ nameSetElemsStable nameSet
+    -- get current location from the old ones
+    let spanNames = hieAstSpanNames nameSet ast
+    let sMap = plusNameEnv_C (<>) importedNameSemanticMap $ mkLocalNameSemanticFromAst hieKind refMap
+    let rangeTokenType = extractSemanticTokensFromNames sMap spanNames
+    return ([], Just $ RangeHsSemanticTokenTypes rangeTokenType)
+    where
+        -- ignore one already in current module
```
michaelpj, line 112: Don't you merge the token types anyway? Is there a problem if we get token types from two sources?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+        getTypeExclude localMap env n nameMap
+            | Nothing <- lookupNameEnv localMap n =
+                do  tyThing <- lookupImported env n
+                    pure $ maybe nameMap (extendNameEnv nameMap n) (tyThing >>= tyThingSemantic)
+            | otherwise = pure nameMap
+        lookupImported :: HscEnv -> Name -> IO (Maybe TyThing)
+        lookupImported env = fmap (fromRight Nothing) . catchSrcErrors (hsc_dflags env) "span" . lookupName env
```
michaelpj, line 119: I can see some inconsistency in how we do this stuff. I think you got this from `Documentation`? But then it looks like we are able to get the imported names from the renamed source in `QualifyImportedNames`? More generally, I wonder if we should have a utility for doing this correctly and re-use it. I guess we can't put the "imported name env" into a rule because we I suspect we can't hash `TyThing` or something like that. But generally this whole business with looking up names seems like it ought to be a common thing to want and it would be nice to make it easy. cc @wz1000

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+
+
+-- | Persistent rule to ensure that semantic tokens doesn't block on startup
+persistentSemanticMapRule :: Rules ()
```
michaelpj, line 123: name is now slightly off?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
michaelpj: nice!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,174 @@
...
+  Decl ConDec _            -> Just TDataCon
+  Decl SynDec _            -> Just TTypeSyn
+  Decl FamDec _            -> Just TTypeFamily
+  -- instance dec is class method
+  Decl InstDec _           -> Just TClassMethod
+  Decl PatSynDec _         -> Just TPatternSyn
+  EvidenceVarUse           -> Nothing
+  EvidenceVarBind {}       -> Nothing
+
+-- * 4. Mapping from LSP tokens to SemanticTokenOriginal.
```
michaelpj, line 137: I'm still thinking about this. I wonder if we can give people a helper like this in `lsp`. It seems much better for debugging :thinking:

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,174 @@
...
+
+typeSemantic :: HieKind hType -> hType -> HsSemanticTokenType
+typeSemantic kind t = case kind of
+  HieFresh -> if isFunType t then TFunction else TVariable
+  HieFromDisk full_file ->
+    if isFixFunction fullType
+      then TFunction
+      else TVariable
+    where
+      fullType = recoverFullType t (hie_types full_file)
```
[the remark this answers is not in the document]
michaelpj: This might be an argument against trying to distinguish between things of function type and non-function type, if that's expensive :thinking: Or is this the sort of thing where instead we should consider indexing this information into `hiedb` itself?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,135 @@
...
+      spanInfos <- -- traceShow ("getting spans:", nameString) $
+        Map.lookup (Right name') rm'
+      let infos = S.unions $ map (identInfo . snd) spanInfos
+      let typeTokenType = fmap (typeSemantic hieKind) $ listToMaybe $ mapMaybe (identType . snd) spanInfos
+      let contextInfoTokenType = contextInfosMaybeTokenType infos
+      maximum <$> NE.nonEmpty (catMaybes [typeTokenType, contextInfoTokenType])
+
+    contextInfosMaybeTokenType :: Set.Set ContextInfo -> Maybe HsSemanticTokenType
+    contextInfosMaybeTokenType details = case NE.nonEmpty $ Set.toList details of
+      Just infos -> maximum $ NE.map infoTokenType infos
```
michaelpj, line 75:
```suggestion
      Just infos -> fold1 $ NE.map infoTokenType infos
```
The monoid instance is supposed to do the right thing, I think?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,135 @@
...
+    (ns, st)
+  where
+    nameSemanticFromRefMap :: RefMap a -> Name -> Maybe HsSemanticTokenType
+    nameSemanticFromRefMap rm' name' = do
+      spanInfos <- -- traceShow ("getting spans:", nameString) $
+        Map.lookup (Right name') rm'
+      let infos = S.unions $ map (identInfo . snd) spanInfos
+      let typeTokenType = fmap (typeSemantic hieKind) $ listToMaybe $ mapMaybe (identType . snd) spanInfos
+      let contextInfoTokenType = contextInfosMaybeTokenType infos
+      maximum <$> NE.nonEmpty (catMaybes [typeTokenType, contextInfoTokenType])
```
michaelpj, line 71:
```suggestion
      fold1 <$> NE.nonEmpty (catMaybes [typeTokenType, contextInfoTokenType])
```

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,135 @@
...
+            (fromIntegral len)
+            (toLspTokenType tokenType)
+            []
+
+extractSemanticTokensFromNames :: NameSemanticMap -> [(Range, Name)] -> [(Range, HsSemanticTokenType)]
+extractSemanticTokensFromNames nsm = mergeNameFromSamSpan . mapMaybe (getSemantic nsm)
+  where
+    -- merge all tokens with same span
+    -- mergeNameFromSamSpan :: [(Span, SemanticTokenType)] -> [(Span, SemanticTokenType)]
+    mergeNameFromSamSpan xs = Map.toList $ Map.fromListWith (<>) xs
```
michaelpj, line 131: Might be cleaner to just work with a `Map Range X` as much as possible until the last minute when we need a list? X could be `Set Name` or something until this point.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,135 @@
...
+
+-- * extract location from HieAST a
+
+-----------------------------------
+
+hieAstSpanNames :: NameSet -> HieAST a -> [(Range, Name)]
+hieAstSpanNames nameSet ast =
+  if null (nodeChildren ast)
+    then getIds ast
+    else concatMap (hieAstSpanNames nameSet) (nodeChildren ast)
```
michaelpj, line 88: Here's a place where it would be better to be using a map. We're doing a lot of list concatenation, and generally we're concatenating the same list prefix `n` times for a tree of depth `n`, so I think this is quadratic in the depth of the tree. Better to use a `Map` or `DList` or something with better concat performance.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
michaelpj: Alternatively, something like:
```
fold $ fmap (Just . infoTokenType) $ Set.toList details
```
using `Monoid Maybe` instance

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
michaelpj: Or even
```
fold [typeTokenType, contextInfoTokenType]
```
using the `Monoid Maybe` instance

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,124 @@
...
+-- It then combines this information to compute the semantic tokens for the file.
+--
+getSemanticTokensRule :: Recorder (WithPriority Log) -> Rules ()
+getSemanticTokensRule recorder =
+  define (cmapWithPrio LogShake recorder) $ \GetSemanticTokens nfp -> do
+    (TcModuleResult {..}) <- use_ TypeCheck nfp
+    (hscEnv -> hsc) <- use_ GhcSessionDeps nfp
+    (HAR {..}) <- use_ GetHieAst nfp
+    Just (_, ast) <- return $ listToMaybe $ Map.toList $ getAsts hieAst
+    let nameSet = nameGetter tmrRenamed
```
wz1000, line 103: Why do we need to look in the renamed source? All the names should already be in the HieAST and RefMap.

wz1000:
It seems like you are consulting a number of sources to construct the tokens, including the hie ast, renamed and typecheck sources, and the global environment using `lookupName`. It would be good to clarify which information you get from which sources. Naively I would think that only consulting the `HieAST` would be good enough.

michaelpj:
I gave this a spin. It's pretty snappy on the files I tried it on, although none of them were gigantic. A few thoughts:
- Consider `f x = x <> mempty`. `f` is highlighted as a function, x is highlighted as a variable, and `mempty` is highlighted like a function because it is a class method, even though in this case it doesn't take any arguments. This is tricky and I think a bit confusing: a name of the same type that wasn't a class method would be highlighted as a variable. I'm not sure what to do about this.
- The special highlighting for functions versus other variables is nice, but I wonder if it's confusing. If you don't know what it's doing it does look like it randomly colours certain things differently...
- We consider `c => Int` to be a function type, not sure if that's deliberate or not. I would probably say it shouldn't be.
- We don't currently use `namespace`. The obvious candidate is modules. This might be a bit awkward but it would be cool to highlight `Set` in `Set.fromList` as a namespace. Perhaps also the `Foo` in  `module Foo where`. Could be future work though.
- Also there are some token types for strings and numbers. Unclear if this is supposed to be for literals or what? Literals seem to get good highlighting anyway so I'm not sure it's necessary. Perhaps worth explicitly saying that somewhere.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: I want to let the rule fail if ast is not here, not sure what is better practice🤔

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: I would like current module stuff's semantic type coming from single source , the hie. Less error prone.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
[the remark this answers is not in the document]
soulomoon: > Calling `recoverFullType` on every type can be very expensive, see https://gitlab.haskell.org/ghc/haddock/-/blob/b0b0e0366457c9aefebcc94df74e5de4d00e17b7/haddock-api/src/Haddock/Backends/Hyperlinker/Utils.hs#L107 for an implementation with sharing.
> 
> 
> 
> I imagine you want to map `hie_types` to something like `A.Array TypeIndex Bool`, ensuring we only look at each index once.

Sure I'll see how to arrange this in the query logic.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: I have been thing about this too, renamed source just provide the visible names.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: Should not be a problem if more comming from here though.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
michaelpj: Sure, but this will fail with a pattern match exception which I don't think will be caught. Not 100% sure. Better to fail explicitly, there are some functions for doing this with things returning Maybe somewehre.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
wz1000: Yes, I don't think it is necessary to consult the renamed source.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
soulomoon: good idea

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: Maybe we return Nothing in the result?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -0,0 +1,121 @@
...
+getSemanticTokensRule recorder =
+  define (cmapWithPrio LogShake recorder) $ \GetSemanticTokens nfp -> do
+    (hscEnv -> hsc) <- use_ GhcSessionDeps nfp
+    (HAR {..}) <- use_ GetHieAst nfp
+    Just (_, ast) <- return $ listToMaybe $ Map.toList $ getAsts hieAst
+    -- get current location from the old ones
+    let spanNamesMap = hieAstSpanNames ast
+    let nameSet = unionNameSets $ Map.elems spanNamesMap
+    -- get imported name semantic map
+    importedNameSemanticMap <- liftIO $ foldrM (getTypeExclude hsc) emptyNameEnv $ nameSetElemsStable nameSet
```
wz1000, line 106: why do we need to consult the `hsc` environment for the type? Can't we get them from the hie ast or the refmap?

soulomoon: we currently use tyThing to ease the implementation, since GHC provide a handful of functions to distinguish the token semantic. Not sure about how to recover tyThing  from hie ast or the refmap. 
Or Is there a better way than using tyThing?

soulomoon: For either way, please enlighten me, I'll take a look at them latter.

wz1000: why can't we use `typeSemantic` instead of `tyThingSemantic`?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -0,0 +1,120 @@
...
+
+-- * extract names from RenamedSource
+
+------------------------------------------------
+
+nameToCollect :: Name -> NameSet
+nameToCollect = unitNameSet
+
+nameGetter :: RenamedSource -> NameSet
+nameGetter = everything unionNameSet (emptyNameSet `mkQ` nameToCollect)
```
wz1000, line 43: These can be deleted now I guess?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: I tried, but I failed find enough semantic information from HieTypeFix, such as class method, pattern synonym etc..

wz1000: what about in the `OccName` and its `NameSpace`?

soulomoon:
# Imported names

What I know:

* for hie types, `IfaceTyCon` in hie type does not contains enough information to distinguish class, type syn, type family etc, they are all type contractor.
* most imported name only provide usage in the `ContextInfo`
* `namespace` in Name is limited, up to 9.8, we can only classify `VarName, FldName, DataName, TvNamem, TcClsName`.
* `WiredIn` names have `TyThing` attached, seems only for type synonym.

## For imported names inferring directly from hie ast, hie types and name(occName, namespace)

Here is the list of imported names and their recoverability to semantic info:

1. [?] type class:
    1.1. [yes?] might be recovered awkwardly from hie ast (ContextInfo. EvidenceVarBind.EvVarSource.EvInstBind.cls) in the same node

    ```haskell
    data Foo = Foo Int deriving (Eq)
    ```

    1.2. [yes] same as 1

    ```haskell
    data Foo = Foo Int 
    instance Eq Foo where
        (==) = undefined
    ```

    1.3. [no] this can not be recover

    ```haskell
    type G = Eq
    ```

    1.4. [no]

    ```haskell
    go :: (Eq a) => a -> Bool
    go = undefined
    ```

3. [yes] field is marked in hie ast' ContextInfo

```haskell
go = Data.Time.Clock.utctDay
```

4. [yes] type synonym: it is not distinguishable from normal type construct in hie ast or types
     but it can be obtained from `Name` since it is wired-in ()

```haskell
v :: String
v = undefined
```

6. [yes] data type: y, just normal type constructor from namespace

```haskell
a :: Maybe Int
a = undefined
```

7. [no] type family: no different from normal type construct in hie ast

```haskell
type A = TypeError (Text "The type " :<>: ShowType Int :<>:
                            Text " is not exportable.")
```

8. [?] class method:  

    7.1. [no] usage

    ```haskell
    x :: Int -> Int -> Bool
    x = (==) 
    ```

    7.2. [yes] instance decl, in val bind of instance type

    ```haskell
    data Foo = Foo Int 
    instance Eq Foo where
        (==) = undefined
    ```

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: no luck here too.

soulomoon:
@michaelpj 
Since we are using GHC funtion through `Development.IDE.GHC.Compat`, `Development.IDE.GHC.Compat.Core.FunTy(pattern synonym)` hides FunTyFlag which is used to distinguish
 (->, =>, etc..).
 This is why currently why we inferring `c => x` as function for now.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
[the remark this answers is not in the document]
soulomoon: cool, I'll take a look at this.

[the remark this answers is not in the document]
soulomoon: adjusted

wz1000:
Thanks @soulomoon for that characterization. I believe you can get the `TyThings` for all type constuctors using the kindMap field of the `GetDocMap` rule, so we can avoid doing these lookups multiple times.

I guess you could relax this check to also collect information about variables and constructors in the same rule: https://github.com/haskell/haskell-language-server/blob/0be6fa7d8591bebbd6e34ee78b79a10b11762973/ghcide/src/Development/IDE/Spans/Documentation.hs#L64

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
michaelpj: See e.g. `handleMaybe`

michaelpj:
@wz1000 if this information is not in HIE files, should it be? Should we make a ticket to add it? I can imagine other consumers of HIE files wanting to know "what kind of thing" a name is.

>  This is why currently why we are inferring c => x as function for now.

Okay, worth a comment explaining that.

> For the UnicodeSyntax, I am sorry that I do not understand much of Unicode stuff. It would be nice if you can help me to add the test case you mentioned ?

Just make a file and use e.g. [this](https://github.com/haskell/lsp/blob/master/lsp/test/VspSpec.hs#L245) as an identifier and see what happens.

> I believe you can get the TyThings for all type constuctors using the kindMap field of the GetDocMap rule, so we can avoid doing these lookups multiple times.

Yeah, but then we should generalize the rule a bit. It's weird to be using a "doc map" for this.

Having a single rule that produces a map of summary information about all names in a module does seem reasonable, though.

soulomoon:
I made a change to the compat to expose the function flag, not sure if it would be good idea

soulomoon:
Thank @wz1000 to point this out. Sure I'll adapt the `GetDocMap` for this latter.

soulomoon:
@michaelpj I am having a wierd off match on colors and an LSP: incoming message parse error:
Screenshot 2023-12-30 at 23 21 09


```
Error condition, please check your setup and/or the [issue tracker](https://github.com/haskell/haskell-language-server/issues): 
LSP: incoming message parse error:
Error in $: Failed reading: Cannot decode input: Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream. Expecting object value at '}]}}'
when processing
{"jsonrpc":"2.0","method":"textDocument/didChange","params":{"textDocument":{"uri":"file:///Users/ares/src/haskell-language-server/plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs","version":216},"contentChanges":[{"range":{"start":{"line":107,"character":15},"end":{"line":107,"character":16}},"rangeLength":1,"text":"\ud801"}]}}
```

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: Thanks I will try this out.

soulomoon: done

wz1000:
Changing the name of KindMap seems sensible.

michaelpj:
> having a wierd off match on colors and an LSP: incoming message parse error:

I would expect the positions to be wrong. The parse error is a bit weird, but probably just another sign of everything being generally broken. Maybe in fact not worth including it here for that reason, until/unless we ever sort it out properly.

soulomoon:
> > having a wierd off match on colors and an LSP: incoming message parse error:
> 
> 
> 
> I would expect the positions to be wrong. The parse error is a bit weird, but probably just another sign of everything being generally broken. Maybe in fact not worth including it here for that reason, until/unless we ever sort it out properly.

I guess we need to convert codepoints in Haskell source to range using the convertor in this pull request? https://github.com/haskell/lsp/pull/407
With lsp spec and ghc using different standard of Unicode.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
soulomoon: outputable is too user friendly(Being verbose), I just want a more compact output that fits the source better here for debugging

soulomoon:
1. Ajusted to use codePointRangeToRange, it should handle unicode correctly now, with newly added realSrcSpanToCodePointRange, realSrcLocToCodePointPosition

2. Rename KindMap to TyThingMap

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
```diff
@@ -119,4 +123,18 @@ recoverFunMaskArray flattened = unflattened
     go (HTyConApp _ _)             = False
 
 
+realSrcSpanToCodePointRange :: RealSrcSpan -> CodePointRange
```
michaelpj, line 126: let's put these somewhere more central, like in `hls-plugin-api`. Ideally more people would start to use these!

michaelpj:
I think this may be the first plugin that handles unicode source correctly :tada:

michaelpj (commented):
I would be great if we could transfer some of the reasoning you've written in the PR into code comments. In particular, we've had a bit of discussion about:
1. What token types to use for what kinds of Haskell entity
2. What information we can and can't get from what sources

This seems likely to be useful to future people working on this plugin and will get lost if it's just in PR comments.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
soulomoon: Yep ,these two are meant to be a replacement for the  `realSrcSpanToRange , realSrcLocToPosition`, I think we should put the codePoint version beside these two, and deprecate them(ideally `Range, Position` should be in lsp standard, `CodePointRange, CodePointPosition` should be in ghc standard). 

In the mean time, I think the lsp package could cache the ghc position to lsp position mapping in the virtual file type, making the conversion to be more efficient. Hopefully, every other part of the hls could support unicode eventually.

soulomoon:
> I would be great if we could transfer some of the reasoning you've written in the PR into code comments. In particular, we've had a bit of discussion about:
> 
> 1. What token types to use for what kinds of Haskell entity
> 2. What information we can and can't get from what sources
> 
> This seems likely to be useful to future people working on this plugin and will get lost if it's just in PR comments.

sure, I'll collect them some times later

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: I could help to open an pr to ship `recoverSemanticTokens` helper like function up to `lsp` in the future. 
I guess they should be in the `test` module of `lsp` package?

Bodigrim:
> but since text reading involved, the conversion taken place one by one is not efficient, maybe we can create a position mapping for the whole file at one. And cache that mapping to do the conversion instead.

To avoid performance degradation `text-rope` supports mixed indexing mode in [`Data.Text.Utf16.Rope.Mixed`](https://hackage.haskell.org/package/text-rope-0.2/docs/Data-Text-Utf16-Rope-Mixed.html). It would be great if someone revived https://github.com/haskell/lsp/pull/436 and switched `lsp` to use it.

soulomoon:
> > but since text reading involved, the conversion taken place one by one is not efficient, maybe we can create a position mapping for the whole file at one. And cache that mapping to do the conversion instead.
> 
> To avoid performance degradation `text-rope` supports mixed indexing mode in [`Data.Text.Utf16.Rope.Mixed`](https://hackage.haskell.org/package/text-rope-0.2/docs/Data-Text-Utf16-Rope-Mixed.html). It would be great if someone revived [haskell/lsp#436](https://github.com/haskell/lsp/pull/436) and switched `lsp` to use it.

Sure, I can take a look at them later to see if I could be helpful.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
soulomoon: Indeed, a more friendly maker function for lsp's semantic token is surely helpful.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
soulomoon: I goes through the status, the lsp3.17 protocal does support codePoint, but need to wait for the client support, up to today, vscode still does not support https://github.com/microsoft/vscode-languageserver-node/blob/3f725d8368ec0c75f390a37e4e159d78bc1527ac/client/src/common/client.ts#L1240, :(

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -0,0 +1,224 @@
...
+hieKindFunMasksKind :: HieKind a -> HieFunMaskKind a
+hieKindFunMasksKind hieKind = case hieKind of
+  HieFresh -> HieFreshFun
+  HieFromDisk full_file -> HieFromDiskFun $ recoverFunMaskArray (hie_types full_file)
+
+-- wz1000 offered
+-- the idea from https://gitlab.haskell.org/ghc/haddock/-/blob/b0b0e0366457c9aefebcc94df74e5de4d00e17b7/haddock-api/src/Haddock/Backends/Hyperlinker/Utils.hs#L107
+recoverFunMaskArray
+  :: A.Array TypeIndex HieTypeFlat -- ^ flat types
+  -> A.Array TypeIndex Bool-- ^ full AST
```
michaelpj, line 110: comments generally seem inaccurate for what this does? (looking for which types are functions)

michaelpj:
Could you make this disabled by default, like the stan plugin here? https://github.com/haskell/haskell-language-server/pull/3917

I think given how big and new this is it's sensible to do that for one release and get some testing.

Then I think this is close to being mergeable. I don't know if there's more tidying you want to do - maybe a bit more documentation/cleaning up commented out code etc.

And we should remember to make issues for the bits of future work you have in mind.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: thanx, fixed

soulomoon:
> Could you make this disabled by default, like the stan plugin here? #3917
> 
> I think given how big and new this is it's sensible to do that for one release and get some testing.
> 
> Then I think this is close to being mergeable. I don't know if there's more tidying you want to do - maybe a bit more documentation/cleaning up commented out code etc.
> 
Yep, Plugin is off by default now and tidying is done. 

> And we should remember to make issues for the bits of future work you have in mind.

definitely, I will open an issues to track them down.

michaelpj (approved):

soulomoon:
track list have been made in #3931

PR #3893, opened 2023-12-09
Generate FileTarget for all possible targetLocations

### LINKED ISSUE #3794: HLS fails to find modules with .hs-boot files in the root project directory
Hey,

This is a continuation of a discussion we had with @fendor on IRC in #haskell-language-server.

Long story short, HLS seems to fail to find two self-recursive modules with .hs-boot files if they happen to live in the project root and you have a minimal `hie.yaml`. That is `hs-source-dirs: .`. This with GHC v9.2.8, HLS v2.2.0.0. 

Here's the gist/Gist reproducing this and the error message it results in: https://gist.github.com/moll/591c08e1c6fa75b237214c298e3a97e5. There's also what gen-hie generates and what `hie.yaml` looked like.

I also confirmed that moving the X and Y modules to subdirectory (like `Lib`), while leaving `hs-source-dirs: .`, works.

Cheers

fendor:
If a target file has multiple possible locations, then we
assume they are all separate file targets.
This happens with '.hs-boot' files if they are in the root directory of the project.
GHC reports options such as '-i. A' as 'TargetFile A.hs' instead of 'TargetModule A'.
In 'fromTargetId', we dutifully look for '.hs-boot' files and add them to the
targetLocations of the TargetDetails. Then we add everything to the 'knownTargetsVar'.
However, when we look for a 'Foo.hs-boot' file in 'FindImports.hs', we look for either

 * TargetFile Foo.hs-boot
 * TargetModule Foo

If we don't generate a TargetFile for each potential location, we will only have
'TargetFile Foo.hs' in the 'knownTargetsVar', thus not find 'TargetFile Foo.hs-boot'
and also not find 'TargetModule Foo'.

It took a while, but this fixes #3794

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -41,8 +41,8 @@ import           Data.Hashable                        hiding (hash)
 import qualified Data.HashMap.Strict                  as HM
 import           Data.IORef
 import           Data.List
-import           Data.List.NonEmpty                   (NonEmpty (..))
 import           Data.List.Extra                      as L
+import           Data.List.NonEmpty                   (NonEmpty (..))
 import qualified Data.List.NonEmpty                   as NE
 import qualified Data.Map.Strict                      as Map
 import           Data.Maybe
@@ -113,22 +113,23 @@ import           System.Random                        (RandomGen)
 
 import qualified Development.IDE.Session.Implicit     as GhcIde
 
-import Development.IDE.GHC.Compat.CmdLine
+import           Development.IDE.GHC.Compat.CmdLine
 
 
 -- See Note [Guidelines For Using CPP In GHCIDE Import Statements]
 #if MIN_VERSION_ghc(9,3,0)
 import qualified Data.Set                             as OS
 
-import GHC.Driver.Errors.Types
-import GHC.Driver.Env (hscSetActiveUnitId, hsc_all_home_unit_ids)
-import GHC.Driver.Make (checkHomeUnitsClosed)
-import GHC.Unit.State
-import GHC.Types.Error (errMsgDiagnostic)
-import GHC.Data.Bag
+import           GHC.Data.Bag
+import           GHC.Driver.Env                       (hscSetActiveUnitId,
+                                                       hsc_all_home_unit_ids)
+import           GHC.Driver.Errors.Types
+import           GHC.Driver.Make                      (checkHomeUnitsClosed)
+import           GHC.Types.Error                      (errMsgDiagnostic)
+import           GHC.Unit.State
 #endif
 
-import GHC.ResponseFile
+import           GHC.ResponseFile
 
 data Log
   = LogSettingInitialDynFlags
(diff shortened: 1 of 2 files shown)


michaelpj (approved):
Not sure I entirely understand, but seems plausible!

fendor:
Test failures seem genuine :(

wz1000:
Patch seems fine to me. Did you check locally to see why tests are failing?

fendor:
Not yet, but will look into it!

fendor:
The existence check seems to be removing valid file targets. I.e., there are many file targets that don't exist on disk. Probably related to the VFS? I am confused why it works with Module Targets in that case, since we always filter those.

wz1000:
> The existence check seems to be removing valid file targets. I.e., there are many file targets that don't exist on disk. Probably related to the VFS? I am confused why it works with Module Targets in that case, since we always filter those.

Many files from the testsuite only exist in the VFS, and those files are always `FileTarget`s. I guess we should also check for existence in the VFS.

wz1000:
We could also ensure that explicit file targets aren't filtered out.

fendor:
> We could also ensure that explicit file targets aren't filtered out.

What are explicit file targets?
For now, I've just decided that all file target locations are valid FileTargets.

wz1000:
> For now, I've just decided that all file target locations are valid FileTargets.

This is incorrect. We might end up trying to typecheck non-existent hs-boot files, or files in the wrong include directory

> What are explicit file targets? 

The FilePath argument to `TargetFile :: FilePath -> Target` is the explicit target. The `[FilePath]` in the `targetLocations` still needs to be filtered. We just need to ensure that the explicit target is included in the final result:

```haskell
case targetTarget of
  TargetFile f -> do
    found <- filterM (IO.doesFileExist . fromNormalizedFilePath) targetLocations
    return [(targetTarget, nubOrd (f:found))]
```

fendor:
Ah, I see!
Ok, the code presented is still not correct with respect to files that only exist in the VFS, right?

wz1000:
No, but the problematic situation will only be when a file is listed as a Target but not saved to disk, i.e. the user has edited the cabal file to include the module, but has not saved the module yet.

fendor:
@wz1000 Looks good now?

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -479,12 +480,28 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
         -- files in the project so that `knownFiles` can learn about them and
         -- we can generate a complete module graph
     let extendKnownTargets newTargets = do
-          knownTargets <- forM newTargets $ \TargetDetails{..} ->
+          knownTargets <- concatForM  newTargets $ \TargetDetails{..} ->
```
wz1000, line 483: minor nit: Does this still need to be a concatForM? I think we can just use `forM` now and return a single element.

wz1000 (approved):

ghcide/session-loader/Development/IDE/Session.hs
fendor: For a single `TargetFile`, I am still generating multiple `TargetFile`'s, so I still need it.

 https://github.com/haskell/haskell-language-server/pull/3893/files#diff-d7140bbfca1b7aaafd8c82d1d6aa21a3f941eb27ba8f08e4d9fb13766b21b418R501

wz1000: ah ok then.

Issue #3885, opened 2023-12-02
HLS 2.5.0.0 generates Stan? related error


Release-Candidate:
### Your environment

Which CPU do you use?

- Apple ARM64 M1

Which OS do you use?

- MacOS Sonoma 14.1.2

Which version of GHC do you use and how did you install it?

- 9.6.3 ghcup (ARM64)

How is your project built (alternative: link to the project)?

- Using stack.

Which LSP client (editor/plugin) do you use?

- VS Code + Haskell 

Which version of HLS do you use and how did you install it?

- 2.5.0.0 ghcup (ARM64)

Have you configured HLS in any way (especially: a `hie.yaml` file)?

- Does not work with or without an explicit `hie.yaml` - HLS 2.4.0.0 is working with the same `hie.yaml`.

### Steps to reproduce

- I don't know. Works with a "simple" project with 5 source files in total - 1x app, 2x lib 2x test - but not in two other, but still small, projects.

### Expected behaviour

- No errors are generated. Works when building manually and with HLS 2.4.0.0

### Actual behaviour

Some Haskell source files - not all in the same subdirectory and not all files of a subdirectory - generate a HLS error.

### Debug information

- Error message:

```text
nodeInfo
CallStack (from HasCallStack):
  error, called at src/Relude/Debug.hs:296:11 in rld-1.2.1.0-ae82aff6:Relude.Debug
  error, called at src/Stan/Hie/Compat904.hs:56:24 in stn-0.1.0.2-d29f6af7:Stan.Hie.Compat904
```

I do not see any indication of errors in the "Output" pane of "Haskell" in VS Code.

fendor:
Hi, thank you for your bug report!

Pinging @0rphee as the maintainer of the stan plugin.

fendor:
I can reproduce on the HLS codebase itself.

Release-Candidate:
Thank you very much, that spares me from the work of generating a minimal project to reproduce the error

fendor:
The latest VSCode 2.4.2 should disable the stan plugin by default. Can you give it a try?

Release-Candidate:
Yes, thanks, it is working now.

And as soon as I enable Stan in the extension's config, the error appears.

fendor:
Keeping this open since this is a real issue we have to fix, but we have a work around :)

0rphee:
This is an issue with the stan compatibility layer for =<9.4. 
I'm very busy right now, but next week I will take a look.

[...]

PR #3894, opened 2023-12-11
Update index-state to get latest stan version

0rphee:
Fixes: https://github.com/haskell/haskell-language-server/issues/3885


fendor (approved):
Awesome, thank you for the prompt fix!

michaelpj:
If the previous version was bad, can we add a lower-bound so we know we're going to get the fixed version?

Issue #3897, opened 2023-12-12
Mark `hls-fourmolu-plugin`, `hls-ormolu-plugin` buildable with GHC 9.8


moonshxne:
I noticed that many plugins are marked as `buildable: false` for GHC 9.8 because of a dependency on `ghc-exactprint`. However, having tried to build HLS with nix for GHC 9.8.1, I'm fairly certain that `hls-fourmolu-plugin` and `hls-ormolu-plugin` actually don't depend on exactprint (and the latest `fourmolu` and `ormolu` already support GHC 9.8).

michaelpj:
Great, should be easy to verify, CI will tell us.

PR #3899, opened 2023-12-13
Remove some buildability blockers that aren't needed

michaelpj:
Fixes https://github.com/haskell/haskell-language-server/issues/3897


joyfulmantis (approved):
LGTM

Issue #1552, opened 2021-03-10
Wingman eta reduction gets stymied by pattern matches


isovector:
This doesn't eta reduce:

```haskell
test :: Monad m => () -> a -> m a
test = _
```

But this does:

```haskell
test2 :: Monad m => b -> a -> m a
test2 = _
```

PR #3901, opened 2023-12-14
Don't run hlint on testdata directories

### LINKED ISSUE #3900: CI: Hlint should not lint test data files.
Current Hlint in CI is linting files that should not be lint, and report bug and warning on test data.
most testdata is in `test/testdata` for each sub project.
We can easily turn them off by using:
`hlint ./ --ignore-glob="**test/testdata/*.hs"`

fendor:
closes #3900


fendor:
@soulomoon Does this look good?

soulomoon:
Wired, It is still there,

soulomoon:
weird, it is still there, take a look at https://github.com/haskell/haskell-language-server/actions/runs/7206841824/job/19632462805?pr=3901#step:4:133.

soulomoon:
perhaps it is some kind of HLint bug?

fendor:
Hm, maybe it only accepts the `ignore-glob` param once?

fendor:
No, no idea, can't reproduce locally.

soulomoon:
May be we can bumbup the version as in https://github.com/rwe/actions-hlint-setup
```
name: lint
on:
  pull_request:
  push:
    branches:
      - master
      - 'releases/*'

jobs:
  hlint:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: 'Set up HLint'
      uses: rwe/actions-hlint-setup@v1
      with:
        version: '3.1.6'

    - name: 'Run HLint'
      uses: rwe/actions-hlint-run@v1
      with:
        path: src/
        fail-on: warning
```

fendor:
No, ignore-glob is seemingly only recognised by the cli. But on CI, not even that seems to be true.

fendor:
I did it! It looks like `@actions/exec` does something funky with cmd arguments that contain quotes. I don't really know what's going, but I am suspecting that a shell is acting up, maybe escaping or resolving `*`?

soulomoon:
LGTM, it seems checks for the testdata's files is gone.

michaelpj:
What's this file doing, then? Isn't it exactly what we want? https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-semantic-tokens-plugin/.hlint.yaml

fendor:
Yeah that looks good, I was looking for a different option, apparently. Will update the PR and delete the respective .hlint.yaml files.

michaelpj (approved):

fendor:
Doesn't seem to be working yet

soulomoon:
Since none of us can repro locally, I tend to think there is a Github action CI up stream bug.

fendor:
Adding 
```
- ignore: { "within": '**/testdata/**' }
- ignore: { "within": '**/test/data/**' }
```
does not work for me locally either. That may be a hlint bug worth discussing upstream.

soulomoon:
create a ticket to hlint #1552

soulomoon:
I realize, It might be a conlict between `--git` and `--ignore-glob`  @fendor

soulomoon:
I found the source, this pull request https://github.com/ndmitchell/hlint/issues/1552 should fix the ignore-glob problem.

soulomoon (commented):
looking good, the problem can be fixed with it once the hlint support ignore glob with arg `--git`

michaelpj:
This seems semantically correct even if there's a hlint problem, so let's merge it anyway.

soulomoon (approved):

Issue #3881, opened 2023-11-27
HLS HLint plugin doesn't preserve HLint's severities


dariooddenino:
## Your environment 

Which OS do you use?
Ubuntu 20.04
Which version of GHC do you use and how did you install it?
9.2.8 form ghcup

Which LSP client (editor/plugin) do you use?
helix
Which version of HLS do you use and how did you install it?
2.4.0.0 from ghcup
Have you configured HLS in any way (especially: a `hie.yaml` file)?
I have a `hie.yaml` file, but I don't think it's relevant to this issue.

## What's wrong?
I have added this line to my `.hlint.yaml` file:
```yaml
- error: {name: Avoid restricted module}
```

When running `hlint` I correctly get:
```
src/Domain/EntityEvent.hs:8:1-24: Error: Avoid restricted module
Found:
  import Infrastructure.DB
Note: may break the code

1 error
```

HLS reports the same issue but with `severity: 3`, making it render as a suggestion instead of an error in my editor:
```
{"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"diagnostics":[{"code":"Avoid restricted module","message":"Avoid restricted module\nFound:\n  import Infrastructure.DB ( foo )\nmay break the code\n","range":{"end":{"character":30,"line":7},"start":{"character":0,"line":7}},"severity":3,"source":"hlint"}],"uri":".../src/Domain/EntityEvent.hs","version":1}}
```

Is this something that's not supported (yet)? Is it possible that it's `hlint`'s fault instead?

Thanks

IAmPara0x:
I have created a PR for it: https://github.com/haskell/haskell-language-server/pull/3902. Hope this solves the issue.

michaelpj:
Just as a note for anyone coming here in future: we decided on the following behaviour:
- `error` severities are kept
- all other severities are downgraded

This is because `hlint` pervasively uses "warning" to mean something much less severe than GHC's warnings, but its "error"s are still pretty bad.

PR #3902, opened 2023-12-15
Fix issue:  HLS HLint plugin doesn't preserve HLint's severities #3881

IAmPara0x:
preserve severity from HLint

### DIFF
--- a/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
+++ b/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
@@ -125,8 +125,7 @@ import           Ide.Plugin.Resolve
 import           Ide.PluginUtils
 import           Ide.Types                                          hiding
                                                                     (Config)
-import           Language.Haskell.HLint                             as Hlint hiding
-                                                                             (Error)
+import           Language.Haskell.HLint                             as Hlint
 import qualified Language.LSP.Protocol.Lens                         as LSP
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types                        hiding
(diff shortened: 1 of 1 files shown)


michaelpj (approved):
Code looks good. I think we need to think carefully about whether we definitely want. This means that _most_ hlint hints will turn up as warnings now. Is that too strong? I guess not - that's what hlint says they are, and if a hint seems too strong at that level then arguably that's a hlint bug.

IAmPara0x:
> This means that most hlint hints will turn up as warnings now. Is that too strong? I guess not - that's what hlint says they are, and if a hint seems too strong at that level then arguably that's a hlint bug.

I agree with you here, mainly because if people are using hlint plugin then they are using hlint and hlint gives warnings most of the times than not for the hints. So, if we don't propagate the severity of hlint, it's not an expected behaviour?

michaelpj:
I do think hlint uses "warning" for a bunch of things that don't seem worthy of it to me. I've opened an issue on the hlint repo to discuss: https://github.com/ndmitchell/hlint/issues/1549

fendor (approved):
Code changes look good to me, but I am also unsure whether we want to treat hlint warnings as warnings.

michaelpj:
@IAmPara0x I'm not sure how to progress really. We could demote all of hlint's diagnostics by one level, but that doesn't really address your original annoyance that a hlint error wasn't showing up as an error. And the only real difference it would make is that hlint errors would be warnings instead of infos.

I do think that passing through hlint's severities as they are is likely to seem worse to our users :/

IAmPara0x:
> @IAmPara0x I'm not sure how to progress really. We could demote all of hlint's diagnostics by one level, but that doesn't really address your original annoyance that a hlint error wasn't showing up as an error. And the only real difference it would make is that hlint errors would be warnings instead of infos.
> 
> I do think that passing through hlint's severities as they are is likely to seem worse to our users :/

Yes, I think you are right here. Propagating _default_ errors and warnings would confuse the users. 

Maybe an alternative, solution will be to only propagate severities of only those hints that are user defined?

michaelpj:
I don't think we have that information from hlint though, do we?

michaelpj:
@IAmPara0x it looks like we got a suggestion from Neil that might work: preserve error severity and downgrade the others to info. How does that sound to you?

IAmPara0x:
> @IAmPara0x it looks like we got a suggestion from Neil that might work: preserve error severity and downgrade the others to info. How does that sound to you?

Sounds, good. I will update the PR then

IAmPara0x:
I have updated the PR as we have decided, we now only preserve error severity from Hlint and rest everything is info. Let me know if the PR needs any further updates.

Initially I had an idea of changing it severity levels if they were defined explicitly. Though I am not sure, how much work it will be.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -241,25 +240,36 @@ rules recorder plugin = do
...
+            , _relatedInformation = Nothing
+            , _tags     = Nothing
+            , _codeDescription = Nothing
+            , _data_ = Nothing
+            }
+
+        where
+          codePre = if null $ ideaRefactoring idea then "" else "refact:"
+
+          ideaSeverityToDiagnosticSeverity :: Hlint.Severity -> Maybe LSP.DiagnosticSeverity
```
michaelpj, line 268: Could you explain what we are doing here and link to the issue where we discussed it?

IAmPara0x: Done!. Let me know if the comment is detailed enough

michaelpj: Great, thanks!

PR #3917, opened 2023-12-21
Disable stan plugin by default

### LINKED ISSUE #3916: stan plugin (enabled by default) gives obnoxious suggestions

stan gives a warning on every single lazy field (!) of a data structure that one defines, with the dubious suggestion to add `StrictData` pragmas.

Haskell is a lazy language and the idea that lazy fields cause space leaks is confused. It is embarrassing that the official language server would behave like this out of the box!

Stan suggestions should not pop up or the default options should disable the warning about defining non-strict data structures.

## Additional context

image

fendor:
Closes #3916

### DIFF
--- a/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
+++ b/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
@@ -32,6 +32,7 @@ import           GHC.Generics                   (Generic)
 import           Ide.Plugin.Config
 import           Ide.Types                      (PluginDescriptor (..),
                                                  PluginId, configHasDiagnostics,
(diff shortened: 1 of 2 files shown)


fendor:
Even with this change:
```
    "haskell.plugin.stan.globalOn": {
        "default": true,
        "description": "Enables stan plugin",
        "scope": "resource",
        "type": "boolean"
    }
 ```
 
 What do I have to do to set the default to `false`?

michaelpj:
Yeah, `generate-default-config` seems to show it off by default, so I think you're right.

0rphee (approved):
Seems fine. It works with my editor.

michaelpj (approved):

Issue #3919, opened 2023-12-23
The current positionmappings is composing in the wrong way in `lastValueIO`


soulomoon:
I discover the problem when developing  Semantic tokens

              I observed a bizzared behaviour when using possition mappings
              in vscode.
              in the first 6 space insert, the using possition mapping to ident the code seems to be working,
              Then the column at the start position stop to update
              ```
              Range {_start = Position {_line = 139, _character = 16}, _end = Position {_line = 139, _character = 25}} spanNames_atz7
              Range {_start = Position {_line = 139, _character = 17}, _end = Position {_line = 139, _character = 26}} spanNames_atz7
              Range {_start = Position {_line = 139, _character = 17}, _end = Position {_line = 139, _character = 27}} spanNames_atz7
              ```
              
              <img width="259" alt="Screenshot 2023-12-23 at 13 19 25" src="https://github.com/haskell/haskell-language-server/assets/14073857/3e98cb7f-9f45-4c60-a37a-e499b3cca129">
              
              here is another one
              ```
              Range {_start = Position {_line = 25, _character = 13}, _end = Position {_line = 25, _character = 19}} xxxxx_aTsW
              Range {_start = Position {_line = 25, _character = 13}, _end = Position {_line = 25, _character = 20}} xxxxx_aTsW
              # with
              [TextDocumentContentChangeEvent (InL (#range .== Range {_start = Position {_line = 25, _character = 9}, _end = Position {_line = 25, _character = 9}} .+ #rangeLength .== Just 0 .+ #text .== " "))]
              ```

_Originally posted by @soulomoon in https://github.com/haskell/haskell-language-server/issues/3892#issuecomment-1868204948_
When useWithStale, the old delta is mistakenly be considered as new.
I will open a pull request to fix this.

soulomoon:
Actually the error happens on the 6th insertions.

The behaviour is subtle on just inserting spaces consistantly:
say the each insertion of whitespace at left of the following.
(x, y), (x+1, y), , (x+2, y) ...
the code `let spanNames` start at (x,y), notice `spanNames` start at (x+5). (This is what the stale version is)
the code after 5 insertions is `let spanNames` start at (x+5,y).  Notice `spanNames` moves from the stale version
from `x+5` to `x+10`. Also notice the 6th insertion would happens at left of `x+6`.
Here `x+6 > x+5`, now the `x+5` which is the start of `spanNames` at the stale version.
[...]

PR #3920, opened 2023-12-23
Fix positionMapping in stale data

soulomoon:
Fix https://github.com/haskell/haskell-language-server/issues/3919#issue-2054752246

### DIFF
--- a/ghcide/src/Development/IDE/Core/PositionMapping.hs
+++ b/ghcide/src/Development/IDE/Core/PositionMapping.hs
@@ -10,7 +10,6 @@ module Development.IDE.Core.PositionMapping
   , fromCurrentPosition
   , toCurrentPosition
   , PositionDelta(..)
-  , addDelta
   , addOldDelta
   , idDelta
   , composeDelta
@@ -120,12 +119,12 @@ idDelta = PositionDelta pure pure
 mkDelta :: [TextDocumentContentChangeEvent] -> PositionDelta
 mkDelta cs = foldl' applyChange idDelta cs
 
--- | Add a new delta onto a Mapping k n to make a Mapping (k - 1) n
-addDelta :: PositionDelta -> PositionMapping -> PositionMapping
-addDelta delta (PositionMapping pm) = PositionMapping (composeDelta delta pm)
-
--- | Add a old delta onto a Mapping k n to make a Mapping (k - 1) n
-addOldDelta :: PositionDelta -> PositionMapping -> PositionMapping
+-- | addOldDelta
+-- Add a old delta onto a Mapping k n to make a Mapping (k - 1) n
+addOldDelta ::
+    PositionDelta -- ^ delta from version k - 1 to version k
+    -> PositionMapping -- ^ The input mapping is from version k to version n
+    -> PositionMapping -- ^ The output mapping is from version k - 1 to version n
 addOldDelta delta (PositionMapping pm) = PositionMapping (composeDelta pm delta)
 
 -- TODO: We currently ignore the right hand side (if there is only text), as
--- a/ghcide/src/Development/IDE/Core/PositionMapping.hs
+++ b/ghcide/src/Development/IDE/Core/PositionMapping.hs
@@ -11,6 +11,7 @@ module Development.IDE.Core.PositionMapping
   , toCurrentPosition
   , PositionDelta(..)
   , addDelta
+  , addOldDelta
   , idDelta
   , composeDelta
   , mkDelta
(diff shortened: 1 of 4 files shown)


michaelpj:
I have absolutely no idea how this stuff works. Maybe @wz1000 does?

soulomoon (commented):
the problem lays in `updatePositionMapping`.

soulomoon:
To share the old keys being in acc, we should use `mapAccumWithKey`

soulomoon:
I add tests on this. 
Just run `cabal test --test-option="-p /enumMapMappingTest/" `

1. Not many understand how is the bug a problem, the test case can serve as explaination.
2. The behaviour of the bug is subtle, it can prevent re-introduction. 

I will explain how this is going on,
Suppose you have a text starting from (0,0) `let x = 1 in 1`(version 0, with x in posision (0, 4)).
For i in [1..5], you are inserting a space at (0, i) consecutively.
After the 6th insertion, the text is now `     let x = 1 in 1`(version 6, with x in position (0, 9)).
The bug would case the mapping of (0, 4)(version0) to  (0, 8)(version5) which should be (0, 9)(version5)

It could be easily reproduced by swtich `addOldDelta` back to `addDelta` in `updatePositionMappingHelper` at /Users/ares/src/haskell-language-server/ghcide/src/Development/IDE/Core/Shake.hs


```
enumMapMappingTest :: TestTree
enumMapMappingTest = testCase "enumMapMappingTest" $ do
  let mkChangeEvent :: Range -> Text -> TextDocumentContentChangeEvent
      mkChangeEvent r t = TextDocumentContentChangeEvent $ InL $ #range .== r .+ #rangeLength .== Nothing .+ #text .== t
      mkCE :: UInt -> UInt -> UInt -> UInt -> Text -> TextDocumentContentChangeEvent
      mkCE l1 c1 l2 c2 = mkChangeEvent (Range (Position l1 c1) (Position l2 c2))
      events :: [(Int32, [TextDocumentContentChangeEvent])]
      events = map (second return) [(0, mkCE 0 0 0 0 ""), (1, mkCE 0 1 0 1 " "), (2, mkCE 0 2 0 2 " "), (3, mkCE 0 3 0 3 " "), (4, mkCE 0 4 0 4 " "), (5, mkCE 0 5 0 5 " ")]
      finalMap = Prelude.foldl (\m (i, e) -> updatePositionMappingHelper i e m) mempty events
  let updatePose fromPos = do
        mapping <- snd <$> EM.lookup 0 finalMap
        toCurrentPosition mapping fromPos
  updatePose (Position 0 4) @?= Just (Position 0 9)
  updatePose (Position 0 5) @?= Just (Position 0 10)
```

soulomoon:
hi @michaelpj , I have added a test case, it should be easy to spot the bug now.

ghcide/src/Development/IDE/Core/PositionMapping.hs
```diff
@@ -123,6 +124,10 @@ mkDelta cs = foldl' applyChange idDelta cs
 addDelta :: PositionDelta -> PositionMapping -> PositionMapping
 addDelta delta (PositionMapping pm) = PositionMapping (composeDelta delta pm)
 
+-- | Add a old delta onto a Mapping k n to make a Mapping (k - 1) n
```
wz1000, line 127: Could you expand on the documentation for this as well as `addDelta` clarifying which is appropriate to use in which situation with something like the example you gave in the PR?

ghcide/src/Development/IDE/Core/PositionMapping.hs
```diff
@@ -123,6 +124,10 @@ mkDelta cs = foldl' applyChange idDelta cs
 addDelta :: PositionDelta -> PositionMapping -> PositionMapping
 addDelta delta (PositionMapping pm) = PositionMapping (composeDelta delta pm)
```
wz1000, line 125: Is `addDelta` used anywhere now?

wz1000 (changes requested):
Thanks, great catch. All this is quite subtle!

ghcide/src/Development/IDE/Core/PositionMapping.hs
soulomoon: No, not that I know of now.

soulomoon: Maybe we should delete addDelta?

wz1000: Yes please

wz1000 (approved):

ghcide/src/Development/IDE/Core/PositionMapping.hs
soulomoon: done, the added haddock comment should be clarifying things out.

michaelpj:
Thanks so much for doing the work to add a test!

PR #3928, opened 2024-01-05
Fix multi unit session when some packages have reexported modules.

wz1000:
If we are loading multiple home packages, we need to explicitly take reexports
into account when searching for target files.

If we can't find a module via the usual mean, but it is listed as a reexport of
a unit in scope, we need to look for the module from the perspective of that unit.

This is not necessary for non-home modules because GHC already handles this for
modules in the package DB.

Unfortunately we can't fix this in GHC 9.2 because it doesn't support multiple
home units and we have no way of knowing if a unit reexports modules


ghcide/src/Development/IDE/Import/FindImports.hs
```diff
@@ -143,10 +155,12 @@ locateModule env comp_info exts targetFor modName mbPkgName isSource = do
     Nothing -> do
 #endif
 
-      mbFile <- locateModuleFile ((homeUnitId_ dflags, importPaths dflags) : other_imports) exts targetFor isSource $ unLoc modName
+      mbFile <- locateModuleFile ((homeUnitId_ dflags, importPaths dflags, S.empty) : other_imports) exts targetFor isSource $ unLoc modName
```
fendor, line 158: This needs a comment explaining why `S.empty` is correct here.

fendor (approved):
LGTM, one comment needed.

ghcide/src/Development/IDE/Import/FindImports.hs
wz1000: added

fendor (approved):

PR #3938, opened 2024-01-09
add doc and ci test for semantic tokens

soulomoon:

Issue #3937, opened 2024-01-09
Configurability for semantic tokens


michaelpj:
As discussed in https://github.com/haskell/haskell-language-server/issues/3931, the current semantic tokens implementation makes various fine distinctions that people may not want. This seems like a case where we may genuinely want some configurability. Configuration can also offer some discoverability: people can see what the highlighter is doing by looking at the default configuration.

----

Possible configuration approaches:

1. For each "hs token type" add an option to either give a specific LSP token type or a generic one

For example, we would have an option: "Specific token type for data types? Y/N (default: Y)". If yes, we would do what we currently do, if no we would fall back to the generic token type for that thing, i.e. "type".

This would basically let people opt-out of a particular highlighting distinction.

2. For each "hs token type" add an option for exactly which LSP token type to use

For example, we would have an option: "Token type for data types: type, function, interface,... (default: enum)".

This would effectively make the mapping completely configurable. We would still control the kinds of thing that can be mapped (depending on what kinds of hs token type we recognize), but in principle users could configure it to then do whatever they want.

This would also make the default behaviour maximally discoverable, since the default configuration would tell you specifically what it does for every hs token type.

The downsides are: the configuration is much more verbose and complex; you can set nonsensical things if you like, e.g. "use the 'method' token type for data types".

----

I think I'm mildly leaning towards approach 2, partly because I think the discoverability aspect is quite nice. I don't see another good way for users to find out that we are highlighting things of function type as "functions" and other variables as "variables", without writing it in the docs (which people don't read) and painstakingly keeping it in sync.

WDYT @soulomoon

soulomoon:
@michaelpj I agree with you, approach 2 is better.
It can solve major problems here.
1. Hs token type match to LSP default is always debatable, every one would have its own flavor.
2. Different color scheme might behave vastly different on same settings. This approach 2 open the door for easier fine tuning(without them to change the color scheme).
3. Discoverability aspect as you said.
I'll open a pull request for this.

michaelpj:
Done

PR #3940, opened 2024-01-10
add config  for semantic-tokens-plugin for mapping from hs token type to LSP default token type

soulomoon:
This would provide configurability for semantic tokens #3937
Things that is done:
1. [x] Add `tokenMappings` to config of semantic-token-plugin
2. [x] Add test to see if modification to `tokenMappings` is working
3. [x] Fix some leftover issue in https://github.com/haskell/haskell-language-server/pull/3938
4. [x] ajust this part in the global config test  

default config for semantic tokens:
```
        "semanticTokens": {
            "config": {
                "class": "class",
                "classMethod": "method",
                "dataCon": "enumMember",
                "function": "function",
                "patternSyn": "macro",
                "recField": "property",
                "typeCon": "enum",
                "typeFamily": "interface",
                "typeSyn": "type",
                "typeVariable": "typeParameter",
                "variable": "variable"
            },
            "globalOn": false
        },

```


soulomoon:
Configuration is implemented, @michaelpj @fendor 
See if this look good ?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -60,23 +58,24 @@ import           Language.LSP.Protocol.Types          (NormalizedFilePath,
...
 
-computeSemanticTokens :: IdeState -> NormalizedFilePath -> ExceptT PluginError Action SemanticTokens
-computeSemanticTokens st nfp = do
-  logActionWith st Debug $ "Computing semantic tokens:" <> show nfp
+computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeState -> NormalizedFilePath -> ExceptT PluginError Action SemanticTokens
+computeSemanticTokens recorder pid _ nfp = do
+  logWith recorder Debug (LogMsg "computeSemanticTokens start")
+  config :: SemanticTokensConfig  <- lift $ usePropertyAction #tokenMapping pid semanticConfigProperties
+  logWith recorder Debug (LogMsg $ show $ toJSON config)
+  logWith recorder Debug (LogConfig config)
```
michaelpj, line 71: Surely you don't need both of these?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -60,23 +58,24 @@ import           Language.LSP.Protocol.Types          (NormalizedFilePath,
...
 -----------------------
 ---- the api
 -----------------------
 
-computeSemanticTokens :: IdeState -> NormalizedFilePath -> ExceptT PluginError Action SemanticTokens
-computeSemanticTokens st nfp = do
-  logActionWith st Debug $ "Computing semantic tokens:" <> show nfp
+computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeState -> NormalizedFilePath -> ExceptT PluginError Action SemanticTokens
+computeSemanticTokens recorder pid _ nfp = do
+  logWith recorder Debug (LogMsg "computeSemanticTokens start")
```
michaelpj, line 68: I think this is probably not useful usually. Or we should be logging such things generically...

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,24 @@
...
+            "recField": "property",
+            "typeCon": "enum",
+            "typeFamily": "interface",
+            "typeSyn": "type",
+            "typeVariable": "typeParameter",
+            "variable": "variable"
+        },
+        "markdownDescription": "Configuration of the map from hs semantic token type to LSP default token type",
+        "scope": "resource",
+        "type": "object"
```
michaelpj, line 268: This isn't very discoverable, I think it's not going to be clear to people how to edit this in the UI. Could we instead have an option per mapping? Then we could have an enum type for the token types as the option type, too?

michaelpj: So more like 
```
haskell.plugin.semanticTokens.config.tokenMapping.class": {
        "default": "class",
        "markdownDescription": "LSP semantic token type to use for typeclass names",
        "type": "string",
        "enum": [ "class", "interface", ... ]
        ...
```

Then a) we get a specific description (so we can say exactly what the token type corresponds to, "class" is suggestive but better to be explicit), and people will get a dropdown menu of options for the token type.

I don't know how easy this is with our config mechanism, though.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -32,30 +34,22 @@ import           Language.LSP.VFS                hiding (line)
...
+  TTypeVariable -> runIdentity $ stTypeVariable conf
+  TDataCon      -> runIdentity $ stDataCon conf
+  TClass        -> runIdentity $ stClass conf
+  TTypeCon      -> runIdentity $ stTypeCon conf
+  TTypeSyn      -> runIdentity $ stTypeSyn conf
+  TTypeFamily   -> runIdentity $ stTypeFamily conf
+  TRecField     -> runIdentity $ stRecField conf
+  TPatternSyn   -> runIdentity $ stPatternSyn conf
 
 lspTokenReverseMap :: Map.Map SemanticTokenTypes HsSemanticTokenType
```
michaelpj, line 51: seems like this might need the config too? Or I guess this is only used in testing? I'd be inclined to make it explicit anyway

plugins/hls-semantic-tokens-plugin/hls-semantic-tokens-plugin.cabal
```diff
@@ -52,12 +52,15 @@ library
     , array
     , deepseq
     , hls-graph == 2.5.0.0
+    , data-default
+    , rank2classes
```
michaelpj, line 56: we don't currently depend on this so it would be mildly better to avoid it if we can. It still looks actively developed, but you never know, and often small stuff like this lags with GHC version upgrades etc.

michaelpj (commented):
The main thing is I think it would be better with one option per mapping, although I'm not sure how to do that. Maybe @fendor knows having looked at the config stuff recently?

michaelpj:
https://github.com/haskell/haskell-language-server/blob/034b33ebd522e1d6826a2e7ba3df34b59b69180c/ghcide/src/Development/IDE/Plugin/TypeLenses.hs#L107

Seems to have an example of enum properties. Looks like each one will need its own descriptions of the enum members, but you can share that on the Haskell side.

plugins/hls-semantic-tokens-plugin/hls-semantic-tokens-plugin.cabal
soulomoon: It just provide a rank2 liftA2. We can customize one of these.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
michaelpj: Would it be better to take the config that you're using and error in that case? Since otherwise you could end up with misleading test output when you have customized the config?

plugins/hls-semantic-tokens-plugin/test/Main.hs
```diff
@@ -141,14 +121,49 @@ semanticTokensValuePatternTests =
...
+                    (mkSemanticConfig funcVar)
+                    def {ignoreConfigurationRequests = False}
+                    fullCaps
+                    fs $ do
+                    -- modifySemantic funcVar
+                    void waitForBuildQueue
+                    doc <- openDoc "Hello.hs" "haskell"
+                    void waitForBuildQueue
+                    result1 <- docLspSemanticTokensString doc
+                    liftIO $ unlines (map show result1) @?= "2:1-3 SemanticTokenTypes_Variable \"go\"\n"
```
michaelpj, line 159: so here for example if you did the backwards mapping you'd presumably get the wrong thing?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: I consider it too, it won't be a confusion, because I have now parametrized the test output, for result with customized configuration, it won't convert the result back to hs semantic, but show the lap semantic token instead.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: The ToJason would be deleted after we finalize the choice of configuration.

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
soulomoon: Customization might be needed, I'll take a look at the relevant code.

plugins/hls-semantic-tokens-plugin/test/Main.hs
soulomoon: Yes, we should not do backward after customize the config. Maybe we should make a note somewhere.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: `error in that case` does seems to be good though.

soulomoon: done

soulomoon:
> https://github.com/haskell/haskell-language-server/blob/034b33ebd522e1d6826a2e7ba3df34b59b69180c/ghcide/src/Development/IDE/Plugin/TypeLenses.hs#L107
> 
> 
> 
> Seems to have an example of enum properties. Looks like each one will need its own descriptions of the enum members, but you can share that on the Haskell side.

thanks for the link. Switching to this method.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: done

plugins/hls-semantic-tokens-plugin/hls-semantic-tokens-plugin.cabal
soulomoon: removed to in favor of template haskell

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: removed

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
         "scope": "resource",
         "type": "boolean"
     },
+    "haskell.plugin.semanticTokens.config.class": {
+        "default": "class",
+        "description": "LSP semantic token type to use for Class",
```
michaelpj, line 254:
```suggestion
        "description": "LSP semantic token type to use for typeclasses",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "modifier",
+            "comment",
+            "string",
+            "number",
+            "regexp",
+            "operator",
+            "decorator"
+        ],
+        "enumDescriptions": [
+            "LSP Semantic Token Type:namespace",
```
michaelpj, line 281: these are kind of pointless actually, aren't they? maybe just skip them since they don't actually say anything new

soulomoon:
See if this looks good, I use Enum property for each token type? @michaelpj

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.classMethod": {
+        "default": "method",
+        "description": "LSP semantic token type to use for ClassMethod",
```
michaelpj, line 310:
```suggestion
        "description": "LSP semantic token type to use for typeclass methods",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.dataCon": {
+        "default": "enumMember",
+        "description": "LSP semantic token type to use for DataCon",
```
michaelpj, line 366:
```suggestion
        "description": "LSP semantic token type to use for data constructors",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:string",
+            "LSP Semantic Token Type:number",
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.dataCon": {
```
michaelpj, line 364:
```suggestion
    "haskell.plugin.semanticTokens.config.dataConstructor": {
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.function": {
+        "default": "function",
+        "description": "LSP semantic token type to use for Function",
```
michaelpj, line 422:
```suggestion
        "description": "LSP semantic token type to use for functions",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.patternSyn": {
+        "default": "macro",
+        "description": "LSP semantic token type to use for PatternSyn",
```
michaelpj, line 478:
```suggestion
        "description": "LSP semantic token type to use for pattern synonyms",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:string",
+            "LSP Semantic Token Type:number",
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.patternSyn": {
```
michaelpj, line 476:
```suggestion
    "haskell.plugin.semanticTokens.config.patternSynonym": {
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:string",
+            "LSP Semantic Token Type:number",
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.recField": {
```
michaelpj, line 532:
```suggestion
    "haskell.plugin.semanticTokens.config.recordField": {
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:string",
+            "LSP Semantic Token Type:number",
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.typeCon": {
```
michaelpj, line 588:
```suggestion
    "haskell.plugin.semanticTokens.config.typeConstructor": {
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.typeVariable": {
+        "default": "typeParameter",
+        "description": "LSP semantic token type to use for TypeVariable",
```
michaelpj, line 758:
```suggestion
        "description": "LSP semantic token type to use for type variables",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
...
+            "LSP Semantic Token Type:regexp",
+            "LSP Semantic Token Type:operator",
+            "LSP Semantic Token Type:decorator"
+        ],
+        "scope": "resource",
+        "type": "string"
+    },
+    "haskell.plugin.semanticTokens.config.variable": {
+        "default": "variable",
+        "description": "LSP semantic token type to use for Variable",
```
michaelpj, line 814:
```suggestion
        "description": "LSP semantic token type to use for variables",
```

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
```diff
@@ -249,6 +249,622 @@
         "scope": "resource",
         "type": "boolean"
     },
+    "haskell.plugin.semanticTokens.config.class": {
```
michaelpj, line 252: I don't know whether we might want more configuration options later, but perhaps it would be wise to nest all these in `config.tokenMapping` or something?

michaelpj (commented):
Just some comments about the naming and stuff. I think this does matter since it's user facing so it's worth making sure it's comprehensible to people

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/SemanticConfig.hs
```diff
@@ -0,0 +1,95 @@
+{-# LANGUAGE GADTs             #-}
+{-# LANGUAGE OverloadedLabels  #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TypeFamilies      #-}
+{-# LANGUAGE TypeOperators     #-}
+
+module Ide.Plugin.SemanticTokens.SemanticConfig where
```
michaelpj, line 8: Is this less code than defining them manually? ;) It's only 11 items. Up to you though :shrug:

michaelpj:
Yeah, the general output looks great

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
soulomoon: Yes, but we should leave it here better than the uggly list of empty string.

michaelpj: Oh, I didn't realise they were mandatory. Then I guess its fine

test/testdata/schema/ghc92/vscode-extension-schema.golden.json
michaelpj: I guess this is the main remaining question. Not sure if it's going to make that much of a difference or not :thinking: 

I guess the alternative would be to suffix the option name with `Token`? So we could have `config.classToken` which is relatively clear what it's doing...

soulomoon: > I don't know whether we might want more configuration options later, but perhaps it would be wise to nest all these in config.tokenMapping or something?

I think this one would be better. I am going to extend the properties api to better support nested configuration.

fendor:
> Maybe @fendor knows having looked at the config stuff recently?

Sorry, I don't have any insight in the config specification itself

michaelpj:
I think in the short term just suffix-ing the options is fine, e.g. `classToken`.

soulomoon:
Sure, let's fix the name and try to merge this one first. 
I can come back to open another pull request for the nested solution once things in #3944 is done.

michaelpj:
Yes, although generally it's better to avoid changing config names since it messes up users

config.vscode
```diff
@@ -0,0 +1,886 @@
+{
```
michaelpj, line 1: shouldn't be here?

generate.bash
```diff
@@ -0,0 +1,4 @@
+ghcup set ghc 9.2.8; cabal clean; cabal test --test-option="-p /generate schema/"
```
michaelpj, line 1: shouldn't be here?

michaelpj (commented):
Nearly there, just a couple of things that shouldn't have been checked in, then I think we can merge.

generate.bash
soulomoon: sorry for get to stash it before commit

michaelpj:
Has conflicts now

michaelpj:
Test failures seem genuine?

soulomoon:
🤔 fail again I'll take a look

soulomoon:
The merge from master create some problems, need manually fixing.

soulomoon:
I should be fixed now

michaelpj (approved):

Issue #3939, opened 2024-01-10
Release for 9.6.4


michaelpj:
This is a release-tracking ticket.

fendor:
hls-cabal-plugin needs to honour `globalOn` configuration. Currently, diagnostics are produced, even if the plugin is turned off.

wz1000:
I have a fix for https://github.com/haskell/haskell-language-server/issues/3542 failures with 9.8 in hiedb.

soulomoon:
> It would be nice to get #3937 done but not a blocker, depends on when @soulomoon gets to it.

Sure, I've open a pull request for this in https://github.com/haskell/haskell-language-server/pull/3940.

benz0li:
Building of HLS 2.5.0.0 currently fails using GHC 9.6.4:

* Aarch64: https://gitlab.b-data.ch/ghc/ghc-musl/-/jobs/92056
* x86_64: https://gitlab.b-data.ch/ghc/ghc-musl/-/jobs/92057

---

### My environment

Which OS do you use? Alpine Linux 3.19.0

Which version of GHC do you use and how did you install it? 9.6.4  
👉 Docker image `glcr.b-data.ch/ghc/ghc-musl:9.6.4` ([https://github.com/benz0li/ghc-musl](https://github.com/benz0li/ghc-musl/blob/ba4b87499b3dd5cea81cdfbbb755a068f41a8777/prior/9.6.4.Dockerfile))

### Steps to reproduce

**On a host with docker installed**

```bash
docker run --rm -ti glcr.b-data.ch/ghc/ghc-musl:9.6.4 bash
```

**Inside the container**

```
export HLS_VERSION=2.5.0.0 \
  && apk add --no-cache patchelf findutils \
  && cd /tmp \
  && curl -sSL "https://github.com/haskell/haskell-language-server/archive/refs/tags/$HLS_VERSION.tar.gz" \
    -o "haskell-language-server-$HLS_VERSION.tar.gz" \
  && tar -xzf "haskell-language-server-$HLS_VERSION.tar.gz" \
  && cd "haskell-language-server-$HLS_VERSION" \
  && . .github/scripts/env.sh \
  && . .github/scripts/common.sh \
  && sed -i.bak -e '/DELETE MARKER FOR CI/,/END DELETE/d' cabal.project \
  && GHCS="$GHC_VERSION" \
  && export GHCS \
  && ARTIFACT="$(uname -m)-linux-alpine" \
[...]

PR #3941, opened 2024-01-10
Don't produce diagnostics if plugin is turned off

fendor:
Relevant for #3939 

(I didn't test this change, but it is identical to https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs#L82)

### DIFF
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Types.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Types.hs
@@ -37,14 +37,14 @@ instance Pretty Log where
     LogUseWithStaleFastNoResult -> "Package description couldn't be read"
     LogMapLookUpOfKnownKeyFailed key -> "Lookup of key in map failed even though it should exist" <+> pretty key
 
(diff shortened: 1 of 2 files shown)


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Types.hs
```diff
@@ -39,7 +39,7 @@ instance Pretty Log where
 
 type instance RuleResult ParseCabal = Parse.GenericPackageDescription
 
-data ParseCabal = ParseCabal
+data ParseCabal = GetCabalDiagnostics
```
michaelpj, line 42: rename the type constructor too? Also the old name was somewhat plausible, right, since it does produce the GPD. So it's a bit like Typecheck: produces the typechecked module and diagnostics. Unsure.

fendor: Good call, I had too much trust in the rename feature to also read my mind :see_no_evil:

PR #3947, opened 2024-01-13
Properties API: Remove unsafe coerce in favor of type class based method in 

soulomoon:
After removal of unsafe coerce the api would not be affected


soulomoon:
The redundant-constraints can also be removed.

michaelpj (approved):
Looks plausible, although I haven't worked it all through in my head. Definitely worth considering if this stuff can be simplified.

Note that I think it would be pretty reasonable for us to actually depend on `singletons` - the base library is pretty small, and it has a lot of vocabulary for this kind of thing.

soulomoon:
> Looks plausible, although I haven't worked it all through in my head. Definitely worth considering if this stuff can be simplified.
> 
> Note that I think it would be pretty reasonable for us to actually depend on `singletons` - the base library is pretty small, and it has a lot of vocabulary for this kind of thing.

Yes, it is for the singleton pattern. And also we can take a look at Oleg Kiselyov's HList for for type level list and record. But it it would be complete rewrite of current implementations, I think we can considered them in the future.

Issue #3950, opened 2024-01-13
Semantic token omitting record field in `{-# LANGUAGE DuplicateRecordFields #-}`


soulomoon:
`recordField` name in instance decl sometimes considered as generated OccName by hieAst even if it is written out by user.  Like in `$sel:plcGlobalOn:PluginConfig`. in hls code base:
```{haskell}
instance Default PluginConfig where
  def = PluginConfig
      { plcGlobalOn         = True
      , plcCallHierarchyOn  = True
      , plcCodeActionsOn    = True
      , plcCodeLensOn       = True
      , plcDiagnosticsOn    = True
      , plcHoverOn          = True
      , plcSymbolsOn        = True
      , plcCompletionOn     = True
      , plcRenameOn         = True
      , plcSelectionRangeOn = True
      , plcFoldingRangeOn   = True
      , plcSemanticTokensOn = True
      , plcConfig           = mempty
      }
```

Stop omit generated OccName could fix this.

soulomoon:
But I do not really understand the cause, could not repro in single file. and hence could not write out test for it.

soulomoon:
I am assuming it is generated back from the hie disk file?

PR #3951, opened 2024-01-13
fix: semantic token omitting record field in `{-# LANGUAGE DuplicateRecordFields  #-}` #3950

soulomoon:
fix omitting record field and {-# LANGUAGE DuplicateRecordFields #-}
and add test case for it


michaelpj (approved):
The test failed before the change, I assume?

soulomoon:
Yes, the I add a test that would fail if no change to current source code.

PR #3952, opened 2024-01-13
3944 extend the properties api to better support nested configuration

### LINKED ISSUE #3944: Extend the Properties api to better support nested configuration
> I don't know whether we might want more configuration options later, but perhaps it would be wise to nest all these in config.tokenMapping or something?

I think this one would be better. I am going to extend the properties api to better support nested configuration.

_Originally posted by @soulomoon in https://github.com/haskell/haskell-language-server/pull/3940#discussion_r1451019549_
            
Since the plugin api's support for nested configuration is poor. It would be better to extend it to enable a type safe configuration for nested configuration. 

### Steps
Towards this goal there are two steps of implementation.
1. The current implementation use (unsafecoerce), and the safty is provided by hand obeying the rules in property map.
Which is not really safe. We replace it with safer implementation. doing in #3947 
2. Extend the configuration api under the above safer implementation to provide nicer nested configuration. doing in #3952

It should support #3937 better

soulomoon:
The implementation closely aligns with the original design, extensively incorporating existing code to minimize workload costs. The new API maintains a consistent style with the old API, which remains unchanged.
Features
With new expose stuff:
`KeyNamePath` -- path to search for properties
`definePropertiesProperty` -- define nested property
`usePropertyByPath` -- extract property by path
`usePropertyByPathEither` -- same as above
`usePropertyByPathAction` -- action api for `usePropertyByPath`
`HasPropertyByPath` -- constraint for using `usePropertyByPath` like the `HasProperty`

We can now define properties upon properties to create nested one. And use KeyNamePath to retrieve the property
```
    nestedPropertiesExample = emptyProperties
        & definePropertiesProperty #parent "parent" (emptyProperties & defineStringProperty #foo "foo" "foo")
        & defineStringProperty #baz "baz" "baz"

    nestedPropertiesExample2 = emptyProperties
        & definePropertiesProperty #parent "parent" (emptyProperties & defineStringProperty #foo "foo" "xxx")
        & defineStringProperty #baz "baz" "baz"

    examplePath1 = SingleKey #baz
    examplePath2 = ConsKeysPath #parent (SingleKey #foo)
```
To retrieve we can have 
```
usePropertyByPathEither examplePath2 nestedPropertiesExample object
```
 

Todo list:
* [x] add key name path
* [x] add usePropertyByPathEither, usePropertyByPath to cope with usage of key name path
* [x] add more test
* [x] add action to use the above

This should close #3944, old api would remains the same.


soulomoon:
I have complete the implementation of extending the Properties api to support nested configuration in a type safe way.
The api style making a lot of use of the original code which simplifies the implementation and follow the original design closely.

fendor (changes requested):
First, sorry for taking forever to get this reviewed! It is rather difficult code to review :) 
Second, thanks for taking care of the Properties module!

I mostly have some small nitpicks, especially regarding the formatting. Could you make sure to add a newline between instance declarations and normalise the formatting a little bit? E.g. `,HasProperty` vs ` , ParsePropertyPath`.

Also, a little bit more documentation in the code would be nice, to explain what problem `TProperties` is supposed to solve.

Otherwise, looks good to go to me!

soulomoon:
Thanks for the review @fendor. This indeed is hard to review with all the type level fiddling. I've adjusted the test and add a comment, see if it is better?

fendor (approved):
LGTM, thanks!

PR #3953, opened 2024-01-14
Remove unused dependencies in hls-refactor-plugin

jhrcek:
I'd like to do general cleanup of unused dependencies (stuff reported by `-Wunused-packages`).

This PR starts by just doing this for one plugin and checking that people agree with the approach to removing unused ghc  (direct - I know that most plugins actually do depend on GHC transitively) dependency.

If you find such initiative useful, I can continue with other plugins.

### DIFF
--- a/plugins/hls-refactor-plugin/test/Main.hs
+++ b/plugins/hls-refactor-plugin/test/Main.hs
@@ -2388,7 +2388,7 @@ addTypeAnnotationsToLiteralsTest = testGroup "add type annotations to literals t
                , ""
                , "f = 1"
                ])
-#if MIN_VERSION_ghc(9,4,0)
+#if MIN_VERSION_GLASGOW_HASKELL(9,4,0,0)
     [ (DiagnosticSeverity_Warning, (3, 4), "Defaulting the type variable") ]
 #else
     [ (DiagnosticSeverity_Warning, (3, 4), "Defaulting the following constraint") ]
@@ -2409,7 +2409,7 @@ addTypeAnnotationsToLiteralsTest = testGroup "add type annotations to literals t
                , "    let x = 3"
                , "    in x"
                ])
-#if MIN_VERSION_ghc(9,4,0)
+#if MIN_VERSION_GLASGOW_HASKELL(9,4,0,0)
     [ (DiagnosticSeverity_Warning, (4, 12), "Defaulting the type variable") ]
 #else
     [ (DiagnosticSeverity_Warning, (4, 12), "Defaulting the following constraint") ]
@@ -2431,7 +2431,7 @@ addTypeAnnotationsToLiteralsTest = testGroup "add type annotations to literals t
                , "    let x = let y = 5 in y"
                , "    in x"
                ])
-#if MIN_VERSION_ghc(9,4,0)
+#if MIN_VERSION_GLASGOW_HASKELL(9,4,0,0)
     [ (DiagnosticSeverity_Warning, (4, 20), "Defaulting the type variable") ]
 #else
     [ (DiagnosticSeverity_Warning, (4, 20), "Defaulting the following constraint") ]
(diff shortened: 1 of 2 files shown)


plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -2388,7 +2388,7 @@ addTypeAnnotationsToLiteralsTest = testGroup "add type annotations to literals t
                , ""
                , "f = 1"
                ])
-#if MIN_VERSION_ghc(9,4,0)
```
jhrcek, line 2391: Do you see any issue with this approach?
Instead of using MIN_VERSION_{package} macros defined by cabal, I switched to using [MIN_VERSION_GLASSWOG_HASKELL](https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html?highlight=min_version_glasgow_haskell#standard-cpp-macros) macro defined by GHC.
This allows us to remove unused (reported by -Wunused-packages) ghc dependency from test suite.

soulomoon:
LGTM

michaelpj:
This is very useful. I definitely think we could turn on a lot more warnings generally. I think it's probably worth putting `-Wunused-packages` into the GHC options for a package when you do this - it means hopefully we stay in the good state afterwards!

plugins/hls-refactor-plugin/test/Main.hs
michaelpj: I _think_ this should be fine, wdyt @wz1000 ?

michaelpj (approved):

plugins/hls-refactor-plugin/test/Main.hs
jhrcek: To be clear, multiple approaches are used across the codebase. 
Clearly the advantage of the first 2 mentioned below is that they don't require unused dependency on GHC.
But there might be issues with them that I'm not aware of.

Currently on master:

The form #if __GLASGOW_HASKELL__ >= XYZ is used 11x
The form `#if MIN_VERSION_GLASGOW_HASKELL(W,X,Y,Z)` is used 4x
The form `#if MIN_VERSION_ghc(X,Y,Z)` is used 354x

jhrcek:
> I think it's probably worth putting -Wunused-packages into the GHC options for a package when you do this - it means hopefully we stay in the good state afterwards!

Can do. What would be your general preference for doing that (so we have some consistency) across packages.
1. Adding -Wunused-packages to each cabal target (library, tests) separately? Or
2. Adding common common-warnings 
                     ghc-options: -Wunused-packages
   and then importing for each target?

plugins/hls-refactor-plugin/test/Main.hs
michaelpj: So I guess one way of looking at it is that that the first two condition on the version of GHC-the-compiler, whereas the latter conditions of the version of ghc-the-library. Arguably it's slightly more semantically correct to give conditionals based on ghc-the-library in most cases, since it's differences between versions of that library that cause problems. 

In this case it is _also_ true that the difference is arguably caused by a difference in (our transitive dependency) ghc-the-library, but we can't _say_ that without depending on it directly. In that sense using one of the first two forms is a bit of a hack that relies on the fact that the version of ghc-the-library matches the version of GHC-the-compiler. I have no idea if those can ever come apart, but I guess it's semantically a bit weird?

michaelpj: Also! I'm not sure we even need the CPP! We have a value `ghcVersion` that is defined (with CPP) to give us the GHC version. So we can just use that here: `if ghcVersion >= GHC94 then ... else ...`.

michaelpj:
> Can do. What would be your general preference for doing that (so we have some consistency) across packages.

Our plugins are pretty inconsistent in general, since they were often written by different people. Adding some consistency would be _lovely_. I personally like common stanzas for stuff like `default-language` and `ghc-options` controlling warnings, so it would be great to use those more consistent.y

plugins/hls-refactor-plugin/test/Main.hs
jhrcek: Ok, in this particular case I switched to ghcVersion as I think no-CPP solution is generally preferable if it can be made to work.

jhrcek:
I'll work on making these more consistent across multiple plugins, but it's too easy to get sidetracked trying to fix too many things at once. Let me try to focus on fixing as many unused-packages as possible and we can agree on how to enforce it consistently at later stage.

fendor (approved):

Issue #3957, opened 2024-01-15
support module name and split out module name in qualified name


soulomoon:
Mainly for `quolified names`.
Semantic highlighting looks worse than syntax highlighting for quolified name.

1. A Syntax highlighting easily support splitting the module part and the name part of a quolified name. 
2. Semantic token is not identifing them, since the `range` of `Name` for quolified name include the module part and the name part as a whole. 

Two steps to solve the problems.
1. First add suppport for module names in the import section, it is easy since there is module name as identifier in hieAst #3958
2. split out the module part and name part, handles ```(Prelude.+), `Data.List.elem` ``` #3958

soulomoon:
Step 2 might requires peek into the source code, which breaks the abstraction of `LSP` and reach into the `LSP`'s package dependency `Text.Rope`, I am not quite sure if it is a good idea to implement Rope related function here. Excepcially when https://github.com/haskell/lsp/pull/542(unrelease) get into release and HLS picks it up, Rope related function implemented in HLS would surely break.

michaelpj:
> Step 2 requires peeking into the source code

Hmm. Why do we need to go into the source code? Don't we have the `OccName` which I think includes the full name text? Then we could just look at that, count how many characters until the last `.`, and split the `Range` at that point? Will need a bit of care about what "character" means, but I don't think we necessarily need the whole file source.

> for (Prelude.+), it is original highlight (Prelude.+) as a whole, now we need to split them. 

Wait, we currently highlight the parens? That's surprising! Is that because GHC gives that as part of the span?

soulomoon:
> Wait, we currently highlight the parens? That's surprising! Is that because GHC gives that as part of the span?
Yes, Ast gives span of `(Prelude.+) `, and OccName give only `+`.

This is the same reason we need to peek into the source code.

michaelpj:
That seems... surprising to me. The _span_ includes all the other text, but the `OccName` doesn't? how odd. Is that expected @wz1000 ?

soulomoon:
You can see that from the renamed source as well.

PR #3958, opened 2024-01-15
Semantic tokens: add module name support and improve performance and accuracy by traversing the hieAst along with source code

soulomoon:
fix https://github.com/haskell/haskell-language-server/issues/3957

Things have been done:
1. Switch `Name` to `Identifier` in the implementation and add `ModuleName` to the `HsSemanticTokenType`
2. Strip ``` ` ` ``` and `()`, and split out qualified names. e.g.``` `Preclude.length` ``` to ```Preclude.```  `length`
3. add tokenizer to walk ast with the souce rope to get more accurate result and faster. Should fix https://github.com/haskell/haskell-language-server/issues/3983.
4. add type sig to semanticConfig's TH result


michaelpj:
This looks like it's still a work in progress?

soulomoon:
> This looks like it's still a work in progress?
Yes, I decicded to to the module name split in here a well.

michaelpj:
Looks like we could also do `module <name> where` but maybe that's not in the HIE ast?

michaelpj:
Okay, so the parentheses are only included in the span if it's an operator section. That seems fine. The question is whether we _should_ highlight the parens when it's an operator section.

soulomoon:
Before we support overlapping tokens.

The following might be a good idea:
Drop the "``, ()" for now, as suggested by @konn https://github.com/haskell/haskell-language-server/issues/3969#issuecomment-1894718215.
We do not need to consider the special split involed "``, ()" for now.

michaelpj:
But if we drop the parens for operator sections around qualified names, we should do it for normal operator sections like `(+)` too for consistency

soulomoon:
Ready to merge.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -5,12 +5,15 @@
...
 {-# LANGUAGE OverloadedLabels      #-}
+{-# LANGUAGE OverloadedRecordDot   #-}
 {-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PolyKinds             #-}
 {-# LANGUAGE RecordWildCards       #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE UnicodeSyntax         #-}
+{-# OPTIONS_GHC -Wno-missing-signatures #-}
```
michaelpj, line 16: do you want this? better to add the signatures maybe?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -62,39 +70,43 @@ nameNameSemanticFromHie hieKind rm ns = do
...
       where
-        xs = mkNameSet $ rights $ M.keys $ M.filterWithKey inclusion bd
+        xs = S.fromList $ M.keys $ M.filterWithKey inclusion bd
         inclusion :: Identifier -> IdentifierDetails a -> Bool
         inclusion a b = not $ exclusion a b
         exclusion :: Identifier -> IdentifierDetails a -> Bool
         exclusion idt IdentifierDetails {identInfo = infos} = case idt of
-          Left _  -> True
+          Left _  -> False
           Right _ -> any isEvidenceContext (S.toList infos)
```
michaelpj, line 99: These cases could do with a comment explaining why we're excluding or not

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/SemanticConfig.hs
```diff
@@ -34,6 +34,7 @@ docName tt = case tt of
     TTypeSynonym     -> "type synonyms"
     TTypeFamily      -> "type families"
     TRecordField     -> "record fields"
+    TModuleName      -> "module names"
```
michaelpj, line 37:
```suggestion
    TModuleName      -> "modules"
```
I think this is more consistent with the others

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Types.hs
```diff
@@ -43,6 +44,7 @@ data HsSemanticTokenType
   | TTypeSynonym -- Type synonym
   | TTypeFamily -- type family
   | TRecordField -- from match bind
+  | TModuleName -- from match bind
```
michaelpj, line 47: comment is wrong
```suggestion
  | TModule -- from match bind
```

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
```diff
@@ -99,3 +101,63 @@ showSpan x = show (srcSpanStartLine x) <> ":" <> show (srcSpanStartCol x) <> "-"
...
+
+-- nameLength is in code points unit.
+-- while Range might not in code points unit.
+-- but we can still use it to get the length
+-- since it is only used to exclude some names
+-- currently, we only break `(ModuleA.b)` into `(ModuleA` and `.b)`
+splitModuleNameAndOccName :: VirtualFile -> Range -> Identifier -> [(Range,Identifier)]
+splitModuleNameAndOccName _ ran (Left m) = [(ran, Left m)]
+splitModuleNameAndOccName vf ran@(Range (Position startLine startColumn) (Position _endLine endColumn)) (Right name)
+    | nameLength name < fromIntegral (endColumn - startColumn), (Just text) <- getTextByCodePointRangeFromVfs vf ran =
```
michaelpj, line 114: I'm still confused about why we can't use the text from the `OccName`? In particular, doesn't `occNameString . nameOccName` give you what you want? You even use that below!

michaelpj: What I would kind of expect is something like:
- Check if the text is delimted by parens or backticks, remember this
- Match the name text against a regex to pull out the module prefix if there is one
- Compute the ranges based on the previous things:
   - If delimited, start and end one character earlier
   - If there is a module prefix, add a token for that bit, and start the operator bit later

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
```diff
@@ -99,3 +101,63 @@ showSpan x = show (srcSpanStartLine x) <> ":" <> show (srcSpanStartCol x) <> "-"
...
 mkRange startLine startCol len =
     Range (Position (fromIntegral startLine) (fromIntegral startCol)) (Position (fromIntegral startLine) (fromIntegral $ startCol + len))
+
+
+-- nameLength is in code points unit.
+-- while Range might not in code points unit.
+-- but we can still use it to get the length
+-- since it is only used to exclude some names
+-- currently, we only break `(ModuleA.b)` into `(ModuleA` and `.b)`
+splitModuleNameAndOccName :: VirtualFile -> Range -> Identifier -> [(Range,Identifier)]
```
michaelpj, line 111: This could do with a longer comment explaining what it's doing.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
```diff
@@ -99,3 +101,63 @@ showSpan x = show (srcSpanStartLine x) <> ":" <> show (srcSpanStartCol x) <> "-"
 mkRange :: (Integral a1, Integral a2) => a1 -> a2 -> a2 -> Range
 mkRange startLine startCol len =
     Range (Position (fromIntegral startLine) (fromIntegral startCol)) (Position (fromIntegral startLine) (fromIntegral $ startCol + len))
+
+
+-- nameLength is in code points unit.
+-- while Range might not in code points unit.
+-- but we can still use it to get the length
+-- since it is only used to exclude some names
+-- currently, we only break `(ModuleA.b)` into `(ModuleA` and `.b)`
```
michaelpj, line 110: comment is wrong

plugins/hls-semantic-tokens-plugin/test/testdata/T1.expected
```diff
@@ -72,7 +73,8 @@
 41:1-3 TFunction "go"
 41:6-9 TRecordField "foo"
 42:1-4 TFunction "add"
-42:7-18 TClassMethod "(Prelude.+)"
+42:8-16 TModuleName "Prelude."
```
michaelpj, line 76: Can we have a test with a multiple-segment module name too? i.e. `(Foo.Bar.+)` or something

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
soulomoon: I update the doc, should be clear now.

plugins/hls-semantic-tokens-plugin/test/testdata/T1.expected
soulomoon: improve in TQualifiedName.expected

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: wait until the prioerities properties added.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Utils.hs
soulomoon: > I'm still confused about why we can't use the text from the OccName? In particular, doesn't occNameString . nameOccName give you what you want? You even use that below!
the reason is here https://github.com/haskell/haskell-language-server/issues/3957#issuecomment-1893485590

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
soulomoon: done

soulomoon:
@michaelpj  Thanx for the review. Should be good now.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -62,39 +70,45 @@ nameNameSemanticFromHie hieKind rm ns = do
...
+        xs = S.fromList $ M.keys $ M.filterWithKey inclusion bd
         inclusion :: Identifier -> IdentifierDetails a -> Bool
         inclusion a b = not $ exclusion a b
         exclusion :: Identifier -> IdentifierDetails a -> Bool
         exclusion idt IdentifierDetails {identInfo = infos} = case idt of
-          Left _  -> True
-          Right _ -> any isEvidenceContext (S.toList infos)
+          Left _moduleName -> False
+          -- filter out the evidence names since they are visible,
+          -- derived names are not filtered out since they are not visible at use site.
```
michaelpj, line 100: Not the point of this PR, but this makes me wonder: won't such identifiers have a missing or zero-length span? So would it be enough to filter them when we create tokens by ignoring anything that has bad spans?

soulomoon: Zero span sound like a ghc error.

michaelpj: what spans do the evidence names have?

soulomoon:
> I'm also concerned that this is going to have a performance impact. We now query the VFS and do text processing for many (every?) token, which is a lot of additional work compared to just using the pre-computed spans in the HIE AST.
 
 yes, this is what I concerned too.
But this should be more ergonomic than the conversion we have done to support code points(this one is every token).
Since we have skipped a lot of case by comparing the source span and length of the occName

soulomoon:
> I'd like a comment from @wz1000 about whether there is another way to get the full name text before we merge.

It would be great if there is way to get the source text for token more efficient.

wz1000:
It does not seem ideal to do range lookups into the source text repeatedly. A better approach would be to walk the lexed source and the hie ast simultaneously, like the haddock hyperlinker:

https://gitlab.haskell.org/ghc/haddock/-/blob/ghc-head/haddock-api/src/Haddock/Backends/Hyperlinker/Renderer.hs?ref_type=heads#L78

michaelpj:
It might also be fine if we can manage to only do this for things we know are either a) functions used infix, or b) operator sections. Can we recognize that from the HIE AST?

soulomoon:
> It might also be fine if we can manage to only do this for things we know are either a) functions used infix, or b) operator sections. Can we recognize that from the HIE AST?

we can, we just check occname length and its span. If not equal, it is either: function used infix, operator section, qualified names.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
soulomoon: the span it binds the evidence on

soulomoon:
Walk along with the source code similar to https://github.com/haskell/haskell-language-server/pull/3958#issuecomment-1898180693

wz1000:
I will review this over the weekend.

michaelpj:
BTW @soulomoon you should join the matrix channel

soulomoon:
> BTW @soulomoon you should join the matrix channel

sure, is it the address ? https://app.element.io/#/room/#haskell-tooling:matrix.org

wz1000:
https://matrix.to/#/#haskell-language-server:matrix.org

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Types.hs
```diff
@@ -25,6 +25,8 @@ import           Development.IDE.Graph.Classes (Hashable)
 import           GHC.Generics                  (Generic)
 import           Language.LSP.Protocol.Types
 -- import template haskell
+import           Data.Map                      (Map)
+import           Data.Set                      (Set)
```
wz1000, line 29: strict maps

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,202 @@
...
+          occStr <- lift $ case (occNameString . nameOccName) name of
+            -- the generated selector name with {-# LANGUAGE DuplicateRecordFields #-}
+            '$' : 's' : 'e' : 'l' : ':' : xs -> Just $ takeWhile (/= ':') xs
+            ['$']                            -> Just "$"
+            -- other generated names that should not be visible
+            '$' : _                          -> Nothing
+            ns                               -> Just ns
+          case ranSplit of
+            (NoSplit (tk, r)) -> do
+              guard $ T.unpack tk == occStr
```
wz1000, line 119: string comparisions can be expensive, using something like `Development.IDE.Types.Export.renderOcc = decodeUtf8 . bytesFS . occNameFS`, or even something like `encodeUtf8 text == bytesFS (occNameFS)` since encoding UTF8 with text 2.0+ is just a simple copy. Ideally we could avoid even the copy if we just compared the raw bytes in the `Text` and the `FastString` from `occNameFS` but this is not very easy to do.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,202 @@
...
+  let span = nodeSpan leaf
+  (ran, token) <- focusTokenAt leaf
+  -- if `focusTokenAt` succeed, we can safely assume we have shift the cursor correctly
+  -- we do not need to recover the cursor state, even if the following computation failed
+  liftMaybeM $ do
+    -- only handle the leaf node with single column token
+    guard $ srcSpanStartLine span == srcSpanEndLine span
+    splitResult <-  lift $ splitRangeByText token ran
+    modify $ \s -> s {currentRange = ran, currentRangeContext = splitResult}
+    mapM_ combineNodeIds $ Map.filterWithKey (\k _ -> k == SourceInfo) $ getSourcedNodeInfo $ sourcedNodeInfo leaf
```
wz1000, line 99: Could we remove `currentRange` and `currentRangeContext` from the state if we passed as arguments to `combineNodeIds`?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,202 @@
...
+  let (startPos, endPos) = srcSpanCharPositions span
+  startOff <- lift $ startPos `sub` cursor
+  tokenOff <- lift $ endPos `sub` startPos
+  (gap, startRope) <- lift $ charSplitAtPositionMaybe startOff rope
+  (token, remains) <- lift $ charSplitAtPositionMaybe tokenOff startRope
+  let ncs = newColumn columnsInUtf16 gap
+  let nce = newColumn ncs token
+  -- compute the new range for utf16, tuning the columns is enough
+  let ran = codePointRangeToRangeWith ncs nce $ realSrcSpanToCodePointRange span
+  modify $ \s -> s {columnsInUtf16 = nce, rope = remains, cursor = endPos}
```
wz1000, line 143: This stretch of code is a bit hard to follow. Perhaps it could use more comments explaining what is going on?

Here is my attempt:

We start with a `rope` representing the document remaining to be processed. `cursor` is a position (utf8?) relative to the start of the document, and is relative to the entire document, not the fragment we have in the `rope`. We want to focus on the position given by the `nodeSpan` of the `HieAST`, and return the token `(Range, Text)` at that span, advancing the `cursor` and the `rope` past the span. `nodeSpan`(utf8?) is also relative to the start of the entire document and not the fragment. `columnsInUtf16` is the number of columns in the utf8 encoding of the current line.

To obtain offsets corresponding to the `nodeSpan` in the fragment, we must subtract the `cursor` from the start position of the `nodeSpan`. This gives us a (utf8?) offset into `rope`. We subtract the start position from the endposition to obtain the (utf8?) length of the token. We can then advance the `cursor` and the `rope` past the token, obtaining the (utf8?) text of the current token in the process.

To obtain the utf16 range, we must advance the current columns first past the initial segment of the rope that we skip past (using its utf16 length, and accounting for newlines by only using the length from the last newline to the end of the segment), and then again past the current token (similarly). The first skip gives us the start column in utf16 bytes and the second skip gives us the end column, also in utf16 bytes. We finally update the `columnsInUtf16` using the end column.

wz1000 (changes requested):
Looks good, just a few minor strictness and performance wiggles and a request for a bit more documentation.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,202 @@
...
+{-# LANGUAGE RankNTypes          #-}
+
+module Ide.Plugin.SemanticTokens.Tokenize (hieAstSpanIdentifiers) where
+
+import           Control.Lens                    (Identity (runIdentity))
+import           Control.Monad                   (forM_, guard)
+import           Control.Monad.State             (MonadState (get),
+                                                  MonadTrans (lift), execStateT,
+                                                  gets, modify, put)
+import           Control.Monad.Trans.State       (StateT)
```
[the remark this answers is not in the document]
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Types.hs
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,202 @@
...
+import           Ide.Plugin.SemanticTokens.Types (RangeIdSetMap)
+import           Language.LSP.Protocol.Types     (Position (Position),
+                                                  Range (Range), UInt, mkRange)
+import           Language.LSP.VFS                hiding (line)
+import           Prelude                         hiding (length, span)
+
+type Tokenizer m a = forall t. StateT (PTokenState t) m a
+
+
+data PTokenState t = PTokenState
```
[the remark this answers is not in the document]
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Query.hs
```diff
@@ -1,100 +1,71 @@
...
 {-# LANGUAGE ScopedTypeVariables #-}
 
 -- |
 -- The query module is used to query the semantic tokens from the AST
 module Ide.Plugin.SemanticTokens.Query where
 
-import           Data.Either                          (rights)
 import           Data.Foldable                        (fold)
 import qualified Data.Map                             as M
 import qualified Data.Map                             as Map
```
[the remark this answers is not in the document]
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -21,8 +23,10 @@ import           Control.Monad.Except                     (ExceptT, liftEither,
                                                            withExceptT)
 import           Control.Monad.Trans                      (lift)
 import           Control.Monad.Trans.Except               (runExceptT)
-import           Data.Aeson                               (ToJSON (toJSON))
+import           Data.Map                                 (Map)
+import qualified Data.Map                                 as M
```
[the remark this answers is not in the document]
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -101,27 +106,28 @@ getSemanticTokensRule recorder =
...
-    let names = nameSetElemsStable $ unionNameSets $ Map.elems spanNamesMap
-    let localSemanticMap = mkLocalNameSemanticFromAst names (hieKindFunMasksKind hieKind) refMap
+    let spanIdMap = M.filter (not . null) $ hieAstSpanIdentifiers virtualFile ast
+    let names = S.toList $ S.unions $ Map.elems spanIdMap
+    let localSemanticMap = mkLocalIdSemanticFromAst names (hieKindFunMasksKind hieKind) refMap
     -- get imported name semantic map
-    let importedNameSemanticMap = foldr (getTypeExclude localSemanticMap getTyThingMap) emptyNameEnv names
-    let sMap = plusNameEnv_C (<>) importedNameSemanticMap localSemanticMap
-    let rangeTokenType = extractSemanticTokensFromNames sMap spanNamesMap
+    let importedIdSemanticMap = foldr (getTypeExclude localSemanticMap getTyThingMap) mempty names
```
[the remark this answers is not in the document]
soulomoon: good idea, done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
soulomoon: yes, you are right, some renaming and doc are added now to make the code more readable

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
wz1000: Did you still mean to use `foldr` here? Why not `fromSet`?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,202 @@
...
+      case idt of
+        Left _moduleName -> addRangeIdSetMap ran idt
+        Right name -> do
+          ranSplit <- gets currentRangeContext
+          occStr <- lift $ case (occNameString . nameOccName) name of
+            -- the generated selector name with {-# LANGUAGE DuplicateRecordFields #-}
+            '$' : 's' : 'e' : 'l' : ':' : xs -> Just $ takeWhile (/= ':') xs
+            ['$']                            -> Just "$"
+            -- other generated names that should not be visible
+            '$' : _                          -> Nothing
```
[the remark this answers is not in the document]
soulomoon: lets handle those as well to be sure.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
soulomoon: revised, see if current appoach would be better.

soulomoon:
@wz1000 Thank you for your detailed reviews. I have made some adjustment to the code.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
soulomoon: Some comments are added to PTokenState
and Some added to this stretch of the code.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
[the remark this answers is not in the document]
soulomoon: I was under the impression using `fromSet` would still have the same time complexity since I was using `elem` in getTypeExclude.

Now, giving a second thought, combining `fromSet` with `Set.difference` does reduce time complexity a bit for some cases.  reviced!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
wz1000: I would try `encodeUtf8 text == bytesFS (occNameFS occ)` as that skips UTF-8 validation.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,198 @@
...
+      case idt of
+        Left _moduleName -> addRangeIdSetMap ran idt
+        Right name -> do
+          occStr <- lift $ T.pack <$> case (occNameString . nameOccName) name of
+            -- the generated selector name with {-# LANGUAGE DuplicateRecordFields #-}
+            '$' : 's' : 'e' : 'l' : ':' : xs -> Just $ takeWhile (/= ':') xs
+            ['$']                            -> Just "$"
+            -- other generated names that should not be visible
+            '$' : _                          -> Nothing
+            _c : ':' : _                     -> Nothing
```
wz1000, line 107: I think you need a `isAlpha _c` guard.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
soulomoon: We are comparing the extracted text(`Text`) from rope with the occStr(`Text`) that have stripped the generated prefix.
I am not sure it is going to be better if we `encodeUtf8` on both side and compares them.🤔

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -0,0 +1,200 @@
...
+    getIdentifier ran ranSplit idt = liftMaybeM $ do
+      case idt of
+        Left _moduleName -> addRangeIdSetMap ran idt
+        Right name -> do
+          occStr <- lift $ T.pack <$> case (occNameString . nameOccName) name of
+            -- the generated selector name with {-# LANGUAGE DuplicateRecordFields #-}
+            '$' : 's' : 'e' : 'l' : ':' : xs -> Just $ takeWhile (/= ':') xs
+            ['$']                            -> Just "$"
+            -- other generated names that should not be visible
+            '$' : _                          -> Nothing
```
wz1000, line 108: You need the guard here as well I think

soulomoon: True, and added here and the previous case.

wz1000 (approved):

soulomoon:
> another thing, though this doesn't have to be addressed in this MR and could be a follow up - now that we are traversing the entire document in order, do we still need to accumulate results into a `M.Map Range HsSemanticTokenType`? Couldn't we instead directly accumulate a `[(Range, HsSemanticTokenType)]`?

Good point, I'll take a look some time later if there is optimization space we can do on this.

soulomoon:
> another thing, though this doesn't have to be addressed in this MR and could be a follow up - now that we are traversing the entire document in order, do we still need to accumulate results into a `M.Map Range HsSemanticTokenType`? Couldn't we instead directly accumulate a `[(Range, HsSemanticTokenType)]`?

One thing I am not sure about would be wether the mappings of range change would change the range ordering.
If the ordered would change, we are forced to sort it again.

wz1000:
> One thing I am not sure about would be wether the mappings of range change would change the range ordering. If the ordered would change, we are forced to sort it again.

you mean the `positionMappings`? I'm pretty sure that they are guaranteed to be monotonic, so I don't think they would change sort order.

soulomoon:
Yes, thanx for clearing this out!

Issue #3098, opened 2022-08-12
Display the derived instances while hovering on type constructors and data constructors


July541:
## Is your enhancement request related to a problem? Please describe.
No

## Describe the solution you'd like
Display the derived instances while hovering on type constructors and data constructors.

For example, for `data A = A deriving Show; a = A`, it would be great to display it derived the `Show` while hovering over all `A`s

## Describe alternatives you've considered
Not sure if there is a better way to show these instances

Jashweii:
Is there a reason for only derived instances rather than all instances, like :instances <type> in GHCi?
```
ghci> :instances Int
instance Foreign.Storable.Storable Int
  -- Defined in `Foreign.Storable'
instance GHC.Bits.Bits Int -- Defined in `GHC.Bits'
instance GHC.Bits.FiniteBits Int -- Defined in `GHC.Bits'
instance Integral Int -- Defined in `GHC.Real'
instance GHC.Ix.Ix Int -- Defined in `GHC.Ix'
instance Num Int -- Defined in `GHC.Num'
instance Real Int -- Defined in `GHC.Real'
instance Bounded Int -- Defined in `GHC.Enum'
instance Enum Int -- Defined in `GHC.Enum'
instance Eq Int -- Defined in `GHC.Classes'
instance Ord Int -- Defined in `GHC.Classes'
instance Read Int -- Defined in `GHC.Read'
instance Show Int -- Defined in `GHC.Show'
```
Also, class constraints on local type variables?

July541:
> Is there a reason for only derived instances rather than all instances, like :instances in GHCi?

When I said `derived`, it also includes these instances you mentioned.

> Also, class constraints on local type variables?

Not sure about that, an example here?

Jashweii:
> > Also, class constraints on local type variables?
> 
> Not sure about that, an example here?

Like ``a`` in these examples though I guess it isn't necessarily restricted to type variables
```Haskell
eg1 :: forall a . Show a => a -> String
eg1 = show @a

data Eg2 a where
  MkEg2 :: Show a => Eg2 a
eg2 :: forall a . Eg2 a -> a -> String
eg2 MkEg2 = show @a
```

michaelpj:
Showing *all* the instances might be a lot. Some types have a lot of instances. But perhaps that's fine, so long as it goes at the bottom or something, like in Haddock.

July541:
[...]

PR #3963, opened 2024-01-15
Display instances on hover for type and data constructors

DeviousStoat:
An attempt to implement the requested enhancement in https://github.com/haskell/haskell-language-server/issues/3098

It displays the list of typeclass instances a type has when requesting hover information.
It should work on type constructors as well as the data constructors, using the related type.

It is added as a new section at the end of the hover information.

Here is an example of how it looks when hovering on `Maybe` or `Just` or `Nothing`
maybe_instances

It seems to also be scope aware, if I import new instances then they will be added to the list. For example if I install `hashable` and `import Data.Hashable` with the `Hashable` instance for `Int` and hover on `Int` I get an additional
hashable_int
which is pretty cool.

### DIFF
--- a/ghcide/src/Development/IDE/Spans/AtPoint.hs
+++ b/ghcide/src/Development/IDE/Spans/AtPoint.hs
@@ -316,7 +316,10 @@ atPoint IdeOptions{} (HAR _ hf _ _ (kind :: HieKind hietype)) (DKMap dm km) env 
                 instancesForName :: IO (Maybe [ClsInst])
                 instancesForName = runMaybeT $ do
                     typ <- MaybeT . pure $ lookupNameEnv km n >>= tyThingAsDataType
-                    liftIO $ evalGhcEnv env $ getInstancesForType typ
+                    clsInst <- liftIO $ evalGhcEnv env $ getInstancesForType typ
+                    -- Avoid creating an empty wrapped section if no instances are found
+                    guard $ not $ null clsInst
+                    return clsInst
 
                 -- | Gets the datatype `Type` corresponding to a TyThing, if it repressents a datatype or
                 -- a data constructor.


soulomoon:
It looks cool, 🤔 how is the result sorted, should we sort it out?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -59,6 +60,7 @@ import           Data.List.Extra                      (dropEnd1, nubOrd)
 
 import           Data.Version                         (showVersion)
 import           Development.IDE.Types.Shake          (WithHieDb)
+import           GHC                                  (getInstancesForType)
```
soulomoon, line 63: Can we get it from Compat?

soulomoon (commented):
look good, some questions

ghcide/src/Development/IDE/Spans/AtPoint.hs
soulomoon: Since in general we import ghc stuff from `Development.IDE.GHC.Compat`

DeviousStoat: so I should just export it in `Development.IDE.GHC.Compat`? Why is this needed? It seems to be available in this module in all the ghc versions supported by hls

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -308,6 +311,24 @@ atPoint IdeOptions{} (HAR _ hf _ _ (kind :: HieKind hietype)) (DKMap dm km) env
...
+        prettyInstances :: (Either ModuleName Name, IdentifierDetails hietype) -> IO (Maybe T.Text)
+        prettyInstances (Right n, _) =
+            fmap (wrapHaskell . T.unlines . fmap printOutputable) <$> instancesForName
+            where
+                instancesForName :: IO (Maybe [ClsInst])
+                instancesForName = runMaybeT $ do
+                    typ <- MaybeT . pure $ lookupNameEnv km n >>= tyThingToType
+                    liftIO $ evalGhcEnv env $ getInstancesForType typ
+
+                tyThingToType :: TyThing -> Maybe Type
```
michaelpj, line 323: Name change since otherwise this is confusing with e.g. `safeTyThingType` which gets the type _of_ a `TyThing`
```suggestion
                -- | Gets the datatype 'Type' corresponding to a TyThing, if it represents a datatype or
                -- a data constructor
                tyThingAsDataType :: TyThing -> Maybe Type
```

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -308,6 +311,24 @@ atPoint IdeOptions{} (HAR _ hf _ _ (kind :: HieKind hietype)) (DKMap dm km) env
             UnhelpfulLoc {} | isInternalName name || isSystemName name -> Nothing
             _ -> Just $ "*Defined " <> printOutputable (pprNameDefnLoc name) <> "*"
 
+        prettyInstances :: (Either ModuleName Name, IdentifierDetails hietype) -> IO (Maybe T.Text)
+        prettyInstances (Right n, _) =
+            fmap (wrapHaskell . T.unlines . fmap printOutputable) <$> instancesForName
+            where
+                instancesForName :: IO (Maybe [ClsInst])
+                instancesForName = runMaybeT $ do
+                    typ <- MaybeT . pure $ lookupNameEnv km n >>= tyThingToType
```
michaelpj, line 320: It looks like we look things up repeatedly in `km`. Should we do this once and then just pass in the `TyThing` to all the places that need it?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -308,6 +311,24 @@ atPoint IdeOptions{} (HAR _ hf _ _ (kind :: HieKind hietype)) (DKMap dm km) env
             UnhelpfulLoc {} | isInternalName name || isSystemName name -> Nothing
             _ -> Just $ "*Defined " <> printOutputable (pprNameDefnLoc name) <> "*"
 
+        prettyInstances :: (Either ModuleName Name, IdentifierDetails hietype) -> IO (Maybe T.Text)
+        prettyInstances (Right n, _) =
```
michaelpj, line 315: In cases like this I'd usually just make the function take what it needs (just a `Name`) and get the caller to worry about massaging the list into the right form.

ghcide/src/Development/IDE/Spans/AtPoint.hs
michaelpj: I'm unsure of our policy here, wdyt @wz1000 ?

michaelpj (commented):
Looks plausible to me, I'd like a review from @wz1000 . And it needs some tests.

I note that we have a problem in our hovers in that there are no explanations for what the various sections of the hover mean. In this case I think it's obvious, but perhaps we could take the opportunity to add titles to the sections? For this bit I think it could just be `Instances in scope for <name>`?

ghcide/src/Development/IDE/Spans/AtPoint.hs
DeviousStoat: I am not sure what you mean. I think the lookup only happens once per `Name` and we need to look it up for each `Name`

ghcide/src/Development/IDE/Spans/AtPoint.hs
DeviousStoat: Sounds better, I switched it up

DeviousStoat:
About the titles, I personally don't think it is too useful, imo it would just add unecessary clutter in the hover text.
You say `In this case I think it's obvious` but I don't really see any other section that is not obvious.
But of course if you all think it is a good idea I am willing to add them (I am kind of starting in the haskell world so I am a sucker for some small not too tricky tasks to improve :D) but maybe in another PR afterwards as it would be awkward to have only one title here. And adding them all in this PR is a bit out of scope.

DeviousStoat:
hum, I used it a bit more and there are a few issues with this feature as it is currently implemented actually.
Hovering on some imported datatypes that are supposed to have instances don't show anything on hover. And it also doesn't show anything for datatypes defined in the current module with instances. I am not sure what's going on yet. I will be looking into it but I guess this PR is still WIP then.

ghcide/src/Development/IDE/Spans/AtPoint.hs
michaelpj: I mean in the other functions we call around here. Probably not a huge deal

michaelpj:
Yes fine, we can leave the title thing for later

DeviousStoat:
I looked a bit into it and I might need some of guidance, there are some things I don't really understand.
So the issue I have is that the instances coming from external libraries apart from GHC base libraries and from the current file are not shown at all.

I believe the issue comes from there
```hs
clsInst <- liftIO $ evalGhcEnv env $ getInstancesForType typ
```

It seems that the `env` doesn't have the required modules loaded at this point.

Doesn't the `GHCSession` action takes care of this? I tried using `GHCSessionDeps` but it didn't change anything. Do you know what the difference is between the 2?

I found this [function](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs#L244) in the `hls-eval-plugin`. It seems to have to manually load the modules in the session. Is it something I would probably have to do to execute interactive functions like the `getInstancesForType`? If so I don't exactly understand what this `env` is all about, is it just an empty ghc execution context?

soulomoon:
> Doesn't the GHCSession action takes care of this? 

Perhaps we are not taking care of things that is not used yet, you can dig deeper to see how do we manage to include the information in the session.

PR #3964, opened 2024-01-16
Fix -Wall and -Wunused-packages in hls-alternate-number-format-plugin

jhrcek:
I'm on a mission to fix this warning across all packages, one plugin at a time :smiley: 

EDIT: It's so tempting to fix all the other warnings that I went away and also fixed -Wall in this plugin.


plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
```diff
@@ -168,10 +167,10 @@ toDecimal :: Integral a => a -> String
 toDecimal = toBase showInt ""
 
 toBinary :: (Integral a, Show a) => a -> String
-toBinary = toBase showBin "0b"
+toBinary = toBase showBin_ "0b"
   where
     -- this is not defined in versions of Base < 4.16-ish
-    showBin = showIntAtBase 2 intToDigit
+    showBin_ = showIntAtBase 2 intToDigit
```
jhrcek, line 173: Avoiding name shadoing warning.
Technically we could raise the lower bound of base (>= 4.16) and just use showBin from base now, that we only officially support ghc 9.2+, but I went with this less invasive "fix".

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
```diff
@@ -35,7 +35,7 @@ import           Ide.Types
 import qualified Language.LSP.Protocol.Lens       as L
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types
-
+import           Prelude                          hiding (log)
```
fendor, line 38: Personally, I'd prefer to rename the log messages instead of adding an import statement. Anyone else agrees?

fendor (approved):
LGTM, thanks!

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/Conversion.hs
fendor: I think fixing it immediately is fine, as it is still trivial to review.

plugins/hls-alternate-number-format-plugin/src/Ide/Plugin/AlternateNumberFormat.hs
jhrcek: I saw several plugins name their logging function log and hide Prelude.
I'm not against renaming it to something else if you suggest a reasonable name to rename it to :smile:

fendor: Yeah, I know a number of plugins are doing that.

I like `msg` when you just pass something through such as `ShakeLog`. But it can also be the more wordy `shakeLog` or even `l`, as the name is not giving you much in this particular situation.

Otherwise, we usually don't have something such as `Log String`, but more concrete types in our logging constructors. These concrete types can have more descriptive names, just like in a regular function.

jhrcek: Alright, renamed to `msg`

michaelpj (approved):

PR #3967, opened 2024-01-16
Semantic tokens: expand type synonym to checkout forall function type when possible

### LINKED ISSUE #3966: Semantic tokens forall function hide under type syn is not recognized
Current implmentation is does expand type synoym for `type` from the `in memory` hieAst. 

So the forall function hide under type syn is not recognized
1. T1 type syn is expanded since `isFunType` check so.
2. T2 is not, since the check only expand forall. 

It result in forall hide behide type syn is not expanded
image

soulomoon:
should fix #3966
1. Add test for function type under type synonym
2. Expand type synoym for type from the in memory hieAst to extract function mask.

Forall function under type syn is checked out now.
image


wz1000 (approved):
looks reasonable to me.

soulomoon:
> looks reasonable to me.

Thanx for the review

PR #3973, opened 2024-01-17
Exactprint plugins for 9.8

wz1000:
I ran out of steam fixing this refactor plugin test:

```
    suggest import actions
      Dont want suggestion
        f = nonEmpty:       FAIL (0.21s)
          test/Main.hs:1772:
          expected: []
           but got: ["import Data.List.NonEmpty (nonEmpty)"]
          Use -p '/f = nonEmpty/' to rerun this test only.
```

Perhaps someone else would like to have a look?

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -217,7 +217,7 @@ common cabal
     cpp-options: -Dhls_cabal
 
 common class
-  if flag(class) && (impl(ghc < 9.8.0) || flag(ignore-plugins-ghc-bounds))
+  if flag(class)
     build-depends: hls-class-plugin == 2.6.0.0
     cpp-options: -Dhls_class
 
@@ -237,12 +237,12 @@ common importLens
     cpp-options: -Dhls_importLens
 
 common rename
-  if flag(rename) && (impl(ghc < 9.8.0) || flag(ignore-plugins-ghc-bounds))
+  if flag(rename)
     build-depends: hls-rename-plugin == 2.6.0.0
     cpp-options: -Dhls_rename
 
 common retrie
-  if flag(retrie) && (impl(ghc < 9.8.0) || flag(ignore-plugins-ghc-bounds))
+  if flag(retrie)
     build-depends: hls-retrie-plugin == 2.6.0.0
     cpp-options: -Dhls_retrie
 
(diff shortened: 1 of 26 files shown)


haskell-language-server.cabal
```diff
@@ -217,7 +217,7 @@ common cabal
     cpp-options: -Dhls_cabal
 
 common class
-  if flag(class) && (impl(ghc < 9.8.0) || flag(ignore-plugins-ghc-bounds))
+  if flag(class)
```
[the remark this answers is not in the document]
wz1000: aah yes I need to rebase

plugins/hls-gadt-plugin/test/Main.hs
```diff
@@ -35,13 +35,13 @@ tests = testGroup "GADT"
     , runTest "ConstructorContext" "ConstructorContext" 2 0 2 38
     , runTest "Context" "Context" 2 0 4 41
     , runTest "Pragma" "Pragma" 2 0 3 29
-    , onlyWorkForGhcVersions (`elem`[GHC92, GHC94, GHC96]) "Single deriving has different output on ghc9.2+" $
+    , onlyWorkForGhcVersions (`elem`[GHC92, GHC94, GHC96, GHC98]) "Single deriving has different output on ghc9.2+" $
```
michaelpj, line 38: we only support GHC 9.2 plus now, so I think we can simplify this stuff

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -703,7 +703,11 @@ typeWildCardActionTests = testGroup "type wildcard actions"
         [ "func :: _"
         , "func x y = x + y"
         ]
+#if MIN_VERSION_ghc(9,7,0)
+        [ "func :: a -> a -> a" -- 9.8 has a different suggestion
```
michaelpj, line 707: this seems kind of bad, is this an upstream regression?

michaelpj (commented):
Changes look plausible, would be nice to get the last few tests sorted!

wz1000:
I have no idea what the failing test is testing for or why it is desirable. It seems require that we don't suggest adding a `nonEmpty` import from `Data.List.NonEmpty` for this code. Why?

```haskell
import Data.List.NonEmpty ()

f = nonEmpty
```

wz1000:
I see, it wants us to not suggest adding a new import, instead we should suggest extending the existing import.

All the tests have been fixed locally now.

michaelpj:
Great, thanks for tracking that down!

michaelpj (approved):

konn (approved):
LGTM as for Splice Plugin, Thanks!

michaelpj:
Yeah, it seems like we can use `ghcVersion` whenever we just have differences in values, which seems like a win!

michaelpj:
Still a build failure in the gadt tests

PR #3975, opened 2024-01-17
Fix most hlint warnings in ghcide

jhrcek:
If nothing else, I hope this will reduce the amount of hlint warnings that are showing up in unrelated PRs.


michaelpj (approved):
I'm not sure I'd consider all of these strict improvements, but nice to quieten things down. I think we should also just be aggressive about turning off stuff we don't like in `hlint.yaml`

jhrcek:
CI passed, ready for merge.

Issue #3556, opened 2023-04-05
Consider packaging HLS as a single cabal package with multiple public libraries


michaelpj:
Today, HLS consists of _37_ packages. This is a huge pain to maintain and release, especially since we currently don't version them all identically. We almost always have issues with the Hackage release where some of the bounds aren't correct.

However, in reality they are all tested and built _together_ by the HLS developers, and while in principle you can mix-and-match the versions of various plugins based on the bounds, it's not really something we _need_ to support, or that brings value to users. Users mostly just want to get a HLS binary! 

We do also have some users who build custom LSP systems using bits of HLS, notably SigmaIDE.

That suggests that we could just package HLS as one cabal package with multiple sub-libraries. They would need to be _public_ sub-libraries in order to still enable the SigmaIDE use case.

That would simplify our lives a lot:
- Only one package
- Only one version
- Only one set of package metadata
- Common stanzas could be used project-wide

The reasons not to do this are:
- Stack doesn't support multiple public libraries (this doesn't seem like too big a deal, it only affects people who want to build stuff on top of HLS, which is a small set of users)
- Hackage [apparently](https://github.com/haskell/hackage-server/issues/1119) supports them, but it's perhaps not well-tested

The uncertainty is probably the best reason not to do this. It would be good to see some successful usage of multi-libs on Hackage before we jump in.

michaelpj:
I think we could consider doing this. I think we should leave `ghcide` split out, but we could merge HLS and its plugins.

michaelpj:
We did this!

PR #3976, opened 2024-01-17
Merge plugins into the HLS package

michaelpj:
This is a POC for merging all the plugin packages into the HLS package, as discussed in https://github.com/haskell/haskell-language-server/issues/3556.

I just did the first two plugins alphabetically to show what it looks like, it's pretty mechanical.

To reiterate, the advantages are:
- Fewer packages to upload
- Less repeated metadata
- Fewer versions
- Ability to use common stanzas to e.g. have the same set of warnings everywhere (I would do this as a follow up if we decide to go this route)

Things seem to still work fine. The main annoyances are:
- Can't `cd` into a plugin directory and `cabal build`/`cabal test`
- Can't `cabal test <plugin-name>`, instead need `cabal test <plugin-name>-tests`

I think overall it's still going to be a pretty big improvement


soulomoon:
One high draw back, it prevent us from working on a single plugin along.

michaelpj:
I don't think it does? You can still just build the component for the plugin and it won't build anything else.

soulomoon:
But we can no longer open an single editor containing just one plugin dir and get hls support 🤔

michaelpj:
I don't see why not? We do support single components of packages just fine? And it'll be better as the multi-component stuff rolls out (but that's already bad if you try to edit multiple packages today)

soulomoon:
> I don't see why not? We do support single components of packages just fine? And it'll be better as the multi-component stuff rolls out (but that's already bad if you try to edit multiple packages today)

Thanx for clearing out my concern on this.

michaelpj:
Another downside: stack supports building packages with internal libraries (the CI run succeeded!), but it doesn't do per-component builds, so the dev experience using stack will probably suck. I'm not sure how much we care about that these days :shrug:

michaelpj:
No longer a POC! 

There is still plenty to do here with cleaning up the now rather monolithic cabal file, making better use of common stanzas etc, but I'm keen to land this before too long since it's very conflict-y.

michaelpj:
cc @fendor @wz1000  as discussed

michaelpj:
Follow up todos:
- Make all components which condition on `pedantic` use the common stanza for that
- Add a common stanza for `default-language`
- Maybe merge in some more packages. I think only `hls-test-utils`, everything else is a dependency of `ghcide` and so needs to stay out for now.
- Maybe try and organize the cabal file to be a bit more navigable

haskell-language-server.cabal
```diff
@@ -63,286 +69,1679 @@ flag ignore-plugins-ghc-bounds
...
+    , text
+    , transformers
+    , unordered-containers
+
+  default-language:   Haskell2010
+  default-extensions:
+    DataKinds
+    TypeOperators
+
+  ghc-options:        -Wno-unticked-promoted-constructors
```
jhrcek, line 588: Can be removed - is now in common warnings.

haskell-language-server.cabal
```diff
@@ -63,286 +69,1679 @@ flag ignore-plugins-ghc-bounds
...
+    , temporary
+    , text
+    , transformers
+    , unordered-containers
+    , ghc-lib-parser
+    , ghc-lib-parser-ex
+    , apply-refact
+
+  cpp-options:   -DHLINT_ON_GHC_LIB
+  ghc-options:
```
jhrcek, line 656: You can import common warnings for this lib and remove this.

jhrcek (commented):
I like this approach more than the previous.
Few nitpicks about ghc-options - warnings, but I will work on those anyway in followup PRs, so feel free to ignore.

haskell-language-server.cabal
```diff
@@ -63,286 +69,1679 @@ flag ignore-plugins-ghc-bounds
...
+  default-language: Haskell2010
+
+test-suite hls-refactor-plugin-tests
+  import: warnings
+  type:             exitcode-stdio-1.0
+  default-language: Haskell2010
+  hs-source-dirs:   plugins/hls-refactor-plugin/test
+  main-is:          Main.hs
+  other-modules:    Test.AddArgument
+  ghc-options:      -O0 -threaded -rtsopts -with-rtsopts=-N -Wunused-imports
```
jhrcek, line 1633: I'm pretty sure that -Wunused-imports are enabled by -Wall, so you could remove it from here since you're importing -Wall via `import: warnings`

michaelpj:
Yeah, I didn't do any cleanup on the cabal file contents, just moved them over as-is. Cleanups later!

michaelpj:
Hmmm... is that a long-path error?

michaelpj:
Yup, the path is 261 characters. I'm going to shorten the component name. Generally I think we could now strip the `hls-` prefixes from everything at least. Frankly, maybe also the `-plugin` suffix?

michaelpj:
Still can't get everything below the limit. At this point it's either do something ad-hoc or turn off GHC9.2+windows

fendor:
> Generally I think we could now strip the hls- prefixes from everything at least. Frankly, maybe also the -plugin suffix?

Changing `hls-fourmolu-plugin` to `fourmolu` will cause a fair amount of confusion. Almost all plugins that essentially wrap a tool would be weird that way.

michaelpj:
Yeah, I think we should probably keep `-plugin`.

fendor:
Personally, I like the explicitness :shrug: It signals an HLS namespace

michaelpj:
Right, but once it's a component of the `haskell-language-server` package it's already namespaced. The full name will be `haskell-language-server:hls-ormolu-plugin` which seems redundant to me.

michaelpj:
You definitely want it when it's _not_ a subcomponent.

michaelpj:
Okay, I think this is now all working, just getting through CI nondeterminism. I've reverted all component name changes for now since I've sidestepped the long path issue. We can discuss that later.

Anyone want to approve?

joyfulmantis (approved):
LGTM

fendor (changes requested):
I think the tests are not run in CI 
image

georgefst:
I suppose the `fourmolu`/`fourmolu-plugin`/`hls-fourmolu-plugin` trade-off comes down to whether sublibraries can only over be referred to outside the package in qualified form. I've dabbled a little with multiple-public-libraries support, but I don't know this off the top of my head, though I'd assume they can't.

In general, this change seems a sensible one. I had assumed when the project started back in 2020 that we'd eventually move towards plugins being maintained outside of the HLS organisation entirely, but this doesn't really work without them being genuinely runtime-pluginnable. And as far as I understand it, this would require significant improvements in GHC dynamic linking support, which no one wants to work on...

tl;dr All good.

joyfulmantis:
> I think the tests are not run in CI image

Yikes! yes we definitely want those tests still

michaelpj:
> I think the tests are not run in CI

What on earth... that is so odd. It works fine for me locally. Investigating

michaelpj:
> I suppose the fourmolu/fourmolu-plugin/hls-fourmolu-plugin trade-off comes down to whether sublibraries can only over be referred to outside the package in qualified form

Note that these are not currently _public_ sublibraries. We could make them public, but I don't know that there are any actual consumers of them. So at the moment we're only using the much older and better-supported non-public internal libraries.

> I had assumed when the project started back in 2020 that we'd eventually move towards plugins being maintained outside of the HLS organisation entirely, but this doesn't really work without them being genuinely runtime-pluginnable

Yes, I think the _reality_ of how plugins are used by HLS is that they are a very handy method of code organization and responsibility allocation, but that's about it. And I think that's fine! My position has always been that nobody actually wants to do runtime customization of their HLS installation, they want it to just work.

michaelpj:
My suspicion is that the test issue was due to tasty-rerun actually working. I've deleted the caches relating to that from this PR and rerun everything, let's see if it now runs the tests.

georgefst:
> Note that these are not currently _public_ sublibraries.

Oh, I just assumed. It is _in principle_ a regression to make all these libraries unusable externally. Though I suspect no one's using them.

michaelpj:
Yes, it's for sure a regression in theory. But personally I _highly_ doubt that anyone is using them. You'd have to be assembling your own custom HLS, and not want to do it using HLS's cabal flags for some reason.

michaelpj:
Nope?

michaelpj:
It was tasty-rerun, turning it off makes the tests run. I don't know whether it's doing something dodgy, but it seems to be orthogonal to this PR at least.

smunix:
@michaelpj - it looks like this change broke `nix` rpath setting for plugins internal libraries. It fails during the `install_phase` for finding references to the `/build/` directory. Have `2.7.0.0` been installed successfully with `nix` at all?

michaelpj:
That seems like a question for the Nix Haskell folks? It builds with cabal, so if it doesn't build with the Nix setup then that's a bug in the Nix setup, I think.

PR #3981, opened 2024-01-18
Fix -Wall and -Wunused-packages in eval plugin

jhrcek:


plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -1,11 +1,11 @@
+{-# LANGUAGE CPP              #-}
```
jhrcek, line 1: Change unrelated to eval plugin, but pre-commit check (which check styling-haskell formatting) was failing after [this PR](https://github.com/haskell/haskell-language-server/pull/3973) was merged to master.

PR #3987, opened 2024-01-19
Update support tables

michaelpj:
Now @wz1000 has fixed the exactprint plugins I think we can declare full support for 9.8.


wz1000:
If this is rendered on to the readthedocs website immediately it might be confusing for users of 2.6.0.0 that exactprint plugins aren't available on their HLS.

michaelpj:
We've always had this problem. If we actually care we need to do per-version docs on RTD, which we can do, but is just More Work. If we wait until the next release we probably won't remember :shrug:

fendor (approved):
Change itself looks good to me!

However, the docs are truly incorrect, as the latest release does not have support for 9.8.1.

We should version the docs, but the additional work seems almost not worth it.

I think we should keep this PR open as a reminder, but not merge quite yet.

michaelpj:
I made versioned docs work

PR #3989, opened 2024-01-19
Remove an allow-newer

michaelpj:


michaelpj:
Hmm, I bumped nixpkgs and now it needs too new a glibc for the GHA runners or something?

fendor:
Now we don't test the nix environment any more? That's not ideal, I think the nix environments are useful and we should maintain them.

The GHA runners are quite annoying at times :sweat_smile:

michaelpj:
It still tests that the nix environment _builds_. I don't know. GLIBC issues really suck and I'm not sure there's an easy solution. This is AFAIK the only way to get a dev shell setup with 9.8, for example :shrug:

fendor (approved):
Change itself looks good.

Let's hope someone who cares about the nix setup will eventually fix it.

PR #3996, opened 2024-01-20
Fix -Wall and -Wunused-packages in explicit-record-fields plugin

jhrcek:


ozkutuk (approved):
LGTM

plugins/hls-explicit-record-fields-plugin/hls-explicit-record-fields-plugin.cabal
```diff
@@ -43,22 +40,17 @@ library
...
-      , ghc-boot-th
-      , unordered-containers
       , containers
       , aeson
     hs-source-dirs:   src
     default-language: Haskell2010
 
   if flag(pedantic)
     ghc-options: -Werror
-                 -Wwarn=incomplete-record-updates
```
ozkutuk, line 55: @jhrcek Now that I look into it, I think the reason it was specified this way was that the plugin _is_ making use of incomplete record updates. The failing test on CI shows where that happens. Off the top of my head I can't think of a straightforward way to eliminate it, as it happens while modifying the AST given by GHC. Maybe you have come across a better way to do this while fixing these warnings for the other plugins?

ozkutuk: Specifically, this section of GHC options means that "We want warnings to be promoted to errors (`-Werror`), except for incomplete record updates". Unless there is a compelling reason to change this, I think this is a sensible setting for this specific plugin.

michaelpj: GHC options on specific files take priority, so I'd be inclined to put the `-Wwarn` option on the module with the problem and add a comment

jhrcek: I understand the semantic of -Werror and how -Wwarn turns it into warning only. I just wasn't sure why it's here (because cabal build --flag=pedantic for this plugin didn't throw any error.
Let me put it back for now and deal with fixing these kinds of warnings separately later.

PR #3998, opened 2024-01-20
Various 9.8 compat

michaelpj:
- New stylish-haskell release supports 9.8
- Ormolu and fourmolu were updated but the test matrix hadn't been
- New version of floskell supports 9.8
- New version of hlint supports 9.8, small test wrangling required

9.8 support shaped up a lot in the last few days!


fendor (approved):
LGTM

PR #4000, opened 2024-01-21
Update base lower bounds for HLS

### LINKED ISSUE #3999: Base lower bounds are wrong
Compiling HLS with GHC-9.0.2 results in compilation errors:

```
[ 6 of 76] Compiling Development.IDE.GHC.Compat.Core ( src/Development/IDE/GHC/Compat/Core.hs, dist/build/Development/IDE/GHC/Compat/Core.o, dist/build/Development/IDE/GHC/Compat/Core.dyn_o )

src/Development/IDE/GHC/Compat/Core.hs:509:1: error:
    Could not load module `GHC.Driver.Env'
    It is a member of the hidden package `ghc-lib-parser-9.4.8.20231111'.
    Perhaps you need to add `ghc-lib-parser' to the build-depends in your .cabal file.
```

Unbuildable configurations must be disallowed by the cabal file, so we don't revive cabal hell.

https://github.com/haskell/haskell-language-server/blob/0a02832ed3d2ef8b5128b1f8de2cbdc98eb4bc2f/haskell-language-server.cabal#L17

https://github.com/haskell/haskell-language-server/blob/0a02832ed3d2ef8b5128b1f8de2cbdc98eb4bc2f/haskell-language-server.cabal#L38

fendor:
Fix #3999

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -73,7 +73,7 @@ library
     , filepath
     , fingertree
     , focus                        >=1.0.3.2
-    , ghc                          >=9.0
+    , ghc                          >=9.2
     , ghc-boot
     , ghc-boot-th
     , ghc-check                    >=0.5.0.8
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -14,7 +14,7 @@ homepage:
   https://github.com/haskell/haskell-language-server/tree/master/ghcide#readme
 
 bug-reports:        https://github.com/haskell/haskell-language-server/issues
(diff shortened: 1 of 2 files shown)


michaelpj (approved):
I'm fine with this. I think the thing that actually caused the compile errors was the `ghc` package, though, and I think it probably makes sense to bound our uses of that too to reflect the GHC's we support?

hasufell (approved):

fendor:
>  I think it probably makes sense to bound our uses of that too to reflect the GHC's we support?

That's true, bumping the lower bound for ghc as well.

PR #4004, opened 2024-01-22
Fix -Wunused-packages in test utils

jhrcek:


hls-test-utils/hls-test-utils.cabal
```diff
@@ -56,9 +53,8 @@ library
     , tasty-rerun
     , temporary
     , text
-    , unordered-containers
     , row-types
-  ghc-options:      -Wall
+  ghc-options:      -Wall -Wunused-packages
```
fendor, line 57: Should this be in a common section?

fendor (approved):
LGTM!

hls-test-utils/hls-test-utils.cabal
jhrcek: This particular package only has one component (library) and no tests etc. So I felt it's overkill to use `common warnings` because it would only be imported in one place.

PR #4007, opened 2024-01-23
Fix -Wall and -Wunused-packages in fourmolu and ormolu plugins

jhrcek:


plugins/hls-ormolu-plugin/test/Main.hs
```diff
@@ -34,9 +34,10 @@ tests = testGroup "ormolu" $
       ]
 
 goldenWithOrmolu :: Bool -> TestName -> FilePath -> FilePath -> (TextDocumentIdentifier -> Session ()) -> TestTree
-goldenWithOrmolu cli title path desc = goldenWithHaskellDocFormatter def ormoluPlugin "ormolu" def title testDataDir path desc "hs"
+goldenWithOrmolu cli title path desc =
+  goldenWithHaskellDocFormatter def ormoluPlugin "ormolu" conf title testDataDir path desc "hs"
  where
-  conf = def{plcConfig = (\(Object obj) -> obj) $ object ["external" .= cli]}
+  conf = def{plcConfig = KM.fromList ["external" .= cli]}
```
jhrcek, line 40: The fact that conf was unused was actual test bug - the tests only ever executed with the default configuration (only def was passed to goldenWithHaskellDocFormatter).

fendor (approved):

PR #4014, opened 2024-01-24
Fix -Wall and -Wunused-packages in stan plugin

jhrcek:
This PR is part of larger effort where I try to reduce the amount of compiler warnings across all packages.

### DIFF
--- a/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
+++ b/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
@@ -80,9 +80,6 @@ stripModifiers = go ""
         Nothing    -> txt
         Just index -> T.drop (index + 1) txt
 
-renderId :: Id a -> T.Text
-renderId (Id t) = "Id = " <> t
-
 instance Pretty Log where
   pretty = \case
     LogShake log -> pretty log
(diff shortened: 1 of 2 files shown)


plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -162,24 +139,21 @@ rules recorder plId = do
 
               -- If envVar is set to 'False', stan will ignore all local and global .stan.toml files
               logWith recorder Debug (LogDebugStanEnvVars env)
-              seTomlFiles <- liftIO $ usedTomlFiles useDefConfig (stanArgsConfigFile stanArgs)
```
jhrcek, line 165: I'm not sure about this, but seTomlFiles unused past this point.
There's also identical statement executed 5 lines above this and the seTomlFiles are just logged.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -162,24 +139,21 @@ rules recorder plId = do
...
+                  ResultL _warnings stanConfig -> do
                       let currentHSAbs = fromNormalizedFilePath file -- hie_hs_file hie
                       currentHSRel <- liftIO $ makeRelativeToCurrentDirectory currentHSAbs
                       cabalExtensionsMap <- liftIO $ createCabalExtensionsMap isLoud (stanArgsCabalFilePath stanArgs) [hie]
 
                       -- Files (keys) in checksMap need to have an absolute path for the analysis, but applyConfig needs to receive relative
                       -- filepaths to apply the config, because the toml config has relative paths. Stan itself seems to work only in terms of relative paths.
                       let checksMap = HM.mapKeys (const currentHSAbs) $ applyConfig [currentHSRel] stanConfig
-
-                      let analysis = runAnalysis cabalExtensionsMap checksMap (configIgnored stanConfig) [hie]
```
jhrcek, line 182: Not sure if this should be removed, but `analysis` - result of this pure computation is unused below, so seems pointless to compute.

michaelpj:
Sounds like there are some genuine questions, perhaps @0rphee can advise.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
0rphee: Yes, it seems I forgot to delete that line when iterating with this function.

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
0rphee: This also slipped past me. The precious (repeated) statement is used to log which .stan.toml files were used to load the stan config. 

The binding should have a better name though. I think `stanTomlUsed` may be more useful.

0rphee:
Everything seems fine. Some errors seem to have eluded me!

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
jhrcek: Ok, I renamed the other binding to `tomlsUsedByStan`

jhrcek:
@0rphee thanks for checking. @michaelpj This PR should be merge-ready.

0rphee (approved):

Issue #3795, opened 2023-09-07
Code action to define undefined variables includes suggestions in type signature


jameshaydon:
### Your environment

MacOS
GHC 9.4.5 from ghcup
cabal
emacs+lsp-mode
HLS v2.2.0.0 from ghcup

### Steps to reproduce

Write some code with an undefined variable, e.g.:
```haskell
foo :: Int -> Int
foo x = plus x (plus x x)
```

Invoke the code action `Define plus...`

Result:
```haskell
plus :: Int -> t0 -> Int Suggested fix: Perhaps use ‘mplus’ (imported from Control.Monad.Reader)
plus = _
```

### Expected behaviour

The suggestion should not be in the type signature, possibly it could be a comment before/after the type signature.

### Actual behaviour

Suggestion in the signature, causing a parse error.

michaelpj:
Looks like a bug in the regex for parsing stuff out of the GHC error message...

jhrcek:
Stumbled upon this while checking for some easy bugs to fix.
It seems the problem is that the bug is coming from [this regex](https://github.com/haskell/haskell-language-server/blob/d6553e001f5134401aef1c676817211947740583/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs#L47) and the fact that 
- the regex works on pre-processed error message in which spaces have been unified (that is replace every sequence of whitespace (including newlines) in the error with a single space).
- it seems that recent GHCs don't contain the "bulletpoint" characters that the regex uses to detect the end of type signature.

This shouldn't be that hard to fix, but I wonder if it's worth trying to make the regexes more robust, or try to use the new structured erros for this (which were only introduced in recent GHCs, something like 9.6+?)

@michaelpj any opinions?

michaelpj:
I think using the structured errors is a whole project, I'm planning to propose it as a GSoC project. So probably not a quick fix. Whereas fixing the regex is a patch, but at least should be quick. What would be extremely helpful in any case would be writing a test that shows the problem!

jhrcek:
Fixed in https://github.com/haskell/haskell-language-server/pull/4016 so this can be closed.

jameshaydon:
Thanks!

PR #4016, opened 2024-01-24
refactor plugin: add reproducer and fix for #3795

jhrcek:
- adding reproducer for https://github.com/haskell/haskell-language-server/issues/3795
- fixing the bug
- fixing bunch of warnings in refactor plugin tests

### DIFF
--- a/plugins/hls-refactor-plugin/hls-refactor-plugin.cabal
+++ b/plugins/hls-refactor-plugin/hls-refactor-plugin.cabal
@@ -102,7 +102,7 @@ test-suite tests
   hs-source-dirs:   test
   main-is:          Main.hs
   other-modules:    Test.AddArgument
-  ghc-options:      -O0 -threaded -rtsopts -with-rtsopts=-N -Wunused-imports
+  ghc-options:      -O0 -threaded -rtsopts -with-rtsopts=-N -Wno-name-shadowing
   build-depends:
     , base
     , filepath
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
@@ -44,7 +44,10 @@ matchVariableNotInScope message
   | otherwise = Nothing
   where
     matchVariableNotInScopeTyped message
-      | Just [name, typ] <- matchRegexUnifySpaces message "Variable not in scope: ([^ ]+) :: ([^*•]+)" =
+      | Just [name, typ0] <- matchRegexUnifySpaces message "Variable not in scope: ([^ ]+) :: ([^*•]+)"
+      , -- When some name in scope is similar to not-in-scope variable, the type is followed by
+        -- "Suggested fix: Perhaps use ..."
+        typ:_ <- T.splitOn " Suggested fix:" typ0 =
           Just (name, typ)
       | otherwise = Nothing
     matchVariableNotInScopeUntyped message
--- a/plugins/hls-refactor-plugin/test/Main.hs
+++ b/plugins/hls-refactor-plugin/test/Main.hs
@@ -9,7 +9,7 @@
 {-# LANGUAGE PolyKinds             #-}
 {-# LANGUAGE RecordWildCards       #-}
 {-# LANGUAGE TypeOperators         #-}
-{-# OPTIONS_GHC -Wno-deprecations -Wno-unticked-promoted-constructors #-}
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
 
 module Main
   ( main
@@ -33,9 +33,7 @@ import           Ide.Types
 import qualified Language.LSP.Protocol.Lens               as L
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types              hiding
-                                                          (SemanticTokenAbsolute (length, line),
-                                                           SemanticTokenRelative (length),
-                                                           SemanticTokensEdit (_start),
+                                                          (SemanticTokensEdit (_start),
                                                            mkRange)
 import           Language.LSP.Test
 import           System.Directory
@@ -81,6 +79,7 @@ tests =
   , completionTests
   ]
 
+initializeTests :: TestTree
 initializeTests = withResource acquire release tests
   where
     tests :: IO (TResponseMessage Method_Initialize) -> TestTree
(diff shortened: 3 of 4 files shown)


plugins/hls-refactor-plugin/hls-refactor-plugin.cabal
```diff
@@ -102,7 +102,7 @@ test-suite tests
   hs-source-dirs:   test
   main-is:          Main.hs
   other-modules:    Test.AddArgument
-  ghc-options:      -O0 -threaded -rtsopts -with-rtsopts=-N -Wunused-imports
+  ghc-options:      -O0 -threaded -rtsopts -with-rtsopts=-N -Wno-name-shadowing
```
jhrcek, line 105: -Wunused-imports is already part of -Wall imported via `import: warnings`.
On the other hand these tests rely on shadowing a lot, so disabling this warning.

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -1782,7 +1786,7 @@ suggestAddRecordFieldImportTests = testGroup "suggest imports of record fields w
       doc <- createDoc "Test.hs" "haskell" before
       waitForProgressDone
       _ <- waitForDiagnostics
-      let defLine = fromIntegral $ 1 + 2
+      let defLine = 3
```
fendor, line 1789: :heart:

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -640,7 +639,10 @@ renameActionTests = testGroup "rename actions"
       doc <- createDoc "Testing.hs" "haskell" content
       _ <- waitForDiagnostics
       actionsOrCommands <- getCodeActions doc (Range (Position 3 12) (Position 3 20))
-      [fixTypo] <- pure [action | InR action@CodeAction{ _title = actionTitle } <- actionsOrCommands, "monus" `T.isInfixOf` actionTitle , "Replace" `T.isInfixOf` actionTitle]
+      [fixTypo] <- pure [action | InR action@CodeAction{ _title = actionTitle } <- actionsOrCommands
+                                , "monus" `T.isInfixOf` actionTitle
+                                , "Replace" `T.isInfixOf` actionTitle
+                        ]
```
fendor, line 645: Perhaps it would make sense to introduce a `findCodeActionsByInfix` as well? Could simplify the code in three locations.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
```diff
@@ -44,7 +44,10 @@ matchVariableNotInScope message
   | otherwise = Nothing
   where
     matchVariableNotInScopeTyped message
-      | Just [name, typ] <- matchRegexUnifySpaces message "Variable not in scope: ([^ ]+) :: ([^*•]+)" =
+      | Just [name, typ0] <- matchRegexUnifySpaces message "Variable not in scope: ([^ ]+) :: ([^*•]+)"
```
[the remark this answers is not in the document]
fendor: Please add this to the comment above so we know what kind of regex we want to match.

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -659,9 +661,11 @@ renameActionTests = testGroup "rename actions"
             , "foo = 'bread"
             ]
       doc <- createDoc "Testing.hs" "haskell" content
-      diags <- waitForDiagnostics
+      _ <- waitForDiagnostics
       actionsOrCommands <- getCodeActions doc (Range (Position 4 6) (Position 4 12))
-      [fixTypo] <- pure [action | InR action@CodeAction{ _title = actionTitle } <- actionsOrCommands, "break" `T.isInfixOf` actionTitle ]
+      [fixTypo] <- pure [action | InR action@CodeAction{ _title = actionTitle } <- actionsOrCommands
+                                , "break" `T.isInfixOf` actionTitle
+                        ]
```
fendor, line 668: `findCodeActionsByInfix`?

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -776,9 +780,9 @@ typeWildCardActionTests = testGroup "type wildcard actions"
         doc <- createDoc "Testing.hs" "haskell" content
         _ <- waitForDiagnostics
         actionsOrCommands <- getAllCodeActions doc
-        let [addSignature] = [action | InR action@CodeAction { _title = actionTitle } <- actionsOrCommands
-                                    , "Use type signature" `T.isInfixOf` actionTitle
-                            ]
+        [addSignature] <- pure [action | InR action@CodeAction { _title = actionTitle } <- actionsOrCommands
+                                       , "Use type signature" `T.isInfixOf` actionTitle
+                               ]
```
fendor, line 785: `findCodeActionsByInfix`?

fendor (changes requested):
Couple of nitpicks, PR looks good to me, thanks for fixing the bug!

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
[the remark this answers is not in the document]
jhrcek: Ok, comment added.

plugins/hls-refactor-plugin/test/Main.hs
jhrcek: It's not so easy, because in this case the two isInfixOf are combined as if by &&.
Whereas the findCodeActions' (used to define findCodeActionsByPrefix) is effectively using || when multiple "search" strings are specified.

plugins/hls-refactor-plugin/test/Main.hs
jhrcek: This case is meaningfuly different - here we're not specifying range in which actions should be find, but getAllCodeActions. So the potential for reuse of findCodeActionsByInfix.

jhrcek: There is some potential for more reuse by defining better helper functions are reusing them more widely, but findCodeActionsByInfix similar to existing "ByPrefix" wouldn't be reusable because
- the place that has 2 infixes needs && but the helper function does ||
- one of the places uses getAllCodeActions and doesn't specify Range
- so I'd prefer not introducing it in this PR.

plugins/hls-refactor-plugin/test/Main.hs
fendor: Ah, I did not see that! Good point, thanks!

fendor (approved):

PR #4018, opened 2024-01-26
update hlint to 3.8 and prevent linting on testdata dir

soulomoon:
This is a following up of #3901

I dig a bit into hlint. 
1. We can only ignore-glob in commandline. config file missing the options for ignore glob.
2. hlint 3.8 can ignore files more correctly.

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -15,7 +15,7 @@ jobs:
     - name: 'Installing'
       uses: rwe/actions-hlint-setup@v1
       with:
-        version: '3.6.1'
+        version: '3.8.0'
 
     - name: 'Checking code'
       uses: rwe/actions-hlint-run@v2
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -20,7 +20,7 @@ jobs:
     - name: 'Checking code'
       uses: rwe/actions-hlint-run@v2
       with:
-        hlint-bin: "hlint --git -j4 --with-group=extra --ignore-glob='**/testdata/**' --ignore-glob='**/test/data/**'"
+        hlint-bin: "hlint --version && hlint --git -j4 --with-group=extra --ignore-glob='**/testdata/**' --ignore-glob='**/test/data/**'"
         fail-on: error
         path: .
 
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -20,7 +20,7 @@ jobs:
     - name: 'Checking code'
       uses: rwe/actions-hlint-run@v2
       with:
-        hlint-bin: "hlint --with-group=extra --ignore-glob='**/testdata/**' --ignore-glob='**/test/data/**'"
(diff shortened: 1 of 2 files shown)


soulomoon:
`hlint --git -j4 --with-group=extra --ignore-glob='**/testdata/**' --ignore-glob='**/test/data/**' -j --json -- . > hlint.result.txt` runs locally ignoring correctly but not in the githubflow.

.github/workflows/hlint.yml
```diff
@@ -15,12 +15,12 @@ jobs:
...
       uses: rwe/actions-hlint-setup@v1
       with:
-        version: '3.6.1'
+        version: '3.8'
 
     - name: 'Checking code'
       uses: rwe/actions-hlint-run@v2
       with:
-        hlint-bin: "hlint --with-group=extra"
+        hlint-bin: "hlint --with-group=extra --ignore-glob=**/testdata/** --ignore-glob=**/test/data/**"
```
michaelpj, line 23: Can you leave a comment with a link to the hlint issue so we know to move this back into the config file when it gets fixed?

soulomoon: I dig into the hlint source code and found this. There is no option to configure the ignore glob, 0 0. Also no issue in hlint about this.
```
        allowFields v ["lhs","rhs","note","name","side"]
        let hintRuleScope = mempty
        pure $
          Left HintRule {hintRuleSeverity = severity, hintRuleLHS = extendInstances lhs, hintRuleRHS = extendInstances rhs, ..}
            : [Right $ Classify severity hintRuleName "" "" | not isBuiltin]
     else do
        names <- parseFieldOpt "name" v >>= maybe (pure []) parseArrayString
        within <- parseFieldOpt "within" v >>= maybe (pure [("","")]) (parseArray >=> concatMapM parseWithin)
        pure [Right $ Classify severity n a b | (a,b) <- within, n <- ["" | null names] ++ names]
```

soulomoon: `within` field is just about `severity`, which we mistakenly thought was about glob.

fendor (approved):

.github/workflows/hlint.yml
soulomoon: The only source code I see about ignoring is 
```
        ignore = (toPredicate $ cmdIgnoreGlob cmd)
        toPredicate :: [FilePattern] -> FilePath -> Bool
        toPredicate [] = const False
        toPredicate globs = \x -> not $ null $ m [((), cleanup x)]
            where m = matchMany (map ((),) globs)
```
where cmdIgnoreGlob is only picked up by commandline arg

michaelpj: I mean, didn't you do https://github.com/ndmitchell/hlint/pull/1554 ? So we can fix put this back once that is released?

soulomoon: Oh that one is about apply the ignoring glob correctly, it is already included in the hlint 3.8. 
Before that, even with commandline ignoring glob, we won't be ignoring some files.

soulomoon: https://github.com/ndmitchell/hlint/pull/1554 is the reason we need to bump up to 3.8

PR #4019, opened 2024-01-26
Fix -Wall and -Wunused-packages in hlint plugin

jhrcek:


fendor (approved):

haskell-language-server.cabal
```diff
@@ -142,7 +142,7 @@ flag cabal
 
 common cabal
   if flag(cabal)
-    build-depends: haskell-language-server:hls-cabal-plugin 
+    build-depends: haskell-language-server:hls-cabal-plugin
```
jhrcek, line 145: Sorry for the noise, but I have editor setup to remove trailing whitespace. I don't think it hurts to clean these up as part of this PR.

PR #4020, opened 2024-01-27
fix isClassNodeIdentifier in hls-class-plugin

### LINKED ISSUE #3942: `DeriveAnyClass` causes "Internal Error: initTcWithGbl failed"
When `DeriveAnyClass` is enabled, attempting to "hover" (inspect) a `deriving A` where `A` is a class with at least one method will cause HLS to have an internal error. 

### Your environment

Which OS do you use?
`Linux 6.7.0, NixOS, 24.05 (Uakari), 24.05.20240108.317484b`

Which version of GHC do you use, and how did you install it?
`Version 9.4.8, installed via nixpkgs`

`N/A`

`VS Codium+vscode-haskell`

Which version of HLS do you use, and how did you install it?
`haskell-language-server version: 2.4.0.0 (GHC: 9.4.8), nixpkgs`

`No`

### Steps to reproduce

```haskell
{-# LANGUAGE Haskell98 #-}
{-# LANGUAGE DeriveAnyClass #-}

module Main where

class C a where
  foo :: a -> Int

newtype Foo = MkFoo Int
  deriving Num

newtype Bar a = MkBar a
  deriving C

main :: IO ()
main = return ()
```

### Expected behaviour

The code above should only result in a `missing-methods` warning on `deriving Num` and `deriving C`

### Actual behaviour

Hovering over `Num` gives you the issue
```
• No explicit implementation for
    ‘+’, ‘*’, ‘abs’, ‘signum’, ‘fromInteger’, and (either ‘negate’
```
And over `C`
```
• No explicit implementation for
    ‘foo’
```

This is expected, however, when hovering over either HLS also gives an error condition pop-up every time they are hovered:
```
Error condition, please check your setup and/or the 
issue tracker: class: Internal Error: initTcWithGbl failed
```

soulomoon:
Partially fix #3942, by handling isClassNodeIdentifier correctly.


plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -198,8 +198,9 @@ codeAction recorder state plId (CodeActionParams _ _ docId _ context) = do
                         _ -> fail "Ide.Plugin.Class.findClassFromIdentifier"
         findClassFromIdentifier _ (Left _) = throwError (PluginInternalError "Ide.Plugin.Class.findClassIdentifier")
 
-isClassNodeIdentifier :: IdentifierDetails a -> Bool
-isClassNodeIdentifier ident = (isNothing . identType) ident && Use `Set.member` identInfo ident
+isClassNodeIdentifier :: Identifier -> IdentifierDetails a -> Bool
```
jhrcek, line 201: Were you able to reproduce the issue? It would be great (and shouldn't be that hard) to add a test that reproduces the original issue and confirms this fix works.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -198,8 +198,9 @@ codeAction recorder state plId (CodeActionParams _ _ docId _ context) = do
...
+isClassNodeIdentifier (Right i) ident  | 'C':':':_ <- unpackFS $ occNameFS $ occName i = (isNothing . identType) ident && Use `Set.member` identInfo ident
```
jhrcek, line 202: Maybe you could use something like: https://hackage.haskell.org/package/ghc-9.8.1/docs/src/GHC.Types.Name.Occurrence.html#isTcOcc on the occName to check if it's a name of a type class, rather than all the unpacking..

soulomoon: I do not see there is a isTcOcc equivalent of checking data constructor for class. 🤔。eventhough DataName appears so, but it should be for all data constructors and not only data contructors for class

soulomoon:
Thank you for your reviews @jhrcek, I have made some responses.

jhrcek:
First of all @soulomoon I appreciate your attempts to fix this issue.
Next I must say that I'm only a newcomer to this codebase. I'm trying to understand how it works to be able to hopefully contribute some fixes in the coming months. With that out of the way..

I tried reproducing https://github.com/haskell/haskell-language-server/issues/3942 with hls 2.4.0.0 (installed via ghcup) and I was able to reproduce it. Then I tried with hls built from master and was still able to reproduce the issue.
Finally I tried the fix in this PR (I'm doing this by setting vscode-haskell extension's hls path to point to haskell-language-server binary in cabal build directory) and I'm still getting that same error, so I don't think this PR will resolve the issue on its own.

Screenshot from 2024-01-28 05-12-18

It's important to understand how [DeriveAnyClass](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/derive_any_class.html?highlight=deriveanyclass#extension-DeriveAnyClass) works:
>  The compiler will simply generate an instance declaration with no explicitly-defined methods.

In the case of `deriving Num` (or more explicitly `deriving anyclass Num` if you use `DerivingStrategies`) ghc creates an empty instance like this
```haskell
instance Num Foo where
-- no methods here
```
This is usually only useful for classes that have default implementation of methods (and is often used with Generics).
But in the case of Num it is not useful, because Num has bunch of methods without default implementation.
So you get the warning "no explicit implementation for ..." (this works as designed).

What this PR seems to be doing is trying to fix the code called on line 105 here:
https://github.com/haskell/haskell-language-server/blob/4788bfca11c366a8d184da2d13f12302429abd57/plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs#L105-L106
I don't think that's correct, because the problem is not that we're not able to find usage of the class Name (there is one usage within `deriving Num` and the code finds it).

But (I think!) the actual error is thrown from the code on line 106 - because it cannot find the empty instance declaration (there is none in the source code - only one generated by the compiler and you cannot add any methods to that).

I don't know how to fix the error, but I don't think this PR fixes it.

jhrcek:
Ooops, it turns out I was testing the local builds incorrectly.
I was mistakenly pasting path to  hls executable into "GHCUP path" instead of "hls path" field in vscode :facepalm: so all the time I was using hls built from master 

I apologize for the confabulated attempts at explanations why this PR doesn't work - this PR  actually DOES prevent the error from being thrown in UI. Though I must admin I don't understand HOW the fix makes the error go away :smile:

soulomoon:
The error happened because his-class-plugin wants a data constructor for a class. While the isClassIdentifier failed to provide the needed check for whether an identifier is actually a data constructor for a class or not. This Pr fix this situation. Hence, the error goes away. (Data constructor for a type class is generate by ghc, it is the way ghc handle type classes)

michaelpj:
Can we add an example from the ticket as a test to ensure this keeps working?

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
michaelpj: This is a little obscure - worth a comment. Perhaps @wz1000 can verify if this is indeed the best way.

soulomoon:
> Can we add an example from the ticket as a test to ensure this keeps working?

See https://github.com/haskell/haskell-language-server/pull/4020#discussion_r1468582794
I fail to construct an example trigger this problem along without triggering another problem that is not fixed in this pr.

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
michaelpj: nag @wz1000

soulomoon:
I managed to come up with a test that focus on the `error` by catching it.
and added a few comments.
See if this look good? @jhrcek @michaelpj

plugins/hls-class-plugin/test/Main.hs
```diff
@@ -120,6 +122,17 @@ codeLensTests = testGroup
             doc <- openDoc "TH.hs" "haskell"
             lens <- getAndResolveCodeLenses doc
             liftIO $ length lens @?= 0
+    , testCase "Don not construct error action!, Ticket3942one" $ do
```
michaelpj, line 125:
```suggestion
    , testCase "Do not construct error action!, Ticket3942one" $ do
```

michaelpj (approved):
Nice!

PR #4021, opened 2024-01-28
auto-approve Dependabot PRs

peterbecich:
Suggestion to auto-approve Dependabot PRs

Copied from:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#approve-a-pull-request

all Dependabot PRs:
https://github.com/haskell/haskell-language-server/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+

This will not auto-merge Dependabot PRs; that's documented here: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request

### DIFF
--- a/.github/workflows/dependabot-auto-approve.yml
+++ b/.github/workflows/dependabot-auto-approve.yml
@@ -1,0 +1,23 @@
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#approve-a-pull-request
+
+name: Dependabot auto-approve
+on: pull_request
+
+permissions:
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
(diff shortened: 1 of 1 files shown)


jhrcek:
Just a heads up that I recently saw some version bump reverts recently, so it might not be good idea to always "blindly" accept changes to latest version.
Example
https://github.com/haskell/haskell-language-server/commit/01413e19e9cd515196dd329793f9bd2235c6d1e0

Maybe @wz1000 can comment in on this PR too?

peterbecich:
Good point, another example https://github.com/haskell/haskell-language-server/pull/3815#issuecomment-1733697453

wz1000:
yes, we can't make any bumps that require upgrading node as the node 20 binaries it pulls in don't work on distributions with older GLIBC versions, which we need to build release binaries.

michaelpj:
Yeah, in normal times this would be useful, but at the moment we are a bit blocked. Perhaps we can explicitly tell dependabot not to update those things for now?

@wz1000 do we have any way to ever get past this problem?

wz1000:
I don't know, perhaps we can convince it to get nodejs binaries from a different source where they are compiled with older GLIBC support.

Issue #3174, opened 2022-09-15
Stan plugin recommends StrictData pragma even when it is in scope


lsmor:
### Your environment

Which OS do you use? 
Ubuntu 20.0.4 Focal
Which version of GHC do you use and how did you install it?
`ghc-8.10.7`  with `ghcup`
How is your project built (alternative: link to the project)?
with cabal 

Which LSP client (editor/plugin) do you use?
VsCode with haskell extension
Which version of HLS do you use and how did you install it?
`hls-1.8.0` I did install it with `ghcup` manually pointing to debian bindist, instead of default fedora as explain in [this](https://github.com/haskell/haskell-language-server/issues/3160#issuecomment-1247230087) comment 
Have you configured HLS in any way (especially: a `hie.yaml` file)?
My hie.yaml is just
```
cradle:
  cabal:
```
### Steps to reproduce
on a normal cabal project write a module

```
{-# LANGUAGE StrictData #-}

module MyModule where

data A = A {field :: Int, field2 :: String}
```

### Expected behaviour

No error or warnings should happen with respect to strict fields

### Actual behaviour

Stan underlies the field's accessors and recommends to add the StrictData pragma. Notice that Stan also recommends adding a bang pattern, and if you do that, then the warning disappears. 

### Debug information

This is the last output of `hls`. It seems that StrictData pragma is reconigze within the module, so I guess the problem is completely up to stan plugin. This is double inconviniently because stan can't be disable or at leat I haven't found the way either: #3157 

```
[Trace - 19:26:53] Received notification 'window/logMessage'.
Using extensions for  NormalizedFilePath "/home/luis/Proyectos/<path/to/module>": [ MonomorphismRestriction
                                                                                  , RelaxedPolyRec
                                                                                  , ForeignFunctionInterface
                                                                                  , ImplicitPrelude
                                                                                  , DoAndIfThenElse
                                                                                  , EmptyDataDecls
                                                                                  , PatternGuards
                                                                                  , DatatypeContexts
                                                                                  , TraditionalRecordSyntax
                                                                                  , EmptyCase
                                                                                  , StrictData
                                                                                  , StarIsType
                                                                                  , CUSKs ]
[Trace - 19:26:53] Received notification 'window/logMessage'.
[Info  - 19:26:53] Typechecking reverse dependencies for NormalizedFilePath "//home/luis/Proyectos/<path/to/module>": [  ]
[Trace - 19:26:53] Received notification 'window/logMessage'.
[...]

PR #3914, opened 2023-12-18
Use stan config files for stan plugin (#3904)

0rphee:
Solves #3904 

I didn't have a clear idea of how it would be best to get the .cabal file for the current file, but the way I did it seems to work fine, it is not critical for stan anyways. Please tell me if there's a better way, and whether the new export of the function from `ghcide/session-loader/Development/IDE/Session/Implicit.hs` is ok.

I also added a test case, in my machine it takes around 5-6s.

I'm not aware if it is possible to load the config just one time for each file, but it should be more appropriate than loading it with every change to the file.

PR #4023, opened 2024-01-29
Use relative file paths for HIE files and Stan's config maps

keithfancher:
Stan expects relative paths. Without this change, file names won't map correctly to their associated language extension data, which means no enabled extensions will be detected. This causes annoying false positives with, e.g., the `StrictData` extension. (See issue #3174.)

Note that the Stan plugin appears to be disabled by default again, so if you'd like to test this change yourself you'll have to enable it in your HLS config.

I've tested this locally and it seems to be working well. Stan is once again respecting both extensions enabled in `.cabal` files and in `LANGUAGE` pragmas.

I've also done a bit of debugging in the CLI version of Stan (which doesn't have this bug), and verified that the file paths in *all* the associated maps are in fact relative.

I'm new-ish to Haskell and I've never contributed to HLS or any "real" Haskell project before, so please let me know if I'm doing something wacky/bad/etc here.

Thanks!


plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -164,24 +164,25 @@ rules recorder plId = do
...
                       logWith recorder Development.IDE.Warning (LogWarnConf es)
                       pure (Map.empty,
                             HM.fromList [(LSP.fromNormalizedFilePath file, inspectionsIds)],
                             [])
                   ResultL warnings stanConfig -> do
-                      let currentHSAbs = fromNormalizedFilePath file -- hie_hs_file hie
-                      currentHSRel <- liftIO $ makeRelativeToCurrentDirectory currentHSAbs
-                      cabalExtensionsMap <- liftIO $ createCabalExtensionsMap isLoud (stanArgsCabalFilePath stanArgs) [hie]
-
-                      -- Files (keys) in checksMap need to have an absolute path for the analysis, but applyConfig needs to receive relative
```
keithfancher, line 178: > Files (keys) in checksMap need to have an absolute path for the analysis

This didn't _seem_ to be the case, when I looked?

I think this logic was added in https://github.com/haskell/haskell-language-server/pull/3914 ... perhaps @0rphee could confirm?

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -164,24 +164,25 @@ rules recorder plId = do
...
-                      -- filepaths to apply the config, because the toml config has relative paths. Stan itself seems to work only in terms of relative paths.
-                      let checksMap = HM.mapKeys (const currentHSAbs) $ applyConfig [currentHSRel] stanConfig
-
-                      let analysis = runAnalysis cabalExtensionsMap checksMap (configIgnored stanConfig) [hie]
+                      -- A Map from *relative* file paths (just one, in this case) to language extension info.
+                      cabalExtensionsMap <- liftIO $ createCabalExtensionsMap isLoud (stanArgsCabalFilePath stanArgs) [hieRelative]
+                      -- HashMap of *relative* file paths to info about enabled checks for those file paths.
+                      let checksMap = applyConfig [relativeHsFilePath] stanConfig
```
keithfancher, line 182: I checked both `cabalExtensionsMap` and `checksMap` in the CLI version of Stan, and both had relative paths as keys. The HIE file path was also relative.

This seemed to have been causing the miss in the Map lookup. (e.g. looking for `/path/to/some/file` when the actual key was `some/file`.)

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
0rphee: Yes this was added in that PR. As I remember it, If I used relative paths, it didn't work, because of how an internal function works in `runAnalysis`, that checks if an absolute path to a file, begins with another filepath, (`currentHSAbs` I think?).

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
0rphee: This seems to be the reason why I initially had issues with relative/absolute paths. Thanks for the investigation!

0rphee:
Everything seems fine, thanks! I'll test this later today or tomorrow. Just in case, I think it would be appropriate to add a test case to guarantee that stan uses pragmas and flags from  `.cabal` files!

keithfancher:
> Everything seems fine, thanks! I'll test this later today or tomorrow. 

Awesome, thank you!

> Just in case, I think it would be appropriate to add a test case to guarantee that stan uses pragmas and flags from `.cabal` files!

Good call! I'll take a crack at adding test cases.

michaelpj:
Maybe make an upstream issue? I know stan is now at least a little bit maintained. This is quite a footgun!

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
```diff
@@ -164,24 +164,25 @@ rules recorder plId = do
...
+              -- making its path relative, the file name(s) won't line up with the associated Map keys.
+              relativeHsFilePath <- liftIO $ makeRelativeToCurrentDirectory $ fromNormalizedFilePath file
+              let hieRelative = hie{hie_hs_file=relativeHsFilePath}
+
               (cabalExtensionsMap, checksMap, confIgnored) <- case configTrial of
                   FiascoL es -> do
                       logWith recorder Development.IDE.Warning (LogWarnConf es)
                       pure (Map.empty,
                             HM.fromList [(LSP.fromNormalizedFilePath file, inspectionsIds)],
                             [])
```
keithfancher, line 177: Oh, I almost forgot to ask: I wasn't sure about desired behavior if we hit this branch.

Of course I could change this file path to `relativeHsFilePath` and even build a `cabalExtensionsMap`, I think... but when I checked the corresponding flow in [Stan-proper](https://github.com/kowainik/stan/blob/main/src/Stan.hs#L90), they seem to bail entirely in the case of a `Fiasco` here. (Assuming I'm reading the code correctly.)

keithfancher:
> Maybe make an upstream issue? I know stan is now at least a little bit maintained. This is quite a footgun!

Also a good call! It seems like it'd be a quick fix for Stan to normalize paths coming in, make them relative if that's what it expects. Or something like that...

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
0rphee: Yes, you read the code correctly. Stan stops completely with an error. 

We could use this opportunity to define what the plugin should do. When I added the capability for reading the config, I left the previous behavior (stan runs without config) on the branch where something is wrong here, however we could instead stop the analysis completely.

The latter could be preferred in the hypothetical case where there is a bad config, and the user would just prefer to ignore possibly annoying suggestions (that would be ignored taking into account a well formed .stan.toml).

However, I'm not partial to any choice. Any suggestions @michaelpj?

0rphee:
The path-issue seems to have arose because the original code for the plugin always used the absolute path. It never occurred to me to change that. It would definitely be nice to normalize the paths from stan itself, or at least add a bit of documentation there!

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
keithfancher: Ah, I see. Yeah, I could see a good argument for going either way.

As you say, it could be annoying for a user to see a bunch of irrelevant diagnostics. But on the other hand, it could be a bad experience for the plugin to just fail silently and for the user to see nothing at all. They could break a config and they might not even notice. (Talking through it, that latter case might be worse in the end?)

In any case, if the intended previous behavior was for it to still run sans config, I'd be fine just defaulting to that. (In which case, I'd have to make some minor tweaks to the code, fix the mappings there as well.)

keithfancher:
Just a quick update here: test cases should be good to go. I've verified that the new tests fail with the previous version of the plugin. Should definitely catch this issue or others like it moving forward :tada: 

Only only question left to clear up (see https://github.com/haskell/haskell-language-server/pull/4023#discussion_r1470264347 ), and then I think this should be ready for merging!

plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
0rphee: Let's stick to running diagnostics anyway and building the `cabalExtensionsMap`, so that it gets taken into account!

keithfancher: > Let's stick to running diagnostics anyway and building the cabalExtensionsMap, so that it gets taken into account!

Done! Pushed up https://github.com/haskell/haskell-language-server/pull/4023/commits/e6a85206dccc08fec9cfe907e308987b52c3b8d5 to take care of that.

Interestingly, I couldn't "organically" hit that code path from the plugin no matter what I tried -- I wound up just forcing the fiasco for testing purposes.

Seems like Stan will always either 1) throw an IO exception (in the case of a TOML parse error), or 2) recover and return some variety of partial/default config. But that path works properly now, just in case that should ever change :smile: 

On the plus side (?!), the "invalid config" case is extremely loud and obvious, thanks to its exception-throwing. Basically brings everything screeching to a halt. Here's how it looks for me in neovim:

hls-stan-toml-parse-error

0rphee (approved):

0rphee:
Everything is fine by me, thanks!

keithfancher:
How odd, these seem like [legitimate (Windows-specific?) test failures](https://github.com/haskell/haskell-language-server/actions/runs/7746229874/job/21124002278?pr=4023). But the same tests passed after I merged `master` in ~16 hours ago. It does look there have been some test suite changes merged in today...

michaelpj:
Yes, CI was a bit broken before. I would guess that those are real failures, especially since you're doing some things with paths!

keithfancher:
>  I would guess that those are real failures, especially since you're doing some things with paths!

Yes indeed! I scared up a Windows VM and did some debugging, turned out to be quite a strange issue. The TL;DR is: Stan builds its extension map by sticking the `hs-source-dirs` value *verbatim* onto the module names. This means that if your `.cabal` file contains `/` as directory separators, Stan's extension map is broken in Windows.

Good new is: the HLS plugin code is fine, and the tests are sound. And in fact, the new tests uncovered this very real issue! :smile:  I just had to work around it by collapsing my test data directory structure.

I don't have a ton of context for how `.cabal` files are supposed to work (mine have all been auto-generated so far!), but this *seems* like a bug to me. Likely in the kowainik `extension` package, probably around [here](https://github.com/kowainik/extensions/blob/main/src/Extensions/Cabal.hs#L150).

Or maybe this is a known thing? I do see a `#if MIN_VERSION_Cabal(3,6,0)` check... maybe it works fine with higher versions of Cabal? I'm happy to file an issue on the `extensions` repo though, if this is a legit bug.

michaelpj:
Great, do please open an upstream issue, even if it's not a heavily maintained package. It really helps if these things come up again!

keithfancher:
> Maybe make an upstream issue? I know stan is now at least a little bit maintained. This is quite a footgun!

> Great, do please open an upstream issue, even if it's not a heavily maintained package. It really helps if these things come up again!

Filed issues for both! For Stan and `extensions`, respectively. They're linked above thanks to Github magic :smile:

PR #4024, opened 2024-01-29
Fix various issues

michaelpj:
The first thing here is quite bad - my mistake! Unfortunately I got it all tangled in some other changes, but I think they're uncontroversial. 

- Make sure that we are always referring to the local libraries and not the published Hackage packages! - This needs a `cabal-version` bump
- Make sure every component imports the default stanzas
- Add a `defaults` stanza for `default-language`
- Add a `tests-default` stanza for making sure test suites are built threaded
- Remove unnecessary warning-related `ghc-options`
- Use the `pedantic` stanza instead of manual checks for `pedantic`


michaelpj:
cc also @jhrcek

haskell-language-server.cabal
```diff
@@ -1914,19 +1840,19 @@ executable haskell-language-server-wrapper
           unix
         , containers
 
-  default-language: Haskell2010
 
 
 test-suite func-test
-  import:             common-deps
+  import:             defaults
+                    , test-defaults
```
jhrcek, line 1847: You can remove "-threaded -rtsopts -with-rtsopts=-N" from func-test now that you're importing test-defaults

haskell-language-server.cabal
```diff
@@ -1992,12 +1917,11 @@ test-suite wrapper-test
   main-is:            Main.hs
 
 benchmark benchmark
-    import:           common-deps
+    import:           defaults, warnings, common-deps
     -- Depends on shake-bench which is unbuildable after this point
     if impl(ghc >= 9.5)
       buildable: False
     type: exitcode-stdio-1.0
-    default-language: Haskell2010
```
jhrcek, line 2000: You can remove -Wall -Wno-name-shadowing from benchmark now that you're importing warnings.

jhrcek (approved):
Two nitpicks, otherwise nice consolidation. I'll update my open PRs once this is merged.

fendor (approved):

michaelpj:
Some `ghcide` tests seem to be failing repeatably, but only on particular platform/GHC version combos. I also can't reproduce it, and I haven't touched ghcide :grimacing:

fendor:
The ghcide tests looks like they suddenly are picking up HLS as the cradle root which is taking forever and probably wrong. I can look into it.

michaelpj:
hooray, green

PR #4025, opened 2024-01-29
Fix documentation/image links

jhrcek:
I built the docs locally and fixed most sphinx warnings.

### DIFF
--- a/docs/contributing/contributing.md
+++ b/docs/contributing/contributing.md
@@ -221,7 +221,7 @@ Run the benchmarks with `cabal bench`.
 
 It should take around 25 minutes and the results will be stored in the `bench-results` folder. To interpret the results, see the comments in the `bench/Main.hs` module.
 
-More details in [bench/README](../../bench/README.md)
+More details in [bench/README](https://github.com/haskell/haskell-language-server/blob/master/bench/README.md)
 
 ### Tracing
 
@@ -233,7 +233,7 @@ Adding support for new editors is fairly easy if the editor already has good sup
 In that case, there will likely be an editor-specific support system for this (like `lsp-mode` for Emacs).
 This will typically provide instructions for how to support new languages.
 
-In some cases you may need to write a small bit of additional client support, or expose a way for the user to set the server's [configuration options](#configuring-haskell-language-server) and
+In some cases you may need to write a small bit of additional client support, or expose a way for the user to set the server's [configuration options](../configuration.md#configuring-haskell-language-server) and
 for them to configure how the server is started.
 
 ## Building the docs
--- a/docs/features.md
+++ b/docs/features.md
@@ -138,7 +138,7 @@ Provided by: `hls-call-hierarchy-plugin`
 
 Shows ingoing and outgoing calls for a function.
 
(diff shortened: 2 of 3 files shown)


docs/features.md
```diff
@@ -138,7 +138,7 @@ Provided by: `hls-call-hierarchy-plugin`
 
 Shows ingoing and outgoing calls for a function.
 
-![Call Hierarchy in VSCode](https://github.com/haskell/haskell-language-server/raw/2857eeece0398e1cd4b2ffb6069b05c4d2308b39/plugins/hls-call-hierarchy-plugin/call-hierarchy-in-vscode.gif)
+![Call Hierarchy in VSCode](../plugins/hls-call-hierarchy-plugin/call-hierarchy-in-vscode.gif)
```
jhrcek, line 141: Didn't test the doc build but most working example images refer to relative paths within the repo, so changing this one to do the same.

docs/contributing/contributing.md
```diff
@@ -221,7 +221,7 @@ Run the benchmarks with `cabal bench`.
 
 It should take around 25 minutes and the results will be stored in the `bench-results` folder. To interpret the results, see the comments in the `bench/Main.hs` module.
 
-More details in [bench/README](../../bench/README.md)
+More details in [bench/README](https://github.com/haskell/haskell-language-server/blob/master/bench/README.md)
```
jhrcek, line 224: These links are referring to stuff outside of docs directory, so sphinx is unable to create proper links.
Rather than trying to change sphinx configuration, i think it's simpler to just link to relevant docs on master branch..?

soulomoon: will it cause a version mismatch though🤔

jhrcek: Let's think through the implications of this change:

1) Before this PR, the links (not images, which are copied during the build!) that point to relative file paths outside of docs directory are not rendered as links:
Example: search for `bench/README` on [this page](https://haskell-language-server.readthedocs.io/en/stable/contributing/contributing.html).

When building the docs by running `make html` in the docs folder you get the following warning:
/home/jhrcek/Devel/github.com/haskell/haskell-language-server/docs/contributing/contributing.md:224: WARNING: Unknown source document '/home/jhrcek/Devel/github.com/haskell/haskell-language-server/bench/README' [myst.xref_missing]

And the resulting html has no links:
Screenshot from 2024-01-30 05-34-52

2) With this PR: the link is renered correctly and points to readme on master:
Screenshot from 2024-01-30 05-40-02

You're right that this is not ideal, because the readme on master can be modified in a way that makes is "incompatible" with the rendered version of the docs or it can be even removed etc. But this is what most of the other links on the page are currently doing (just pointing to master version of the doc on github).

3)The ideal solution would be to replace "master" within those links with "the commit hash of the commit from which the docs were built" which would tie the rendered docs to the state of the repo when they were built. Unfortunately I don't know what's the "canonical" way to do this with sphinx (I can only think of hacks like adding a placeholder in all those links and running `sed` before building with shinx, which feels like hack).

michaelpj (approved):
General direction is great, thanks!

fendor (approved):
LGTM, thanks for taking care of the Docs!

PR #4027, opened 2024-01-30
Reduce the number of ad-hoc helper test functions in refactor plugin tests

jhrcek:
I'd like to add couple of reproducers for refactor plugin issues that annoy me in my daily work.
But the existing tests trigger my OCD.
It seems that everyone who added tests came up with their own set of helper functions and shortcuts :smile: 

This PR shouldn't change the coverage, but it partially reduces the chaos by making the tests reuse smaller number of simple test helper functions.
More improvements could be done, but perfect is the enemy of good..


jhrcek:
Huh, almost no tests were executed :thinking: is this something that's going to be resolved in https://github.com/haskell/haskell-language-server/pull/4024?

fendor (approved):
Changes look good to me, but the CI is confusing me right now.

michaelpj:
> It seems that everyone who added tests came up with their own set of helper functions and shortcuts

It's a huge mess. Tidying that up would be a great blessing!

PR #4028, opened 2024-01-30
Turn off tasty-rerun

michaelpj:


fendor (approved):

Bodigrim:
I think `tasty-rerun` is doing what it's asked for, no? It's just that `cabal test ghcide --test-options="$TEST_OPTS" || cabal test ghcide --test-options="$TEST_OPTS"` is a very strange pattern, especially given that rerun logs are cached.

jhrcek:
I was surprised by this too.
Given that caching is involved, my expectation would be to do
`--rerun-update only` is used in the first run (so that all tests are executed no matter what other branches have chached)
`--rerun-filter filters,exceptions` is used in the second run.

michaelpj:
Yes, this certainly interacts with our slightly idiosyncractic approach to re-running tests. For now I'm just trying to clear the deck so we can get everything green again!

michaelpj:
Somehow, inexplicably, that seems to have got us back to a better place. Could it be that somehow those ghcide tests only pass when run by themselves, which accidentally happens if they fail and then get rerun?

michaelpj:
I don't fully understand, but I think I've restored us to the old state, so I'm going to try and merge this to keep us green.

Issue #3969, opened 2024-01-16
Semantic Tokens: Distinguish (possibly-partially-applied) infix operators from normal functions/variables


konn:
## Is your enhancement request related to a problem? Please describe.

First of all, thank you @soulomoon for implementating fantastic support for semantic tokens! I've long been waiting for this feature :tada:

After the release of 2.6.0.0, I've just tested it and looks really good. I think it will become even better if one can tell (possibly partially-applied) infix operators from ordinary variable/functions semantically.

For example, the image below is what I apply One Monokai + Semantic Tokens to the code fragments containing several `($)`s:

![code-semantic](https://github.com/haskell/haskell-language-server/assets/21860/a3062658-d177-4e67-88c6-d9894c8a120f)

In above, functions being applied and `($)`s are all coloured homogeneously. Without semantic colouring, OTOH, we got:

![code-syntactic](https://github.com/haskell/haskell-language-server/assets/21860/e81b44fb-173c-4259-9333-6226596826ee)

I think the latter case gives more legible highlighting - binary operator shines between functions.

## Describe the solution you'd like

Add something like `TBinaryOperator` constructor to `HsSemanticTokenType`, map its default to `operator` token, and also allowing users to map them to what they like (as we already did for ` haskell.plugin.semanticTokens.config.dataConstructorToken` etc).

## Describe alternatives you've considered

Perhaps, as we have already done to plain symbols, there can be `TBinaryFunction` and `TBinaryVariable`, as the function declaration can take its argument in infix-form.

soulomoon:
Thank you @konn for ideas! I've added it to the tracklist!

michaelpj:
We even have a good LSP token type for this: `operator`.

michaelpj:
Also, what about ` `isSubsetOf` ` vs `+`? Is the former also an operator or should it continue to be a function?

konn:
> We even have a good LSP token type for this: `operator`.

Indeed, as I already suggested in my solution section 😃 

> Also, what about `` `isSubsetOf` `` vs `+`? Is the former also an operator or should it continue to be a function?

Good catch. Both seem making sense, but I prefer treating it as a function.

michaelpj:
Yep. I think you already have support for returning multiple types and picking one.

I note that there is a client capability for supporting overlapping tokens. I guess in principle that means we could return two tokens for the same span? I don't know what that would do.

michaelpj:
I agree

konn:
[...]

PR #4030, opened 2024-01-31
semantic tokens: add infix operator

soulomoon:
Trying to implement #3969
~~I am still thinking by default, we should take infix operator or class method(we have infix class method) or type(we have type operator) as precedent.~~

After discussion, I made a decision, temporally operator have its priority over them by default.


plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
```diff
@@ -66,7 +77,7 @@ tyThingSemantic ty = case ty of
     | isTyVar vid -> Just TTypeVariable
     | isRecordSelector vid -> Just TRecordField
     | isClassOpId vid -> Just TClassMethod
-    | isFunVar vid -> Just TFunction
+    | isFunVar vid -> Just TFunction <> (nameInfixOperator $ getName vid)
```
michaelpj, line 80: Since you're going to merge them all with `<>` later, perhaps `tyThingSemantic` should just return `[HsSemanticTokenType]`, since in general we might allocate several types? That would simplify some things, e.g. here you're implicitly duplicating the fact that `TClassMethod` is preferred over `TFunction` by putting the match for `isClassOpId` first. (So if you ever changed the preference in the Semigroup instance you would still get the old answer!).

So you could do something like:
```
AnId vid -> [TTypeVariable | isTyVar vid] ++ ... ++ [TFunction | isFunVar vid] ++ [TOperator | isSymId vid]
```
and then let the semigroup instance sort it all out. Less important for the other cases where we don't have overlapping token types, though.

plugins/hls-semantic-tokens-plugin/test/testdata/TOperator.hs
```diff
@@ -0,0 +1,12 @@
+module TOperator where
+
+-- imported operator
+go f x = f $ x
+-- operator defined in local module
+($$$$) = b
+x = 1 $$$$ 2
+data a :+: b = Add a b
+-- type take precedence over operator
```
michaelpj, line 9: I actually think people might prefer to see type operators highlighted differently? In "normal" Haskell syntax highlighting they are!

soulomoon: Yeah, I've been thinking about this, may be it better for us to mark all the infix thing as operator.

michaelpj: I feel reasonably sure that we want "operator" to take precedence over "type". I'm still unsure about "operator" vs "class method"

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: yes, and it is going to be more tricky when priorities is configurable. Since we need to make sure the behaviour consistant between information from hieAst and the one from typeThing.

plugins/hls-semantic-tokens-plugin/test/testdata/TOperator.hs
soulomoon: me neither, maybe we can make a pool or something to collect people's opinion on this.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
michaelpj: Well, that should work out okay with the model where we just collect (possibly many) token types and then pick the "best". We'll have to have a "best" function that is derived from the configuration, instead of just the hardcoded one based on the Semigroup instance, but that seems okay?

soulomoon: I agree, we might want to switch to list or set eventually with priorities configurations. Since by then, we might need to add test to ensure the all the results coming from imported or local names would have the same set of token types(instead of the current strategy to have just the same token type).

michaelpj: > test to ensure the all the results coming from imported or local names would have the same set of token types

You mean if I have 

```
module A where

foo = ...
```
```
module B where
import A

bar = foo
```
then `foo` should have the same tokens in both cases? That does seem like a nice property!

plugins/hls-semantic-tokens-plugin/test/testdata/TOperator.hs
soulomoon: Yep, lets stick to the simplest way for now. Make the infix on top of all.
We can sort things out in configuration PR that switching to the `[HsSemanticTokenType]` .

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Mappings.hs
soulomoon: Exactly

michaelpj (approved):

PR #4031, opened 2024-02-01
Fix weird behavior of OPTIONS_GHC completions (fixes #3908)

### LINKED ISSUE #3908: Completion of GHC options in`OPTIONS_GHC` doesn't properly handle `-`s
#### Which version of GHC do you use and how did you install it?

9.8.1 from GHCup

#### Which version of HLS do you use and how did you install it?

2.5.0.0

### Steps to reproduce

```hs
{-# OPTIONS_GHC -ddump-simpl -dno-suppr #-}
```
When typing `-dno-suppr`, select the `-dno-suppress-type-signatures` suggestion.

### Expected behaviour

```hs
{-# OPTIONS_GHC -ddump-simpl -dno-suppress-type-signatures #-}
```

### Actual behaviour

```hs
{-# OPTIONS_GHC -ddump-simpl -dno-dno-suppress-type-signatures #-}
```

jhrcek:
Reproducer for https://github.com/haskell/haskell-language-server/issues/3908

The issue is that getCompletionPrefix on this [line](https://github.com/haskell/haskell-language-server/blob/17aadba75732316edfcc6d01d17841a3f469afab/plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs#L206)  doesn't correctly identify the "already typed prefix of the ghc option".
It thinks the prefix is only the part up to '-' (e.g. in the reproducer it tries to fuzzy match prefix "fo" instead of "dsuppress-constant-fo" against existing flags), which can be also seen in the suggestion dropdown (note only "fo" is highlighted).
Screenshot from 2024-02-01 07-53-49

I'll take some time over weekend to try to fix this. In the meantime I welcome any suggestions as to what would be the proper way to fix this.
1. modify getCompletionPrefix to accept larger prefix (seems too dangerous, as it's used by a lot of code already)?
2. create custom version of "prefix finder" tailored to ghc options (my plan)?

NOTE: although the tests seem to pass, it's due to issue on master due to which most tests are not running - this is being resolved in https://github.com/haskell/haskell-language-server/pull/4028


michaelpj:
`getCompletionPrefix` is mis-named: it finds completion prefixes for _Haskell identifiers_. So yeah, we need a custom one. Maybe we can make it more generic, not sure?

Also note that it's moved to `Logic.hs` in `ghcide`, I'm going to get rid of the one in `lsp` at some point, it should never have been there.

michaelpj:
The use in the cabal plugin is likely also wrong for this reason!

jhrcek:
Here's how it works after the fix in this PR:

Peek 2024-02-04 09-37

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -904,7 +904,7 @@ getCompletionPrefix pos@(Position l c) (VFS.VirtualFile _ _ ropetext) =
             lastMaybe = headMaybe . reverse
 
         -- grab the entire line the cursor is at
-        curLine <- headMaybe $ T.lines $ Rope.toText
+        curLine <- headMaybe $ Rope.lines
```
jhrcek, line 907: This is supposed to be faster than Text.lines, so why not use it here as well..
https://hackage.haskell.org/package/text-rope-0.2/docs/Data-Text-Lines.html#v:lines

fendor:
> The use in the cabal plugin is likely also wrong for this reason!

IIRC, hls-cabal-plugin introduces its own completion prefix function due to `/` in filepaths.

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
michaelpj: Yes! We really need to audit all uses of `Rope.toText`, maybe even give it a hlint hint. It's generally unnecessarily expensive and does work proportional to the length of the document...

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
```diff
@@ -192,30 +195,32 @@ allPragmas =
...
                 result (Just pfix)
                     | "{-# language" `T.isPrefixOf` line
-                    = map buildCompletion
-                        (Fuzzy.simpleFilter (VFS.prefixText pfix) allPragmas)
+                    = map mkLanguagePragmaCompl $
+                        Fuzzy.simpleFilter (VFS.prefixText pfix) allPragmas
                     | "{-# options_ghc" `T.isPrefixOf` line
-                    =  map buildCompletion
-                        (Fuzzy.simpleFilter (VFS.prefixText pfix) flags)
+                    = let flagPrefix = getGhcOptionPrefix cursorPos cnts
```
michaelpj, line 214: The logic here feels a little bad; we're passing in `pfix` here and no longer using it at all. Maybe we can refactor it a bit?

jhrcek: I wanted to do localized change in fear of breaking existing functionality.
But ok, will look into rewriting the whole thing so that it doesn't use VFS.getCompletionPrefix from lsp packae at all. I actually didn't notice there were two versions of that (one in ghcide and one in lsp).

michaelpj: Yes, that's what I meant. I didn't meant to say that the pragmas plugin shouldn't be using the generic `getCompletionPrefix` function (although that's almost certainly true, for the same reasons!), just that the particular branch that you're changing now sits oddly.

> I actually didn't notice there were two versions of that (one in ghcide and one in lsp).

I'm going to kill the `lsp` one, so ignore that.

michaelpj: On reflection, I don't want to block this on doing the refactoring, so I'm happy to leave it up to you whether you want to do that now or not. If not, maybe leave a note explaining the weirdness.

michaelpj (approved):

plugins/hls-pragmas-plugin/src/Ide/Plugin/Pragmas.hs
jhrcek: I'll get back to this and try if I can clean it up. It just needs a bit more focus time that I'm unlikely to get during workdays :smile:

PR #4033, opened 2024-02-01
Use GHC2021

michaelpj:
I think this looks fairly good, cuts down our extension spam a bit.


soulomoon:
I like this idea.

jhrcek (approved):
I like this idea especially in a codebase that's more like an application than like a library.
FYI I resolved merge conflict caused by my PR and pushed to this branch.
Let me know if you find it inappropriate for people to push to other people's branches.

michaelpj:
>  Let me know if you find it inappropriate for people to push to other people's branches.

I think it's generally helpful in the case of resolving conflicts (unless you do it wrong :p ). So thanks!

PR #4037, opened 2024-02-03
Fix -Wunused-imports

jhrcek:
Since -Wall was turned on more universally during recent cabal file consolidations, the build logs are very noisy.
I'd like to fix most of the warnings, doing focused changes per PR to make it easy to review.


plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
```diff
@@ -4,8 +4,7 @@
 {-# LANGUAGE RecordWildCards   #-}
 {-# LANGUAGE ViewPatterns      #-}
 
-{-# OPTIONS_GHC -Wall -Wwarn -fno-warn-type-defaults #-}
-{-# OPTIONS_GHC -Wno-name-shadowing #-}
+{-# OPTIONS_GHC -Wwarn -fno-warn-type-defaults #-}
```
jhrcek, line 7: Removing per-file warnings flags that are enabled for all plugins universally via `import warnings` in haskell-language-server.cabal

michaelpj (approved):
great stuff

michaelpj:
FWIW I wonder if you want to still go per-component, then you can also start including the `pedantic` stanza so we ensure we stay in the good state!

PR #4039, opened 2024-02-03
Set test options via cabal.project

michaelpj:
This means we can set the default properly for people using the project, and also set things more simply in the test workflow.

The option setting in `hls-test-options` also seemed to be simply not working, for reasons I don't fully understand (but it seems like `NumThreads` is a special option.


fendor (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -104,151 +104,147 @@ jobs:
...
-        # run the tests without parallelism, otherwise tasty will attempt to run
-        # all functional test cases simultaneously which causes way too many hls
-        # instances to be spun up for the poor github actions runner to handle
-        #
         # See https://github.com/ocharles/tasty-rerun/issues/22 for why we need
         # to include 'new' in the filters, since many of our test suites are in the
         # same package.
         run: |
-          echo "TEST_OPTS=-j1 --rerun-update --rerun-filter failures,exceptions,new" >> $GITHUB_ENV
+          echo 'test-options: --rerun-update --rerun-filter "failures,exceptions,new"' >> cabal.project.local
```
jhrcek, line 111: Nitpick: isn't this what [cabal configure](https://cabal.readthedocs.io/en/stable/cabal-commands.html#cabal-configure) is for?
e.g.
cabal configure --test-options='--rerun-update --rerun-filter failures,exceptions,new'

creates cabal.project.local with the desired options.

michaelpj: Good point!

.github/workflows/test.yml
```diff
@@ -104,151 +104,147 @@ jobs:
...
-        # run the tests without parallelism, otherwise tasty will attempt to run
-        # all functional test cases simultaneously which causes way too many hls
-        # instances to be spun up for the poor github actions runner to handle
-        #
         # See https://github.com/ocharles/tasty-rerun/issues/22 for why we need
         # to include 'new' in the filters, since many of our test suites are in the
         # same package.
         run: |
-          echo "TEST_OPTS=-j1 --rerun-update --rerun-filter failures,exceptions,new" >> $GITHUB_ENV
+          echo 'cabal configure --test-options="--rerun-update --rerun-filter failures,exceptions,new"' >> cabal.project.local
```
jhrcek, line 111: I mean this, which has the side effect of writing those options into cabal.project.local
```suggestion
          cabal configure --test-options="--rerun-update --rerun-filter failures,exceptions,new"
```

```
$ cabal configure --test-options="--rerun-update --rerun-filter failures,exceptions,new"
$ cat cabal.project.local 
ignore-project: False
test-options: "--rerun-update"
              "--rerun-filter"
              "failures,exceptions,new"
```

michaelpj: That's what I meant to type, just :facepalm:

PR #4043, opened 2024-02-04
Disable caching job with ghc 9.2 on windows

jhrcek:
Not sure if this is desirable, but @michaelpj recently disabled tests for this combination (see [PR](https://github.com/haskell/haskell-language-server/pull/3976/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R81-R83))  and said something like this would be needed to prevent failures of caching job on master.

### DIFF
--- a/.github/workflows/caching.yml
+++ b/.github/workflows/caching.yml
@@ -50,7 +50,6 @@ env:
   cabalBuild: "v2-build --keep-going"
 
 jobs:
-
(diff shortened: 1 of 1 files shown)


michaelpj (approved):

michaelpj:
Thanks!

PR #4047, opened 2024-02-04
Fix warnings in hls-graph, enable pedantic in CI

jhrcek:


hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -5,7 +5,65 @@
 {-# LANGUAGE RecordWildCards    #-}
 {-# LANGUAGE ViewPatterns       #-}
 
-module Development.IDE.Graph.Internal.Types where
+module Development.IDE.Graph.Internal.Types
```
jhrcek, line 8: Why am I adding explicit export list to this module?
I'm aware it's internal. But it exposes UnsafeMkKey + also things like newKey which use global reference to map with all keys.
If we make Key constructor opaque, it should be safe to add  `{-# COMPLETE Key #-}` pragma to get rid of "incomplete pattern match" warnings, because the only way to create new Key values are through newKey smart constructor.

.github/workflows/flags.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           os:  ${{ runner.os }}
 
       - name: Build `hls-graph` with flags
-        run: cabal v2-build hls-graph --flags="embed-files stm-stats"
+        run: cabal v2-build hls-graph --flags="embed-files stm-stats pedantic"
```
michaelpj, line 73: I _think_ this should be implied by the later build with pedantic?

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
michaelpj: I don't have a strong opinion, although it is a rather long export list.

michaelpj (approved):

.github/workflows/flags.yml
jhrcek: Indeed, that's what I expected too (that when you change flags cabal rebuilds all the libraries affected).

But the how do you explain the fact that this flags job passes on master with all those warnings present?

Running these 3 cabal builds locally (after cabal clean) I see that modules from hls-graph are compiled in the first two and the third (`cabal v2-build --flags="pedantic"`) seems to be using hls-graph compiled in the first 2 runs:

```
$ cabal v2-build --flags="pedantic"
...
In order, the following will be built (use -v for more details):
 - ghcide-2.6.0.0 (exe:ghcide-test-preprocessor) (configuration changed)
 - hls-graph-2.6.0.0 (lib) (configuration changed)
 - shake-bench-0.2.0.0 (lib) (first run)
...
Preprocessing executable 'ghcide-test-preprocessor' for ghcide-2.6.0.0..
Building executable 'ghcide-test-preprocessor' for ghcide-2.6.0.0..
Preprocessing library for hls-graph-2.6.0.0..
Building library for hls-graph-2.6.0.0..                                   <<-- the output suggests the individual modules are no longer compiled (it's probably using artifacts from previous 2 builds???
Preprocessing library for shake-bench-0.2.0.0..
Building library for shake-bench-0.2.0.0..
```

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
jhrcek: I could try to make it shorter by splitting off KeyMap and KeySet into separate modules.. How does that sound?

.github/workflows/flags.yml
michaelpj: Weird. I wonder if we could simplify this by setting up a `cabal.project.local` as for the test options and then just doing a simple `cabal build all`?

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
michaelpj: Or `Key` I guess, if that's the thing you want to be opaque. I don't want to derail you too much from your warning-fixing quest, but both of those sound like sensible refactorings to me.

.github/workflows/flags.yml
```diff
@@ -69,14 +69,15 @@ jobs:
           ghc: ${{ matrix.ghc }}
           os:  ${{ runner.os }}
 
-      - name: Build `hls-graph` with flags
-        run: cabal v2-build hls-graph --flags="embed-files stm-stats"
-
-      - name: Build `ghcide` with flags
-        run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe ekg"
+      - name: Configue extra flags for some components
```
jhrcek, line 72: ghc-patched-unboxed-bytecode and bench-exe flags seem to be some leftovers - they don't exist in ghcide.cabal anymore, so I just removed them from here.

The idea behind using cabal configure is that we do only single cabal build with pedantic flags enabled - and don't skip pedantic when building hls-graph and ghcide.

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
jhrcek: Ok, moved Key + KeyMap + KeySet stuff (because they depend on the internals) to separate module and removed explicit export list from Types module.

.github/workflows/flags.yml
```diff
@@ -69,14 +69,21 @@ jobs:
           ghc: ${{ matrix.ghc }}
           os:  ${{ runner.os }}
 
-      - name: Build `hls-graph` with flags
-        run: cabal v2-build hls-graph --flags="embed-files stm-stats"
-
-      - name: Build `ghcide` with flags
-        run: cabal v2-build ghcide --flags="ghc-patched-unboxed-bytecode test-exe executable bench-exe ekg"
```
jhrcek, line 76: ghc-patched-unboxed-bytecode and bench-exe flags seem to be some historic leftovers - they don't exist in ghcide.cabal anymore, so I just removed them from here.

The idea behind using cabal configure is that we do only single cabal build with pedantic flags enabled for everything - and don't skip pedantic when building hls-graph and ghcide.

Also I guessed what is the purpose of this flags job and added some comments - shout if you don't agree with any of this.

jhrcek:
@michaelpj I did the changes you suggested 
1) use cabal configure to turn on pedantic flag for everything and only do single cabal build
2) moved `Key`related internals to separate module with explicit import list
+ fixed minimal set of fixes to make the build pass with the newly enabled pedantic flags.

.github/workflows/flags.yml
```diff
@@ -69,14 +69,21 @@ jobs:
...
+      - name: Configue flags for all components
+        run: |
+          cabal configure \
+              --constraint "hls-graph +embed-files +stm-stats" \
+              --constraint "ghcide +ekg +executable +test-exe" \
+              --constraint "hls-plugin-api -use-fingertree" \
+              --constraint "all +pedantic"
+          cat cabal.project.local
 
       - name: Build with pedantic (-WError)
```
michaelpj, line 85: this is now inaccurate

michaelpj (approved):
Looks great! just one nit

PR #4050, opened 2024-02-04
Optimize semantic token extraction logic

soulomoon:


plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs
```diff
@@ -22,42 +21,50 @@ import           Data.Text.Utf16.Rope.Mixed       (Rope)
...
 data PTokenState = PTokenState
-  { rangeIdSetMap  :: !RangeIdSetMap,
-    rope           :: !Rope, -- the remains of rope we are working on
-    cursor         :: !Char.Position, -- the cursor position of the current rope to the start of the original file in code point position
-    columnsInUtf16 :: !UInt -- the column of the start of the current rope in utf16
+  {
+    rope                  :: !Rope -- the remains of rope we are working on
+    , cursor              :: !Char.Position -- the cursor position of the current rope to the start of the original file in code point position
+    , columnsInUtf16      :: !UInt -- the column of the start of the current rope in utf16
+    , rangeHsSemanticList :: [(Range, HsSemanticTokenType)] -- (range, token type) in reverse order
```
wz1000, line 40: can we move this out the state? It seems like all the other `Tokenizer` functions return `()`, instead we could just return `[(Range, HsSemanticTokenType)]`

soulomoon: Sure, maybe we can do a cps style to collect them one by one.

soulomoon:
cc @wz1000  
see if it is looking good now?

wz1000:
Yes. I'm ambivalent on the `DList`, but whatever seems cleaner to you. We can go back to the old approach with the state too if you prefer that.

wz1000 (approved):

soulomoon:
> Yes. I'm ambivalent on the `DList`

Just curious, I wonder why

wz1000:
Now that I see it, I wonder if maybe the previous approach was cleaner. But it is totally up to you!

soulomoon:
I see, I review it a bit, both should have simliar cleaness(DList cut a few lines), DList seems make the type look better. I'll stick with DList, seems more explicite.

PR #4052, opened 2024-02-05
Bump lsp versions

michaelpj:
Broadly:
- A few places where we need to pipe `ProgressToken`s around.
   - I also just removed the progress reporting from resolve commands, since it's going to often be costly to do progress reporting on something that short. Possibly we could revisit after https://github.com/haskell/lsp/issues/549
- Some changes to the registration options we infer
- A few places where we need to adapt to ignoring registrations or not
- Adapting to use the ghcide verison of `getCompletionPrefix` everywhere
- Adapting to use the new mixed rope format

### DIFF
--- a/ghcide/test/exe/WatchedFileTests.hs
+++ b/ghcide/test/exe/WatchedFileTests.hs
@@ -28,6 +28,7 @@ tests = testGroup "watched files"
     [ testSession' "workspace files" $ \sessionDir -> do
         liftIO $ writeFile (sessionDir </> "hie.yaml") "cradle: {direct: {arguments: [\"-isrc\", \"A\", \"WatchedFilesMissingModule\"]}}"
         _doc <- createDoc "A.hs" "haskell" "{-#LANGUAGE NoImplicitPrelude #-}\nmodule A where\nimport WatchedFilesMissingModule"
+        setIgnoringRegistrationRequests False
         watchedFileRegs <- getWatchedFilesSubscriptionsUntil SMethod_TextDocumentPublishDiagnostics
 
         -- Expect 2 subscriptions: one for all .hs files and one for the hie.yaml cradle
@@ -38,6 +39,7 @@ tests = testGroup "watched files"
         let yaml = "cradle: {direct: {arguments: [\"-i" <> tail(init(show tmpDir)) <> "\", \"A\", \"WatchedFilesMissingModule\"]}}"
         liftIO $ writeFile (sessionDir </> "hie.yaml") yaml
         _doc <- createDoc "A.hs" "haskell" "{-# LANGUAGE NoImplicitPrelude#-}\nmodule A where\nimport WatchedFilesMissingModule"
+        setIgnoringRegistrationRequests False
         watchedFileRegs <- getWatchedFilesSubscriptionsUntil SMethod_TextDocumentPublishDiagnostics
 
         -- Expect 2 subscriptions: one for all .hs files and one for the hie.yaml cradle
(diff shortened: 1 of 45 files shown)


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Completions.hs
```diff
@@ -207,7 +209,7 @@ splitAtPosition pos ls = do
   split <- splitFile
   pure $ reverse $ Rope.lines $ fst split
   where
-    splitFile = Rope.splitAtPosition ropePos ls
+    splitFile = Rope.utf16SplitAtPosition ropePos ls
```
michaelpj, line 212: I think I set this up right, but I'm not 100% sure. We would need a cabal file test with unicode characters...

michaelpj:
I also haven't taken advantage of the fact that we now get an updater that lets us send text-only updates from `withIndefiniteProgress`.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -1084,11 +1086,11 @@ mkFormattingHandlers f = mkPluginHandler SMethod_TextDocumentFormatting ( provid
         mf <- lift $ getVirtualFile $ toNormalizedUri uri
         case mf of
           Just vf -> do
-            let typ = case m of
-                  SMethod_TextDocumentFormatting -> FormatText
-                  SMethod_TextDocumentRangeFormatting -> FormatRange (params ^. L.range)
+            let (typ, mtoken) = case m of
```
jhrcek, line 1089: Seems strange to have the same code in both branches if it doesn't depend on the scrutinee `case m of` ..
I'd add separate `mtoken = params ^. L.workDoneToken` in the let.

ghcide/test/exe/WatchedFileTests.hs
```diff
@@ -28,6 +28,7 @@ tests = testGroup "watched files"
     [ testSession' "workspace files" $ \sessionDir -> do
         liftIO $ writeFile (sessionDir </> "hie.yaml") "cradle: {direct: {arguments: [\"-isrc\", \"A\", \"WatchedFilesMissingModule\"]}}"
         _doc <- createDoc "A.hs" "haskell" "{-#LANGUAGE NoImplicitPrelude #-}\nmodule A where\nimport WatchedFilesMissingModule"
+        setIgnoringRegistrationRequests False
```
wz1000, line 31: what does this do?

jhrcek (approved):
LGTM, just one nitpick.

soulomoon (approved):
LGTM!

ghcide/test/exe/WatchedFileTests.hs
michaelpj: It says "don't pretend that `registerCapability` and `unregisterCapability` messages don't exist", which we mostly want to otherwise it adds more noise. This test actually looks at those messages so needs to set this.

hls-plugin-api/src/Ide/Types.hs
michaelpj: It actually _does_ depend on the scrutinee, because `params` has type `MethodParams m`, so we don't know what it is until we do the case. In this two cases we actually have different types for `params`, they just happen to boht have `range` fields. A bit weird, but you have to do it like this!

jhrcek: Ah, these invisible type-level things don't make reviewing easy :smile:

michaelpj: haha yeah

PR #4054, opened 2024-02-06
Promote more warnings to errors in ghcide

jhrcek:


ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -190,12 +190,10 @@ documentSymbolForDecl (L (locA -> (RealSrcSpan l _)) (ForD _ x)) = Just
     { _name   = case x of
                   ForeignImport{} -> name
                   ForeignExport{} -> name
-                  XForeignDecl{}  -> "?"
     , _kind   = SymbolKind_Object
     , _detail = case x of
                   ForeignImport{} -> Just "import"
                   ForeignExport{} -> Just "export"
-                  XForeignDecl{}  -> Nothing
```
jhrcek, line 198: One of those "trees that grow" constructors that apparently can't happen in parsed AST, so is marked as redundant pattern match.

michaelpj (approved):

PR #4058, opened 2024-02-07
Bump ghcide-test-utils to 2.0.0.0

wz1000:
We need this so we can release a version compatible with 2.6.0.0

I will need to make hackage revisions to all the plugin packages that depend on it

### DIFF
--- a/ghcide/test/ghcide-test-utils.cabal
+++ b/ghcide/test/ghcide-test-utils.cabal
@@ -3,7 +3,7 @@ cabal-version:      3.0
 build-type:         Simple
 category:           Development
 name:               ghcide-test-utils
-version:            1.9.0.0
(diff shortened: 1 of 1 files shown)


ghcide/test/ghcide-test-utils.cabal
```diff
@@ -3,7 +3,7 @@ cabal-version:      3.0
 build-type:         Simple
 category:           Development
 name:               ghcide-test-utils
-version:            1.9.0.0
+version:            2.0.0.0
```
jhrcek, line 6: Nitpick: should we update tested-with field to reflect what's being tested in CI?

soulomoon (approved):

jhrcek (approved):
Thanks. LGTM

Issue #3736, opened 2023-07-31
Overhauling the HLS testsuite


fendor:
## Problem

The test suite of HLS is gigantic. That's great!
Running the test suite of HLS takes a long time. That's bad.
Even worse, the test suite has become immensely flaky. Usually, we require at least one CI rerun for failed CI jobs per commit. Then we add insult to injury, since tests fail for components that the PR doesn't even touch.
We concluded, that a lot of the flakiness and bad performance originates from lsp-test which is basically a parser generator for parsing a very specific order of lsp messages.

However, HLS is not the most deterministic language server out there, so sometimes messages come out of the expected order. For example, published diagnostics for module A are sent before module B, which leads the test to fail. Rerunning the same test might yield a different order of messages, causing the test to succeed. 

The performance and flakiness of our CI has become a real problem, which is why need to come up with solutions.

As a comparison, `rust-analyzers` while test-suite takes 1.30-min while HLS usually takes up to an hour. It is our dedicated goal to be within the 5-min mark for running our *entire* test suite.

## Solution

There are many ways to reach our destination. So we outline a couple here:

1) Reducing flakiness in `lsp-test`. We have identified before some common sources of flakiness. The most important is loading too many modules into the same test session. Making sure, each test-case loads only exactly what it needs for the test, reduces the flakiness.
2) I argue that `lsp-test` is often the wrong tool for testing. To be precise, LSP is the wrong API boundary for most of our tests! `lsp-test` sets up an entire HLS session, negotiates capabilities, etc... All not needed for testing a plugin Code Action! Thus, we take https://matklad.github.io/2021/05/31/how-to-test.html as our role model and try to gain as much as possible from it.

The rest of the issue, will outline how we envision fixing the test suite using the knowledge gained from point 2.

tbd, I have some POC for Code Actions

michaelpj:
Just brainstorming. It seems to me that many plugins could have their tests split into two:
- Tests for the handlers, assuming that the rules all work and return the right things.
- Tests for the rules, checking that they compute the right thing in a real HLS session

And then it's relatively reasonable to assume that the composition of these things works.

The second part does seem like it would benefit from being run in a HLS _build_ session. But it maybe it doesn't even need a LSP server! We could set up the VFS manually, and then just query the state of a rule.

fendor:
> Really nice ideas, how is the current state of it.

[...]

PR #4059, opened 2024-02-07
Improve hls class plugin test

soulomoon:
I don't think we need to pass in the source to match the result, since its result is throwed away.
switching from `waitForDiagnosticsFromSource` to `waitForDiagnosticsFrom` improve the test speed by magnitude.

before
`All 29 tests passed (88.71s)`
after
`All 29 tests passed (5.74s)`

### DIFF
--- a/plugins/hls-class-plugin/test/Main.hs
+++ b/plugins/hls-class-plugin/test/Main.hs
@@ -124,7 +124,7 @@ codeLensTests = testGroup
     , testCase "Do not construct error action!, Ticket3942one" $ do
         runSessionWithServer def classPlugin testDataDir $ do
             doc <- openDoc "Ticket3942one.hs" "haskell"
-            _ <- waitForDiagnosticsFromSource doc (T.unpack sourceTypecheck)
+            _ <- waitForDiagnosticsFrom doc
             lens <- getAllCodeActions doc
             -- should switch to `liftIO $ length lens @?= 2, when Ticket3942 is entirely fixed`
             -- current fix is just to make sure the code does not throw an exception that would mess up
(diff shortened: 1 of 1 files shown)


soulomoon:
fixing the behaviour of `waitForDiagnosticsFromSourceWithTimeout` would be better

plugins/hls-class-plugin/test/Main.hs
```diff
@@ -124,7 +124,7 @@ codeLensTests = testGroup
     , testCase "Do not construct error action!, Ticket3942one" $ do
         runSessionWithServer def classPlugin testDataDir $ do
             doc <- openDoc "Ticket3942one.hs" "haskell"
-            _ <- waitForDiagnosticsFromSource doc (T.unpack sourceTypecheck)
```
michaelpj, line 127: This function is rather weird. I don't know why it isn't just a slight variant of `waitForDiagnostics` that adds an extra predicate...

michaelpj (commented):
Should we just get rid of `waitForDiagnosticsWithSource` entirely? It seems like it's going to introduce unconditional waits everywhere, which seems pretty bad!

soulomoon:
> waitForDiagnosticsWithSource

This one has a timeout, and something I am not so sure about. It seems some test rely on the timeout. 
I tried a naive fix but does not successful here, It seems rather complex. 
Maybe we can open a seperate issue and get rid of(or fixing it to perform better) `waitForDiagnosticsWithSource`

michaelpj:
Okay, well maybe let's at least note it down as something to investigate in https://github.com/haskell/haskell-language-server/issues/3736

michaelpj (approved):

PR #4060, opened 2024-02-08
stack CI: switch to offic. haskell images, bump to lts-22.9 (ghc 9.6.4)

jhrcek:
Motivation:
this week I saw several instances of stack CI jobs failing with errors ala "failed to download stack binaries from fpcomplete servers".

I propose we switch to official haskell images which have latest stack preinstalled so we can avoid this class of failures altogether.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: nightly-2023-07-10 # ghc-9.6.2
+resolver: lts-22.9 # ghc-9.6.4
 
 packages:
   - .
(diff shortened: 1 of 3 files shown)


stack.yaml
```diff
@@ -20,25 +20,12 @@ extra-deps:
 - retrie-1.2.2
 - hiedb-0.5.0.1
 - implicit-hie-0.1.4.0
-- hie-bios-0.13.1
```
jhrcek, line 23: Removing these from extra-deps, because all of these are in lts-22.9

michaelpj (approved):

PR #4061, opened 2024-02-09
Enable pedantic for more components

jhrcek:


cabal.project
```diff
@@ -72,5 +72,3 @@ if impl(ghc >= 9.7)
   -- this is okay
   allow-newer:
     ekg-core:text,
-    -- https://github.com/haskell-primitive/primitive-unlifted/issues/39
-    primitive-unlifted:bytestring,
```
jhrcek, line 76: Seems like primitive-unlifted is not used by anything anymore, so removing it from here.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -70,8 +71,8 @@ renameProvider state pluginId (RenameParams _prog docId@(TextDocumentIdentifier
            indirect references through punned names. To find the transitive closure, we do a pass of
            the direct references to find the references for any punned names.
            See the `IndirectPuns` test for an example. -}
-        indirectOldNames <- concat . filter ((>1) . Prelude.length) <$>
-            mapM (uncurry (getNamesAtPos state) . locToFilePos) directRefs
+        indirectOldNames <- concat . filter notNull <$>
+            mapM (uncurry (getNamesAtPos state)) (mapMaybe locToFilePos directRefs)
```
jhrcek, line 75: Change locToFilePos (see def. below) to return maybe.
I hope that silently ignoring paths for which uriToNormalizedFilePath returns Nothing, rather than crashing is ok.

cabal.project
michaelpj: It's a transitive dependency, but if it works then it works!

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
michaelpj: Hmm, I think we might want to fail the rename, because that means there's a potential reference that we couldn't resolve, so we might not be doing a correct rename.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -187,8 +188,8 @@ refsAtName state nfp name = do
 
 nameLocs :: Name -> (HieAstResult, PositionMapping) -> [Location]
 nameLocs name (HAR _ _ rm _ _, pm) =
-    mapMaybe (toCurrentLocation pm . realSrcSpanToLocation . fst)
-             (concat $ M.lookup (Right name) rm)
+    concatMap (mapMaybe (toCurrentLocation pm . realSrcSpanToLocation . fst))
```
michaelpj, line 191: ... although here's another `mapMaybe`, so I guess we're already silently ignoring some things and maybe it's fine :shrug:

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
michaelpj: Maybe just add a comment at least.

michaelpj (approved):

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
jhrcek: Ok, will dive a bit into how this works and get back to this.

jhrcek: Ok, I found there's a dedicated `getNormalizedFilePathE` function for this that just translates the Nothing to PluginError, which sounds cleaner/more informative than pattern match error..

Issue #4056, opened 2024-02-07
Handling of `Diagnostics` in code action contexts is questionable


michaelpj:
When we get a `textDocument/codeAction` request, the context contains some `Diagnostic`s. These are (allegedly) diagnostics that are a) visible to the user and b) overlap with the requested range.

However, it's fairly unclear what the client is allowed to do in populating this field. In particular, it's not clear that they have to give it back to us as we sent it to them. The spec also indicates that this field is unreliable, the doc for it says 

> They are provided so that the server knows which errors are *currently presented to the user* for the given range. There is *no guarantee* that these accurately reflect the error state of the resource. The primary parameter to compute code actions is the *provided range*.

(emphasis mine)

We rely on the context `Diagnostics` quite a bit:
- Various code actions that trigger off diagnostics inspect them
    - This seems questionable since in most cases we want to trigger based on the objective fact about the code that triggers the diagnostic (e.g. "redundant import"), and _not_ on whether this fact is presented to the user.
- In some places we compare them against existing diagnostics we know about
    - This can go wrong if the client messes with the diagnostic, see e.g. https://github.com/haskell/haskell-language-server/issues/3857
    - If we actually care about this we should put identifying information in the `data` field.

Moreover, we don't really need to do this at all: we can just consult _our_ diagnostic store and filter it based on the given range. This is more direct and simpler, and we should probably just do this in almost all cases. The exception would be cases where we actually _do_ care about presentation: for example, the "suppress this warning" code action _does_ care about whether the diagnostic it is proposing to suppress is in fact visible to the user. But this is an exceptional case.

michaelpj:
See also https://github.com/haskell/haskell-language-server/pull/4051#issuecomment-1927878937

keithfancher:
> See also #4051 (comment)

To clarify/summarize my somewhat-rambly comment: I agree!

A process like this makes a lot of sense to me:

1. Filter all available server-side diagnostics based *only* on range(s) from the incoming context.
2. Based on the results of that filter, i.e. still using the server-side diags, decide which code actions to show. (The messy bits: regexes and other matching logic, etc.)

And I think there's potential to consolidate some of the logic for *both* of these steps to simplify things for plugin authors a bit.

As for edge cases/exceptions... I'll leave it to the experts!

fendor:
I am pretty sure we try to do this now, we have `activeDiagnosticsInRange` and `activeDiagnosticsInRangeMT`, so this should be trivial to do now.

[...]

Issue #3857, opened 2023-10-31
'Remove redundant imports' code action not showing in neovim 0.9


asivitz:
### Your environment

Which OS do you use?
MacOS

Which version of GHC do you use and how did you install it?
9.4.7 from ghcup

How is your project built (alternative: link to the project)?
stack

Which LSP client (editor/plugin) do you use?
nvim + lspconfig (the neovim native lsp)

Which version of HLS do you use and how did you install it?
2.4.0.0 from ghcup

Have you configured HLS in any way (especially: a `hie.yaml` file)?
Yes, a hie.yaml

### Steps to reproduce
Open a file using neovim 0.9. Add a redundant import. Ask for code actions on this line.

### Expected behaviour
There is a code action to 'Remove import' and 'Remove all redundant imports', as well as 'Disable 'unused-imports' warnings'

### Actual behaviour
There is only a code action to disable the warnings

### Debug information
The code actions do appear under neovim 0.7. Here are RPC logs for the two versions. Looks like the request data changed a bit between 0.7 and 0.9.

# NEOVIM 0.7 REQUEST
```
[DEBUG][2023-10-31 14:31:44] .../vim/lsp/rpc.lua:347	"rpc.send"	{  id = 2,  jsonrpc = "2.0",  method = "textDocument/codeAction",  params = {    context = {      diagnostics = { {          code = "-Wunused-imports",          message = "The import of ‘generateDebugUI’\nfrom module ‘DebugInfoScreen’ is redundant",          range = {            end = {              character = 53,              line = 8            },            start = {              character = 38,              line = 8            }          },          severity = 2,          source = "typecheck",          tags = { 1 }        } }    },    range = {      end = <1>{        character = 0,        line = 8      },      start = <table 1>    },    textDocument = {      uri = "file:///Users/axis/dev/ultimate/src/Program/IODriver.hs"    }  }}
```

# NEOVIM 0.9 REQUEST
```
[...]

PR #4063, opened 2024-02-09
Redundant imports/exports: use range only to determine which code actions are in scope

keithfancher:
...rather than doing a full compare with incoming `Diagnostic` objects from the client.

This brings the "remove redundant imports/exports" code actions more in line with behavior described in https://github.com/haskell/haskell-language-server/issues/4056, and has the pleasant side-effect of fixing broken code actions in neovim (https://github.com/haskell/haskell-language-server/issues/3857).

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
@@ -80,7 +80,8 @@ import           GHC.Exts                                          (fromList)
 import qualified GHC.LanguageExtensions                            as Lang
 import           Ide.Logger                                        hiding
                                                                    (group)
-import           Ide.PluginUtils                                   (extractTextInRange,
+import           Ide.PluginUtils                                   (extendToFullLines,
+                                                                    extractTextInRange,
                                                                     subRange)
 import           Ide.Types
 import           Language.LSP.Protocol.Message                     (Method (..),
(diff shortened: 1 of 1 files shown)


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -438,19 +438,25 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
         = [("Remove import", [TextEdit (extendToWholeLineIfPossible contents _range) ""])]
     | otherwise = []
 
+diagInRange :: Diagnostic -> Range -> Bool
+diagInRange Diagnostic {_range = dr} r = dr `subRange` extendedRange
+  where
+    -- Ensures the range captures full lines. Makes it easier to trigger the correct
+    -- "remove redundant" code actions from anywhere on the offending line.
+    extendedRange = extendToFullLines r
```
fendor, line 446: I am a bit unsure about this, so this makes it easier for code actions to overlap, right?

So if we have two refactor code actions in the same line, they'd overlap, potentially making it unclear which code action would be truly triggered for an diagnostic?

keithfancher: > so this makes it easier for code actions to overlap, right?

It does! But (at least in my opinion) in a good way :)

For example, here are two lines with unused imports:

Screenshot from 2024-02-14 10-58-18

*With* extending to full line, I can move my cursor to *anywhere* on the offending `import` line -- say, the beginning of the line -- and show code actions, and I'll see all code actions that are relevant to that `import`, including ones about redundancies:

Screenshot from 2024-02-14 11-01-48

*Without* extending to full line, I need to put my cursor *directly on* `showInfoBrief` to see the relevant code actions. Which also makes things a bit confusing, IMO... especially since some import-related code actions ("Make imports explicit", e.g.) aren't tied to diagnostics at all, it can seem a little random for a user to click on a different part of an import and see a different set of code actions.

> So if we have two refactor code actions in the same line, they'd overlap, potentially making it unclear which code action would be truly triggered for an diagnostic?

There's always the possibility of overlap with code actions regardless, right? Multiple actions can apply to a single piece of code. But it's not a problem -- the code actions still work exactly as expected, and it's still up to the user to select what they'd like to actually *change*. For example, on the unused `import Debug.Trace` line, I get other actions in addition to the redundant import ones:

Screenshot from 2024-02-14 11-05-43

And they all work great, as expected.

To me, extending to the full line in *this particular case* -- with redundant imports -- seemed like a usability improvement. And in fact, as a user I'd be a little surprised if it didn't work that way.

But of course I'm open to changing it if folks feel otherwise!

keithfancher:
What do y'all think? Is this good to go, or any other open questions? Would probably be good for @michaelpj to take a look as well, to make sure I'm getting the intent of #4056 correct.

I'd love to get #3857 closed out, it's such an annoying bug for neovim users...

fendor (approved):
LGTM

keithfancher:
> LGTM

Awesome, thank you!

As for getting it merged... do I just hang tight and wait for a maintainer to pull the trigger?

michaelpj:
Sorry, I am going to take a look at this! I was on holiday last week and it hasn't quite got to the top of my pile yet :)

keithfancher:
> Sorry, I am going to take a look at this! I was on holiday last week and it hasn't quite got to the top of my pile yet :)

Oh, no worries at all, take your time of course! I just wanted to be sure I wasn't missing some step on my end.

Thank you!

PR #4070, opened 2024-02-12
Fix -Wall in qualified imported names plugin

jhrcek:


plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
```diff
@@ -111,7 +110,7 @@ data ImportedBy = ImportedBy {
 }
 
 isRangeWithinImportedBy :: Range -> ImportedBy -> Bool
-isRangeWithinImportedBy range (ImportedBy _ srcSpan) = fromMaybe False $ spanContainsRange srcSpan range
+isRangeWithinImportedBy range ImportedBy{importedBySrcSpan} = fromMaybe False $ spanContainsRange importedBySrcSpan range
```
jhrcek, line 113: Agree that using named field puns could be controversial, I'm just fixing warnings of the type
`Defined but not used: ‘importedBySrcSpan’`

Other ways to make this go away:
- prefix record selectors in data type definition with _
- use the record selectors (e.g. `importedBySrcSpan ib`) in function body

fendor (approved):

plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs
fendor: I think it is fine, and I don't think we have any policy regarding `NamedFieldPuns`.

PR #4071, opened 2024-02-13
Fix -Wall in retrie plugin

jhrcek:


plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -212,20 +211,12 @@ runRetrieInlineThisCmd state token RunRetrieInlineThisParams{..} = do
     --   Run retrie to get a list of changes
     --   Select the change that inlines the identifier in the given position
     --   Apply the edit
-    ast <- runActionE "retrie" state $
```
jhrcek, line 215: All of these bindings were unused. Not sure if they can be removed (if something is relying on the side effects of these?) or not, but the tests are passing so.. :man_shrugging:

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -333,33 +324,32 @@ provider state plId (CodeActionParams _ _ (TextDocumentIdentifier uri) range ca)
...
-      topLevelBinds =
-        [ decl
-          | (_, bagBinds) <- binds,
-            L _ decl <- bagToList bagBinds
-        ]
-  return (tmrModSummary tm, topLevelBinds, posMapping, hs_ruleds, hs_tyclds)
+  case rn of
+    (HsGroup{hs_valds, hs_ruleds, hs_tyclds}, _, _, _) -> do
+      topLevelBinds <- case hs_valds of
+        ValBinds{} -> throwError $ PluginInternalError "getBinds: ValBinds not supported"
```
jhrcek, line 346: The above was giving "incomplete pattern match: constructors not matched ValBinds ..." warning.
Throwing error seems far from ideal but I still consider it step forward (better visibility into where the error is coming from).

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -167,10 +165,11 @@ data RunRetrieParams = RunRetrieParams
     restrictToOriginatingFile :: Bool
   }
   deriving (Eq, Show, Generic, FromJSON, ToJSON)
+
 runRetrieCmd :: CommandFunction IdeState RunRetrieParams
 runRetrieCmd state token RunRetrieParams{originatingFile = uri, ..} = ExceptT $
   withIndefiniteProgress description token Cancellable $ \_updater -> do
-    runExceptT $ do
+    _ <- runExceptT $ do
```
jhrcek, line 172: This is definitely suspicious: wouldn't we want to send the failure to client instead of ignoring it?

soulomoon: Or we can atleast log the error?

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
soulomoon: I don't think they are nessary.

soulomoon (approved):
LGTM

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
jhrcek: On second thought it's actually sending some errors notifications to the client within the block if errors occur, so this is probably ok.

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
fendor: If the tests are passing, let's remove them for now and wait for bug reports :)

fendor (approved):
LGTM, thanks!

PR #4073, opened 2024-02-13
add Method_TextDocumentSemanticTokensFullDelta

soulomoon:
Mainly the following things have been done:
1. add `semanticTokensCache` and `semanticTokensId` to shake extras.
2. implement `Method_TextDocumentSemanticTokensFullDelta`
3. add a few test to test `Method_TextDocumentSemanticTokensFullDelta`
4. remove `persistentGetSemanticTokensRule` since it is of no use since we do not use `useWithStaleFastMT` for semantic tokens rule.


plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
```diff
@@ -156,6 +177,60 @@ semanticTokensConfigTest = testGroup "semantic token config test" [
...
+semanticTokensFullDeltaTests =
+  testGroup "semanticTokensFullDeltaTests" $
+    [ testCase "null delta since unchanged" $ do
+        let file1 = "TModula𐐀bA.hs"
+        let expectDelta = InR (InL (SemanticTokensDelta (Just "1") []))
+        Test.Hls.runSessionWithServerInTmpDir def semanticTokensPlugin (mkFs $ FS.directProjectMulti [file1]) $ do
+          doc1 <- openDoc file1 "haskell"
+          _ <- waitForAction "TypeCheck" doc1
+          fullResult <- Test.getSemanticTokens doc1
+          liftIO $ print fullResult
```
fendor, line 190: debug function

plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
```diff
@@ -156,6 +177,60 @@ semanticTokensConfigTest = testGroup "semantic token config test" [
...
+          _ <- Test.getSemanticTokens doc1
+          -- open the file and append a line to it
+          let change = TextDocumentContentChangeEvent
+                $ InL $ #range .== Range (Position 4 0) (Position 4 6)
+                .+ #rangeLength .== Nothing
+                .+ #text .== "foo = 1"
+          changeDoc doc1 [change]
+          _ <- waitForAction "TypeCheck" doc1
+          delta <- getSemanticTokensFullDelta doc1 "0"
+          liftIO $ print delta
```
fendor, line 211: debug function

plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
```diff
@@ -156,6 +177,60 @@ semanticTokensConfigTest = testGroup "semantic token config test" [
...
+          _ <- Test.getSemanticTokens doc1
+          -- open the file and append a line to it
+          let change = TextDocumentContentChangeEvent
+                $ InL $ #range .== Range (Position 2 0) (Position 2 28)
+                .+ #rangeLength .== Nothing
+                .+ #text .== Text.replicate 28 " "
+          changeDoc doc1 [change]
+          _ <- waitForAction "TypeCheck" doc1
+          delta <- getSemanticTokensFullDelta doc1 "0"
+          liftIO $ print delta
```
fendor, line 230: debug function

plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
```diff
@@ -101,6 +107,21 @@ docLspSemanticTokensString doc = do
...
+getSemanticTokensFullDelta doc lastResultId = do
+  let params = SemanticTokensDeltaParams Nothing Nothing doc lastResultId
+  rsp <- request SMethod_TextDocumentSemanticTokensFullDelta params
+  case rsp ^. L.result of
+    Right x -> return x
+    _       -> error "No tokens found"
+
+-- docLspSemanticTokensFullDeltaString :: TextDocumentIdentifier -> Session [SemanticTokenOriginal Language.LSP.Protocol.Types.SemanticTokenTypes]
+-- docLspSemanticTokensFullDeltaString doc = do
+--   res <- Test.getSemanticTokens doc
```
fendor, line 123:
```suggestion
```

michaelpj:
I'm on holiday but I'll take a look next week!

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -259,6 +261,8 @@ data ShakeExtras = ShakeExtras
     ,publishedDiagnostics :: STM.Map NormalizedUri [Diagnostic]
     -- ^ This represents the set of diagnostics that we have published.
     -- Due to debouncing not every change might get published.
+    ,semanticTokensCache:: STM.Map NormalizedUri SemanticTokens
+    ,semanticTokensId :: TVar Int
```
fendor, line 265: Some docs please, why we have them, etc... :)

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -68,14 +78,38 @@ computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeS
...
+semanticTokensFullDelta :: Recorder (WithPriority SemanticLog) -> PluginMethodHandler IdeState 'Method_TextDocumentSemanticTokensFullDelta
+semanticTokensFullDelta recorder state pid param = do
   nfp <- getNormalizedFilePathE (param ^. L.textDocument . L.uri)
-  items <- runActionE "SemanticTokens.semanticTokensFull" state $ computeSemanticTokens recorder pid state nfp
-  return $ InL items
+  let previousVersionFromParam = param ^. L.previousResultId
+  runActionE "SemanticTokens.semanticTokensFullDelta" state $ computeSemanticTokensFullDelta previousVersionFromParam recorder pid state nfp
+  where
+    computeSemanticTokensFullDelta :: Text -> Recorder (WithPriority SemanticLog) -> PluginId -> IdeState -> NormalizedFilePath -> ExceptT PluginError Action (MessageResult Method_TextDocumentSemanticTokensFullDelta)
+    computeSemanticTokensFullDelta previousVersionFromParam recorder pid state nfp = do
```
fendor, line 103:
```suggestion
    computeSemanticTokensFullDelta :: Recorder (WithPriority SemanticLog) -> Text -> PluginId -> IdeState -> NormalizedFilePath -> ExceptT PluginError Action (MessageResult Method_TextDocumentSemanticTokensFullDelta)
    computeSemanticTokensFullDelta recorder previousVersionFromParam pid state nfp = do
```

Feels a little bit inconsistent otherwise.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -113,3 +144,21 @@ handleError recorder action' = do
...
+  liftIO $ atomically $ do
+    i <- readTVar semanticTokensId
+    modifyTVar' semanticTokensId (+1)
+    return $ T.pack $ show i
+
+getPreviousSemanticTokens :: NormalizedUri -> Action (Maybe SemanticTokens)
+getPreviousSemanticTokens uri = getShakeExtras >>= liftIO . atomically . STM.lookup uri . semanticTokensCache
+
+setSemanticTokens :: NormalizedUri -> SemanticTokens -> Action ()
+setSemanticTokens uri tokens = getShakeExtras >>= liftIO . atomically . STM.insert tokens uri . semanticTokensCache
```
fendor, line 164: I am wondering, since we will keep the SemanticTokens Map around until HLS restarts, could we think of situations where we want to clear this store?
How much additional memory will this take, or is memory usage something we have decided to not care about in this plugin, since it is quite expensive already?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -113,3 +144,21 @@ handleError recorder action' = do
...
+-----------------------
+-- helper functions
+-----------------------
+
+getAndIncreaseSemanticTokensId :: Action Text
+getAndIncreaseSemanticTokensId = do
+  ShakeExtras{semanticTokensId} <- getShakeExtras
+  liftIO $ atomically $ do
+    i <- readTVar semanticTokensId
+    modifyTVar' semanticTokensId (+1)
```
fendor, line 157:
```suggestion
    i <- stateTVar semanticTokensId (\val -> (val, val+1))
```
just some random bikeshed.

fendor (commented):
Changes look good to me, I mostly have minor suggestions.

While I do approve, let's wait until michael also had time to look at this :)

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: done

plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
soulomoon: done

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -113,3 +144,21 @@ handleError recorder action' = do
       logWith recorder Warning msg
       pure $ toIdeResult (Left [])
     Right value -> pure $ toIdeResult (Right value)
+
+-----------------------
+-- helper functions
+-----------------------
+
+getAndIncreaseSemanticTokensId :: Action Text
+getAndIncreaseSemanticTokensId = do
```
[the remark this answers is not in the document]
soulomoon: good idea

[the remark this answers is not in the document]
soulomoon: May be we can go with a type synonym

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: It stores the last response of semantic tokens for each file.
It is semanticTokens in the lsp format, main cost are the tokens, should be about 5 UInt(32-bit) for a token.
Say 1000 tokens for a file. Should be around 20kb a file. Suppose we have opened 1000 files, 20mb. (I think it should be fine.)

The main mem usage would be the tons of dependency files that we type checked when computing the semantic tokens.

soulomoon:
> Changes look good to me, I mostly have minor suggestions.

Thanx for the reviews. Fixing most of them.

> While I do approve, let's wait until michael also had time to look at this :)
Sure.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: done

ghcide/src/Development/IDE/Core/Shake.hs
fendor: Thanks, much better!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
fendor: Thank you for this calculation, seems reasonable to me!

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -68,14 +77,38 @@ computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeS
...
+  semanticId <- lift getAndIncreaseSemanticTokensId
   (RangeHsSemanticTokenTypes {rangeSemanticList}, mapping) <- useWithStaleE GetSemanticTokens nfp
-  withExceptT PluginInternalError $ liftEither $ rangeSemanticsSemanticTokens config mapping rangeSemanticList
+  withExceptT PluginInternalError $ liftEither $ rangeSemanticsSemanticTokens semanticId config mapping rangeSemanticList
 
 semanticTokensFull :: Recorder (WithPriority SemanticLog) -> PluginMethodHandler IdeState 'Method_TextDocumentSemanticTokensFull
-semanticTokensFull recorder state pid param = do
+semanticTokensFull recorder state pid param = runActionE "SemanticTokens.semanticTokensFull" state computeSemanticTokensFullDelta
+  where
+    computeSemanticTokensFullDelta :: ExceptT PluginError Action (MessageResult Method_TextDocumentSemanticTokensFull)
```
michaelpj, line 87: name seems wrong?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -259,6 +261,11 @@ data ShakeExtras = ShakeExtras
     ,publishedDiagnostics :: STM.Map NormalizedUri [Diagnostic]
     -- ^ This represents the set of diagnostics that we have published.
     -- Due to debouncing not every change might get published.
+    ,semanticTokensCache:: STM.Map NormalizedUri SemanticTokens
```
michaelpj, line 264: I think these could probably be variables local to the plugin? I think we only need to put thing in here if they're really shared across multiple rules etc. 

So the function that creates the plugin would run in `IO` or similar and create the TVars, which then get passed into the handlers.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -68,14 +77,38 @@ computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeS
...
+    computeSemanticTokensFullDelta = do
+      nfp <- getNormalizedFilePathE (param ^. L.textDocument . L.uri)
+      items <- computeSemanticTokens recorder pid state nfp
+      lift $ setSemanticTokens (normalizedFilePathToUri nfp) items
```
michaelpj, line 91: why not use the `NormalizedFilePath` as the key? you keep having to convert back to `NormalizedURI`?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -68,14 +77,38 @@ computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeS
...
+    computeSemanticTokensFullDelta recorder previousVersionFromParam  pid state nfp = do
+      semanticTokens <- computeSemanticTokens recorder pid state nfp
+      previousSemanticTokensMaybe <- lift $ getPreviousSemanticTokens (normalizedFilePathToUri nfp)
+      lift $ setSemanticTokens (normalizedFilePathToUri nfp) semanticTokens
+      case previousSemanticTokensMaybe of
+          Nothing -> return $ InL semanticTokens
+          Just previousSemanticTokens ->
+              if Just previousVersionFromParam == previousSemanticTokens^.L.resultId
+              then return $ InR $ InL $ makeSemanticTokensDeltaWithId (semanticTokens^.L.resultId) previousSemanticTokens semanticTokens
+              else return $ InL semanticTokens
```
michaelpj, line 111: Worth doing some debug logging in these cases? This one in particular seems like something has gone wrong and might even deserve a WARN log

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -98,9 +131,6 @@ getSemanticTokensRule recorder =
     let hsFinder = idSemantic getTyThingMap (hieKindFunMasksKind hieKind) refMap
     return $ computeRangeHsSemanticTokenTypeList hsFinder virtualFile ast
 
--- | Persistent rule to ensure that semantic tokens doesn't block on startup
```
michaelpj, line 101: why drop this?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -68,14 +77,38 @@ computeSemanticTokens :: Recorder (WithPriority SemanticLog) -> PluginId -> IdeS
 computeSemanticTokens recorder pid _ nfp = do
   config <- lift $ useSemanticConfigAction pid
   logWith recorder Debug (LogConfig config)
+  semanticId <- lift getAndIncreaseSemanticTokensId
```
michaelpj, line 80: possibly extremely dumb suggestion, but... what if we used the request ID for this? I guess maybe JSON-RPC doesn't guarantee that ids are never reused?

michaelpj: Or the file version?

michaelpj (commented):
Mostly worried about adding lots of stuff to `ShakeExtras` if we don't have to.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: SemanticTokens depends on the file and its dependencies. We might not find a good identifier except we maintain our own.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: good catch

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: True

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: True

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: Agree that putting it in shakeExtras is not optimal, but I am also not sure about local state to plugin too.
Since no other plugins have their own plugin state. It would mean we need to change `idePlugins :: Recorder (WithPriority Log) -> IdePlugins IdeState` to `idePlugins :: Recorder (WithPriority Log) -> IO (IdePlugins IdeState)`.
And open a door to introduce state to plugins breaking the current centralized state management in the hls build system. 🤔

soulomoon:
Thanx for the reviews @michaelpj , I fix some of them, and some might need discussion.

ghcide/src/Development/IDE/Core/Shake.hs
michaelpj: I don't think plugin-local state is inherently problematic. But I admit that it's not totally clear to me what stuff _should_ be in `ShakeExtras` :thinking: thoughts @fendor ?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
michaelpj: Hmm, so does this mean we have to wait quite a while to get responses when we start up?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
michaelpj: Yeah, seems reasonable. Maybe write a comment?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: yes, as we have been. Persistent rule won't help us here(Since it is not `ideAction`, the persistent rule serves only as a fall back when the actual computation of the rule failed and not already having a cache) .

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: Added

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: added a comment to the shakeExtra stating it might not be ideal to store the varaibles there, but we do not find a better place to store it yet.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -257,8 +259,18 @@ data ShakeExtras = ShakeExtras
     ,diagnostics :: STMDiagnosticStore
     ,hiddenDiagnostics :: STMDiagnosticStore
     ,publishedDiagnostics :: STM.Map NormalizedUri [Diagnostic]
+
+    -- storing semantic tokens for each file in shakeExtras might
```
michaelpj, line 263: this comment is in the wrong place, it's interrupting the haddock for the field above

michaelpj: Maybe put it in a Note or something

michaelpj (approved):
@wz1000 still interested in your feedback on the `ShakeExtras` vs plugin-local state vars question, but for now we're going to leave it as it is with a note.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: Sure. Did it

ghcide/src/Development/IDE/Core/Shake.hs
fendor: I do not have any specific thoughts, if the variable could be easily plugin local, I think I would prefer that.

fendor (approved):
Re-approving, LGTM :)

PR #4076, opened 2024-02-15
ghcide: Only try `stat`ing a core file after we ensure it actually exists

wz1000:
This doesn't change any behaviour, but ensures we have a more informative error message
if we somehow request a linkable that does not exist.


soulomoon (approved):
look reasonable

jhrcek (approved):
Seems like definite improvement.

fendor (approved):

Issue #4079, opened 2024-02-16
Refactor plugin shows "bug in srcspan parser" when trying to construct code action


jhrcek:
### Your environment

Which OS do you use?
Linux
Which version of GHC do you use and how did you install it?
9.4.8
How is your project built (alternative: link to the project)?

Which LSP client (editor/plugin) do you use?
VSCode

Which version of HLS do you use and how did you install it?
latest master
Have you configured HLS in any way (especially: a `hie.yaml` file)?
hie.yaml

This compiler error:

```
• Illegal term-level use of the type constructor or class ‘Id’
• imported from ‘PI.Core.Id’ at /home/jhrcek/Devel/github.com/Holmusk/PI/backend/src/PI/Db/CohortExplorer/Builder.hs:65:20-21
• Perhaps use variable ‘id’ (imported from Prelude)
  Perhaps you want to add ‘Id’ to the import list in the import of
  ‘PI.Core.Id’
  (/home/jhrcek/Devel/github.com/Holmusk/PI/backend/src/PI/Db/CohortExplorer/Builder.hs:65:1-22).
• In the second argument of ‘criteriaQuery’, namely
    ‘(Id 1 :: Id Cache)’
  In the second argument of ‘($)’, namely
    ‘criteriaQuery diagnosisLookup (Id 1 :: Id Cache) criteria’
  In a stmt of a 'do' block:
    pure $ criteriaQuery diagnosisLookup (Id 1 :: Id Cache) criteria
```

leads to this error being shown in the clinet:
![Screenshot from 2024-02-16 11-37-07](https://github.com/haskell/haskell-language-server/assets/2716069/c7fec299-dbdc-4896-8cf3-4cd032989a1c)

The error is coming from here:

https://github.com/haskell/haskell-language-server/blob/e93528bae7a747cad7cc22cd50b2e6db84da2baa/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs#L952

After adding some tracing I see that srcspan that's being passed to the parser is this:
```
[...]

PR #4080, opened 2024-02-16
refactor plugin: fix regex for extracting import suggestions

jhrcek:
Fixes https://github.com/haskell/haskell-language-server/issues/4079

The problem was that the original regex was mistakenly using `.` as a regex metacharacter, whereas the GHC error  contains a literal `.`

In cases where GHC error contains further `)`'s this led to 3rd match group (which is used to extract srcspan) to contain too much stuff, leading to errors described in the issue.

If the tests pass, would you still recommend adding a test case for this or is it ok like this?

### DIFF
--- a/plugins/hls-refactor-plugin/test/Main.hs
+++ b/plugins/hls-refactor-plugin/test/Main.hs
@@ -1275,7 +1275,8 @@ extendImportTests = testGroup "extend import actions"
                     , "b :: A"
                     , "b = ConstructorFoo"
                     ])
-        , testSession "extend single line import in presence of extra perens" $ template
+        , brokenForGHC92 "On GHC 9.2, the error doesn't contain \"perhaps you want ...\" part from which import suggestion can be extracted." $
+          testSession "extend single line import in presence of extra parens" $ template
             []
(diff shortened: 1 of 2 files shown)


fendor (approved):
LGTM

Preferably, we have a regression test proving this can't happen again.

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -1275,6 +1275,20 @@ extendImportTests = testGroup "extend import actions"
                     , "b :: A"
                     , "b = ConstructorFoo"
                     ])
+        , testSession "extend single line import in presence of extra perens" $ template
```
fendor, line 1278:
```suggestion
        , testSession "extend single line import in presence of extra parens" $ template
```

If you want, you may just mark this test broken for ghc 92 :see_no_evil:

jhrcek: On 9.2 the error doesn't contain any suggestion, which is why no code action is generated.
I tried for a bit to find an example that would trigger this on 9.2 but failed. I'll go with you suggestion.

PR #4092, opened 2024-02-24
Remove more workarounds for GHCs < 9.0

jhrcek:


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -266,40 +280,7 @@ getParsedModuleRule recorder =
 
     -- We still parse with Haddocks whether Opt_Haddock is True or False to collect information
     -- but we no longer need to parse with and without Haddocks separately for above GHC90.
-    res@(_,pmod) <- if Compat.ghcVersion >= Compat.GHC90 then
```
michaelpj, line 269: good idea, just a thought that we could probably just delete the `Compat.GHC90` enum member entirely and that might point out other places to fix.

jhrcek: Agree. I wasn't sure if I can just remove them but it definitely makes sense.

jhrcek:
There is more cleanup that can be done w.r.t unsupported ghc versions (e.g. `ghcVersion >= GHC92` will also be always True with the versions we support) but I'm going to open separate PR for that to keep things more focused and easy to review.

fendor (approved):
LGTM, thanks!

ghcide/test/exe/TestUtils.hs
```diff
@@ -306,8 +306,7 @@ defToLocation (InR (InR Null)) = []
 
 -- | Ghc 9 doesn't include the $-sign in TH warnings like earlier versions did
 thDollarIdx :: UInt
-thDollarIdx | ghcVersion >= GHC90 = 1
-            | otherwise = 0
+thDollarIdx = 1
```
michaelpj, line 309: inline?

michaelpj (approved):

test/wrapper/Main.hs
```diff
@@ -9,23 +9,29 @@ main = defaultTestRunner $ testGroup "haskell-language-server-wrapper" [projectG
 
 projectGhcVersionTests :: TestTree
 projectGhcVersionTests = testGroup "--project-ghc-version"
-  [ stackTest "9.2.8"
+  [ testCase "stack with global ghc" $ do
+      ghcVer <- ghcNumericVersion
```
michaelpj, line 13: nice

michaelpj (approved):

Issue #4046, opened 2024-02-04
HLS stuck (init then no progress) since multi home support


guibou:
First, I apologize for the lack of reproducibility in this issue. I'm a bit lost here and I need your help. I'll try to gather as much information as possible, but for now I don't have much.

Since we upgraded at work HLS from 2.4.0.0 to 2.6.0.0, HLS is "stuck" on some (but not all) of our projects.

By stuck, I mean that HLS starts, gather information for the build flags thank to our `hie.yaml` / `cradle:bios:shell` command, and then nothing else is happening. CPU runs at 100% (of one core) and the latest entry in the log is:

```
2024-02-04T12:10:01.928343Z | Info | Making new HscEnv. In-place unit ids: [ main-5f50f5bd074bca15292d92df75278f6ab5b479ff
```

### Your environment

We are using Linux. I'm debuging that on NixOS, but users had observed the same problem on Ubuntu.

GHC 9.6.4 (But same problem with 9.8.1), installed from nixpkgs. HLS is installed from nixpkgs too.

The project is built with a custom build system using nix and we use a `cradle:bios:shell` command to output the build flags. Note that the command works fine, so HLS is not blocked on this.

### A few notes about our build system.

Our build system is based on nix, but that's orthogonal to our problem (I think) because mostly what it does is populating the environment with the correct version of GHC and providing flags for HLS.

Note that we have one specifitity which may be responsible for the problem. We have a "main project" and multiples subprojects.

- The main project have a ghc database populated only with packages from hackage and built using nixpkgs and it contains 2000+ haskell module. HLS works fine here!
- Our subprojects have a ghc database populated with packages from hackage as well as from the main project. The ones from hackage are built using nixpkgs haskell infrastructure, but the ones from our main project are built using our custom build system. So maybe there is something wrong / incorrect in the way we are populating the package database. Note that `ghc-pkg check` only warns about missing `haddock` entries. None of our subproject work correctly with HLS, they are all stuck.

### HLS bisect

I've bisected HLS (starting on tag `2.6.0.0` and `2.4.0.0` as "bad" and "good" and I hit:

```
7db6215c6da39b6550d7c1ce6ea5b1bbe6e6dfde is the first bad commit
commit 7db6215c6da39b6550d7c1ce6ea5b1bbe6e6dfde
Author: wz1000 <zubin.duggal@gmail.com>
Date:   Wed Nov 22 16:08:04 2023 -0800

    Add support for multi unit argument syntax (#3462)
[...]

PR #4096, opened 2024-02-26
session-loader: Don't loop forever when we don't find a file in any multi component

wz1000:
We add a check for if the current file is a target we know about, and emit a
diagnostic if that is the case, refusing to load the file in.

This doesn't change the implicit adding of the current file as a target for a
single component case, as we need the old behaviour to support bare GHC/Direct
cradles where not all targets may be listed.

TODO: add test case


michaelpj:
Is this related to https://github.com/haskell/haskell-language-server/issues/4046 ?

wz1000:
> Is this related to #4046 ?

I don't think so

michaelpj (approved):
Seems reasonable, test case would be great.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -585,9 +585,21 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
+          let (all_targets, this_flags_map, this_options)
+                = case HM.lookup _cfp flags_map' of
+                    Just this -> (all_targets', flags_map', this)
+                    Nothing -> (this_target_details : all_targets', HM.insert _cfp this_flags flags_map', this_flags)
+                  where all_targets' = concat all_target_details
+                        flags_map' = HM.fromList (concatMap toFlagsMap all_targets')
+                        this_target_details = TargetDetails (TargetFile _cfp) this_error_env this_dep_info [_cfp]
+                        this_flags = (this_error_env, this_dep_info)
+                        this_error_env = ([this_error], Nothing)
+                        this_error = ideErrorWithSource (Just "crade") (Just DiagnosticSeverity_Error) _cfp
```
fendor, line 598:
```suggestion
                        this_error = ideErrorWithSource (Just "cradle") (Just DiagnosticSeverity_Error) _cfp
```

fendor (approved):
LGTM

PR #4098, opened 2024-02-27
Remove more workarounds for GHCs < 9.2 (#4092)

jhrcek:
Further cleanups based on the fact that we can assume `ghcVersion >= GHC92`


ghcide/test/data/plugin-recorddot/RecordDot.hs
```diff
@@ -1,6 +0,0 @@
-{-# LANGUAGE DuplicateRecordFields, TypeApplications, TypeFamilies, UndecidableInstances, FlexibleContexts, DataKinds, MultiParamTypeClasses, TypeSynonymInstances, FlexibleInstances #-}
```
jhrcek, line 1: Removing this because one of the tests being removed says
`ignoreForGHC92Plus "No need for this plugin anymore!"`

michaelpj (approved):
great

PR #4103, opened 2024-02-29
remove non-ascii name

### LINKED ISSUE #4102: Non-ascii module names break `cabal sdist`
cabal-install uses the `tar` package, which just recently got proper unicode support.

The following module breaks `cabal sdist`: https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-semantic-tokens-plugin/test/testdata/TModula%F0%90%90%80bA.hs

It will truncate the module name to `TModula`:

```
$ tar -tvf dist-newstyle/sdist/haskell-language-server-2.7.0.0.tar.gz | grep plugins/hls-semantic-tokens-plugin/test/testdata/            Thu 11:32
drwxr-xr-x 0/0               0 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/
-rw-r--r-- 0/0             722 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/T1.hs
-rw-r--r-- 0/0              59 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TClass.hs
-rw-r--r-- 0/0             277 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TClassImportedDeriving.hs
-rw-r--r-- 0/0             295 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TDataFamily.hs
-rw-r--r-- 0/0              57 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TDataType.hs
-rw-r--r-- 0/0              75 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TDatatypeImported.hs
-rw-r--r-- 0/0              84 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TDoc.hs
-rw-r--r-- 0/0              76 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TFunction.hs
-rw-r--r-- 0/0              57 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TFunctionLet.hs
-rw-r--r-- 0/0              77 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TFunctionLocal.hs
-rw-r--r-- 0/0             125 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TFunctionUnderTypeSynonym.hs
-rw-r--r-- 0/0             236 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TGADT.hs
-rw-r--r-- 0/0             104 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TInstanceClassMethodBind.hs
-rw-r--r-- 0/0              52 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TInstanceClassMethodUse.hs
-rw-r--r-- 0/0              63 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TModula
-rw-r--r-- 0/0             123 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TModuleB.hs
-rw-r--r-- 0/0              74 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TNoneFunctionWithConstraint.hs
-rw-r--r-- 0/0             306 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TOperator.hs
-rw-r--r-- 0/0              50 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TPatternMatch.hs
-rw-r--r-- 0/0              82 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TPatternSynonym.hs
-rw-r--r-- 0/0              57 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TPatternbind.hs
-rw-r--r-- 0/0             142 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TQualifiedName.hs
-rw-r--r-- 0/0              56 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TRecord.hs
-rw-r--r-- 0/0             122 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TRecordDuplicateRecordFields.hs
-rw-r--r-- 0/0             113 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TTypefamily.hs
-rw-r--r-- 0/0              49 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TUnicodeSyntax.hs
-rw-r--r-- 0/0              73 2001-09-09 09:46 haskell-language-server-2.7.0.0/plugins/hls-semantic-tokens-plugin/test/testdata/TValBind.hs
```

@fendor @soulomoon

soulomoon:
Fix #4102
Since hackage does not accept non-ascii filename.

### DIFF
--- a/plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
+++ b/plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
@@ -178,7 +178,7 @@ semanticTokensFullDeltaTests :: TestTree
 semanticTokensFullDeltaTests =
   testGroup "semanticTokensFullDeltaTests" $
     [ testCase "null delta since unchanged" $ do
-        let file1 = "TModula𐐀bA.hs"
+        let file1 = "TModuleA.hs"
         let expectDelta = InR (InL (SemanticTokensDelta (Just "1") []))
         Test.Hls.runSessionWithServerInTmpDir def semanticTokensPlugin (mkFs $ FS.directProjectMulti [file1]) $ do
           doc1 <- openDoc file1 "haskell"
@@ -188,7 +188,7 @@ semanticTokensFullDeltaTests =
           liftIO $ delta @?= expectDelta
 
       , testCase "add tokens" $ do
-        let file1 = "TModula𐐀bA.hs"
+        let file1 = "TModuleA.hs"
         let expectDelta = InR (InL (SemanticTokensDelta (Just "1") [SemanticTokensEdit 20 0 (Just [2,0,3,8,0])]))
         --                                                                                         r c l t m
         --                                      where r = row, c = column, l = length, t = token, m = modifier
@@ -207,7 +207,7 @@ semanticTokensFullDeltaTests =
           liftIO $ delta @?= expectDelta
 
       , testCase "remove tokens" $ do
-        let file1 = "TModula𐐀bA.hs"
+        let file1 = "TModuleA.hs"
         let expectDelta = InR (InL (SemanticTokensDelta (Just "1") [SemanticTokensEdit 0 20 (Just [])]))
         -- delete all tokens
         Test.Hls.runSessionWithServerInTmpDir def semanticTokensPlugin (mkFs $ FS.directProjectMulti [file1]) $ do
@@ -229,7 +229,7 @@ semanticTokensTests :: TestTree
 semanticTokensTests =
   testGroup "other semantic Token test" $
     [ testCase "module import test" $ do
-        let file1 = "TModula𐐀bA.hs"
+        let file1 = "TModuleA.hs"
         let file2 = "TModuleB.hs"
         Test.Hls.runSessionWithServerInTmpDir def semanticTokensPlugin (mkFs $ FS.directProjectMulti [file1, file2]) $ do
           doc1 <- openDoc file1 "haskell"
@@ -245,14 +245,14 @@ semanticTokensTests =
 
           result <- docSemanticTokensString def doc2
           let expect = unlines [
-                    "3:8-18 TModule \"TModula\\66560bA\""
-                    , "4:18-28 TModule \"TModula\\66560bA\""
+                    "3:8-16 TModule \"TModuleA\""
+                    , "4:18-26 TModule \"TModuleA\""
                     , "6:1-3 TVariable \"go\""
                     , "6:6-10 TDataConstructor \"Game\""
                     , "8:1-5 TVariable \"a\\66560bb\""
-                    , "8:8-19 TModule \"TModula\\66560bA.\""
-                    , "8:19-22 TRecordField \"a\\66560b\""
-                    , "8:23-25 TVariable \"go\""
+                    , "8:8-17 TModule \"TModuleA.\""
+                    , "8:17-20 TRecordField \"a\\66560b\""
+                    , "8:21-23 TVariable \"go\""
                 ]
           liftIO $ result @?= expect,
       goldenWithSemanticTokensWithDefaultConfig "mixed constancy test result generated from one ghc version" "T1",
--- a/plugins/hls-semantic-tokens-plugin/test/testdata/TModuleA.hs
+++ b/plugins/hls-semantic-tokens-plugin/test/testdata/TModuleA.hs
@@ -1,4 +1,4 @@
-module TModula𐐀bA where
+module TModuleA where
 
 data Game = Game {a𐐀b :: Int}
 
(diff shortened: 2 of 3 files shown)


fendor (approved):

hasufell:
this should be backported to the 2.7.0.0 release branch

hasufell:
> Since hackage does not accept non-ascii filename.

No, because it's a bad idea. There's a reason hackage rejects it. It's not portable and tar doesn't specify what to do with non-ascii characters. That means they land uninterpreted on your hard drive (and possibly incompatible with your local encoding). If you go the interpretation route (always store in tar as UTF-8, then try to convert to local encoding), you'll get into all sorts of other problems (no byte equivalence of filepaths etc.).

fendor:
> this should be backported to the 2.7.0.0 release branch

It is not necessary for the HLS release to Hackage, `cabal sdist all` seems to "conveniently" ignore that module already.
Also, it seems like we were missing a couple dozens of `extra-source-files`, I have added #4105 to fix that.

hasufell:
> cabal sdist all seems to "conveniently" ignore that module already.

How so? My tests show that it incorrectly truncates. That will cause failure to run `cabal test`.

fendor:
> That will cause failure to run cabal test.

Yeah that is true. It seems we have been missing a couple of `extra-source-files` and most of our releases can't be tested from a `cabal get`

hasufell:
> > That will cause failure to run cabal test.
> 
> Yeah that is true. It seems we have been missing a couple of `extra-source-files` and most of our releases can't be tested from a `cabal get`.

For the plugins, you have `plugins/**/*.hs` in `extra-source-files`, which includes all the test data. If you look at my post [here](https://github.com/haskell/haskell-language-server/issues/4102#issue-2160935504) you can see the evidence that cabal is not ignoring the module, but storing it as `TModula` without a file extension.

fendor:
Indeed, it truncates, but it looks like  `cabal get` not `cabal sdist` don't choke on the module name.

However, the tests will definitely fail.

hasufell:
> Indeed, it truncates, but it looks like cabal get not cabal sdist don't choke on the module name.

Well, `cabal get` just unpacks the already "malformed" (filename truncated) tar archive. If you manage to upload a unicode filename containing tar archive to hackage somehow, it'll truncate on unpacking.

This would become an absolute nightmare if hackage allowed unicode filenames and some cabal versions handle it correctly and others not.

fendor:
> This would become an absolute nightmare if hackage allowed unicode filenames and some cabal versions handle it correctly and others not.

I agree, and we are getting rid of it!

soulomoon:
> > Since hackage does not accept non-ascii filename.
> 
> 
> 
> No, because it's a bad idea. There's a reason hackage rejects it. It's not portable and tar doesn't specify what to do with non-ascii characters. That means they land uninterpreted on your hard drive (and possibly incompatible with your local encoding). If you go the interpretation route (always store in tar as UTF-8, then try to convert to local encoding), you'll get into all sorts of other problems (no byte equivalence of filepaths etc.).

Thank for the explanation, so the universal support for utf-8 for filename just sucks :(

PR #4107, opened 2024-02-29
Don't distribute gifs or plugin readmes

fendor:


fendor:
As discussed with @michaelpj

michaelpj (approved):

PR #3748, opened 2023-08-08
Prepare release 2.1.0.0

fendor:
## ChangeLog 2.1.0.0

* Binaries for GHC 9.4.6
* Completions for .cabal files
* Performance improvements
* Show package name and its version while hovering on import statements ([#3691](https://github.com/haskell/haskell-language-server/pull/3691))
* Fix code edits in lsp spec compliant editors like helix. ([#3643](https://github.com/haskell/haskell-language-server/pull/3643))

### Pull requests merged

- Update to latest lsp packages
  ([#3747](https://github.com/haskell/haskell-language-server/pull/3747)) by @joyfulmantis
- Remove unnecessary allow-newer in stack.yaml
  ([#3746](https://github.com/haskell/haskell-language-server/pull/3746)) by @July541
- Log fourmolu and ormolu version that hls using
  ([#3744](https://github.com/haskell/haskell-language-server/pull/3744)) by @July541
- Various PluginError PR suggestions I missed earlier
  ([#3737](https://github.com/haskell/haskell-language-server/pull/3737)) by @joyfulmantis
- Add resolve support in refine imports by merging it with explicit imports
  ([#3729](https://github.com/haskell/haskell-language-server/pull/3729)) by @joyfulmantis
- Fix other file goto definition
  ([#3725](https://github.com/haskell/haskell-language-server/pull/3725)) by @nlander
- Fix Nix builds
  ([#3724](https://github.com/haskell/haskell-language-server/pull/3724)) by @cydparser
- Better plugin error infrastructure
  ([#3717](https://github.com/haskell/haskell-language-server/pull/3717)) by @joyfulmantis
- Move Recorder to hls-plugin-api
  ([#3714](https://github.com/haskell/haskell-language-server/pull/3714)) by @joyfulmantis
- Actually force usages
  ([#3713](https://github.com/haskell/haskell-language-server/pull/3713)) by @wz1000
- Best-effort support of Qualified Imports in GHC 9.4
  ([#3712](https://github.com/haskell/haskell-language-server/pull/3712)) by @konn
- Skip test if only CODEOWNERS changed
  ([#3707](https://github.com/haskell/haskell-language-server/pull/3707)) by @July541
- Update stack stuff
  ([#3706](https://github.com/haskell/haskell-language-server/pull/3706)) by @July541
- Mark hls-floskell-plugin as tier 3
  ([#3705](https://github.com/haskell/haskell-language-server/pull/3705)) by @July541
- Remove isovector as an owner
[...]

PR #4110, opened 2024-03-01
Exit with non-zero exitcode if wrapper fails to launch

fendor:
As reported by https://github.com/haskell/haskell-language-server/pull/3748#discussion_r1508794998, the tests are not run correctly, since there is no GHC installed. We install the recommended GHC version to make sure one (hopefully) supported GHC version is installed for running the wrapper test.

Further, if setup fails, HLS-wrapper launches an LSP server that allows users to restart the server once they have fixed potential setup issues. That's only desired if we are indeed launching as an LSP server. Thus, we exit with a non-zero exit code if setup fails and we are not launching in LSP mode.

### DIFF
--- a/.github/scripts/test.sh
+++ b/.github/scripts/test.sh
@@ -48,6 +48,8 @@ test_all_hls() {
             fi
         fi
     done
+    # install the recommended GHC version so the wrapper can launch HLS
+    ghcup install ghc --set recommended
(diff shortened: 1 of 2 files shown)


hasufell:
I can't really judge whether this is correct. Have you tested it?

fendor:
I have tested the Wrapper change, with 2.6.0.0, running the wrapper launches the LSP server if project setup fails. Now, it doesn't :)

hasufell (approved):

michaelpj (approved):

Issue #3915, opened 2023-12-19
Comments renaming should not be allowed


Alecton4:
### Your environment

<details>

#### Which OS do you use?

Operating System: openSUSE Tumbleweed 20231206
KDE Plasma Version: 5.27.9
KDE Frameworks Version: 5.112.0
Qt Version: 5.15.11
Kernel Version: 6.6.3-1-default (64-bit)
Graphics Platform: X11
#### Which version of GHC do you use and how did you install it?

Output of `haskell-language-server-wrapper --probe-tools`:
```text
haskell-language-server version: 2.4.0.0 (GHC: 9.0.2) (PATH: /home/az/.ghcup/hls/2.4.0.0/lib/haskell-language-server-2.4.0.0/bin/haskell-language-server-wrapper)
Tool versions found on the $PATH
cabal:          3.6.2.1
stack:          2.11.1
ghc:            9.4.7

Tool versions in your project
2023-12-18T06:37:13.171524Z | Debug | executing command: ghc --numeric-version
ghc:            9.4.7
```
Installed via `curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh`
#### How is your project built (alternative: link to the project)?
N/A
#### Which LSP client (editor/plugin) do you use?

VSCode 1.85.0 + `haskell.haskell` 2.4.2
#### Which version of HLS do you use and how did you install it?

See above.
#### Have you configured HLS in any way (especially: a `hie.yaml` file)?
No.

</details>

### Steps to reproduce

I discovered the issue accidentally when using a [spell checker](https://github.com/streetsidesoftware/vscode-spell-checker) to correct a typo in my comment. But it caused an error message from HLS. These are the steps to reproduce: 

1. Open a `.hs` file and place the cursor on a comment.
2. Press `F2`.
3. The word will be prompted to rename.
4. Enter a new word and press `Enter`, error occurs:

   ![image](https://github.com/haskell/vscode-haskell/assets/43232804/ec83cba6-82ad-40b6-ba04-658cda734877)

### Expected behaviour

Renaming comments should never be allowed, i.e, pressing `F2` on comments should have no effect at all. 

### Actual behaviour

There is a renaming prompt but renaming will eventually cause error.

### Include debug information

Execute in the root of your project the command `haskell-language-server-wrapper --debug .` and paste the logs here (you can find the executable location [here](https://github.com/haskell/vscode-haskell#downloaded-binaries)):

<details>
<summary>
[...]

PR #4111, opened 2024-03-02
Improve handling of nonsense rename attempts

jhrcek:
First step towards improving behavior when renaming unsupported things (e.g. keywords or words within comments).
Instead of throwing plugin error (reporte here https://github.com/haskell/haskell-language-server/issues/3915)
we detect that there is no Name to rename under cursor and throw less scary looking Invalid Param warning.

Behavior after this PR:

Peek 2024-03-02 08-30

EDIT: Actually in the end I included @michaelpj's suggestion to implement [Prepare Rename](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename) handler that "prefilters" invalid requests on the client and doesn't even allow creating rename request for them.

### DIFF
--- a/hls-plugin-api/src/Ide/Types.hs
+++ b/hls-plugin-api/src/Ide/Types.hs
@@ -475,6 +475,9 @@ instance PluginMethod Request Method_CodeLensResolve where
 instance PluginMethod Request Method_TextDocumentRename where
   handlesRequest = pluginEnabledWithFeature plcRenameOn
 
+instance PluginMethod Request Method_TextDocumentPrepareRename where
+  handlesRequest = pluginEnabledWithFeature plcRenameOn
+
 instance PluginMethod Request Method_TextDocumentHover where
   handlesRequest = pluginEnabledWithFeature plcHoverOn
 
@@ -599,7 +602,7 @@ class PluginMethod Request m => PluginRequestMethod (m :: Method ClientToServer 
 ---
 instance PluginRequestMethod Method_TextDocumentCodeAction where
   combineResponses _method _config (ClientCapabilities _ textDocCaps _ _ _ _) (CodeActionParams _ _ _ _ context) resps =
-      InL $ fmap compat $ filter wasRequested $ concat $ mapMaybe nullToMaybe $ toList resps
+      InL $ fmap compat $ concatMap (filter wasRequested) $ mapMaybe nullToMaybe $ toList resps
     where
       compat :: (Command |? CodeAction) -> (Command |? CodeAction)
       compat x@(InL _) = x
@@ -656,6 +659,10 @@ instance PluginRequestMethod Method_CodeLensResolve where
     combineResponses _ _ _ _ (x :| _) = x
 
 instance PluginRequestMethod Method_TextDocumentRename where
+
+instance PluginRequestMethod Method_TextDocumentPrepareRename where
+    -- TODO more intelligent combining?
+    combineResponses _ _ _ _ (x :| _) = x
 
 instance PluginRequestMethod Method_TextDocumentHover where
   combineResponses _ _ _ _ (mapMaybe nullToMaybe . toList -> hs :: [Hover]) =
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
@@ -57,43 +57,49 @@ import           Language.LSP.Server
 instance Hashable (Mod a) where hash n = hash (unMod n)
 
 descriptor :: Recorder (WithPriority E.Log) -> PluginId -> PluginDescriptor IdeState
-descriptor recorder pluginId = mkExactprintPluginDescriptor recorder $ (defaultPluginDescriptor pluginId "Provides renaming of Haskell identifiers")
-    { pluginHandlers = mkPluginHandler SMethod_TextDocumentRename renameProvider
-    , pluginConfigDescriptor = defaultConfigDescriptor
-        { configCustomConfig = mkCustomConfig properties }
-    }
+descriptor recorder pluginId = mkExactprintPluginDescriptor recorder $
+    (defaultPluginDescriptor pluginId "Provides renaming of Haskell identifiers")
+        { pluginHandlers = mkPluginHandler SMethod_TextDocumentRename renameProvider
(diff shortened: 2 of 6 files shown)


plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -57,43 +57,49 @@ import           Language.LSP.Server
...
+descriptor recorder pluginId = mkExactprintPluginDescriptor recorder $
+    (defaultPluginDescriptor pluginId "Provides renaming of Haskell identifiers")
+        { pluginHandlers = mkPluginHandler SMethod_TextDocumentRename renameProvider
+        , pluginConfigDescriptor = defaultConfigDescriptor
+            { configCustomConfig = mkCustomConfig properties }
+        }
 
 renameProvider :: PluginMethodHandler IdeState Method_TextDocumentRename
 renameProvider state pluginId (RenameParams _prog (TextDocumentIdentifier uri) pos newNameText) = do
-        nfp <- getNormalizedFilePathE uri
```
jhrcek, line 68: Most of this is indentation "fixes", recommended to review with "Hide whitespace"
Peek 2024-03-02 08-36

jhrcek:
Actually it was relatively easy to implement PrepareRename handler and the behavior is now even better with it:

PrepareRename

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -53,47 +53,64 @@ import qualified Language.LSP.Protocol.Lens            as L
...
+              , mkPluginHandler SMethod_TextDocumentPrepareRename prepareRenameProvider
+              ]
+        , pluginConfigDescriptor = defaultConfigDescriptor
+            { configCustomConfig = mkCustomConfig properties }
+        }
+
+prepareRenameProvider :: PluginMethodHandler IdeState Method_TextDocumentPrepareRename
+prepareRenameProvider state _pluginId (PrepareRenameParams (TextDocumentIdentifier uri) pos _progressToken) = do
+    nfp <- getNormalizedFilePathE uri
+    namesUnderCursor <- getNamesAtPos state nfp pos
```
jhrcek, line 74: This is just an initial "approximation" to easily filter out most of things that can't be renamed.
I decided not to follow Rename handler's logic to filter out other cases (e.g. renaming of exported symbols) as I think it's better if we give explicit warning in such cases (renaming of exported symbols is not supported) rather than "silently" prevent such renames using this handler.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -657,6 +660,10 @@ instance PluginRequestMethod Method_CodeLensResolve where
 
 instance PluginRequestMethod Method_TextDocumentRename where
 
+instance PluginRequestMethod Method_TextDocumentPrepareRename where
+    -- TODO more intelligent combining?
+    combineResponses _ _ _ _ (x :| _) = x
```
soulomoon, line 665: I think it is fine we only pick up the newest

jhrcek: I think this is used combining or responses in case multiple plugins happen to handler this type of request.
I'd say it's ok to leave it like this for now, because this PR introduces this new request type and for now rename plugin is the only one that supports this kind of request.

fendor (approved):
Nice job!

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -57,43 +58,66 @@ import           Language.LSP.Server
...
+    -- When this handler says that rename is invalid, VSCode shows "The element can't be renamed"
+    -- and doesn't even allow you to create full rename request.
+    -- This handler deliberately approximates "things that definitely can't be renamed"
+    -- to mean "there is no Name at given position".
+    --
+    -- In particular it allows some cases through (e.g. cross-module renames),
+    -- so that the full rename handler can give more informative error about them.
+    let renameValid = not $ null namesUnderCursor
+    pure $ InL $ PrepareRenameResult $ InR $ InR $ #defaultBehavior .== renameValid
```
michaelpj, line 83: I'm unsure if this is the correct behaviour. It's very hard to tell what the intention of the various options for results of `prepareRename` are. Worth a comment anyway. We could consider returning the range corresponding to the name, although it's unclear if we have that...

jhrcek: I would say that [lsp spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename) is quite clear in this regard:

>  If { defaultBehavior: boolean } is returned (since 3.16) the rename position is valid and the client should use its default behavior to compute the rename range.

I looked at what the vscode-languageclient is doing with that response (see [here](https://github.com/microsoft/vscode-languageserver-node/blob/46f6f9038fee60dbdfeb5307201397adc8feb3cb/client/src/common/rename.ts#L94-L108)) and when you return the boolean it does:
if true it triggers the default rename behavior (open input box where you can specify the new name)
if false it gives you a small error popup saying "The element can't be renamed"

I'll let you read this / react and will can open a PR with some additional clarifying comment later today.

michaelpj: I guess I'm just wondering if we can rely on the "default behaviour" or whether it would be better for us to be returning the actual range of the name? Maybe it's fine?

jhrcek: I played with it for a bit and it seems it would be hard to get this right if we're to provide the Range under cursor.

The thing I had trouble to make work was qualified names. 
e.g. if you rename `IO.putStrLn` the HieAST contains the range of the whole thing (including `IO.`) so if you trigger rename on it, you get the whole thing prefilled into the rename input which is confusing.

I'd be more comfortable keeping the current implementation which seems to work robustly on all cases we have in tests.

michaelpj: Huh, and vscode does the right thing and just changes `putStrLn`? IDK, seems potentially fragile but maybe it's fine.

Anyway, this all seems like super useful information, so worth writing down!

PR #4112, opened 2024-03-02
Improve isolation of build artefacts of test runs

fendor:
Even though we copy test files into temporary directories, we used to reuse the same cache directory for build artefacts, hiedb and compilation artefacts. While there is practially no chance this causes any issues for the test runs themselves, it litters the cache directory with a lot of files. So, we create one main directory in the temporary directory, and generate all caches and in there. This makes it trivial to delete all test caches, without risking deleting the cache that is still used.

### DIFF
--- a/hls-test-utils/hls-test-utils.cabal
+++ b/hls-test-utils/hls-test-utils.cabal
@@ -46,6 +46,7 @@ library
     , lens
     , lsp-test                ^>=0.17
     , lsp-types               ^>=2.1
+    , safe-exceptions
     , tasty
     , tasty-expected-failure
     , tasty-golden
--- a/hls-test-utils/src/Test/Hls.hs
+++ b/hls-test-utils/src/Test/Hls.hs
@@ -63,9 +63,9 @@ where
 import           Control.Applicative.Combinators
 import           Control.Concurrent.Async           (async, cancel, wait)
 import           Control.Concurrent.Extra
-import           Control.Exception.Base
+import           Control.Exception.Safe
 import           Control.Lens.Extras                (is)
-import           Control.Monad                      (guard, unless, void)
+import           Control.Monad                      (guard, unless, void, when)
 import           Control.Monad.Extra                (forM)
 import           Control.Monad.IO.Class
 import           Data.Aeson                         (Result (Success),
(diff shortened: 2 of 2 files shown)


hls-test-utils/hls-test-utils.cabal
```diff
@@ -46,6 +46,7 @@ library
     , lens
     , lsp-test                ^>=0.17
     , lsp-types               ^>=2.1
+    , safe-exceptions
```
jhrcek, line 49: Seems you're not using anything from this package in this PR. Is it needed?

hls-test-utils/src/Test/Hls.hs
```diff
@@ -447,6 +452,32 @@ runSessionWithServerInTmpDir' plugins conf sessConf caps tree act = withLock loc
...
+setupTestEnvironment = do
+  tmpDirRoot <- getTemporaryDirectory
+  let testRoot = tmpDirRoot </> "hls-test-root"
+      testCacheDir = testRoot </> ".cache"
+  createDirectoryIfMissing testRoot
+  createDirectoryIfMissing testCacheDir
+  setEnv "XDG_CACHE_HOME" testCacheDir
+  pure testRoot
+  where
+    createDirectoryIfMissing fp = do
```
jhrcek, line 476: Why not use [createDirectoryIfMissing](https://hackage.haskell.org/package/directory-1.3.8.3/docs/System-Directory.html#v:createDirectoryIfMissing) from directory package? It can also create intermediate dirs on path, so can be called just once..

fendor: That's a good suggestion :D My brain didn't want to see that function

hls-test-utils/hls-test-utils.cabal
fendor: I wanted to use `finally` from `safe-exceptions`, but missed the import `Control.Concurrent.Base`. Thanks for catching this!

jhrcek (approved):
LGTM

hls-test-utils/src/Test/Hls.hs
```diff
@@ -447,6 +451,26 @@ runSessionWithServerInTmpDir' plugins conf sessConf caps tree act = withLock loc
...
+-- Additionally, this overwrites the 'XDG_CACHE_HOME' variable to isolate
+-- the tests from existing caches. 'hie-bios' and 'ghcide' honour the
+-- 'XDG_CACHE_HOME' environment variable and generate their caches there.
+setupTestEnvironment :: IO FilePath
+setupTestEnvironment = do
+  tmpDirRoot <- getTemporaryDirectory
+  let testRoot = tmpDirRoot </> "hls-test-root"
+      testCacheDir = testRoot </> ".cache"
+  createDirectoryIfMissing False testRoot
+  createDirectoryIfMissing False testCacheDir
```
jhrcek, line 471: Nitpick: I think you can replace these 2 with just 
```suggestion
  createDirectoryIfMissing True testCacheDir
```

fendor: I wanted to make sure that `$TMPDIR` exists on disk before running the tests (`getTemporaryDirectory` honour the `$TMPDIR` environment variable). I am not sure why I think that's a good idea, if you think this is unhelpful, I am happy to update :)

jhrcek: I don't insist, keep it as is if you think it's better in some way.

michaelpj (approved):

Issue #3500, opened 2023-02-21
Eval-plugin: Evaluate `-- >>>` does not work when there is one or more newlines at end of the file


turbotimon:
### Your environment

Which OS do you use? Tested on Windows, Ubuntu and MacOS

Which version of GHC do you use and how did you install it?

```bash
ghcup list
[ Info  ] downloading: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml as file C:\ghcup\cache\ghcup-0.0.7.yaml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed  
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0 
[ Warn  ] New ghc version available. If you want to install this latest version, run 'ghcup install ghc 9.4.4'      
[ Warn  ] New cabal version available. If you want to install this latest version, run 'ghcup install cabal 3.8.1.0'
[ Warn  ] New stack version available. If you want to install this latest version, run 'ghcup install stack 2.9.3'
[ Info  ] verifying digest of: gs.exe
   Tool  Version  Tags                      Notes      
X  ghc   7.10.3   base-4.8.2.0
X  ghc   8.0.2    base-4.9.1.0
X  ghc   8.2.2    base-4.10.1.0
X  ghc   8.4.1    base-4.11.0.0
X  ghc   8.4.2    base-4.11.1.0
X  ghc   8.4.3    base-4.11.1.0
X  ghc   8.4.4    base-4.11.1.0
X  ghc   8.6.1    base-4.12.0.0
X  ghc   8.6.2    base-4.12.0.0
X  ghc   8.6.3    base-4.12.0.0
X  ghc   8.6.4    base-4.12.0.0
X  ghc   8.6.5    base-4.12.0.0
X  ghc   8.8.1    base-4.13.0.0
X  ghc   8.8.2    base-4.13.0.0
X  ghc   8.8.3    base-4.13.0.0
X  ghc   8.8.4    base-4.13.0.0
X  ghc   8.10.1   base-4.14.0.0
X  ghc   8.10.2   base-4.14.1.0
X  ghc   8.10.3   base-4.14.1.0
X  ghc   8.10.4   base-4.14.1.0
X  ghc   8.10.5   base-4.14.2.0
X  ghc   8.10.6   base-4.14.3.0
[...]

Issue #3312, opened 2022-11-01
Eval plugin appears to stop working after a function def that has more than one pattern match


googleson78:
#### Which OS do you use?
NixOS

#### Which version of GHC do you use and how did you install it?
```
[nix-shell:/tmp]$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.2.4
```
via nix

#### How is your project built (alternative: link to the project)?
No project

##### Which LSP client (editor/plugin) do you use?
Neovim+LanguageClient-neovim

#### Which version of HLS do you use and how did you install it?
```
[nix-shell:/tmp]$ haskell-language-server --version
2022-11-01T15:02:05.261746Z | Info | No log file specified; using stderr.
haskell-language-server version: 1.8.0.0 (GHC: 9.2.4) (PATH: /nix/store/d717g1x6mr0310lijj6671pws1k5cvn7-haskell-language-server-1.8.0.0/bin/.haskell-language-server-9.2.4-unwrapped)
```
via `nix-shell -p haskell.packages.ghc924.haskell-language-server`

#### Have you configured HLS in any way (especially: a `hie.yaml` file)?
No

### Steps to reproduce
For the following file
```haskell
-- >>> 5
-- 5
bla :: Bool -> Bool
bla True = True
bla False = False
```
The `Evaluate...` code lens pops up. However, for the following file
```haskell
bla :: Bool -> Bool
bla True = True
bla False = False
-- >>> 5
```
it doesn't. 

The thing that triggers this issue appears to be any multiline (more than one match) function definition.

This issue does not reproduce if we're using an hls built with 9.0.2

googleson78:
Here're the logs from the language server and from the client from the following experiment:

0. Start with the file that has `Evaluate`.
1. Wait for `Evaluate...` to pop up
2. Move the `-- >>> 5` below the function.
3. Wait some time.
4. Close the file

[LanguageClient.log](https://github.com/haskell/haskell-language-server/files/9911553/LanguageClient.log)
[LanguageServer.log](https://github.com/haskell/haskell-language-server/files/9911554/LanguageServer.log)

turbotimon:
Maybe it't the same problem? #3500+

PiDelport:
This looks like the same issue as:

- https://github.com/haskell/haskell-language-server/issues/3129

jfong2702:
I am having the same issue what was the resolution?

googleson78:
[...]

PR #4113, opened 2024-03-03
eval: more robust way to extract comments from ParsedModule

jhrcek:
Fixes https://github.com/haskell/haskell-language-server/issues/3500 and https://github.com/haskell/haskell-language-server/issues/3312

The problem in those issues can be illustrated by this simpler reproducer.
Peek 2024-03-03 13-47
The issue is that the "end of line comment" is associated with the 2nd match group of function definition (confirmed this by compiling with `-ddump-parsed-ast -dkeep-comments`), which was part of AST  that eval plugin was ignoring when extracting comments, thus it didn't create "Evaluate..." code lens.

By using `biplate` combinator from lens we can more simply extract all occurrences of LEpaComment from the parsed source, wherever they might be.


fendor (approved):
LGTM :)

PR #4121, opened 2024-03-07
Remove bitrotted CPP gated code

RaoulHC:
A lot of the HLINT_ON_GHC_LIB gated code has been bitrotting since this flag was removed. This could be reintroduced if we wanted to directly work on the same parsed AST, but as the hlint ghc plugin showed this may not make much difference: https://www.haskellforall.com/2023/09/ghc-plugin-for-hlint.html

Split out from #3757 as this is less controversial.


michaelpj:
merged the other PR

PR #4124, opened 2024-03-09
Integrate stylish-haskell into hls executable with ghc 9.8

jhrcek:
The version of stylish-haskell where @michaelpj added support for ghc 9.8 is on hackage:
https://hackage.haskell.org/package/stylish-haskell-0.14.6.0/changelog
Let's enable it in hls!

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -8,7 +8,7 @@ packages:
          ./hls-plugin-api
          ./hls-test-utils
 
-index-state: 2024-02-25T00:00:00Z
+index-state: 2024-03-09T08:17:00Z
 
 tests: True
 test-show-details: direct
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -1433,7 +1433,7 @@ library hls-stylish-haskell-plugin
     , hls-plugin-api   == 2.7.0.0
     , lsp-types
     , mtl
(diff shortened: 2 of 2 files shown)


cabal.project
```diff
@@ -8,7 +8,7 @@ packages:
          ./hls-plugin-api
          ./hls-test-utils
 
-index-state: 2024-02-25T00:00:00Z
+index-state: 2024-03-09T08:17:00Z
```
jhrcek, line 11: Bumping index state just to avoid this warning:
```
Warning: Requested index-state 2024-02-25T00:00:00Z is newer than 
'hackage.haskell.org'! Falling back to older state (2024-02-24T17:42:44Z).
```

fendor (approved):
I am a little bit confused, why didn't CI report the command `cabal test hls-stylish-haskell` plugin as failing? https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/test.yml#L170

jhrcek:
Hmm, good question.
What motivated me to open this PR was that I was playing with ghc 9.8 and formatting hs files didn't do anything, so I thought "maybe stylish-haskell is not yet enabled", so I looked at hls cabal file and saw `if impl(ghc < 9.8.0)` and just removed it.
After closer look I see that this conditional is part of `common stylishHaskell` pragma, which is not used anywhere :smile:

michaelpj:
Oops, I thought I had done this :facepalm:

PR #4125, opened 2024-03-09
Enable floskell and hlint plugins for ghc 9.8

jhrcek:
Followup on https://github.com/haskell/haskell-language-server/pull/4124
two more plugins that I think we can now allow to be integrated into hls with ghc 9.8


fendor (approved):

fendor (approved):
Thanks!

michaelpj (approved):
Again, argh. I think I even updated the docs incorrectly :(

PR #4126, opened 2024-03-09
hls-notes-plugin: Initial implementation

jvanbruegge:
For some reason I could not reopen the other pull request, so I have to open a new one (#3629). Sorry there were a lot of real live issues compounding a few days ago and closing this was just irrational.

In case you are still interested in this, I would still like to have this merged and would maintain it in the future.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,7 +74,7 @@ jobs:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
-        test: 
+        test:
           - true
           - false
         exclude:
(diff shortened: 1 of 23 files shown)


fendor (approved):
Hi, thank you for coming back for this :) 

I am wondering, how expensive the note generation is on a huge code base such as GHC. 
I am a little bit inclined to disable this plugin by default, to avoid performance regressions, but given that the rule is only triggered on a `Goto Definition` request, it might be irrelevant.

Also, you should have to add the configuration for the notes plugin to the configuration integration tests: https://github.com/haskell/haskell-language-server/tree/master/test/testdata/schema and https://github.com/haskell/haskell-language-server/blob/master/test/functional/ConfigSchema.hs, I don't understand why the CI is green wthout that 0.o

jvanbruegge:
I don't think we need to disable this by default because as you said the notes are only searched for when you do `goto definition` **and** your cursor is currently on a note reference. Otherwise nothing happens.
If you use this in GHC, the first time you go to a note it takes a few seconds to index, every subsequent note jump is fast

jvanbruegge:
For the configuration, is that something that I also need to handle in the plugin itself? Or does the hls infrastructure take care of that?

fendor:
> For the configuration, is that something that I also need to handle in the plugin itself? Or does the hls infrastructure take care of that?

You should have to add entries to the respective json file, so that I don't forget to update the vscode config during the next release. But we check that in CI, so I don't understand what is going wrong right now.

fendor:
You are good to go, I documented the bug at #4127

jhrcek (approved):
LGTM

jvanbruegge:
thanks :)

fendor:
Congrats, sorry it took so long :)

jhrcek:
Hello @jvanbruegge 
today I rebuilt my local hls with this change merged and I'm running into this issue:

The new notes plugin seems too eager to interpret everything as a note 😅
I just want to ctrl+v some text..
Peek 2024-03-12 07-13

Would you have time to look into this and fix this behavior?
If not, I can take a look.
What I would expect is some kind of silent failure if there is no note at given position, as opposed to error being shown to client.
Now that fendor gave you write access to the repo, you could also join the haskell-language-server matrix channel if you prefer more interactive communication.

jvanbruegge:
I agree this should be changed, I'm on it. Have you mapped go-to-definition to ctrl? Also how do I join the matrix room?

jhrcek:
I didn't do any keybinding changes in vscode, Seems like ctrl+click is the default way in vscode to trigger "go to definition"
https://code.visualstudio.com/docs/editor/editingevolved#_go-to-definition

How to join the matrix room:
You can use this link https://matrix.to/#/#haskell-language-server:matrix.org
I randomly picked browser-based app element as my matrix client and joined using that.

fendor:
@jvanbruegge Just want to let you know, this plugin is awesome! It is awesome to explore a codebase using `Goto Definition` for the notes. It also motivates me to write more notes, as they are trivial to discover now and make it much easier to document architectural decisions.
This message is just to let you know, I appreciate this plugin a lot, I deeply underestimated how useful it would be!

jvanbruegge:
thanks :)
Then it does what I hoped to achieve.

PR #4128, opened 2024-03-10
Use Set.member instead of Foldable.elem

jhrcek:
`Set.member` is generally more efficient because it can use Ord constraint unlike `Foldable.elem` which can only use Eq class and has to traverse the whole thing if the element is not present.

using Set/Map.singleton is less important but I still find it useful because it reduces visual noise.


fendor (approved):
LGTM, thanks!

Issue #4036, opened 2024-02-02
Ghcide: ExceptionTests are incorrect


fendor:
All exception tests are run in the `ghcide` directory, inferring a `cabal` cradle and failing, as the test files are not part of ghcide.
The tests themselves do not actually test what they set out to and merely increase the test suite execution time.

fendor:
This issue shows that we should unify the test util infrastructure of ghcide and plugins.

PR #4129, opened 2024-03-10
Fix references to old names

fendor:
See https://github.com/haskell/haskell-language-server/blob/master/haskell-language-server.cabal#L1821


fendor:
Genuine bugs uncovered :)

michaelpj:
How did this not just break before??

michaelpj:
The horror :scream:

fendor:
A lot of our tests are running (or not running) but failing or not failing for the wrong reasons, also see #4130 and #4036.

PR #4135, opened 2024-03-11
improve logging in semantic tokens rule

soulomoon:
Switch `use_` to `useE` to capture meaningful error.


michaelpj:
Do we need to do this manually everywhere? What does this do, concretely?

michaelpj:
(I am pretty confused about how all this works tbh!)

soulomoon:
`BadDependency` is swallow here Is around here.

[pure (Nothing, ([ideErrorText file $ T.pack $ show e | not $ isBadDependency e],Nothing))](https://github.com/soulomoon/haskell-language-server/blob/e686a4d5c6c256288822402862a15f7cd58ae4c3/ghcide/src/Development/IDE/Core/Shake.hs#L1208)

PR #4136, opened 2024-03-12
hls-notes-plugin: Do not error if no note is under the cursor

jvanbruegge:
Otherwise this would also trigger when trying to go to a function definition.

### DIFF
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -84,13 +84,15 @@ jumpToNote state _ param
             =<< lift (LSP.getVirtualFile uriOrig)
         line <- err "Line not found in file" (listToMaybe $ Rope.lines $ fst
             (Rope.splitAtLine 1 $ snd $ Rope.splitAtLine (fromIntegral l) contents))
-        note <- err "No note at this position" $ listToMaybe $
-            mapMaybe (atPos $ fromIntegral c) $ matchAllText noteRefRegex line
-        notes <- runActionE "notes.definedNotes" state $ useE MkGetNotes nfp
-        (noteFp, pos) <- err ("Note definition (a comment of the form `{- Note [" <> note <> "]\\n~~~ ... -}`) not found") (HM.lookup note notes)
-        pure $ InL (Definition (InL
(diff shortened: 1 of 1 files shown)


jvanbruegge:
@jhrcek this should do it

fendor (commented):
Perhaps a regression test might be helpful to avoid this in the future?

jvanbruegge:
How can I assert that an error happened? Is there a test I could copy from?

jhrcek:
> How can I assert that an error happened? Is there a test I could copy from?

I did something like that in https://github.com/haskell/haskell-language-server/pull/4111/files#diff-ecd5eff03edc6cac02343f9fe14aeb61fa616f9069e4b3db048d13affb648d17R70-R75

But I think fendor probably means regression test of the shape "if there is no note at given position then the plugin doesn't throw any error". Not how to best test something like that.

Otherwise I agree that this fixes my issues with the plugin.

jhrcek:
I looked into implementing a regression test for this and it's not so easy.
There is a deeper issue in the [NotesTest](https://github.com/haskell/haskell-language-server/blob/82148dc22124dab2f9874e0b61dffeeabd97abe9/plugins/hls-notes-plugin/test/NotesTest.hs), namely that the results that we get from getDefinitions in all the tests are combination of results from hls-notes-plugin and results from Ghcide's [ghcide-hover-and-symbols](https://github.com/haskell/haskell-language-server/blob/82148dc22124dab2f9874e0b61dffeeabd97abe9/ghcide/src/Development/IDE/Plugin/HLS/GhcIde.hs#L36) plugin, which is added to all test sessions by default.

This means that even if the notes plugin throws an error (like " No note at this position" that this PR avoids throwing) we still get a success response in the tests, because based on [this logic](https://github.com/haskell/haskell-language-server/blob/82148dc22124dab2f9874e0b61dffeeabd97abe9/ghcide/src/Development/IDE/Plugin/HLS.hs#L257-L259), if any plugin returns a success response (which in this case ghcide plugin gives success response with empty list of definitions), we just log the error and return the success response.

So I think we should accept this PR without tests because I don't see an easy way to disable that ghcide plugin that masks notes plugin's errors in tests. This will require some discussion and refactoring of test setup logic.

soulomoon:
It should be quite common, and I am not sure even we should be giving out warnning,
since if a click on names other than note, it triggers

jhrcek (approved):
LGTM

PR #4137, opened 2024-03-12
Update contact info in docs

jhrcek:
- don't mention discord (can't find any haskell-tooling channel in there, maybe I'm just searching incorrectly)?
- replace the generic haskell-tooling matrix with haskell-language-server which seems more appropriate for hls documentation
- move IRC down in the list to incentivize people to join matrix as first option (I'm ok with deleting IRC as @michaelpj suggested, just waiting for some feedback). EDIT got feedback from geekosaur that he's maintaining the irc bridge, so I'd just keep it for now.

### DIFF
--- a/docs/contributing/contributing.md
+++ b/docs/contributing/contributing.md
@@ -4,10 +4,8 @@ The Haskell tooling dream is near, we need your help!
 
 ## How to contact the haskell ide team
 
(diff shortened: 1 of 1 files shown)


michaelpj:
Maybe just add a note that the matrix channel is primary and the IRC channel is bridged?

michaelpj (approved):

Issue #4139, opened 2024-03-12
`Prelude.head: empty list` eval result when `main` function present 


mattapet:
### Your environment

OS: macOS 13.2.1 (22D68)
GHC: 9.8.1, and 9.8.2
LSP: 2.2.0.0, 2.5.0.0, 2.6.0.0, 2.7.0.0, and head
LSP Plugin: `hls-eval-plugin`

### Steps to reproduce

I tried a number of of combinations of GHC/LSP/file setup, but it seems that with GHC 9.8.x any time when there's a `main` function present in the module, the result after evaluation is `Prelude.head: empty list`.

E.g.

```haskell
module Test where

-- >>> 'x'

main :: IO ()
main = putStrLn "Hello World!"
```

### Expected behaviour

```haskell
module Test where

-- >>> 'x'
-- 'x'

main :: IO ()
main = putStrLn "Hello World!"
```

### Actual behaviour

```haskell
module Test where

-- >>> 'x'
-- Prelude.head: empty list

main :: IO ()
main = putStrLn "Hello World!"
```

### Debug information

After a bit of debugging myself, I found the error originating from [this use](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs#L88) of `head`. After a bit of tinkering, it seems that the underlying [hscParsedDecls](https://hackage.haskell.org/package/ghc-9.8.2/docs/src/GHC.Driver.Main.html#hscParsedDecls) function, which `TyThings` output is `[Identifier `$trModule']`, rather than the expected `[Identifier `evalPrint', Identifier `$trModule']`.

As the problem seems to arise from the GHC, I am not quite sure whether there's a bug in GHC, or Eval plugin implementation is working under the wrong assumptions.

I am happy to try to fix this if someone is able to provide some additional direction/advice.

fendor:
Awesome analysis!

Well, the use of `head` seems worthy of doubt at the very least :)

I **can not** reproduce the issue with ghc 9.4.8, so some assumption might be violated nowadays.

Can you write a failing test case? I think it is weird that such a trivial example fails, but the hls-eval test suite is green.

I am afraid, I don't have much more insight than that, we currently don't have any active maintainers for this plugin.

mattapet:
> reproduce the issue with ghc 9.4.8

[...]

PR #4140, opened 2024-03-12
hls-eval-plugin: Replicate #4139

mattapet:


fendor:
Thanks! As a note, our CI jobs are rather flaky and fail unpredictably, so don't be afraid and suspect an error in your code if some random plugin tests fail.

fendor:
That's a good start! Perhaps now we can investigate why the other test cases are fine, and only this scenario is affected?

Since it might not be obvious, you can run your individual test case locally like this:

```sh
> TASTY_PATTERN="T4139" cabal test test:hls-eval-plugin-tests
```

To see debug logs:
```sh
> HLS_TEST_PLUGIN_LOG_STDERR=1 TASTY_PATTERN="T4139" cabal test test:hls-eval-plugin-tests
```

And to log *everything* in HLS (likely too much information):

```sh
> HLS_TEST_LOG_STDERR=1 TASTY_PATTERN="T4139" cabal test test:hls-eval-plugin-tests
```

mattapet:
The debug output was not much of a help. In continuing the search, I tried explicitly filtering out the `main` function from the renamed environment, which made the tests pass again (at least locally).

@fendor do you know if there are any specific properties of the module, or the hsc environment that it might be worth to check for differences?

fendor:
> do you know if there are any specific properties of the module, or the hsc environment that it might be worth to check for differences?

Unfortunately no, I am not familiar with that part of GHC's codebase. Perhaps @wz1000 knows.

mattapet:
Continuing the search, I found the [tcRnDeclsi](https://hackage.haskell.org/package/ghc-9.8.2/docs/src/GHC.Tc.Module.html#tcRnDeclsi) in [hscParsedDecls](https://hackage.haskell.org/package/ghc-9.8.2/docs/src/GHC.Driver.Main.html#hscParsedDecls) to be the point where the `evalPrint` is last seen in the environment. After switching the `ddump-rn-trace` I also noticed the difference in the exports there:

`rnExports: Exports: [main]` vs `rnExports: Exports: [evalPrint]` when we filter out the `main` symbol from the initial env. I'll check if traces further for additional leads. I can share them if that would help

fendor:
Sounds like good detective work! You can share whatever you find!

I will likely only get to it some time in the next couple of days.

mattapet:
Attaching the diffs. They're produced by running `cabal test test:hls-eval-plugin-tests > <<dump-file.txt>>`. I'll also update the PR with version adding the dump flag itself in a bit.

[rn-dump-main.txt](https://github.com/haskell/haskell-language-server/files/14593149/rn-dump-main.txt)
[rn-dump-no-main.txt](https://github.com/haskell/haskell-language-server/files/14593150/rn-dump-no-main.txt)

mattapet:
Scratch the previous comment. It turned out to be a dead-end. After comparing with renamer traces with GHC `9.4.8`, which is the version where the test works, I found the same `Exports`. But I decided to add couple more dump flags and compare the traces across the `9.4.8` and `9.8.2` without changes, and `9.8.2` with `main` filtering.

It appears that for `9.8.2` with the `main` function present, the `evalPrint` function gets removed between `ddump-ds-preopt` and `ddump-ds` steps. Adding traces files here:

[output-dump-9.4.8.txt](https://github.com/haskell/haskell-language-server/files/14593733/output-dump-9.4.8.txt)
[output-dump-9.8.2.txt](https://github.com/haskell/haskell-language-server/files/14593734/output-dump-9.8.2.txt)
[output-dump-no-main-9.8.2.txt](https://github.com/haskell/haskell-language-server/files/14593735/output-dump-no-main-9.8.2.txt)

michaelpj:
Can we merge the test as an expected failure for now? It's a good contribution regardless!

plugins/hls-eval-plugin/test/Main.hs
```diff
@@ -88,6 +88,7 @@ tests =
   , goldenWithEval "Reports an error for an incorrect type with :kind" "T13" "hs"
   , goldenWithEval' "Returns a fully-instantiated type for :type" "T14" "hs" (if ghcVersion >= GHC98 then "ghc98.expected" else "expected") -- See https://gitlab.haskell.org/ghc/ghc/-/issues/24069
   , knownBrokenForGhcVersions [GHC92, GHC94, GHC96, GHC98] "type +v does not work anymore with 9.2" $ goldenWithEval "Returns an uninstantiated type for :type +v, admitting multiple whitespaces around arguments" "T15" "hs"
+  , goldenWithEval "T4139" "T4139" "hs"
```
jhrcek, line 91: Looks good.
Now just mark it as known broken (knownBrokenForGhcVersions [GHC98] $ goldenWithEval ...
and give it more human readable test name, like "Doesn't break in module containing main function".

jhrcek, line 91: Looks good.
Now just mark it as known broken and give it more human readable test name, like:

```suggestion
  , knownBrokenForGhcVersions [GHC98] $ goldenWithEval "Doesn't break in module containing main function" "T4139" "hs"
```

soulomoon:
It might worth add a eval benchmark test to see if this would cause any leak or undesired behaviour.

mattapet:
@soulomoon I added a couple of experiments, one with very simple, single line evaluation, and one a bit more complex. After inspecting the output logs locally, the benchmarks seem to actually exercise the desired behavior. Is this what you had in mind though?

soulomoon:
> @soulomoon I added a couple of experiments, one with very simple, single line evaluation, and one a bit more complex. After inspecting the output logs locally, the benchmarks seem to actually exercise the desired behavior. Is this what you had in mind though?

Thank you for add new benchmarks, it is looking good.
If we are not seeing any more memory usage etc,. It should be good to go.

soulomoon (approved):
LGTM

PR #4144, opened 2024-03-19
Speed up hls-hlint-plugin-tests

fendor:
Migrate `hls-hlint-plugin-tests` to use VirtualFileTree infrastructure for test data.
Avoid `waitForDiagnosticsWithSource` which makes the test wait for 5 seconds, even though it might have received the diagnostics already.

### DIFF
--- a/hls-test-utils/src/Test/Hls.hs
+++ b/hls-test-utils/src/Test/Hls.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE OverloadedLists       #-}
 {-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE OverloadedStrings     #-}
 module Test.Hls
   ( module Test.Tasty.HUnit,
     module Test.Tasty,
@@ -57,6 +58,7 @@ module Test.Hls
     WithPriority(..),
     Recorder,
     Priority(..),
+    captureKickDiagnostics,
     )
 where
 
@@ -124,6 +126,9 @@ import           Test.Tasty.ExpectedFailure
 import           Test.Tasty.Golden
 import           Test.Tasty.HUnit
 import           Test.Tasty.Ingredients.Rerun
+import Language.LSP.Protocol.Lens qualified as L
+import Data.Maybe (mapMaybe)
+import Control.Lens ((^.))
 
 data Log
   = LogIDEMain IDEMain.Log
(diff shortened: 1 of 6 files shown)


fendor:
This doesn't quite work yet, as the test asserts sometimes that *no* diagnostic is emitted. However, on initial load, HLS doesn't publish empty diagnostics, so we don't know when processing finished. We probably need to insert some dummy diagnostics, and clear them.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -208,8 +216,16 @@ rules recorder plugin = do
...
-    files <- getFilesOfInterestUntracked
-    void $ uses GetHlintDiagnostics $ Map.keys files
+    files <- Map.keys <$> getFilesOfInterestUntracked
+    Shake.ShakeExtras{ideTesting = Options.IdeTesting testing, lspEnv} <- Shake.getShakeExtras
+    let signal :: KnownSymbol s => Proxy s -> Action ()
+        signal msg = when testing $ liftIO $ Shake.mRunLspT lspEnv $
+            LSP.sendNotification (LSP.SMethod_CustomMethod msg) $
+            toJSON $ map fromNormalizedFilePath files
+
+    signal (Proxy @"kick/start/hlint")
```
michaelpj, line 226: Can we do this generically? Like, for every rule, send custom method notifications for starting and finishing? It seems like many (most?) tests want this sort of thing, just having a generic signal when rules start or finish would be quite handy.

fendor: I think only diagnostic tests need it, but yeah, we can definitely generalise it.

fendor:
Tests are much quicker now, 15 seconds instead of 2 minutes! Just not sure how to refactor this to make it reusable :sweat_smile:

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
soulomoon: I support it, since we are already tracing the action any way, see `otTracedAction`.

michaelpj: We could add a log handler that intercepts rule start/finish logs and sends the custom messages, like we do for sending `window/logMessage` etc.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -712,6 +717,17 @@ setHlsConfig config = do
   -- requests!
   skipManyTill anyMessage (void configurationRequest)
 
+captureKickDiagnostics :: Session [Diagnostic]
+captureKickDiagnostics = do
+    _ <- skipManyTill anyMessage nonTrivialKickStart2
+    messages <- manyTill anyMessage nonTrivialKickDone2
```
soulomoon, line 723: Would it work if we are using the combination of `nonTrivialKickDone` and `nonTrivialKickDone2` work?

soulomoon: Sorry, one of them is `nonTrivialKickStart`. But never mind, I think it won't work anyway.
To make sure it works we would have to customize a signal for every rule. But It should only turned on if we enable it(Provide a switch as we did for ideTesting🤔), otherwise there would be too many rules being logged.

michaelpj: We don't even need to change the method name, we could put the information about what rule it was in the method parameters.

Or we could just use `$/logTrace`, which we don't use at all at the moment and is maybe quite suited to this?

fendor:
Closing in favour of #4339

PR #4151, opened 2024-03-25
Remove dead code in ghcide and hls-graph for priority

soulomoon:
In the current status of hls-graph, action priority is handled by rts scheduler. 
We do not have a scheduler in hls-graph.


michaelpj (approved):
This seems fine. What is there would not be hard to put back again if we needed it.

soulomoon:
Thank for the review

PR #4153, opened 2024-03-25
Fix hanging after exit is sent #4152

### LINKED ISSUE #4152: Tests that are blocked for 3 second waitting for the server to exit
We have some tests that are blocked for 3 second waitting for the shutdown of the server.
 https://github.com/soulomoon/haskell-language-server/blob/693f2a13fdc9319ed4f85d1ac7e2c45548888a29/hls-test-utils/src/Test/Hls.hs#L644-L645

For example in this test 
https://github.com/soulomoon/haskell-language-server/blob/693f2a13fdc9319ed4f85d1ac7e2c45548888a29/plugins/hls-refactor-plugin/test/Main.hs#L80

we can see the bug:
```
cabal test haskell-language-server:test:hls-refactor-plugin-tests --test-options="-p /refactor/"  
```

My guess is that

~~The SMethod_Shutdown is sent to the server at the end of runSessionWithHandles
It seems for this case the shutdown handler SMethod_Shutdown is not set up even after runSessionWithHandles is done.
I added a sleep 0.1 to the test to wait for the setup, then the server is closed as normal.~~

soulomoon:
Trying to fix #4152

Reactor thread offered a db connect to hieDB, if it exist, the db connection exit with it.
Operation might need the db connection should only run if the reactor thread is still running. 

Otherwise a race condition might happened: 
Shutdown handler would stop the reactor thread, cancel the shake seesion.
If an operation is still trying to reach the db, in this case, the config update callback would try to call `setSomethingModified` which need db connections. It hangs HLS forever.

This PR ships the `config update callback` to the reactor thread to fix the deadlock.
Alternative would be to make the shutdown handler take away the ideaVar.


soulomoon:
We need to send the update of config to reactor thread to avoid the race condition.

jhrcek:
I don't understand how the fix works (will leave review to people who understand it better) but great that you managed to resolve it.
Do you have any estimate how much time this saves across one test suite run?

soulomoon:
> Do you have any estimate how much time this saves across one test suite run?

I have not measured, But I don't think there is much. Instead of the initial motivation to speed up the testcase. The PR is rather a bug fix that is to solve the potential dead lock situation from happenning.

michaelpj:
Hmm, this is tricky. I'm not sure this is quite the right fix. Can we instead make it so that we throw in this case instead of hanging?

soulomoon:
Alternative we could make the shutdown handler take away the ideaVar to prevent the race condition. This would not change the thread the configupdate callback is running on @michaelpj

michaelpj:
I think that's a better approach. But I think maybe we can work at a lower level of granularity? Basically, I think the pattern we want is something like this:
- During shutdown, various resources or capabilities may become unavailable
- If we try to access those resources or capabilities then we should fail instead of getting stuck

In this case I think you've narrowed it down to the `hiedb` connection, so I think what we want is something like:
- During shutdown, the database connection will become unavailable
- If we try to access the database connection after this, we should fail

What you're suggesting with the `ideVar` is one possible way of doing this: we remove the handle that our code uses to find the resources. I see two problems with it:
1. I'm sure in many other places we use `readMVar`. Now _these_ will hang if they get triggered during shutdown. If they would all throw then I would be okay with it.
2. This prevents things from working that _could_ work. Maybe that doesn't matter but I could imagine it causing other issues (e.g. what if we had code to flush the values of persistent rules at shutdown and now it can't access the rule system?)

So I think it would be best to try and localise the guard more. Ideally we would just make the thing that currently hangs fail instead. I don't know exactly where that is, but maybe you do?

soulomoon:
> * During shutdown, the database connection will become unavailable
> * If we try to access the database connection after this, we should fail.

Yes, I agree about this.

In summary,  we actually have two problem to solve here.
* First problem, Access to database connection should fail if it is closed instead of hang. (This is root from `HieDB` and can be fixed there? unfortunatly, I am not familiar with it, maybe @wz1000 or @jhrcek do)
* Second problem, The config update callback should not be invoke after the shutdown handler is called.

> * I'm sure in many other places we use readMVar. Now these will hang if they get triggered during shutdown. If they would all throw then I would be okay with it.
> * This prevents things from working that could work. Maybe that doesn't matter but I could imagine it causing other issues (e.g. what if we had code to flush the values of persistent rules at shutdown and now it can't access the rule system?)

I have update this branch with the gurad fix.
I think this can resolve your concerns:
We are now having only one user for `ideStateVar` after we put the `ideaState` value in it. It is the config update callback. 
* This is used to guard against the config update callback before the initilization
After the fix, we would another a second user: the shutdown handler. So we have a second function for the `ideStateVar`
* Guard against the config update callback after the shutdown.

So I believe `ideStateVar` is an ideal guard to solve the second problem,

michaelpj:
> Access to database connection should fail if it is closed instead of hang.

Did you manage to figure out exactly which function is hanging?

> The config update callback should not be invoke after the shutdown handler is called(It is not only trying to access the db connection, also it would try to restart the shake session).

Again, I'm not sure this is necessarily a problem. What happens if we restart the shake session after it is closed? It should fail! So I think it's fine if we allow "higher-level" things to run, so long as the "lower-level" things that can't work fail appropriately.

> So there would be a second functionality for the ideStateVar.

... and now this is an invariant we have to remember to use whenever we work with `ideStateVar`! In that way it makes our lives more complicated, I feel.

soulomoon:
> Again, I'm not sure this is necessarily a problem. What happens if we restart the shake session after it is closed? It should fail! So I think it's fine if we allow "higher-level" things to run, so long as the "lower-level" things that can't work fail appropriately.

🤔 If we only make the "lower-level" to fail instead of hanging,
the "higher-level" -- the main LSP thread handling the callback would still fail.
We would be left with something error like this:

```
Test suite hls-refactor-plugin-tests: RUNNING...
refactor
  initialize response capabilities
       code action:     FAIL
      Exception: onConfigChange: db already shutdown
```

soulomoon:
I am creating a minimal example to see if I can reproduce the hanging.

michaelpj:
> We would be left with something error like this:

Sure, but then we can work from there, e.g. we can ask: should we be catching exceptions during shutdown, or in the config handler, or whatever. At least the problem is not masked!

>  if the thread running runWithDb is done, the writing to hieChan would hang(Not exactly the Access to database connection for correction).

Huh, that is weird. AFAIK writing to a `TChan` shouldn't hang, ever. I'll take a look at your reproduction code...

I also note that there's some stuff in there about `retryIfSqlitBusy` that looks like it might retry for up to 10 seconds, which might be a problem?

soulomoon:
> AFAIK writing to a TChan shouldn't hang, ever. I'll take a look at your reproduction code...

You are right writing the TChan won't hang if the `TChan`  is already computed.

We have a very interesting situation:
The problematic part is `~que <- unsafeInterleaveIO $ takeMVar queVar`, it delay the computation of `que`.
`x <- atomically $ readTQueue que` hangs, when it tries to actually compute the `que` and write to it.

michaelpj:
> when it tries to actually compute the que and write to it.

I also am not sure how that can happen! Surely creating the queue and then writing to it should be fine?

soulomoon:
Notice the computation of the que like this. `takeMVar queVar`. 
`queVar` might never be fill with the `que` value

michaelpj:
Yeah, that's super weird and I don't know why we do it like that. If it works the other way that looks reasonable to me, what do you think @wz1000 ?

In general it seems like anywhere we have a `takeMVar` we have a potential hang.

The other question of course is why the hanging thread doesn't get killed. I think it's probably because we process notifications synchronously, so if we get stuck processing a notification we will never see the `exit` notification that comes in. Tricky!

I think we could also make this simpler on the `lsp` side. The spec says we're allowed to ignore notifications after `shutdown` and error requests, so it could keep track of whether we have been shutdown, and if so block all further requests and notifications. That would stop us getting into this situation.

soulomoon:
@michaelpj 
I see we talked about it in the biweekly meeting, is there something new you guys have come up with?

michaelpj:
That was just about the generic shutdown stuff. I would like to understand why there is a hang here. But it seems like at least one thing we should try is your commit that gets rid of the weirdness. Does that pass the tests? And does it solve the problem?

(I still think a comment from @wz1000 would be helpful here.)

soulomoon:
Yes, it does pass the test and solve the problem.

Simply put, it hangs because we are trying to `takeMvar` of the `queVar`(It should contain the `write chan` of the db), but `queVar` is empty(Reactor thread is shutdown before it put the `write chan` to `queVar`). 
What worse, this `takeMvar` happens in `update callback`(Not in its callsite),  because we use `unsafeInterleaveIO` to delay the `takeMvar` IO operation. That is why it is so hassling.

So two things have been done here so far:
1. `update callback` won't be running after the shutdown.
2. Removal of `unsafeInterleaveIO`. It wait until reactor thread(The one create the db connection) put the write chan of db to `queVar` in the initialization.

michaelpj:
What happens if we just do the second bit? I would still prefer to avoid complicating the config update callback if we can

soulomoon:
> What happens if we just do the second bit? I would still prefer to avoid complicating the config update callback if we can

If only the second bit is done. There is still some hangs occasionally, It is some where inside the `shakeRestart`(In the proccess of update config), I still could not configure out where it hangs.

IMO, do a `shakeRestart` after the `shutdownHandler` is NB. We should avoid it all togather, that is why we need the first bit.

soulomoon:
Great, it seems to be working with lsp updated with correct shutdown behaviour.

soulomoon:
```
--> {
    "jsonrpc": "2.0",
    "method": "initialized",
    "params": {}
}
--> {
    "id": 0,
    "jsonrpc": "2.0",
    "method": "shutdown"
}
2024-04-08T14:20:03.274166Z | Info | LSP: received shutdown
2024-04-08T14:20:03.274196Z | Debug | Received shutdown message
--> {
    "id": 1,
    "jsonrpc": "2.0",
    "result": [
        {
            "cabalFormattingProvider": "cabal-gild",
            "checkParents": "CheckOnSave",
            "checkProject": true,
            "formattingProvider": "ormolu",
            "maxCompletions": 40,
            "plugin": {}
        }
    ]
}
<-- {
    "id": 0,
    "jsonrpc": "2.0",
    "result": null
}
--> {
    "jsonrpc": "2.0",
    "method": "exit"
}
2024-04-08T14:20:03.274515Z | Debug | LSP: set new config: {
  "cabalFormattingProvider": "cabal-gild",
  "checkParents": "CheckOnSave",
  "checkProject": true,
  "formattingProvider": "ormolu",
  "maxCompletions": 40,
  "plugin": {  }
}
2024-04-08T14:20:03.274588Z | Debug | Configuration changed: Config {checkParents = CheckOnSave, checkProject = True, formattingProvider = "ormolu", cabalFormattingProvider = "cabal-gild", maxCompletions = 40, plugins = fromList []}
2024-04-08T14:20:03.276135Z | Info | Reactor thread stopped
Server does not exit in 3s, canceling the async task...
Finishing canceling (took 0.41ss)

```

It seems we are still trying to "set new config" after the shutdown.

soulomoon:
So this is what happens:
1 server sendRequest `SMethod_WorkspaceConfiguration` to client (callback is not take effect)
2 client send shutdown to server
3 response to `SMethod_WorkspaceConfiguration` is recieved, update callback begins here.(We hangs for some reason)

michaelpj:
Right, I wondered about doing that (ignoring responses after shutdown). The spec doesn't say that we should, so I guess it's optional. I guess the reason it seems risky to me is that in principle a server might be waiting for a response to happen. That might be foolish, but it could happen. Suppose e.g.:
1. The server tries to start a progress session, and requests a progress token, blocking some thread until it receives it
2. The client sends shutdown
3. The client sends the progress token back (but the server ignores it)
4. The server hangs until the blocking in 1 is resolved some other way (e.g. a timeout)

Now, this is maybe bad behaviour from the server implementor, but I thought it seemed reasonable to still handle responses.

What I could do is have us not call the config update callback after shutdown :thinking: 

But yeah, it would still be great if we could work out why `shakeRestart` is hanging. And I think your change to get rid of the `unsafePerformIO` is good anyway. Maybe put that in a separate PR and we can try and get that bit done?

soulomoon:
I am also considering adding a current thread ID to the hls log.
It should help us with a more effective debugging eperience on such race issues.

soulomoon:
Something like this
```
ThreadId 19 | 2024-04-11T03:17:33.383209Z | Info | Started LSP server in 0.01s
ThreadId 19 | 2024-04-11T03:17:33.452743Z | Debug | ghc --print-libdir
ThreadId 19 | 2024-04-11T03:17:33.453779Z | Debug | Setting initial dynflags...
ThreadId 19 | 2024-04-11T03:17:33.453819Z | Debug | shouldRunSubset: True
ThreadId 29 | 2024-04-11T03:17:33.453879Z | Debug | Initializing exports map from hiedb
ThreadId 29 | 2024-04-11T03:17:33.454034Z | Debug | Done initializing exports map from hiedb. Size: 0
```

soulomoon:
Fixed at lsp https://github.com/haskell/lsp/pull/567 and upgraded lsp in #4166

soulomoon:
I found another hang https://github.com/soulomoon/haskell-language-server/blob/eaa4a94b7f62533132c437a460971cdf9f3af8bd/ghcide/session-loader/Development/IDE/Session.hs#L627

Issue #2979, opened 2022-06-21
Remove the ghcide executable


michaelpj:
The HLS executable can already be built without all the plugins, there are flags for that. So it's unclear what purpose the `ghcide` executable serves any more. Removing it would let us simplify some code, and perhaps move some of the glue code out of `ghcide` (is that desirable?).

Anyone have any strong reasons to keep it?

pepeiborra:
Previous discussion: 

https://github.com/haskell/haskell-language-server/issues/813
https://github.com/haskell/ghcide/pull/939

fendor:
I did take a look, and it is currently non-trivial because the ghcide-testsuite depends on the ghcide executable and instructs it to emit certain LSP messages for the tests.

To remove the ghcide executable, we need to make the testsuite depend on HLS which requires a bit more untangling of the hls-plugin-api/ghcide/HLS dependency graph.

soulomoon:
Do you guys think this might be a good approach? we start by adding a core plugin, then move most of the ghcide functionality to the core plugin bit by bit.
If you guys can agree with me on this approach, I'll add a core plugin and move a bit of ghcide functionality and test over.

fendor:
To be perfectly clear, ghcide would not expose a plugin any more, consequentially there would be barely any tests in ghcide? I am not sure that's enough, I think we should first focus on the dependency graph and look at what dependency edges we wish to sever. I created this:

![hls-deps drawio(1)](https://github.com/haskell/haskell-language-server/assets/8463814/ea6ecfb5-a8e8-4c41-adda-96092e3ea1ae)

I hope there are no mistakes, but it can help identifying why it is non-trivial to refactor :upside_down_face:

wz1000:
I find the `ghcide` executable very useful for quickly testing changes without waiting for all of HLS and its dependencies to build, and without manually disabling all the plugin flags.

michaelpj:
> I am not sure that's enough, I think we should first focus on the dependency graph and look at what dependency edges we wish to sever

In the long run I'm not sure even that is enough. If we were starting again today, I doubt we would have a separate package like `ghcide` at all. I imagine we might have:

- `hls-core`: the rule system, the handlers and rules that deal with file system changes, position mapping etc.
- `hls-haskell-plugin`: a "normal" plugin that provides handlers for Haskell stuff, the `Typecheck` rules etc.

[...]

PR #4156, opened 2024-03-29
POC: push plugin enablement into plugin components

michaelpj:
This implements a version of what I describe in [this comment] on the `fourmolu` plugin as a POC. The goal here is that:
- CPP moves to plugin components, rather than the integration point
- Plugin components always export a plugin descriptor, it just might provide nothing and say it is disabled
   - This can be helpful for our users, to tell them that things are disabled
- Plugin components always build, even if the flag is turned off
   - This is useful since cabal actually always tries to solve for all the buildable components, pulling in unwanted dependencies. This way that will happen, but it will be fine. (An alternative would be to just mark them as not buildable in this situation)
- Plugin _tests_ are not buildable if the flag is off.
    - We could try and make them always buildable, and just have no tests, but that seems worse?

I haven't tried to make "being disabled" a first-class status or anything, a disabled plugin is just a plugin with no handlers and a description that calls that out. Maybe that's a mistake? I'm not sure.

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -1370,19 +1370,18 @@ flag fourmolu
   default:     True
   manual:      True
 
-common fourmolu
-  if flag(fourmolu)
-    build-depends: haskell-language-server:hls-fourmolu-plugin
-    cpp-options: -Dhls_fourmolu
-
 library hls-fourmolu-plugin
   import:           defaults, pedantic, warnings
   exposed-modules:  Ide.Plugin.Fourmolu
   hs-source-dirs:   plugins/hls-fourmolu-plugin/src
+
+  if flag(fourmolu)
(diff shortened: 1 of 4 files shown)


maralorn:
As a downstream consumer I like this a lot. It will make things easier for us.

In regard to the code I just wonder how good this scales. Does this mean that every file in every plugin contains a big CPP statement. I wonder if that is annoying to maintain?

michaelpj:
I pushed an alternative version that instead has multiple different source directories.

michaelpj:
WDYT @fendor ? It turns one module into four, which is not that nice...

soulomoon:
> An alternative would be to just mark them as not buildable in this situation
Will it improve the build time if most of the plugins are disabled.
It would be helpful to create the minimal HLS described in https://github.com/haskell/haskell-language-server/issues/2979 ?

PR #4159, opened 2024-04-05
Fix references to old CPP names in tests, update tests

jhrcek:
Picks up where https://github.com/haskell/haskell-language-server/pull/4129 left off + includes test fixes.
One of the fixed tests seems to be flaky so might require some stabilization..

### DIFF
--- a/test/functional/Progress.hs
+++ b/test/functional/Progress.hs
@@ -37,7 +37,7 @@ tests =
 
               (codeLensResponse, activeProgressTokens) <- expectProgressMessagesTill
                 (responseForId SMethod_TextDocumentCodeLens lspId)
-                ["Setting up testdata (for T1.hs)", "Processing", "Indexing"]
+                ["Setting up testdata (for T1.hs)", "Processing"]
                 []
 
               -- this is a test so exceptions result in fails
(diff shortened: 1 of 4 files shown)


jhrcek:
Converting to draft as the fixed eval plugin test is quite unstable and slow. Will check it out in detail tomorrow.

test/functional/Progress.hs
```diff
@@ -37,7 +37,7 @@ tests =
 
               (codeLensResponse, activeProgressTokens) <- expectProgressMessagesTill
                 (responseForId SMethod_TextDocumentCodeLens lspId)
-                ["Setting up testdata (for T1.hs)", "Processing", "Indexing"]
+                ["Setting up testdata (for T1.hs)", "Processing"]
```
jhrcek, line 40: Removing this Indexing because it was making the test flaky (sometimes appeared _after_ the code lens reponse).
As suggested by @fendor the only Indexing message that happens is related to files not related to this test, so it also semantically doesn't belong here.

fendor (approved):

fendor (approved):
LGTM, thanks for pushing it over the finish line!

PR #4160, opened 2024-04-06
Mark plugins as not buildable if the flag is disabled

### LINKED ISSUE #4100: Merged plugins pull in disabled deps when building with `Setup.hs`
I'm taking a look at packaging `haskell-language-server-2.7.0.0` for Gentoo Linux. Currently, I have `2.6.0.0` and most plugins installed to my system, although I do not have the `hls-fourmolu-plugin` or `fourmolu` packages installed. Just experimenting with the `Setup.hs` interface, I notice that the `fourmolu` dependency is being pulled in even when the `fourmolu` flag is disabled:

```bash
cd haskell-language-server-2.7.0.0/
runhaskell Setup.hs configure --flag=-fourmolu
```

As you can see, it is still looking for `fourmolu`:

```
Configuring haskell-language-server-2.7.0.0...
Setup.hs: Encountered missing or private dependencies:
fourmolu >=0.14 && <0.16,
ghcide ==2.7.0.0,
hiedb >=0.6.0.0 && <0.7,
hls-graph ==2.7.0.0,
hls-plugin-api ==2.7.0.0,
lsp >=2.4 && <2.5
```

The changes made in https://github.com/haskell/haskell-language-server/pull/3976 seem to be selectively choosing which sub-libraries to build using `build-depends`:

https://github.com/haskell/haskell-language-server/blob/41de40ea637231be3ceb1355513f6a99c0c6026f/haskell-language-server.cabal#L1298-L1301

It seems to me that `Setup.hs` does not seem to have the functionality to handle this, and instead tries to build _all_ sub-libraries mentioned in the `.cabal` file, regardless of what flags are toggled.

This experiment was done using `ghc-9.2.8` and `Cabal-3.6.3.0`.

michaelpj:
This ensures that cabal does not consider them at all, and won't try to solve for their dependencies. So if we turn off the fourmolu plugin, cabal really won't consider fourmolu at all.

This gets us some of the benefits of #4156 with much less work.

Fixes #4100.

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -118,7 +118,7 @@ flag isolateCabalfmtTests
 
 library hls-cabal-fmt-plugin
   import:           defaults, pedantic, warnings
-  if !flag (cabalfmt)
+  if !flag(cabalfmt)
     buildable: False
   exposed-modules:  Ide.Plugin.CabalFmt
   hs-source-dirs:   plugins/hls-cabal-fmt-plugin/src
@@ -136,7 +136,7 @@ library hls-cabal-fmt-plugin
 
 test-suite hls-cabal-fmt-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
-  if !flag (cabalfmt)
+  if !flag(cabalfmt)
     buildable: False
   type:             exitcode-stdio-1.0
   hs-source-dirs:   plugins/hls-cabal-fmt-plugin/test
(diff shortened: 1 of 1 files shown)


haskell-language-server.cabal
```diff
@@ -158,7 +162,7 @@ flag cabalgild
   manual:      True
 
 common cabalgild
-  if flag(cabalgild)
+  if !flag(cabalgild)
```
jhrcek, line 165: This looks undesirable?

jhrcek: Also nitpick: I see 4x `flag(` and 56 x `flag (` (with space after flag.
I'd unify those for easier searchability..

jhrcek (approved):
Looks good, just..

haskell-language-server.cabal
michaelpj: OOps thanks, I fixed that locally but forgot to push!

PR #4163, opened 2024-04-06
Remove EKG

### LINKED ISSUE #3993: Remove EKG from ghcide
It's not clear anyone uses it and it's a pain to update. 

Please chime in if you do use it or care!

michaelpj:
It's always been a pain and it's not really used.

Fixes #3993


fendor (approved):
LGTM. I am a bit sad, because I think these monitors are interesting in itself, but I cant say I have used them.

michaelpj:
Yes, I think it's a good idea, but the interface is still there if we want to resurrect it. And the EKG library has been nothing but a pain.

PR #4167, opened 2024-04-09
get rid of the `unsafeInterleaveIO` at start up

soulomoon:
It contribute to a hang in #4152
we discover it when doing #4153


soulomoon:
Does it look good @michaelpj

michaelpj (approved):
This seems reasonable to me. The existing comment doesn't make much sense to me. If it passes the tests I think it should be fine!

Issue #4168, opened 2024-04-09
Formatters block on session setup


michaelpj:
At the moment, most of our formatter plugins require the GHC session in order to acquire the `DynFlags` for a module. This is even true for the CLI modes - most of the plugins still get the flags so that they can pass them to the CLI executable.

This means that formatting is generally going to block until the session is set up. This isn't great: in particular, it's not uncommon to have some kind of format-on-save set up, and that means you can't save a file until HLS is finished setting up! (I think there's also a client problem there: I think it should be the [client's responsibility](https://github.com/microsoft/language-server-protocol/issues/1916) to time out formatting in that case)

So it would be nice if this didn't happen. The best thing I can think of doing is just to let the formatters do whatever they do if they don't get the flags specified. That presumably means guessing them from cabal files etc, but crucially in a way that doesn't block on HLS. It would also make them more likely to line up with how a CLI formatter would operate. Or at least if we're in the CLI mode I think we shouldn't take the flags from HLS, since the whole point is to match what the CLI formatter does.

Thoughts from formatting plugin maintainers? @georgefst @peterbecich ?

michaelpj:
See https://github.com/haskell/haskell-language-server/pull/4170

michaelpj:
We could also have an option, of course, although generally better to just do the right thing.

PR #4170, opened 2024-04-09
ormolu-plugin: don’t request GhcSession for useCLI

Profpatsch:
Probably not the best solution.

When ormolu is requested from the command line, it should do its own detection of cabal features. If it has to set up the GhcSession just to get some flags, it blocks saving a file (`formatOnSave = true`), which is a horrible user experience in cases where the session takes a minute or more to set up.

Instead, now the plugin only depends on ormolu and the file itself.

Fixes https://github.com/haskell/haskell-language-server/issues/4168


michaelpj:
cc @georgefst 

This seems plausible to me. We would need to change the `fourmolu` one to match as they are generally kept in sync.

Issue #4173, opened 2024-04-13
Extracts ghcide-tests and migrate it to use hls-test-utils


soulomoon:
I think we should pick up https://github.com/haskell/haskell-language-server/pull/3315,
This can be served as part of https://github.com/haskell/haskell-language-server/issues/3736,
Extracts ghcide-tests, and migrate the test to use hls-test-utils. This can be served as a pre-step to move the all the functionalities of ghcide to the stand alone plugin https://github.com/haskell/haskell-language-server/issues/2979

Some benifits:
1. Removal of code duplication.
2. Test speed up, we can refactor the test that is doing unconditional wait along the way. 
3. Test correctness, we can refactor some tests that are not designed properly.

When I have done a proof of concept of https://github.com/haskell/haskell-language-server/issues/2979,
I migrate some of the tests to use hls-test-utils, I have some positive result, some tests can be run 2 to 3 times faster.

Jobs that need to be done

- [x] merging ghcide-test-utils and hls-test-utils
  - https://github.com/haskell/haskell-language-server/pull/4197
  - #4274
  - https://github.com/haskell/haskell-language-server/pull/4272
- [x]  move ghcide-tests to haskell-language-server.cabal and make it depend on hls-test-utils #4176
- [x]  fix up each test to use hls-test-utils. 
  - [x]     AsyncTests #4199
  - [x]     BootTests #4227
  - [x]     ClientSettingsTests #4258 
  - [x]     CodeLensTests https://github.com/haskell/haskell-language-server/pull/4231
  - [x]     CompletionTests https://github.com/haskell/haskell-language-server/pull/4192
  - [x]     CPPTests https://github.com/haskell/haskell-language-server/pull/4231
  - [x]     CradleTests https://github.com/haskell/haskell-language-server/pull/4231
  - [x]     DependentFileTest #4195
  - [x]     DiagnosticTests https://github.com/haskell/haskell-language-server/pull/4207
  - [x]     ExceptionTests #4231
  - [x]     FindDefinitionAndHoverTests #4202
  - [x]     FuzzySearch
  - [x]     GarbageCollectionTests https://github.com/haskell/haskell-language-server/pull/4231
  - HaddockTests
  - HieDbRetry
  - [x]     HighlightTests  #4202
  - [x]     IfaceTests https://github.com/haskell/haskell-language-server/pull/4201
  - [x]     InitializeResponseTests
  - LogType
  - [x]     NonLspCommandLine
[...]

PR #4176, opened 2024-04-14
move ghcide-tests to haskell-language-server.cabal and make it depend on hls-test-utils

soulomoon:
A part of the job that break the import cycle migrating the ghcide-tests to use hls-test-util https://github.com/haskell/haskell-language-server/issues/4173.


michaelpj:
Fine - do you want to do this by itself or also tweak a few tests?

soulomoon:
I am going migrate the `InitializeResponseTests.hs` along with it.

soulomoon:
Done migrating the InitializeResponseTests.hs

michaelpj (approved):

PR #4179, opened 2024-04-16
Allow users to specify whether to use `cabal`'s multi-repl feature

### LINKED ISSUE #4178: Users should be able to specify `multi-repl` support and opt-out of it, if unsupported.
We added support for loading multiple component loading in https://github.com/haskell/haskell-language-server/pull/3462 and since then, we had a couple of bug reports and other issues, such as https://github.com/haskell/cabal/issues/9860.

We improve the situation by adding a `hie-bios` configuration and expand this configurability to HLS itself: It should be possible to use the cabal `multi-repl` feature by toggling an option in the lsp client.

To achieve that, introduce a new LSP config:

**Description:** Specify the project loading strategy of the Language Server. It allows the build tool (such as `cabal` or `stack`) to [load multiple components at once](https://github.com/haskell/cabal/pull/8726). While this can lead to an improved IDE experience, it is an experimental feature. Only change to \"Single Component\" when you encounter issues with `cabal`.

**Configs:**
* **Single Component**: Always load only a single component at a time. This is the most reliable option if you encountered any issues with the other options.
* **Multiple Components**: Allow the build tool to setup a multiple component session, if the build tool supports it. At the moment, only `cabal` supports loading with multiple components. If the `cabal` version does not support load multiple components at once, HLS falls gracefully back to \"Single Component\" mode
* **Auto**: Let HLS decide how to load your project. For build tools that support it, HLS will use \"Multiple Components\" and otherwise \"Single Component\". Starting from the `cabal` version `3.12`, `cabal` supports loading of multiple components at once.

fendor:
We add an option to `Config` that allows clients to specify how HLS should load components.

The three options are:

* SessionLoadSingleComponent: Always load only a single component when a new component is discovered.
* SessionLoadMultipleComponents: Always allow the cradle to load multiple components at once. This might not be always possible, e.g., if the tool doesn't support multiple components loading. The cradle can decide how to handle these situations.
* SessionAuto: Leave the decision to the individual cradle type. Some cradles have good support for loading multiple components, while others do not. So use multiple components loading for tools, such as cabal, and force single component loading for tools that do not support multiple components, such as stack.

This allows users to opt-out of `cabal` `multi-repl` whenever they want and vice versa.

Closes #4178 

Depends on https://github.com/haskell/hie-bios/pull/433

Sibling PR for vscode-haskell: https://github.com/haskell/vscode-haskell/pull/1077

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -25,7 +25,7 @@ import           Control.Concurrent.Async
 import           Control.Concurrent.Strict
 import           Control.Exception.Safe               as Safe
 import           Control.Monad
-import           Control.Monad.Extra
+import           Control.Monad.Extra                  as Extra
 import           Control.Monad.IO.Class
 import qualified Crypto.Hash.SHA1                     as H
 import           Data.Aeson                           hiding (Error)
@@ -52,13 +52,13 @@ import           Development.IDE.Core.RuleTypes
 import           Development.IDE.Core.Shake           hiding (Log, Priority,
                                                        knownTargets, withHieDb)
 import qualified Development.IDE.GHC.Compat           as Compat
-import qualified Development.IDE.GHC.Compat.Util      as Compat
 import           Development.IDE.GHC.Compat.Core      hiding (Target,
                                                        TargetFile, TargetModule,
                                                        Var, Warning, getOptions)
 import qualified Development.IDE.GHC.Compat.Core      as GHC
 import           Development.IDE.GHC.Compat.Env       hiding (Logger)
 import           Development.IDE.GHC.Compat.Units     (UnitId)
+import qualified Development.IDE.GHC.Compat.Util      as Compat
 import           Development.IDE.GHC.Util
 import           Development.IDE.Graph                (Action)
 import           Development.IDE.Session.VersionCheck
@@ -70,6 +70,7 @@ import           Development.IDE.Types.Location
 import           Development.IDE.Types.Options
 import           GHC.Check
 import qualified HIE.Bios                             as HieBios
+import qualified HIE.Bios.Cradle                      as HieBios
 import           HIE.Bios.Environment                 hiding (getCacheDir)
 import           HIE.Bios.Types                       hiding (Log)
 import qualified HIE.Bios.Types                       as HieBios
@@ -80,6 +81,8 @@ import           Ide.Logger                           (Pretty (pretty),
                                                        nest,
                                                        toCologActionWithPrio,
                                                        vcat, viaShow, (<+>))
+import           Ide.Types                            (SessionLoadingPreferenceConfig (..),
+                                                       sessionLoading)
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Server
 import           System.Directory
@@ -123,7 +126,8 @@ import           GHC.Data.Bag
 import           GHC.Driver.Env                       (hsc_all_home_unit_ids)
 import           GHC.Driver.Errors.Types
 import           GHC.Driver.Make                      (checkHomeUnitsClosed)
-import           GHC.Types.Error                      (errMsgDiagnostic, singleMessage)
+import           GHC.Types.Error                      (errMsgDiagnostic,
+                                                       singleMessage)
 import           GHC.Unit.State
 #endif
 
@@ -149,6 +153,7 @@ data Log
   | LogNoneCradleFound FilePath
   | LogNewComponentCache !(([FileDiagnostic], Maybe HscEnvEq), DependencyInfo)
   | LogHieBios HieBios.Log
+  | LogSessionLoadingChanged
 deriving instance Show Log
 
 instance Pretty Log where
@@ -219,6 +224,8 @@ instance Pretty Log where
     LogNewComponentCache componentCache ->
       "New component cache HscEnvEq:" <+> viaShow componentCache
     LogHieBios msg -> pretty msg
+    LogSessionLoadingChanged ->
+      "Session Loading config changed, reloading the full session."
 
 -- | Bump this version number when making changes to the format of the data stored in hiedb
 hiedbDataVersion :: String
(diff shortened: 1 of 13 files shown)


fendor:
The only thing is kind of bad is that changing the load style requires a restart of the language server. We could track when the config changes and unload all components instead.

michaelpj:
> The only thing is kind of bad is that changing the load style requires a restart of the language server. We could track when the config changes and unload all components instead.

Yeah, we already do something like this when e.g. we have a change to a cabal file. Can we do the same here?

fendor:
> Can we do the same here?

Yeah, I believe we can.

wz1000:
I'm skeptical about the utility of the `auto` configuration option, but other than that it seems like it is OK to move ahead with this. We should produce an hie-bios release before merging though.

fendor:
> I'm skeptical about the utility of the auto configuration option, but other than that it seems like it is OK to move ahead with this. We should produce an hie-bios release before merging though.

 Yes, after some more discussion, I agree that the `auto` feature is overkill for now. I propose the following changes to the PR:
 
 * Single Component: Default option
 * Multiple Components: Allow the usage of `mutli-repl: True` if the cabal cradle allows it.

michaelpj:
Then isn't this really just a flag saying "allow use of cabal multi-repl: yes/no"? I continue to be a little suspicious of pretending that it's more general than it is.

fendor:
Yes it is, and the docs can make this clear, but there is nothing that is specific to cabal cradles right now. `bios` cradles could implement the same mechanism, we just didn't do it yet in `hie-bios`.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -206,6 +210,38 @@ data CheckParents
...
+data SessionLoadingConfig
+    = SessionLoadSingleComponent
+    -- ^ Always load only a single component when a new component
+    -- is discovered.
+    | SessionLoadMultipleComponents
+    -- ^ Always allow the cradle to load multiple components
+    -- at once. This might not be always possible, if the tool doesn't
+    -- support multiple components loading.
+    --
+    -- The cradle can decide how to handle these situations.
```
michaelpj, line 223: I think maybe we should just describe these options as _preferences_. i.e. "prefer loading a single component" or whatever, since the implication is that the cradle may do something else if it can't honour it. You could imagine a cradle type that can't honour single component loading (e.g. arguably a direct cradle that loads all components is like that).

test/testdata/schema/ghc98/default-config.golden.json
```diff
@@ -151,5 +151,6 @@
         "stan": {
             "globalOn": false
         }
-    }
+    },
+    "sessionLoading": "Single Component"
```
michaelpj, line 155: This looks different from most of our other option values:
- None have spaces
- Most start lower-case
- The uppercase ones are clearly aeson-derived so look more like `SingleComponent`

It's all a bit inconsistent but I think this makes it _more_ inconsistent. So I'd go with `SingleComponent` or even better `singleComponent`.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -194,6 +197,7 @@ instance Default Config where
     -- , cabalFormattingProvider     = "cabal-fmt"
     -- this string value needs to kept in sync with the value provided in HlsPlugins
     , maxCompletions              = 40
+    , sessionLoading              = SessionLoadSingleComponent
```
michaelpj, line 200: I do think we should make the _default_ multi-component. We want people to benefit from this without knowing they have to turn it on! The main thing is that they can turn it _off_ if they need to.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -217,6 +222,8 @@ instance Pretty Log where
     LogNewComponentCache componentCache ->
       "New component cache HscEnvEq:" <+> viaShow componentCache
     LogHieBios msg -> pretty msg
+    LogSessionLoadingChanged ->
+      "Session Loading Config change, reload the full session."
```
michaelpj, line 226:
```suggestion
      "Session Loading config change, reloading the full session."
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -461,6 +469,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
   runningCradle <- newVar dummyAs :: IO (Var (Async (IdeResult HscEnvEq,[FilePath])))
 
   return $ do
+    clientConfig <- getClientConfigAction
```
michaelpj, line 472: I'm not sure this actually works? This will get the client config at _this_ point, but we consult it inside `consultCradle` which is an IO function that maybe runs later? So might we just always see a stale version of the config? I think we might need to get the config in IO inside `consultCradle`.

fendor: It does work, in combination with invalidating the `GhcSessionIO` rule.

hls-plugin-api/src/Ide/Types.hs
fendor: People have raised the concern that this is basically an untested cabal feature that may contain many hard to understand bugs, so we should disable it by default, and then enable it by default in the next release.

test/testdata/schema/ghc98/default-config.golden.json
fendor: Ok, will change as suggested!

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -679,11 +688,38 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
                return (res, maybe [] pure hieYaml ++ concatMap cradleErrorDependencies err)
 
+    let
+        -- | We allow users to specify a loading strategy.
+        -- Check whether this config was changed since the last time we have loaded
+        -- a session.
+        --
+        -- If the loading configuration changed, we likely should restart the session
+        -- in its entirety.
+        didSessionLoadingConfigChange :: IO Bool
```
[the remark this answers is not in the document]
fendor: It is, but I don't have the time to fix it all, if we want to do a release for GHC 9.6.5 this week.

ghcide/session-loader/Development/IDE/Session.hs
fendor: Thanks!

hls-plugin-api/src/Ide/Types.hs
fendor: Updated the comments, will also update the description in vscode-haskell once this PR is merged.

wz1000 (approved):

ghcide/session-loader/Development/IDE/Session.hs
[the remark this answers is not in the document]
michaelpj: yeah, I wasn't suggesting that you do fix it!

michaelpj (approved):
Thanks for tackling this!

PR #4174, opened 2024-04-14
enable ThreadId for when testing

soulomoon:
It seems we do have it already, I enable it for default when testing.

PR #4180, opened 2024-04-17
Fix up the format problem

soulomoon:
Fix up format introduce in #4174


jhrcek:
I included this fix in https://github.com/haskell/haskell-language-server/pull/4181

soulomoon:
closing since we have https://github.com/haskell/haskell-language-server/pull/4181

PR #4182, opened 2024-04-19
4173 extracts ghcide tests and migrate it to use hls test utils [Migrate OutlineTests.hs]

soulomoon:


hls-test-utils/src/Test/Hls.hs
```diff
@@ -363,6 +365,78 @@ initialiseTestRecorder envVars = do
 -- ------------------------------------------------------------
 -- Run an HLS server testing a specific plugin
 -- ------------------------------------------------------------
+class TestRunner cont res where
```
michaelpj, line 368: Why do we need this? I can see you're abstracting over some common patterns, but I would have thought we could do that with a wrapper function or similar rather than needing a typeclass?

hls-test-utils/src/Test/Hls/FileSystem.hs
```diff
@@ -66,6 +67,7 @@ data VirtualFileTree =
 data FileTree
   = File FilePath Content
   | Directory FilePath [FileTree]
+  | CopiedDirectory FilePath
```
michaelpj, line 70: how is this different? some haddock would be helpful

michaelpj (commented):
ghcide test changes look good, I'm a little unsure about the refactoring of the test support functions

hls-test-utils/src/Test/Hls/FileSystem.hs
fendor: Looks like it corresponds to a directory glob.

michaelpj: I guess I'm confused about how it differs from the previous entry :thinking: Do we need two ways of specifying directories?

soulomoon: I have to manually specified the whole virtual tree down to the file levels.
`CopiedDirectory` allow us to simply copied an existing folder to vfs.
Matching those test in ghcide depend on just a folder in the `test/data`.
Yep, maybe a doc comment need to be added

hls-test-utils/src/Test/Hls.hs
soulomoon: It provide us a way to expose the root of the vsf during the test.
Some test in ghcide test need that.

soulomoon: It can not be don with `wrapper ` since it need to be exposed from the inside, I do find a nicer way to do it.

hls-test-utils/src/Test/Hls/FileSystem.hs
```diff
@@ -99,15 +101,23 @@ materialise rootDir' fileTree testDataDir' = do
...
+        nodes <- copyDir' testDataDir' name
+        mapM_ (persist root) nodes
 
   traverse_ (persist rootDir) fileTree
   pure $ FileSystem rootDir fileTree testDataDir
+    where -- | Copy a directory into a test project.
+        copyDir' :: FilePath -> FilePath -> IO [FileTree]
+        copyDir' root dir = do
+            files <- listDirectory (root </> dir)
+            traverse (\f -> pure $ copy (dir </> f)) files
```
fendor, line 120: I think we should only copy files that have been added to the git index. Cabal testsuite does something similar: https://github.com/haskell/cabal/blob/master/cabal-testsuite/src/Test/Cabal/Monad.hs#L417

soulomoon: Great, let's copy some of the cabal code

soulomoon: One concern though, does it mean we have to stage the files if newly test files are added.

soulomoon: Right the `--modified` should list unstaged files

hls-test-utils/src/Test/Hls.hs
michaelpj: Could we just expose it to all tests and mostly ignore it?

soulomoon: I updated to use wrapper version. But not sure if it is better than typeclass version🤔

hls-test-utils/src/Test/Hls/FileSystem.hs
soulomoon: added

soulomoon:
I have done some update base on the reviews, see what is still need to be improve?

Issue #4093, opened 2024-02-24
Flaky test failure result in error of `GetLinkable`


soulomoon:
update

=>IsEvaluating is running and already get the files for evaluations,
=>enqeueue file for evaluation
=>setSomethingModified dirty the IsEvaluating key in shakeExtre
=>IsEvaluating is finished, diry key removed from shakeExtra
=>session restarting and flushing the dirty keys of shakeExtra to hls-graph database
=>IsEvaluating is not updating since hls-graph database do not know it is dirty.

=======================
This is the issue discovered by @jhrcek 
we can run the test
```
cabal test hls-eval-plugins-test --test-options="-p /:i behaves exactly the same as  :info/"
```

> Today I spent some time looking into flaky test failures in CI.
Downloaded the logs from all failed testing jobs over the past 14 days.
The good news is that there are like 10 tests that keep failing relatively often, so fixing some of those could significantly reduce the overall flakiness :)

> One of the most frequently failing tests is this [group of tests](https://github.com/haskell/haskell-language-server/blob/409bf3ba830b3635290f43973ebad2211bb63018/plugins/hls-eval-plugin/test/Main.hs#L171-L210) whose failures have the same root cause.

> They essentially run bunch of eval plugin's code lenses and then wait for workspace/applyEdit message from the server, which very often comes, but from time to time (can reproduce this in like once every 10 runs) it doesn't come. In such cases there is this error in server logs:

> 2024-02-13T12:32:05.980696Z | Error | eval: Internal Error: BadDependency "GetLinkable"

> Few lines before that there's this: diagnostic message: "message": "/home/jhrcek/.cache/ghcide/test-0.1.0.0-inplace-05a647850b95fcd2b14858d164f368e84f7f8463/TI_Info.hi.core: getFileStatus: does not exist (No such file or directory)",

> There's only one usage of this this GetLinkable thing in eval plugin [here](https://github.com/haskell/haskell-language-server/blob/409bf3ba830b3635290f43973ebad2211bb63018/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs#L252)

> I'm pretty sure the "file doesn't exist" exception is coming from [this line](https://github.com/haskell/haskell-language-server/blob/409bf3ba830b3635290f43973ebad2211bb63018/ghcide/src/Development/IDE/Core/Rules.hs#L1121)

This merge from @wz1000 alleviate the situation https://github.com/haskell/haskell-language-server/pull/4076
but

[...]

PR #4185, opened 2024-04-21
Characterize eval plugin race leading to GetLinkable errors

jhrcek:
Here's the full characterization of the race that leads to occasional GetLinkable errors that are still reproducible on master ([see steps to reproduce](https://github.com/haskell/haskell-language-server/issues/4093#issuecomment-2061255229))

Here's my best attempt to describe the race condition succinctly:
1. we end up in [this error](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L1068) branch sometimes because hieCoreFp is Nothing
2. It is Nothing because getLinkableType [here](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L939) returns Nothing, which leads to writeCoreFileIfNeeded on [this line](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L943C29-L943C50) to NOT write a core file.

So the question is why does getLinkableType sometimes return Nothing (instead of `Just BCOLinkable` which would be required for writing core file)?

3. eval plugin [redefines the `NeedsCompilation` rule](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs#L122-L131) - so it returns `linkableType == Just BCOLinkable` when the file is evaluating
4. This rule uses [`IsEvaluating` rule](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs#L109-L114) to decide if given file is evaluating. This rule consults a global variable that stores "set of files being evaluated"
5. Now comes the racy part: the file is present in the "is evaluating" set only temporarily. It is inserted there by [queueForEvaluation](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs#L56-L65) and then later removed by unqueueForEvaluation.

EDIT:  @soulomoon pointed out mistake in my explanation - the error we see in the tests is a different than the original explanation claimed. Explanation corrected slightly, but it doesn't change what the actual race condition is.
Here is the corrected explanation in a video form walking through the code
[Explanation.webm](https://github.com/haskell/haskell-language-server/assets/2716069/4eaa3413-e4d9-4951-943f-fe1107e9c593)

The race condition is that IsEvaluating returns False (and thus linkableType ends up as Nothing, so .core file is not generated) in case that queueForEvaluation is called AFTER the isEvaluating check is done or if unqueueForEvaluation is called BEFORE isEvaluating check is done.

Here are 2 examples of one successful test and one failing test.
The thing that makes the difference is whether IsEvaluating rule is triggered WHILE given file is in the "set of files being evaluated".
Screenshot from 2024-04-21 07-44-20
Screenshot from 2024-04-21 07-37-38


soulomoon:
I am wondering under what circumstances we are using `unqueueForEvaluation`, is that another shutdown problem?

jhrcek:
No, it's only used in one place, in this bracket_:
https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/CodeLens.hs#L213-L218

soulomoon:
```
ThreadId 403 | 2024-04-17T13:02:14.285792Z | Error | eval: Internal Error: BadDependency "GetLinkable"
...
 "message": "called GetLinkable for a file without a linkable: NormalizedFilePath \"/tmp/hls-test-root/extra-dir-56738026422965/TI_Info.hs\"\nCallStack (from HasCallStack):\n  error, called at src/Development/IDE/Core/Rules.hs:1068:18 in ghcide-2.7.0.0-inplace:Development.IDE.Core.Rules"
```
I don't quite follow why we have the above error if we have getLinkableType(Result = Nothing) It seems we have 
`GetLinkable => GetModIface(result hirCoreFp=Nothing) => writeCoreFileIfNeeded(result mbHiFile=Nothing) => writeCoreFileIfNeeded(condition: Maybe LinkableType is not Nonthing)` => getLinkableType(result is not Nothing)

```haskell
writeCoreFileIfNeeded :: ShakeExtras -> HscEnv -> Maybe LinkableType -> Action (IdeResult ModGuts) -> TcModuleResult -> Action (IdeResult HiFileResult)
writeCoreFileIfNeeded _ hsc Nothing _ tmr = do
  incrementRebuildCount
  res <- liftIO $ mkHiFileResultNoCompile hsc tmr
  pure ([], Just $! res)
 ....
  
getLinkableRule :: Recorder (WithPriority Log) -> Rules ()
getLinkableRule recorder =
  defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GetLinkable f -> do
    ModSummaryResult{msrModSummary = ms} <- use_ GetModSummary f
    HiFileResult{hirModIface, hirModDetails, hirCoreFp} <- use_ GetModIface f
    let obj_file  = ml_obj_file (ms_location ms)
        core_file = ml_core_file (ms_location ms)
    case hirCoreFp of
      Nothing -> error $ "called GetLinkable for a file without a linkable: " ++ show f
...


getModIfaceRule :: Recorder (WithPriority Log) -> Rules ()
getModIfaceRule recorder = defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GetModIface f -> do
  fileOfInterest <- use_ IsFileOfInterest f
  res <- case fileOfInterest of
    IsFOI status -> do
      -- Never load from disk for files of interest
      tmr <- use_ TypeCheck f
      linkableType <- getLinkableType f
      hsc <- hscEnv <$> use_ GhcSessionDeps f
      let compile = fmap ([],) $ use GenerateCore f
      se <- getShakeExtras
      (diags, !mbHiFile) <- writeCoreFileIfNeeded se hsc linkableType compile tmr
      let fp = hiFileFingerPrint <$> mbHiFile
      hiDiags <- case mbHiFile of
        Just hiFile
          | OnDisk <- status
          , not (tmrDeferredError tmr) -> liftIO $ writeHiFile se hsc hiFile
        _ -> pure []
      return (fp, (diags++hiDiags, mbHiFile))
....
```

jhrcek:
I'm not sure I follow your reasoning. I don't understand what you mean by

> writeCoreFileIfNeeded(condition: Maybe LinkableType is not Nonthing)

I think the problem is that when you call `writeCoreFileIfNeeded` with `Maybe LinkableType` argument set to Nothing, the .core file is not written to disk. But the rule fails even before touching the core file, because we throw error every time when getLinkableType results in Nothing: https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L1072

soulomoon:
> I'm not sure I follow your reasoning. I don't understand what you mean by
> 
> 
> 
> > writeCoreFileIfNeeded(condition: Maybe LinkableType is not Nonthing)
> 
> 
> 
> I think the problem is that when you call `writeCoreFileIfNeeded` with `Maybe LinkableType` argument set to Nothing, the .core file is not written to disk. But the rule fails even before touching the core file, because we throw error every time when getLinkableType results in Nothing: https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L1072

This seems to be a different error than the one in https://github.com/haskell/haskell-language-server/issues/4093#issuecomment-2061255229

jhrcek:
Ah, you're right. We end up in the [first error](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L1068) branch because `hirCoreFp` is Nothing.

But I still don't get why you think we should end up calling `writeCoreFileIfNeeded(condition: Maybe LinkableType is not Nonthing)`

soulomoon:
> Ah, you're right. We end up in the [first error](https://github.com/haskell/haskell-language-server/blob/9593d04a76e024942981b1333bfb2558a6ae0dab/ghcide/src/Development/IDE/Core/Rules.hs#L1068) branch because `hirCoreFp` is Nothing.
> 
> But I still don't get why you think we should end up calling `writeCoreFileIfNeeded(condition: Maybe LinkableType is not Nonthing)`

If `writeCoreFileIfNeeded(result mbHiFile=Nothing)`, than its arguments `Maybe LinkableType` must not be Nothing, since if `Maybe LinkableType` is Nothing, `writeCoreFileIfNeeded(result mbHiFile=Just xxx)`

jhrcek:
But as you can see in the screenshot of failing test case linkableType=Nothing (printed by [this trace](https://github.com/haskell/haskell-language-server/pull/4185/files#diff-65a56362c2b3593800cd8369e6cfab1e92c347a9be97d1f0b8793f0e5dd9d524R943))

soulomoon:
> But as you can see in the screenshot of failing test case linkableType=Nothing (printed by [this trace](https://github.com/haskell/haskell-language-server/pull/4185/files#diff-65a56362c2b3593800cd8369e6cfab1e92c347a9be97d1f0b8793f0e5dd9d524R943))

Right, It confused me :(

jhrcek:
@soulomoon description updated based on your pointers. I also added a video walkthrough of the code showing how we get to the error branch.

wz1000:
I still don't understand why isEvaluating returns false after we queue a file for evaluation, because `setSomethingModified VFSUnmodified st [toKey IsEvaluating nfp] "Eval"` is supposed to invalidate the `IsEvaluating` key and then restart the build session.

jhrcek:
> I still don't understand why isEvaluating returns false after we queue a file for evaluation, because setSomethingModified VFSUnmodified st [toKey IsEvaluating nfp] "Eval" is supposed to invalidate the IsEvaluating key and then restart the build session.

I didn't check this deeply yet, so this is just me theorizing now. But if you invalidate IsEvaluating, the NeedsCompilation rule might already be in flight (e.g in the True if branch [here](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs#L125)) and might have already decided not to (re)generate core file - I think this is what we see in the failed test screenshot.

wz1000:
I believe that any on going compile should have been cancelled by the restart and the NeedsCompilation rule should be re-evaluated.

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Rules.hs
```diff
@@ -56,13 +56,13 @@ instance IsIdeGlobal EvaluatingVar
 queueForEvaluation :: IdeState -> NormalizedFilePath -> IO ()
 queueForEvaluation ide nfp = do
     EvaluatingVar var <- getIdeGlobalState ide
-    atomicModifyIORef' var (\fs -> (Set.insert nfp fs, ()))
+    atomicModifyIORef' var (\fs -> (trace ("TRACE: queueForEvaluation: " <> show nfp ) $ Set.insert nfp fs, ()))
```
soulomoon, line 59: Maybe we can add the trace here instead to make sure the restart session take effect.
```
              (do queueForEvaluation st nfp
                  setSomethingModified VFSUnmodified st [toKey IsEvaluating nfp] "Eval"
                  (trace ("TRACE: queueForEvaluation done")
```

soulomoon:
Some reproduction in `/aarch64-osx/ghc-9.8.2/`
```
TRACE: isMemberEvaluatingVar=False(NormalizedFilePath "/private/var/folders/36/_743psv11gv2wrj9dclrpd500000gn/T/hls-test-root/extra-dir-93896921344/TI_Info.hs")
TRACE: isEvaluating=False(NormalizedFilePath "/private/var/folders/36/_743psv11gv2wrj9dclrpd500000gn/T/hls-test-root/extra-dir-93896921344/TI_Info.hs")
TRACE: queueForEvaluation: NormalizedFilePath "/private/var/folders/36/_743psv11gv2wrj9dclrpd500000gn/T/hls-test-root/extra-dir-93896921344/TI_Info.hs"
TRACE: linkableType=Nothing (NormalizedFilePath "/private/var/folders/36/_743psv11gv2wrj9dclrpd500000gn/T/hls-test-root/extra-dir-93896921344/TI_Info.hs")
TRACE: unqueueForEvaluation: NormalizedFilePath "/private/var/folders/36/_743psv11gv2wrj9dclrpd500000gn/T/hls-test-root/extra-dir-93896921344/TI_Info.hs"
```

jhrcek:
Indeed, the "Timed out waiting to receive a message from the server." is the standard way this test fails.
The error itself doesn't appear in the test output by default - You have to enable logging of messages/server output via lsp-test's env variables to see the error printed:
`LSP_TEST_LOG_MESSAGES=1 LSP_TEST_LOG_STDERR=1 cabal test ...`

soulomoon:
Right, I accidentally added a line break after `LSP_TEST_LOG_MESSAGES=1 LSP_TEST_LOG_STDERR=1` in my run script

jhrcek:
I see we're already logging shake session restarts, so here's the restart within the failing tests in the context of other messages.

The restart works a bit differently than I imagined in the sense that the "currently running build actions" are simply cancelled (unless they are enqueued in which case they are requeued for later execution).
See this comment:
https://github.com/haskell/haskell-language-server/blob/9b0699dd92fbfa986600e6262220ef305df84816/ghcide/src/Development/IDE/Core/Shake.hs#L759-L761

Screenshot from 2024-04-22 14-55-42

jhrcek:
Here is the full log of a single failing test run if you want to verify some things about it: https://gist.github.com/jhrcek/74fbd87d3d34190aef2ce0de4e99801f

jhrcek:
I got a test failure with your PR branch after 95 iterations.
Here are the logs of the failed run: https://gist.github.com/jhrcek/96cde598185cb7abf0842ef0b0a47dea

soulomoon:
> I got a test failure with your PR branch after 95 iterations.
Yep, it won't help, now I realized. `initialiseSessionForEval` is already running using the `shakeEnqueue`, which should already pick up the `IsEvaluating` changed when running for rule `GetLinkable` in the session.

soulomoon:
For the above change, after doing 500 runs, the error does not appear locally.

jhrcek:
Updated based on lateset findings from @soulomoon 
added few more traces to demonstrate that IsEvaluating key is removed from dirty key set before restart of shake session happens. CC @wz1000 

Screenshot from 2024-04-24 07-10-43

jhrcek:
No longer needed as the issue has been fixed in https://github.com/haskell/haskell-language-server/pull/4190.
Thanks @soulomoon!

PR #4190, opened 2024-04-22
Stabilize the build system by correctly house keeping the dirtykeys and rule values [flaky test #4185 #4093]

soulomoon:
## What is currently did up to [240254e](https://github.com/haskell/haskell-language-server/pull/4190/commits/240254e9cdaf4f392b08b7b30ddd9e1eab19d369)
The main problem is the out of sync state in the build system. Several states involved, the shake database running result state for key. shake extra's dirty state for key and shake extra's rule values state.
To stablize the build system. This PR force some of the updates of these state into a single STM block.

1.  [x] collect dirtykeys and ship it to session restart directly. No more invalid removal before session restart. Fixing #4093 [Verified fixed locally 500 rounds]
2. [x] move the dirtykey removal and rules values updating to hls-graph by adding a call back fucntion to `RunResult`. Properly handle the dirtykeys and rule value state after session start and closely followed by another session restart Fixing #4194 [Verified fixed locally 500 rounds]
3. [x] properly handle clean up by wrapping the `asyncWithCleanUp` to `refreshDeps`
### possible future improvement (Should be in some other PR)
*. we can avoid giving multiple update to dirty states across multiple session if the key is still dirty but does not marked as new dirty, by sending the collected keys directly to newsession. But `garbageCollectKeys` is using it, so it is not a good idea to do it now. Maybe in the future. 
*. Maybe we can schedule some debouncer to the session restart. I've observed some times the consective session restart might happen.

## The problem
Some proof of concept for fixing problem at https://github.com/haskell/haskell-language-server/pull/4185 #4093 discovered and tracked down by @jhrcek with his impressive bug tracking technique.

Here should be what might be happeing for the problem:
```
=>IsEvaluating is running and already get the files for evaluations,
=>enqeueue file for evaluation
=>setSomethingModified dirty the IsEvaluating key in shakeExtre
=>IsEvaluating is finished, diry key removed from shakeExtra
=>session restarting and flushing the dirty keys of shakeExtra to hls-graph database
=>IsEvaluating is not updating since hls-graph database do not know it is dirty.
```
The essential problem is that a running key is being marked as dirty, and finished running and removed itself from the dirtyset in shakeExtra just before a sessionRestart flush it to rerun in the hls-graph database, and hence the dirty mark is incorrectly lost.

The worm bed is that we are bookkeeping two set of dirtiness, one in hls-graph database(has three states, running, dity, clean) and one in shake-extra(two states, dirty, clean).
It is nasty to keep them in sync.

## Possible fixes

Some possible solutions:

0. We are passing keys need to be updated directly to the `restartShakeSession` instead of just recording it in the dirtykeys of `shakeExtra`. Which keep them from being removed right before the restartShakeSession.
But might need to track out each such case and safe guard them to restartShakeSession.

1. when mark the dirtykeys in `shakeExtra`, we also mark them in `shakeDatabase`, but we might need to dirty all its reversedeps to make sure it would break dirtiness of them, I do not see how much better can this be better than shakerestart
  

2. When keys being added to dirtykeys in `shakeExtra`, we restart the shake session at the same time, leaving no gap in between. @michaelpj 'idea that we can use VarT instead of MVar and wrap the key changed and session restart both togather. 

3. ~we add a running state for keys in `shakeExtra`, avoid remove the dirtykeys in `shakeExtra` if the running state is gone. Or some thing similar to expand the possible state that can be represent in `shakeExtra`. We might need extra book keeping a lot more keys.  but simplestz~ not workable since the update have to be in hls-graph
 
 4. create a new temp dirtykeyset to record new dirty keys, and only flush it to dirtykeyset inside `restartShakeSession`


michaelpj:
So basically we don't want to delete the dirty keys outside the `withMVar` lock on the `shakeSession`? That makes sense. I guess the thing that's a shame here is that these modifications don't compose - which suggests to me it would be nicer if we could use STM. Would it be possible to make `shakeSession` a `TVar` instead of an `MVar`?

soulomoon:
> So basically we don't want to delete the dirty keys outside the `withMVar` lock on the `shakeSession`? That makes sense. I guess the thing that's a shame here is that these modifications don't compose - which suggests to me it would be nicer if we could use STM. Would it be possible to make `shakeSession` a `TVar` instead of an `MVar`?

We might have dirty key loss if we are marking them when the shake session is running

soulomoon:
Another option would be to expand the state of keys in shakeExtra to have three state 
shake-extra(running, dirty, clean), we have git rid of the dirty key lost problem

wz1000:
This PR seems to have a bunch of unrelated changes.

Could you also write down a commit message explaining the problem and solution when you find the time please?

michaelpj:
> Another option would be to expand the state of keys in shakeExtra to have three state
shake-extra(running, dirty, clean), we can get rid of the dirty key lost problem.

Yes, that could also work. So then if a rule finishes it updates the state to 'clean' if it was 'running', but not if it has been set back to 'dirty' while it was in progress.

soulomoon:
> This PR seems to have a bunch of unrelated changes.
> 
> Could you also write down a commit message explaining the problem and solution when you find the time please?

This PR is just a proof of concept to fix the problem we have encounter at https://github.com/haskell/haskell-language-server/pull/4185 for now. Sorry I forget to addresss that in the main sector

soulomoon:
> Would it be possible to make shakeSession a TVar instead of an MVar?

I review relevant part a bit, it seems to be hard to do so, since a lot of IO is entangling

soulomoon:
> > Another option would be to expand the state of keys in shakeExtra to have three state
> > shake-extra(running, dirty, clean), we can get rid of the dirty key lost problem.
> 
> Yes, that could also work. So then if a rule finishes it updates the state to 'clean' if it was 'running', but not if it has been set back to 'dirty' while it was in progress.

This seems to have the least changed, I have pushed this branch.

jhrcek:
> tracked down by @jhrcek with his impressive bug tracking technique.

That's kind of you to say, but it's just good old "printf debugging", I wish there was easier way :smile:

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1234,7 +1252,9 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
                     (if eq then ChangedRecomputeSame else ChangedRecomputeDiff)
                     (encodeShakeValue bs) $
                     A res
-        liftIO $ atomicallyNamed "define - dirtyKeys" $ modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
+        liftIO $ atomicallyNamed "define - (runningKeys, dirtyKeys)" $ do
+            running <- readTVar runningKeys
+            when (memberKeySet theKey running) $ return (deleteKeySet theKey running) >> modifyTVar' dirtyKeys (deleteKeySet theKey)
```
jhrcek, line 1257: This looks suspicious.
deleteKeySet is pure function which creates a new set with the key deleted.
But you seem to be throwing the that new set away (using >>).
I'd expect something like
```haskell
when (memberKeySet theKey running) $ do 
   modifyTVar' runningKeys (deleteKeySet theKey)
   modifyTVar' dirtyKeys (deleteKeySet theKey)
```

soulomoon: You are right, I make a mistake, fixed now

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -573,11 +576,22 @@ recordDirtyKeys
...
     modifyTVar' dirtyKeys $ \x -> foldl' (flip insertKeySet) x (toKey key <$> file)
     return $ withEventTrace "recordDirtyKeys" $ \addEvent -> do
         addEvent (fromString $ unlines $ "dirty " <> show key : map fromNormalizedFilePath file)
 
+recordDirtyKeySet
+  :: ShakeExtras
+  -> [Key]
+  -> STM (IO ())
+recordDirtyKeySet ShakeExtras{dirtyKeys, runningKeys} keys = do
+    modifyTVar' runningKeys $ \x -> foldl' (flip deleteKeySet) x keys
```
michaelpj, line 590: I think this is a little counter-intuitive. It means that we will remove a key from the running set _while it is in fact running_. But I think maybe we just need to rename the "running keys" set. It's something like "running keys which will produce non-dirty results". With that interpretation it is fine to remove the key here, since we are denying that it will produce a non-dirty result when it finishes.

This also all definitely needs a Note to explain what's going on!

michaelpj (commented):
I was wondering about writing a test for this, and that made me wonder why dirtiness checking isn't part of `hls-graph`, where it would be easier to test in isolation. Does anyone know?

michaelpj:
In fact `hls-graph` does have dirtiness-related code, e.g.: https://github.com/haskell/haskell-language-server/blob/master/hls-graph/src/Development/IDE/Graph/Database.hs#L40

Why do we have two versions of this? :(

michaelpj:
And indeed, it already correctly handles this case!

michaelpj:
Answers are in https://github.com/haskell/haskell-language-server/pull/1862, sadly no Note explaining it. I'm still suspicious, though.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: Yep, maybe rename to something like `runningCleanlyKeys` to indicate that it have not been polluted

michaelpj:
So it looks like ultimately the intention is that these dirty keys get passed to `hls-graph`, which should handle whether or not to rerun things. And it does look like it should handle the case of things being marked as dirty while something is running, but we don't currently allow that? I don't really understand the details enough. But in principle I think it might simplify things a lot if we were simply able to tell `hls-graph` that things are dirty as soon as we know that, and allow it to track the status. I have no idea how difficult that would be, though.

soulomoon:
To mark a key in hls-graph dirty, we need to track all the reverse dependency keys, otherwise, we might have invalid build state.

> if we were simply able to tell hls-graph that things are dirty as soon as we know that

It is the kind of thing our `RestartShakeSession` is supposed to do. So the dirtiness kept in `ShakeExpra` is more like a preparation to mark the dritiness in `hls-graph`.

I can imagine we can do it without stop the session. That is to mark keys and its inverse deps in one atomic session. But it does seems a lot of things depend on the `RestartShakeSession` right now in our codebase.

michaelpj:
Yeah, it very much looks like `hls-graph` is supposed to be run in distinct sessions, not "fully-online" where you can poke it while it's going. So probably that would be a big change.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -328,6 +329,8 @@ data ShakeExtras = ShakeExtras
       -- ^ Default HLS config, only relevant if the client does not provide any Config
     , dirtyKeys :: TVar KeySet
       -- ^ Set of dirty rule keys since the last Shake run
+    , runningKeys:: TVar KeySet
+      -- ^ Set of running rule keys since the last Shake run
```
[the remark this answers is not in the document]
soulomoon: Yep, do worth documentation.
They are related in only one way, do not remove it from `dirtyKeys` if not in the `runningKeys`(should rename to `runningCleanlyKeys`).

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -573,11 +576,22 @@ recordDirtyKeys
   -> k
   -> [NormalizedFilePath]
   -> STM (IO ())
-recordDirtyKeys ShakeExtras{dirtyKeys} key file = do
+recordDirtyKeys ShakeExtras{dirtyKeys, runningKeys} key file = do
```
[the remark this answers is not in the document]
soulomoon: yep, done

michaelpj:
We _could_ also imitate `hls-graph` here, which has instead a key status enumeration which can be clean/running/dirty. Might be slower since we won't be using `KeySet`.

ghcide/src/Development/IDE/Core/Shake.hs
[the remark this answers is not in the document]
soulomoon: done

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: done

soulomoon:
I think the flakiness in the test is also fading away.
Not seeing them yet!

wz1000:
Instead of this dance with `runningCleanlyKeys`, can we mark keys and dirty after stopping the current session (the call to `cancelShakeSession`) and before restarting the session?

soulomoon:
> Instead of this dance with `runningCleanlyKeys`, can we mark keys and dirty after stopping the current session (the call to `cancelShakeSession`) and before restarting the session?

Yes we can, that was the original solution, just safe guard the dirty keys to restart session.

Cons: less state to maintain.
Downs: We may need refactor a lot of places.

wz1000:
I think I prefer that, there don't seem to be too many places to refactor. We also avoid making difficult to understand changes to the garbage collection etc.

soulomoon:
ok, I push that branch up now, is that something that you have in your mind?

Some thing need to be done refactor `recordDirtyKeys` to return keys, and modified those places to tug thoes keys directly to shakeRestart.

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -251,7 +251,7 @@ setSomethingModified vfs state keys reason = do
         writeTQueue (indexQueue $ hiedbWriter $ shakeExtras state) (\withHieDb -> withHieDb deleteMissingRealFiles)
         modifyTVar' (dirtyKeys $ shakeExtras state) $ \x ->
```
wz1000, line 252: we don't need this anymore?

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -224,7 +224,7 @@ setFileModified recorder vfs state saved nfp = do
           CheckOnSave -> saved
           _           -> False
     join $ atomically $ recordDirtyKeys (shakeExtras state) GetModificationTime [nfp]
```
wz1000, line 226: this needs to be pushed into the `restartShakeSession`?

wz1000:
It seems much cleaner to tell the build system which keys are dirty when restarting.

ghcide/src/Development/IDE/Core/FileStore.hs
soulomoon: right, I'll need track those usage down.

wz1000: I think we can just delete the `recordDirtyKeys` function and replace it by supplying arguments to `restartShakeSession`

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -759,13 +760,14 @@ delayedAction a = do
 -- | Restart the current 'ShakeSession' with the given system actions.
 --   Any actions running in the current session will be aborted,
 --   but actions added via 'shakeEnqueue' will be requeued.
-shakeRestart :: Recorder (WithPriority Log) -> IdeState -> VFSModified -> String -> [DelayedAction ()] -> IO ()
-shakeRestart recorder IdeState{..} vfs reason acts =
+shakeRestart :: Recorder (WithPriority Log) -> IdeState -> VFSModified -> String -> [DelayedAction ()] -> [Key] -> IO ()
+shakeRestart recorder IdeState{..} vfs reason acts keys =
```
wz1000, line 764: Perhaps we want something like `IO ()` instead of `Key` so that we can safely perform arbitrary actions in between build sessions, so that things like `SMethod_WorkspaceDidChangeWatchedFiles` which update the state can be safely executed.

wz1000: Or perhaps `IO KeySet`, so that we can perform effects and return a set of keys to be invalidated.

soulomoon: make sense, I'll see how is that working out.

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -106,11 +107,11 @@ getFileExistsMapUntracked = do
   return v
 
 -- | Modify the global store of file exists.
-modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO ()
+modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO [Key]
 modifyFileExists state changes = do
   FileExistsMapVar var <- getIdeGlobalState state
   -- Masked to ensure that the previous values are flushed together with the map update
-  join $ mask_ $ atomicallyNamed "modifyFileExists" $ do
+  keys <- join $ mask_ $ atomicallyNamed "modifyFileExists" $ do
```
wz1000, line 114: we don't need the `join` if we get rid of the double return.

soulomoon: right done

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -611,19 +610,16 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
-          invalidateShakeCache
-
           -- The VFS doesn't change on cradle edits, re-use the old one.
-          restartShakeSession VFSUnmodified "new component" []
+          restartShakeSession VFSUnmodified "new component" [] $ do
+            void $ modifyVar' fileToFlags $
+                Map.insert hieYaml this_flags_map
+            void $ modifyVar' filesMap $
+                flip HM.union (HM.fromList (map ((,hieYaml) . fst) $ concatMap toFlagsMap all_targets))
+            key1 <- extendKnownTargets all_targets
```
soulomoon, line 619: I am not sure running all these between sessions is a good idea @wz1000

wz1000: `fileToFlags` and `filesMap` don't need to updated in between sessions because they aren't part of the shake state, but I can't thinking that could go wrong by updating them here.

It is probably a good idea to call `extendKnownTargets` and `invalidateShakeCache` inside the callback though.

soulomoon: I do have a performance concern, pushing too many things into the restart callback would slow down the build system by increasing its halting time.

soulomoon:
Performance choice Put more logic between session restart vs Put less logic between session restart
choice 1: gap between sessions would be longer.
choice 2:dirty key can not be correctly invalidated before session restart.

ghcide/session-loader/Development/IDE/Session.hs
wz1000: I think its fine to move the `modifyVar'` outside the callback.

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -106,11 +107,11 @@ getFileExistsMapUntracked = do
   return v
 
 -- | Modify the global store of file exists.
-modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO ()
+modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO [Key]
 modifyFileExists state changes = do
   FileExistsMapVar var <- getIdeGlobalState state
   -- Masked to ensure that the previous values are flushed together with the map update
-  join $ mask_ $ atomicallyNamed "modifyFileExists" $ do
+  keys <- mask_ $ atomicallyNamed "modifyFileExists" $ do
```
wz1000, line 114:
```suggestion
  mask_ $ atomicallyNamed "modifyFileExists" $ do
```

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -244,14 +247,13 @@ typecheckParentsAction recorder nfp = do
...
+setSomethingModified vfs state reason actionBetweenSession = do
     -- Update database to remove any files that might have been renamed/deleted
-    atomically $ do
-        writeTQueue (indexQueue $ hiedbWriter $ shakeExtras state) (\withHieDb -> withHieDb deleteMissingRealFiles)
-        modifyTVar' (dirtyKeys $ shakeExtras state) $ \x ->
-            foldl' (flip insertKeySet) x keys
-    void $ restartShakeSession (shakeExtras state) vfs reason []
+    void $ restartShakeSession (shakeExtras state) vfs reason [] $ do
+        keys <- actionBetweenSession
+        atomically $ writeTQueue (indexQueue $ hiedbWriter $ shakeExtras state) (\withHieDb -> withHieDb deleteMissingRealFiles)
```
wz1000, line 255: This doesn't need to be in the callback.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -568,15 +569,13 @@ deleteValue ShakeExtras{dirtyKeys, state} key file = do
...
-  :: Shake.ShakeValue k
-  => ShakeExtras
-  -> k
-  -> [NormalizedFilePath]
+  :: ShakeExtras
+  -> [Key]
   -> STM (IO ())
-recordDirtyKeys ShakeExtras{dirtyKeys} key file = do
-    modifyTVar' dirtyKeys $ \x -> foldl' (flip insertKeySet) x (toKey key <$> file)
+recordDirtyKeys ShakeExtras{dirtyKeys} keys = do
```
wz1000, line 575: I don't think this function serves much of a purpose anymore, it can be deleted. The trace isn't very important.

soulomoon:
I am seeing some strange error production, look llike is being garbage collected
```
Keys: [ GhcSessionIO; 
      , GetKnownTargets; 
      , GetModificationTime; /tmp/extra-dir-39469908493613/A.hs
      , IsFileOfInterest; /tmp/extra-dir-39469908493613/A.hs ]
Aborting previous build session took 0.00s 
Debug | Finished: InitialLoad Took: 0.00s
Debug | Finished build session
Could not find key: NonLocalCompletions; /tmp/extra-dir-39469908493613/A.hs
  errorIO, called at src/Development/IDE/Graph/Internal/Rules.hs:49:25 in hls-graph-2.7.0.0-inplace:Development.IDE.Graph.Internal.Rules
```

soulomoon:
This is ready for another review. What is done is recorded at https://github.com/haskell/haskell-language-server/pull/4190#issue-2256944513

ghcide/src/Development/IDE/Core/FileStore.hs
soulomoon: done

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -516,10 +515,10 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
                             HM.unionWith (<>) k $ HM.fromList knownTargets
                 hasUpdate = if known /= known' then Just (unhashed known') else Nothing
             writeTVar knownTargetsVar known'
-            logDirtyKeys <- recordDirtyKeys extras GetKnownTargets [emptyFilePath]
-            return (logDirtyKeys >> pure hasUpdate)
+            return (pure hasUpdate)
```
jhrcek, line 518: I think you can remove the `join` wrapping this block if you remove this extraneous return or pure

```suggestion
            pure hasUpdate
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -516,10 +515,10 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
           for_ hasUpdate $ \x ->
             logWith recorder Debug $ LogKnownFilesUpdated x
+          return $ toKey GetKnownTargets emptyFilePath
```
jhrcek, line 521:
```suggestion
          return $ toNoFileKey GetKnownTargets
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -612,18 +611,14 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
-
+          void $ modifyVar' fileToFlags $ Map.insert hieYaml this_flags_map
+          void $ modifyVar' filesMap $ flip HM.union (HM.fromList (map ((,hieYaml) . fst) $ concatMap toFlagsMap all_targets))
+          key1 <- extendKnownTargets all_targets
+          key2 <- invalidateShakeCache
           -- The VFS doesn't change on cradle edits, re-use the old one.
-          restartShakeSession VFSUnmodified "new component" []
+          restartShakeSession VFSUnmodified "new component" [] $ do
+            return [key1, key2]
+            -- Invalidate all the existing GhcSession build nodes by restarting the Shake session
```
jhrcek, line 621: I think this Comment should be moved above `invalidateShakeCache`

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -106,11 +107,11 @@ getFileExistsMapUntracked = do
   return v
 
 -- | Modify the global store of file exists.
-modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO ()
+modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO [Key]
```
jhrcek, line 110: Can you please extend the haddock to make it clear what the returned `[Key]` means?
This doesn't make it clear what should be done with these returned keys, so I'd maybe add a note-style comment about this somewhere stating what problem these extra IO [Key] arguments/return types mean and how they are used.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1232,9 +1224,9 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
                         _                                     -> False
                 return $ RunResult
                     (if eq then ChangedRecomputeSame else ChangedRecomputeDiff)
-                    (encodeShakeValue bs) $
-                    A res
-        liftIO $ atomicallyNamed "define - dirtyKeys" $ modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
+                    (encodeShakeValue bs)
+                    (A res)
+                    (setValues state key file res (Vector.fromList diags) >> modifyTVar' dirtyKeys (deleteKeySet $ toKey key file))
```
jhrcek, line 1229: Nitpis: this is a bit hard to read on one line..

```suggestion
                    (encodeShakeValue bs)
                    (A res) $ do
                        setValues state key file res (Vector.fromList diags)
                        modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
```

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -119,10 +120,10 @@ modifyFileExists state changes = do
     -- flush previous values
     let (fileModifChanges, fileExistChanges) =
             partition ((== FileChangeType_Changed) . snd) changes
-    mapM_ (deleteValue (shakeExtras state) GetFileExists . fst) fileExistChanges
-    io1 <- recordDirtyKeys (shakeExtras state) GetFileExists $ map fst fileExistChanges
-    io2 <- recordDirtyKeys (shakeExtras state) GetModificationTime $ map fst fileModifChanges
-    return (io1 <> io2)
+    keys0 <- concat <$> mapM (deleteValue (shakeExtras state) GetFileExists . fst) fileExistChanges
```
jhrcek, line 123: Maybe slightly controversial tip (up to you if you like it more)
but you can use this instance here `Monoid a => Monoid (STM a)`:
```suggestion
    keys0 <- foldMap (deleteValue (shakeExtras state) GetFileExists . fst) fileExistChanges
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -562,21 +563,11 @@ deleteValue
   => ShakeExtras
   -> k
   -> NormalizedFilePath
-  -> STM ()
+  -> STM [Key]
 deleteValue ShakeExtras{dirtyKeys, state} key file = do
```
jhrcek, line 567:
```suggestion
deleteValue ShakeExtras{state} key file = do
```

ghcide/src/Development/IDE/Main.hs
```diff
@@ -18,8 +18,8 @@ import           Control.Concurrent.STM.Stats             (dumpSTMStats)
 import           Control.Exception.Safe                   (SomeException,
                                                            catchAny,
                                                            displayException)
-import           Control.Monad.Extra                      (concatMapM, unless,
-                                                           when)
+import           Control.Monad.Extra                      (concatMapM, join,
```
jhrcek, line 21: I suspect some unused imports here :smile: 
I'd do `cabal clean` and rebuilt to see them all ..

jhrcek (commented):
Thanks for working on this!
I don't claim to understand how all of that works, but flaky tests disappearing seem to be a good indication that it does.
Let's wait for reviews from others though

ghcide/src/Development/IDE/Core/FileExists.hs
jhrcek: Sorry for bad tip - it looks like this instance was introduces with base coming with ghc 9.4 and is not yet available with ghc 9.2 :/

soulomoon: right, I discovered that before, STM do not have the monoid instance, hit by it again lol

ghcide/session-loader/Development/IDE/Session.hs
soulomoon: done

ghcide/session-loader/Development/IDE/Session.hs
soulomoon: done

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -200,7 +200,9 @@ compute db@Database{..} stack key mode result = do
                     (getResultDepsDefault mempty previousDeps)
                     deps
         _ -> pure ()
-    atomicallyNamed "compute" $ SMap.focus (updateStatus $ Clean res) key databaseValues
+    atomicallyNamed "compute and run hook" $ do
+        runHook
+        SMap.focus (updateStatus $ Clean res) key databaseValues
```
michaelpj, line 205: Even though these are in the same STM transaction, I think it probably makes sense to do this first? Potentially the hook could look at this state? I'm not sure.

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -202,11 +203,10 @@ data RunResult value = RunResult
         -- ^ The value to store in the Shake database.
     ,runValue   :: value
         -- ^ The value to return from 'Development.Shake.Rule.apply'.
+    ,runHook    :: STM ()
+        -- ^ The hook to run after the rule completes.
```
michaelpj, line 207: Maybe give an example of what this is for? e.g. "This is useful if the rule needs to modify state which is being tracked outside the rule system"?

michaelpj: Although I guess it's then not clear to me why this has to happen in this hook rather than in the action. I'm sure you know - write it down somewhere!

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -474,10 +474,9 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
     clientConfig <- getClientConfigAction
     extras@ShakeExtras{restartShakeSession, ideNc, knownTargetsVar, lspEnv
                       } <- getShakeExtras
-    let invalidateShakeCache :: IO ()
-        invalidateShakeCache = do
+    let invalidateShakeCache = do
```
michaelpj, line 477: does this in fact invalidate anything any more?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -562,21 +563,11 @@ deleteValue
   => ShakeExtras
   -> k
   -> NormalizedFilePath
-  -> STM ()
-deleteValue ShakeExtras{dirtyKeys, state} key file = do
+  -> STM [Key]
```
michaelpj, line 566: what is this?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -759,12 +750,14 @@ delayedAction a = do
 -- | Restart the current 'ShakeSession' with the given system actions.
 --   Any actions running in the current session will be aborted,
 --   but actions added via 'shakeEnqueue' will be requeued.
-shakeRestart :: Recorder (WithPriority Log) -> IdeState -> VFSModified -> String -> [DelayedAction ()] -> IO ()
-shakeRestart recorder IdeState{..} vfs reason acts =
+shakeRestart :: Recorder (WithPriority Log) -> IdeState -> VFSModified -> String -> [DelayedAction ()] -> IO [Key] -> IO ()
+shakeRestart recorder IdeState{..} vfs reason acts ioActionBetweenShakeSession =
```
michaelpj, line 754: It's more like "compute dirty keys", right?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -759,12 +750,14 @@ delayedAction a = do
...
     withMVar'
         shakeSession
         (\runner -> do
               (stopTime,()) <- duration $ logErrorAfter 10 $ cancelShakeSession runner
+              keys <- ioActionBetweenShakeSession
+              atomically $ modifyTVar' (dirtyKeys shakeExtras) $ \x -> foldl' (flip insertKeySet) x keys
               res <- shakeDatabaseProfile shakeDb
               backlog <- readTVarIO $ dirtyKeys shakeExtras
```
michaelpj, line 762: Do we even need the `dirtyKeys` TVar now? I believe we now only set it just above and then we immediately read it!

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -148,24 +149,24 @@ isInterface :: NormalizedFilePath -> Bool
...
 -- | Reset the GetModificationTime state of interface files
-resetInterfaceStore :: ShakeExtras -> NormalizedFilePath -> STM ()
+resetInterfaceStore :: ShakeExtras -> NormalizedFilePath -> STM [Key]
 resetInterfaceStore state f = do
     deleteValue state GetModificationTime f
 
 -- | Reset the GetModificationTime state of watched files
 --   Assumes the list does not include any FOIs
-resetFileStore :: IdeState -> [(NormalizedFilePath, LSP.FileChangeType)] -> IO ()
+resetFileStore :: IdeState -> [(NormalizedFilePath, LSP.FileChangeType)] -> IO [Key]
```
michaelpj, line 158: This could be in STM, I think? Unless it's deliberately not to reduce contention.

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -105,12 +106,12 @@ getFileExistsMapUntracked = do
   FileExistsMapVar v <- getIdeGlobalAction
   return v
 
--- | Modify the global store of file exists.
-modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO ()
+-- | Modify the global store of file exists and return the keys that need to be mark as dirty
+modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO [Key]
 modifyFileExists state changes = do
   FileExistsMapVar var <- getIdeGlobalState state
```
michaelpj, line 112: I think `getIdeGlobalState` could potentially be in STM, in which case this whole thing could be in STM.

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -105,12 +106,12 @@ getFileExistsMapUntracked = do
   FileExistsMapVar v <- getIdeGlobalAction
   return v
 
--- | Modify the global store of file exists.
-modifyFileExists :: IdeState -> [(NormalizedFilePath, FileChangeType)] -> IO ()
+-- | Modify the global store of file exists and return the keys that need to be mark as dirty
```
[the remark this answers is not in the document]
michaelpj: Various other places that need this additional documentation!

ghcide/src/Development/IDE/Core/FileStore.hs
```diff
@@ -244,14 +247,11 @@ typecheckParentsAction recorder nfp = do
 -- | Note that some keys have been modified and restart the session
 --   Only valid if the virtual file system was initialised by LSP, as that
 --   independently tracks which files are modified.
-setSomethingModified :: VFSModified -> IdeState -> [Key] -> String -> IO ()
-setSomethingModified vfs state keys reason = do
+setSomethingModified :: VFSModified -> IdeState -> String -> IO [Key] -> IO ()
+setSomethingModified vfs state reason actionBetweenSession = do
```
michaelpj, line 251: again, it mostly computes dirty keys, right?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1232,9 +1224,10 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
                         _                                     -> False
                 return $ RunResult
                     (if eq then ChangedRecomputeSame else ChangedRecomputeDiff)
-                    (encodeShakeValue bs) $
-                    A res
-        liftIO $ atomicallyNamed "define - dirtyKeys" $ modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
+                    (encodeShakeValue bs)
+                    (A res) $ do
+                        setValues state key file res (Vector.fromList diags)
+                        modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
```
michaelpj, line 1230: Oh hmm, we still have to modify the dirty keys here? How awkward. Why do we need to do that specially here?

michaelpj: Actually, this makes me think I don't understand what's going on. If it's safe to modify the TVar here (presumably because we're doing it in STM so we can put it in a transaction appropriately?), then why can't we do that elsewhere (i.e. mark things dirty in STM as needed).

michaelpj (commented):
Okay, I think I get the idea. Is this right?

- We no longer statefully track dirty keys. It turns out that we always immediately restart the session after we make something dirty, so we can easily just pass in the dirty keys directly.
- In fact, we usually need to do some computation to work out the dirty keys, but we don't want to do this while the rule system is running, otherwise it can get out-of-sync.
- So we pass in a computation to compute the dirty keys.

(I'm not sure this is right, in particular, I don't quite get how the `ruleHook` fits into this, and why we can't use that approach everywhere...

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: yes, I am thinking about removing it too, but our garbage collector depend on it.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: We are putting it into the hook to run after all computation is done and put it to run with marking the key clean in https://github.com/haskell/haskell-language-server/pull/4190#discussion_r1593175057

yep it should need an explicit note to demonstrate the reason

soulomoon: So they get in sync. Pretty nasty.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: So we have to keep it this way for now. Maybe until we find a better way to handle the garbage collector, but this should be in another PR.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: The rule result cache, storing the rule result and its diagnostic.

soulomoon:
@michaelpj  We have two things to do. 
1. The dirty keys insertion, which you have describe.
2. Things to keep sync when we mark the key as clean, the rule cache and the dirty key removal
This is what the ruleHook fits into.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -474,10 +474,9 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
             void $ modifyVar' version succ
```
wz1000, line 478: this should be moved out of the restart.

wz1000 (approved):
Awesome work and investigation! This should make HLS much more robust.

ghcide/src/Development/IDE/Core/Shake.hs
michaelpj: It looks to me like the garbage collector calls `getDirtySet` which gets the dirty keys from `hls-graph`?

ghcide/src/Development/IDE/Core/Shake.hs
michaelpj: I mean, the return value. It's the dirty keys, we just need to document it.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: Yes, but it also mark the key as dirty at `foldM (removeDirtyKey dirtyKeys state) (0,[]) agedKeys` accompany to the rule cache removal.

soulomoon: This is odd since we might have clean it in the rule cahce, 
the a part of the rule cahce would still stays in the hls-graph database.
We might want to improve this in the future.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: done

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
soulomoon: done

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: done

soulomoon:
Yes，this seems to be a potential improvement to do as a follow up to this PR. We can see how it would fit into the monitoring and garbage collect stuff by then

michaelpj:
Okay, I'm happy for you to look into this more after this PR!

michaelpj (approved):

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1258,6 +1256,32 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
         --  * creating bogus "file does not exists" diagnostics
         | otherwise = useWithoutDependency (GetModificationTime_ False) fp
 
+-- Note [Housekeeping rule cache and dirty key out side of hls-graph]
```
jhrcek, line 1259: Nit: fix spelling and all the note references
```suggestion
-- Note [Housekeeping rule cache and dirty key outside of hls-graph]
```

ghcide/src/Development/IDE/Main.hs
```diff
@@ -362,9 +363,10 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
                     Nothing -> pure ()
                     Just ide -> liftIO $ do
                         let msg = T.pack $ show cfg
-                        logWith recorder Debug $ LogConfigurationChange msg
-                        modifyClientSettings ide (const $ Just cfgObj)
-                        setSomethingModified Shake.VFSUnmodified ide [toKey Rules.GetClientSettings emptyFilePath] "config change"
+                        setSomethingModified Shake.VFSUnmodified ide "config change" $ do
+                            logWith recorder Debug $ LogConfigurationChange msg
+                            modifyClientSettings ide (const $ Just cfgObj)
+                            return [toKey Rules.GetClientSettings emptyFilePath]
```
jhrcek, line 369:
```suggestion
                            return [toNoFileKey Rules.GetClientSettings]
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1258,6 +1256,32 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+-- Hls-graph contains its own internal running state for each key in the shakeDatabase.
+-- Rule result cache and dirty key are in ShakeExtras that is not visible to the hls-graph
```
jhrcek, line 1262: It's not clear to me which field in ShakeExtras represents rule result cache.
Could you please mention it here?
```suggestion
-- ShakeExtrac contains ??? field (rule result cache) and dirtyKeys (keys that became dirty in between build sessions) that is not visible to the hls-graph
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1258,6 +1256,32 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+-- Essentially, we need to keep the rule cache and dirty key and hls-graph's internal state
+-- in sync.
+
+-- 1. A dirty key collect in a session should not be clean out in the same session.
```
jhrcek, line 1266: Do I understand it right or what do you mean by clean out?
```suggestion
-- 1. A dirty key collected in a session should not be removed from dirty keys in the same session.
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1258,6 +1256,32 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+-- Since if we clean out the dirty key in the same session,
+--     1.1. we will lose the chance to dirty it's reverse dependencies. Since it only happened during session restart.
```
jhrcek, line 1268:
```suggestion
--     1.1. we will lose the chance to dirty its reverse dependencies. Since it only happens during session restart.
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1258,6 +1256,32 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+--     1.2. a key might marked as dirty at the same time it's already being run to a point that it should not be clean,
+--          then invalidly clean it out.
```
jhrcek, line 1270:
```suggestion
--     1.2. a key might be marked as dirty in ShakeExtras while it's being recomputed by hls-graph which could lead to it's premature removal from dirtyKeys.
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1258,6 +1256,32 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+--          See issue https://github.com/haskell/haskell-language-server/issues/4093 for more details.
+
+-- 2. When a key is marked clean in the hls-graph's internal running
+-- state, the rule cache and dirty key are updated in the same transaction.
```
jhrcek, line 1274:
```suggestion
-- state, the rule cache and dirty keys are updated in the same transaction.
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -612,18 +611,14 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
-
+          void $ modifyVar' fileToFlags $ Map.insert hieYaml this_flags_map
+          void $ modifyVar' filesMap $ flip HM.union (HM.fromList (map ((,hieYaml) . fst) $ concatMap toFlagsMap all_targets))
           -- The VFS doesn't change on cradle edits, re-use the old one.
-          restartShakeSession VFSUnmodified "new component" []
+          -- Invalidate all the existing GhcSession build nodes by restarting the Shake session
+          key2 <- invalidateShakeCache
+          restartShakeSession VFSUnmodified "new component" [] $ do
+            key1 <- extendKnownTargets all_targets
+            return [key1, key2]
```
jhrcek, line 621:
```suggestion
          keys2 <- invalidateShakeCache
          restartShakeSession VFSUnmodified "new component" [] $ do
            keys1 <- extendKnownTargets all_targets
            return [keys1, keys2]
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1232,9 +1227,12 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
                         _                                     -> False
                 return $ RunResult
                     (if eq then ChangedRecomputeSame else ChangedRecomputeDiff)
-                    (encodeShakeValue bs) $
-                    A res
-        liftIO $ atomicallyNamed "define - dirtyKeys" $ modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
+                    (encodeShakeValue bs)
+                    (A res) $ do
+                        -- this hook need to be run in the same transaction as the key is marked clean
```
jhrcek, line 1232:
```suggestion
                        -- this hook needs to be run in the same transaction as the key is marked clean
```

jhrcek (approved):
Few stylistic suggestions, otherwise LGTM

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -759,12 +751,16 @@ delayedAction a = do
...
-shakeRestart recorder IdeState{..} vfs reason acts =
+shakeRestart :: Recorder (WithPriority Log) -> IdeState -> VFSModified -> String -> [DelayedAction ()] -> IO [Key] -> IO ()
+shakeRestart recorder IdeState{..} vfs reason acts ioActionBetweenShakeSession =
     withMVar'
         shakeSession
         (\runner -> do
               (stopTime,()) <- duration $ logErrorAfter 10 $ cancelShakeSession runner
+              keys <- ioActionBetweenShakeSession
+              -- it is every important to update the dirty keys after we enter the critical section
+              -- see Note [Housekeeping rule cache and dirty key out side of hls-graph]
```
jhrcek, line 762: Sorry that I keep nitpicking, but this should be made consistent across the 2 note references
```suggestion
              -- see Note [Housekeeping rule cache and dirty key outside of hls-graph]
```

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1232,9 +1227,12 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
                         _                                     -> False
                 return $ RunResult
                     (if eq then ChangedRecomputeSame else ChangedRecomputeDiff)
-                    (encodeShakeValue bs) $
-                    A res
-        liftIO $ atomicallyNamed "define - dirtyKeys" $ modifyTVar' dirtyKeys (deleteKeySet $ toKey key file)
+                    (encodeShakeValue bs)
+                    (A res) $ do
+                        -- this hook needs to be run in the same transaction as the key is marked clean
+                        -- see Note [Housekeeping rule cache and dirty key out side of hls-graph]
```
jhrcek, line 1233:
```suggestion
                        -- see Note [Housekeeping rule cache and dirty key outside of hls-graph]
```

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: No problem, I am actually greatful that you helped keeping an eyes for these.

Issue #2915, opened 2022-05-20
Rename a constructor renames all record fields


deemp:
### Your environment

Which OS do you use:
Ubuntu 20.04
Which LSP client (editor/plugin) do you use:
VS Code, Haskell extension v2.2.0, ghc 9.2.0, HLS 1.7.0.0
Describe your project (alternative: link to the project):
A .yml parser.  

### Steps to reproduce

See examples in https://github.com/haskell/haskell-language-server/issues/3106#issue-1338974881

### Expected behaviour

Only the constructor name is updated

### Actual behaviour

All fields of a record are renamed the same as the new consructor name

OliverMadine:
Hi, 

Thanks for the report!

Please could you update the link to the commit for the steps to reproduce? It looks like the repository might be private.

@br4ch1st0chr0n3

br4ch1st0chr0n3:
Hi, sorry for that. Should be public now, @OliverMadine.

OliverMadine:
It seems like the issue is hie files including record field names as references to the constructor.
This is only happening with ghc > 9.

@wz1000 maybe you have some more insight?

pepeiborra:
> It seems like the issue is hie files including record field names as references to the constructor.
> This is only happening with ghc >= 9.

Is this a bug or a feature? 

If it is a bug, we will need to report it in the GHC bug tracker and ideally contribute a fix. And back port the fixes to hie-compat: https://github.com/haskell/haskell-language-server/tree/master/hie-compat

michaelpj:
Is this still a problem?

konn:
Yes. I've gotten bitten by this yesterday at least with HLS 2.5.0.0.

konn:
I've just tested and confirmed that it still occurs with HLS 2.6.0.0.

soulomoon:
Perhaps the `type RefMap a = Map Identifier [(Span, IdentifierDetails a)]` should contains the [NodeOrigin](https://hackage.haskell.org/package/ghc-9.8.2/docs/src/GHC.Iface.Ext.Types.html#NodeOrigin) to indicate if it is one of the evidence variable

jhrcek:
Note that I took a stab at fixing this in https://github.com/haskell/haskell-language-server/pull/4089

That fix works, but requires changing on hiedb side (skipping indexing of all Identifiers whose NodeOrigin is GeneratedInfo).
[...]

PR #4202, opened 2024-05-01
Migrate FindDefinitionAndHoverTests and HighlightTests part of #4173 and fix HighlightTests

soulomoon:


ghcide/test/exe/HighlightTests.hs
```diff
@@ -44,8 +43,9 @@ tests = testGroup "highlight"
             , DocumentHighlight (R 6 10 6 13) (Just DocumentHighlightKind_Read)
             , DocumentHighlight (R 7 12 7 15) (Just DocumentHighlightKind_Read)
             ]
-  , knownBrokenForGhcVersions [GHC92, GHC94, GHC96, GHC98] "Ghc9 highlights the constructor and not just this field" $
-        testSessionWait "record" $ do
+  ,
+--   knownBrokenForGhcVersions [GHC92, GHC94, GHC96, GHC98] "Ghc9 highlights the constructor and not just this field" $
```
michaelpj, line 47: delete

ghcide/test/exe/HighlightTests.hs
```diff
@@ -55,7 +55,12 @@ tests = testGroup "highlight"
           ]
         highlights <- getHighlights doc (Position 3 17)
         liftIO $ highlights @?=
-          [ DocumentHighlight (R 3 17 3 23) (Just DocumentHighlightKind_Write)
+          [
+            DocumentHighlight (R 3 11 3 14) (Just DocumentHighlightKind_Write)
+          , DocumentHighlight (R 3 17 3 23) (Just DocumentHighlightKind_Read)
+          , DocumentHighlight (R 3 32 3 38) (Just DocumentHighlightKind_Read)
+          , DocumentHighlight (R 4 4 4 7) (Just DocumentHighlightKind_Read)
+          , DocumentHighlight (R 3 17 3 23) (Just DocumentHighlightKind_Write)
```
michaelpj, line 63: are these outputs correct?

michaelpj (approved):

ghcide/test/exe/HighlightTests.hs
soulomoon: good point should we highlight the constructor if we just hover on its field?

soulomoon: Is it the desire behaviour? If not I think we should kept the broken test and open an issue for it instead!

soulomoon:
This should be the relating issue for the highlight test
https://github.com/haskell/haskell-language-server/issues/2915

ghcide/test/exe/HighlightTests.hs
soulomoon: I don't think this is desire, but a fix is relatively easy. I have impelmented it.

ghcide/test/exe/HighlightTests.hs
soulomoon: done

fendor (approved):
LGTM, thanks for taking great care of the test suite!

michaelpj:
The fix seems plausible, but also I think there is still an underlying issue, right? So it would be nice to not forget about that.

soulomoon:
The root is the introducion of ghc generated nodes,
but filter out them is fine for our highlights.

ghcide/src/Development/IDE/GHC/Compat.hs
```diff
@@ -471,7 +472,8 @@ isQualifiedImport ImportDecl{ideclQualified = NotQualified} = False
 isQualifiedImport ImportDecl{}                              = True
 isQualifiedImport _                                         = False
 
-
+getSourceNodeIds :: HieAST a -> Map.Map Identifier (IdentifierDetails a)
```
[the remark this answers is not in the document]
soulomoon: done

PR #4204, opened 2024-05-02
Add thread id in ghcide tests log

soulomoon:
What is done: 
* hls-test-utils have flip this on at #4174, but ghcide-tests have not. Enable it here too.
 We need to enable it in two places of ghcide, one in the ghcide exe directly, other is in the main entry of the ghcide test
* remove redundant log when initializing file system during test

### DIFF
--- a/ghcide/exe/Main.hs
+++ b/ghcide/exe/Main.hs
@@ -21,7 +21,7 @@ import qualified Development.IDE.Main                     as IDEMain
 import qualified Development.IDE.Monitoring.OpenTelemetry as OpenTelemetry
 import qualified Development.IDE.Plugin.HLS.GhcIde        as GhcIde
 import           Development.IDE.Types.Options
-import           Ide.Logger                               (LoggingColumn (DataColumn, PriorityColumn),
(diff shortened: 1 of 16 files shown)


ghcide/exe/Main.hs
```diff
@@ -73,7 +73,7 @@ main = withTelemetryRecorder $ \telemetryRecorder -> do
     -- stderr recorder just for plugin cli commands
     pluginCliRecorder <-
       cmapWithPrio pretty
-      <$> makeDefaultStderrRecorder (Just [PriorityColumn, DataColumn])
+      <$> makeDefaultStderrRecorder (Just [ThreadIdColumn, PriorityColumn, DataColumn])
```
michaelpj, line 76: Hmm, this seems like it will affect people using the ghcide executable directly? Well, hopefully that will go away in due course anyway...

soulomoon: Yes, I am under the impression ghcide is just using for test purpose now and we are considering removing it?

michaelpj (approved):

PR #4205, opened 2024-05-02
Wait For token using Barrier when indexHieFile

soulomoon:


michaelpj:
Could we refactor this to actually use the helpers from `lsp` for running progress sessions? I think that should handle all of this properly, rather than us re-implementing it here?

wz1000:
This code manages its progress token in a somewhat unique way which is not easy to statically scope with the lsp progress functions.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -942,8 +942,9 @@ indexHieFile se mod_summary srcPath !hash hf = do
             Nothing -> pure Nothing
             Just env -> LSP.runLspT env $ do
               u <- LSP.ProgressToken . LSP.InR . T.pack . show . hashUnique <$> liftIO Unique.newUnique
-              -- TODO: Wait for the progress create response to use the token
-              _ <- LSP.sendRequest LSP.SMethod_WindowWorkDoneProgressCreate (LSP.WorkDoneProgressCreateParams u) (const $ pure ())
+              b <- liftIO newBarrier
+              void $ LSP.sendRequest LSP.SMethod_WindowWorkDoneProgressCreate (LSP.WorkDoneProgressCreateParams u) $ liftIO . signalBarrier b
+              ready <- liftIO $ waitBarrier b
```
wz1000, line 947: I don't know if its a good idea to delay indexing by an unbounded amount of time waiting for the client to respond with a progress token. Perhaps the progress reporting should be on another thread which doesn't block indexing.

soulomoon: yes, but indexHieFile return without reportProgress actually report done🤔.

soulomoon:
I've spawn a new thread to run the indexing progress updated, send through TQueue.

michaelpj:
I think we can probably do something like https://github.com/haskell/haskell-language-server/pull/4218 now

soulomoon:
> I think we can probably do something like #4218 now

we might need to change the update logic ?
IndexHieFile's progress using push update.
#4218 is using a pull update 🤔

michaelpj:
> Also, I am wondering if we can make  Development.IDE.Core.ProgressReporting more general, so we can actually reuse it.

Yes, I think there's almost something there. It's something like a standalone counter that also has some triggers for killing and restarting it?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -911,69 +913,20 @@ indexHieFile se mod_summary srcPath !hash hf = do
...
-          bracket_ (pre optProgressStyle) post $
-            withHieDb (\db -> HieDb.addRefsFromLoaded db targetPath (HieDb.RealFile $ fromNormalizedFilePath srcPath) hash hf')
+          tok <- modifyVar indexProgressToken $ fmap (first Just . dupe) . \case
+              Just x -> return x
+              -- create a progressReport if we don't already have one
+              Nothing -> do
+                  tt <- progressReporting (lspEnv se) "Indexing" optProgressStyle
+                  progressUpdate tt ProgressStarted
+                  return tt
+          inProgress tok srcPath
```
soulomoon, line 923: I bump of progress needed should move to the ouside scope.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -898,6 +899,7 @@ indexHieFile se mod_summary srcPath !hash hf = do
     _ -> do
       -- hiedb doesn't use the Haskell src, so we clear it to avoid unnecessarily keeping it around
       let !hf' = hf{hie_hs_src = mempty}
+      -- todo, this is the real pending count
       modifyTVar' indexPending $ HashMap.insert srcPath hash
```
soulomoon, line 903: We should to a bump of waitting progress here

soulomoon:
The scope quite complex and is hard to fix in.

To solve the scope problem, I refactor the  `Development.IDE.Core.ProgressReporting` to add `InProgressStateOutSide`. We can let the ousider program manage the state, and ProgressReporting just pull in the todo and done count.

soulomoon:
But it seems we lose the finish notification. Any idea? @michaelpj

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -988,16 +940,11 @@ indexHieFile se mod_summary srcPath !hash hf = do
         when (coerce $ ideTesting se) $
           LSP.sendNotification (LSP.SMethod_CustomMethod (Proxy @"ghcide/reference/ready")) $
             toJSON $ fromNormalizedFilePath srcPath
-      whenJust mdone $ \done ->
+      whenJust mdone $ \_ ->
         modifyVar_ indexProgressToken $ \tok -> do
-          whenJust (lspEnv se) $ \env -> LSP.runLspT env $
```
soulomoon, line 993: we lose it.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,156 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
-    | KickCompleted
+  = ProgressStarted
+  | ProgressCompleted
 
-data ProgressReporting  = ProgressReporting
-  { progressUpdate :: ProgressEvent -> IO ()
-  , inProgress     :: forall a. NormalizedFilePath -> Action a -> Action a
-  , progressStop   :: IO ()
+data ProgressReporting m = ProgressReporting
+  { progressUpdate :: ProgressEvent -> IO (),
```
michaelpj, line 45: it's a bit surprising that these are still in `IO` rather than `m`? I think they could be, since you're requiring `MonadUnliftIO` for `m` everywhere?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,156 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
 import qualified Language.LSP.Server            as LSP
 import qualified StmContainers.Map              as STM
-import           UnliftIO                       (Async, async, cancel)
+import           UnliftIO                       (Async, MonadUnliftIO, async,
+                                                 bracket, cancel)
 
 data ProgressEvent
-    = KickStarted
-    | KickCompleted
+  = ProgressStarted
```
michaelpj, line 41: :+1:

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -917,63 +916,16 @@ indexHieFile se mod_summary srcPath !hash hf = do
...
-                  , _percentage = Nothing
-                  }
-
+    pre optProgressStyle = modifyVar_ indexProgressToken $ fmap Just . \case
+              Just x -> return x
+              -- create a progressReport if we don't already have one
+              Nothing -> do
+                  tt <- progressReportingOutsideState
+                    (liftM2 (+) (HashMap.size <$> readTVar indexPending) (readTVar indexCompleted))
+                    (readTVar indexCompleted)
```
michaelpj, line 925: This seems a bit backwards to me? Can we not use the usual `ProgressReporting` and then call `progressUpdate` etc. instead of touching these TVars in the indexing code?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,156 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
+  | ProgressCompleted
```
michaelpj, line 42: Can we just get rid of `progressStop`, given that it just does the same thing as `progressUpdate ProgressCompleted`?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -988,16 +940,11 @@ indexHieFile se mod_summary srcPath !hash hf = do
...
-          whenJust (lspEnv se) $ \env -> LSP.runLspT env $
-            whenJust tok $ \token ->
-              LSP.sendNotification LSP.SMethod_Progress  $ LSP.ProgressParams token $
-                toJSON $
-                LSP.WorkDoneProgressEnd
-                  { _kind = LSP.AString @"end"
-                  , _message = Just $ "Finished indexing " <> T.pack (show done) <> " files"
-                  }
+          case tok of
+            Just token -> progressUpdate token ProgressCompleted >> progressStop token
```
michaelpj, line 946: This in fact does the same thing twice! (Leading to my comment below about getting rid of one of them)

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,156 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
+        todoVar    :: TVar Int,
+        -- | Number of files done
+        doneVar    :: TVar Int,
+        currentVar :: STM.Map NormalizedFilePath Int
+      }
+  | InProgressStateOutSide
+      { -- | Number of files to do
+        todo :: STM Int,
+        -- | Number of files done
+        done :: STM Int
```
michaelpj, line 90: This is kind of neat, we just rely on STM transaction retrying to notice when things change. Worth explaining! 

I wonder if we could just use this as the primary mechanism, and then have the other kind of progress reporting as a little wrapper around it?

ghcide/src/Development/IDE/Core/Compile.hs
michaelpj: Okay, I think I get it. Now I wonder if it would be simpler to make the other progress tracker work this way?

michaelpj (commented):
Looks plausible, I wonder if we can simplify `ProgressReporting` and just have one way to do it.

ghcide/src/Development/IDE/Core/Compile.hs
soulomoon: As along as they are already tracking the progress status external. 

But I take a look at the use case at `kick`, the state(todo count,done count) tracking is at `defineEarlyCutoff'` and independent for each file. No easy way to swtich the `Kick` progress tracker work this way.

michaelpj: Right, but when we _create_ the progress tracker we could set that stuff up externally, so that the `ProgressReporting` doesn't see it. Let me see if I can just write it down.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: `Kick` progress tracking working differently than `indexHIeFile`.

HLS begins: Progress tracking started

loop:
  Kick start ProgressStarted
  Kick end ProgressCompleted

HLS begins: Progress tracking Stop

ghcide/src/Development/IDE/Core/Compile.hs
michaelpj: Here: https://github.com/haskell/haskell-language-server/commit/2ad92a01e89480e06d60647f127fae052054c71e

If we did this then I think `indexHieFile` could just use `progressCounter` directly?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: Yes, I've already turned the `lspShakeProgressNew` of `InProgressState` to `LSP.runLspT lspEnv $ withProgress title Nothing NotCancellable $ \update -> loop update 0 (readTVar todoVar) (readTVar doneVar)`
We are using it as the `primary mechanism` now.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
michaelpj: Oh yes, I missed that the state update is different, `ProgressCompleted` goes back to `NotStarted`. It's not clear that that gets us much, though :thinking:

michaelpj: `NotStarted` and `Stopped` are the same state, except you can't start again from `Stopped`. But why does that even matter? In both cases you stop the session, it's not like `Stopped` cleans up any more resources!

michaelpj: So the way it is structured we get a restartable progress tracker, which is a reasonable thing to have when you keep a single reference to it alive for a long time.. But I think it could be simpler.

ghcide/src/Development/IDE/Core/Compile.hs
soulomoon: Oh I see. I've done that to the `loop`, but exract that to `progressCounter` is cleaner

michaelpj: Yeah, it's a nice standalone thing in fact

soulomoon: I ported 2ad92a0 here

michaelpj: then I think if you export it you can just use it in `indexHieFile`

soulomoon: Great, extract it to standalone now! @michaelpj

soulomoon: If using it barely in `indexHieFile`, we still need some kind of wrapping around `progressCounter`.  `async` and `cancel` the `progressCounter`.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: > So the way it is structured we get a restartable progress tracker

Yes, I think it is intended for multiple `Kick`'s.

soulomoon: or we need to initialize progress tracker at each kick.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: right, we let's swtich it to m

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: I guess, `Stopped ` state is used to forbit us to start the progress tracker again after the we shutdown hls

ghcide/src/Development/IDE/Core/Compile.hs
michaelpj: Yeah, we do need that. But we could just do it locally. Or maybe you want to have a standalone concept of a "resumable/cancellable progress counter" and then just put the file-tracking on top for the compilation progress tracker?

soulomoon: May be the latter, I am moving up the `indexHieFile` reporting initilization to `shakeOpen` as that of  the `Kick` reporting. Should be much cleaner than the preivous version.

soulomoon: "resumable/cancellable progress counter" may need the additional Event `ProgressTryToStart`.

soulomoon:
Now it becomes `resumable/restartable/cancellable progress reporting`

soulomoon:
After introducing new event capturing more progress reporting logic, I was able to change the `indexProgressToken :: Var (Maybe LSP.ProgressToken)` to `indexProgressReporting :: ProgressReporting IO`,

ghcide/src/Development/IDE/Core/ProgressReporting.hs
michaelpj: That doesn't seem actually useful, though. I'd be inclined to remove it.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,170 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
+  | ProgressStarted
 
-data ProgressReporting  = ProgressReporting
-  { progressUpdate :: ProgressEvent -> IO ()
-  , inProgress     :: forall a. NormalizedFilePath -> Action a -> Action a
-  , progressStop   :: IO ()
+data ProgressReporting m = ProgressReporting
+  { progressUpdate :: ProgressEvent -> m (),
+    inProgress     :: forall a. NormalizedFilePath -> m a -> m a,
+    progressStop   :: IO ()
```
michaelpj, line 48: why is this not in m?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,170 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
-    }
+data InProgressState
+  = InProgressState
+      { -- | Number of files to do
+        todoVar    :: TVar Int,
+        -- | Number of files done
+        doneVar    :: TVar Int,
+        currentVar :: STM.Map NormalizedFilePath Int
+      }
+  | InProgressStateOutSide
```
michaelpj, line 89: I still find this split a bit weird, but we've spent enough time on this!

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -911,69 +909,14 @@ indexHieFile se mod_summary srcPath !hash hf = do
...
-                    , _percentage = Nothing
-                    }
-                NoProgress -> LSP.WorkDoneProgressReport
-                  { _kind = LSP.AString @"report"
-                  , _cancellable = Nothing
-                  , _message = Nothing
-                  , _percentage = Nothing
-                  }
-
+    pre = progressUpdate indexProgressReporting ProgressStarted
```
michaelpj, line 919: I had actually thought we could just initialize a `progressCounter` here and use it just in the scope of this function, but this works too!

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -33,119 +34,170 @@ import           Language.LSP.Server            (ProgressAmount (..),
...
-mRunLspT :: Applicative m => Maybe (LSP.LanguageContextEnv c ) -> LSP.LspT c m () -> m ()
+progressReporting ::
+  (MonadUnliftIO m, MonadIO m) =>
+  Maybe (LSP.LanguageContextEnv c) ->
+  T.Text ->
+  ProgressReportingStyle ->
+  IO (ProgressReporting m)
+progressReporting = progressReporting' newInProgress
+
+progressReportingOutsideState ::
```
michaelpj, line 128: At the moment it's pretty unclear when you would use one of these or the other. Can we write down why in the haddock? I think we maybe also learned some other things that we should write down, like the stuff about whether `Stopped` means anything.

michaelpj (approved):
Looks good, maybe just write down some of what we've learned!

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: `shakeShut :: IdeState -> IO ()` use `progressStop`.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: yes, especially we have `inProgress` that is useless `InProgressStateOutSide`.
Maybe in some following ups, we can come up with alternative design such as using gadt to eliminate the oddness

michaelpj:
Needs conflicts fixed then good to go

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: done

PR #4206, opened 2024-05-03
Refactor delayedProgressReporting lsp progress api

soulomoon:
It seems we have not been using lsp progress helper api any where.
Refactor `delayedProgressReporting` to use lsp progress api. 
It is an easy target, we can see see how it works out here.
cc @michaelpj

### DIFF
--- a/ghcide/src/Development/IDE/Core/ProgressReporting.hs
+++ b/ghcide/src/Development/IDE/Core/ProgressReporting.hs
@@ -39,8 +39,10 @@ import           Language.LSP.Server            (ProgressAmount (..),
 import qualified Language.LSP.Server            as LSP
 import qualified StmContainers.Map              as STM
 import           System.Time.Extra
+import qualified UnliftIO                       as MonadUnliftIO
 import           UnliftIO                       (MonadUnliftIO (..),
-                                                 UnliftIO (unliftIO), toIO)
+                                                 UnliftIO (unliftIO), newMVar,
+                                                 putMVar, toIO)
 import           UnliftIO.Exception             (bracket_)
 
 data ProgressEvent
@@ -64,18 +66,18 @@ noProgressReporting = return $ ProgressReporting
 data State
     = NotStarted
     | Stopped
-    | Running (Async ())
+    | Running (IO (IO ()))
 
 -- | State transitions used in 'delayedProgressReporting'
 data Transition = Event ProgressEvent | StopProgress
 
-updateState :: IO (Async ()) -> Transition -> State -> IO State
+updateState :: IO (IO ()) -> Transition -> State -> IO State
 updateState _      _                    Stopped     = pure Stopped
-updateState start (Event KickStarted)   NotStarted  = Running <$> start
-updateState start (Event KickStarted)   (Running a) = cancel a >> Running <$> start
-updateState _     (Event KickCompleted) (Running a) = cancel a $> NotStarted
+updateState start (Event KickStarted)   NotStarted  = pure $ Running start
+updateState start (Event KickStarted)   (Running a) = join a $> Running start
+updateState _     (Event KickCompleted) (Running a) = join a $> NotStarted
 updateState _     (Event KickCompleted) st          = pure st
-updateState _     StopProgress          (Running a) = cancel a $> Stopped
+updateState _     StopProgress          (Running a) = join a $> Stopped
 updateState _     StopProgress          st          = pure st
 
 -- | Data structure to track progress across the project
@@ -107,6 +109,12 @@ recordProgress InProgressState{..} file shift = do
         return (prev, new)
     alter x = let x' = maybe (shift 0) shift x in Just x'
 
+-- | Runs the action until it ends or until the given MVar is put.
+--   Rethrows any exceptions.
+untilMVar :: MonadUnliftIO m => MVar () -> m () -> m ()
+untilMVar mvar io = void $
+    MonadUnliftIO.waitAnyCancel =<< traverse MonadUnliftIO.async [ io , MonadUnliftIO.readMVar mvar ]
+
 -- | A 'ProgressReporting' that enqueues Begin and End notifications in a new
 --   thread, with a grace period (nothing will be sent if 'KickCompleted' arrives
 --   before the end of the grace period).
@@ -130,8 +138,18 @@ delayedProgressReporting before after (Just lspEnv) optProgressStyle = do
             -- first sleep a bit, so we only show progress messages if it's going to take
             -- a "noticable amount of time" (we often expect a thread kill to arrive before the sleep finishes)
             liftIO $ sleep before
-            async $ LSP.runLspT lspEnv $ withProgress "Processing" Nothing NotCancellable $ \update -> loop update 0
+            cancelProgress <- Control.Concurrent.Strict.newEmptyMVar
+            LSP.runLspT lspEnv $ do
+                u <- ProgressToken . InR . T.pack . show . hashUnique <$> liftIO newUnique
+                b <- liftIO newBarrier
+                LSP.sendRequest SMethod_WindowWorkDoneProgressCreate
+                    LSP.WorkDoneProgressCreateParams { _token = u } $ liftIO . signalBarrier b
+                return $ async $ do
+                    ready <- waitBarrier b
+                    LSP.runLspT lspEnv $ withProgress "Processing" (Just u) Cancellable $ \update -> loopUntil cancelProgress update 0
+            return (Control.Concurrent.Strict.putMVar cancelProgress ())
             where
+                loopUntil m a b = untilMVar m $ loop a b
                 loop _ _ | optProgressStyle == NoProgress =
                     forever $ liftIO $ threadDelay maxBound
                 loop update prevPct = do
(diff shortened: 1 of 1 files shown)


ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -102,6 +111,12 @@ recordProgress InProgressState{..} file shift = do
         return (prev, new)
     alter x = let x' = maybe (shift 0) shift x in Just x'
 
+-- | Runs the action until it ends or until the given MVar is put.
+--   Rethrows any exceptions.
+untilMVar :: MonadUnliftIO m => MVar () -> m () -> m ()
+untilMVar mvar io = void $
+    waitAnyCancel =<< traverse async [ io , readMVar mvar ]
```
jhrcek, line 118: I'm not using async stuff very often, so take this suggestion with a grain of salt.
But I think you can write this using the more higher-level function [race_](https://hackage.haskell.org/package/async-2.2.4/docs/Control-Concurrent-Async.html#v:race_) (resp. its unliftio counterpart)

soulomoon: That is simpler, but [race_](https://hackage.haskell.org/package/async-2.2.4/docs/Control-Concurrent-Async.html#v:race_) would hide the exception?🤔

jhrcek: What exception? And do you or do you not want that exception to be hidden?

soulomoon: I do not want exception to be hidden if any.

jhrcek: In that case race_ should be just fine, because it [rethrows](https://hackage.haskell.org/package/async-2.2.4/docs/src/Control.Concurrent.Async.html#race) the exception if any of the 2 actions throws any.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -117,76 +132,42 @@ delayedProgressReporting before after (Just lspEnv) optProgressStyle = do
...
+                    if todo == 0 then loop update 0 else do
+                        let nextFrac :: Double
                             nextFrac = fromIntegral done / fromIntegral todo
                             nextPct :: UInt
                             nextPct = floor $ 100 * nextFrac
                         when (nextPct /= prevPct) $
-                          LSP.sendNotification SMethod_Progress $
-                          LSP.ProgressParams
-                              { _token = token
-                              , _value = case optProgressStyle of
```
michaelpj, line 173: We've lost this progress-style configurability, but honestly, configurability is bad, so better to get rid of it IMO!

michaelpj: We should get rid of the option as well, I guess.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -117,76 +132,42 @@ delayedProgressReporting before after (Just lspEnv) optProgressStyle = do
...
-
+        updateStateVar tran = do
+            start <- lspShakeProgressNew inProgressState
+            modifyVar_ progressState $ updateState start tran
         inProgress = updateStateForFile inProgressState
     return ProgressReporting{..}
     where
-        lspShakeProgress InProgressState{..} = do
+        lspShakeProgressNew InProgressState{..} = do
             -- first sleep a bit, so we only show progress messages if it's going to take
```
michaelpj, line 142: Ah, I would actually really like to do this in `lsp` as well, see https://github.com/haskell/lsp/issues/549

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -117,76 +132,42 @@ delayedProgressReporting before after (Just lspEnv) optProgressStyle = do
...
-                LSP.runLspT lspEnv $ for_ ready $ const $ bracket_ (start u) (stop u) (loop u 0)
+            cancelProgress <- newEmptyMVar
+            LSP.runLspT lspEnv $ do
+                u <- ProgressToken . InR . T.pack . show . hashUnique <$> liftIO newUnique
+                b <- liftIO newBarrier
+                LSP.sendRequest SMethod_WindowWorkDoneProgressCreate
+                    LSP.WorkDoneProgressCreateParams { _token = u } $ liftIO . signalBarrier b
+                async $ do
+                    ready <- liftIO $ waitBarrier b
+                    withProgress "Processing" (Just u) Cancellable $ \update -> loopUntil cancelProgress update 0
```
michaelpj, line 153: This isn't quite right, you should skip sending the `WorkDoneProgressCreate` message and pass `Nothing` here. Then it creates the token for you automatically. Passing `Just token` is for the case where you can get a progress token in the request, which isn't what's happening here.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -117,76 +132,42 @@ delayedProgressReporting before after (Just lspEnv) optProgressStyle = do
...
             liftIO $ sleep before
-            u <- ProgressToken . InR . T.pack . show . hashUnique <$> liftIO newUnique
-
-            b <- liftIO newBarrier
-            void $ LSP.runLspT lspEnv $ LSP.sendRequest SMethod_WindowWorkDoneProgressCreate
-                LSP.WorkDoneProgressCreateParams { _token = u } $ liftIO . signalBarrier b
-            liftIO $ async $ do
-                ready <- waitBarrier b
-                LSP.runLspT lspEnv $ for_ ready $ const $ bracket_ (start u) (stop u) (loop u 0)
+            cancelProgress <- newEmptyMVar
```
michaelpj, line 145: Make this a `Barrier`? Then indeed I think you can just do ```waitBarrier cancelProgress `race` loop update 0``` below.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
michaelpj: Then you can skip all the other stuff and just go straight to the `withProgress` call.

Also, while we're here, can we change the title to something other than the incredibly generic "Processing"? I've always found that quite annoying. We might need to pass in an informative string from the callers, though...

michaelpj (commented):
Looks plausible, we can get rid of even more. And I would really like to do that generic delay ticket in `lsp` so we can get rid of even more !

soulomoon:
good suggestion. Swtiching to `race` and `barrier`

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: done

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -59,18 +68,18 @@ noProgressReporting = return $ ProgressReporting
...
 data Transition = Event ProgressEvent | StopProgress
 
-updateState :: IO (Async ()) -> Transition -> State -> IO State
+updateState :: Barrier () -> Transition -> State -> IO State
 updateState _      _                    Stopped     = pure Stopped
-updateState start (Event KickStarted)   NotStarted  = Running <$> start
-updateState start (Event KickStarted)   (Running a) = cancel a >> Running <$> start
-updateState _     (Event KickCompleted) (Running a) = cancel a $> NotStarted
+updateState start (Event KickStarted)   NotStarted  = pure $ Running start
+updateState start (Event KickStarted)   (Running a) = signalBarrier a () $> Running start
```
michaelpj, line 79: Hmmm. Does this work? We signal the barrier and then go back into the running state, but won't the barrier have cancelled the whole progress reporting??

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -117,76 +127,35 @@ delayedProgressReporting before after (Just lspEnv) optProgressStyle = do
...
-            b <- liftIO newBarrier
-            void $ LSP.runLspT lspEnv $ LSP.sendRequest SMethod_WindowWorkDoneProgressCreate
-                LSP.WorkDoneProgressCreateParams { _token = u } $ liftIO . signalBarrier b
-            liftIO $ async $ do
-                ready <- waitBarrier b
-                LSP.runLspT lspEnv $ for_ ready $ const $ bracket_ (start u) (stop u) (loop u 0)
+            cancelProgress <- newBarrier
+            LSP.runLspT lspEnv $ withProgress "Processing" Nothing Cancellable $ \update ->
+                race (liftIO $ waitBarrier cancelProgress) (loop update 0)
+            return cancelProgress
```
michaelpj, line 143: This can't work: `withProgress` runs until the inner operation terminates! So we won't ever get here!

michaelpj: (Sorry, I don't have a suggestion for how to do it right, I need to read it carefully again...)

soulomoon: Right, I forget to async it. I'll add it and see how would this go

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: we cancel the previous job and leave the new job cancellation here.

soulomoon:
Closing, now Micheal picks this up in his PR https://github.com/haskell/haskell-language-server/pull/4218

PR #4207, opened 2024-05-04
[Migrate diagnosticTests] part of #4173 Migrate ghcide tests to hls test utils 

soulomoon:
* [x] migrate diagnosticTests, figure out how to pass `--test-no-kick`. 
* [x] migrate openCloseTests
* [x] Lift a few functions from ghcide-test-utils to hls-test-utils
* [x] fixed `deeply nested cyclic module dependency`
* [x] modify `runSessionWithServer'` and ghcIde arguments to admit additional a switch for `no-kick`.


ghcide/test/exe/Config.hs
```diff
@@ -31,9 +34,15 @@ runWithDummyPlugin' = runSessionWithServerInTmpDirCont' def dummyPlugin
 runWithDummyPluginAndCap :: ClientCapabilities -> Session () -> IO ()
 runWithDummyPluginAndCap cap = runSessionWithServerAndCapsInTmpDir def dummyPlugin cap (mkIdeTestFs [])
 
+runWithDummyPluginAndCap' :: ClientCapabilities -> (FileSystem -> Session ()) -> IO ()
```
michaelpj, line 37: FWIW, I'm not sure all these functions are necessarily worth it. How bad would it be to just have the `FileSystem -> Session ()` variants and then write `_ -> whatever` at the use sites?

soulomoon:
It seems `pass --test-no-kick` might require a lot of code change in setup in hls-graph, I don't know if there is alternative to this.

ghcide/test/exe/Config.hs
soulomoon: Emm, we are at somewhat funny situation, that the one without the extra FileSystem is the most used one. 
Sure, for not frequent used ones, it would be fine we write `_ -> whatever`.

michaelpj: I guess I'm wondering how bad it would be. Otherwise we have a lot of very similar variant functions :thinking:

michaelpj: Not a blocker, anyway

michaelpj:
What's the significance of `test-no-kick`?

soulomoon:
> What's the significance of test-no-kick?

My guess is that so the test can trigger the typecheck manually without a lot of rules being triggered by kick and hence capture the diagnostic in a more stable way

ghcide/test/exe/Config.hs
soulomoon: Yep, I mean since we have a lot of test that does not need the `FileSystem `,
adding a tons of `_ -> whatever` is visually bad and and not convenient for us to do the migration.

That is why I want to either do the typeclass polymorphic or just duplicate the test runner function variant.

soulomoon:
Two new thing have been done, I modify runSessionWithServer' and ghcIde arguments to admit additional a switch for no-kick. And also some setup rearrangement.
Additionally, I put more test environment variables into pluginTestRecorder so we can use pluginTestRecorder to generate test recorder generically, any reason not to ?

ghcide/src/Development/IDE/Main.hs
```diff
@@ -293,7 +295,10 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
         plugins = hlsPlugin <> argsGhcidePlugin
         options = argsLspOptions { LSP.optExecuteCommandCommands = LSP.optExecuteCommandCommands argsLspOptions <> Just hlsCommands }
         argsParseConfig = getConfigFromNotification argsHlsPlugins
-        rules = argsRules >> pluginRules plugins
+        rules = argsRules >> (unless argsDisableKick $ action kick)  >> pluginRules plugins
```
fendor, line 298:
```suggestion
        rules = do
            argsRules
            unless argsDisableKick $ action kick
            pluginRules plugins
```

No reason to code golf :upside_down_face:

ghcide/test/exe/DiagnosticTests.hs
```diff
@@ -238,11 +240,9 @@ tests = testGroup "diagnostics"
       _ <- createDoc "ModuleB.hs" "haskell" contentB
       _ <- createDoc "ModuleB.hs-boot" "haskell" contentBboot
       expectDiagnostics [("ModuleB.hs", [(DiagnosticSeverity_Warning, (3,0), "Top-level binding")])]
-  , testSession' "bidirectional module dependency with hs-boot" $ \path -> do
-      let cradle = unlines
-            [ "cradle:"
-            , "  direct: {arguments: [ModuleA, ModuleB]}"
-            ]
+  , testWithDummyPlugin "bidirectional module dependency with hs-boot"
+        (mkIdeTestFs [file "hie.yaml" $ text $ T.unlines ["cradle:", "  direct: {arguments: [ModuleA, ModuleB]}"]])
```
fendor, line 244:
```suggestion
        (mkIdeTestFs [directCradle ["ModuleA", "ModuleB"]])
```

ghcide/test/exe/DiagnosticTests.hs
```diff
@@ -426,9 +425,9 @@ tests = testGroup "diagnostics"
             Lens.filtered (T.isInfixOf ("/" <> name <> ".hs:"))
           failure msg = liftIO $ assertFailure $ "Expected file path to be stripped but got " <> T.unpack msg
       Lens.mapMOf_ offenders failure notification
-  , testSession' "-Werror in cradle is ignored" $ \sessionDir -> do
-      liftIO $ writeFile (sessionDir </> "hie.yaml")
-        "cradle: {direct: {arguments: [\"-Wall\", \"-Werror\"]}}"
+  , testWithDummyPlugin "-Werror in cradle is ignored"
+        (mkIdeTestFs [file "hie.yaml" $ text "cradle: {direct: {arguments: [\"-Wall\", \"-Werror\"]}}" ])
```
fendor, line 429:
```suggestion
        (mkIdeTestFs [directCradle ["-Wall", "-Werror"]])
```

hls-test-utils/src/Test/Hls.hs
```diff
@@ -323,8 +326,14 @@ mkPluginTestDescriptor'
 mkPluginTestDescriptor' pluginDesc plId _recorder = IdePlugins [pluginDesc plId]
 
 -- | Initialise a recorder that can be instructed to write to stderr by
--- setting the environment variable "HLS_TEST_PLUGIN_LOG_STDERR=1" before
--- running the tests.
+-- setting the environment variable "HLS_TEST_PLUGIN_LOG_STDERR|LSP_TEST_LOG_STDERR
+-- |LSP_TEST_LOG_STDERR|HLS_TEST_LOG_STDERR=1" before running the tests.
```
fendor, line 330: I think haddock will get confused by the `-- |`. 

Perhaps more lines make it better?

```suggestion
-- setting one of the environment variables:
--
-- * HLS_TEST_PLUGIN_LOG_STDERR=1
-- * HLS_TEST_LOG_STDERR=1
--
-- before running the tests.
```

I would omit 'LSP_TEST_LOG_STDERR', so that no one thinks about using it any more.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -323,8 +326,14 @@ mkPluginTestDescriptor'
...
+--
+-- Because "LSP_TEST_LOG_STDERR" has been used before,
+-- (thus, backwards compatibility) and "HLS_TEST_HARNESS_STDERR" because it
+-- uses a more descriptive name to be used with "HLS_TEST_HARNESS_NO_TESTDIR_CLEANUP".
```
fendor, line 334: This is not the intention, a plugin test recorder should not log on `HLS_TEST_HARNESS_STDERR`.

`HLS_TEST_HARNESS_STDERR` should log information about our test helpers, such as `runSessionWithServerInTmpDirCont`, it shouldn't enable plugin logging. Same goes for `HLS_TEST_HARNESS_NO_TESTDIR_CLEANUP`, that environment variable shouldn't be used for logging at all, it doesn't remove a `VirtualFileTree` after the test is run.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -339,7 +348,8 @@ mkPluginTestDescriptor' pluginDesc plId _recorder = IdePlugins [pluginDesc plId]
 -- @
 pluginTestRecorder :: Pretty a => IO (Recorder (WithPriority a))
 pluginTestRecorder = do
-  initialiseTestRecorder ["HLS_TEST_PLUGIN_LOG_STDERR", "HLS_TEST_LOG_STDERR"]
+  initialiseTestRecorder ["HLS_TEST_PLUGIN_LOG_STDERR",
+        "LSP_TEST_LOG_STDERR", "HLS_TEST_HARNESS_STDERR", "HLS_TEST_LOG_STDERR"]
```
fendor, line 352: I don't think all these environment variables should be used to enable logging for plugins. I don't think this should be changed.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -374,28 +384,27 @@ runSessionWithServerAndCapsInTmpDir config plugin caps tree act = runSessionWith
...
+    runSessionWithServerInTmpDirCont False plugin config def caps tree act
 
 runSessionWithServerInTmpDir' ::
+    Pretty b =>
     -- | Plugins to load on the server.
     --
     -- For improved logging, make sure these plugins have been initalised with
     -- the recorder produced by @pluginTestRecorder@.
-    IdePlugins IdeState ->
+    PluginTestDescriptor b ->
```
fendor, line 399: Comment is out of date.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -419,22 +428,24 @@ runSessionWithServerInTmpDir' plugins conf sessConf caps tree act = runSessionWi
...
 runSessionWithServerInTmpDirCont ::
+    Pretty b =>
+    -- | whether we disable the kick action or not
+    Bool ->
     -- | Plugins to load on the server.
     --
     -- For improved logging, make sure these plugins have been initalised with
     -- the recorder produced by @pluginTestRecorder@.
-    IdePlugins IdeState ->
+    PluginTestDescriptor b ->
```
fendor, line 438: Comment is out of date.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -605,11 +614,14 @@ lockForTempDirs = unsafePerformIO newLock
...
 runSessionWithServer' ::
+  (Pretty b) =>
+  -- | whether we disable the kick action or not
+  Bool ->
   -- | Plugins to load on the server.
   --
   -- For improved logging, make sure these plugins have been initalised with
   -- the recorder produced by @pluginTestRecorder@.
-  IdePlugins IdeState ->
+  PluginTestDescriptor b ->
```
fendor, line 624: Wrong comment

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -48,24 +47,21 @@ import           Text.Regex.TDFA                          ((=~))
...
-      , Refactor.bindingsPluginDescriptor exactprintLog "ghcide-code-actions-bindings"
-      , Refactor.fillHolePluginDescriptor exactprintLog "ghcide-code-actions-fill-holes"
-      , Refactor.extendImportPluginDescriptor exactprintLog "ghcide-completions-1"
-      ] ++ GhcIde.descriptors ghcideLog
+  return $ mkPluginTestDescriptor Refactor.iePluginDescriptor "ghcide-code-actions-imports-exports"
+      <> mkPluginTestDescriptor Refactor.typeSigsPluginDescriptor "ghcide-code-actions-type-signatures"
+      <> mkPluginTestDescriptor Refactor.bindingsPluginDescriptor "ghcide-code-actions-bindings"
+      <> mkPluginTestDescriptor Refactor.fillHolePluginDescriptor "ghcide-code-actions-fill-holes"
+      <> mkPluginTestDescriptor Refactor.extendImportPluginDescriptor "ghcide-completions-1"
+
```
fendor, line 64:
```suggestion
  mkPluginTestDescriptor Refactor.iePluginDescriptor "ghcide-code-actions-imports-exports"
      <> mkPluginTestDescriptor Refactor.typeSigsPluginDescriptor "ghcide-code-actions-type-signatures"
      <> mkPluginTestDescriptor Refactor.bindingsPluginDescriptor "ghcide-code-actions-bindings"
      <> mkPluginTestDescriptor Refactor.fillHolePluginDescriptor "ghcide-code-actions-fill-holes"
      <> mkPluginTestDescriptor Refactor.extendImportPluginDescriptor "ghcide-completions-1"

```

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -48,24 +47,21 @@ import           Text.Regex.TDFA                          ((=~))
...
+import qualified Development.IDE.GHC.ExactPrint
 import qualified Development.IDE.Plugin.CodeAction        as Refactor
-import qualified Development.IDE.Plugin.HLS.GhcIde        as GhcIde
 import qualified Test.AddArgument
 
 main :: IO ()
 main = defaultTestRunner tests
 
-refactorPlugin :: IO (IdePlugins IdeState)
+refactorPlugin :: IO (PluginTestDescriptor Development.IDE.GHC.ExactPrint.Log)
```
fendor, line 57:
```suggestion
refactorPlugin :: PluginTestDescriptor Development.IDE.GHC.ExactPrint.Log
```

If no IO is necessary any more, let's get rid of it :)

fendor (changes requested):
Changes look generally good. Some nitpicks, and the environment variables need to be discussed!

hls-test-utils/src/Test/Hls.hs
soulomoon: done

hls-test-utils/src/Test/Hls.hs
soulomoon: so `HLS_TEST_LOG_STDERR` enable both the helper and plugin, while `HLS_TEST_HARNESS_STDERR ` and `HLS_TEST_PLUGIN_LOG_STDERR ` enable either.

fendor: Yes, that's exactly what I meant, thanks!

ghcide/test/exe/DiagnosticTests.hs
```diff
@@ -198,28 +204,24 @@ tests = testGroup "diagnostics"
...
-            "cradle: {direct: {arguments: [ModuleA, ModuleB, ModuleC, ModuleD]}}"
-      liftIO $ writeFile (path </> "ModuleA.hs") contentA
-      liftIO $ writeFile (path </> "ModuleB.hs") contentB
-      liftIO $ writeFile (path </> "ModuleC.hs") contentC
-      liftIO $ writeFile (path </> "hie.yaml") cradle
+  , let contentA = T.unlines [ "module ModuleA where" , "import ModuleB" ]
+        contentB = T.unlines [ "module ModuleB where" , "import ModuleA" ]
+        contentC = T.unlines [ "module ModuleC where" , "import ModuleB" ]
+        contentD = T.unlines [ "module ModuleD where" , "import ModuleC" ]
+        cradle = "cradle: {direct: {arguments: [ModuleA, ModuleB, ModuleC, ModuleD]}}"
```
fendor, line 211: I overlooked this last time:

```suggestion
        cradle = directCradle ["ModuleA", "ModuleB", "ModuleC", "ModuleD"]
```

ghcide/test/exe/DiagnosticTests.hs
```diff
@@ -198,28 +204,24 @@ tests = testGroup "diagnostics"
...
+    in testWithDummyPlugin "deeply nested cyclic module dependency"
+        (mkIdeTestFs [
+            file "ModuleA.hs" (text contentA)
+            ,file "ModuleB.hs" (text contentB)
+            ,file "ModuleC.hs" (text contentC)
+            ,file "hie.yaml" (text cradle)
```
fendor, line 217:
```suggestion
            ,cradle
```

hls-test-utils/src/Test/Hls.hs
```diff
@@ -322,9 +325,27 @@ mkPluginTestDescriptor'
   -> PluginTestDescriptor b
 mkPluginTestDescriptor' pluginDesc plId _recorder = IdePlugins [pluginDesc plId]
 
--- | Initialise a recorder that can be instructed to write to stderr by
--- setting the environment variable "HLS_TEST_PLUGIN_LOG_STDERR=1" before
--- running the tests.
+-- | Initialize a recorder that can be instructed to write to stderr by
+-- setting one of the environment variables:
+-- "HLS_TEST_HARNESS_STDERR", "HLS_TEST_LOG_STDERR"
```
fendor, line 330:
```suggestion
-- "HLS_TEST_HARNESS_STDERR=1", "HLS_TEST_LOG_STDERR=1"
```

Since `HLS_TEST_HARNESS_STDERR=0` will actually not work.

Perhaps it would be even nicer if the comment looked like:
```suggestion
-- * HLS_TEST_HARNESS_STDERR=1
-- * HLS_TEST_LOG_STDERR=1
```

hls-test-utils/src/Test/Hls.hs
```diff
@@ -419,22 +435,21 @@ runSessionWithServerInTmpDir' plugins conf sessConf caps tree act = runSessionWi
 --
 -- Note: cwd will be shifted into a temporary directory in @Session a@
 runSessionWithServerInTmpDirCont ::
+    Pretty b =>
+    -- | whether we disable the kick action or not
+    Bool ->
```
fendor, line 440: This looks a little bit ad-hoc, perhaps we might want a full `TestConfig` record in the future?
Just a thought, let's leave it as is, it is easy to refactor later.

fendor (approved):
LGTM, thank you very much! 

Only one comment nitpick and a thought :)

hls-test-utils/src/Test/Hls.hs
soulomoon: Yes, I am thinking about it too. perhaps tucking in things like `Config, SessionConfig, ClientCapabilities` into the one `TestConfig` too.

ghcide/test/exe/DiagnosticTests.hs
soulomoon: done

ghcide/test/exe/DiagnosticTests.hs
soulomoon: done

Issue #3577, opened 2023-05-06
haskell-language-server is chatty about progress


mitchellwrosen:
## Is your enhancement request related to a problem? Please describe.

haskell-language-server seems rather chatty with its [progress updates](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress). In particular, after the initial long load, while I'm actively editing a buffer, the server sends a huge number of very short-lived progress updates (where the "end" notification occurs a fraction of a second after the "begin" notification) that just say the word "Processing".

## Describe the solution you'd like

I'm not exactly sure - on one hand, I should think that a language-server would always be in the background "processing" things, so perhaps these notifications simply shouldn't be emitted at all. Alternatively, they could be made more informative.

Thanks!

michaelpj:
Generally our progress notifications are pretty bad, I'm not sure what to do about that.

I suspect what's going on here is that we're sending progress for every module we're loading or similar. That's not a _crazy_ thing to do, but it's perhaps unnecessary, and it would be good if the notification actually said something useful, like what the module was.

Anrock:
Personally I like current progress reports. I don't know if sending lots of short-lived notifications have any detrimental effect but seeing `140/180 modules` message is pretty neat in my neovim. It also gives you a hint that I should split my modules further when after changing a single definition it reprocesses 40 other modules :sweat_smile:

mitchellwrosen:
@Anrock Please note this issue is not about progress notifications in general. I also like seeing messages like "X/Y modules", especially when Y is large or getting to Y/Y takes more than a couple seconds.

However, as mentioned in the issue description there are notifications that literally just say the word "Processing" and no more.

I think whether or not there are detrimental effects of sending many notifications is mostly a function of client settings. I recently toyed around with a fancier graphical notification plugin, which is what drew my attention to hls; I'm sure you can imagine what getting a hundred toasts a minute while editing text would be like. But if you just have notifications drawn on a single line somewhere, and each notification just overwrites whatever is there, then I'm not sure it matters much either way ;)

July541:
You can only see `Processing` from hls is here, should the text be optimized?

https://github.com/haskell/haskell-language-server/blob/41cd52e603ab4e1c7d8fcbf95140610c2a1fe04e/ghcide/src/Development/IDE/Core/ProgressReporting.hs#L137-L146

mortezadadgar:
any progress on this?

michaelpj:
[...]

Issue #381, opened 2020-09-07
The formatting notification is too distracting 


banacorn:
I think there's really nothing worth notifying when formatting the document, especially when the formatting command has to be triggered by user by default. It becomes even more distracting when you enable, say, `format on save`.

<img width="584" alt="截圖 2020-09-07 下午2 22 29" src="https://user-images.githubusercontent.com/797844/92354890-bdd9db80-f115-11ea-9b4b-f05707e5f2ab.png">

There should be an option in the settings for disabling these notifications.

bubba:
I noticed this as well, formatting doesn't usually seem to be that time intensive of an operation so it would make sense to debounce it first.

pepeiborra:
Formatting can block for other reasons though, like consulting a cradle. Debouncing is always good, even better if done on the LSP client side since it is a crosscutting concern across all progress notifications.

georgefst:
Worth noting, currently the progress notification only shows for Ormolu/Fourmolu, presumably because the others aren't calling [withIndefiniteProgress](https://hackage.haskell.org/package/haskell-lsp-0.22.0.0/docs/Language-Haskell-LSP-Core.html#t:LspFuncs).

The notification doesn't annoy me *per se*, but what's odd is that it hangs around for almost a second after formatting appears to have finished, even on a trivial file. This could be because we call Ormolu with the idempotency check, which means that everything is actually formatted twice, but it's longer than I'd expect even accounting for that. Anyway, having a threshold of even a second would make this a non-issue for me.

jneira:
I am gonna use `withIndefiniteProgress` in the hlint plugin, so this would affect it

Profpatsch:
I get the problem that with `format-on-save`, formatting will block saving the file when hls is working on something in the background (especially when it crashes/hangs, which happens if the file has syntax errors on hls start for example).

Requiring me to manually stop hls and interrupt the formatting before I can save the file.

michaelpj:
Should be fixed now, we will only get a progress notification if it takes some time.

PR #4218, opened 2024-05-08
Another attempt at using the lsp API for some progress reporting

michaelpj:


michaelpj:
Okay, I think the tests should pass now. I had to get rid of some of the functional tests because they test the real HLS executable, so that now includes a delay on starting progress. So we can't observe it unless we can make things take long enough, and we can't make the delay shorter because we're using the real HLS (unlike the `ghcide` tests which use a special executable with different options). I was able to arrange for one of the tests to have a long enough delay, so that might have to do for now.

As far as I can tell, this seems to work reliably. I would also like to rewrite the other special progress reporter, not sure if I'll do that in this PR or not.

michaelpj:
Okay, many tests rely on waiting for progress notifications and are now somewhat broken. I shall continue on it.

michaelpj:
This will fix https://github.com/haskell/haskell-language-server/issues/381, in that we won't show progress for fast operations, which will include most formatting operations.

It should also fix https://github.com/haskell/haskell-language-server/issues/3577, or at least mitigate it, in that updates will be throttled and won't be sent for short-lived "processing" operations.

michaelpj:
Ready for a look, perhaps @soulomoon ?

test/functional/Progress.hs
```diff
@@ -52,29 +46,10 @@ tests =
                           (command ^. L.command)
                           (decode $ encode $ fromJust $ command ^. L.arguments)
 
-                      expectProgressMessages ["Evaluating"] activeProgressTokens
+                      expectProgressMessages ["Evaluating"] createdProgressTokens activeProgressTokens
                   _ -> error $ "Unexpected response result: " ++ show response
-        , requiresOrmoluPlugin $ testCase "ormolu plugin sends progress notifications" $ do
```
soulomoon, line 57: So these two tests are not working anymore, because of the way we handle the notification?

test/functional/Progress.hs
```diff
@@ -25,19 +23,15 @@ tests :: TestTree
 tests =
     testGroup
         "window/workDoneProgress"
-        [ testCase "sends indefinite progress notifications" $
-            runSession hlsLspCommand progressCaps "test/testdata/diagnostics" $ do
-                let path = "Foo.hs"
-                _ <- openDoc path "haskell"
-                expectProgressMessages [pack ("Setting up diagnostics (for " ++ path ++ ")"), "Processing", "Indexing"] []
-        , requiresEvalPlugin $ testCase "eval plugin sends progress reports" $
```
soulomoon, line 33: And this one

soulomoon (approved):
Looks good ! 
Just want to ask a thing, so we lose some of the notification tests since we now have 1 s delay for notification ? I wonder if we can set the shorter delay for testing and gain those test back?

michaelpj:
Yes, so we have some special arguments for the `ghcide` executable that get used during testing that set the delay to 0. But for the HLS functional tests we don't call the main function of HLS we actually run the built executable, so I'm not sure how to make that nicely configurable...

soulomoon:
Look like a limitation of `runSessionWithConfig`

michaelpj:
Yeah, or maybe I can use an environment variable or something.

michaelpj:
We already had a `--test` argument to HLS that turns on the testing stuff, which I had already made change the delay. So I just needed to pass that in the test invocation and it works again.

ghcide/src/Development/IDE/Main.hs
```diff
@@ -238,7 +238,15 @@ defaultArguments recorder plugins = Arguments
             { optCheckProject = pure $ checkProject config
             , optCheckParents = pure $ checkParents config
             }
-        , argsLspOptions = def {LSP.optCompletionTriggerCharacters = Just "."}
+        , argsLspOptions = def
+            { LSP.optCompletionTriggerCharacters = Just "."
+            -- Generally people start to notice that something is taking a while at about 1s, so
+            -- that's when we start reporting progress
+            , LSP.optProgressStartDelay = 1_00_000
```
fendor, line 245: What a weird number. Shouldn't this be `1_000_000`

michaelpj: :facepalm:

PR #4222, opened 2024-05-11
Cleanup cabal files, ghc compat code, fix ghc warnings

jhrcek:
Mostly cleaning up and reorganizing imports to avoid most dodgy imports/exports warnings and to follow the guildelines for using CPP in imports. Other changes highlighted in comments.

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -14,7 +14,7 @@ homepage:
   https://github.com/haskell/haskell-language-server/tree/master/ghcide#readme
 
 bug-reports:        https://github.com/haskell/haskell-language-server/issues
-tested-with:        GHC ==9.8.2 || ==9.6.4 || ==9.4.8 || ==9.2.8
+tested-with:        GHC ==9.8.2 || ==9.6.5 || ==9.4.8 || ==9.2.8
 extra-source-files:
   CHANGELOG.md
   README.md
@@ -98,7 +98,6 @@ library
     , prettyprinter-ansi-terminal
     , random
     , regex-tdfa                   >=1.3.1.0
-    , row-types
     , safe-exceptions
     , sorted-list
     , sqlite-simple
@@ -277,20 +276,7 @@ library ghcide-test-utils
     Development.IDE.Test.Runfiles
 
   build-depends:
-      aeson,
-      base > 4.9 && < 5,
-      containers,
-      data-default,
-      directory,
-      extra,
-      filepath,
-      ghcide,
-      lsp-types,
-      hls-plugin-api,
-      lens,
-      lsp-test ^>= 0.17,
-      tasty-hunit >= 0.10,
-      text,
+      base > 4.9 && < 5
 
   default-extensions:
       LambdaCase
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -271,7 +271,7 @@ library ghcide-test-utils
   visibility: public
   default-language:   GHC2021
 
-  hs-source-dirs: test/src test/cabal
+  hs-source-dirs: test/cabal
   exposed-modules:
     Development.IDE.Test.Runfiles
 
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -482,11 +482,7 @@ mkHiFileResultNoCompile session tcm = do
       tcGblEnv = tmrTypechecked tcm
   details <- makeSimpleDetails hsc_env_tmp tcGblEnv
   sf <- finalSafeMode (ms_hspp_opts ms) tcGblEnv
-  iface' <- mkIfaceTc hsc_env_tmp sf details ms
-#if MIN_VERSION_ghc(9,5,0)
-                      Nothing
-#endif
-                      tcGblEnv
+  iface' <- mkIfaceTc hsc_env_tmp sf details ms Nothing tcGblEnv
   let iface = iface' { mi_globals = Nothing, mi_usages = filterUsages (mi_usages iface') } -- See Note [Clearing mi_globals after generating an iface]
   pure $! mkHiFileResult ms iface details (tmrRuntimeModules tcm) Nothing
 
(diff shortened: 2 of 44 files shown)


ghcide/ghcide.cabal
```diff
@@ -272,25 +271,12 @@ library ghcide-test-utils
   visibility: public
   default-language:   GHC2021
 
-  hs-source-dirs: test/src test/cabal
+  hs-source-dirs: test/cabal
```
jhrcek, line 274: It seems to me that we can now get rid of ghcide-test-utils entirely now that it consists of this one module?
https://github.com/haskell/haskell-language-server/blob/master/ghcide/test/cabal/Development/IDE/Test/Runfiles.hs

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1148,36 +1143,23 @@ needsCompilationRule file = do
...
       xopt LangExt.TemplateHaskell dflags || xopt LangExt.QuasiQuotes dflags
 
--- | How should we compile this module?
--- (assuming we do in fact need to compile it).
--- Depends on whether it uses unboxed tuples or sums
-computeLinkableTypeForDynFlags :: DynFlags -> LinkableType
-computeLinkableTypeForDynFlags d
-          = BCOLinkable
-  where -- unboxed_tuples_or_sums is only used in GHC < 9.2
-        _unboxed_tuples_or_sums =
```
jhrcek, line 1178: Interesting simplification #1: this is unused on GHCs we support, which means this function doesn't depend on DynFlags and we can just hardcode BCOLinkable in bunch of places leading to bunch of simplifications in this module.

ghcide/src/Development/IDE/GHC/Compat/Plugins.hs
```diff
@@ -58,18 +58,17 @@ import qualified GHC.Parser.Lexer                      as Lexer
 type PsMessages = (Bag WarnMsg, Bag ErrMsg)
 #endif
 
-getPsMessages :: PState -> DynFlags -> PsMessages
-getPsMessages pst _dflags = --dfags is only used if GHC < 9.2
```
jhrcek, line 62: Interesting simplification #2: few more unused DynFlags removed because supported GHCs don't need them anymore.

ghcide/src/Development/IDE/GHC/Orphans.hs
```diff
@@ -56,11 +58,22 @@ instance NFData SafeHaskellMode where rnf = rwhnf
 instance Show Linkable where show = unpack . printOutputable
 instance NFData Linkable where rnf (LM a b c) = rnf a `seq` rnf b `seq` rnf c
 instance NFData Unlinked where
-  rnf (DotO f)   = rnf f
-  rnf (DotA f)   = rnf f
-  rnf (DotDLL f) = rnf f
-  rnf (BCOs a b) = seqCompiledByteCode a `seq` liftRnf rwhnf b
-  rnf  _         = error "rnf: not implemented for Unlinked"
```
jhrcek, line 63: Completing this NFData implementation for more recent GHCs

ghcide/src/Development/IDE/GHC/Compat.hs
```diff
@@ -241,9 +238,9 @@ nonDetOccEnvElts = occEnvElts
 
 type ModIfaceAnnotation = Annotation
 
-#if MIN_VERSION_ghc(9,3,0)
-nameEnvElts :: NameEnv a -> [a]
-nameEnvElts = nonDetNameEnvElts
+#if !MIN_VERSION_ghc(9,3,0)
+nonDetNameEnvElts :: NameEnv a -> [a]
+nonDetNameEnvElts = nameEnvElts
```
jhrcek, line 243: This name flip warrants explanation:
Since this function is named `nonDetNameEnvElts` in ghc 9.4+
I flipped the "compat" function to this new name and aliased the old function with this new name.
With the expectation that the compat function can be dropped when when we remove ghc 9.2 as opposed to keeping the old name even when it's no longer needed.

ghcide/ghcide.cabal
michaelpj: cc @soulomoon ?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -482,11 +472,7 @@ mkHiFileResultNoCompile session tcm = do
       tcGblEnv = tmrTypechecked tcm
   details <- makeSimpleDetails hsc_env_tmp tcGblEnv
   sf <- finalSafeMode (ms_hspp_opts ms) tcGblEnv
-  iface' <- mkIfaceTc hsc_env_tmp sf details ms
-#if MIN_VERSION_ghc(9,5,0)
```
michaelpj, line 486: Surely we must still need this?

ghcide/src/Development/IDE/Core/Rules.hs
michaelpj: Seems plausible!

ghcide/src/Development/IDE/Core/Compile.hs
michaelpj: Ah I see, you changed the definition

michaelpj (approved):
Looks great, thanks for doing this! Do you think we're at the point where we could have `pedantic` for `ghcide` to keep things tidier?

ghcide/ghcide.cabal
soulomoon: Yes, test/src get merged with the hls-test-utils

soulomoon (approved):
LGTM, more cleaner

fendor (approved):
LGTM!

ghcide/ghcide.cabal
soulomoon: We will eventually drop all the uses of `locateGhcideExecutable`, after we are done with #4173

jhrcek:
> Looks great, thanks for doing this! Do you think we're at the point where we could have pedantic for ghcide to keep things tidier?

We're close, but not yet. There are still bunch of other warnings that need to be fixed. I'll work towards that next.

PR #4223, opened 2024-05-12
CI: enable pedantic for each cabal package separately

jhrcek:
It seems that the original "all +pedantic" doesn't enable pedantic flag for everything 
If you run `cabal configure --constraint "all +pedantic"` followed by 

```bash
cabal build -v2   #among other things verbosity level 2 outputs "Flags chosen: ... pedantic=False... " for each target
# or 
cabal build -v2 all
# or 
cabal build -v2 <specific-component>
```
you' won't see any pedantic=True anywhere.
Doing it on per-cabal-file basis seems to work fine (and leads to desirable CI failures).

### DIFF
--- a/.github/workflows/flags.yml
+++ b/.github/workflows/flags.yml
@@ -76,10 +76,11 @@ jobs:
       - name: Configue non-default flags for all components
         run: |
           cabal configure \
-              --constraint "hls-graph +embed-files +stm-stats" \
-              --constraint "ghcide +ekg +executable +test-exe" \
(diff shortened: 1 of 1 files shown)


michaelpj:
Baffling.

jhrcek:
Closing this for now as I enabled pedantic for most subprojects in https://github.com/haskell/haskell-language-server/pull/4224
Will work towards fixing all warnings in ghcide in separate PR.

PR #4224, opened 2024-05-12
Actually enable pedantic flag in ci flags job

jhrcek:
As found in https://github.com/haskell/haskell-language-server/pull/4223 we're not actually enabling pedantic in CI at all.
This enables it for all cabal projects for which most warnings have been fixed so far (not yet for ghcide).

### DIFF
--- a/.github/workflows/flags.yml
+++ b/.github/workflows/flags.yml
@@ -76,10 +76,10 @@ jobs:
       - name: Configue non-default flags for all components
         run: |
           cabal configure \
-              --constraint "hls-graph +embed-files +stm-stats" \
+              --constraint "haskell-language-server +pedantic" \
+              --constraint "hls-graph +embed-files +pedantic +stm-stats" \
               --constraint "ghcide +ekg +executable +test-exe" \
-              --constraint "hls-plugin-api -use-fingertree" \
-              --constraint "all +pedantic"
+              --constraint "hls-plugin-api +pedantic -use-fingertree"
           cat cabal.project.local
 
       - name: Build everything with non-default flags
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -340,6 +340,8 @@ test-suite hls-class-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
   if !flag(class)
     buildable: False
+  if impl(ghc >= 9.8)
+    ghc-options: -Wwarn=x-partial
   type:             exitcode-stdio-1.0
   hs-source-dirs:   plugins/hls-class-plugin/test
   main-is:          Main.hs
@@ -1183,6 +1185,8 @@ test-suite hls-change-type-signature-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
   if !flag(changeTypeSignature)
     buildable: False
+  if impl(ghc >= 9.8)
+    ghc-options: -Wwarn=x-partial
   type:             exitcode-stdio-1.0
   hs-source-dirs:   plugins/hls-change-type-signature-plugin/test
   main-is:          Main.hs
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -1680,25 +1680,23 @@ test-suite hls-refactor-plugin-tests
   ghc-options:      -O0
   build-depends:
     , base
+    , data-default
+    , directory
+    , extra
     , filepath
+    , ghcide:ghcide
     , haskell-language-server:hls-refactor-plugin
     , hls-test-utils      == 2.8.0.0
     , lens
-    , lsp-types
-    , text
-    , hls-plugin-api
-    , parser-combinators
-    , data-default
-    , extra
-    , ghcide:ghcide
-    , shake
-    , hls-plugin-api
     , lsp-test
-    , directory
+    , lsp-types
+    , parser-combinators
     , regex-tdfa
-    , tasty-hunit
-    , tasty-expected-failure
+    , shake
     , tasty
+    , tasty-expected-failure
+    , tasty-hunit
+    , text
 
 -----------------------------
 -- semantic tokens plugin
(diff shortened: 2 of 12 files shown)


fendor (approved):
LGTM, some nitpicks

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1087,8 +1087,9 @@ suggestImportDisambiguation df (Just txt) ps fileContents diag@Diagnostic {..}
                 ++ [HideOthers restImports | not (null restImports)]
             ] ++ [ ( renderUniquify mode T.empty symbol True
               , disambiguateSymbol ps fileContents diag symbol mode
-              ) | local, not (null targetsWithRestImports)
-                , let mode = HideOthers (uncurry (:) (head targetsWithRestImports))
+              ) | local
+                , (m,ms) <- take 1 targetsWithRestImports
+                , let mode = HideOthers (m:ms)
```
fendor, line 1092: Before this change, the plugin would crash because some precondition was violated. With this change, it will be ignored. I don't think that's strictly an improvement, I would prefer it if we  refactored this code to use `NonEmpty` or some other proof why this will always give us at least one result.

fendor (changes requested):
One thing I noticed:

plugins/hls-change-type-signature-plugin/test/Main.hs
```diff
@@ -1,3 +1,8 @@
+{-# LANGUAGE CPP #-}
+#if __GLASGOW_HASKELL__ >= 908
+{-# OPTIONS_GHC -Wwarn=x-partial #-}
+#endif
+
```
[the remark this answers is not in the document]
jhrcek: I went for this to make the disabling as local as possible.
But since you're proponent of using partial functions in tests, I think it makes sense to use this cabal file variant in  specific test suite targets that use partial functions.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
jhrcek: I don't think the original impl. would crash, because it had `not (null targetsWithRestImports)` so the list comprehension would also "fail silently" returning empty list before evaluating head.
But I agree the `take 1` is confusing, so I switched to using direct pattern match on the list.

haskell-language-server.cabal
```diff
@@ -1763,19 +1765,17 @@ test-suite hls-semantic-tokens-plugin-tests
     , aeson
     , base
     , containers
+    , data-default
     , filepath
+    , ghcide                == 2.8.0.0
     , haskell-language-server:hls-semantic-tokens-plugin
-    , hls-test-utils              == 2.8.0.0
-    , hls-plugin-api
```
jhrcek, line 1769: I sorted these 2 dependency lists alphabetically, because I kept getting "unused package hls-plugin-api" warning, even though I deleted it. It turns out it was in the list twice :facepalm:

haskell-language-server.cabal
```diff
@@ -340,6 +340,8 @@ test-suite hls-class-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
   if !flag(class)
     buildable: False
+  if impl(ghc >= 9.8)
+    ghc-options: -Wwarn=x-partial
```
jhrcek, line 344: Given that @fendor is for using partial functions in tests, I'm even considering switching to disabling these warnings for tests as there's no point in making the build output more noisy for non-actionable stuff (My favorite [broken window theory](https://en.wikipedia.org/wiki/Broken_windows_theory) - the more warnings / flaky tests there are, the more likely people are to ignore them).
```suggestion
    ghc-options: -Wno-x-partial
```

jhrcek: Maybe I can even put it into `common test-defaults` as there's no reason to have additional mechanism to prevent new occurrences (we already have hlint in ci)

fendor: Sounds like a great idea!

haskell-language-server.cabal
fendor: Perhaps we should always format our .cabal files using a formatter? I know, such a suggestion is deeply abhorrent to most haskelleres, but now we have two great formatters, might as well use them :upside_down_face: 

I'll add it to the list of talking points for the next meeting.

jhrcek: Count me in as a supporter. I cabal-fmt all my private project's cabal files.

.github/workflows/flags.yml
```diff
@@ -76,10 +76,10 @@ jobs:
       - name: Configue non-default flags for all components
         run: |
           cabal configure \
-              --constraint "hls-graph +embed-files +stm-stats" \
+              --constraint "haskell-language-server +pedantic" \
+              --constraint "hls-graph +embed-files +pedantic +stm-stats" \
               --constraint "ghcide +ekg +executable +test-exe" \
-              --constraint "hls-plugin-api -use-fingertree" \
-              --constraint "all +pedantic"
```
jhrcek, line 82: This `all +pedantic` did not have any effect - it needs to be enabled on per-cabal-file basis.

fendor (approved):
LGTM

.github/workflows/flags.yml
jhrcek: I didn't open any ticket about this in cabal, nor did I research why it doesn't work/whether it should work at all.
I just mistakenly assumed this works when I refactored this job.
Let me ask about it in cabal issue tracker later today.

jhrcek: It turns out that the way to enable the flag for all local packages is via
`cabal configure --flag pedantic`

I'll make use of that in upcoming PR after I fix warnings in ghcide

PR #4227, opened 2024-05-13
[Migrate BootTests] part of #4173 Migrate ghcide tests to hls test utils

soulomoon:


hls-test-utils/src/Test/Hls.hs
```diff
@@ -676,6 +677,25 @@ runSessionWithServer' disableKick pluginsDp conf sconf caps root s =  withLock l
             putStrLn $ "Finishing canceling (took " <> showDuration t <> "s)"
     pure x
 
+-- | Host a server, and run a test session on it
+-- Note: cwd will be shifted into @root@ in @Session a@
```
soulomoon, line 681: I do not think it is valid?

fendor: What do you mean? Looks correct to me. I don't think the tests will work though, if you don't lock, as shifting the cwd will affect all tests.

soulomoon: Oh I see, the shift happens in lsp code, right?
We might want one with no lock too, so that it can be run inside the one with the lock. Some test in ghcide need to nest the runs.

fendor: No, the shift happens in the executable https://github.com/haskell/haskell-language-server/blob/master/src/Ide/Main.hs#L126

Ok, I understand that nesting might be necessary sometimes. However, the comment should make it clear that all integration tests need to be locked, since we launch a thread, not a new process, for the tests, which all share the same `CWD` variable.

soulomoon: Thanks for the classification, it probably be a good idea to eliminate the need for `getCurrentDirectory` inside `IDEMain.defaultMain` now. I'll try to do it.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -632,15 +632,15 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
...
     let consultCradle :: Maybe FilePath -> FilePath -> IO (IdeResult HscEnvEq, [FilePath])
         consultCradle hieYaml cfp = do
-           lfpLog <- flip makeRelative cfp <$> getCurrentDirectory
+           let lfpLog = makeRelative dir cfp
            logWith recorder Info $ LogCradlePath lfpLog
 
            when (isNothing hieYaml) $
              logWith recorder Warning $ LogCradleNotFound lfpLog
 
            cradle <- loadCradle recorder hieYaml dir
```
soulomoon, line 641: Is loadCradle depend on `CWD`, or would it change it ?

soulomoon: Probably fine,  I am not any usage of that in HieBios.

hls-test-utils/src/Test/Hls.hs
soulomoon: EMM, it might be hard, rule have no access to root dir, but just currentworking dir
https://github.com/soulomoon/haskell-language-server/blob/542ea2603f361ba43b839fee0779d1d1ac034c2b/ghcide/src/Development/IDE/Core/Rules.hs#L722

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -848,7 +846,7 @@ getModIfaceFromDiskAndIndexRule recorder =
       hie_loc = Compat.ml_hie_file $ ms_location ms
   fileHash <- liftIO $ Util.getFileHash hie_loc
   mrow <- liftIO $ withHieDb (\hieDb -> HieDb.lookupHieFileFromSource hieDb (fromNormalizedFilePath f))
-  hie_loc' <- liftIO $ traverse (makeAbsolute . HieDb.hieModuleHieFile) mrow
```
soulomoon, line 851: the doc said hieModuleHieFile would return a full path, so there is no need to `makeAbsolute` here

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -58,15 +57,19 @@ updateHscEnvEq oldHscEnvEq newHscEnv = do
   let update newUnique = oldHscEnvEq { envUnique = newUnique, hscEnv = newHscEnv }
   update <$> Unique.newUnique
 
+toAbsolute :: FilePath -> FilePath -> FilePath
```
soulomoon, line 60: need to find a place to put all the `toAbsolute` functions so that it can be shared

hls-test-utils/src/Test/Hls.hs
soulomoon: Also we might not find the file in `TemplateHaskell` here ? If current dir is not set.
https://github.com/soulomoon/haskell-language-server/blob/67438ef6aa23ad404af2f142e52dce78b617f95c/ghcide/test/exe/DependentFileTest.hs#L40

soulomoon: Attemp moved to https://github.com/haskell/haskell-language-server/pull/4231
Focusing on BootTests migration now.

fendor (approved):
LGTM, thanks!

PR #4229, opened 2024-05-13
Draft: Format .cabal files with cabal-gild

fendor:


haskell-language-server.cabal
```diff
@@ -303,264 +302,273 @@ common qualifyImportedNames
...
+    common-deps
+    -- configuration
+    , warnings
+    , pedantic
+
+  main-is: Main.hs
+  hs-source-dirs: exe
   ghc-options:
-    -threaded
     -- allow user RTS overrides
```
michaelpj, line 437: bad comment placement

michaelpj: I think it's not a blocker, we can move these comments above the field and clarify

hls-test-utils/hls-test-utils.cabal
```diff
@@ -1,61 +1,59 @@
...
+    aeson,
+    async,
+    base >=4.12 && <5,
+    bytestring,
+    containers,
+    data-default,
+    directory,
+    extra,
+    filepath,
+    ghcide ==2.8.0.0,
```
jhrcek, line 44: One more reason I prefer cabal-gild, is that it doesn't try to preserve version constraint indentation.
This leads to more diff friendly changes (adding/removing dependency with longer than longest name, doesn't lead to indentation changes for other version constraints)
e.g. 
```
short >=1.0
longer >= 2.0   < Adding longer dependency re-indents version constraint for shorter, which makes for longer diffs
```

ghcide/ghcide.cabal
```diff
@@ -1,20 +1,18 @@
...
+name: ghcide
+version: 2.8.0.0
+license: Apache-2.0
+license-file: LICENSE
+author: Digital Asset and Ghcide contributors
+maintainer: Ghcide contributors
+copyright: Digital Asset and Ghcide contributors 2018-2020
+synopsis: The core of an IDE
+description: A library for building Haskell IDE's on top of the GHC API.
+homepage: https://github.com/haskell/haskell-language-server/tree/master/ghcide#readme
```
tfausak, line 13: Which version of Gild was used here? As of [1.2.0.0](https://github.com/tfausak/cabal-gild/releases/tag/1.2.0.0) it shouldn't have re-formatted this line. (The URL should've stayed on its own line after `homepage:`.)

soulomoon:
can `cabal-gild` have `,` prefixing line breaking for items?

ghcide/ghcide.cabal
```diff
@@ -45,77 +43,76 @@ common warnings
...
-    , text
-    , text-rope
-    , time
-    , transformers
-    , unliftio                     >=0.2.6
-    , unliftio-core
-    , unordered-containers         >=0.2.10.0
-    , vector
+    Diff ^>=0.4.0,
+    Glob,
```
tfausak, line 49: I chose to make Gild sort things case sensitively, but I'm not set on it. See https://github.com/tfausak/cabal-gild/issues/46.

ghcide/ghcide.cabal
```diff
@@ -45,77 +43,76 @@ common warnings
...
+    stm-containers,
+    syb,
+    text,
+    text-rope,
+    time,
+    transformers,
+    unliftio >=0.2.6,
+    unliftio-core,
+    unordered-containers >=0.2.10.0,
+    vector,
```
tfausak, line 112: A lot of noise here from switching to trailing commas. I see that @soulomoon had a question about it too. Gild isn't (and hopefully won't be) configurable. I think trailing commas are generally preferable, especially when a final trailing comma is allowed (like it is in Cabal files for recent enough versions). It's a bit unfortunate in cases like this, but overall I think trailing commas are better.

hie-compat/hie-compat.cabal
```diff
@@ -1,41 +1,59 @@
...
-license:             Apache-2.0
+cabal-version: 1.22
+name: hie-compat
+version: 0.3.1.2
+synopsis: HIE files for GHC 8.8 and other HIE file backports
+license: Apache-2.0
 description:
   Backports for HIE files to GHC 8.8, along with a few other backports
   of HIE file related fixes for ghcide.
-
```
tfausak, line 9: This line was deleted because blank lines in the `description` field must be replaced by a lone period (`foo\n.\nbar`). You can see on Hackage that the description is all one paragraph: https://hackage.haskell.org/package/hie-compat-0.3.1.2

If you don't want to do that, you could upgrade to `cabal-version: 3.0`. You can see where Gild handles this here: https://github.com/tfausak/cabal-gild/blob/51425c05c67e6c335ba9024ced779669e3a4aca0/source/library/CabalGild/Unstable/Action/ReflowText.hs#L34

Here's the upstream Cabal PR for changing this behavior: https://github.com/haskell/cabal/pull/5981

hie-compat/hie-compat.cabal
```diff
@@ -1,41 +1,59 @@
...
     Compat.HieAst
     Compat.HieBin
-    Compat.HieTypes
     Compat.HieDebug
+    Compat.HieTypes
     Compat.HieUtils
 
-  if (impl(ghc >= 9.2) && impl(ghc < 9.3))
-    hs-source-dirs: src-ghc92 src-reexport-ghc9
+  if (impl(ghc >= 9.2)&& impl(ghc < 9.3))
```
tfausak, line 53: Hmm this formatting is no good. There should be a space before the `&&`. I made an issue for this: https://github.com/tfausak/cabal-gild/issues/72

tfausak (commented):
👋 I'm the author/maintainer of Gild. Let me know if I can do anything to help here!

ghcide/ghcide.cabal
jhrcek: I'm fan of this approach (non-configurability).
Similar approach is taken by elm-format, which became standard within elm community.
I like that most libraries you can get is formatted with it so there's very consistent feel to the whole ecosystem.
It eliminates a lot of pointless bickering on what format to choose.
One gets used to it pretty quickly.

ghcide/ghcide.cabal
fendor: 1.1.1.0 was used, will reformat with 1.2.0.0, thanks for the hint!

ghcide/ghcide.cabal
soulomoon: Thanx for the reasonning, I am convinced that it is good.

haskell-language-server.cabal
```diff
@@ -1792,225 +1819,236 @@ common notes
...
-                  , stylishHaskell
-                  , refactor
-                  , overloadedRecordDot
-                  , semanticTokens
-                  , notes
+  import:
+    defaults
+    , warnings
+    , pedantic
+    -- plugins
```
michaelpj, line 1876: @tfausak here! This comment is (surprisingly?) preserved in situ.

fendor:
Reformatted the `.cabal` files with cabal-gild-1.3.1.0 for accurate comparison.

hie-compat/hie-compat.cabal
tfausak: This has been fixed in version 1.3.1.3: https://github.com/tfausak/cabal-gild/releases/tag/1.3.1.3

haskell-language-server.cabal
tfausak: Starting with [version 1.4.0.0](https://github.com/tfausak/cabal-gild/releases/tag/1.4.0.0), Gild will always float comments up to the top of fields.

fendor:
@VeryMilkyJoe Thanks for rebasing this, I think we should merge this PR now and see how annoying it is for future contributors.

Formatted with cabal-gild-1.5.0.1

michaelpj:
Shipit!

ghcide/ghcide.cabal
```diff
@@ -1,32 +1,32 @@
...
+copyright: Digital Asset and Ghcide contributors 2018-2020
+synopsis: The core of an IDE
+description: A library for building Haskell IDE's on top of the GHC API.
 homepage:
   https://github.com/haskell/haskell-language-server/tree/master/ghcide#readme
 
-bug-reports:        https://github.com/haskell/haskell-language-server/issues
-tested-with:        GHC == {9.12.2, 9.10.1, 9.8.4, 9.6.7}
+bug-reports: https://github.com/haskell/haskell-language-server/issues
+tested-with: ghc ==9.12.2 || ==9.10.1 || ==9.8.4 || ==9.6.7
```
tfausak: Starting with version 1.5.0.3, Gild will correctly leave this in set notation. See https://github.com/tfausak/cabal-gild/pull/118.

shake-bench/shake-bench.cabal
```diff
@@ -1,45 +1,46 @@
...
-        filepath,
-        lens,
-        lens-aeson,
-        mtl,
-        shake,
-        text
+    Chart,
+    Chart-diagrams,
+    aeson,
+    base >=4 && <5,
```
tfausak: Same note here: Starting with version 1.5.0.3, Gild will correctly retain the wildcard version constraint. https://github.com/tfausak/cabal-gild/pull/118

haskell-language-server.cabal
```diff
@@ -124,54 +123,55 @@ common cabalfmt
...
+  main-is: Main.hs
+  build-depends:
+    directory,
+    filepath,
+    haskell-language-server:hls-cabal-fmt-plugin,
+    haskell-language-server:hls-cabal-plugin,
+    hls-plugin-api ==2.11.0.0,
+    hls-test-utils ==2.11.0.0,
+
+  if flag(isolatecabalfmttests)
```
tfausak, line 164: Cabal converts flag names to lowercase. Gild does the same in order to make this conversion obvious. However it could retain the case from the input if desired. See https://github.com/haskell/haskell-language-server/pull/4230#discussion_r1600436999.

haskell-language-server.cabal
```diff
@@ -124,54 +123,55 @@ common cabalfmt
...
-
-  if flag(isolateCabalfmtTests)
+  type: exitcode-stdio-1.0
+  hs-source-dirs: plugins/hls-cabal-fmt-plugin/test
+  main-is: Main.hs
+  build-depends:
+    directory,
+    filepath,
+    haskell-language-server:hls-cabal-fmt-plugin,
+    haskell-language-server:hls-cabal-plugin,
```
tfausak, line 160: Just so you know, these can use set notation:
```suggestion
    haskell-language-server:{hls-cabal-fmt-plugin, hls-cabal-plugin},
```

haskell-language-server.cabal
```diff
@@ -1047,581 +1066,602 @@ test-suite hls-alternate-number-format-plugin-tests
...
-  main-is:          Main.hs
+  type: exitcode-stdio-1.0
+  hs-source-dirs: plugins/hls-fourmolu-plugin/test
+  main-is: Main.hs
+
   -- Work around https://gitlab.haskell.org/ghc/ghc/-/issues/24648
-  if os(darwin)
-    ghc-options: -optl-Wl,-ld_classic
+  if os(osx)
+    ghc-options: "-optl-Wl,-ld_classic"
```
tfausak, line 1515: Hmm, I wonder why this got quoted. Maybe because there's a comma in it? 

Also note that `darwin` and `osx` are aliases. Gild canonicalizes operating systems (and architectures).

haskell-language-server.cabal
```diff
@@ -1635,148 +1675,151 @@ library hls-refactor-plugin
...
-    , hls-graph
-    , dlist
-    , deepseq
-    , mtl
-    , lens
-    , time
+
+  hs-source-dirs: plugins/hls-refactor-plugin/src
+  build-depends:
     -- FIXME: Only needed to workaround for qualified imports in GHC 9.4
```
tfausak: This comment got floated up to the top of the `build-depends`, so it doesn't really make sense anymore. I would suggest something like this:

```
build-depends:
  bytestring,
  ...

-- FIXME: Only needed to workaround for qualified imports in GHC 9.4
build-depends:
  parser-combinators,
  regex-applicative
```

haskell-language-server.cabal
```diff
@@ -1785,217 +1828,231 @@ common notes
...
-                  , floskell
-                  , fourmolu
-                  , ormolu
-                  , stylishHaskell
-                  , refactor
-                  , overloadedRecordDot
-                  , semanticTokens
-                  , notes
+  import:
+    -- plugins
```
tfausak, line 1885: Same note here about comment placement.

haskell-language-server.cabal
```diff
@@ -2179,119 +2243,140 @@ test-suite ghcide-tests
...
+  build-depends:
+    aeson,
+    async,
+    binary,
+    bytestring,
+    deepseq,
+    directory,
+    extra,
+    filepath,
+    ghcide,
```
tfausak: In this case I don't think it changes the semantics, but starting with version 1.5.0.3, Gild would keep this as `ghcide:{ghcide}`. See https://github.com/tfausak/cabal-gild/pull/115.

tfausak (commented):
FYI https://github.com/marketplace/actions/setup-gild

PR #4231, opened 2024-05-14
Refactor hls-test-util and reduce getCurrentDirectory after initilization

soulomoon:
What's done
* [x] Refactor the `runSession*` family function, properly add `TestConfig`, `runSessionWithTestConfig`, as the most generic `runSession*` function.
* [x] remove raraly used variants of `runSession*` functions and replaced by `runSessionWithTestConfig`. 
* [x] migrate `ExceptionTests ClientSettingsTests CodeLensTests CPPTests CradleTests` to use the `hls-test-utils`
* [x] Only shift to lsp root when current root is different from the lsp root in DefaultMain of ghcide. 
* [x] Remove most usage for `getCurrentDirectory`(After DefaultMain is called), Only remain those in top level of wrapper and exe, implement https://github.com/haskell/haskell-language-server/issues/3736#issuecomment-1924507928

What's remains:
* golden tests are not using `TestConfig`, maybe we can refactor them to use `TestConfig` and cut the number of variants too.
* Hlint, Stan and some Template haskell tests still depend on current working dir. But they are out of our reach.
* `hie.yaml` specify relative files 
  ```
  cradle:
  direct:
    arguments: ["-unit" ,"@a-1.0.0-inplace"
               ,"-unit" ,"@b-1.0.0-inplace" 
               ,"-unit" ,"@c-1.0.0-inplace"
               ]

  ```

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -811,7 +811,7 @@ fromTargetId :: [FilePath]          -- ^ import paths
              -> TargetId
              -> IdeResult HscEnvEq
              -> DependencyInfo
-             -> FilePath
+             -> FilePath -- ^ root dir -- see Note [Root Directory]
              -> IO [TargetDetails]
 -- For a target module we consider all the import paths
 fromTargetId is exts (GHC.TargetModule modName) env dep dir = do
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -913,7 +913,7 @@ newComponentCache
          -> HscEnv             -- ^ An empty HscEnv
          -> [ComponentInfo]    -- ^ New components to be loaded
          -> [ComponentInfo]    -- ^ old, already existing components
-         -> FilePath           -- ^ root dir
+         -> FilePath           -- ^ root dir -- see Note [Root Directory]
          -> IO [ [TargetDetails] ]
 newComponentCache recorder exts cradlePath _cfp hsc_env old_cis new_cis dir = do
     let cis = Map.unionWith unionCIs (mkMap new_cis) (mkMap old_cis)
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -440,7 +440,7 @@ loadSession recorder = loadSessionWithOptions recorder def
 
 loadSessionWithOptions :: Recorder (WithPriority Log) -> SessionLoadingOptions -> FilePath -> IO (Action IdeGhcSession)
 loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir = do
-  let toAbsolutePath = toAbsolute rootDir
+  let toAbsolutePath = toAbsolute rootDir -- see Note [Root Directory]
   cradle_files <- newIORef []
   -- Mapping from hie.yaml file to HscEnv, one per hie.yaml file
   hscEnvs <- newVar Map.empty :: IO (Var HieMap)
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -811,7 +811,7 @@ fromTargetId :: [FilePath]          -- ^ import paths
              -> TargetId
              -> IdeResult HscEnvEq
              -> DependencyInfo
-             -> FilePath -- ^ root dir -- see Note [Root Directory]
+             -> FilePath -- ^ root dir, see Note [Root Directory]
              -> IO [TargetDetails]
 -- For a target module we consider all the import paths
 fromTargetId is exts (GHC.TargetModule modName) env dep dir = do
@@ -913,7 +913,7 @@ newComponentCache
          -> HscEnv             -- ^ An empty HscEnv
          -> [ComponentInfo]    -- ^ New components to be loaded
          -> [ComponentInfo]    -- ^ old, already existing components
-         -> FilePath           -- ^ root dir -- see Note [Root Directory]
+         -> FilePath           -- ^ root dir, see Note [Root Directory]
          -> IO [ [TargetDetails] ]
 newComponentCache recorder exts cradlePath _cfp hsc_env old_cis new_cis dir = do
     let cis = Map.unionWith unionCIs (mkMap new_cis) (mkMap old_cis)
@@ -1170,8 +1170,13 @@ addUnit unit_str = liftEwM $ do
   putCmdLineState (unit_str : units)
 
 -- | Throws if package flags are unsatisfiable
-setOptions :: GhcMonad m => NormalizedFilePath -> ComponentOptions -> DynFlags -> FilePath -> m (NonEmpty (DynFlags, [GHC.Target]))
-setOptions cfp (ComponentOptions theOpts compRoot _) dflags dir = do
+setOptions :: GhcMonad m
+    => NormalizedFilePath
+    -> ComponentOptions
+    -> DynFlags
+    -> FilePath -- ^ root dir, see Note [Root Directory]
+    -> m (NonEmpty (DynFlags, [GHC.Target]))
+setOptions cfp (ComponentOptions theOpts compRoot _) dflags rootDir = do
     ((theOpts',_errs,_warns),units) <- processCmdLineP unit_flags [] (map noLoc theOpts)
     case NE.nonEmpty units of
       Just us -> initMulti us
@@ -1194,7 +1199,7 @@ setOptions cfp (ComponentOptions theOpts compRoot _) dflags dir = do
         --
         -- If we don't end up with a target for the current file in the end, then
         -- we will report it as an error for that file
-        let abs_fp = toAbsolute dir (fromNormalizedFilePath cfp)
+        let abs_fp = toAbsolute rootDir (fromNormalizedFilePath cfp)
         let special_target = Compat.mkSimpleTarget df abs_fp
         pure $ (df, special_target : targets) :| []
     where
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -744,7 +744,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir = do
         void $ wait as
         asyncRes <- async $ getOptions file
         return (asyncRes, wait asyncRes)
-      pure $ (fmap . fmap) toAbsolutePath opts
+      pure opts
 
 -- | Run the specific cradle on a specific FilePath via hie-bios.
 -- This then builds dependencies or whatever based on the cradle, gets the
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -713,6 +713,7 @@ loadGhcSession recorder ghcSessionDepsConfig = do
 
     defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GhcSession file -> do
         IdeGhcSession{loadSessionFun} <- useNoFile_ GhcSessionIO
+        -- loading is always returning a absolute path now
         (val,deps) <- liftIO $ loadSessionFun $ fromNormalizedFilePath file
 
         -- add the deps to the Shake graph
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -23,6 +23,7 @@ module Development.IDE.Core.Rules(
     getParsedModuleWithComments,
     getClientConfigAction,
     usePropertyAction,
+    usePropertyByPathAction,
     getHieFile,
     -- * Rules
     CompiledLinkables(..),
@@ -147,9 +148,13 @@ import qualified Ide.Logger                                   as Logger
 import           Ide.Plugin.Config
 import           Ide.Plugin.Properties                        (HasProperty,
                                                                KeyNameProxy,
+                                                               KeyNamePath,
                                                                Properties,
                                                                ToHsType,
-                                                               useProperty)
+                                                               useProperty,
+                                                               usePropertyByPath,
+                                                               HasPropertyByPath
+                                                               )
 import           Ide.Types                                    (DynFlagsModifications (dynFlagsModifyGlobal, dynFlagsModifyParser),
                                                                PluginId)
 import           Language.LSP.Protocol.Message                (SMethod (SMethod_CustomMethod, SMethod_WindowShowMessage))
@@ -1059,6 +1064,16 @@ usePropertyAction ::
 usePropertyAction kn plId p = do
   pluginConfig <- getPluginConfigAction plId
   pure $ useProperty kn p $ plcConfig pluginConfig
+
+usePropertyByPathAction ::
+  (HasPropertyByPath props path t) =>
+  KeyNamePath path ->
+  PluginId ->
+  Properties props ->
+  Action (ToHsType t)
+usePropertyByPathAction path plId p = do
+  pluginConfig <- getPluginConfigAction plId
+  pure $ usePropertyByPath path p $ plcConfig pluginConfig
 
 -- ---------------------------------------------------------------------
 
--- a/ghcide/src/Development/IDE/Core/Service.hs
+++ b/ghcide/src/Development/IDE/Core/Service.hs
@@ -67,7 +67,7 @@ initialise :: Recorder (WithPriority Log)
            -> WithHieDb
            -> IndexQueue
            -> Monitoring
-           -> FilePath
+           -> FilePath -- ^ Root directory see Note [Root Directory]
            -> IO IdeState
 initialise recorder defaultConfig plugins mainRule lspEnv debouncer options withHieDb hiedbChan metrics rootDir = do
     shakeProfiling <- do
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -535,7 +535,7 @@ newtype ShakeSession = ShakeSession
 -- It helps to remove most usage for getCurrentDirectory(After DefaultMain of GhcIde is called),
 -- Using it instead of `getCurrentDirectory` allows us to avoid issues if we `setCurrentDirectory`
 -- somewhere else in the code.
--- And in turn, it is helps with testing in parallel, where we can keep the root directory
+-- And in turn, it helps with testing in parallel, where we can keep the root directory
 -- and the current directory separate.
 --
 -- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -532,18 +532,23 @@ newtype ShakeSession = ShakeSession
 -- We keep track of the root directory explicitly, which is the directory of the project root.
 -- We might be setting it via these options with decreasing priority:
 --
--- 1. from LSP workspace root 
--- 2. command line (--cwd) 
+-- 1. from LSP workspace root
+-- 2. command line (--cwd)
 -- 3. default to the current directory.
 --
--- It helps to remove most usage for `getCurrentDirectory`(After DefaultMain of GhcIde is called),
--- Using it instead of `getCurrentDirectory` allows us to avoid issues if we `setCurrentDirectory`
--- somewhere else in the code.
--- And in turn, it helps with testing in parallel, where we can keep the root directory
--- and the current directory separate.
+-- use `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
+-- If we modify the global Variable CWD, via `setCurrentDirectory`, all other test threads are suddenly affected,
+-- forcing us to run all integration tests sequentially.
+--
+-- Also, there might be a race condition if we depend on the current directory, as some plugin might change it.
+-- e.g. stylish's `loadConfig`. https://github.com/haskell/haskell-language-server/issues/4234
 --
 -- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
--- This is already deprecated and we can drop it in the future when the time comes.
+-- The root dir is deprecated, but we still need this now,
+-- since a lot of places in the codebase still rely on it.
+-- We can drop it in the future when the condition meets:
+-- 1. We can get rid all the usages of root directory in the codebase.
+-- 2. LSP version we support actually removes the root directory from the protocol.
 --
 
 -- | A Shake database plus persistent store. Can be thought of as storing
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -536,7 +536,7 @@ newtype ShakeSession = ShakeSession
 -- 2. command line (--cwd)
 -- 3. default to the current directory.
 --
--- use `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
+-- Using `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
 -- If we modify the global Variable CWD, via `setCurrentDirectory`, all other test threads are suddenly affected,
 -- forcing us to run all integration tests sequentially.
 --
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -554,7 +554,7 @@ data IdeState = IdeState
     ,shakeExtras          :: ShakeExtras
     ,shakeDatabaseProfile :: ShakeDatabase -> IO (Maybe FilePath)
     ,stopMonitoring       :: IO ()
-    -- see Note [Root Directory]
+    -- | See Note [Root Directory]
     ,rootDir              :: FilePath
     }
 
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -662,10 +662,9 @@ shakeOpen recorder lspEnv defaultConfig idePlugins debouncer
             atomically $ modifyTVar' exportsMap (<> em)
             logWith recorder Debug $ LogCreateHieDbExportsMapFinish (ExportsMap.size em)
 
-        progress <- do
-            let (before, after) = if testing then (0,0.1) else (0.1,0.1)
+        progress <-
             if reportProgress
-                then delayedProgressReporting before after lspEnv optProgressStyle
+                then progressReporting lspEnv optProgressStyle
                 else noProgressReporting
         actionQueue <- newQueue
 
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -544,8 +544,11 @@ newtype ShakeSession = ShakeSession
 -- e.g. stylish's `loadConfig`. https://github.com/haskell/haskell-language-server/issues/4234
 --
 -- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
--- The root dir is deprecated, but we still need this now,
--- since a lot of places in the codebase still rely on it.
+-- The root dir is deprecated, that means we should cleanup dependency on the project root(Or $CWD) thing gradually,
+-- so multi-workspaces can actually be supported when we use absolute path everywhere(might also need some high level design).
+-- That might not be possible unless we have everything adapt to it, like 'hlint' and 'evaluation of template haskell'.
+-- But we should still be working towards the goal.
+--
 -- We can drop it in the future once:
 -- 1. We can get rid all the usages of root directory in the codebase.
 -- 2. LSP version we support actually removes the root directory from the protocol.
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -546,7 +546,7 @@ newtype ShakeSession = ShakeSession
 -- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
 -- The root dir is deprecated, that means we should cleanup dependency on the project root(Or $CWD) thing gradually,
 -- so multi-workspaces can actually be supported when we use absolute path everywhere(might also need some high level design).
--- That might not be possible unless we have everything adapt to it, like 'hlint' and 'evaluation of template haskell'.
+-- That might not be possible unless we have everything adapted to it, like 'hlint' and 'evaluation of template haskell'.
 -- But we should still be working towards the goal.
 --
 -- We can drop it in the future once:
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -539,6 +539,7 @@ newtype ShakeSession = ShakeSession
 --
 -- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
 -- This is already deprecated and we can drop it in the future.
+--
 
 -- | A Shake database plus persistent store. Can be thought of as storing
 --   mappings from @(FilePath, k)@ to @RuleResult k@.
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -546,7 +546,7 @@ newtype ShakeSession = ShakeSession
 -- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
 -- The root dir is deprecated, but we still need this now,
 -- since a lot of places in the codebase still rely on it.
--- We can drop it in the future when the condition meets:
+-- We can drop it in the future once:
 -- 1. We can get rid all the usages of root directory in the codebase.
 -- 2. LSP version we support actually removes the root directory from the protocol.
 --
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -532,7 +532,7 @@ newtype ShakeSession = ShakeSession
 -- We keep track of the root directory explicitly, which is the directory of the project root.
 -- We might be setting it via these options with decreasing priority:
 --
--- 1. from LSP workspace root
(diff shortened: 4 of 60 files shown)


soulomoon:
`canonicalizePath` is blocking, should make path absolute before `canonicalizePath`

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -848,7 +846,7 @@ getModIfaceFromDiskAndIndexRule recorder =
       hie_loc = Compat.ml_hie_file $ ms_location ms
   fileHash <- liftIO $ Util.getFileHash hie_loc
   mrow <- liftIO $ withHieDb (\hieDb -> HieDb.lookupHieFileFromSource hieDb (fromNormalizedFilePath f))
-  hie_loc' <- liftIO $ traverse (makeAbsolute . HieDb.hieModuleHieFile) mrow
```
soulomoon, line 851: The doc said hieModuleHieFile would return a full path, I guess it is not necessary to absolute it.

soulomoon:
Stan is failing.
https://github.com/soulomoon/haskell-language-server/blob/2ebfafcea751adb8d0be9c1ad9c40a4f8771f261/plugins/hls-stan-plugin/test/Main.hs#L33

soulomoon:
The ghcide-tests seems to be running faster now, run under 3m50s in ghc9.8

konn (approved):
Thanks a lot! LGTM as for Splice Plugin :+1:

soulomoon:
@fendor I've implement the `TestConfig` you mentioned at https://github.com/haskell/haskell-language-server/pull/4207#discussion_r1597660237.
See if this part look good?

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -176,28 +177,32 @@ setupLSP  recorder getHieDbLoc userHandlers getIdeState clientMsgVar = do
...
+  let doInitialize = handleInit root recorder getHieDbLoc getIdeState reactorLifetime exit clearReqId waitForCancel clientMsgChan
 
   let interpretHandler (env,  st) = LSP.Iso (LSP.runLspT env . flip (runReaderT . unServerM) (clientMsgChan,st)) liftIO
 
   pure (doInitialize, asyncHandlers, interpretHandler)
 
 
 handleInit
-    :: Recorder (WithPriority Log)
+    :: FilePath
```
michaelpj, line 188: ditto

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -176,28 +177,32 @@ setupLSP  recorder getHieDbLoc userHandlers getIdeState clientMsgVar = do
...
     -> LSP.LanguageContextEnv config -> TRequestMessage Method_Initialize -> IO (Either err (LSP.LanguageContextEnv config, IdeState))
-handleInit recorder getHieDbLoc getIdeState lifetime exitClientMsg clearReqId waitForCancel clientMsgChan env (TRequestMessage _ _ m params) = otTracedHandler "Initialize" (show m) $ \sp -> do
+handleInit rootDir recorder getHieDbLoc getIdeState lifetime exitClientMsg clearReqId waitForCancel clientMsgChan env (TRequestMessage _ _ m params) = otTracedHandler "Initialize" (show m) $ \sp -> do
     traceWithSpan sp params
-    let root = LSP.resRootPath env
-    dir <- maybe getCurrentDirectory return root
-    dbLoc <- getHieDbLoc dir
+    let rootMaybe = LSP.resRootPath env
+    -- only shift if lsp root is different from the rootDir
+    when (rootMaybe /= Just rootDir) $ do
```
michaelpj, line 202: I'm a bit confused about what's going on here. Why do we have two sources of root directory? Can't we just use the one from the LSP setup?

michaelpj (commented):
Haven't fully reviewed but I'm a bit confused about what is going on with the root dirs.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
soulomoon: Lsp have a maybe one specified by the client, wrapper and exe could pass in one as the default one.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -714,13 +713,13 @@ loadGhcSession recorder ghcSessionDepsConfig = do
...
     defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GhcSession file -> do
         IdeGhcSession{loadSessionFun} <- useNoFile_ GhcSessionIO
+        -- loading is always returning a absolute path now
         (val,deps) <- liftIO $ loadSessionFun $ fromNormalizedFilePath file
 
         -- add the deps to the Shake graph
         let addDependency fp = do
                 -- VSCode uses absolute paths in its filewatch notifications
-                afp <- liftIO $ makeAbsolute fp
-                let nfp = toNormalizedFilePath' afp
```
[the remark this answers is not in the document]
soulomoon: We absolute them in up stream function. The 'loadSessionFun'

ghcide/src/Development/IDE/LSP/LanguageServer.hs
michaelpj: Okay, so the idea is:
- The wrapper or top-level exe will do `getCurrentDirectory` once, and then pass that around as the root
- If we get a root directory from the client, then we will cd into that directory
- After that, we need to keep around whatever root directory we are using so we can use it to make paths absolute

Is that right? I wonder if we should implement this on the `lsp` side. We could then make it so that the `LspEnv` always contains a root directory, either passed in or taken from the client.

soulomoon: That's right. 
For `lsp` side implmentation.
I suppose that would not make much different even if we pass the one from server to initilized the `LspEnv`. Most of places we use root are in `sessionLoader` and `Rules` system. So the only time we need to query the `LspEnv`'s root is here, when we are doing initialization for shake.

michaelpj: I think maybe this is already not correct. `rootUri` is deprecated in favour of `workspaceFolders`: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders

So in general the server might have several open folders. Which means there is no single answer to "how do I make a path absolute". I hope that means that clients won't send us many relative paths... Perhaps we should instead just stick with our original working directory?

soulomoon: I see,In fact, I would very much like to remove the make a path absolute thing,
The initial goal here is to remove the dependency on `cwd` for our test as much as possible.
So it's more like a temperal thing if we wish to still support `rootUri` and remove the `cwd` lock.

michaelpj: Yes, I'm not really sure what we're supposed to do with the workspace folders. I think maybe the thing to do for now would just be to preserve the "original working directory" and thread it around so we don't need to change directory as you say. So maybe just drop this bit?

soulomoon: Do you mean we drop the `setCurrentDirectory` entirly here?
That's I originally did, but I found we still have some tests that is dependent on shift the cwd to lsp's rootdir. 

The original `setCurrentDirectory` is in `getIdeState`. We used to do the shift no matter the rootdir is comming from defaultmain's arguement or lsp' root. It create a problem for threadings with the shift.

So I move it up here instead of completely dropping it,  it means we only shift if lsp's root and defaultmain's argument's root have conflict. And hence remove the most shift cases in the test.

ghcide/src/Development/IDE/Main.hs
```diff
@@ -326,22 +326,18 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
             logWith recorder Info $ LogLspStart (pluginId <$> ipMap argsHlsPlugins)
 
             ideStateVar <- newEmptyMVar
-            let getIdeState :: LSP.LanguageContextEnv Config -> Maybe FilePath -> WithHieDb -> IndexQueue -> IO IdeState
+            let getIdeState :: LSP.LanguageContextEnv Config -> FilePath -> WithHieDb -> IndexQueue -> IO IdeState
                 getIdeState env rootPath withHieDb hieChan = do
-                  traverse_ IO.setCurrentDirectory rootPath
```
soulomoon, line 331: @michaelpj we used to do the shift here.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -714,13 +713,13 @@ loadGhcSession recorder ghcSessionDepsConfig = do
 
     defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GhcSession file -> do
         IdeGhcSession{loadSessionFun} <- useNoFile_ GhcSessionIO
+        -- loading is always returning a absolute path now
         (val,deps) <- liftIO $ loadSessionFun $ fromNormalizedFilePath file
 
         -- add the deps to the Shake graph
         let addDependency fp = do
                 -- VSCode uses absolute paths in its filewatch notifications
```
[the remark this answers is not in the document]
soulomoon: Let keep it here as the remainder.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
soulomoon: done

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -125,15 +125,16 @@ runLanguageServer recorder options inH outH defaultConfig parseConfig onConfigCh
 
 setupLSP ::
      forall config err.
-     Recorder (WithPriority Log)
+     FilePath -- ^ root directory
+  -> Recorder (WithPriority Log)
```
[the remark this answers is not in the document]
soulomoon: done

0rphee:
Thanks a lot! LGTM as for Stan Plugin 👍

0rphee (approved):

fendor:
Sorry for not reviewing so far! Will do it later today!

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -316,3 +319,10 @@ escapedTextParser = concat <$> P.many (outsideStringLiteral P.<|> stringLiteral)
           inside' = concatMap f inside
 
       pure $ "\"" <> inside' <> "\""
+
+-- ---------------------------------------------------------------------
+
+toAbsolute :: FilePath -> FilePath -> FilePath
+toAbsolute dir file
+    | isAbsolute file = file
+    | otherwise = dir </> file
```
fendor, line 328: If I am not mistaken, this is equivalent to `toAbsolute = (</>)`, see https://hackage.haskell.org/package/filepath-1.5.2.0/docs/System-FilePath.html#v:-60--47--62-

ghcide/src/Development/IDE/Core/Rules.hs
fendor: Perhaps we should also introduce an abstraction that tracks whether we expect a path to be absolute or relative?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -535,6 +535,7 @@ data IdeState = IdeState
     ,shakeExtras          :: ShakeExtras
     ,shakeDatabaseProfile :: ShakeDatabase -> IO (Maybe FilePath)
     ,stopMonitoring       :: IO ()
+    ,rootDir              :: FilePath
```
fendor, line 538: This needs some more documentation. What does this `rootDir` point to? Is it the root of the cradle we found, the argument provided by `--cwd DIR`.

If we want to follow this approach to the extreme, we need a `Note` that explains our high level strategy how we avoid the global variable `$CWD` (queries and modified by `getCurrentDirectory` and `setCurrentDirectory`)

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -127,14 +127,15 @@ runLanguageServer recorder options inH outH defaultConfig parseConfig onConfigCh
 setupLSP ::
      forall config err.
      Recorder (WithPriority Log)
+  -> FilePath -- ^ root directory
```
fendor, line 130: For example, this needs to reference a note, explaining what's going on, why we have it, etc...

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -186,19 +187,22 @@ setupLSP  recorder getHieDbLoc userHandlers getIdeState clientMsgVar = do
...
     -> (SomeLspId -> IO ())
     -> Chan ReactorMessage
     -> LSP.LanguageContextEnv config -> TRequestMessage Method_Initialize -> IO (Either err (LSP.LanguageContextEnv config, IdeState))
-handleInit recorder getHieDbLoc getIdeState lifetime exitClientMsg clearReqId waitForCancel clientMsgChan env (TRequestMessage _ _ m params) = otTracedHandler "Initialize" (show m) $ \sp -> do
+handleInit recorder defaultRoot getHieDbLoc getIdeState lifetime exitClientMsg clearReqId waitForCancel clientMsgChan env (TRequestMessage _ _ m params) = otTracedHandler "Initialize" (show m) $ \sp -> do
     traceWithSpan sp params
-    let root = LSP.resRootPath env
-    dir <- maybe getCurrentDirectory return root
-    dbLoc <- getHieDbLoc dir
+    -- only shift if lsp root is different from the rootDir
```
fendor, line 201: Do we know why this might happen?

ghcide/src/Development/IDE/Types/HscEnvEq.hs
```diff
@@ -59,14 +59,14 @@ updateHscEnvEq oldHscEnvEq newHscEnv = do
...
-newHscEnvEq cradlePath hscEnv0 deps = do
+newHscEnvEq :: FilePath -> FilePath -> HscEnv -> [(UnitId, DynFlags)] -> IO HscEnvEq
+newHscEnvEq root cradlePath hscEnv0 deps = do
     let relativeToCradle = (takeDirectory cradlePath </>)
         hscEnv = removeImportPaths hscEnv0
 
     -- Make Absolute since targets are also absolute
     importPathsCanon <-
-      mapM makeAbsolute $ relativeToCradle <$> importPaths (hsc_dflags hscEnv0)
+      mapM (return . toAbsolute root) $ relativeToCradle <$> importPaths (hsc_dflags hscEnv0)
```
fendor, line 69:
```suggestion
      map (toAbsolute root) $ relativeToCradle <$> importPaths (hsc_dflags hscEnv0)
```

ghcide/test/exe/Config.hs
```diff
@@ -52,37 +56,50 @@ dummyPlugin = mkPluginTestDescriptor (\_ pid -> defaultPluginDescriptor pid "dum
...
+testSession' :: TestName -> (FilePath -> Session ()) -> TestTree
+testSession' name = testCase name . run'
+
+run :: Session a -> IO a
+run = runSessionWithTestConfig def
+    {testDirLocation=Right (mkIdeTestFs []), testPluginDescriptor=dummyPlugin}
+    . const
+
+run' :: (FilePath -> Session a) -> IO a
+run' = runSessionWithTestConfig def {testDirLocation=Right (mkIdeTestFs []), testPluginDescriptor=dummyPlugin}
```
fendor, line 102: Similar suggestion here

hls-test-utils/src/Test/Hls.hs
```diff
@@ -617,87 +614,118 @@ lock = unsafePerformIO newLock
...
-  FilePath ->
-  Session a ->
-  IO a
-runSessionWithServerNoRootLock disableKick pluginsDp conf sconf caps root s =  do
+data TestConfig b = TestConfig
+  {
+    testDirLocation          :: Either FilePath VirtualFileTree
+    -- ^ The file tree to use for the test, either a directory or a virtual file tree
+
+    -- if using a virtual file tree,
```
fendor, line 622:
```suggestion
    -- ^ The file tree to use for the test, either a directory or a virtual file tree
    -- if using a virtual file tree,
```

hls-test-utils/src/Test/Hls.hs
```diff
@@ -617,87 +614,118 @@ lock = unsafePerformIO newLock
...
+    --
+    -- For plugin test logs, look at the documentation of 'mkPluginTestDescriptor'.
+  , testShiftRoot            :: Bool
+    -- ^ Whether to shift the current directory to the root of the project
+  , testDisableKick          :: Bool
+    -- ^ Whether to disable the kick action
+  , testDisableDefaultPlugin :: Bool
+    -- ^ Whether to disable the default plugin comes with ghcide
+  , testCheckProject         :: Bool
+    -- ^ Whether to disable the default plugin comes with ghcide
```
fendor, line 646: Incorrect comment

plugins/hls-notes-plugin/test/NotesTest.hs
```diff
@@ -16,19 +16,20 @@ main = defaultTestRunner $
...
 gotoNoteTests = testGroup "Goto Note Definition"
-    [ testCase "single_file" $ runSessionWithServer def plugin testDataDir $ do
+    [
+      testCase "single_file" $ runSessionWithServer def plugin testDataDir $ do
         doc <- openDoc "NoteDef.hs" "haskell"
         waitForKickDone
         defs <- getDefinitions doc (Position 3 41)
         liftIO $ do
-          fp <- canonicalizePath "NoteDef.hs"
+          fp <- canonicalizePath $ testDataDir </> "NoteDef.hs"
```
fendor, line 25: We should probably add a `Todo` here, is the `canonicalizePath` even necessary?

fendor (changes requested):
Thanks for working on this, this looks extremely promising!
I think what we are especially missing is documentation for why we try to avoid `$CWD` and how we plan to achieve this. So a couple of `Note` might be very helpful!

I do have some nitpicks, though, regarding the coding style. You switch between `, field` and `field, ` a lot, could you standardise this for this PR?
Also, try to keep consistent whitespace, e.g. around `=` in record updates.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
soulomoon: It mgiht be somehow we are starting hls in some folder, and lsp client provide us with the actual project real root.

plugins/hls-notes-plugin/test/NotesTest.hs
soulomoon: Let remove it and pass the the root instead.

hls-plugin-api/src/Ide/PluginUtils.hs
soulomoon: oh right, make it `toAbsolute = (</>)` now, but keep alias `toAbsolute` since it state our intension better

hls-test-utils/src/Test/Hls.hs
soulomoon: fixed

ghcide/test/exe/Config.hs
soulomoon: I am thinking if we can make our formator to auto format record update

plugins/hls-module-name-plugin/src/Ide/Plugin/ModuleName.hs
```diff
@@ -150,7 +151,10 @@ pathModuleNames recorder state normFilePath filePath
       let paths = map (normalise . (<> pure pathSeparator)) srcPaths
       logWith recorder Debug (NormalisedPaths paths)
 
-      mdlPath <- liftIO $ makeAbsolute filePath
+      -- TODO, this can be avoid if the filePath is already absolute,
+      -- we can avoid the toAbsolute call in the future.
+      -- see Note [Root Directory]
+      let mdlPath = (toAbsolute $ rootDir state) filePath
```
soulomoon, line 157: We have rare usage of rootDir, this is one of it.

plugins/hls-retrie-plugin/src/Ide/Plugin/Retrie.hs
```diff
@@ -762,7 +761,7 @@ reuseParsedModule state f = do
 
 getCPPmodule :: Recorder (WithPriority Log) -> IdeState -> HscEnv -> FilePath -> IO (FixityEnv, CPP AnnotatedModule)
 getCPPmodule recorder state session t = do
-    nt <- toNormalizedFilePath' <$> makeAbsolute t
+    let nt = toNormalizedFilePath' $ (toAbsolute $ rootDir state) t
```
soulomoon, line 764: We have rare usage of rootDir, this is one of it.

soulomoon: This one is comming GHC, it should depend on how we load the file into ghc, I think we can drop this if we already loading the file into ghc with absolute path.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
soulomoon: done

ghcide/src/Development/IDE/Core/Rules.hs
soulomoon: Seems like a good idea, some places we might need this.
path from [ghc, hiedb, lsp client]. 

The ones from ghc should depend on how we load the file into `ghc` in sessionloader?

soulomoon: It is quite a mess in our codebase, we might need to open an issue and try to sweep and standardize this.

ghcide/src/Development/IDE/Core/Shake.hs
soulomoon: done

hls-plugin-api/src/Ide/PluginUtils.hs
soulomoon: done

plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
```diff
@@ -155,20 +136,22 @@ semanticTokensConfigTest =
...
-            fs
-            $ do
-              -- modifySemantic funcVar
-              void waitForBuildQueue
-              doc <- openDoc "Hello.hs" "haskell"
-              void waitForBuildQueue
-              result1 <- docLspSemanticTokensString doc
-              liftIO $ unlines (map show result1) @?= "2:1-3 SemanticTokenTypes_Variable \"go\"\n"
+        Test.Hls.runSessionWithTestConfig def {
+          testPluginDescriptor = semanticTokensPlugin
```
fendor, line 140:
```suggestion
        Test.Hls.runSessionWithTestConfig def
          { testPluginDescriptor = semanticTokensPlugin
```

plugins/hls-semantic-tokens-plugin/test/SemanticTokensTest.hs
```diff
@@ -155,20 +136,22 @@ semanticTokensConfigTest =
...
+          , testConfigSession = def {
+            ignoreConfigurationRequests = False
+          }
+          , testConfigCaps = fullCaps
+          , testDirLocation = Right fs
+          , testLspConfig = mkSemanticConfig funcVar
+        }
```
fendor, line 147:
```suggestion
          }
```

fendor (approved):
LGTM, I have a bunch of more nitpicks (feel free to ignore some of them if I start getting on your nerves), but we do need a little bit more documentation for the `Note` (as it is really crucial to give us much information as possible) and when we have to use it.

ghcide/test/exe/DiagnosticTests.hs
```diff
@@ -178,11 +184,11 @@ tests = testGroup "diagnostics"
             [ "module ModuleB where"
             , "import ModuleA ()"
             ]
-      _ <- createDoc (tmpDir `toAbsFp` "ModuleB.hs") "haskell" contentB
-      expectDiagnostics [(tmpDir `toAbsFp` "ModuleB.hs", [(DiagnosticSeverity_Error, (1, 7), "Could not find module")])]
+      _ <- createDoc (tmpDir </> "ModuleB.hs") "haskell" contentB
+      expectDiagnostics [(tmpDir </> "ModuleB.hs", [(DiagnosticSeverity_Error, (1, 7), "Could not find module")])]
```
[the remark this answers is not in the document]
soulomoon: We are unifying functions that actually use a temp dir and the one does not create a temp dir. 
So, `tmpDir` switch to `FilePath` now. I guess in the example we should be using `tmpDir </> "src" </> "Foo.hs
"`

ghcide/test/exe/DependentFileTest.hs
```diff
@@ -15,28 +14,32 @@ import           Language.LSP.Protocol.Types    hiding
...
-import           Test.Tasty
+import           Test.Hls
+
 
 tests :: TestTree
 tests = testGroup "addDependentFile"
-    [testGroup "file-changed" [testWithDummyPluginEmpty' "test" test]
+    [testGroup "file-changed" [testCase "test" $ runSessionWithTestConfig def
+        {testShiftRoot=True
+        , testDirLocation=Right (mkIdeTestFs [])
```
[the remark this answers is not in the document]
soulomoon: I'll do an abstraction over this if I am seeing more occurence of this in other places when migrating tests. Not worth it now. Since it only occurs in one file now.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -915,8 +913,9 @@ newComponentCache
          -> HscEnv             -- ^ An empty HscEnv
          -> [ComponentInfo]    -- ^ New components to be loaded
          -> [ComponentInfo]    -- ^ old, already existing components
+         -> FilePath           -- ^ root dir
```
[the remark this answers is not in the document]
soulomoon: done

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -814,19 +811,20 @@ fromTargetId :: [FilePath]          -- ^ import paths
              -> TargetId
              -> IdeResult HscEnvEq
              -> DependencyInfo
+             -> FilePath
              -> IO [TargetDetails]
 -- For a target module we consider all the import paths
-fromTargetId is exts (GHC.TargetModule modName) env dep = do
+fromTargetId is exts (GHC.TargetModule modName) env dep dir = do
```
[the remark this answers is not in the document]
soulomoon: add comment and issue https://github.com/haskell/haskell-language-server/issues/4257

ghcide/session-loader/Development/IDE/Session.hs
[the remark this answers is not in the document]
soulomoon: done

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -439,4 +459,10 @@ goldenResolveTest testCaseName goldenFilename point hintText =
 
 setupGoldenHlintResolveTest :: TestName -> FilePath -> (TextDocumentIdentifier -> Session ()) -> TestTree
 setupGoldenHlintResolveTest testName path =
-  goldenWithHaskellAndCaps def codeActionResolveCaps hlintPlugin testName testDir path "expected" "hs"
+    goldenWithTestConfig def {
+    testConfigCaps=codeActionResolveCaps
```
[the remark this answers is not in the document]
soulomoon: done

ghcide/src/Development/IDE/Main.hs
```diff
@@ -271,11 +271,11 @@ defaultArguments recorder plugins = Arguments
         }
 
 
-testing :: Recorder (WithPriority Log) -> IdePlugins IdeState -> Arguments
-testing recorder plugins =
+testing :: Recorder (WithPriority Log) -> FilePath -> IdePlugins IdeState -> Arguments
+testing recorder fp plugins =
```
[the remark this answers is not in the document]
soulomoon: I'll rename it to projectRoot so it can explain itself.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -747,7 +744,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} dir = do
         void $ wait as
         asyncRes <- async $ getOptions file
         return (asyncRes, wait asyncRes)
-      pure opts
+      pure $ (fmap . fmap) toAbsolutePath opts
```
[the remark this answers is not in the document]
soulomoon: Great, let's drop it. done.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -1171,8 +1170,8 @@ addUnit unit_str = liftEwM $ do
   putCmdLineState (unit_str : units)
 
 -- | Throws if package flags are unsatisfiable
-setOptions :: GhcMonad m => NormalizedFilePath -> ComponentOptions -> DynFlags -> m (NonEmpty (DynFlags, [GHC.Target]))
-setOptions cfp (ComponentOptions theOpts compRoot _) dflags = do
+setOptions :: GhcMonad m => NormalizedFilePath -> ComponentOptions -> DynFlags -> FilePath -> m (NonEmpty (DynFlags, [GHC.Target]))
+setOptions cfp (ComponentOptions theOpts compRoot _) dflags dir = do
```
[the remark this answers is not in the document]
soulomoon: done

ghcide/src/Development/IDE/Core/Service.hs
```diff
@@ -67,8 +67,9 @@ initialise :: Recorder (WithPriority Log)
            -> WithHieDb
            -> IndexQueue
            -> Monitoring
+           -> FilePath
```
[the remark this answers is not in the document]
soulomoon: done

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -623,11 +640,12 @@ shakeOpen :: Recorder (WithPriority Log)
           -> ShakeOptions
           -> Monitoring
           -> Rules ()
+          -> FilePath
```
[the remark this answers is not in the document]
soulomoon: done

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -527,6 +527,21 @@ newtype ShakeSession = ShakeSession
...
+-- We might be setting it from LSP workspace root > command line > from the current directory.
+--
+-- It helps to remove most usage for `getCurrentDirectory`(After DefaultMain of GhcIde is called),
+-- Using it instead of `getCurrentDirectory` allows us to avoid issues if we `setCurrentDirectory`
+-- somewhere else in the code.
+-- And in turn, it helps with testing in parallel, where we can keep the root directory
+-- and the current directory separate.
+--
+-- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
+-- This is already deprecated and we can drop it in the future when the time comes.
```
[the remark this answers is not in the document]
soulomoon: Indeed done.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -527,6 +527,21 @@ newtype ShakeSession = ShakeSession
...
   }
 
+-- Note [Root Directory]
+-- ~~~~~~~~~~~~~~~~~~~~~
+-- We are keep tracking of the root directory explicitly, which is the directory where the project is located.
+-- We might be setting it from LSP workspace root > command line > from the current directory.
+--
+-- It helps to remove most usage for `getCurrentDirectory`(After DefaultMain of GhcIde is called),
+-- Using it instead of `getCurrentDirectory` allows us to avoid issues if we `setCurrentDirectory`
+-- somewhere else in the code.
```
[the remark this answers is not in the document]
soulomoon: Detail added

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -527,6 +527,30 @@ newtype ShakeSession = ShakeSession
...
+-- Note [Root Directory]
+-- ~~~~~~~~~~~~~~~~~~~~~
+-- We keep track of the root directory explicitly, which is the directory of the project root.
+-- We might be setting it via these options with decreasing priority:
+--
+-- 1. from LSP workspace root, `resRootPath` in `LanguageContextEnv`.
+-- 2. command line (--cwd)
+-- 3. default to the current directory.
+--
+-- use `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
```
fendor, line 539:
```suggestion
-- use `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
```
```suggestion
-- Using `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
```

soulomoon:
Again, thank your for the reviews and suggestions,  expecially since this PR contains a lot of changes, you have done me a great favor to track everything down ! I could not do it without your help.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -527,6 +527,30 @@ newtype ShakeSession = ShakeSession
...
+-- If we modify the global Variable CWD, via `setCurrentDirectory`, all other test threads are suddenly affected,
+-- forcing us to run all integration tests sequentially.
+--
+-- Also, there might be a race condition if we depend on the current directory, as some plugin might change it.
+-- e.g. stylish's `loadConfig`. https://github.com/haskell/haskell-language-server/issues/4234
+--
+-- But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
+-- The root dir is deprecated, but we still need this now,
+-- since a lot of places in the codebase still rely on it.
```
fendor, line 548: Sorry, I don't quite get this, which `root dir` is this now?

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -527,6 +527,30 @@ newtype ShakeSession = ShakeSession
...
+-- We can drop it in the future when the condition meets:
```
fendor, line 549:
```suggestion
-- We can drop it in the future when the condition meets:
```
```suggestion
-- We can drop it in the future once:
```

fendor (approved):
Just some more attention to the `Note`, I still don't quite get the `LSP root` thingy, where does this affect us?

soulomoon:
`LSP root` is deprecated @michaelpj send me the leak, 
I guess that means we should cleanup dependency on the project root(Or $CWD) thing gradually, so muti workspaces can be actually support when we use absolute path everywhere.

michaelpj:
I think we might want to seriously consider using one of the typed path libraries so we can assert that we have absolute paths everywhere. I agree it's just extremely unclear how you are supposed to deal with relative paths of any kind. There's also the issue that some of the tools we call like `cabal` are working directory sensitive, I don't know if we currently call cabal with `--project-file` or whatever we need to get it to run "as if" it were in a specific directory...

fendor:
Since we spawn a new process for `cabal`, we use https://hackage.haskell.org/package/process-1.6.20.0/docs/System-Process.html#t:CreateProcess to set the working directory for cabal, so that is unaffected.

PR #4235, opened 2024-05-16
Provide explicit import in inlay hints

jetjinser:
This PR introduces new feature to provide explicit imports via inlay hints.

`textEdits` are applied when the editor accepts inlay hints (double click in vscode). The effect of applying `textEdits` is to complete imports.This behavior is consistent with code lens.

(For example, in vscode) When the cursor is hover on the label of inlay hints, the tooltip will be displayed as a popup. tooltip is always "Make this import explicit".

Special logic, when the lsp client supports inlay hints, explicit import will be provided in the inlay hints, otherwise it will **fallback** to the code lens.

```haskell
module ExplicitUsualCase where                                                                                                                                               

import ExplicitA

main :: IO ()
main = putStrLn $ "hello " ++ a1
```

```haskell
module ExplicitUsualCase where                                                                                                                                               

import ExplicitA ( a1 ) -- in inlay hints

main :: IO ()
main = putStrLn $ "hello " ++ a1
```


plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -98,9 +102,11 @@ descriptorForModules recorder modFilter plId =
          -- This plugin provides code lenses
            mkPluginHandler SMethod_TextDocumentCodeLens (lensProvider recorder)
         <> mkResolveHandler SMethod_CodeLensResolve (lensResolveProvider recorder)
-          -- This plugin provides code actions
+        -- This plugin provides inlay hints
+        <> mkPluginHandler SMethod_TextDocumentInlayHint (inlayHintProvider recorder)
+        -- <> mkResolveHandler SMethod_InlayHintResolve (inlayHintResolveProvider recorder)
```
michaelpj, line 107: I think just delete this for now, we can worry about resolve later.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -166,6 +173,36 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
    throwError $ PluginInvalidParams (T.pack $ "Unexpected argument for lens resolve handler: " <> show rd)
 
+
+inlayHintProvider :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState 'Method_TextDocumentInlayHint
+inlayHintProvider _ state _ InlayHintParams {_textDocument = TextDocumentIdentifier {_uri}, _range = visibleRange} = do
+    nfp <- getNormalizedFilePathE _uri
+    (ImportActionsResult {forLens, forResolve}, pm) <- runActionE "ImportActions" state $ useWithStaleE ImportActions nfp
+    let inlayHints = [ generateInlayHints newRange ie
+                     | (range, int) <- forLens
+                     , range < visibleRange
```
michaelpj, line 183: I think you want `isSubRangeOf`, I'm not sure this does what you want.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -166,6 +173,36 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+                     , Just newRange <- [toCurrentRange pm range]
+                     , Just ie <- [forResolve IM.!? int]]
+    pure $ InL inlayHints
+  where
+    generateInlayHints :: Range -> ImportEdit -> InlayHint
+    generateInlayHints Range {_end} ie =
+      InlayHint { _position = _end
+                , _label = mkLabel ie
+                , _kind = Nothing
+                , _textEdits = Nothing
```
michaelpj, line 193: We should definitely set this, it's kind of the point!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -444,6 +479,10 @@ abbreviateImportTitle input =
           else actualPrefix <> suffixText
   in title
 
+squashedAbbreviateImportTitle :: T.Text -> T.Text
```
michaelpj, line 482: How is this different to what we do on line 170? Why are we doing different things?

michaelpj: I think I worked it out: the previous thing has the whole import spec in it. I think it would be better to just create the two bits of the import spec separately when we generate the title, then we can use only the pieces we need.

We'd then need to tweak `abbreviateImportTitle` to just abbreviate the list of imported items, and then we can glue it onto the previous part of the import or not depending on whether we're creating an inlay hint or a code lens.

cabal.project
```diff
@@ -16,6 +16,24 @@ benchmarks: True
 
 write-ghc-environment-files: never
 
+source-repository-package
```
michaelpj, line 19: you can use the released version now, just bump the index-state

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -166,6 +173,36 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+    let inlayHints = [ generateInlayHints newRange ie
+                     | (range, int) <- forLens
+                     , range < visibleRange
+                     , Just newRange <- [toCurrentRange pm range]
+                     , Just ie <- [forResolve IM.!? int]]
+    pure $ InL inlayHints
+  where
+    generateInlayHints :: Range -> ImportEdit -> InlayHint
+    generateInlayHints Range {_end} ie =
+      InlayHint { _position = _end
```
michaelpj, line 190: This needs a comment. We are assuming that the appropriate position for the hint to begin is the end of the range for the lens, we should document that carefully.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: God, the spec is weird. Inlay hints themselves only have text edits, but label parts can have commands??? I'm not even sure what you're meant to do there.

michaelpj (commented):
Okay, so the things we definitely need before we can land this:
- Clicking the hints needs to do something. I think that means either including the `textEdits` field or working out how to connect things to a command.
- We need to be able to configure these so you can have one or the other. A simple thing to do would be to add some generic config for "inlay hints on or off", like we have for code lenses etc. Then people at least have some way of configuring it.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jetjinser: Not only that, `abbreviateImportTitle` will retain a lot of indentation (shown as spaces after cutting by line), and inlay hints label will need remove those spaces because they are displayed at the end of each imports statement to avoid being too long.

michaelpj: The thing I was suggesting is that in the `ImportActions` rule we could compute both a) the prefix of the import (`import Foo`) and the suffix (`(bar, baz)`). Then for the code lenses we can concatenate them together and abbreviate, and for the lens we can just use the suffix?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -97,17 +105,35 @@ descriptorForModules recorder modFilter plId =
...
+        -- This plugin provides inlay hints
+        <> mkPluginHandler SMethod_TextDocumentInlayHint (inlayHintProvider recorder)
+        -- This plugin provides code actions
         <> codeActionHandlers
-
     }
 
+isInlayHintsSupported :: MonadIO m => IdeState -> m Bool
+isInlayHintsSupported state = do
+    Shake.ShakeExtras{lspEnv} <- liftIO $ runAction "" state Shake.getShakeExtras
```
michaelpj, line 116: This seems a bit complicated, compare with https://github.com/haskell/haskell-language-server/blob/master/ghcide/src/Development/IDE/Plugin/Completions.hs#L208 (I just searched for "clientCaps"!)

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -129,13 +155,17 @@ runImportCommand _ _ _ rd = do
...
 lensProvider :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState 'Method_TextDocumentCodeLens
-lensProvider _  state _ CodeLensParams {_textDocument = TextDocumentIdentifier {_uri}} = do
-    nfp <- getNormalizedFilePathE _uri
-    (ImportActionsResult{forLens}, pm) <- runActionE "ImportActions" state $ useWithStaleE ImportActions nfp
-    let lens = [ generateLens _uri newRange int
-                | (range, int) <- forLens
-                , Just newRange <- [toCurrentRange pm range]]
-    pure $ InL lens
+lensProvider _ state _ CodeLensParams {_textDocument = TextDocumentIdentifier {_uri}} = do
+    isIHSupported <- liftIO $ isInlayHintsSupported state
```
michaelpj, line 159: We need a note explaining what we're doing here.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -129,13 +155,17 @@ runImportCommand _ _ _ rd = do
...
+    if isIHSupported
+    then do pure $ InR Null
```
michaelpj, line 161: We might be able to return a more useful error here. I think we can maybe use `PluginRequestRefused`? Or maybe it's just the most sensible thing to succeed but return nothing.

plugins/hls-explicit-imports-plugin/test/Main.hs
```diff
@@ -34,10 +34,43 @@ main = defaultTestRunner $ testGroup "import-actions"
...
+           , _data_ = Nothing
+           }]
     , codeActionOnlyGoldenTest "ExplicitOnlyThis" 3 0
     , codeActionOnlyResolveGoldenTest "ExplicitOnlyThis" 3 0
+    , inlayHintsTest "ExplicitOnlyThis" 3 $ (@=?)
+        [InlayHint
+           { _position = Position {_line = 3, _character = 16}
+           , _label = InL "( b1 )"
+           , _kind = Just InlayHintKind_Type
+           , _textEdits = Just [TextEdit (Range (Position 3 0) (Position 3 16)) "import ExplicitB ( b1 )"]
```
michaelpj, line 55: These text edits are arguably overly aggressive. We _could_ just insert the export list where it needs to go, rather than replacing the entire import. That would have the advantage of preserving as much as possible of what the user wrote.

I'm not sure it makes a big difference, but we should think about it and write down why we chose this one. I think in general the expectation is that an inlay hint inserts just the text in the hint...

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -165,6 +196,50 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+    --                  |^-padding
+    --                  ^-_position
+    generateInlayHints :: Range -> ImportEdit -> PositionMapping -> InlayHint
+    generateInlayHints (Range _ end) ie pm =
+      InlayHint { _position = end
+                , _label = InL $ mkLabel ie
+                , _kind = Just InlayHintKind_Type -- for type annotations
+                , _textEdits = fmap singleton $ toTEdit pm ie
+                , _tooltip = Nothing
+                , _paddingLeft = Just True
```
michaelpj, line 232: Did you figure out what these padding fields do? It has been very unclear to me form the spec!

plugins/hls-explicit-imports-plugin/test/Main.hs
```diff
@@ -34,10 +34,43 @@ main = defaultTestRunner $ testGroup "import-actions"
...
+           , _kind = Just InlayHintKind_Type
+           , _textEdits = Just [TextEdit (Range (Position 2 0) (Position 2 16)) "import ExplicitA ( a1 )"]
+           , _tooltip = Nothing
+           , _paddingLeft = Just True
+           , _paddingRight = Nothing
+           , _data_ = Nothing
+           }]
     , codeActionOnlyGoldenTest "ExplicitOnlyThis" 3 0
     , codeActionOnlyResolveGoldenTest "ExplicitOnlyThis" 3 0
+    , inlayHintsTest "ExplicitOnlyThis" 3 $ (@=?)
```
michaelpj, line 50: We should test that we _don't_ get code actions when we have inlay hints enabled and vice versa.

michaelpj (commented):
This is looking generally good! It's also quite non-obvious to me why the failing test is failing, it doesn't look like you've changed anything there. I'll think about it. Maybe @fendor has an idea?

plugins/hls-explicit-imports-plugin/test/Main.hs
jetjinser: Yes, append is what I did first, but I did run into some problems, like this code:
```haskell
import           Development.IDE                      hiding (pluginHandlers,
                                                       pluginRules)
```

If we just append it, it will obviously break the import statement.
So I chose the same behavior as the original code lens & code action, replacing the entire import statement, so that users only need to re-format it.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jetjinser: In neovim and vscode, it means:

with padding
```haskell
import Development.IDE.Graph.Classes  ( Hashable )
```

withoug padding
```haskell
import Development.IDE.Graph.Classes ( Hashable )
```

fendor:
Which test is failing unexpectedly? `ExplicitBreakFile code action (golden)`?

fendor:
I debugged this to the point that replacing `isInlayHintsSupported` with `pure False` makes the test succeed, but `isInlayHintsSupported` in the `lensProvider` makes it crash.
I don't understand this issue, yet.

fendor:
@jetjinser Change the client capabilities of `codeActionBreakFile` to use `codeActionNoInlayHintsCaps`, that worked for me locally.

However, the error message itself is extremely puzzeling.

jetjinser:
> @jetjinser Change the client capabilities of `codeActionBreakFile` to use `codeActionNoInlayHintsCaps`, that worked for me locally.
> 
> However, the error message itself is extremely puzzeling.

Yes, this works on my machine too.

Should I apply this modification? I'll add a comment pointing to the discussion here.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -97,17 +105,25 @@ descriptorForModules recorder modFilter plId =
...
+        <> mkPluginHandler SMethod_TextDocumentInlayHint (inlayHintProvider recorder)
+        -- This plugin provides code actions
         <> codeActionHandlers
-
     }
 
+isInlayHintsSupported :: Applicative f => IdeState -> f Bool
+isInlayHintsSupported ideState = do
+  let clientCaps = Shake.clientCapabilities $ shakeExtras ideState
+  pure $ case clientCaps of
```
michaelpj, line 117: Could do
```
isJust $ clientCaps ^? textDocument . _Just . inlayHint . _Just
```

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -129,13 +145,19 @@ runImportCommand _ _ _ rd = do
...
-    let lens = [ generateLens _uri newRange int
-                | (range, int) <- forLens
-                , Just newRange <- [toCurrentRange pm range]]
-    pure $ InL lens
+lensProvider _ state _ CodeLensParams {_textDocument = TextDocumentIdentifier {_uri}} = do
+    -- Code lens are not provided when the client supports inlay hints,
+    -- otherwise it will be provided as a fallback
+    isIHSupported <- liftIO $ isInlayHintsSupported state
+    if isIHSupported
+    then do pure $ InL []
```
michaelpj, line 153: Comment about why this is the right thing to return here

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -165,6 +188,52 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+        (ImportActionsResult {forLens, forResolve}, pm) <- runActionE "ImportActions" state $ useWithStaleE ImportActions nfp
+        let inlayHints = [ generateInlayHints newRange ie pm
+                         | (range, int) <- forLens
+                         , Just newRange <- [toCurrentRange pm range]
+                         , isSubrangeOf newRange visibleRange
+                         , Just ie <- [forResolve IM.!? int]]
+        pure $ InL inlayHints
+    -- When the client does not support inlay hints, fallback to the code lens,
+    -- so this is Null
+    else do pure $ InR Null
```
michaelpj, line 211: Why do we return null here but an empty list of code lenses above? It helps to be consistent!

Also, we should add: if the client doesn't support inlay hints, this method should never get called!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: Okay, so it basically means "show an extra space before the inlay hint. I guess that makes sense.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -165,6 +188,52 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+                , _textEdits = fmap singleton $ toTEdit pm ie
+                , _tooltip = Nothing
+                , _paddingLeft = Just True
+                , _paddingRight = Nothing
+                , _data_ = Nothing
+                }
+    mkLabel :: ImportEdit -> T.Text
+    mkLabel (ImportEdit{ieResType, ieText}) =
+      let title ExplicitImport = abbreviateImportTitle . T.dropWhile (/= '(') $ ieText
+          title RefineImport   = "Refine imports to " <> T.intercalate ", " (T.lines ieText)
```
michaelpj, line 233: This seems odd. Do we have any tests showing this? These perhaps _shouldn't_ be shown as inlay hints? I'm unsure.

michaelpj: It's possible these need to always be code lenses, since they don't represent implicit content that can be added to the document by accepting it, which is pretty much what inlay hints are supposed to be. Maybe I'm wrong though, not sure.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -165,6 +188,52 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+    -- is the end of the range for the code lens.
+    --   import Data.Char (isSpace)
+    --   |--- range ----|-- IH ---|
+    --                  |^-paddingLeft
+    --                  ^-_position
+    generateInlayHints :: Range -> ImportEdit -> PositionMapping -> InlayHint
+    generateInlayHints (Range _ end) ie pm =
+      InlayHint { _position = end
+                , _label = InL $ mkLabel ie
+                , _kind = Just InlayHintKind_Type -- for type annotations
```
michaelpj, line 223: This is surely the wrong kind. I think this should just be omitted.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -165,6 +188,52 @@ lensResolveProvider _ ideState plId cl uri rd@(ResolveOne _ uid) = do
...
+                , _textEdits = fmap singleton $ toTEdit pm ie
+                , _tooltip = Nothing
```
michaelpj, line 225: I think we should definitely consider using this! For example, the tooltip for the minimal imports inlay hint could say "This is an import list for exactly the names used by the current module", or something. That way it's a bit more discoverable what it means. It's a nice UX feature!

michaelpj (commented):
I do think we still need a way to turn these off at the plugin level, like we have for code lenses etc. See https://github.com/haskell/haskell-language-server/blob/master/hls-plugin-api/src/Ide/Types.hs#L262

plugins/hls-explicit-imports-plugin/test/Main.hs
michaelpj: Interesting! How does the inlay hint display in that case? I guess here the "perfect" thing to do would be to append the import list and also delete any "hiding" clauses. Complicated, though.

michaelpj: Also, needs a comment, and maybe a test?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jetjinser: I'm trying to figure out what's the difference between the two, and from this comment (https://github.com/microsoft/language-server-protocol/issues/1200#issuecomment-779094756), it seems like `[]` is no different from `null`.

But we do need to uniformly choose a way to indicate “no information”.  Do you think Right `[]` is better?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jetjinser: Could this be as simple as "Make this import explicit"? This doesn't even require resolve, even though the LSP spec recommends that the tooltip be obtained via resolve, but it's simple enough.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jetjinser: Well, I was going to support refine with inlay hints, it may looks like:

`import qualified RefineA <inlay>from RefineB</inlay> as RA`

or:
`import RefineD <inlay>from RefineE</inlay> <inlay>( e2 )</inlay>`

But this is a little tricky for me, I'm trying to locate where the module name position is. In addition, I don't know if I understand it wrong, there may be _many_ original modules? This makes inlay hints longer and jagged.
So I want to ask your opinion.

plugins/hls-explicit-imports-plugin/test/Main.hs
jetjinser: > I guess here the "perfect" thing to do would be to append the import list and also delete any "hiding" clauses. Complicated, though.

The current approach is to _append inlay hints directly_ at the end.
I don't think it's possible to have inlay hints "cover" a part of your code.

> Also, needs a comment, and maybe a test?

Okay, I'll add both.

michaelpj: >  I don't think it's possible to have inlay hints "cover" a part of your code.

Right, that's definitely true for the hint. I meant that the _text edits_ could only overwrite the hiding clause rather than the entire line.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: Yes, that seems reasonable to me.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: Yes, I think it would be pretty sensible to leave refine as a code lens and not an inlay hint. Let's do that for now and maybe open an issue to discuss what the best UX would be for refine imports?

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
michaelpj: Yes, I think either of those would be fine. I think we can just make it a constant string; the hint itself shows the things that are going to be inserted. So no need for resolve. Resolve is only for things that are expensive to compute.

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
```diff
@@ -129,13 +145,19 @@ runImportCommand _ _ _ rd = do
...
-    let lens = [ generateLens _uri newRange int
-                | (range, int) <- forLens
-                , Just newRange <- [toCurrentRange pm range]]
-    pure $ InL lens
+lensProvider _ state _ CodeLensParams {_textDocument = TextDocumentIdentifier {_uri}} = do
+    -- Code lens are not provided when the client supports inlay hints,
+    -- otherwise it will be provided as a fallback
+    if isInlayHintsSupported state
+    -- `[]` is no different from `null`, we chose to use all `[]` to indicate "no information"
+    then pure $ InL []
```
michaelpj, line 153: Okay, so I think the only thing remaining here is that we were going to leave the refine import actions as code lenses, so we need to return them in both cases. And never return them for inlay hints.

michaelpj (commented):
One last thing then I think this is good to go!

plugins/hls-explicit-imports-plugin/src/Ide/Plugin/ExplicitImports.hs
jetjinser: Should this be checked at lens resolve? If I want to know whether `ImportEdit` is explicit or refine, it seems that can't do it in `lensProvider`.

If I can't tell whether the `lensProvider` should respond at the time, then every import statement actually needs to be resolved even that is not refine. I'm worried this will rely on the expensive resolve.

michaelpj: Hmmm, that's annoying. I would have thought we could determine what type it is beforehand? Perhaps we need to tweak the rule to include that information?

jetjinser: I think now should be okay.

michaelpj:
Great stuff, let's merge it :tada:

michaelpj (approved):

PR #4238, opened 2024-05-17
Fix resultBuilt(dirty mechanism) in hls-graph

### LINKED ISSUE #4237: Clarify the dirty mechanism of hls-graph
comment should added to specify how the hls-graph dirty mechanism is working.

soulomoon:
Fix  #4237
`resultBuilt` might updated to an older value, which make a rule result that should be clean back to dirty. 
It might trigger un-needed recomputation. Fixing it could minimize the recomputation further.


hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -179,14 +182,23 @@ compute db@Database{..} stack key mode result = do
     deps <- newIORef UnknownDeps
     (execution, RunResult{..}) <-
         duration $ runReaderT (fromAction act) $ SAction db deps stack
-    built <- readTVarIO databaseStep
+    curStep <- readTVarIO databaseStep
     deps <- readIORef deps
-    let changed = if runChanged == ChangedRecomputeDiff then built else maybe built resultChanged result
-        built' = if runChanged /= ChangedNothing then built else changed
```
soulomoon, line 185: set the new build time to `changed` might update the build time when we should not

hls-graph/src/Development/IDE/Graph/Internal/Types.hs
```diff
@@ -187,7 +187,6 @@ instance NFData RunMode where rnf x = x `seq` ()
 -- | How the output of a rule has changed.
 data RunChanged
     = ChangedNothing -- ^ Nothing has changed.
-    | ChangedStore -- ^ The stored value has changed, but in a way that should be considered identical (used rarely).
```
soulomoon, line 190: I do not think it is ever used for would be used, we could added back if there is a need for this.

michaelpj:
Do we have any tests for this mechanism? `hls-graph` is reasonably stand-alone, I would hope we could write a test for this?

wz1000:
this seems to make sense. how did you discover this?

soulomoon:
Actually, I've discovered the `Changed` and `Built` is somewhat off when fixing the drity problem, but I forget about it at the time.
I am reviewing the code and it come back to me.

Yep, I'll see if some test could be added.

soulomoon:
I've added two test, one from the lower layer and one from higher layer to demonstate the bug and ensure no regression.

michaelpj (approved):

michaelpj:
I'd like @wz1000 to check the tests make sense just to be sure!

michaelpj:
nag @wz1000

soulomoon:
Let's merge it if no one object to this. I've been using this branch and no regression yet.

wz1000 (approved):

Issue #717, opened 2019-12-15
Hover does not work for type symbols from external packages


jacg:
... as documented by the tests titled *type constructor from other package*, *value from other package* and *external class is signature*.

To be precise, hover *does* work for external values, but not types.

jneira:
Test in hls are in:

https://github.com/haskell/haskell-language-server/blob/d4f5d436641efdca15f68f1d435901b5057e351d/ghcide/test/exe/Main.hs#L2487-L2488

https://github.com/haskell/haskell-language-server/blob/d4f5d436641efdca15f68f1d435901b5057e351d/ghcide/test/exe/Main.hs#L2517-L2518

Anton-Latukha:
I believe it works now.

jneira:
Afaik the broken mark in tests should make the fail if they are fixed. Although maybe they fail for another reason.
In other similar issue i removed the broken mark to check if the test continue failing for the same reason.

jacg:
I regret that I don't have the time to work on this any more. :disappointed: 

> Afaik the broken mark in tests should make the test fail if they are fixed.

Yes, that was the original intention when I wrote these tests. IIRC, it did actually work like that, at least at the beginning.

jneira:
>  Although maybe they fail for another reason.

It happened to me for the `top level signature`/`inner signature` (they are handled in another issue)
I removed the broken mark and both continue failing obviusly. But visually `top-level signature` was fixed!
The test was failing due to the hover position, slightly different. Fixing hover position in the test made it green.

hasufell:
I'm not sure if this ticket is still relevant. There are other tickets for goto definition like https://github.com/haskell/haskell-language-server/issues/708

Feel free to re-open this ticket with a clarification.

soulomoon:
We should enable the test for it too.

PR #4241, opened 2024-05-17
Enable test for #717

soulomoon:
I think we can enable the test for #717  now. Not sure why it won't work before, but it should work now


jhrcek (approved):
LGTM

soulomoon:
thanks for the review

Issue #4251, opened 2024-05-26
Unify critical session running in hls


soulomoon:
There are some places we serialize critical session.
* Reactor thread, running the notification handler, by running it in the reactor thread.
* sessionLoader, we load session one by one, spawn a thread each time, and wait for the preivous thread to end.
* sessionRestart, we runs it in seperate threads but using an `MVar` to do the serlization.

The latter two create a challenge for our shutdown handler to halt the hls. Since they are not cancelable.
A possible solution, would be to switch the latter two to the first approach. Create a thread specifically to run the target critical session.

soulomoon:
This might fix the hanging problem happend in windows
https://github.com/haskell/haskell-language-server/pull/2497

michaelpj:
We did this

PR #4256, opened 2024-05-27
Unify critical session running in hls 

soulomoon:
see #4251
What's done

1. add `Development.IDE.Core.Thread` to provide common api to run a woker thread
2. rewrite `runWithDb` using the `Development.IDE.Core.Thread`
3. add session loading thread and session restart thread using `Development.IDE.Core.Thread`
4. send session loading runs to session loading thread
5. send session restart runs to session restart thread

The above modification ensure the shutdown handler can cancel the session loading and session restart. And hence we would have a more correct shutdown behaviour. Potentially make the test less likely to hang.


ghcide/src/Development/IDE/Core/Thread.hs
```diff
@@ -0,0 +1,67 @@
...
+            -> IO ()
+}
+
+-- | runWithThread
+-- Run a long running action with a additional running thread
+-- The additional thread will serialize runs of the actions from the TQueue.
+-- Return ContT to run the action
+runWithThread :: ThreadRun input workerResource resource arg -> input -> ContT () IO (resource, TQueue arg)
+runWithThread ThreadRun{..} ip = ContT $ \f -> do
+    tRunWithResource ip $ \w r -> do
```
michaelpj, line 48: I guess this kind of looks like two separate things to me:
1. A domain-specific `withResource` which creates the resource and passes it to a continuation.
2. A standard "worker queue"

My instinct is that we could focus on defining the worker queue, and just leave the creation of the resource, where necessary, to the call site. Indeed, many of the call sites don't create a resource at all! I think we could then quite naturally write:
```
withMyResource $ \resource -> withWorkerQueue (\arg -> doSomething resource arg) $ \queue -> ...
```
or something?

soulomoon: Make sense, let's implement `withWorkerQueue` instead

ghcide/src/Development/IDE/Core/Thread.hs
```diff
@@ -0,0 +1,67 @@
...
+    where
+        writerThread r q =
+            forever $ do
+                l <- atomically $ readTQueue q
+                tWorker ip r l
+
+
+-- | blockRunInThread run and wait for the result
+-- Take an action from TQueue, run it and
+-- use barrier to wait for the result
```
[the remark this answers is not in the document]
soulomoon: done

ghcide/src/Development/IDE/Core/WorkerThread.hs
```diff
@@ -0,0 +1,46 @@
...
+--
+-- Originally we used various ways to implement this, but it was hard to maintain and error prone.
+-- Moreover, we can not stop these threads uniformly when we are shutting down the server.
+--
+-- `Development.IDE.Core.WorkerThread` module provides a simple api to implement this easily.
+-- * `withWorkerQueue`: accepts an action to run in separate thread and returns a `TQueue` to send the actions to run.
+-- * `blockRunInThread` : accepts a `TQueue` and an action to run in separate thread and waits for the result.
+
+
+-- | withWorkerQueue creates a new TQueue and runs the workerAction in a separate thread.
```
michaelpj, line 26:
```suggestion
-- | 'withWorkerQueue' creates a new 'TQueue', and launches a worker
-- thread which polls the queue for requests and runs the given worker 
-- function on them. 
```

ghcide/src/Development/IDE/Core/WorkerThread.hs
```diff
@@ -0,0 +1,46 @@
...
+withWorkerQueue workerAction = ContT $ \mainAction -> do
+    q <- newTQueueIO
+    withAsync (writerThread q) $ \_ -> mainAction q
+    where
+        writerThread q =
+            forever $ do
+                l <- atomically $ readTQueue q
+                workerAction l
+
+-- | blockRunInThread run and wait for the result
```
michaelpj, line 37:
```suggestion
-- | 'blockRunInThread' queues up an 'IO' action to be run by a worker thread,
-- and then blocks until the result is computed.
```

ghcide/src/Development/IDE/Core/WorkerThread.hs
```diff
@@ -0,0 +1,46 @@
...
+ where
+
+import           Control.Concurrent.Async
+import           Control.Concurrent.STM
+import           Control.Concurrent.Strict (newBarrier, signalBarrier,
+                                            waitBarrier)
+import           Control.Monad             (forever)
+import           Control.Monad.Cont        (ContT (ContT))
+
+-- Note [Serializing runs in separate thread]
```
michaelpj, line 12: I do recommend using multiline comments for stuff like this, it's nicer!

ghcide/src/Development/IDE/Core/WorkerThread.hs
```diff
@@ -0,0 +1,46 @@
...
+    q <- newTQueueIO
+    withAsync (writerThread q) $ \_ -> mainAction q
+    where
+        writerThread q =
+            forever $ do
+                l <- atomically $ readTQueue q
+                workerAction l
+
+-- | blockRunInThread run and wait for the result
+blockRunInThread :: TQueue (IO ()) -> IO result -> IO result
```
michaelpj, line 38: Subjective, but I think we more often use "await" rather than "block" in this sort of context. So maybe call it `awaitRunInThread`?

michaelpj (approved):
This looks better, makes sense to me.

I do wonder: if we are using `blockRunInThread`, then it looks like it's pretty much identical to having a resource that's guarded by a lock, and then having the waiting threads block on acquiring the lock. If we put work into the queue and the move on so it can be executed asynchronously, then it is different. But if we immediately block and wait then there's no real effect of it happening in another thread!

ghcide/src/Development/IDE/Core/WorkerThread.hs
soulomoon: Yep

ghcide/src/Development/IDE/Core/WorkerThread.hs
soulomoon: good idea done

soulomoon:
>  it's pretty much identical to having a resource that's guarded by a lock

It remind me of something, we might probably want to limit some of the queue size to 1, so the waiting to enqueue action can be cancelled when we are doing shakeRestart.

PR #4262, opened 2024-05-27
add back withLongTimeout to CradleTests

soulomoon:
In previous migrate of `CradleTests`, removal of `withLongTimeout` seems to be causing trouble to windows.
Add it back fix that.


michaelpj:
Please add a comment explaining why it's there for future people!

soulomoon:
Yeah, I'll continue this if there is more timeout can be observed in the CI

soulomoon:
I've done some tests and I am not sure if it is just a timeout problem.
This test is fairly broken in windows. I've created an issue for this.

PR #4264, opened 2024-05-28
Remove unused and outdated CHANGELOG files

fendor:
The CHANGELOG files in plugins are neither up-to-date nor can they even be displayed in Hackage (since plugins are all internal libraries and Hackage). I also don't think they are displayed in flora.

Thus, delete them, so users aren't confused why the ChangeLogs contain barely any information.

Thanks to @Kleidukos, who kindly told me that we had CHANGELOGs :)


jhrcek (approved):
LGTM

PR #4271, opened 2024-05-29
CI change, only run bench on performance label

soulomoon:
We already have the `performance` label

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -18,6 +18,7 @@ jobs:
   pre_job:
     runs-on: ubuntu-latest
     outputs:
+      if: contains(github.event.pull_request.labels.*.name, 'performance')
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
     steps:
       - id: skip_check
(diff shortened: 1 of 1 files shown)


.github/workflows/bench.yml
```diff
@@ -17,6 +17,7 @@ on:
 jobs:
   pre_job:
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'performance')
```
soulomoon, line 20: I am not quite sure if this is enough ?

fendor:
CI looks like it skipped the benchmark job.

To we want to make nightly runs of the benchmarks perhaps? Ideally, there would be an easy to access them later.

soulomoon:
CI already skipped bench on some occasions.

And we almost never looking at them unless we touch performance related code.

I think for now, if we touch performance related, manually adding the performance label should be enough.

But we can still add the nightly runs, if someone wants it. I want to know which branch would the nightly runs be running, for all PR or just the mater?

fendor (approved):
LGTM

PR #4272, opened 2024-06-01
Delete testUtil from ghcide-tests

soulomoon:
After a serious of migration https://github.com/haskell/haskell-language-server/issues/4173 
Only `NonLspCommandLine` use `TestUtil`, we can delete `TestUtil` now
And some more cleanup of unused function in ghcide-tests


fendor (approved):
Nice!

PR #4273, opened 2024-06-02
Fix flaky test by a test env setup refactor

### LINKED ISSUE #4200: ghcide test `IfaceTests`, we should remove interface cache dir before we run the test, or the diganostic simply gone
This can happened in the second round doing the test.
## Reproduce 
Run this consectively
```
LSP_TEST_LOG_MESSAGES=1 LSP_TEST_LOG_STDERR=1  TASTY_PATTERN="iface-error-test-1" cabal test ghcide-tests
```

we are seeing 
```
Debug | SUCCEEDED LOADING HIE FILE FOR /Users/ares/.cache/ghcide/main-67f5d3e80e9a5dab967334c225648bdcd080465f-67f5d3e80e9a5dab967334c225648bdcd080465f/P.hie
```

The test past again if we remove the cache dir entirely

komikat:
Fixes #4200 by creating a separate temporary `.cache` directory for every test instead of having all tests share the same `.cache` directory. This makes sure the test `iface-error-test-1` is run with a clean directory — causing the diagnostic to be generated every time.


soulomoon:
It seems `hls-refactor-plugin-tests` depend on the cache directory.🤔

fendor:
That seems impossible to me, there must be another race condition :D

komikat:
Weird because `hls-refactor-plugin-tests` seems to pass on my system (MacOS, ghc9.8)

komikat:
The setup wasn't creating the `hls-test-root` directory. 
`ghcide-tests` is still failing, will look into this soon.

fendor:
This is good! 

Let's merge this as is, add the comment to the test case, create an issue containing the knowledge you have gained so far, and then accept the test case changes. E.g., update the assertions in the test. Even though we have discovered an interesting race (?) condition, the test case itself will be stable.

@komikat Could you do that, please?

fendor (approved):
Please add a comment to the test case that fails now and create a new tracking issue! LGTM otherwise

soulomoon:
> ### definition.Imported symbol (reexported) fails without `$TEMPDIR/hls-test-root`
> running `HLS_TEST_HARNESS_NO_TESTDIR_CLEANUP=1 HLS_TEST_LOG_STDERR=1 TASTY_PATTERN="/definition.Imported symbol (reexported)/" cabal test ghcide-tests` FAILS (>=GHC94)
> 
> ```
> expected: Location {_uri = Uri {getUri = "$TMPDIR/hls-test-root/extra-dir-85406152527/Bar.hs"}, _range = Range {_start = Position {_line = 3, _character = 5}, _end = Position {_line = 3, _character = 8}}}
>  but got: Location {_uri = Uri {getUri = "$TMPDIR/hls-test-root/extra-dir-85406152527/Bar.hs"}, _range = Range {_start = Position {_line = 3, _character = 0}, _end = Position {_line = 3, _character = 14}}}
> ```
> 
> ### Notes
> * AtPoint.hs.gotoDefinition calls locationsAtPoint
> * it is using nameToLocation when failing
> * `itExists <- liftIO $ doesFileExist fs` returns true
> * doesn't seem to be falling back to db
> * in `srcSpanToLocation sp`
>   `rng <- srcSpanToRange src`
>   reports a different location from what's expected in the test
> * the `sp` comes from
>   `nameSrcSpan name`
>   in that branch.

This is a bug in our code, It worth a seperate issue to fix this.

PR #4274, opened 2024-06-03
Delete library ghcide test utils

soulomoon:
Finally we can get rid of library ghcide test util!


fendor (approved):
:confetti_ball: A job well done, congratulations on untangling this ball of spaghetti! Much appreciated!

PR #4284, opened 2024-06-08
FindImports: ThisPkg means some home unit, not "this" unit

wz1000:
This bug prevents package imports from working within a multi unit session.


michaelpj:
What happened before? A crash?

wz1000:
It just wasn't able to resolve those imports, because it tried to look from them in the current unit instead of *some* home unit.

soulomoon (approved):
LGTM

michaelpj:
I guess I mean did the test fail before this? but I assume so

PR #4286, opened 2024-06-08
Add common stanza to completion data

### LINKED ISSUE #4285: Cabal file completion does not consider common stanzas
Fields and their values are not completed correctly when the cursor is inside the common stanza of a cabal file.

Field completion suggests fields from previous stanza instead of common stanza:
image

VeryMilkyJoe:
This allows fields and values to be completed correctly inside common stanzas
Fixes #4285.


fendor (approved):

fendor (changes requested):
LGTM, thanks for looking into this! There is no test for the completions in common sections, yet, could you perhaps add one?

VeryMilkyJoe:
> LGTM, thanks for looking into this! There is no test for the completions in common sections, yet, could you perhaps add one?

Will do, good catch!

michaelpj (approved):

PR #1914, opened 2021-06-13
Limit the number of valid hole fits

expipiplus1:
Originally https://github.com/haskell/ghcide/pull/888

<a href="https://gitpod.io/#https://github.com/haskell/haskell-language-server/pull/1914"></a>

PR #4288, opened 2024-06-08
Limit number of valid hole fits to 10

### LINKED ISSUE #4198: Applying a code action to a _ hole produces too many suggestions
I'm often using code actions to import out-of-scope identifiers. But sometimes I accidentally run a code action on a line that contains both an out-of-scope identifier _and_ an underscore, such as:

```haskell
TwoWay.twoWay (zipDefnsWith (alignWith _) undefined) conflicts
                             ^^^^^^^^^
                                this is out of scope, and I want HLS to suggest
                                `import Data.Semialign (alignWith)` with a code action.
```

In this case, the `_` on the same line will cause 1800 (!) suggestions that look like

```
1: replace _ with foo
2: replace _ with bar
...
```

for (I guess?) every single identifier in scope.

akshaymankar:
Fixes #4198
Supersedes #1914

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Compat.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat.hs
@@ -133,6 +133,7 @@ import           Compat.HieTypes                         hiding
                                                          (nodeAnnotations)
 import qualified Compat.HieTypes                         as GHC (nodeAnnotations)
 import           Compat.HieUtils
+import           Control.Applicative                     ((<|>))
 import qualified Data.ByteString                         as BS
 import           Data.Coerce                             (coerce)
 import           Data.List                               (foldl')
@@ -434,7 +435,7 @@ setHieDir _f d = d { hieDir = Just _f}
 dontWriteHieFiles :: DynFlags -> DynFlags
 dontWriteHieFiles d = gopt_unset d Opt_WriteHie
 
-setUpTypedHoles ::DynFlags -> DynFlags
+setUpTypedHoles :: DynFlags -> DynFlags
 setUpTypedHoles df
   = flip gopt_unset Opt_AbstractRefHoleFits    -- too spammy
   $ flip gopt_unset Opt_ShowDocsOfHoleFits     -- not used
@@ -447,9 +448,13 @@ setUpTypedHoles df
   $ flip gopt_unset Opt_SortValidHoleFits
   $ flip gopt_unset Opt_UnclutterValidHoleFits
   $ df
-  { refLevelHoleFits = Just 1   -- becomes slow at higher levels
-  , maxRefHoleFits   = Just 10  -- quantity does not impact speed
-  , maxValidHoleFits = Nothing  -- quantity does not impact speed
+  { refLevelHoleFits = refLevelHoleFits df <|> Just 1   -- becomes slow at higher levels
+
+   -- Sometimes GHC can emit a lot of hole fits, this causes editors to be slow
+   -- or just crash, we limit the hole fits to 10. The number was chosen
+   -- arbirtarily by the author.
+  , maxRefHoleFits   = maxRefHoleFits df <|> Just 10
+  , maxValidHoleFits = maxValidHoleFits df <|> Just 10
   }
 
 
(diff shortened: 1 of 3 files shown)


ghcide/src/Development/IDE/GHC/Compat.hs
```diff
@@ -449,7 +449,7 @@ setUpTypedHoles df
   $ df
   { refLevelHoleFits = Just 1   -- becomes slow at higher levels
   , maxRefHoleFits   = Just 10  -- quantity does not impact speed
-  , maxValidHoleFits = Nothing  -- quantity does not impact speed
+  , maxValidHoleFits = Just 10  -- quantity does not impact speed
```
michaelpj, line 452: outdated comment also on the previous line

Can we have a comment explaining the choice of the magic number? "I made this up" is fine, it indicates we don't have a good reason for it, which is helpful information for the future

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -69,7 +69,8 @@ processHoleSuggestions mm = (holeSuggestions, refSuggestions)
             (mrAfter . (=~ t " *Valid (hole fits|substitutions) include"))
             validHolesSection
       let holeFit = T.strip $ T.takeWhile (/= ':') holeFitLine
-      guard (not $ T.null holeFit)
+      guard $ not $ holeFit =~ t "Some hole fits suppressed"
```
michaelpj, line 72: What's going on here? Why are we doing this?

ghcide/src/Development/IDE/GHC/Compat.hs
michaelpj: Is this the version we want? I don't really know.

ghcide/src/Development/IDE/GHC/Compat.hs
```diff
@@ -449,7 +449,7 @@ setUpTypedHoles df
   $ df
   { refLevelHoleFits = Just 1   -- becomes slow at higher levels
```
michaelpj, line 450: So, I think these things are also user-configurable. Generally it's nice if we don't overwrite user settings if they've set them. So can we adjust these so that they use the value from `df` if it's not `Nothing` and otherwise use these values? I actually think that's what the old PR did?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
akshaymankar: GHC prints this line where we expect a hole fit to be when the number of hole fits is limited. This line just ignores that. We do this in the refinement hole fits too.

ghcide/src/Development/IDE/GHC/Compat.hs
akshaymankar: Fixed the comments.

ghcide/src/Development/IDE/GHC/Compat.hs
akshaymankar: Now respecting previous values like the other PR.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
michaelpj: Write it down!

michaelpj:
Genuine test failure in `hls-refactor-plugin`, it's looking for a hole-fit suggestion code action that presumably now gets filtered out. Not sure quite how we want to fix this - we could change the test to look for one that does get selected, or maybe we should just increase the number of fits to 20 or something.

akshaymankar:
> Genuine test failure in `hls-refactor-plugin`, it's looking for a hole-fit suggestion code action that presumably now gets filtered out. Not sure quite how we want to fix this - we could change the test to look for one that does get selected, or maybe we should just increase the number of fits to 20 or something.

Just replaced the test with a type which has fewer hole-fits. For me getting 20 code suggestions for a hole feels a bit overwhelming (10 is already pushing it honestly), but its just a feeling.

michaelpj:
Great, happy to take your read on the UX.

michaelpj (approved):

Tritlo:
Hmm. Does this then now not inform the users when there are hole-fits found but are suppressed by this setting? That seems confusing and counterintuitive, though I might be biased.

michaelpj:
This doesn't change how the normal diagnostic from GHC is produced, I think.

akshaymankar:
We do get this info from GHC that some hole-fits are suppressed. I don't know what would be a good way to communicate this info.

michaelpj:
If you mean e.g. in offering completions, I don't think we have a good way of conveying that information...

PR #4294, opened 2024-06-08
Call useWithStale instead of useWithStaleFast when calling ParseCabalFields

### LINKED ISSUE #4293: Shown cabal completions are wrong when queried too quickly after a change in the file
When prompting completions directly after changing the cabal file, incorrect completions are shown.

This is the result of `useWithStaleFast` which is used when calling `CabalParseFields` as the results are queried before the computation has actually finished.

VeryMilkyJoe:
Also adds some documentation about why we use each of these versions of `use`.
Fixes #4293.


michaelpj:
This doesn't seem right to me. I thought completions were one of the typical examples where we definitely want fast response speed at the expense of stale output?

fendor:
In general true, but `readFields` is incredibly fast as it is basically a single pass over the file contents.
In blow testing, this felt like the better default beaviour..

fendor (approved):

PR #4296, opened 2024-06-08
Make workerqueue cancellable if not started already configurable

soulomoon:


soulomoon:
cc @michaelpj , take a look. it should improive the semantic of the workerQueue

michaelpj:
Okay, now I really am suspicious. Session loading uses `awaitRunInQueue` and the queue can only have one thing in it. That really _is_ just a lock! I think it might just make sense to use a different mechanism there, and keep the worker queue for cases where we really do have a queue of things we want to do asynchronously.

I'm also sceptical about the restart queue. Are we really going to have multiple restarts queued up? And if that can be seen as just needing to take a lock, well, that's already there in the form of the `withMVar` call.

soulomoon:
restart queue should queue up, since we do not want the restart to get lost if the thread waiting to do it is cancelled. 

Session loading is a different case, since it is called by the rule system. The rule system would handle the restart case. 

But we would want them to run in a separate thread. 
1. Session loading need to run in separate thread one way or another since the thread calling it might be cancelled anytime. Exit session loading would cause problem.
2. handle the shutdown correctly

soulomoon:
Two more thing have been done.
* An action is only removed from the queue after it is done running.
* add waitUntilWorkerQueueEmpty for the test

soulomoon:
Not relevant anymore and superpass by https://github.com/haskell/haskell-language-server/pull/4445

PR #4298, opened 2024-06-09
Try turning off optimization for testing

michaelpj:
I want to see if this cuts down the time we spend building in CI a bit.


michaelpj:
eh, this doesn't even reliably make a difference for me locally

PR #4303, opened 2024-06-09
Test Rope.getLine

jhrcek:
Just testing out newly introduced utility from text-rope

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -26,6 +26,11 @@ test-options: -j1
 package *
   ghc-options: -haddock
 
+source-repository-package
+  type:git
+  location: https://github.com/jhrcek/text-rope.git
+  tag: 37e808b26c8e8c862a04c985663fefd7c64335e2 
+
 constraints:
   -- C++ is hard to distribute, especially on older GHCs
   -- See https://github.com/haskell/haskell-language-server/issues/3822
@@ -41,3 +46,4 @@ constraints:
   -- We want to be able to benefit from the performance optimisations
   -- in the future, thus: TODO: remove this flag.
   bitvec -simd
+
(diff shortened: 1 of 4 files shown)


michaelpj:
We also do this a bunch in `lsp`, I think.

jhrcek:
> We also do this a bunch in lsp, I think.
Hmm, in lsp there's some [extractLine](https://github.com/haskell/lsp/blob/dfdff1380b747dcde47d497cee8772775a963957/lsp/src/Language/LSP/VFS.hs#L405) 
 in LSP which returns a Rope and does further manipulation on that Rope, which means that the new function I introduced is probably useless. Or do you think that the function I introduced should return a Rope?

michaelpj:
Hmm, that's a little annoying. We return a `Rope` so we can use the mixed encoding indexing which `text-rope` provides, but conceptually it's just a one-line `Text`. I don't know if there are comparable mixed-indexing functions for plain `Text`? If so we can switch to use your `getLine`.

michaelpj:
Note that it's quite likely that functions in HLS also need this facility. Remember that much of HLS incorrectly assumes that it can use LSP positions as GHC positions or as indexes into `Text`. So e.g. [this line](https://github.com/haskell/haskell-language-server/pull/4303/files#diff-d3f4379ba2a24de34059ba42f4c4ec61f373218f8d77274b06eed536824c2613R911) is wrong: it needs to drop `c` _UTF-16 code units_, whereas `Text.drop` drops _code units_. 

So HLS might also prefer to get a `Rope`. Or it should use `rangeLinesFromVFS` from `lsp`, which does this correctly.

jhrcek:
@michaelpj I opened a followup PR in text-rope that makes getLine stay in the Rope world. Would it make more sense to you like that? https://github.com/Bodigrim/text-rope/pull/7

Issue #4057, opened 2024-02-07
Handlers should run in a custom monad that doesn't expose stuff that people shouldn't use


michaelpj:
Today our handlers have access to all of `LspM`, which lets them:

- Access the VFS and the configuration
   - This is incorrect - plugin handlers should be accessing the snapshots of these that we have in the shake graph
- Send arbitrary messages
   - This makes things harder to test, we can't just look at the handler return value to see the main thing that the handler "does"

We should probably run in a custom monad that is a wrapper around `LspM` that restricts some stuff, and maybe also change our handler types to mostly return their results.

awjchen:
I'd like to work on this!

PR #4304, opened 2024-06-09
Use restricted monad for plugins (#4057)

awjchen:
This PR defines new monad, `PluginM`, a newtype wrapper around the `LspM` monad, and substitutes it for `LspM` in all plugin-related code. This is an initial step towards https://github.com/haskell/haskell-language-server/issues/4057 that refactors code but does not change the capabilities of the plugin interface.


hls-plugin-api/src/Ide/Types.hs
```diff
@@ -890,9 +893,52 @@ instance GEq IdeNotification where
...
+-- | Restricted version of 'LspM' specific to plugins
+newtype PluginM config a = PluginM { _runPluginM :: LspM config a }
+  deriving newtype (Applicative, Functor, Monad, MonadIO, MonadUnliftIO)
+
+runPluginM :: PluginM config a -> LspM config a
+runPluginM = _runPluginM
+
+-- | Wrapper of 'getVirtualFile' for PluginM
+--
+-- TODO: To be replaced by a lookup of the Shake build graph
```
michaelpj, line 905: YES! This is exactly the sort of thing we want to spot, excellent.

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -890,9 +893,52 @@ instance GEq IdeNotification where
 instance GCompare IdeNotification where
   gcompare (IdeNotification a) (IdeNotification b) = gcompare a b
 
+-- | Restricted version of 'LspM' specific to plugins
+newtype PluginM config a = PluginM { _runPluginM :: LspM config a }
```
michaelpj, line 897: Suggest `HandlerM`? since it's the monad that's appropriate for handlers?

hls-plugin-api/src/Ide/Types.hs
```diff
@@ -890,9 +893,52 @@ instance GEq IdeNotification where
 instance GCompare IdeNotification where
   gcompare (IdeNotification a) (IdeNotification b) = gcompare a b
 
+-- | Restricted version of 'LspM' specific to plugins
```
michaelpj, line 896: Worth explaining why this is here! Something like:

We use this instead of `LspM`, since there are parts of the LSP server state which
plugins should not access directly, but instead only via the build system.

michaelpj (approved):
Looks like very much the right direction?

hls-plugin-api/src/Ide/Types.hs
awjchen: Renamed

hls-plugin-api/src/Ide/Types.hs
awjchen: I've added your explanation to the comment!

michaelpj:
Needs stylish-haskell run on things

PR #4305, opened 2024-06-09
Add completions for import statements in cabal files

VeryMilkyJoe:
Currently any common stanzas in the cabal file will be suggested as values for an import field.
In the future, we should only suggest names of common stanzas, defined before the cursor position to adhere to the cabal specification.


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Completer/Simple.hs
```diff
@@ -41,6 +45,22 @@ constantCompleter completions _ cData = do
...
+importCompleter :: Completer
+importCompleter l cData = do
+  cabalFieldsM <- getCabalFields cData
+  case cabalFieldsM of
+    Just cabalFields -> do
+        let commonNames = mapMaybe (\case
+              Syntax.Section (Syntax.Name _ "common") commonNames _ -> getOptionalSectionName commonNames
+              _ -> Nothing)
+              cabalFields
+        constantCompleter commonNames l cData
```
fendor, line 61: I think this should be potentially a rule to avoid expensive recomputations.

fendor (changes requested):

Issue #3246, opened 2022-10-05
Link from error codes to the error message index


david-christiansen:
## Is your enhancement request related to a problem? Please describe.

Some GHC error messages and warnings can be confusing.

## Describe the solution you'd like

The Haskell Foundation has started a project to assign unique numbers to each error or warning emitted by GHC, and a community-driven web site that associates documentation with each code. The codes will be part of GHC 9.6. I'd like to see HLS include links to this documentation from messages in some convenient way.

## Describe alternatives you've considered

None.

## Additional context

All codes emitted by GHC match `GHC-(0-9){5}`. The message index will have addresses of the form `https://errors.haskell.org/messages/GHC-NNNNN`.

michaelpj:
Need to come back to this when we've got basic 9.6 support.

michaelpj:
We can actually put this in the `Diagnostic`: it has a `code` field and a `codeDescription` field that's a URI, so perfect.

michaelpj:
I think we can do this already without structured diagnostics: we can just regex it out of the diagnostic message easily enough.

michaelpj:
I have some code for this, but annoyingly we don't actually render the error codes already! So it's confusing what to do - do we render them (thus making users see them in the messages), or not (and not be able to pull them out)... :thinking:

michaelpj:
Hacky code here: https://github.com/haskell/haskell-language-server/tree/mpj/error-index

michaelpj:
Hmmm. I feel like I am going in the wrong direction here by smuggling additional information into messages only to parse it out later. Perhaps the right thing to do is just to implement the structured diagnostics stuff :thinking:

dylan-thinnes:
Working on this for ZuriHac

xsebek:
It seems that this was indeed successfully implemented two Zurihacks prior and this issue could be closed? 🙂 

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/e7c8840c-557a-40dd-8600-a1bf4720c238" />

<sup>The [GHC-83865](https://errors.haskell.org/messages/GHC-83865) link is clickable!</sup>

PR #4311, opened 2024-06-10
Support structured diagnostics

dylan-thinnes:
Addresses https://github.com/haskell/haskell-language-server/issues/2014

Should hopefully enable https://github.com/haskell/haskell-language-server/issues/3246

TODOs:
- [x] ~~Resolve TODOs in code~~
- [x] ~~Add diagnostic codes to more tests~~
- [x] ~~Make PR on GHC side so that we can eventually remove `ghcide/src/Development/IDE/GHC/Compat/Driver.hs`~~
  - Issue https://gitlab.haskell.org/ghc/ghc/-/issues/24996 and MR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12891
- [ ] Deal with CPP to make this work properly on all versions
- [x] Fix `hls-plugin-api` using `_code` field set by `withWarnings` to determine which warning to use.

Additional work (future PRs):
- Replacing regexes with structured diagnostic checks
- Improve typing so that it can be known that an error has a structured diagnostic, instead of defaulting to `Maybe`


ghcide/test/exe/CradleTests.hs
```diff
@@ -210,7 +210,7 @@ sessionDepsArePickedUp = testWithDummyPluginEmpty'
         "cradle: {direct: {arguments: []}}"
     -- Open without OverloadedStrings and expect an error.
     doc <- createDoc "Foo.hs" "haskell" fooContent
-    expectDiagnostics [("Foo.hs", [(DiagnosticSeverity_Error, (3, 6), "Couldn't match type")])]
+    expectDiagnostics [("Foo.hs", [(DiagnosticSeverity_Error, (3, 6), "Couldn't match type", Nothing)])]
```
michaelpj, line 213: I was expecting at least some of these to now have codes?

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -106,10 +181,10 @@ prettyDiagnostics :: [FileDiagnostic] -> Doc Terminal.AnsiStyle
 prettyDiagnostics = vcat . map prettyDiagnostic
 
 prettyDiagnostic :: FileDiagnostic -> Doc Terminal.AnsiStyle
-prettyDiagnostic (fp, sh, LSP.Diagnostic{..}) =
+prettyDiagnostic FileDiagnostic { fdFilePath, fdShouldShowDiagnostic, fdLspDiagnostic = LSP.Diagnostic{..} } =
     vcat
-        [ slabel_ "File:    " $ pretty (fromNormalizedFilePath fp)
-        , slabel_ "Hidden:  " $ if sh == ShowDiag then "no" else "yes"
+        [ slabel_ "File:    " $ pretty (fromNormalizedFilePath fdFilePath)
+        , slabel_ "Hidden:  " $ if fdShouldShowDiagnostic == ShowDiag then "no" else "yes"
```
michaelpj, line 187: I think we should print the code from the LSP diagnostic here also!

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -86,7 +125,43 @@ instance NFData ShowDiagnostic where
...
+  compare NoStructuredMessage NoStructuredMessage = EQ
+  compare SomeStructuredMessage {} SomeStructuredMessage {} = EQ
+  compare NoStructuredMessage SomeStructuredMessage {} = GT
+  compare SomeStructuredMessage {} NoStructuredMessage = LT
+
+instance NFData StructuredMessage where
+  rnf NoStructuredMessage = ()
+  rnf SomeStructuredMessage {} = ()
+
+modifyFdLspDiagnostic :: (Diagnostic -> Diagnostic) -> FileDiagnostic -> FileDiagnostic
```
michaelpj, line 162: We use lenses a lot, this seems like a really obvious place for a lens! Just generate lenses for the `FileDiagnostic` type.

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -86,7 +125,43 @@ instance NFData ShowDiagnostic where
...
+  { fdFilePath :: NormalizedFilePath
+  , fdShouldShowDiagnostic :: ShowDiagnostic
+  , fdLspDiagnostic :: Diagnostic
+  , fdStructuredMessage :: StructuredMessage
+  }
+  deriving (Eq, Ord, Show, Generic)
+
+instance NFData FileDiagnostic
+
+data StructuredMessage
```
michaelpj, line 138: This is really only a GHC-specific message. Is the idea that we might expand it to other types with some `Typeable` wizardry or something? If so, write it down!

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -86,7 +125,43 @@ instance NFData ShowDiagnostic where
...
+  deriving (Generic)
+
+instance Show StructuredMessage where
+  show NoStructuredMessage = "NoStructuredMessage"
+  show SomeStructuredMessage {} = "SomeStructuredMessage"
+
+instance Eq StructuredMessage where
+  (==) NoStructuredMessage NoStructuredMessage = True
+  (==) SomeStructuredMessage {} SomeStructuredMessage {} = True
+  (==) _ _ = False
```
michaelpj, line 150: Are these instances safe? Do we need them? Worth a comment.

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +56,53 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
...
-    _data_ = Nothing
-    })
+  -> Maybe (MsgEnvelope GhcMessage)
+  -> FileDiagnostic
+ideErrorWithSource source sev fdFilePath msg origMsg =
+  let lspDiagnostic =
+        LSP.Diagnostic {
+          _range = noRange,
+          _severity = sev,
+          _code = Nothing,
```
michaelpj, line 96: set this I guess

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +56,53 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
...
+          _source = source,
+          _message = msg,
+          _relatedInformation = Nothing,
+          _tags = Nothing,
+          _codeDescription = Nothing,
```
michaelpj, line 101: and this

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +56,53 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
...
+
+ideErrorFromLspDiag
+  :: LSP.Diagnostic
+  -> NormalizedFilePath
+  -> Maybe (MsgEnvelope GhcMessage)
+  -> FileDiagnostic
+ideErrorFromLspDiag lspDiag fdFilePath origMsg =
+  let fdShouldShowDiagnostic = ShowDiag
+      fdStructuredMessage =
+        maybe NoStructuredMessage SomeStructuredMessage origMsg
```
michaelpj, line 71: prefer `case`

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +56,53 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
...
+      fdLspDiagnostic = lspDiag
+        { _code = fmap ghcCodeToLspCode . diagnosticCode . errMsgDiagnostic =<< origMsg
```
michaelpj, line 73: Hmm, we set it here but not below? confusing. Maybe these should share code?

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +56,53 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
...
+        }
+      ghcCodeToLspCode :: DiagnosticCode -> Int32 LSP.|? T.Text
```
michaelpj, line 75: This probably needs a pre-structured errors path too...

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +56,53 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
 -- | an IdeResult with a fingerprint
 type IdeResultNoDiagnosticsEarlyCutoff  v = (Maybe ByteString, Maybe v)
 
-ideErrorText :: NormalizedFilePath -> T.Text -> FileDiagnostic
-ideErrorText = ideErrorWithSource (Just "compiler") (Just DiagnosticSeverity_Error)
+ideErrorText :: Maybe (MsgEnvelope GhcMessage) -> NormalizedFilePath -> T.Text -> FileDiagnostic
```
michaelpj, line 59: This is now a somewhat odd function, since if we have the GhcMessage we could get the text from it...

ghcide/src/Development/IDE/GHC/Warnings.hs
```diff
@@ -33,15 +33,12 @@ withWarnings diagSource action = do
   warnings <- newVar []
   let newAction :: DynFlags -> LogActionCompat
       newAction dynFlags logFlags wr _ loc prUnqual msg = do
-        let wr_d = map ((wr,) . third3 (attachReason wr)) $ diagFromErrMsg diagSource dynFlags $ mkWarnMsg dynFlags wr logFlags loc prUnqual msg
+        let wr_d = map ((wr,) . modifyFdLspDiagnostic (attachReason wr)) $ diagFromSDocErrMsg diagSource dynFlags (mkWarnMsg dynFlags wr logFlags loc prUnqual msg)
         modifyVar_ warnings $ return . (wr_d:)
       newLogger env = pushLogHook (const (logActionCompat (newAction (hsc_dflags env)))) (hsc_logger env)
```
michaelpj, line 38: This whole code is rather suspect. I thought you told me you came up with a way to avoid this entirely?

ghcide/src/Development/IDE/GHC/Warnings.hs
```diff
@@ -33,15 +33,12 @@ withWarnings diagSource action = do
   warnings <- newVar []
   let newAction :: DynFlags -> LogActionCompat
       newAction dynFlags logFlags wr _ loc prUnqual msg = do
-        let wr_d = map ((wr,) . third3 (attachReason wr)) $ diagFromErrMsg diagSource dynFlags $ mkWarnMsg dynFlags wr logFlags loc prUnqual msg
+        let wr_d = map ((wr,) . modifyFdLspDiagnostic (attachReason wr)) $ diagFromSDocErrMsg diagSource dynFlags (mkWarnMsg dynFlags wr logFlags loc prUnqual msg)
```
michaelpj, line 36: Note that `attachReason` arguably does the wrong thing here: it prints the warning _flag_ into the code, which seems wrong.

ghcide/src/Development/IDE/GHC/Error.hs
```diff
@@ -51,30 +55,36 @@ import           Language.LSP.VFS                  (CodePointPosition (CodePoint
...
-    , _severity = Just sev
-    , _source   = Just diagSource -- not shown in the IDE, but useful for ghcide developers
-    , _message  = msg
-    , _code     = Nothing
-    , _relatedInformation = Nothing
-    , _tags     = Nothing
-    , _codeDescription = Nothing
-    , _data_   = Nothing
-    }
+diagFromText :: T.Text -> D.DiagnosticSeverity -> SrcSpan -> T.Text -> Maybe (MsgEnvelope GhcMessage) -> FileDiagnostic
```
michaelpj, line 58: This module is a real twisty maze of similar functions :sweat_smile: I don't have a good suggestion at the moment but it would be nice if there are fewer, since at the moment I don't feel confident that all of them are even doing the right thing. Perhaps we could rely on fewer functions even if it means making some of the call sites more verbose? I'm not sure.

ghcide/src/Development/IDE/GHC/Compat/Driver.hs
```diff
@@ -0,0 +1,125 @@
+-- This module copies parts of the driver code in GHC.Main.Driver to provide
```
michaelpj, line 1: Make an upstream issue to expose this and link to it here so we know when we can get rid of this

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1442,7 +1442,7 @@ getAllDiagnostics ::
     STMDiagnosticStore ->
     STM [FileDiagnostic]
 getAllDiagnostics =
-    fmap (concatMap (\(k,v) -> map (fromUri k,ShowDiag,) $ getDiagnosticsFromStore v)) . ListT.toList . STM.listT
+    fmap (concatMap (\(k,v) -> map (\diag -> ideErrorFromLspDiag diag (fromUri k) Nothing) $ getDiagnosticsFromStore v)) . ListT.toList . STM.listT -- TODO: Do we need the structured message here?
```
michaelpj, line 1445: Hmm, this is potentially bad? i.e. do we just throw away the extra info when we put stuff into here? Honestly I have no idea how this thing fits into the diagnostics provided by rules..

michaelpj: i.e. I wonder if this is going to bite us when we try to match on the structured error in a handler, maybe it will have been thrown away by this...

michaelpj (commented):
Possibly insane suggestion: could we write some orphan JSON instances for GHC's message types and stuff the `GhcMessage` into the `data` field on the `Diagnostic`? Then we don't have to carry anything else around, and we won't lose things going via the `DiagnosticStore`?

ghcide/test/exe/CradleTests.hs
dylan-thinnes: Yeah, that was a TODO in the PR's opening comment - fixed now in c4d5eddb and 4b452c98

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Fixed in 25b02fa6

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Fixed in 12b364ae

ghcide/src/Development/IDE/GHC/Warnings.hs
dylan-thinnes: This code is no longer used for anything in the typechecking, but it is still used for bytecode/object code generation, as well as a few other places.

I suspect some of these functions (e.g. codegen) will need deeper changes to be able to get diagnostics as a list, though I don't have great evidence for that atm. I haven't taken a look to see if those functions that are wrapped with this could produce diagnostics another way. I'll take a look another day.

That said, I prefer to defer this to another PR anyways, if at all possible. The current approach gives us structured diagnostics for renaming and typechecking, and doesn't require us to copy too much code from GHC or make any deeper changes, and lets us get started with the bulk of the useful plugin work.

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Fixed in 25bfcb3a

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Fixed in a46f0a26

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: The idea behind `StructuredMessage` is that it's a thin wrapper around `GhcMessage` that pretends the type has Show, Eq, Ord, Generic instances. This allows us to derive those typeclasses. The fact that it isn't a law-abiding typeclass is not important since it's just metadata in `FileDiagnostic`.

Documented in 003c15c1

This is a bit cursed so I'm leaving this open - wdyt?

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Documented in 003c15c1 , further discussion in https://github.com/haskell/haskell-language-server/pull/4311#discussion_r1641871002

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: `CodeDescription` is (misleadingly) a URI that should link to more advice - I don't think there's anything we can reasonably populate it with here?

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: It gets set via `ideErrorFromLspDiag`, which is the underlying function that `ideErrorWithSource` calls out to.

I agree that the code path here is a bit strange, but I can't really decide on a better separation of concerns.

ghcide/src/Development/IDE/Core/Shake.hs
dylan-thinnes: Fixed in 2cd33dfc

ghcide/src/Development/IDE/GHC/Compat/Driver.hs
dylan-thinnes: This was a TODO "Make PR on GHC side so that we can eventually remove ghcide/src/Development/IDE/GHC/Compat/Driver.hs" - done now :)

Issue https://gitlab.haskell.org/ghc/ghc/-/issues/24996 and MR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12891

dylan-thinnes: Added to the source code as well in 3500ac3e

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: The diagnostic generating functions are all a bit of a pile (in the Enrico Fermi sense of the word) right now - let's discuss refactoring that in https://github.com/haskell/haskell-language-server/pull/4311#discussion_r1636130917

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Continued in https://github.com/haskell/haskell-language-server/pull/4311#discussion_r1636130917

ghcide/src/Development/IDE/GHC/Error.hs
dylan-thinnes: I've been thinking the same thing as I went along. The functions in `Development.IDE.Types.Diagnostics` and `Development.IDE.GHC.Error` are a mess. Lots of them restate information or pass in information twice that could be derived from a single piece of state.

That said, fixing that up feels like a relatively low priority - the current functions are a bit more complicated than they need to be, but ultimately don't take *too* long to grok IMO, and are a minimum of changes necessary to get the system working without changing many call sites too much.

Could we push to a ticket, another PR? Considering I'm the most likely candidate for doing this, I'm alright with that.

ghcide/src/Development/IDE/GHC/Warnings.hs
dylan-thinnes: That was a source of a bit of confusion when I got started testing my changes, I forgot that it's technically a bug - I've dropped it in 4fed9877, I doubt anyone relied on this.

That said, there are a few other places where I see the `_code` field being used weirdly.

ghcide/src/Development/IDE/GHC/Error.hs
dylan-thinnes: (In the meantime, reverted)

dylan-thinnes: Looking deeper, `attachReason` was introduced in f03a7fa55d909c864771f0989abace8793b5b15a, explicitly for code actions that would disable a warning.

If we reimplement the code action in some other way, then we should be good.

dylan-thinnes:
Re: deriving JSON instances, the messages often contain fragments of AST and other things which aren't terribly friendly to ToJSON/FromJSON. I tried deriving an instance and immediately came up against

```
No instance for ‘ToJSON
                         (Language.Haskell.Syntax.Binds.HsBindLR
                            GHC.Hs.Extension.GhcRn GHC.Hs.Extension.GhcRn)’
```

which makes me think this approach is not going to be easy.

dylan-thinnes:
Gonna take a break for today, but two big TODOs remain, and a few open questions.

ghcide/src/Development/IDE/GHC/Warnings.hs
dylan-thinnes: (In the meantime, reverted)

dylan-thinnes: Looking deeper, `attachReason` was introduced in f03a7fa55d909c864771f0989abace8793b5b15a, explicitly for code actions that would disable a warning.

If we reimplement the code action in some other way, then we should be good.

michaelpj:
> Re: deriving JSON instances, the messages often contain fragments of AST and other things which aren't terribly friendly to ToJSON/FromJSON.

This tickled my brain and I remembered that GHC has its own JSON conversion functions! https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Utils/Json.hs?ref_type=heads

And they definitely have instances for diagnostics, since they're used in the JSON output for GHC, see e.g. here: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11093

Having written this I realise that that only handles one direction, so it doesn't help us much, oh well.

michaelpj:
They might have another serialization format, though. There's nothing stopping us from serializing them to bytes and the smuggling a bytestring as a JSON string in `data` :joy:

ghcide/src/Development/IDE/GHC/Error.hs
michaelpj: Fine with me. It's just hard for me to review since I'm failing to grok them D:

ghcide/src/Development/IDE/GHC/Warnings.hs
michaelpj: Yes, that all sounds right. Ideally we'd have the structured diagnostics and then we'd be able to do this directly, in the mean time using `_data` seems better.

ghcide/src/Development/IDE/GHC/Warnings.hs
michaelpj: Fine with me. If you learned something about it, maybe write it down in a note so we don't forget?

ghcide/src/Development/IDE/Types/Diagnostics.hs
michaelpj: We can link to the error index!

ghcide/src/Development/IDE/Types/Diagnostics.hs
michaelpj: I guess my question is: why do we need those instances? And if we need them for a good reason, area we sure the use is safe if the instances are busted? I think the answer is "yes it's okay, because the LSP diagnostic should always 'determine' the `StructuredMessage`, so we can safely ignore it", but we should write that reasoning down!

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: We can only link to the error index if it has that error in the index, which is currently a bit over 100 (probably 200 once the PRs get merged). I've been putting a little merge request together on the HEI site for generating a machine-readable metadata.json that we can then request once and cache each time the language server launches.

Since that dance will likely take some time, I've created an issue - could you assign it to me?

Issue here: https://github.com/haskell/haskell-language-server/issues/4327
MR for HEI: <incoming>

michaelpj: I was actually thinking we could just construct the link anyway. If you link to an error that isn't documented yet, you just get a page saying it's not documented yet. That's what GHC does, it now just blindly links there!

ghcide/src/Development/IDE/GHC/Warnings.hs
dylan-thinnes: I've made a note be37756d

wz1000:
I don't think we should serialise the raw data over the wire, instead we should stash it in some sort of `Map` and send over an identifier/key which we can later use to look up the value from the `Map`. It requires care to ensure that the values are properly garbage collected when no longer valid though.

michaelpj:
> I don't think we should serialise the raw data over the wire, instead we should stash it in some sort of Map and send over an identifier/key which we can later use to look up the value from the Map. It requires care to ensure that the values are properly garbage collected when no longer valid though.

There's no viable serialization format anyway. I do think it would be nice if there _was_ one, though.

That said, I don't think we _should_ need this info in diagnostics that we e.g. get back from the client. So the main issue is when we e.g. ask for the diagnostics covering a range so we can then match on them to work out code actions. At _that_ point we need the GHC diagnostics. So I don't think we need a separate store, at worst we might have to augment our server-side diagnostic store.

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: I'd rather push back on that flow, even if it is what GHC already does - the user who clicks that link is already likely very confused, adding confusion by pointing them to places that don't exist is worse and feels unfriendly.

In GHC I understand that the added complexity is not worth the cost, but HLS should be more user-friendly and helpful.

ghcide/src/Development/IDE/GHC/Warnings.hs
dylan-thinnes: Alright, I've solved this in 7bb82d0b

ghcide/src/Development/IDE/Types/Diagnostics.hs
dylan-thinnes: Finally made some time to commit up my proof of concept of how I think the Haskell Error Index to CodeDescription could work - 414c845a

I'm open to criticism here:
- I can defer this to another PR or cut it out completely if you think this is too much for now
- Currently this PR would generate `dailyHaskellUsers * dailyInvocationsOfHLS * apiResponseSize` bytes of traffic every day. I didn't look very hard for a decent caching location - if there's some way we can persist this info to disk so we only make one request every week that may be a good shout.
- Deserializing and serializing the HEI API should probably be managed by HEI - I can make a PR to turn the HaskellErrorIndex module into a package in the HEI repo instead iyl
- Maybe we do want a link on every error, though I do feel strongly that this would confuse novices who need it most. The HEI is not necessarily very complete, even on errors I'd consider common.

WDYT?

dylan-thinnes:
One remaining point to address from review: https://github.com/haskell/haskell-language-server/pull/4311#discussion_r1651757991

I'm looking into the CI failures now - I assume some CPP shenanigans are at least part of the issue.

ghcide/src/Development/IDE/Types/Diagnostics.hs
michaelpj: I think at least let's do this in another PR, this one is complex enough already!

I continue to think that just unconditionally linking to the error index is pretty reasonable: in particular, since that's the behaviour that GHC has implemented, there's reasonable pressure on the error index to present something sensible to users in that case. Plus, it's decent amount of complexity to avoid that. We can also just see if people report being confused!

dylan-thinnes: After having seen how much this has sprawled out, I'm happy to go with that - I also expected that the code URL would be more prominent, but in reality it looks something like this:



which is unassuming enough.

I'll go ahead to bundle everything related to that into its own PR.

noughtmare:
@dylan-thinnes are you still planning to work on this? The immediate CI issue seems very simple to solve. It seems there is just a redundant import. What else still needs to happen?

noughtmare:
Removing these four redundant imports fixes all issues for me locally:

```diff
diff --git a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
index 5425c419..a0cea4dc 100644
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
@@ -13,8 +13,7 @@ where
 
 import           Control.Lens                      ((&), (.~))
 import qualified Data.Text                         as T
-import           Development.IDE                   (FileDiagnostic,
-                                                    ShowDiagnostic (ShowDiag))
+import           Development.IDE                   (FileDiagnostic)
 import           Development.IDE.Types.Diagnostics (fdLspDiagnosticL,
                                                     ideErrorWithSource)
 import           Distribution.Fields               (showPError, showPWarning)
diff --git a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
index c9ce440f..35f41ea3 100644
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
@@ -87,7 +87,6 @@ import           Language.LSP.Protocol.Types                       (ApplyWorkspa
                                                                     CodeActionKind (CodeActionKind_QuickFix),
                                                                     CodeActionParams (CodeActionParams),
                                                                     Command,
-                                                                    Diagnostic (..),
                                                                     MessageType (..),
                                                                     Null (Null),
                                                                     ShowMessageParams (..),
diff --git a/plugins/hls-refactor-plugin/test/Main.hs b/plugins/hls-refactor-plugin/test/Main.hs
index 377a6758..434fe48b 100644
--- a/plugins/hls-refactor-plugin/test/Main.hs
+++ b/plugins/hls-refactor-plugin/test/Main.hs
@@ -21,7 +21,6 @@ import           Data.Foldable
 import           Data.List.Extra
 import           Data.Maybe
 import qualified Data.Text                                as T
-import           Data.Tuple.Extra
 import           Development.IDE.GHC.Util
 import           Development.IDE.Plugin.Completions.Types (extendImportCommandId)
 import           Development.IDE.Test
diff --git a/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs b/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
index 1fc7fa42..51e00550 100644
--- a/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
+++ b/plugins/hls-stan-plugin/src/Ide/Plugin/Stan.hs
@@ -14,7 +14,6 @@ import qualified Data.Text                         as T
 import           Development.IDE
 import           Development.IDE.Core.Rules        (getHieFile)
 import qualified Development.IDE.Core.Shake        as Shake
-import           Development.IDE.Types.Diagnostics
 import           GHC.Generics                      (Generic)
 import           Ide.Plugin.Config                 (PluginConfig (..))
 import           Ide.Types                         (PluginDescriptor (..),
```

fendor:
Closing in favour of #4433

Issue #3628, opened 2023-06-08
Document symbols could include local definitions


michaelpj:
At the moment we include all the "top-level" stuff. But we _could_ include local definitions as well. The thing that would be awkward is reflecting nested scopes, but often they are also associated with a symbol.

It might also be rather noisy, but arguably that's a client problem? e.g. things should be shown collapsed by default, if you open up a function, sure you get lots of stuff.

Examples to think about:
```
f x = g x
  where 
    g = ...

f x = 
  let g = ..
  in g x

f x =
   -- awkward nesting and shadowing
   let g = let g = ... in g
   in g x

f x = do
    -- nest g inside res? what if we discarded the result?
   res <- let g = ... in g x
   pure res
```

wz1000:
We have an interval map in `Development.IDE.Spans.LocalBindings` that could be used to implement this

sgillespie:
Hi, I've been hacking at this, and here's an example of how it looks in emacs with eglot:

![ss-20240522-213934](https://github.com/haskell/haskell-language-server/assets/139144/73ec4f2e-6416-438c-9bfe-bfa61ae1ace4)

The markup itself looks like this:

````
* * *
Local bindings:

```haskell
hf :: HieASTs a
rm :: RefMap a
<-- Snip -->
prettyNames :: [Text]

```
* * *
````

How is this?

**Comment by michaelpj** [comment]
I'm a little confused about where you're adding this? I thought the idea was to put it into the document symbols, so it would show up in the outline, or similar?

**Comment by sgillespie** [comment]
After some discussion on Matrix, it sounds like what we really want is a something like
this:

 * f
   * g
     * h

(assuming f, g, and h are nested local functions). The IntervalMap in
`Development.IDE.Spans.LocalBindings` does not currently seem to give us this. Instead we
end up with a flat list of names `[f, g, h]`.

This does not seem to be so straightforward. What am I missing?

**Comment by michaelpj** [comment]
> This does not seem to be so straightforward. What am I missing?

We might just need to change it! There's no guarantee the existing code does what we want. I guess one question is whether we would be okay with the non-nested version or whether that's going to be more confusing than not having local bindings...

> Instead we end up with a flat list of names [f, g, h].

This is a little surprising to me, since AFAICT completions seem to work properly, i.e. in this program
```
f = goto 
  where
     goto = hello
      where
        hello = 1
```
I get "goto" in the completions at the top level but not "hello".

sgillespie:
[...]

PR #4312, opened 2024-06-11
Add local definitions to outline

sgillespie:
Addresses #3628. Adds all local symbols (functions/vars) as children to function document symbols.

There are a few potential issues:

 1. The position/range points to the toplevel function
 2. `getFuzzyScope` might be abused here (should I create another function like `getLocalScope`?)

Thoughts?


sgillespie:
This is what the outline looks like in emacs (via imenu-list)

ss-20240611-103517

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -66,11 +72,16 @@ moduleOutline ideState _ DocumentSymbolParams{ _textDocument = TextDocumentIdent
...
 
-
     Nothing -> pure $ InL []
 
-documentSymbolForDecl :: LHsDecl GhcPs -> Maybe DocumentSymbol
-documentSymbolForDecl (L (locA -> (RealSrcSpan l _)) (TyClD _ FamDecl { tcdFam = FamilyDecl { fdLName = L _ n, fdInfo, fdTyVars } }))
+getRefMap :: HieAstResult -> RefMap Type
+getRefMap HAR{refMap=refMap, hieKind=hieKind} =
+  case hieKind of
+    HieFromDisk _ -> mempty
```
soulomoon, line 80: I have a question, why we use `mempty` for `HieFromDisk ` here

sgillespie: This is mostly because I needed a `RefMap Type`, and I didn't know how to get there from `RefMap TypeIndex`. If this is important, I can keep hacking.

jhrcek: Maybe you can find some inspiration for how to recover full types here?
https://github.com/haskell/haskell-language-server/blob/c11f32b62a199200951dbdcacce31516033ab031/ghcide/src/Development/IDE/Spans/AtPoint.hs#L298-L299

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -282,4 +305,11 @@ hsConDeclsBinders cons
              -> [LFieldOcc GhcPs]
     get_flds flds = concatMap (cd_fld_names . unLoc) (unLoc flds)
 
-
+getLocalBindings :: RefMap Type -> LHsDecl GhcPs -> [Name]
```
michaelpj, line 308: This could do with some explanation: I'm not sure reading this why this is the right way to get the local bindings.

ghcide/src/Development/IDE/LSP/Outline.hs
```diff
@@ -282,4 +305,11 @@ hsConDeclsBinders cons
...
+getLocalBindings refmap (L (locA -> (RealSrcSpan l _)) _) =
+  nubOrdOn getOccFS . filter isVarName . map fst $ locals
+  where
+    locals = getFuzzyScope (bindings refmap) start end
```
michaelpj, line 312: This doesn't recurse, right? So presumably this won't handle something like:
```
f = ...
  where
     g = ...
       where
          h = ...
```
?

Worth adding as a test case even if we don't support it, but also maybe we just can support it?

ghcide/src/Development/IDE/LSP/Outline.hs
sgillespie: What is confusing? Is it the use of `getFuzzyScope` or is it something else?

michaelpj: I guess I'm just rather unsure why this is the right thing to do? Why is it okay to use `getFuzzyScope` rather than `getLocalScope`? What's going on with position mapping?

michaelpj:
A test case I'd like to see
```
f 1 = g
  where g = 2
f _ = g 
  where g = 3
```

ghcide/src/Development/IDE/LSP/Outline.hs
sgillespie: It actually does recurse, I've just pushed an update that adds this and the other test case below

ghcide/src/Development/IDE/LSP/Outline.hs
sgillespie: According to my testing, `getLocalScope` only returns symbols at the current start point (the start of the given function), which does not appear to have the locals in scope. `getFuzzyScope` returns all symbols that have intersecting scopes. This makes sense to me when I look at the docs for `IntervalMap.FingerTree`'s `dominator` (`getLocalScope`) and `intersections` (`getFuzzyScope`):

`dominators`:
> O(k log (n/k)). All intervals that contain the given interval, in lexicographical order.

`intersections`
>O(k log (n/k)). All intervals that intersect with the given interval, in lexicographical order.

It's clear that there at least needs a comment there, but maybe it would be better to create a new function with a clearer name?

michaelpj: > According to my testing, getLocalScope only returns symbols at the current start point (the start of the given function), which does not appear to have the locals in scope.

I feel confused about what `getLocalScope` is for then, if not this? Are we using it wrong?

ghcide/test/exe/OutlineTests.hs
```diff
@@ -78,6 +78,27 @@ tests =
...
+      testSymbolsA
+        "function with nested scope"
+        ["a x = g x where g = h where h = id"]
+        [ docSymbolWithChildren
+            "a"
+            SymbolKind_Function
+            (R 0 0 0 34)
+            [ docSymbol "x" SymbolKind_Function (R 0 0 0 34)
+            , docSymbol "g" SymbolKind_Function (R 0 0 0 34)
+            , docSymbol "h" SymbolKind_Function (R 0 0 0 34)
```
michaelpj, line 90: These locations are wrong? They all have the range of the original binding?

michaelpj: Also, shouldn't `h` be a child of `g`, not of `a`?

ghcide/test/exe/OutlineTests.hs
```diff
@@ -78,6 +78,27 @@ tests =
...
+            ]
+        ],
+      testSymbolsA
+        "function with patterns"
+        ["f 1 = g where g = 2", "f _ = g where g = 3"]
+        [ docSymbolWithChildren
+            "f"
+            SymbolKind_Function
+            (R 0 0 1 19)
+            [docSymbol "g" SymbolKind_Function (R 0 0 1 19)]
```
michaelpj, line 100: This seems wrong, there should be two `g`s?

michaelpj (commented):
Tests are helpful! I think the results look a bit off?

PR #4316, opened 2024-06-12
Bump stack configs + CI to 9.6.5 and 9.8.2

jhrcek:
Partly motivated as a workaround for https://github.com/haskell/haskell-language-server/pull/4314#issuecomment-2162553156 but also as part of general migration towards more recent ghc versions.

Also using more restrictive [allow-newer-deps](https://docs.haskellstack.org/en/stable/yaml_configuration/#allow-newer-deps) to prevent issues like in the fourmolu upgrade PR, where allow-newer caused fourmolu version bounds from cabal file to be ignored.


ghcide/ghcide.cabal
```diff
@@ -62,7 +62,7 @@ library
     , deepseq
     , dependent-map
     , dependent-sum
-    , Diff                         ^>=0.4.0
+    , Diff                         ^>=0.5
```
jhrcek, line 65: Seems like this version has nice perf improvements: https://github.com/seereason/Diff/blob/master/CHANGELOG.md

Also needed to bump to avoid adding Diff-0.4 to extra deps because it's not in recent stack resolvers.

fendor (approved):
LGTM, thanks :)

ghcide/ghcide.cabal
maralorn: Small feedback: This change is making trouble in distributing.
There was as I see it no reason to raise the lower bound here. (Raising the upper bound is good.) Especially since Diff 0.4.0 is the current Diff version in stackage-lts this lower bound bump seems a bit hasty.

jhrcek: Sorry for that. I didn't think about it too deeply. Do you want me to try to widen the range again?

maralorn: Not sure. Would probably be a good idea, but I have a workaround downstream so I don’t care too much.

PR #4318, opened 2024-06-13
Fix quadratic memory usage in GetLocatedImports

### LINKED ISSUE #4317: GetLocatedImports uses quadratic amount of memory in KnownTargets at startup
At startup `GetLocatedImports` is called on all known files. Say you have 10000 modules in your project then this leads to 10000 calls to GetLocatedImports running concurrently. 

In `GetLocatedImports` the known targets are consulted and the targetsMap is created by mapping the known targets. This map
is used for introducing sharing amongst filepaths.  This operation copies a local copy of the `target` map which is local to the rule. 

```
let targetsMap = HMap.mapWithKey const targets
```

So now each rule has a hashmap of size 10000 held locally to it and depending on how the threads are scheduled there will be 
10000^2 elements in total allocated in hashmaps. This used a lot of memory.

Solution: Return the normalising map in the result of the `GetKnownTargets` rule so it is shared across threads.

mpickering:
At startup `GetLocatedImports` is called on all known files. Say you have 10000 modules in your project then this leads to 10000 calls to GetLocatedImports running concurrently.

In `GetLocatedImports` the known targets are consulted and the targetsMap is created by mapping the known targets. This map is used for introducing sharing amongst filepaths.  This operation copies a local copy of the `target` map which is local to the rule.

```
let targetsMap = HMap.mapWithKey const targets
```

So now each rule has a hashmap of size 10000 held locally to it and depending on how the threads are scheduled there will be 10000^2 elements in total allocated in hashmaps. This used a lot of memory.

Solution: Return the normalising map in the result of the `GetKnownTargets` rule so it is shared across threads.

Fixes #4317

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -505,13 +505,12 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
                 return [(targetTarget, Set.fromList found)]
           hasUpdate <- atomically $ do
             known <- readTVar knownTargetsVar
-            let known' = flip mapHashed known $ \k ->
-                            HM.unionWith (<>) k $ HM.fromList knownTargets
+            let known' = flip mapHashed known $ \k -> unionKnownTargets k (mkKnownTargets knownTargets)
                 hasUpdate = if known /= known' then Just (unhashed known') else Nothing
             writeTVar knownTargetsVar known'
             pure hasUpdate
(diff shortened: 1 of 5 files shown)


michaelpj:
Please write down this subtle reasoning in the code! Otherwise we will be unlikely to retain this property over time.

fendor:
I added the comment to the documentation of 'normalisingMap'.

michaelpj (approved):

PR #4320, opened 2024-06-14
Cleanup allow-newer for ghc 9.10

jhrcek:


cabal.project
```diff
@@ -12,8 +12,6 @@ index-state: 2024-06-10T12:08:58Z
 tests: True
 test-show-details: direct
 
--- benchmarks: True
```
jhrcek, line 15: Removing because we have `benchmarks: True` / `benchmarks: False` in the if/else section below.

fendor (approved):
LGTM, thanks! :)

PR #4322, opened 2024-06-14
Add missing documentation for cabal formatters

### LINKED ISSUE #4313: Lack of documentation of cabal-fmt & cabal-gild

I tried to format a cabal file with hls through the `cabal-fmt` plugin, but failed. Looking at the logs, I found that I needed `cabal-gild`:

<details><summary>Details</summary>
<p>

```
2024-06-11T15:28:37.314 helix_term::application [WARN] unhandled window/showMessage: ShowMessageParams { typ: Error, message: "Error condition, please check your setup and/or the [issue tracker](https://github.com/haskell/haskell-language-server/issues): \ncabal-gild: Internal Error: No installation of cabal-gild could be found. Please install it globally, or provide the full path to the executable." }
```

</p>
</details> 

I looked through the `latest` docs and didn't find anything...

(Looked here: https://haskell-language-server.readthedocs.io/en/latest/features.html#formatting and here: https://haskell-language-server.readthedocs.io/en/latest/support/plugin-support.html#current-plugin-support-tiers)

Add documentation about:
- `cabal-gild` (I didn't find any mention to it in the docs)
- the default cabal formatter used
- how to configure the cabal formatter used

-

## Additional context

-

fendor:
Fixes #4313

Adds documentation about the default formatters. I am not sure that's the correct location, though, to document the defaults.


docs/features.md
```diff
@@ -111,6 +111,7 @@ Completions for language pragmas.
 ## Formatting
 
 Format your code with various Haskell code formatters.
+The default Haskell code formatter is `ormolu`, and the Haskell formatter be configured via the `formattingProvider` option.
```
jhrcek, line 114:
```suggestion
The default Haskell code formatter is `ormolu`, and the Haskell formatter cah be configured via the `formattingProvider` option.
```

jhrcek (approved):
LGTM

docs/features.md
```diff
@@ -119,12 +120,17 @@ Format your code with various Haskell code formatters.
 | Ormolu          | `hls-ormolu-plugin`          |
 | Stylish Haskell | `hls-stylish-haskell-plugin` |
 
+---
```
michaelpj, line 123: does this do anything?

docs/features.md
```diff
@@ -119,12 +120,17 @@ Format your code with various Haskell code formatters.
...
+
 Format your cabal files with a cabal code formatter.
+The default cabal code formatter is `cabal-gild`, which needs to be available on the `$PATH`,
```
michaelpj, line 126: This isn't a great experience, actually. So the default behaviour will send people errors, since most people won't have installed this? I think our default behaviour shouldn't rely on any external programs, which makes me think we should set the default cabal formatter to "none".

michaelpj (commented):
I'm not sure we should document default values for options here. Someone was working on auto-generating option documentation at Zurihac, that would be the ideal place for it...

docs/features.md
fendor: Hm, I thought we did. We definitely should have a "none" option.

fendor:
Seems like I forgot to add the `cabalFormattingProvider` to vscode :see_no_evil: My mistake.

https://github.com/haskell/vscode-haskell/issues/1099

docs/features.md
jhrcek: I think this markdown syntax is usually translated to `<hr/>` in html, so should be fine.

docs/features.md
fendor: Yeah, the initial experience is not good. In the case of cabal-gild, we can likely ship it with all supported GHC versions.

However, perhaps we should start shipping the binaries instead of linking it into HLS directly. Then every HLS version would have the same formatter version. Let's give #411 another kick.

fendor:
@michaelpj I think we should merge this as is, as it merely documents the status-quo

Issue #4282, opened 2024-06-07
Cabal plugin outline view


VenInf:
I think that it would be convenient to see an [outline view](https://code.visualstudio.com/docs/getstarted/userinterface#_outline-view) of a cabal file in VSCode.

In the implementation, I propose to use the AST from the cabal's parser, using the `readFileds` rule.
(the latest usage can be found [here](https://github.com/haskell/haskell-language-server/pull/4186))

It can be converted to the outline tree and used like [this](https://github.com/haskell/haskell-language-server/blob/master/ghcide/src/Development/IDE/Plugin/HLS/GhcIde.hs#L49). 

I would like to implement this feature.

PR #4323, opened 2024-06-15
Cabal plugin outline view

VenInf:
This is an attempt to solve an issue #4282.
It adds an Outline for the cabal files.

### Implementation details
Using [runIdeAction](https://github.com/VenInf/haskell-language-server/blob/cabal-plugin-outline-view/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs#L38) we get parsed [cabal fields](https://hackage.haskell.org/package/Cabal-syntax-3.12.0.0/docs/Distribution-Fields.html) with positions of the keywords.

This can be transformed with a [series of functions](https://github.com/VenInf/haskell-language-server/blob/cabal-plugin-outline-view/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs#L46) to the [DocumentSymbol](https://hackage.haskell.org/package/lsp-types-2.3.0.0/docs/Language-LSP-Protocol-Types.html#t:DocumentSymbol). It is returned, and the `moduleOutline` can be plugged in to the [cabal descriptor](https://github.com/VenInf/haskell-language-server/blob/cabal-plugin-outline-view/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs#L92).


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,131 @@
...
+          where
+            allSymbols = mapMaybe documentSymbolForField fieldPositions
+        Nothing -> pure $ LSP.InL []
+    Nothing -> pure $ LSP.InL []
+
+documentSymbolForField :: Field Position -> Maybe LSP.DocumentSymbol
+documentSymbolForField (Field (Name pos fieldName) fieldLines) =
+  Just
+    (defDocumentSymbol range)
+      { LSP._name = decodeASCII fieldName,
```
michaelpj, line 50: This will throw if the input isn't ascii, but that's probably fine, I think that won't crash HLS.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,131 @@
...
+            )
+      }
+  where
+    range = cabalPositionToLSPRange pos `addNameLengthToLSPRange` decodeASCII fieldName
+
+documentSymbolForSectionArgs :: SectionArg Position -> Maybe LSP.DocumentSymbol
+documentSymbolForSectionArgs (SecArgName pos identifier) =
+  Just
+    (defDocumentSymbol range)
+      { LSP._name = decodeLatin1 identifier,
```
michaelpj, line 74: please document your choices for decoding in a Note

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,131 @@
...
+  Just
+    (defDocumentSymbol range)
+      { LSP._name = decodeLatin1 line, -- since there is no ascii invariant (?)
+        LSP._kind = LSP.SymbolKind_Field,
+        LSP._children = Nothing -- can't delete even though the base case covers this (?)
+      }
+  where
+    range = cabalPositionToLSPRange pos `addNameLengthToLSPRange` decodeASCII line
+
+cabalPositionToLSPRange :: Position -> LSP.Range
```
michaelpj, line 110: Worth noting that this just makes a single point range!

michaelpj (commented):
Needs some tests. I'll let the others comment on the symbol kind choices, I'm not sure what those constructors actually correspond to.

VeryMilkyJoe:
Thank you for working on this feature, it seems like a great addition to the plugin!
Could you add a screenshot of what the feature looks like in action?

VenInf:
> Thank you for working on this feature, it seems like a great addition to the plugin! Could you add a screenshot of what the feature looks like in action?

Yes, sure.
There is how the outline looks for `haskell-language-server.cabal` file and `simple.cabal` file from cabal plugin tests.
hls_cabal_outline_screenshot
simple_cabal_outline_sway-screenshot

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
VenInf: Yes, that was motivated by various [unusual data](https://hackage.haskell.org/package/Cabal-syntax-3.10.2.0/docs/Distribution-Fields-Parser.html#g:1) that could be stored in the `SecArgName`'s _name.
I couldn't find an example of an edge case, though.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
VenInf: I will document this as well.
The `readFields` returns a single position per field, so this choice of range seems reasonable for a default case.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
fendor: The cabal file is definitely utf-8. However, field names can contain non-ascii characters, such as `x-的: Test` (which is a custom field). Why not `decodeUtf8`?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
fendor: I think this should be [`decodeUtf8`](https://hackage.haskell.org/package/text-2.1.1/docs/Data-Text-Encoding.html#v:decodeUtf8) either way?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
VenInf: Missed that it's utf-8.
I will change it to the `decodeUtf8` thanks.

VeryMilkyJoe:
I was trying out the feature, it looks like a great start! 
But, I think there is some unwanted behaviour here.
Firstly, I think it would be less overwhelming to only see the parents in the outline by default, with everything else collapsed.
Secondly, I think it would be easier to use the outline if the names of stanzas were part of the respective elements instead of being a child.
Also, I think the if conditions are not grouped correctly as can be seen below.
image

VeryMilkyJoe:
I think, it could also make sense to not show leaf values in the outline at all?
Specifically for long descriptions or exposed-modules content, the text becomes rather long.

VenInf:
> I was trying out the feature, it looks like a great start! But, I think there is some unwanted behaviour here. Firstly, I think it would be less overwhelming to only see the parents in the outline by default, with everything else collapsed.

I thought that this will be an easy fix, but it looks like that it's unclear, who can provide such settings. It certanly can be done [manualy](https://github.com/microsoft/vscode/issues/53262), but I can't tell if it's possible to do by default. I will look into it, but I will be glad to hear any ideas for a soltion.

> Secondly, I think it would be easier to use the outline if the names of stanzas were part of the respective elements instead of being a child. Also, I think the if conditions are not grouped correctly as can be seen below.

Right now the outline displays the cabal's AST without any modifications. I think it will be nice to see the if statement in one line, and looks like it can be done by jamming all `SectionArg`s on one level in one `DocumentSymbol`. I will test this.

VenInf:
> I think, it could also make sense to not show leaf values in the outline at all? Specifically for long descriptions or exposed-modules content, the text becomes rather long.

At least for me it's not obvious, what data should be displayed, and what omitted. As far as I can tell, it can be independent with the depth in the cabal's AST.

I think a collapsed outline would solve this issue as well.

michaelpj:
I don't think we have any control over whether or not the outline is expanded or not.

> At least for me it's not obvious, what data should be displayed, and what omitted.

As I understand it, the outline view is generally used for navigation between named entities in the file. So I definitely don't think we should be including the _values_ of fields, that's quite odd.

I think having the top-level field names is probably useful, in that it allows jumping to them, but I think it would be nice to group them in a "top-level stanza" element, perhaps?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,131 @@
...
+            allSymbols = mapMaybe documentSymbolForField fieldPositions
+        Nothing -> pure $ LSP.InL []
+    Nothing -> pure $ LSP.InL []
+
+documentSymbolForField :: Field Position -> Maybe LSP.DocumentSymbol
+documentSymbolForField (Field (Name pos fieldName) fieldLines) =
+  Just
+    (defDocumentSymbol range)
+      { LSP._name = decodeASCII fieldName,
+        LSP._kind = LSP.SymbolKind_Object,
```
michaelpj, line 51:
```suggestion
        LSP._kind = LSP.SymbolKind_Field,
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,131 @@
...
+  Just
+    (defDocumentSymbol range)
+      { LSP._name = decodeLatin1 string,
+        LSP._kind = LSP.SymbolKind_String,
+        LSP._children = Nothing
+      }
+  where
+    range = cabalPositionToLSPRange pos `addNameLengthToLSPRange` decodeASCII string
+
+documentSymbolForFieldLine :: FieldLine Position -> Maybe LSP.DocumentSymbol
```
michaelpj, line 99: I think what I'm suggesting is that we get rid of this.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,131 @@
...
+        LSP._children =
+          Just
+            ( mapMaybe documentSymbolForField fields
+                ++ mapMaybe documentSymbolForSectionArgs sectionArgs
+            )
+      }
+  where
+    range = cabalPositionToLSPRange pos `addNameLengthToLSPRange` decodeASCII fieldName
+
+documentSymbolForSectionArgs :: SectionArg Position -> Maybe LSP.DocumentSymbol
```
michaelpj, line 70: I'm unsure we want to include these either, what do they correspond to? It looks like this is responsible for e.g. the weirdness with conditionals.

VeryMilkyJoe: I think they can also include the names of stanzas so we definitely want to process them somehow but probably not into their own DocumentSymbol element

VenInf:
Fixed some minor issues, and hopefully solved the problem with displaying values of the fields.
Also now the `Section` and `SectionArgs` are displayed in one line.
For example `source-repository head` and `if impl(ghc >= 9.8)` 

cabal_outline_ifs_oneline_sway-screenshot

VeryMilkyJoe:
> Fixed some minor issues, and hopefully solved the problem with displaying values of the fields. Also now the `Section` and `SectionArgs` are displayed in one line. For example `source-repository head` and `if impl(ghc >= 9.8)`
> 
> cabal_outline_ifs_oneline_sway-screenshot

This looks great!

VeryMilkyJoe:
To fix the pre-commit failing, you will want to run the `stylish haskell` formatter on your code, especially the imports.

VenInf:
The tests are now implemented as well.

VeryMilkyJoe (approved):

plugins/hls-cabal-plugin/test/Outline.hs
```diff
@@ -0,0 +1,86 @@
...
+                                    { _name = "if os ( windows )",
+                                      _kind = T.SymbolKind_Object,
+                                      _children = Just $ [sectionArgChildrenDocumentSymbol]                                    }
+    sectionArgChildrenDocumentSymbol :: DocumentSymbol
+    sectionArgChildrenDocumentSymbol = (defDocumentSymbol (Range {_start = Position {_line = 1, _character = 4},
+                                                                     _end = Position {_line = 1, _character = 17}}))
+                                    { _name = "build-depends",
+                                      _kind = T.SymbolKind_Field,
+                                      _children = Nothing
+                                    }
```
fendor, line 74: This formatting looks hard to read to me, could you format this section? Parts of the hls-cabal-plugin codebase are formatted with `fourmolu`.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,103 @@
...
+      case fmap fst mFields of
+        Just fieldPositions -> pure $ LSP.InR (LSP.InL allSymbols)
+          where
+            allSymbols = mapMaybe documentSymbolForField fieldPositions
+        Nothing -> pure $ LSP.InL []
+    Nothing -> pure $ LSP.InL []
+
+-- | Creates a DocumentSumbol object for the
+--   cabal AST, without displaying fieldLines and
+--   displaying Section name and SectionArgs in one line
```
fendor, line 46: Could you expand on this with a concrete example? E.g. 


We turn an cabal ast such as
```
-- @
--   library
--     build-depends: base
-- @
-- 
-- into 
-- 
-- @ 
--   library
--     build-depends
-- @ 
-- 
-- omitting the value of the field `build-depends`, as it scaled poorly on big projects.

File: plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Types.hs [hunk 9]
```diff
@@ -180,3 +180,6 @@ lspPositionToCabalPosition :: Position -> Syntax.Position
 lspPositionToCabalPosition pos = Syntax.Position
   (fromIntegral (pos ^. JL.line) + 1)
   (fromIntegral (pos ^. JL.character) + 1)
+
+cabalPositionToLSPPosition :: Syntax.Position -> Position
```
fendor, line 184: Docs please, basically copy paste the docs from `lspPositionToCabalPosition`.

fendor (approved):
LGTM, thanks! Just some documentation nitpicks, code looks code otherwise!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Outline.hs
```diff
@@ -0,0 +1,103 @@
...
+
+joinedNameForSectionArgs :: [SectionArg Position] -> T.Text
+joinedNameForSectionArgs sectionArgs = joinedName
+  where
+    joinedName = T.unwords $ map getName sectionArgs
+
+    getName :: SectionArg Position -> T.Text
+    getName (SecArgName _ identifier)  = decodeUtf8 identifier
+    getName (SecArgStr _ quotedString) = decodeUtf8 quotedString
+    getName (SecArgOther _ string)     = decodeUtf8 string
```
fendor, line 78: Docs please here as well, what does this do? Bonus points for practical examples! A better name perhaps would be `prettySectionArgs`

This function should likely be a part of https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/CabalFields.hs#L12

fendor (commented):
One more thing

plugins/hls-cabal-plugin/test/Outline.hs
VenInf: This formatting was accidental. Switched to the `fourmolu` with slight modifications.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/CabalFields.hs
```diff
@@ -66,3 +66,19 @@ getOptionalSectionName (x:xs) = case x of
     Syntax.SecArgName _ name -> Just (T.decodeUtf8 name)
     _                        -> getOptionalSectionName xs
 
+
+-- | Makes a single text line out of multiple
+--   @SectionArg@s. Allowes to display conditions,
```
fendor, line 71:
```suggestion
--   @SectionArg@s. Allows to display conditions,
```

VeryMilkyJoe:
@VenInf You need to format some imports, but afterwards I think we're good to go to merge!

VenInf:
> I don't understand why this isn't merging... Or rather, why CircleCI seems to be hating @VenInf in particular. 

Maybe I have to register somewhere? Or provide some sort of credentials?

PR #4325, opened 2024-06-17
[WIP] Introduce new code action: suggest missing fields

### LINKED ISSUE #462: Feature Request: Automatic Expansion of uninitialized record fields
When a warning for "Fields of ‘SomeRecord’ not initialised: a, b, c, d" fires, I'd like a `CodeAction` that allows it to initialize all of those rows with typed holes.

Flow would look like this:
1. let thing = SomeRecord{}
2. Ctrl+'.' (or other code action trigger)
3. ~~let thing = SomeRecord{}~~ --> let thing = SomeRecord { a = _, b = _, c = _, d = _ }

The reason that this is somewhat of a big deal is that in the presence of long record names with the conventional prefix based row names (someRecordA, someRecordB, ...), the typing for this is nontrivial and also entirely mechanical and so the editor is perfectly capable of doing this for me.

akshaymankar:
Detects `-Wmissing-fields` warnings and suggests a code action to add all the missing fields.

Resolves #462

Pending tasks:
- [ ] The basic test is failing because there are extra new lines
- [ ] Add cases for `HsProc`
- [ ] De-duplicate some copy-pasted code
- [ ] Extract the code action into its own module
- [ ] Add a few more test cases

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -1627,6 +1627,7 @@ library hls-refactor-plugin
                     Development.IDE.Plugin.Plugins.Diagnostic
                     Development.IDE.Plugin.Plugins.FillHole
                     Development.IDE.Plugin.Plugins.FillTypeWildcard
+                    Development.IDE.Plugin.Plugins.SuggestAddMissingFields
                     Development.IDE.Plugin.Plugins.ImportUtils
   default-extensions:
     CPP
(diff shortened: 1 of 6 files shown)


akshaymankar:
>  The basic test is failing because there are extra new lines

I need help with this one. I tried adding `traceAst` everywhere, but it still doesn't make sense to me.

PR #4326, opened 2024-06-19
Drop 9.2

michaelpj:
Per our new deprecation policy, since the current binding constraint is GHCup recommended, which is on 9.4. 

I did my usual "evaluate away the CPP" trick, I haven't attempted to simplify what's left. I do wonder if we're close to not needing `hie-compat` at all?

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -940,7 +940,6 @@ library hls-splice-plugin
     , extra
     , foldl
     , ghc
-    , ghc-exactprint
     , ghcide                == 2.9.0.0
     , hls-plugin-api        == 2.9.0.0
     , haskell-language-server:hls-refactor-plugin


wz1000:
I have pushed a fairly aggressive cleanup of session loader stuff which is more likely than not broken in some horribly subtle ways. Let's see what CI says.

jhrcek (approved):
I did my best to remove the rest of dead code, configs and fix few warnings :smiley: 

I found one more opportunity to simplify things here: https://github.com/haskell/haskell-language-server/blob/287ee42ab566e5c088d80ec859d991f5be04f66e/ghcide/src/Development/IDE/GHC/Compat/Plugins.hs#L86-L91

Though I'm missing some ghc domain knowledge to be able to act on it.
@wz1000 @michaelpj  feel free to do something about it or leave it as is.
Otherwise LGTM as is.

konn (approved):
LGTM as for Splice Plugin, thanks!

soulomoon (commented):
LGTM!

wz1000:
Let's not squash all of these changes before merging, in particular I would like to keep 7665e2c5eba22f0d29a79ce15dc769cb2b4e79dc as a separate commit in the history so that it is easy to view these changes in isolation or possibly revert.

michaelpj:
Great work  team.

michaelpj:
Let's land this, it seems to work and we can do more cleanup as we want.

michaelpj:
> Let's not squash all of these changes before merging

I hit "rebase and merge", so it's all in there.

PR #4329, opened 2024-06-20
Remove final allow-newer for 9.10

michaelpj:
Revised on hackage

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -7,10 +7,12 @@ packages:
          ./hls-plugin-api
          ./hls-test-utils
 
-index-state: 2024-06-13T17:12:34Z
+index-state: 2024-07-20T00:00:00Z
 
 tests: True
 test-show-details: direct
+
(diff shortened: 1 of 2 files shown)


michaelpj:
I had to bound `cabal-gild` since we seem to have issues with newer versions: https://github.com/tfausak/cabal-gild/issues/89

cabal.project
```diff
@@ -40,12 +42,6 @@ constraints:
...
-
-
 if impl(ghc >= 9.9)
-  allow-newer:
-    haddock-library:base,
-    haddock-library:containers,
+  -- https://github.com/haskell/haskell-language-server/issues/4324
   benchmarks: False
-else
-  benchmarks: True
```
soulomoon, line 51: I guess `benchmarks: True` is default right?

michaelpj: I moved it up above.

PR #4332, opened 2024-06-21
making hls-cabal and hls-hlint faster

komikat:
`expectNoMoreDiagnostics 1` waits for 1 second to establish that no diagnostics have been created, it has been replaced with `waitForTypecheck` and `waitForAction "gethieast"` for Haskell and cabal files respectively, shaving off 1 second each time. I'm not sure if I should wait for some other action for cabal files.

### DIFF
--- a/plugins/hls-cabal-plugin/test/Main.hs
+++ b/plugins/hls-cabal-plugin/test/Main.hs
@@ -110,14 +110,15 @@ pluginTests =
                 liftIO $ newDiags @?= []
             , runCabalTestCaseSession "No Diagnostics in .hs files from valid .cabal file" "simple-cabal" $ do
                 hsDoc <- openDoc "A.hs" "haskell"
-                expectNoMoreDiagnostics 1 hsDoc "typechecking"
+                _ <- waitForTypecheck hsDoc
                 cabalDoc <- openDoc "simple-cabal.cabal" "cabal"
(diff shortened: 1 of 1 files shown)


soulomoon:
you may also need to consider the debouncer delay

komikat:
looking into it, thanks!

plugins/hls-cabal-plugin/test/Main.hs
```diff
@@ -110,14 +110,15 @@ pluginTests =
...
-                expectNoMoreDiagnostics 1 cabalDoc "parsing"
+                _ <- waitForAction "gethieast" cabalDoc
+                return ()
             , runCabalTestCaseSession "Diagnostics in .hs files from invalid .cabal file" "simple-cabal" $ do
                     hsDoc <- openDoc "A.hs" "haskell"
-                    expectNoMoreDiagnostics 1 hsDoc "typechecking"
+                    _ <- waitForTypecheck hsDoc
                     cabalDoc <- openDoc "simple-cabal.cabal" "cabal"
-                    expectNoMoreDiagnostics 1 cabalDoc "parsing"
+                    _ <- waitForAction "gethieast" cabalDoc
```
fendor, line 121: I can imagine this works, cabal files don't run this action, right?

fendor (changes requested):

PR #4333, opened 2024-06-23
Enable stylish-haskell with ghc 9.10

jhrcek:
Blocked on https://github.com/haskell/stylish-haskell/pull/480 which in turn is blocked on ghc 9.10 becoming available in stackage snapshot..


michaelpj:
Great work, and the tests all pass. So looks like we just need a new stackage snapshot, a review, and a release there.

jhrcek:
ghc 9.10 compatibility for stylish-haskell has been enabled in https://github.com/haskell/haskell-language-server/pull/4559

PR #4335, opened 2024-06-25
Formalize the ProgressReporting Type

soulomoon:
follow up of https://github.com/haskell/haskell-language-server/pull/4205
This is intended to remove the useless `inProgress` when tracking the `inProgressState` from the outside.
1. Add kind `data ProgressStateTrackingType = ProgressStateInsideTracking | ProgressStateOutsideTracking` to distinguish the two type of `inProgressState` tracking.
2. Spread the `TProgressState` to related type location to mark what kind of tracking we used. Then we won't have `inProgress` for `ProgressStateOutsideTracking`

### DIFF
--- a/ghcide/src/Development/IDE/Core/ProgressReporting.hs
+++ b/ghcide/src/Development/IDE/Core/ProgressReporting.hs
@@ -83,16 +83,13 @@ instance ProgressReportingClass (ProgressReporting m) where
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The progress of tasks can be tracked in two ways:
 
-1. `InProgressState`: This is an internal state that actively tracks the progress.
+1. `ProgressReporting`: we have an internal state that actively tracks the progress.
    Changes to the progress are made directly to this state.
 
-2. `InProgressStateOutSide`: This is an external state that tracks the progress.
+2. `ProgressReportingNoTrace`: there is an external state that tracks the progress.
    The external state is converted into an STM Int for the purpose of reporting progress.
 
-The `inProgress` function is only useful when we are using `InProgressState`.
-
-An alternative design could involve using GADTs to eliminate this discrepancy between
-`InProgressState` and `InProgressStateOutSide`.
+The `inProgress` function is only useful when we are using `ProgressReporting`.
 -}
 
 noProgressReportingNoTrace :: (MonadUnliftIO m) => (ProgressReportingNoTrace m)
--- a/ghcide/src/Development/IDE/Core/ProgressReporting.hs
+++ b/ghcide/src/Development/IDE/Core/ProgressReporting.hs
@@ -194,8 +194,7 @@ progressReporting (Just lspEnv) title optProgressStyle = do
   inProgressState <- newInProgress
   progressReportingInner <- progressReportingNoTrace (readTVar $ todoVar inProgressState)
                                 (readTVar $ doneVar inProgressState) (Just lspEnv) title optProgressStyle
-  let
-      inProgress :: forall a. NormalizedFilePath -> m a -> m a
+  let inProgress :: forall a. NormalizedFilePath -> m a -> m a
       inProgress = updateStateForFile inProgressState
   return ProgressReporting {..}
   where
--- a/ghcide/src/Development/IDE/Core/ProgressReporting.hs
+++ b/ghcide/src/Development/IDE/Core/ProgressReporting.hs
@@ -40,41 +40,36 @@ import           Language.LSP.Server            (ProgressAmount (..),
                                                  withProgress)
 import qualified Language.LSP.Server            as LSP
 import qualified StmContainers.Map              as STM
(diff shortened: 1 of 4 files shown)


michaelpj:
I'm not sure I see the attraction of this design, to be honest. I think we could instead see the tracking with files as a _wrapper_ on top of the simpler `progressCounter` that just works with the `STM Int`s. That would mean we have two different types for the two different counters, rather than just one with a type parameter, but I think it would be simpler?

soulomoon:
Good suggestions @michaelpj , 
I've managed to use the wrapper version,
Seems much simpler using the wrapper version here.

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -42,38 +48,60 @@ data ProgressEvent
   | ProgressCompleted
   | ProgressStarted
 
-data ProgressReporting m = ProgressReporting
-  { progressUpdate :: ProgressEvent -> m (),
-    inProgress     :: forall a. NormalizedFilePath -> m a -> m a,
-    -- ^ see Note [ProgressReporting API and InProgressState]
-    progressStop   :: IO ()
+data ProgressReportingNoTrace m = ProgressReportingNoTrace
+  { progressUpdateI :: ProgressEvent -> m (),
```
michaelpj, line 52: You might not need to add name suffixes if you use `DuplicateRecordFields`?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -42,38 +48,60 @@ data ProgressEvent
...
   }
 
+data ProgressReporting m = ProgressReporting
+  {
+    inProgress             :: forall a. NormalizedFilePath -> m a -> m a,
+    -- ^ see Note [ProgressReporting API and InProgressState]
+    progressReportingInner :: ProgressReportingNoTrace m
+  }
+
+class ProgressReportingClass a where
```
michaelpj, line 65: Maybe just `ProgressReporter`? Usually we read class constraints as "X is a Y" or similar, so "X is a ProgressReporter" reads sensibly?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -42,38 +48,60 @@ data ProgressEvent
...
+    type M a :: * -> *
```
michaelpj, line 66: I think I asked this before, but can everything just be in IO? The only place we use another monad is in `inProgress`, and I think we could also just provide that with IO and get people to use `MonadUnliftIO` to use it?

ghcide/src/Development/IDE/Core/ProgressReporting.hs
```diff
@@ -42,38 +48,60 @@ data ProgressEvent
...
-    -- ^ see Note [ProgressReporting API and InProgressState]
-    progressStop   :: IO ()
+data ProgressReportingNoTrace m = ProgressReportingNoTrace
+  { progressUpdateI :: ProgressEvent -> m (),
+    progressStopI   :: IO ()
     -- ^ we are using IO here because creating and stopping the `ProgressReporting`
     -- is different from how we use it.
   }
 
+data ProgressReporting m = ProgressReporting
```
michaelpj, line 58: I would name the other one `ProgressReporting` and call this `PerFileProgressReporting` or something?

michaelpj:
Looks much better

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: `inProgress`'s monad `m` is invariant (Both covariant and contra variant locations). I doubt we can have `forall a.IO a -> IO a` -> `forall a. m a -> m a`.🤔

soulomoon: Yes, we can turn them into IOs and save the `m`

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: No, we have class function and data field collision. `DuplicateRecordFields ` is not helpful here

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: good name

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: make sense, I like this naming

ghcide/src/Development/IDE/Core/ProgressReporting.hs
soulomoon: But maybe we can make it prefix with '_' and not export it.

michaelpj (approved):
Great

PR #4338, opened 2024-06-26
Avoid setCurrentDirectory in hls-stylish-haskell-plugin (closes #4234)

### LINKED ISSUE #4234: StylishHaskell shifting the working dir, this could lead to race condition
StylishHaskell shifting the working dir, it could make other part of hls depending on cwd result in error.

https://github.com/haskell/haskell-language-server/blob/61fd5c464842448dfebe91d3e01de425eacf07b7/plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs#L81

Such as `Hlint, Stan, Template haskel, load cradle` etc. We should not shifting the working dir after initilization.

jhrcek:
A quick attempt at fixing #4234

Depends on upstream fix: https://github.com/haskell/stylish-haskell/pull/482


jhrcek:
The fix from this PR has been included as part of https://github.com/haskell/haskell-language-server/pull/4559

PR #4259, opened 2024-05-27
Migrate PluginSimpleTests

soulomoon:

PR #4339, opened 2024-06-27
Using captureKicksDiagnostics to speed up multiple plugin tests

komikat:
Extending #4144 to `cabal-plugin-tests` and merging upstream changes on `hlint-plugin-tests`
Update: Extending to multiple plugins as noted below.
Update 2: Scope reverted to cabal and hlint plugins. (ref #4371)

### TODO
- [x] update the remaining tests which waitForDiagnostics unconditionally
- [x] resolve the one hlint-plugin test failing after the update
- [x] fix pre-commit issues


komikat:
Yes, I'm looking into it, will hopefully have something by tomorrow.

PS: the failing testcase is ignoreHintGoldenResolveTest.

fendor:
This looks like... the testcase is wrong? I can't explain it, but changing `"Eta reduce"` with `"Redundant id"` works for me locally. The testcase itself is still wrong, you have to accept the differences, but no unknown code actions any more.

fendor:
Perhaps we lack a `.hlint.yaml` in the temp directory or something?

soulomoon:
@komikat since I do not familiar with cabal plugin, I can only give very general suggestions. Take consideration of the following.
1. the time gap the debouncer needed to emit the diagnostic
2. Be sure the number of cabal kicks that would happened in the test logic. Seems like @fendor 's new finding suggest some incorrect logic in the test case?

komikat:
`hlint --hint=test-hlint-config.yaml UnrecognizedPragmasOff.hs`
only returns Eta Reduce so it is a hlint config file issue

komikat:
Setting `.hlint.yaml` correctly breaks the testcase "provides 3.8 code actions including apply all" while not fixing this one at the same time, seems like the `.hlint.yaml` might not have been copied correctly for the unrecognized pragma test case.

Will also have to look into "provides 3.8 code actions including apply all" since it's correctly failing after ignoring redundant id, it expects both redundant id and eta reduce currently.

komikat:
Moved hlint tests to the virtual file system and fixed a config issue with the resolveTests. 
`hlint-plugin-tests` takes 6s on my system now.

soulomoon (approved):
good work！

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -227,7 +231,14 @@ function invocation.
...
 kick = do
   files <- HashMap.keys <$> getCabalFilesOfInterestUntracked
+  Shake.ShakeExtras{ideTesting = Options.IdeTesting testing, lspEnv} <- Shake.getShakeExtras
+  let signal :: KnownSymbol s => Proxy s -> Action ()
+      signal msg = when testing $ liftIO $ Shake.mRunLspT lspEnv $
+        LSP.sendNotification (LSP.SMethod_CustomMethod msg) $
+        toJSON $ map fromNormalizedFilePath files
+  signal (Proxy @"kick/start/cabal")
   void $ uses Types.ParseCabalFile files
+  signal(Proxy @"kick/done/cabal")
```
fendor, line 241: Can this wrapping functions be extracted and documented? While I remember what I did here, most other devs will not remember why they might need this.

plugins/hls-cabal-plugin/test/Utils.hs
```diff
@@ -49,6 +52,15 @@ runCabalSession subdir =
...
 testDataDir = "plugins" </> "hls-cabal-plugin" </> "test" </> "testdata"
 
+cabalKickDone :: Session ()
+cabalKickDone = kick (Proxy @"kick/done/cabal") >>= guard . not . null
+
+cabalKickStart :: Session ()
+cabalKickStart = kick (Proxy @"kick/start/cabal") >>= guard . not . null
+
+cabalCaptureKick :: Session [Diagnostic]
+cabalCaptureKick = captureKickDiagnostics cabalKickStart cabalKickDone
```
fendor, line 62: Docs please!

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -202,8 +207,16 @@ rules recorder plugin = do
...
   action $ do
-    files <- getFilesOfInterestUntracked
-    void $ uses GetHlintDiagnostics $ Map.keys files
+    files <- Map.keys <$> getFilesOfInterestUntracked
+    Shake.ShakeExtras{ideTesting = Options.IdeTesting testing, lspEnv} <- Shake.getShakeExtras
+    let signal :: KnownSymbol s => Proxy s -> Action ()
+        signal msg = when testing $ liftIO $ Shake.mRunLspT lspEnv $
+            LSP.sendNotification (LSP.SMethod_CustomMethod msg) $
+            toJSON $ map fromNormalizedFilePath files
+
```
fendor, line 216: Please extract to some location. Probably to ghcide?

plugins/hls-hlint-plugin/test/Main.hs
```diff
@@ -429,16 +463,15 @@ goldenTest testCaseName goldenFilename point hintText =
...
+    { testConfigCaps = config
     , testShiftRoot = True
     , testPluginDescriptor = hlintPlugin
-    , testDirLocation = Left testDir
-    }
-    testName testDir path "expected" "hs"
-
+    , testDirLocation = Right tree
+    } testName tree path "expected" "hs"
+  where tree = (mkVirtualFileTree testDir (directProject (path <.> "hs")))
```
fendor, line 474: Please remove unneeded parenthesis!

fendor (changes requested):
Thanks for pushing this over the finish line! Please also make sure to remove dead code, e.g. code that is commented out.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
komikat: would hls-test-utils be more suitable?

fendor: I don't think so, then each plugin would need to depend on `hls-test-utils` right?

komikat: ah, makes sense.

komikat: seems like using `ghcide/src/Development/IDE/GHC/Util.hs` causes a cycle in the module graph, would creating a new file be better?

fendor: Yeah, that looks like a non-ideal location. It should be in the `LSP` namespace in ghcide, perhaps `ghcide/src/Development/IDE/Core/Shake.hs`?

komikat:
Updated PR scope, please refer TODO.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -202,8 +204,11 @@ rules recorder plugin = do
     liftIO $ argsSettings flags
 
   action $ do
-    files <- getFilesOfInterestUntracked
-    void $ uses GetHlintDiagnostics $ Map.keys files
+    files <- Map.keys <$> getFilesOfInterestUntracked
+    Shake.ShakeExtras{ideTesting = Options.IdeTesting testing, lspEnv} <- Shake.getShakeExtras
+    Shake.kickSignal testing lspEnv files (Proxy @"kick/start/hlint")
+    void $ uses GetHlintDiagnostics files
+    Shake.kickSignal testing lspEnv files (Proxy @"kick/done/hlint")
```
fendor, line 211: Couldn't we wrap this to avoid some of this overhead? E.g. `runWithSignal`

```haskell
    runWithSignal "hlint" files $ do
      void $ uses GetHlintDiagnostics files
```

fendor (changes requested):

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
komikat: sure, uses "kick/start/hlint" ... for now

fendor (approved):
LGTM, doing a final CI run to make sure the tests didn't get more flaky

komikat:
seems like the timeout still happens erratically (unmarked as broken in #4259)

PR #4342, opened 2024-06-28
capture error in worker thread

### LINKED ISSUE #4340: HLS gets stuck if session setup fails for any reason
When the session set up fails, then the HLS thread crashes and will not be restarted.

I added a `catch` to the session setup, logging the error. After the error is thrown, HLS appears to be completely stuck.

Example trace:

```
2024-06-27T15:50:04.706392Z | Debug | stack setup --silent
2024-06-27T15:50:05.070647Z | Debug | stack exec ghc -- --numeric-version
[Trace - 17:50:05] Received notification '$/progress'.
2024-06-27T15:50:05.070866Z | Debug | Session loading result: Right (ComponentOptions {componentOptions = ["-i","-odir=/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/.stack-work/odir","-hidir=/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/.stack-work/odir","-hide-all-packages","-XGHC2021","-i/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable/.stack-work/dist/x86_64-linux/ghc-9.6.5/build/explainable-exe","-i/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable/app","-i/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable/.stack-work/dist/x86_64-linux/ghc-9.6.5/build/explainable-exe/autogen","-i/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable/.stack-work/dist/x86_64-linux/ghc-9.6.5/build/global-autogen","-i/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable/.stack-work/dist/x86_64-linux/ghc-9.6.5/build/explainable-exe/explainable-exe-tmp","-stubdir=/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable/.stack-work/dist/x86_64-linux/ghc-9.6.5/build","-package-id=QuickCheck-2.14.3-7WdOShjYsRTQ5DwHd4Gtb","-package-id=aeson-2.1.2.1-F3hDGK91Wx42rTliAGkUq","-package-id=base-4.18.2.1","-package-id=boxes-0.1.5-C5KrYVvXsQG8cA3g3hv7uh","-package-id=bytestring-0.11.5.3","-package-id=containers-0.6.7","-package-id=directory-1.3.8.4","-package-id=effectful-2.3.0.0-3V3X0RsGjod7jCYr6Wti","-package=explainable-0.1.0.0","-package-id=fgl-5.8.2.0-85Qkt6u9x2z5NSStlZTxr6","-package-id=filepath-1.4.300.1","-package-id=graphviz-2999.20.2.0-64QWSEOt6dh36pYl7wtiek","-package-id=hspec-2.11.9-4wTKmuILwwUEq4fqjWEoZg","-package-id=lens-5.2.3-G0LWaXeEo0UDN0iZ5b4CcG","-package-id=megaparsec-9.5.0-D2uz1Hosetnt14WoHo7Ea","-package-id=mtl-2.3.1","-package-id=numeric-extras-0.1-LgQWj8Z8pmPKPk9V7di9SO","-package-id=openapi3-3.2.4-8OMi17vBjaf4LVrZUUMg5C","-package-id=optparse-applicative-0.18.1.0-KKw4zpvajYE9Msk9T1Ikpr","-package-id=parsec-3.1.16.1","-package-id=parser-combinators-1.3.0-20UTEUZ5CKDJdRu747ID48","-package-id=pcre-heavy-1.0.0.3-6TcStAgMWa5DIw1maQ5Z9t","-package-id=prettyprinter-1.7.1-60yVE7QePDs8FHIPsacPFF","-package-id=prettyprinter-interp-0.2.0.0-5T6TkajMYyiGGBwTvUfYk3","-package-id=quickcheck-instances-0.3.30-KikxF4MVt6S9929xX2tsLH","-package-id=scientific-0.3.7.0-7ih7NuxY8AT9Jmz8D0EZgy","-package-id=servant-0.20.1-6oNY4TzoYI63JyjY0sLHWq","-package-id=servant-openapi3-2.0.1.6-FQDfjqNkbplHlJNuW2u5gJ","-package-id=servant-server-0.20-DtT9lGedLuzJu9wRxFzqT4","-package-id=servant-swagger-ui-0.3.5.5.0.0-7dLcEYwWM2i53FUQ5BcZKg","-package-id=string-interpolate-0.3.3.0-FdcmmUSqZzY5RLaNHS2LG4","-package-id=text-2.0.2","-package-id=transformers-0.6.1.0","-package-id=unordered-containers-0.2.20-ED65WLJGTgHKKHANJYMwZr","-package-id=wai-3.2.4-6PqrRGq5uVCcx5GJE074k","-package-id=wai-logger-2.4.0-K2ThryFHdX15W72WGMUmgk","-package-id=warp-3.3.31-BQkvXyHwhQ02XfAxYTn4Y8","-Wall","-Wcompat","-Widentities","-Wincomplete-record-updates","-Wincomplete-uni-patterns","-Wmissing-export-lists","-Wmissing-home-modules","-Wpartial-fields","-Wredundant-constraints","-Wall","-rtsopts","-with-rtsopts=-N","-fwrite-ide-info","-optP-include","-optP/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/.stack-work/ghci/69ba514d/cabal_macros.h","-ghci-script=/home/hugin/.cache/stack/ghci-script/3484a1de/ghci-script","-package-db","/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/.stack-work/install/x86_64-linux/b436016af794c3fa6328af8976a3bc0bcf56a1156d8c4cc82a3d6f67ee396726/9.6.5/pkgdb","-package-db","/home/hugin/.stack/snapshots/x86_64-linux/b436016af794c3fa6328af8976a3bc0bcf56a1156d8c4cc82a3d6f67ee396726/9.6.5/pkgdb","-package-db","/home/hugin/.ghcup/ghc/9.6.5/lib/ghc-9.6.5/lib/package.conf.d"], componentRoot = "/home/hugin/Documents/haskell/cclaw/dsl/lib/haskell/explainable", componentDependencies = ["explainable/explainable.cabal","explainable/package.yaml","stack.yaml"]},"/home/hugin/.ghcup/ghc/9.6.5/lib/ghc-9.6.5/lib","9.6.5")
2024-06-27T15:50:05.083638Z | Info | Interface files cache directory: /home/hugin/.cache/ghcide/main-d8039823af9b65242e8d42ce90d5528ab12596c0-d8039823af9b65242e8d42ce90d5528ab12596c0
2024-06-27T15:50:05.090341Z | Info | Making new HscEnv. In-place unit ids: [ main-d8039823af9b65242e8d42ce90d5528ab12596c0 ]
cannot satisfy -package explainable-0.1.0.0
    (use -v for more information)
 ```
 
 The error itself is a manifestation of https://github.com/commercialhaskell/stack/issues/5380, not relevant to this issue, though.
 
 However,  we should catch such errors and handle them gracefully.

### PR DESCRIPTION
**Pull request by soulomoon**
Fix #4340
let the caller handle the error happened during `awaitRunInThread`. In such case the HLS won't be stuck if the session setup fails and the LSP should be able to show the error to the client as other error cases happening in the rule system.
cc @fendor

### DIFF
--- a/ghcide/src/Development/IDE/Core/WorkerThread.hs
+++ b/ghcide/src/Development/IDE/Core/WorkerThread.hs
@@ -15,7 +15,7 @@ import           Control.Concurrent.Async  (AsyncCancelled (AsyncCancelled),
 import           Control.Concurrent.STM
 import           Control.Concurrent.Strict (newBarrier, signalBarrier,
                                             waitBarrier)
-import           Control.Exception         (Exception (fromException),
+import           Control.Exception.Safe    (Exception (fromException),
                                             SomeException, throwIO, try)
 import           Control.Monad             (forever)
 import           Control.Monad.Cont        (ContT (ContT))
@@ -52,11 +52,7 @@ awaitRunInThread q act = do
     -- Take an action from TQueue, run it and
     -- use barrier to wait for the result
     barrier <- newBarrier
-    atomically $ writeTQueue q $ do
-        resultOrException <- try act
-        case resultOrException of
-            Left e@(fromException -> Just AsyncCancelled) -> throwIO e
-            _ -> signalBarrier barrier resultOrException
+    atomically $ writeTQueue q $ try act >>= signalBarrier barrier
     resultOrException <- waitBarrier barrier
     case resultOrException of
         Left e  -> throwIO (e :: SomeException)
--- a/ghcide/src/Development/IDE/Core/WorkerThread.hs
+++ b/ghcide/src/Development/IDE/Core/WorkerThread.hs
@@ -53,11 +53,12 @@ awaitRunInThread q act = do
     barrier <- newBarrier
     atomically $ writeTQueue q $ do
         resultOrException <- try act
-        case resultOrException of
-            Left e -> case fromException e of
-                Just AsyncCancelled -> throwIO e  -- Rethrow if it's an AsyncCancelled exception
-                Nothing             -> signalBarrier barrier resultOrException  -- Handle other exceptions as before
-            Right _ -> signalBarrier barrier resultOrException
+        let result
+                -- Rethrow if it's an AsyncCancelled exception
+                -- we need to do this because we can handled the exit of the worker thread
+                | Left e <- resultOrException, Just AsyncCancelled <- fromException e = throwIO e
+                | otherwise = signalBarrier barrier resultOrException
+        result
     resultOrException <- waitBarrier barrier
     case resultOrException of
         Left e  -> throwIO (e :: SomeException)
--- a/ghcide/src/Development/IDE/Core/WorkerThread.hs
+++ b/ghcide/src/Development/IDE/Core/WorkerThread.hs
@@ -54,12 +54,9 @@ awaitRunInThread q act = do
     barrier <- newBarrier
     atomically $ writeTQueue q $ do
         resultOrException <- try act
-        let result
-                -- Rethrow if it's an AsyncCancelled exception
-                -- we need to do this because we can handled the exit of the worker thread
-                | Left e <- resultOrException, Just AsyncCancelled <- fromException e = throwIO e
-                | otherwise = signalBarrier barrier resultOrException
-        result
+        case resultOrException of
+            Left e@(fromException -> Just AsyncCancelled) -> throwIO e
+            _ -> signalBarrier barrier resultOrException
     resultOrException <- waitBarrier barrier
     case resultOrException of
         Left e  -> throwIO (e :: SomeException)
--- a/ghcide/src/Development/IDE/Core/WorkerThread.hs
+++ b/ghcide/src/Development/IDE/Core/WorkerThread.hs
@@ -10,10 +10,13 @@ module Development.IDE.Core.WorkerThread
     (withWorkerQueue, awaitRunInThread)
  where
 
-import           Control.Concurrent.Async  (withAsync)
+import           Control.Concurrent.Async  (AsyncCancelled (AsyncCancelled),
+                                            withAsync)
 import           Control.Concurrent.STM
 import           Control.Concurrent.Strict (newBarrier, signalBarrier,
                                             waitBarrier)
+import           Control.Exception         (Exception (fromException),
+                                            SomeException, throwIO, try)
 import           Control.Monad             (forever)
 import           Control.Monad.Cont        (ContT (ContT))
 
@@ -49,6 +52,13 @@ awaitRunInThread q act = do
     -- use barrier to wait for the result
     barrier <- newBarrier
     atomically $ writeTQueue q $ do
-        res <- act
-        signalBarrier barrier res
-    waitBarrier barrier
+        resultOrException <- try act
+        case resultOrException of
+            Left e -> case fromException e of
+                Just AsyncCancelled -> throwIO e  -- Rethrow if it's an AsyncCancelled exception
+                Nothing             -> signalBarrier barrier resultOrException  -- Handle other exceptions as before
+            Right _ -> signalBarrier barrier resultOrException
+    resultOrException <- waitBarrier barrier
+    case resultOrException of
+        Left e  -> throwIO (e :: SomeException)
+        Right r -> return r
--- a/ghcide/src/Development/IDE/Core/WorkerThread.hs
+++ b/ghcide/src/Development/IDE/Core/WorkerThread.hs
@@ -45,7 +45,8 @@ withWorkerQueue workerAction = ContT $ \mainAction -> do
                 workerAction l
 
 -- | 'awaitRunInThread' queues up an 'IO' action to be run by a worker thread,
--- and then blocks until the result is computed.
+-- and then blocks until the result is computed. If the action throws an
+-- non-async exception, it is rethrown in the calling thread.
 awaitRunInThread :: TQueue (IO ()) -> IO result -> IO result
 awaitRunInThread q act = do
     -- Take an action from TQueue, run it and

### CODE REVIEW

**Comment by michaelpj** [comment]
Okay, so what are the cases here?
1. The action is maybe expected to throw exceptions. Then probably the action supplied by the user should include exception-handling as appropriate.
2. The action is not expected to throw exceptions, or anyway the user has not prepared for that. Then something has gone wrong, and probably it's okay to just let everything die?

So the case we need to handle is 2... but I think we have exactly the tool for that in the form of `Async.link`. We just need to link the "main thread" to the "worker thread", so that if the worker thread dies the exception is rethrown in the main thread. So I think all we need is just a call to `link` on line 37.

**Comment by soulomoon** [comment]
Session loading belong to a part of the Rule `GhcSession` on per file bases.
If #4340 happen for one file means it would happen for all files, let everything dies seems to be a good way to go.
I worried handling in 2 is a bit too aggressive for just one file failed cases ? 

Maybe @fendor knows the cases ?

**Comment by michaelpj** [comment]
Actually I think I'm wrong and you were right: an unhandled exception can either go to the thread that launched the worker queue, or the thread that put the action into the queue. Probably the second option actually makes most sense, since it means that "run this action directly" and "run this action via the queue" behave the same in terms of where the exception goes.

I wonder if there's a nice way to implement that behaviour using some of the existing concurrency pieces we have. This sort of thing can be very tricky, so I'd feel more comfortable if we were leveraging `async` or something somehow.

**Comment by michaelpj** [comment]
To be honest, I continue to think about my comment here: https://github.com/haskell/haskell-language-server/pull/4256#pullrequestreview-2105858952

A lot of this would be simpler if we were using locks! 
- Consumers who want to block until they get the result can just wait until they get the lock and then do the thing
- Consumers who don't want to wait can fork a thread that waits for the lock

In both cases the exception handling is pretty straightforward and under the control of the consumer.

Haskell MVars are fair, so if we use a MVar-based lock we also should get things being scheduled reasonably. GHC does them in FIFO fashion, so we would even get the same queueing behaviour, but arguably this is a bit less good because it doesn't give us quite the same sequencing guarantees (does that matter?).

**Comment by michaelpj** [comment]
So basically everywhere we have a `ThreadQueue` we would instead just have a `Lock`.

**Comment by soulomoon** [comment]
> I wonder if there's a nice way to implement that behaviour using some of the existing concurrency pieces we have.
Currently I do not find an existing api for such mechanism. Since the caller are different rule evaluation threads. The scoping is nested, so the `link` function might not be in our favor.

**Comment by soulomoon** [comment]
> So basically everywhere we have a ThreadQueue we would instead just have a Lock.

`Lock` solve the competition problem. However, it can not solve the cancel problem.
We have two main cancel points for them. 
The session restart and the shutdown.
We want the shutdown can cancel it as fast as it can. 
But we do not want to session restart to cancel them.

**Comment by michaelpj** [comment]
Don't the restarts cancel the `Actions` which trigger the job enqueuing? If those can cancel their spawned actions then there's maybe not a problem there?

**Comment by soulomoon** [comment]
no, unless we can rewrite session loader like other rules and update their STM all togather in the hls-graph computation post hooker. Which is a rather hard work to do.

File: ghcide/src/Development/IDE/Core/WorkerThread.hs [hunk 1]
```diff
@@ -42,13 +45,19 @@ withWorkerQueue workerAction = ContT $ \mainAction -> do
...
     -- use barrier to wait for the result
     barrier <- newBarrier
     atomically $ writeTQueue q $ do
-        res <- act
-        signalBarrier barrier res
-    waitBarrier barrier
+        resultOrException <- try act
+        case resultOrException of
+            Left e@(fromException -> Just AsyncCancelled) -> throwIO e
+            _ -> signalBarrier barrier resultOrException
```
fendor, line 59: perhaps https://hackage.haskell.org/package/safe-exceptions-0.1.7.4/docs/Control-Exception-Safe.html#v:try is the better `try` here? I don't think we want to catch any kind of `AsyncException` and swallow it.

ghcide/src/Development/IDE/Core/WorkerThread.hs
```diff
@@ -42,13 +45,19 @@ withWorkerQueue workerAction = ContT $ \mainAction -> do
...
+    resultOrException <- waitBarrier barrier
+    case resultOrException of
+        Left e  -> throwIO (e :: SomeException)
+        Right r -> return r
```
fendor, line 63: This will bubble up to the rules engine? Or how does the logging work here?

soulomoon: It will bubble up to the rules engine. And being catch here https://github.com/haskell/haskell-language-server/blob/67680ea347308f8f6819cb6f55c9a60c9bb19313/ghcide/src/Development/IDE/Core/Shake.hs#L1248

ghcide/src/Development/IDE/Core/WorkerThread.hs
soulomoon: Yes, Nice tip

ghcide/src/Development/IDE/Core/WorkerThread.hs
soulomoon: Just like other kind of rule failure.

PR #4343, opened 2024-06-29
Fix cabal.project index-state

soulomoon:
Time travel back since the new release cabal-install 12 is not happy with us and refuse to build


jhrcek (approved):
LGTM

PR #4344, opened 2024-06-29
Simplify initPlugins

soulomoon:
@jhrcek found it. I take a look at relevant code in ghc, we do not need to update modsummary any more.

### DIFF
--- a/ghcide/src/Development/IDE/GHC/Compat/Plugins.hs
+++ b/ghcide/src/Development/IDE/GHC/Compat/Plugins.hs
@@ -52,12 +52,12 @@ initializePlugins :: HscEnv -> IO HscEnv
 initializePlugins env = do
     Loader.initializePlugins env
 
--- | Plugins aren't stored in ModSummary anymore since GHC 9.2, but this
--- function still returns it for compatibility with 8.10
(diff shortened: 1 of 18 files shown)


ghcide/src/Development/IDE/GHC/Compat/Plugins.hs
```diff
@@ -52,12 +52,12 @@ initializePlugins :: HscEnv -> IO HscEnv
 initializePlugins env = do
     Loader.initializePlugins env
 
--- | Plugins aren't stored in ModSummary anymore since GHC 9.2, but this
--- function still returns it for compatibility with 8.10
-initPlugins :: HscEnv -> ModSummary -> IO (ModSummary, HscEnv)
+-- | Plugins aren't stored in ModSummary anymore since GHC 9.2, so this
+-- function not returns ModSummary
+initPlugins :: HscEnv -> ModSummary -> IO HscEnv
```
michaelpj, line 57: We could consider just inlining this, and also the defintion of `initializePlugins` above, and so just call `Loader.initializePlugins` directly at the call site?

fendor (approved):
LGTM, thanks!

PR #4347, opened 2024-07-03
Fix core file location in `GetLinkable`

### LINKED ISSUE #4145: HLS often cannot find a `.hi.core` file in `~/.cache/ghcide/`
### Your environment

**Which OS do you use?**
MacOS 11.7.10 and 14
**Which version of GHC do you use and how did you install it?**
9.2.8 via ghcup (0.1.22.0)
**How is your project built (alternative: link to the project)?**
cabal (or nix + cabal)
https://github.com/plow-technologies/inferno

**Which LSP client (editor/plugin) do you use?**
vscode (v1.87.2) + Haskell plugin (v2.4.3)
**Which version of HLS do you use and how did you install it?**
2.7.0.0 via ghcup
**Have you configured HLS in any way (especially: a `hie.yaml` file)?**
No

### Steps to reproduce

I cannot reproduce it myself deterministically, but it happens so often that I'm opening a bug report. Today I tried on a fresh laptop, fresh ghcup + ghc + hls install, fresh install of vscode's Haskell plugin, and within an hour I saw this error again. Here's the steps:
- Install ghcup using official instructions
- Install ghc 9.2.8 and hls 2.7.0.0 using ghcup
- Install vscode and Haskell plugin 
- Wait for HLS to startup, find ghc and hls using ghcup, and build the project. It all works perfectly for a while.
- But at some point, some files start to show an error on the first line that it cannot find a `.hi.core` file in `~/.cache/ghcide/`...

### Expected behaviour

No such error.

### Actual behaviour

I see a red squiggle on the first line of the file, with the error:

> /Users/sid/.cache/ghcide/inferno-core-0.11.1.0-inplace-bdea44f530a52e915c4b2a4fb12efa71e6fd5f24/Inferno/Infer.hi.core: getFileStatus: does not exist (No such file or directory)

### Debug information

The vscode output window for Haskell shows some errors, but I don't know if they are related:
```
2024-03-21T13:20:36.759358Z | Warning | No plugin handles this "textDocument/semanticTokens/full" request.
[Error - 15:20:36] Request textDocument/semanticTokens/full failed.
  Message: No plugins are available to handle this SMethod_TextDocumentSemanticTokensFull request.
semanticTokens is disabled globally in your config.
  Code: -32601 
[Error - 15:23:06] Request textDocument/foldingRange failed.
  Message: codeRange: Rule Failed: GetCodeRange
  Code: -32803 
...

```

Thank you for all your awesome work 🙌 

---

EDIT:  Clearing the `~/.cache/ghcide/` and `~/.cache/hie-bios/` directories helps for a while, but the error is back soon.

soulomoon:
Fix #4145
The error case is demonstrated in https://github.com/haskell/haskell-language-server/issues/4145#issuecomment-2206965649
1. Include ModLocation in the `ModSummaryResult`  fingerprint.
2. Instead of getting the core file location from `GetModSummary`, get it from the result of `GetModIface` directly since that is the actual location the core file written to. While the `GetModSummary` contains the future core file location that would be written only if we would have made some change cause the rewrite. See Note [Avoiding bad interface files]

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -125,7 +125,6 @@ import           GHC.Driver.Errors.Types
 import           GHC.Types.Error                     (errMsgDiagnostic,
                                                       singleMessage)
 import           GHC.Unit.State
-import Debug.Trace (traceM)
 
 data Log
   = LogSettingInitialDynFlags
@@ -531,10 +530,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
                   -- See Note [Avoiding bad interface files]
                   let hscComponents = sort $ map show uids
                       cacheDirOpts = hscComponents ++ componentOptions opts
-                  let opts_hash = B.unpack $ B16.encode $ H.finalize $ H.updates H.init (map B.pack cacheDirOpts)
-                  traceM $ "Setting cache dirs for " ++ show rawComponentUnitId ++ " " ++ opts_hash ++ " " ++ show cacheDirOpts
                   cacheDirs <- liftIO $ getCacheDirs prefix cacheDirOpts
-
                   processed_df <- setCacheDirs recorder cacheDirs df2
                   -- The final component information, mostly the same but the DynFlags don't
                   -- contain any packages which are also loaded
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -125,6 +125,7 @@ import           GHC.Driver.Errors.Types
 import           GHC.Types.Error                     (errMsgDiagnostic,
                                                       singleMessage)
 import           GHC.Unit.State
+import Debug.Trace (traceM)
 
 data Log
   = LogSettingInitialDynFlags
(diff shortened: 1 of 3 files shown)


wz1000:
Can you explain why the core file in the `ModSummaryResult` and `HiFileResult` are different? If one is getting updated and not the other then it seems like this is more of a dependency tracking/invalidation issue.

soulomoon:
@wz1000, The cause of this different is that `GetModSummary`'s fingerprint is irrelevant to the core file. So some of its' reverse deps are being earlycutoff even core file location changed. Is this expected ?

wz1000:
> It's not hard to change the fingerprint algorithm to reflect the core file location though.
> 
> https://github.com/haskell/haskell-language-server/blob/3747ae515045adbfd5b54bdd94b46ef0b8c2a789/ghcide/src/Development/IDE/Core/Compile.hs#L1017

Yes I think the correct solution is to include the fingerprint of the entire ModLocation in `computeFingerprint`.

This patch is a good cleanup though, so perhaps you can apply it after testing that just changing `computeFingerprint` to take the `ModLocation` into account also fixes the bug.

soulomoon:
Yes, local testing show that fingierprinting the ModLocation fix the `.hi.core` problem.

wz1000 (approved):

PR #4351, opened 2024-07-07
Support Inlay hints for record wildcards

### LINKED ISSUE #4211: Inlay hints for record wildcards

Record wildcards implicitly bind the record field names. We can use inlay hints to display that information explicitly, and triggering it could insert an explicit binding for the field.

```
f Foo{...}
```
becomes
```
f Foo{<hint>bar,</hint><hint>baz,</hint>...}
```

jetjinser:
resolve #4211.

This PR adds the feature to provide explicit records by inlay hints.

The inlay hints are located after dotdot and have left padding, It also provides textEdit, which is consistent with the textEdit of code actions.

e.g
```haskell
data MyRec = MyRec
  { foo :: Int
  , bar :: Int
  , baz :: Char
  }

convertMe :: MyRec -> String
convertMe MyRec {foo, ..<inlay> bar, baz</inlay>} = show foo ++ show bar ++ show baz
```

```
MyRec {foo, .. bar, baz}
  _paddingLeft^  | |  |
        part  |1 |2|3 |
```
part1: `bar`
part2: `, `
part3: `baz`

image

image


michaelpj (commented):
This already looks pretty good, nice work!

I think the position is fine, I'm not sure we can do much better than that. It would look weird to put it before the `...` as well. I think this way will also look better if it is split over multiple lines, e.g.
```
Foo {
  a,
  b,
  ... <hint goes here>
}
```

Do you want to investigate the richer label parts? I think that's optional but would be cool. I'd be interested to know what UI we actually get from vscode if we e.g. supply a source location for a label part...

michaelpj:
> This seems reasonable. However, when the inlay hints and definition are in the same file, vscode can correctly preview the definition in the file; but when the definition is in another file, vscode will try to use the location where the definition is located to preview it in the current file, which is obviously an incorrect location.

This seems weird. Are we definitely returning `Location`s with the correct `uri`?

> I tried to use getDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [Location]) to get the definition at dotdot

I think we might need to dig in a bit more than that. I can see that that eventually calls `locationsAtPoint` which gets the `Name`s at that location and then maps them to their locations. We could either copy what it's doing, or we could consider teaching `getDefinition` to return `(Name, Location)` pairs. That seems like it might be useful in other contexts as well, since it basically serves to disambiguate the case where we have several implicit nodes in the same location, like with `RecordWildCards`.

jetjinser:
> This seems weird. Are we definitely returning `Location`s with the correct `uri`?

I'll try again and recheck.
 
> I think we might need to dig in a bit more than that. I can see that that eventually calls `locationsAtPoint` which gets the `Name`s at that location and then maps them to their locations. We could either copy what it's doing, or we could consider teaching `getDefinition` to return `(Name, Location)` pairs. That seems like it might be useful in other contexts as well, since it basically serves to disambiguate the case where we have several implicit nodes in the same location, like with `RecordWildCards`.

Yes, this is one of my attempts... I forgot why I gave up, but it will be done soon.

jetjinser:
I pushed some commits but it's broken on GHC 9.10.1, I'm working on fixing it.

jetjinser:
I have spent a lot of time trying to fix the problem in higher GHC versions, but so far I can't figure out how to solve it.

The failed tests are [`HsExpanded1`](https://github.com/haskell/haskell-language-server/blob/96a2550e093759213affec57015f2ad91c849d23/plugins/hls-explicit-record-fields-plugin/test/testdata/HsExpanded1.hs) and [`HsExpanded2`](https://github.com/haskell/haskell-language-server/blob/96a2550e093759213affec57015f2ad91c849d23/plugins/hls-explicit-record-fields-plugin/test/testdata/HsExpanded2.hs).

I tested it in GHC 9.6.5 and 9.10.1. In 9.6.5, there are `MyRec { .. }` and `MyRec { foo = foo }` in [`tcg_binds`](https://github.com/haskell/haskell-language-server/blob/96a2550e093759213affec57015f2ad91c849d23/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs#L259), but in 9.10.1 there is only `MyRec { foo = foo }`.

`MyRec { .. }` should be processed in this function. But if there are no `MyRec { .. }` (9.10.1), no records would be found.
https://github.com/haskell/haskell-language-server/blob/96a2550e093759213affec57015f2ad91c849d23/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs#L496-L503


Please take a look, thanks! @wz1000, @fendor.

fendor:
@jetjinser Sorry for the delay! I spent a little bit of time on this and while I cannot really tell you what's going wrong, I think this is either a bug in our HLS parsing/type checking pipeline or straight-up a GHC bug.

For example, the following `Unused2.hs` works completely fine:
```haskell
-- {-# LANGUAGE RebindableSyntax #-}

module HsExpanded2 where
import Prelude

-- ifThenElse :: Int -> Int -> Int ->  Int
-- ifThenElse x y z = x + y + z

data MyRec = MyRec
  { foo :: Int }

data YourRec = YourRec
  { bar :: Bool }

myRecExample = MyRec 5

yourRecExample = YourRec True

convertMe :: Int
convertMe =
  if (let MyRec {..} = myRecExample
          YourRec {..} = yourRecExample
      in bar) then 1 else 2
 ```
 
 However, this does also not work:
 ```haskell
convertMe :: Int
convertMe =
     (let MyRec {..} = myRecExample
          YourRec {..} = yourRecExample
      in bar) + 1 + 2
```
 
 I tested the file with `ghc-exactprint` and that the AST contains these nodes (which is what we are looking for): 
 ```
                              (RecCon
                              (HsRecFields
                               []
                               (Just
                                (L
                                 (EpaSpan { 24:20-21 })
                                 (RecFieldsDotDot
                                  (0))))))))
```
That means that at least the parser is accurate.

My suggestion is to mark this test as broken for GHC 9.10.1 and move on for now.

**Comment by fendor** [comment]
I found something else, take this ast representation of the `tcg_binds`

```
Found an expr: if (let
      MyRec {..} = myRecExample
      YourRec {..} = yourRecExample
    in bar_azh) then
    1
else
    2
(L
 (EpAnn
  (EpaSpan { (23,3)-(25,27) })
  (AnnListItem
   [])
  (EpaComments
   []))
 (XExpr
  (ExpandedThingTc
   (OrigExpr
    (HsIf
     (NoExtField)
     (L
      (EpAnn
       (EpaSpan { (23,6)-(25,13) })
       (AnnListItem
        [])
       (EpaComments
        []))
      (HsPar
       (NoExtField)
       (L
        (EpAnn
         (EpaSpan { (23,7)-(25,12) })
         (AnnListItem
          [])
         (EpaComments
          []))
        (HsLet
         (NoExtField)
         (HsValBinds
          (EpAnn
           (EpaSpan { (23,11)-(24,39) })
           (AnnList
            (Just
             (EpaSpan { (23,11)-(24,39) }))
            (Nothing)
            (Nothing)
            []
            [])
           (EpaComments
            []))
          (XValBindsLR
           (NValBinds
            [((,)
              (NonRecursive)
              {Bag(LocatedA (HsBind Name)):
               [(L
                 (EpAnn
                  (EpaSpan { 24:11-39 })
                  (AnnListItem
                   [])
                  (EpaComments
                   []))
                 (PatBind
                  {NameSet:
                   [{Name: YourRec}
                   ,{Name: bar}
                   ,{Name: yourRecExample}]}
                  (L
                   (EpAnn
                    (EpaSpan { 24:11-22 })
                    (AnnListItem
                     [])
                    (EpaComments
                     []))
                   (ConPat
                    (NoExtField)
                    (L
                     (EpAnn
                      (EpaSpan { 24:11-17 })
                      (NameAnnTrailing
                       [])
                      (EpaComments
                       []))
                     {Name: YourRec})
                    (RecCon
                     (HsRecFields
                      [(L
                        (EpAnn
                         (EpaSpan { 24:20-21 })
                         (AnnListItem
                          [])
                         (EpaComments
                          []))
                        (HsFieldBind
                         []
                         (L
                          (EpAnn
                           (EpaSpan { 24:20-21 })
                           (AnnListItem
                            [])
                           (EpaComments
                            []))
                          (FieldOcc
                           {Name: bar}
                           (L
                            (EpAnn
                             (EpaSpan { 24:20-21 })
                             (NameAnnTrailing
                              [])
                             (EpaComments
                              []))
                            (Unqual
                             {OccName: bar}))))
                         (L
                          (EpAnn
                           (EpaSpan { 24:20-21 })
                           (AnnListItem
                            [])
                           (EpaComments
                            []))
                          (VarPat
                           (NoExtField)
                           (L
                            (EpAnn
                             (EpaSpan { 24:20-21 })
                             (NameAnnTrailing
                              [])
                             (EpaComments
                              []))
                            {Name: bar_azh})))
                         (False)))]
                      (Just
                       (L
                        (EpaSpan { 24:20-21 })
                        (RecFieldsDotDot
                         (0))))))))
```

It looks like `getRecCons (unLoc -> XExpr (ExpandedThingTc a _)) = (collectRecords a, True)` doesn't quite work as intended. Changing it to `collectRecords a, False)` seems to be working.

File: plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs [hunk 1]
```diff
@@ -144,17 +166,71 @@ codeActionResolveProvider ideState pId ca uri uid = do
...
+           textEdits = maybeToList (renderRecordInfoAsTextEdit nameMap record)
+                    <> maybeToList (pragmaEdit enabledExtensions pragma)
+           names = renderRecordInfoAsLabelName record
+       in do
+         end <- fmap _end range
+         names' <- names
+         defnLocs' <- defnLocs
+         let excludeDotDot (Location _ (Range _ pos)) = pos /= end
+             -- find location from dotdot definitions that name equal to label name
+             findLocation t = fmap fst . find (either (const False) ((==) t) . snd) . filter (excludeDotDot . fst)
```
fendor, line 207: This feels hard to read, could you add explicit parameters and intermediate results?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -249,8 +321,8 @@ instance Show CollectNamesResult where
 type instance RuleResult CollectNames = CollectNamesResult
 
 data RecordInfo
-  = RecordInfoPat RealSrcSpan (Pat (GhcPass 'Renamed))
-  | RecordInfoCon RealSrcSpan (HsExpr (GhcPass 'Renamed))
+  = RecordInfoPat RealSrcSpan (Pat GhcTc)
+  | RecordInfoCon RealSrcSpan (HsExpr GhcTc)
```
fendor, line 325: Ah, does that mean this plugin now only works if the source file can be typechecked? And previously, it already worked if the source was renamed?

fendor (changes requested):
Thanks, this looks quite nice already! Some small nitpicks and proposed changes :)

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -333,17 +415,47 @@ preprocessRecord getName names flds = flds { rec_dotdot = Nothing , rec_flds = r
...
+    puns' = map (mapLoc (\fld -> fld { hfbPun = True })) puns
+
+
+showRecordPat :: Outputable (Pat GhcTc) => UniqFM Name [Name] -> Pat GhcTc -> Maybe Text
 showRecordPat names = fmap printOutputable . mapConPatDetail (\case
   RecCon flds -> Just $ RecCon (preprocessRecordPat names flds)
   _           -> Nothing)
 
+showRecordPatFlds :: Pat GhcTc -> Maybe [Name]
+showRecordPatFlds (ConPat _ _ args) = fmap (fmap ((\case FieldOcc x _ -> getName x) . unLoc . hfbLHS . unLoc) . rec_flds) (m args)
```
fendor, line 439: Maybe a couple of newlines and intermediate results, maybe?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -333,17 +415,47 @@ preprocessRecord getName names flds = flds { rec_dotdot = Nothing , rec_flds = r
...
+  where
+    m (RecCon flds) = Just $ processRecordFlds flds
+    m _             = Nothing
```
fendor, line 442: Proper name for `m` please :)

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -333,17 +415,47 @@ preprocessRecord getName names flds = flds { rec_dotdot = Nothing , rec_flds = r
...
 showRecordCon expr@(RecordCon _ _ flds) =
   Just $ printOutputable $
     expr { rcon_flds = preprocessRecordCon flds }
 showRecordCon _ = Nothing
 
+showRecordConFlds :: p ~ GhcTc => HsExpr p -> Maybe [Name]
+showRecordConFlds (RecordCon _ _ flds) = mapM (m . unLoc . hfbRHS . unLoc) (rec_flds $ processRecordFlds flds)
+  where
+    m (HsVar _ lidp) = Just $ getName lidp
+    m _              = Nothing
```
fendor, line 455: here as well

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
jetjinser: Yes

fendor (approved):
This looks good to me now :)

michaelpj (approved):

Issue #3853, opened 2023-10-25
Use cabal-add to fill missing build-depends automatically


Bodigrim:
When HLS gets an error "module `Bla.Bla.Bla` is a member of a hidden package `bla`" it should suggest a quick fix to insert the dependency on `bla` into Cabal file using https://github.com/Bodigrim/cabal-add and carry on. 

`cabal-add` draft API can be examined here: https://hackage.haskell.org/package/cabal-add-0.1/candidate/docs/Distribution-Client-Add.html. My impression is that HLS should be very much capable to create [`Config`](https://hackage.haskell.org/package/cabal-add-0.1/candidate/docs/Distribution-Client-Add.html#t:Config) and write the result of [`executeConfig`](https://hackage.haskell.org/package/cabal-add-0.1/candidate/docs/Distribution-Client-Add.html#v:executeConfig) to the disk. 

CC @VeryMilkyJoe on behalf of `hls-cabal-plugin`.

georgefst:
This has been on the todo list for a _long_ time: https://github.com/haskell/haskell-language-server/issues/155 (but maybe `cabal-add` is the solution!).

Bodigrim:
`cabal-add` preserves the formatting of the original file. I tested it on complete set of Cabal files from Hackage.

fendor:
Adding `cabal-add` should be a matter of parsing the error message, guessing the relevant `.cabal` file and then running `cabal-add` on it. #3778 can be used as an inspiration for this particular code action.
Nothing difficult about it, just someone needs to take the time to implement it!

fendor:
Implemented by #4360

Issue #3841, opened 2023-10-12
evaluate plugin error  hPutChar: invalid argument (cannot encode character '\8216')


Delli:
### Your environment

Which OS do you use?
windows 10 

Which version of GHC do you use and how did you install it?
ghc 9.8.1      / tried other versions 9.2.2  same issue

How is your project built (alternative: link to the project)?
stack build 

Which LSP client (editor/plugin) do you use?

vscode latest  + extension Haskell v2.4.1

Which version of HLS do you use and how did you install it?
hls v 2.4.0.0
ghcup installed it 

Have you configured HLS in any way (especially: a `hie.yaml` file)?
no 

### Steps to reproduce

stack new project 

in any hs file 

try add evaluation 
{-| double 
>>> double 2
<stderr>: hPutChar: invalid argument (cannot encode character '\8216')
-}

double :: Num a => a -> a
double x = (2*x)

### Expected behaviour

{-| double 
>>> double 2
4
-}

### Actual behaviour

{-| double 
>>> double 2
<stderr>: hPutChar: invalid argument (cannot encode character '\8216')
-}

Bodigrim:
Year 2023, Unicode in Windows strikes again. 

This is likely https://gitlab.haskell.org/ghc/ghc/-/commit/74132c2b0992849f83ef87c8a56ac3975738e767

fendor:
Thank you for the context!

Does that mean, if we set the environment variable `GHC_NO_UNICODE` in HLS, these kinds of issues shouldn't arise?
Or should we mirror ghc's logic in determining the unicode support and only then set the environment variable?

Delli:
thx , confirm solved the issue on my machine
@ fendor what i did  
![image](https://github.com/haskell/haskell-language-server/assets/4728038/1701cebb-4c31-494a-9321-500061d4dd85)
set system variable GHC_CHARENC to value  GHC_NO_UNICODE in windows os

Delli:
@fendor  

To set the environment variable GHC_CHARENC for GHC, you can use the following steps:

Windows:

Open the Control Panel.
Click on "System and Security".
Click on "System".
Click on "Advanced system settings".
In the "System Properties" window, click on the "Environment Variables" button.
In the "Environment Variables" window, under "System variables", find the variable GHC_CHARENC. If it does not exist, click on "New" and create it.
Set the value of the variable to the desired character encoding.
Click on "OK" to save the changes.

Bodigrim:
[...]

PR #4360, opened 2024-07-12
cabal-add integration as a CodeAction

VenInf:
### This is a proof-of-concept solution for the issue #3853

If HLS detects a message like  "module `Bla.Bla.Bla` is a member of a hidden package `bla-1.2.3`" it suggests a quick fix, that finds the closest cabal file and adds the dependency there.

Solution uses [`Distribution.Client.Add`](https://hackage.haskell.org/package/cabal-add-0.1/candidate/docs/Distribution-Client-Add.html) from the `cabal-add` and automaticly adds version requirement, if it's detected.

CC @Bodigrim

Video with an example:

https://github.com/user-attachments/assets/02be3290-e25c-4d39-a299-ba135b274218

### Implementation details

For now, the `cabal-add` project was linked using [remote package specification](https://cabal.readthedocs.io/en/3.4/cabal-project.html#specifying-packages-from-remote-version-control-locations). Some parts were heavily inspired by the `cabal-add` code in the main module and might be rewritten later. 

`CodeAction` works by parsing all haskell diagnostics, and is constructed if a suited message was found. Parsed information is passed down to a new command, which itself uses tools provided by `cabal-add`. The command conducts IO actions with found cabal file. 

### What isn't impemented
 - [x] automatic tests, and existing ones lack variety (Since the error is not documented it's hard to make it's different versions. Also GHC(?) does not register that .cabal file was changed and still shows error)
 - [x] no recorders
 - [ ] no action for "missing dependencies" error, since I couldn't recreate it
 - [ ] no multiple packages suggestion


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,160 @@
...
+                                                         takeExtension, (</>))
+import           Text.Regex.TDFA
+
+
+-- | Given a path to a haskell file, finds all cabal files paths
+--   sorted from the closest to the farthest.
+--   Gives all found paths all the way to the root directory.
+findResponsibleCabalFile :: FilePath -> IO [FilePath]
+findResponsibleCabalFile uriPath = do
+  contents <- mapM listDirectory allDirPaths
```
Bodigrim, line 59: File IO is expensive, I'd suggest against travesing everything up to the very top. You can spray it generously with `unsafeInterleveIO`, or rewrite more explicitly with checking each level before proceeding to scan an upper one.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,160 @@
...
+        command = mkLspCommand plId (CommandId cabalAddCommand) "Execute Code Action" (Just [toJSON params])
+      in CodeAction title (Just CodeActionKind_QuickFix) (Just []) Nothing Nothing Nothing (Just command) Nothing
+
+-- | Gives a mentioned number of hidden packages given
+--   a specific error message
+missingDependenciesSuggestion :: Int -> T.Text -> [(T.Text, T.Text)]
+missingDependenciesSuggestion maxCompletions msg = take maxCompletions $ getMatch (msg =~ regex)
+  where
+    regex :: T.Text -- TODO: Support multiple packages suggestion
+    regex = "Could not load module \8216.*\8217.\nIt is a member of the hidden package \8216([a-z]+)[-]?([0-9\\.]*)\8217"
```
Bodigrim, line 93: Depending on environment variables, GHC can use `\8216...\8217`, or simply `'...'`. It would be prudent to match on both.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,160 @@
...
+
+
+-- | Given a path to a haskell file, finds all cabal files paths
+--   sorted from the closest to the farthest.
+--   Gives all found paths all the way to the root directory.
+findResponsibleCabalFile :: FilePath -> IO [FilePath]
+findResponsibleCabalFile uriPath = do
+  contents <- mapM listDirectory allDirPaths
+  let filesWithPaths = concat $ zipWith (\path content -> map (path </>) content) allDirPaths contents
+  let cabalFiles = filter (\c -> takeExtension c == ".cabal") filesWithPaths
```
Bodigrim, line 61: I'd check that `cabalFiles` are files and not folders.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,160 @@
...
+
+-- | Constructs prerequisets for the @executeConfig@
+--   and runs it, given path to the cabal file and
+--   a dependency message.
+--
+--   Inspired by @main@ in cabal-add,
+--   Distribution.Client.Main
+addDependency :: FilePath -> NonEmpty String -> IO ()
+addDependency cabalFilePath dependency = do
+  let rcnfComponent = Nothing
```
Bodigrim, line 144: This is a good first step, but it would be nice to cover components other than the main library. Is there a way to ask Cabal which component a certain file belongs to?

Bodigrim (commented):
Thanks for working on this!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: Ok, I will fix that. Is there a way to alter the environment to test the second case? On my system, I get the first one by default.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: Personally, I don't know any convenient tools like this.
I think it can be done by parsing the cabal file and looking, where its module name is mentioned. But this approach will not be trivial to implement.

Bodigrim: Could https://hackage.haskell.org/package/Cabal-3.12.1.0/docs/Distribution-Simple-BuildTarget.html#t:BuildTarget be of any help? Presumably it should be possibly to pass file name to `readBuildTargets` (?) and get `BuildTargetModule ComponentName ModuleName`, where `ComponentName` is what we are looking for.

fendor: This is a great suggestion and seems to be working on small projects at least!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
fendor: Try running VSCode with the environment variable `GHC_NO_UNICODE=1`.

See https://github.com/haskell/haskell-language-server/issues/3841#issuecomment-1762048408

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
michaelpj: I also wonder if we don't know this already. Do we not know what component a file belongs to, and thus hopefully what cabal file is relevant? At the very least we could compute it with a Rule.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: I am working on the integration of this function, but I faced some problems.
Unfortunately, `readBuildTargets` does not resolve them completely and makes it impossible to use the found target.

For example, if there is a library and an executable with the same name, it will not resolve them into `["exe:common-name", "lib:common-name"]` and will just give `["common-name"]`. (oversimplified demonstration).

It will be impossible to use just "common-name" since it's not resolved.

Is there a way to just get all the targets? Or a way to resolve them? That would be very convenient.
I think it's possible to bruteforce them, but I would rather avoid it.

Bodigrim: @fendor @michaelpj is there any ready-to-use machinery for this in HLS? I'd imagine determining component by filename is a common task.

Catching `VerboseException` is not too bad, I think.

fendor: >  I'd imagine determining component by filename is a common task.

No, as HLS is build-tool independent, there are very few locations where we handle build-tools differently and the hls-cabal-plugin is young by comparison to the other infrastructure.

fendor: I discussed with @VenInf a possible solution

VenInf:
Have made some significant changes to the solution.
- Integrated `readBuildTargets`. It doesn't give all the targets in the cabal file, so the result of the action is similar. But it supports multiple targets, if there will be a way to get them.
- Now sometimes `genericPackageDescription` is received via cabal rule.There are a lot of IO actions in the solution, and it will be hard/unnecessary(?) to add them all.
- General refactoring

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,185 @@
...
+--   Gives all found paths all the way to the root directory.
+findResponsibleCabalFile :: FilePath -> IO [FilePath]
+findResponsibleCabalFile haskellFilePath = do
+  contents <- mapM (unsafeInterleaveIO . listDirectory) allDirPaths
+  let objectWithPaths = concat $ zipWith (\path content -> map (path </>) content) allDirPaths contents
+  let objectCabalExtension = filter (\c -> takeExtension c == ".cabal") objectWithPaths
+  cabalFiles <- filterM (\c -> doesFileExist c) objectCabalExtension
+  pure $ reverse cabalFiles -- sorted from closest to the haskellFilePath
+  where dirPath = dropFileName haskellFilePath
+        allDirPaths = scanl1 (</>) (splitPath dirPath)
```
Bodigrim, line 78: The function as written will `listDirectory` for all parent folders up to the root, then pick up the closest one. `listDirectory` could be extremely expensive, we should strive to do as little IO as possible. 

At the very least, let's reverse `allDirPaths` upfront instead of reversing `cabalFiles` after all `listDirectory` and `filterM` was done. But I would recommend to rewrite this function with explicit recursion, stripping components of `haskellFilePath` on each revursive call.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,185 @@
...
+        command = mkLspCommand plId (CommandId cabalAddCommand) "Execute Code Action" (Just [toJSON params])
+      in CodeAction title (Just CodeActionKind_QuickFix) (Just []) Nothing Nothing Nothing (Just command) Nothing
+
+-- | Gives a mentioned number of hidden packages given
+--   a specific error message
+hiddenPackageSuggestion :: Int -> T.Text -> [(T.Text, T.Text)]
+hiddenPackageSuggestion maxCompletions msg = take maxCompletions $ getMatch (msg =~ regex)
+  where
+    regex :: T.Text -- TODO: Support multiple packages suggestion
+    regex = "Could not load module \8216.*\8217.\nIt is a member of the hidden package [\8216\\']([a-z]+)[-]?([0-9\\.]*)[\8217\\']"
```
Bodigrim, line 114: Why `[\8216\\']` instead of `[\8216']`? 

`\8216.*\8217` should also account for non-Unicode quotes.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,185 @@
...
+
+-- | Constructs prerequisets for the @executeConfig@
+--   and runs it, given path to the cabal file and a dependency message.
+--
+--   Inspired by @main@ in cabal-add,
+--   Distribution.Client.Main
+addDependency :: FilePath -> Maybe String -> NonEmpty String -> IO ()
+addDependency cabalFilePath buildTarget dependency = do
+
+  cnfOrigContents <- readCabalFile cabalFilePath
```
Bodigrim, line 166: I don't know if HLS does it in other cases, but it could be a good idea to lock the handle using https://hackage.haskell.org/package/lukko-0.1.2/docs/Lukko.html#v:hLock.

Bodigrim: Or without `lukko`, open file for read/write access and manipulate it without closing?.. Or am I talking nonsense?..

fendor: Instead of manually reading and parsing, use https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs#L260, which hooks into the rule system.

E.g. 
```haskell
liftIO $ runAction "cabal-plugin.cabal-add" ide $ useWithStale ParseCabalFile $ toNormalizedFilePath $ toNormalizedFilePath cabalFilePath
```
This will give you the GPD already parsed and cached.

fendor: Or if you need the fields, the original link will give you that

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: That's an escape for the `'` symbol in this flavour or [regex](https://hackage.haskell.org/package/regex-tdfa-1.3.2.2/docs/Text-Regex-TDFA.html). Or at the very least an attempt to make one.

I would rather not try to catch all symbols in this regex, since I am looking for the package and a version separately, and add the version to the dependency if it's found.

Bodigrim: You don't normally need to escape quotes inside regexps, so I imagine `[\8216']` should work as well.

My point was that the module name could be `\8216.*\8217` or `'.*'`, we need to account for both possibilities.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: Since the `handle` using approach is not very in the style of hls I used `runAction` approach instead. It falls back on reading files, if the data wasn't found in rules.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: Rewrote the function to an explicit recursion.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
jhrcek: I think this is not about module names, but about quotes that GHC uses to surround the module name in the error message.
This particular error is generated [here](https://gitlab.haskell.org/ghc/ghc/-/blob/b57792a804ad71986238467abe6917406ffaeae5/compiler/GHC/Iface/Errors/Ppr.hs#L169) and the [quote](https://gitlab.haskell.org/ghc/ghc/-/blob/b57792a804ad71986238467abe6917406ffaeae5/compiler/GHC/Utils/Outputable.hs#L747) function used there can output both ascii quotes as well as unicode quotes.
So to make this regex robust, I think we should support both.

fendor: Ah right, I misunderstood, thanks!

VenInf:
Have made a lot of changes to the solution:
- Refactoring and fixed some naming issues
- Write to a cabal file was changed to a comprehensive `edit`
- Added logs
- Tests (Many thanks for helping out with them to the @fendor)
- Others

plugins/hls-cabal-plugin/test/Main.hs
```diff
@@ -220,10 +223,108 @@ codeActionTests = testGroup "Code Actions"
...
+                                   ]
     ]
   where
     getLicenseAction :: T.Text -> [Command |? CodeAction] -> [CodeAction]
     getLicenseAction license codeActions = do
         InR action@CodeAction{_title} <- codeActions
         guard (_title == "Replace with " <> license)
         pure action
+    generateHiddenPackageTestSession :: FilePath -> FilePath -> T.Text -> [Int] -> Session ()
+    generateHiddenPackageTestSession cabalFile haskellFile dependency indicesRes = do
```
VeryMilkyJoe, line 314:
```suggestion
    suggestAddDependencyTestSession :: FilePath -> FilePath -> T.Text -> [Int] -> Session ()
    suggestAddDependencyTestSession cabalFile haskellFile dependency indicesRes = do
```

VeryMilkyJoe (changes requested):
Thank you for this awesome feature, I just have a few nitpicks and will check it out and manually test a little bit :)

plugins/hls-cabal-plugin/test/Main.hs
```diff
@@ -220,10 +223,108 @@ codeActionTests = testGroup "Code Actions"
...
+                                   , ("AWin32Console", "1.19.1")
+                                   ]
+    , expectFailBecause "TODO fix regex for these cases" $
+      testHiddenPackageSuggestions "Check CabalAdd's parser, with version, unicode comma"
+                                   [ "It is a member of the hidden package \82163d-graphics-examples\8217"
+                                   , "It is a member of the hidden package \82163d-graphics-examples-1.1.6\8217"
+                                   ]
+                                   [ ("3d-graphics-examples", T.empty)
+                                   , ("3d-graphics-examples", "1.1.6")
+                                   ]
```
VeryMilkyJoe, line 305: I think these should be moved to a different file, since they are simple unit tests.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,264 @@
...
+      LogCreatedEdit edit -> "Created inplace edit:\n" Logger.<+> Logger.pretty edit
+      LogExecutedCommand -> "Executed CabalAdd command"
+
+
+-- | Gives a code action that calls the command,
+--   if a suggestion for a missing dependency is found.
+--   Disabled action if no cabal files given.
+--   Conducts IO action on a cabal file to find build targets.
+hiddenPackageAction :: Logger.Recorder (Logger.WithPriority Log) -> PluginId -> VersionedTextDocumentIdentifier -> Int -> Diagnostic -> FilePath -> FilePath -> GenericPackageDescription -> IO [CodeAction]
+hiddenPackageAction recorder plId verTxtDocId maxCompletions diag haskellFilePath cabalFilePath gpd = do
```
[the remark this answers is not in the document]
VenInf: I hoped that later other related errors will have their own actions, for example a `missing module` one. But maybe this is a better naming choice.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -82,6 +86,23 @@ instance Pretty Log where
         <+> "for cursor position:"
         <+> pretty position
     LogCompletions logs -> pretty logs
+    LogCabalAdd logs -> pretty logs
+
+
+haskellFilesDescriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
```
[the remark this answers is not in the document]
VenInf: I added this descriptor because the `descriptor` below is related to the cabal files and works on them only. Since the `cabalAddCodeAction` is called from a haskell file it requires a separate descriptor.

I will clarify this in the documentation, thank you.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -82,6 +86,24 @@ instance Pretty Log where
         <+> "for cursor position:"
         <+> pretty position
     LogCompletions logs -> pretty logs
+    LogCabalAdd logs -> pretty logs
+
+-- | Some actions with cabal files originate from haskell files.
+-- This descriptor is needed to handle these cases.
```
fendor, line 92:
```suggestion
-- | Some actions with cabal files originate from haskell files.
-- This descriptor allows to hook into the diagnostics of haskell source files, and
-- allows us to provide code actions and commands that interact with `.cabal` files.
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -82,6 +86,24 @@ instance Pretty Log where
...
+haskellInteractionDescriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
+haskellInteractionDescriptor recorder plId =
+  (defaultPluginDescriptor plId "Provides the cabal-add code action in haskell files")
+    { pluginHandlers =
+        mconcat
+          [ mkPluginHandler LSP.SMethod_TextDocumentCodeAction $ cabalAddCodeAction recorder
+          ]
+    , pluginCommands = [PluginCommand CabalAdd.cabalAddCommand "add a dependency to a cabal file" (CabalAdd.command cabalAddRecorder)]
+    , pluginRules = pure () -- TODO: change to haskell files only (?)
```
fendor, line 101:
```suggestion
    , pluginRules = pure ()
```

this is fine as is.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -82,6 +86,24 @@ instance Pretty Log where
         <+> "for cursor position:"
         <+> pretty position
     LogCompletions logs -> pretty logs
+    LogCabalAdd logs -> pretty logs
+
+-- | Some actions with cabal files originate from haskell files.
+-- This descriptor is needed to handle these cases.
+haskellInteractionDescriptor :: Recorder (WithPriority Log) -> PluginId -> PluginDescriptor IdeState
+haskellInteractionDescriptor recorder plId =
+  (defaultPluginDescriptor plId "Provides the cabal-add code action in haskell files")
```
fendor, line 95: Change the description to be slightly more generic, this descriptor will not always only work for `cabal-add` only.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +299,33 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
       let completionTexts = fmap (^. JL.label) completions
       pure $ FieldSuggest.fieldErrorAction uri fieldName completionTexts _range
 
+cabalAddCodeAction :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState 'LSP.Method_TextDocumentCodeAction
+cabalAddCodeAction recorder state plId (CodeActionParams _ _ (TextDocumentIdentifier uri) _ CodeActionContext{_diagnostics=diags}) = do
+  maxCompls <- fmap maxCompletions . liftIO $ runAction "cabal.cabal-add" state getClientConfigAction
+  let mbHaskellFilePath = uriToFilePath uri
+  case mbHaskellFilePath of
```
fendor, line 306:
```suggestion
  case uriToFilePath uri of
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +299,33 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+    Nothing -> pure $ InL []
+    Just haskellFilePath -> do
+      mbCabalFile <- liftIO $ CabalAdd.findResponsibleCabalFile haskellFilePath
+      case mbCabalFile of
+        Nothing -> pure $ InL $ fmap InR [noCabalFileAction]
```
fendor, line 311:
```suggestion
        Nothing -> pure $ InL [InR noCabalFileAction]
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +299,33 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
       let completionTexts = fmap (^. JL.label) completions
       pure $ FieldSuggest.fieldErrorAction uri fieldName completionTexts _range
 
+cabalAddCodeAction :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState 'LSP.Method_TextDocumentCodeAction
+cabalAddCodeAction recorder state plId (CodeActionParams _ _ (TextDocumentIdentifier uri) _ CodeActionContext{_diagnostics=diags}) = do
```
fendor, line 303: I think this code action requires a little bit of code re-organisation.

The issue is that some expensive operations are executed, even though we don't know whether we even need them. 

The current structure is:
1. Find a .cabal file for a given filepath
2. Parse said cabal file
3. You try to guess the component
4. Only now you check whether the diagnostic is even a hidden package description!

This means, for each button click where the client asks for code actions, we perform the steps (1)-(3). Please change this such that (4) happens first :)

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+    "CabalAdd parameters:\n" Logger.<+>
+    "| cabal path: " Logger.<+> Logger.pretty cabalPath Logger.<+> "\n" Logger.<+>
+    "| target: " Logger.<+> Logger.pretty buildTarget Logger.<+> "\n" Logger.<+>
+    "| dependendency: " Logger.<+> Logger.pretty dependency Logger.<+> "\n" Logger.<+>
+    "| version: " Logger.<+> Logger.pretty version Logger.<+> "\n"
+
+-- | Gives a code action that calls the command,
+--   if a suggestion for a missing dependency is found.
+--   Disabled action if no cabal files given.
+--   Conducts IO action on a cabal file to find build targets.
```
fendor, line 126: Comment is a bit ambiguous, what's `the command`?

Hard to read, could expand on this a little bit more?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+  deriving (Generic, Show)
+  deriving anyclass (FromJSON, ToJSON)
+
+instance Logger.Pretty CabalAddCommandParams where
+  pretty CabalAddCommandParams{..} =
+    "CabalAdd parameters:\n" Logger.<+>
+    "| cabal path: " Logger.<+> Logger.pretty cabalPath Logger.<+> "\n" Logger.<+>
+    "| target: " Logger.<+> Logger.pretty buildTarget Logger.<+> "\n" Logger.<+>
+    "| dependendency: " Logger.<+> Logger.pretty dependency Logger.<+> "\n" Logger.<+>
+    "| version: " Logger.<+> Logger.pretty version Logger.<+> "\n"
```
fendor, line 121: In the logs, this is a bit annoying to read :D 

```suggestion
    "CabalAdd parameters:" <+> vcat
      [ "cabal path:" <+> pretty cabalPath
      , "target:" <+> pretty buildTarget
      , "dependendency:" <+> pretty dependency
      , "version:" <+> pretty version
      ]
```

And I recommend to import
```
import Ide.Logger
import qualified Distribution.Pretty                           as Pretty
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+      [] -> pure $ mkCodeAction cabalFilePath Nothing <$> hiddenPackageSuggestion maxCompletions (_message diag)
+      targets -> pure $ concat [mkCodeAction cabalFilePath (Just $ buildTargetToStringRepr target) <$>
+                          hiddenPackageSuggestion maxCompletions (_message diag) | target <- targets]
+  where
+    buildTargetToStringRepr target = render $ pretty $ buildTargetComponentName target
+
+    getBuildTargets :: GenericPackageDescription -> FilePath -> FilePath -> IO [BuildTarget]
+    getBuildTargets gpd cabalFilePath haskellFilePath = do
+      let haskellFileRelativePath = makeRelative (dropFileName cabalFilePath) haskellFilePath
+      readBuildTargets (verboseNoStderr silent) (flattenPackageDescription gpd) [haskellFileRelativePath]
```
fendor, line 150: Comment please why you call `makeRelative` here! I know you know, but the next developer won't know ;)

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+          Just t  -> "  target " <> T.pack t
+        title = "Add dependency " <> suggestedDep <> versionTitle <> targetTitle
+        version = if T.null suggestedVersion then Nothing else Just suggestedVersion
+
+        params = CabalAddCommandParams {cabalPath = cabalFilePath
+                                      , verTxtDocId = verTxtDocId
+                                      , buildTarget = target
+                                      , dependency = suggestedDep
+                                      , version=version}
+        command = mkLspCommand plId (CommandId cabalAddCommand) "Execute Code Action" (Just [toJSON params])
```
fendor, line 167:
```suggestion
        command = mkLspCommand plId (CommandId cabalAddCommand) "Add missing dependency" (Just [toJSON params])
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+      in CodeAction title (Just CodeActionKind_QuickFix) (Just []) Nothing Nothing Nothing (Just command) Nothing
+
+-- | Gives a mentioned number of @(dependency, version)@ pairs
+--   found in the "hidden package" message
```
fendor, line 171:
```suggestion
-- | Given a ghc error message that looks like this:
-- 
-- @
-- Could not load module ‘Data.List.Split’
-- It is a member of the hidden package ‘split-0.2.5’.
-- Perhaps you need to add ‘split’ to the build-depends in your .cabal file.
-- @
-- 
-- Extract all mentioned package names and version numbers. In this example, we extract `[("split", "0.2.5")]`
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+--   if a suggestion for a missing dependency is found.
+--   Disabled action if no cabal files given.
+--   Conducts IO action on a cabal file to find build targets.
+addDependencySuggestCodeAction
+  :: Logger.Recorder (Logger.WithPriority Log)
+  -> PluginId
+  -> VersionedTextDocumentIdentifier -- ^ Cabal's versioned text identifier
+  -> Int -- ^ Maximum number of suggestions to return
+  -> Diagnostic -- ^ Diagnostic from a code action
+  -> FilePath -- ^ Path to the haskell file
```
fendor, line 133: Please clarify, which "haskell file"? The one the diagnostic originates from, right?  Add that to the documentation!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+--   Inspired by @main@ in cabal-add,
+--   Distribution.Client.Main
+getDependencyEdit :: MonadIO m => Logger.Recorder (Logger.WithPriority Log) -> (IdeState, ClientCapabilities, VersionedTextDocumentIdentifier) ->
+    FilePath -> Maybe String -> NonEmpty String -> ExceptT PluginError m WorkspaceEdit
+getDependencyEdit recorder env cabalFilePath buildTarget dependency = do
+  let (state, caps, verTxtDocId) = env
+  (mbCnfOrigContents, mbFields, mbPackDescr) <- liftIO $ runAction "cabal.cabal-add" state $ do
+        contents <- Development.IDE.useWithStale GetFileContents $ toNormalizedFilePath cabalFilePath
+        inFields <- Development.IDE.useWithStale ParseCabalFields $ toNormalizedFilePath cabalFilePath
+        inPackDescr <- Development.IDE.useWithStale ParseCabalFile $ toNormalizedFilePath cabalFilePath
```
fendor, line 209:
```suggestion
  runActionE "cabal.cabal-add" state $ do
          contents <- useWithStaleE GetFileContents $ toNormalizedFilePath cabalFilePath
          inFields <- useWithStaleE ParseCabalFields $ toNormalizedFilePath cabalFilePath
          inPackDescr <- useWithStaleE ParseCabalFile $ toNormalizedFilePath cabalFilePath
```

Functions from `PluginUtils` are generally useful

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+      let objectsWithPaths = map (\obj -> path <> obj) objects
+          objectsCabalExtension = filter (\c -> takeExtension c == ".cabal") objectsWithPaths
+      cabalFiles <- filterM (\c -> doesFileExist c) objectsCabalExtension
+      case safeHead cabalFiles of
+        Nothing        -> go ps
+        Just cabalFile -> pure $ Just cabalFile
+
+-- | Gives cabal file's contents or throws error.
+--   Inspired by @readCabalFile@ in cabal-add,
+--   Distribution.Client.Main
```
fendor, line 266: Please add this is a *fallback* path and not how you should read the .cabal file in the average case.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,272 @@
...
+    -- Have to do this matching because `Regex.TDFA` doesn't(?) support
+    -- not-capturing groups like (?:message)
+    getMatch :: (T.Text, T.Text, T.Text, [T.Text]) -> [(T.Text, T.Text)]
+    getMatch (_, _, _, []) = []
+    getMatch (_, _, _, [dependency, _, cleanVersion]) = [(dependency, cleanVersion)]
+    getMatch (_, _, _, _) = error "Impossible pattern matching case"
+
+command :: Logger.Recorder (Logger.WithPriority Log) -> CommandFunction IdeState CabalAddCommandParams
+command recorder state _ params@(CabalAddCommandParams {cabalPath = path, verTxtDocId = verTxtDocId, buildTarget = target, dependency = dep, version = mbVer}) = do
+  Logger.logWith recorder Logger.Info $ LogCalledCabalAddCommand params
```
fendor, line 186: This should be a `Debug` level.

fendor (changes requested):
Very nice, just a couple of changes, then this is fine by me :)

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
VenInf: Oh, good catch! I have missed that, thank you.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
VenInf: Forgot about this hurdle. I will make notes about it.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +300,35 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+  let suggestions = concatMap (\diag -> CabalAdd.hiddenPackageSuggestion maxCompls diag) diags
+  case suggestions of
+    [] -> pure $ InL []
+    _ ->
+      case uriToFilePath uri of
+        Nothing -> pure $ InL []
+        Just haskellFilePath -> do
+          mbCabalFile <- liftIO $ CabalAdd.findResponsibleCabalFile haskellFilePath
+          case mbCabalFile of
+            Nothing -> pure $ InL [InR noCabalFileAction]
```
fendor, line 315: You shouldn't offer an empty code action, imo. Just produce `InL []`

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +300,35 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
       let completionTexts = fmap (^. JL.label) completions
       pure $ FieldSuggest.fieldErrorAction uri fieldName completionTexts _range
 
+cabalAddCodeAction :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState 'LSP.Method_TextDocumentCodeAction
+cabalAddCodeAction recorder state plId (CodeActionParams _ _ (TextDocumentIdentifier uri) _ CodeActionContext{_diagnostics=diags}) = do
+  maxCompls <- fmap maxCompletions . liftIO $ runAction "cabal.cabal-add" state getClientConfigAction
+  let suggestions = concatMap (\diag -> CabalAdd.hiddenPackageSuggestion maxCompls diag) diags
```
fendor, line 306:
```suggestion
  let suggestions = take maxCompls $ concatMap (\diag -> CabalAdd.hiddenPackageSuggestion maxCompls diag) diags
```

Otherwise, you may produce `maxCompls * length diags` number of code actions. Or is that fine?
If you call `take maxCompls` here, you can drop it from `hiddenPackageSuggestion`.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,310 @@
...
+  -> PluginId
+  -> VersionedTextDocumentIdentifier -- ^ Cabal's versioned text identifier
+  -> [(T.Text, T.Text)] -- ^ A dependency-version suggestion pairs
+  -> FilePath -- ^ Path to the haskell file (source of diagnostics)
+  -> FilePath -- ^ Path to the cabal file (that will be edited)
+  -> GenericPackageDescription
+  -> IO [CodeAction]
+addDependencySuggestCodeAction recorder plId verTxtDocId suggestions haskellFilePath cabalFilePath gpd = do
+    buildTargets <- liftIO $ getBuildTargets gpd cabalFilePath haskellFilePath
+    logWith recorder Info LogCalledCabalAddCodeAction
```
fendor, line 146: Delete this kind of log message, I don't think it adds any value beyond this point.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,310 @@
...
+                                                                makeRelative,
+                                                                splitPath,
+                                                                takeExtension,
+                                                                (</>))
+import           Text.PrettyPrint                              (render)
+import           Text.Regex.TDFA
+
+data Log
+  = LogFoundResponsibleCabalFile FilePath
+  | LogCalledCabalAddCodeAction
```
fendor, line 88:
```suggestion
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,310 @@
...
+  -> [(T.Text, T.Text)] -- ^ A dependency-version suggestion pairs
+  -> FilePath -- ^ Path to the haskell file (source of diagnostics)
+  -> FilePath -- ^ Path to the cabal file (that will be edited)
+  -> GenericPackageDescription
+  -> IO [CodeAction]
+addDependencySuggestCodeAction recorder plId verTxtDocId suggestions haskellFilePath cabalFilePath gpd = do
+    buildTargets <- liftIO $ getBuildTargets gpd cabalFilePath haskellFilePath
+    logWith recorder Info LogCalledCabalAddCodeAction
+    case buildTargets of
+      [] -> pure $ mkCodeAction cabalFilePath Nothing <$> suggestions
```
fendor, line 148: What would `[]` mean? Could you perhaps add a comment?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,310 @@
...
+      targets -> pure $ concat [mkCodeAction cabalFilePath (Just $ buildTargetToStringRepr target) <$>
+                                                            suggestions | target <- targets]
+  where
+    -- | Note the use of `pretty` funciton.
```
fendor, line 152:
```suggestion
    -- | Note the use of `pretty` function.
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -0,0 +1,310 @@
...
+  (cnfFields, origPackDescr, cnfComponent, cnfDependencies) <- case inputs of
+    Left err   -> throwE $ PluginInternalError $ T.pack err
+    Right pair -> pure pair
+
+  case executeConfig (validateChanges origPackDescr) (Config {..}) of
+    Nothing -> throwE $ PluginInternalError $ T.pack $ "Cannot extend build-depends in " ++ cabalFilePath
+    Just newContents  -> do
+              let edit = diffText caps (verTxtDocId, T.decodeUtf8 cnfOrigContents) (T.decodeUtf8 newContents) SkipDeletions
+              logWith recorder Debug $ LogCreatedEdit edit
+              pure edit
```
fendor, line 279:
```suggestion
      let edit = diffText caps (verTxtDocId, T.decodeUtf8 cnfOrigContents) (T.decodeUtf8 newContents) SkipDeletions
      logWith recorder Debug $ LogCreatedEdit edit
      pure edit
```

Uniform indentation

fendor (approved):
LGTM, a couple nitpicks left :innocent:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
VenInf: Yes, it was not the desired behavior.

VeryMilkyJoe (approved):

Bodigrim:
I built HLS from sources, but apparently something is off in my configuration, because when clicking on "Add dependency ..." I keep getting log messages like
```
haskell-ide-engine: 2024-08-21T22:41:41.162085Z | Warning | No plugin handles this "codeAction/resolve" request.
:: [23:41:41.162] <~~ haskell-ide-engine (124) (duration: 0ms): {'code': -32601, 'message': 'No plugins are available to handle this SMethod_CodeActionResolve request.\n Plugins installed for this method, but not available to handle this request are:\nexplicit-fields does not handle resolve requests for (unable to determine resolve owner)).\nhlint does not handle resolve requests for (unable to determine resolve owner)).\noverloaded-record-dot does not handle resolve requests for (unable to determine resolve owner)).\nimportLens does not handle resolve requests for (unable to determine resolve owner)).'}
```
Do I need to enable something specific in my HLS config? I put there 
```
			"settings": {
				"haskell.plugin.ghcide-completions.config.snippetsOn": false,
				"haskell.formattingProvider": "fourmolu",
				"haskell.plugin.stan.globalOn": false,
				"haskell.plugin.semanticTokens.globalOn": true,
				"haskell.plugin.cabalHaskellIntegration.globalOn": true,
				"haskell.plugin.cabal.globalOn": true
			},
```
but seemingly I didn't get plugin name right.

fendor:
The config shouldn't need any modification as it is disabled by default.
You may have to tell your language server to also send notifications for `.cabal` files, but I don't know how to do that in non-vscode settings.

I also get errors such:
```
2024-08-22T08:20:24.550779Z | Warning | No plugin handles this "codeAction/resolve" request.
[Trace - 10:20:24] Received response 'codeAction/resolve - (71)' in 1ms. Request failed: No plugins are available to handle this SMethod_CodeActionResolve request.
explicit-fields does not handle resolve requests for (unable to determine resolve owner)).
hlint 's feature that handles this request is disabled in your config.
overloaded-record-dot does not handle resolve requests for (unable to determine resolve owner)).
importLens does not handle resolve requests for (unable to determine resolve owner)). (-32601).
[Error - 10:20:24] Request codeAction/resolve failed.
  Message: No plugins are available to handle this SMethod_CodeActionResolve request.
explicit-fields does not handle resolve requests for (unable to determine resolve owner)).
hlint 's feature that handles this request is disabled in your config.
overloaded-record-dot does not handle resolve requests for (unable to determine resolve owner)).
importLens does not handle resolve requests for (unable to determine resolve owner)).
  Code: -32601 
  ```
But it works otherwise fine on my machine after building from source.

Could you post the full logs?

Issue #4277, opened 2024-06-04
Consider using a typed path library


michaelpj:
## Is your enhancement request related to a problem? Please describe.

Making paths absolute can be a bit tricky for us, since it's somewhat unclear where HLS is _supposed_ to be running, and hence what relative paths are supposed to be relative _to_.

But at the moment we don't even really know the scope of the problem. Hopefully, most of the time we are dealing with absolute paths, but we have various random places where we make things absolute. It would make things a lot clearer if this was tracked in the types.

## Describe the solution you'd like

Use one of the typed path libraries for file paths. We might also need to adjust our `NormalizedFilePath` and `NormalizedUri` types to track whether they are relative or absolute, but ideally we would use the same mechanics there as whatever typed path library we use.

## Describe alternatives you've considered

Roll our own typed path library. This might be worth it if it's not much code.

michaelpj:
We could also track other information in the types, e.g. "is it a Haskell or Cabal file" or "is it a project or non-project file".

PR #4364, opened 2024-07-21
[WIP] parallelising ghcide tests

komikat:
Using `getCurrentDirectory` and `setCurrentDirectory` makes it difficult to run tests in parallel since one thread modifying the global `CWD` variable would affect all other test threads. The goal of this PR is to deliberate on ways to avoid using the aforementioned functions (by making paths absolute) and have type guarantees for these absolute paths.
<details><summary>Note [Root Directory] from <i>Development.IDE.Core.Shake</i> </summary>
<p>

We keep track of the root directory explicitly, which is the directory of the project root.
We might be setting it via these options with decreasing priority:

1. from LSP workspace root, `resRootPath` in `LanguageContextEnv`.
2. command line (--cwd)
3. default to the current directory.

Using `getCurrentDirectory` makes it more difficult to run the tests, as we spawn one thread of HLS per test case.
If we modify the global Variable CWD, via `setCurrentDirectory`, all other test threads are suddenly affected,
forcing us to run all integration tests sequentially.

Also, there might be a race condition if we depend on the current directory, as some plugin might change it.
e.g. stylish's `loadConfig`. https://github.com/haskell/haskell-language-server/issues/4234

But according to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
The root dir is deprecated, that means we should cleanup dependency on the project root(Or $CWD) thing gradually,
so multi-workspaces can actually be supported when we use absolute path everywhere(might also need some high level design).
That might not be possible unless we have everything adapted to it, like 'hlint' and 'evaluation of template haskell'.
But we should still be working towards the goal.

We can drop it in the future once:
1. We can get rid all the usages of root directory in the codebase.
2. LSP version we support actually removes the root directory from the protocol.

</p>
</details> 

Thank you @fendor for mentioning https://github.com/haskell/cabal/pull/9718 which seems to be doing something similar albeit for completely different reasons.

- [ ] #4277
      ref #4372 
- [x] remove references to `setCurrentDirectory` where possible
   - [x] hls-stylish-plugin
- [ ] investigate using lspEnv's resRootPath as the source of truth for the root directory.
    <details><summary>Details</summary><p>`doInitialize` uses the `defaultRoot` parameter, which is passed down from `setupLSP` (`argsProjectRoot`) in `Arguments{..}` in `defaultArguments` which seems to be taken from `getCurrentDirectory`.</p></details>

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -129,6 +129,7 @@ library
   exposed-modules:
     Control.Concurrent.Strict
     Development.IDE
+    Development.IDE.Core.AbstractPath
     Development.IDE.Core.Actions
     Development.IDE.Core.Compile
     Development.IDE.Core.Debouncer
--- a/ghcide/src/Development/IDE/Core/AbstractPath.hs
+++ b/ghcide/src/Development/IDE/Core/AbstractPath.hs
@@ -1,0 +1,11 @@
+module Development.IDE.Core.AbstractPath where
+
+import           System.FilePath
+
+data AbstractPath = RelativePath FilePath
+                    | AbsolutePath FilePath
+                    deriving (Show)
+
+mkAbstract :: FilePath -> AbstractPath
+mkAbstract x | isRelative x = RelativePath x
+             | otherwise = AbsolutePath x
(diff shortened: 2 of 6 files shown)


komikat:
## Tracking `setCurrentDirectory` usage
- hls-test-utils/src/Test/
`runSessionWithTestConfig` uses `keepCurrentDirectory` if `testShiftRoot`

PR #4368, opened 2024-08-01
Add sig inlay hints for where clause

jetjinser:
ref #2966.

Image_1722433614391


ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,140 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+  -- The location of signature name `f`(first line) is equal to
+  -- the definition of `f`(second line).
+  }
+
+-- | All where clauses from type checked source.
+findWhereQ :: GenericQ [HsLocalBinds GhcTc]
+findWhereQ = everything (<>) $ mkQ [] (pure . findWhere)
+  where
+    findWhere :: GRHSs GhcTc (LHsExpr GhcTc) -> HsLocalBinds GhcTc
+    findWhere = grhssLocalBinds
```
fendor, line 417: This looks very expensive, don't we traverse the full AST each time the client requests InlayHints in this case? I think this needs to be tucked away in a rule, so that we at least don't recompute when nothing changes.

Do you happen to know when the client is supposed to ask for inlay hints?

fendor (changes requested):
This looks like it is in a rather good shape, I have only a couple of nitpicks and the suggestion to move the `syb` traversal code into a `defineRule`  to avoid expensive recomputations when nothing changes.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -89,28 +98,34 @@ instance Pretty Log where
...
+  '[ 'PropertyKey "whereInlayHintOn" 'TBoolean,
+     'PropertyKey "mode" ('TEnum Mode)]
 properties = emptyProperties
   & defineEnumProperty #mode "Control how type lenses are shown"
     [ (Always, "Always displays type lenses of global bindings")
     , (Exported, "Only display type lenses of exported global bindings")
     , (Diagnostics, "Follows error messages produced by GHC about missing signatures")
     ] Always
+  & defineBooleanProperty #whereInlayHintOn
+    "Enable type lens on instance methods"
```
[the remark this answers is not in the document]
jetjinser: I think this must be wrong. I'll fix it.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
jetjinser: Yes, this should be tucked away in the rule.

I think editor will request InlayHints every time the buffer text changes.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -305,26 +323,40 @@ rules recorder = do
     result <- liftIO $ gblBindingType (hscEnv <$> hsc) (tmrTypechecked <$> tmr)
     pure ([], result)
 
+bindToSig :: Id -> HscEnv -> GlobalRdrEnv -> IOEnv (Env TcGblEnv TcLclEnv) (Name, String)
+bindToSig id hsc rdrEnv = do
+    env <-
+#if MIN_VERSION_ghc(9,7,0)
+      liftZonkM
+#endif
+      tcInitTidyEnv
```
fendor, line 332: https://github.com/haskell/haskell-language-server/pull/2966#discussion_r900008476

> is this expensive? It seems like we might do it a lot...

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,140 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+          pure $ generateWhereInlayHints range (maybe ("", "") (bimap (T.pack . printName) T.pack) sig) offset
+
+    inlayHints <- sequence
+      [ bindingToInlayHints bindingId bindingRange offset
+      | WhereBindings{..} <- localBindings
+      , let sigSpans = getSrcSpan <$> existedSigNames
+      , WhereBinding{..} <- bindings
+      , let bindingSpan = getSrcSpan (idName bindingId)
+      , bindingSpan `notElem` sigSpans
+      -- , Just bindingRange <- maybeToList $ toCurrentRange pm <$> srcSpanToRange bindingLoc
```
fendor, line 483: Either remove or fix!

fendor (approved):
The code looks good to me, Ill try to give it a spin tomorrow as well. 
I do need the latest version of `vscode-haskell` to try this out, right?

jetjinser:
> The code looks good to me, Ill try to give it a spin tomorrow as well. I do need the latest version of `vscode-haskell` to try this out, right?

Yes! The latest `vscode-haskell` can enable inlay-hints.

fendor:
Thanks, I was able to try it out and it works as advertised, very handy!

Personally, I kind of dislike the inlay hints and would prefer them as lenses as inlay hints may make the code even less aligned for definitions with multiple function heads. See this example where at least I was briefly confused. 

image

This is obviously unrelated to your PR and something we need to hash out in the future, I am mainly wondering what the default behaviour should be.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
fendor: Reading the GHC comment, 

> We initialise the "tidy-env", used for tidying types before printing,
> by building a reverse map from the in-scope type variables to the
> OccName that the programmer originally used for them

It *sounds* at least expensive. However, it is now cached by a rule, so should be fine.

jetjinser:
> inlay hints may make the code even less aligned for definitions with multiple function heads.

Yes, that is indeed a problem.
A better look I can think of is to make the font size of the inlay hints smaller in the editor config, which will better highlight the original code. But this still doesn't solve the problem of it breaking the alignment.

BTW, the main problem with code lens is that its position is not predictable and is determined by the editor.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,165 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+  define (cmapWithPrio LogShake recorder) $ \GetWhereBindingTypeSigs nfp -> do
+    tmr <- use TypeCheck nfp
+    -- we need session here for tidying types
+    hsc <- use GhcSession nfp
+    result <- liftIO $ whereBindingType (tmrTypechecked <$> tmr) (hscEnv <$> hsc)
+    pure ([], result)
+
+whereBindingType :: Maybe TcGblEnv -> Maybe HscEnv -> IO (Maybe WhereBindingTypeSigsResult)
+whereBindingType (Just gblEnv) (Just hsc) = do
+  let wheres = findWhereQ (tcg_binds gblEnv)
```
michaelpj, line 438: I wonder if we should just go ahead and do lets at the same time? I would imagine it's very similar code?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,165 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+                           (col (getLoc fun_id) - col (getLoc bind))
+        in Just $ pure whereBinding
+      PatBind{..} -> Just $ (everything (<>) $ mkQ [] (fmap (uncurry wb) . maybeToList . findIdFromPat)) pat_lhs
+        where
+          wb id srcSpan = WhereBinding id srcSpan (col srcSpan - col (getLoc pat_lhs))
+      _           -> Nothing
+      where
+        col = srcSpanStartCol . realSrcSpan
+
+    -- | Example: Find `a` and `b` from @(a,b) = (1,True)@
```
michaelpj, line 479: This doesn't seem quite right, it's only finding one thing, not multiple things?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,165 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+
+-- | Provide code lens for where bindings.
+whereClauseInlayHints :: PluginMethodHandler IdeState Method_TextDocumentInlayHint
+whereClauseInlayHints state plId (InlayHintParams _ (TextDocumentIdentifier uri) visibleRange)  = do
+  enabled <- liftIO $ runAction "inlayHint.config" state $ usePropertyAction #whereInlayHintOn plId properties
+  if not enabled then pure $ InL [] else do
+    nfp <- getNormalizedFilePathE uri
+    (WhereBindingTypeSigsResult (localBindings, sigMap), pm)
+        <- runActionE "InlayHint.GetWhereBindingTypeSigs" state $ useWithStaleE GetWhereBindingTypeSigs nfp
+    let bindingToInlayHints id sig = generateWhereInlayHints (T.pack $ printName (idName id)) (maybe "_" T.pack sig)
```
michaelpj, line 496: I think we should probably just omit things where we don't have a signature, rather than putting in `_`?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,165 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+        -- like @(a, b) = (42, True)@, there are `a` and `b`
+        -- in one binding.
+        inlayHints =
+          [ bindingToInlayHints bindingId bindingSig bindingRange offset
+          | WhereBindings{..} <- localBindings
+          , let sigSpans = getSrcSpan <$> existingSigNames
+          , WhereBinding{..} <- bindings
+          , let bindingSpan = getSrcSpan (idName bindingId)
+          , let bindingSig = Map.lookup bindingId sigMap
+          , bindingSpan `notElem` sigSpans
```
michaelpj, line 508: This seems a little complicated? I wonder if we could just track the `Id`s that have signatures already?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,165 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+          , Just bindingRange <- maybeToList $ toCurrentRange pm <$> srcSpanToRange bindingLoc
+          -- , Just bindingRange <- [srcSpanToRange bindingLoc]
```
michaelpj, line 510: commented out

ghcide/test/exe/InlayHintTests.hs
```diff
@@ -0,0 +1,179 @@
...
+                     , _label = InL ":: Map.Map Bool Char"
+                     , _textEdits = Just [mkTextEdit 7 9 "g :: Map.Map Bool Char\n         "]
+                     }
+      ]
+    ]
+  ]
+
+editTest :: String -> TestTree
+editTest file =
+  testWithDummyPlugin (file <> " (InlayHint EditText)") (mkIdeTestFs [copyDir "local-sig-lens"]) $ do
```
michaelpj, line 120: I guess the directory should be "local-sig-hint", perhaps?

ghcide/test/exe/InlayHintTests.hs
```diff
@@ -0,0 +1,179 @@
...
+    , expectFail $ editTest "ScopedTypeVariables"
+    , editTest "Nest"
+    , editTest "NoLens"
+    , expectFail $ editTest "Typeclass"
+    , editTest "Qualified"
+    ]
+  , testGroup "apply EditText"
+    [ hintTest "Simple" $ (@=?)
+      [defInlayHint { _position = Position 5 9
+                    , _label = InL ":: Bool"
```
michaelpj, line 66: I wonder if we should display the hints on the previous line rather than right after the binding. i.e. make the hint look like the edit currently does.

Advantages:
1. It makes the effect of applying the hint align with how the hint looks, which is I think what we're supposed to do.
2. It makes the situation less bad if the type is very long. That might already be annoying, but it could be even more annoying if it's right in the middle of something you're trying to read. Maybe we should look at what other language servers do?

michaelpj (approved):
This is looking pretty good, I think!

I wonder if we should roll in some other cases. I think the top-level code lenses should also be inlay hints (if that's supported), and we could maybe do let-bindings too, since I'm sure that's very similar to where-bindings?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
jetjinser: It seems that `Sig GhcTc` does not exists (not sure), that is to say, `Id` (`IdP GhcTc`) does not exist in `HsLocalBinds`, and currently only `Name` (`IdP GhcRn`) can be obtained from `Sig GhcRn`.
ref: https://hackage.haskell.org/package/ghc-9.10.1/docs/GHC-Hs-Binds.html#v:NValBinds

On the other hand, the pass of `LHsBinds` is `GhcTc`, and the `Name` obtained from `bindingId` is different from the`Name` in `Sig` because their `Unique` is different.
ref: https://hackage.haskell.org/package/ghc-9.10.1/docs/GHC-Types-Unique.html#v:nonDetCmpUnique

If the `Id`s that have signatures is not easily available, I've tried comparing via `OccName`, but maybe `SrcSpan` is still better?

jetjinser:
image

should this intend?
@michaelpj

ghcide/test/exe/InlayHintTests.hs
jetjinser: I don't think inlay hints can do this.
I found this issue: https://github.com/microsoft/language-server-protocol/issues/1821.

AFAIK, standard inlay hints cannot be in virtual line.
It's more like a fixed position code lens?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -340,3 +372,188 @@ pprPatSynTypeWithoutForalls p = pprPatSynType pWithoutTypeVariables
...
+#else
+    findLetExpr (HsLet _ binds _)           = [binds]
+#endif
+    findLetExpr (HsDo _ _ (unLoc -> stmts)) = concatMap (findLetStmt . unLoc) stmts
+    findLetExpr _                           = []
+
+    findLetStmt :: ExprStmt GhcTc -> [HsLocalBinds GhcTc]
+    findLetStmt (LetStmt _ binds) = [binds]
+    -- TODO(jinser): why `foo <- expr` does not exist
+    -- findLetStmt (BindStmt _ _ expr) = findLetExpr (unLoc expr)
```
jetjinser, line 470: I'm wondering why this doesn't work, `expr` in `BindStmt` contains Var but not Bind.

cassandracomar:
this needs an update for ghc-9.12 -- can send a PR for that but not sure where to send it. I can either open a PR on this repo with this PR included or I can send a PR on the originating repo.

also, after using this PR for a bit, I must say that including the inlay hints on the same line makes these hints tremendously less useful than they would be if placed on the line prior. I understand there were issues with the older approach but perhaps we could expose a setting to use the older behavior, rendering them as lenses? long inferred type signatures completely mangle local bindings, placing function arguments many lines away from the name of the local function. and that's if you have word-wrapping enabled. without it, the local variable binding zips off-screen with no warning and it appears that some unbound/non-local definition is in use (editor dependent).

e.g.:
screenshot-multiline-signature

fendor:
> I can either open a PR on this repo with this PR included or I can send a PR on the originating repo.

Just open a new PR, this one has been dormant for some time.

> this needs an update for ghc-9.12 -- can send a PR for that

🎊 

> I must say that including the inlay hints on the same line makes these hints tremendously less useful than they would be if placed on the line prior

We probably simply want an option that allows us to toggle this behaviour. I think as well, by default, we should have code lenses... But it is quite easy to switch between inlay hints and codelenses in the plugin.

cassandracomar:
@fendor ok will do. just need to apply CPP guards so it continues to work with older ghcs first.

> We probably simply want an option that allows us to toggle this behaviour.

yeah, this is what I was suggesting as well. let me see if I can figure out how the lenses were created before and how they were disabled so we can reexpose them in a new option.

jian-lin:
@fendor 

> I think as well, by default, we should have code lenses

Why?  Is it because inlay hints were placed in the same line and code lenses were places in the previous line?

Like code lenses, we can place inlay hints in the previous line, too.  IIRC, my LSP client (Emacs Eglot) shows those hints without problem.  Not sure about other LSP clients and editors though.

michaelpj:
If anything, we should prefer inlay hints for anything that _requires_ a particular position, because inlay hints actually specify a position whereas code lenses don't. This causes trouble: e.g. the emacs lsp-mode has an option to show code lenses at the _end_ of the line, which makes our type signature lenses look like nonsense.

I think we can have both, but ideally we don't have the code lenses as a fallback for people who don't have a inlay-hint-capable client, or possibly as an option.

cassandracomar:
@jian-lin 
> Like code lenses, we can place inlay hints in the previous line, too. 

apparently we're not supposed to do that per the spec. see [this previous comment](https://github.com/haskell/haskell-language-server/pull/4368#discussion_r1740472510) about it.

jian-lin:
@cassandracomar 

Putting an inlay hint in the previous line works.  Just set the position of the hint to a position in the previous line.  I did an experiment about that a few months ago.  Not sure why "multi-line hints" or "virtual lines" are relevant here.  

> apparently we're not supposed to do that per the spec

It is not apparent to me.  Could you provide links to the specific places in the spec?  (offtopic: I find many places in the LSP spec not clear to me.)

---

edit:

Oh, if the previous line is not empty, inlay hints may not be nicely rendered.

cassandracomar:
> It is not apparent to me. Could you provide links to the specific places in the spec? (offtopic: I find many places in the LSP spec not clear to me.)

I'm not positive. the comment I linked had a discussion about it. there's an open issue on the language-server-protocol repo where the notion of multi-line inlay hints are regarded [as an abuse of the feature](https://github.com/microsoft/language-server-protocol/issues/1821#issuecomment-1739602648). as you've noticed, we can't just render the hint on the previous line as it may not be empty. we instead have to render the hint in-place, insert a new line, then restore the original line.

michaelpj:
> there's an open issue on the language-server-protocol repo where the notion of multi-line inlay hints are regarded https://github.com/microsoft/language-server-protocol/issues/1821#issuecomment-1739602648.

The person saying that is just a random commenter!

Jinser's observation that it just plain doesn't work is more troubling. I had thought it would be possible because of the discussion in https://github.com/microsoft/language-server-protocol/issues/1558, however I see that that was inconclusive.

I'm unsure what we should do if inlay hints on preceding lines don't work and code lenses don't have reliable positions :/

PR #4370, opened 2024-08-01
Speedup waitfordiagnosticsfrom

komikat:
ref #4339
- [ ] fix `waitForDIagnosticsFrom` which waits recursively 
   - [ ] hls-class-plugin
   - [ ] hls-stan-plugin
   - [ ] hls-rename-plugin
   - [ ] hls-pragma-plugin


komikat:
accidentally branched from local, ref #4371

PR #4371, opened 2024-08-01
Speedup waitForDiagnosticsFrom

komikat:
ref #4339
- [ ] fix `waitForDiagnosticsFrom` which waits recursively 
   - [ ] hls-stan-plugin
   - [ ] hls-rename-plugin
   - [ ] hls-pragma-plugin


komikat:
Each test in `hls-stan-plugin-tests` takes around 5s, investigating if this can be refactored.

PR #4372, opened 2024-08-01
Path newtype declaration

komikat:
- Phantom type with one parameter (Abs/Rel)
- ghcide refactor in progress


komikat:
`simple plugin` failing assertion.
mkAbsFromFp being called for `cabal.project` somewhere.

PR #4376, opened 2024-08-07
Disable ghcide-bench by default

### LINKED ISSUE #4359: Is ghcide-bench supposed to be built and installed by default?
## Your environment 

Which OS do you use? Fedora Linux

Which version of GHC do you use and how did you install it? ghc-9.4.5 from Fedora
https://github.com/fedora-haskell/haskell-language-server

Which version of HLS do you use and how did you install it? 2.9.0

## What's wrong?

bindir/ghcide-bench gets installed.
Is it intended for end users?

### Debug information

https://download.copr.fedorainfracloud.org/results/petersen/haskell-language-server/fedora-rawhide-aarch64/07729207-haskell-language-server/builder-live.log.gz

fendor:
Introduce a cabal flag to control the building of `ghcide-bench`, which is an internal tool for benchmarking HLS.

This fixes a regression, where running `cabal install haskell-language-server` to install HLS from Hackage would install `ghcide-bench` as well as the required HLS executables.

Fixes #4359 

POC, so we can iterate on this.

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -93,6 +93,11 @@ flag dynamic
   default:     True
   manual:      True
 
+flag ghcide-bench
+  description: Enable building of the ghcide-bench executable
+  default:     False
+  manual:      True
+
 ----------------------------
 ----------------------------
 -- PLUGINS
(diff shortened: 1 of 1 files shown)


soulomoon:
Do we need to update the github CI as well ?

fendor:
Yeah probably, thanks for the reminder!

PR #4380, opened 2024-08-14
Cabal go to module's definition

VenInf:
### Go to the module's definition

If you click go-to definition on the field under `exposed-module` or `other-module`
it will open the file where this module was defined.

Video with an example:

https://github.com/user-attachments/assets/c5cdcb9e-4d24-4d6b-a83a-de34d8dcedf6

### Implementation details

The go-to definition function compares the highlighted text with modules in the cabal file.
If there is a match, it takes the respective build target and tries to fetch their `hsSourceDirs` from the `PackageDescription`. (by looking at all `buildInfos` with matching names).

After finding them, it constructs a path using directory where the cabal file is located, the info from `hsSourceDirs` and a name of the module converted to a path.
If the file exists it returns the `Definition` with the acquired location.

### What isn't implemented
 - [x] No tests
 - [x] Current solution can't differentiate modules with same names.
 - [x] Current solution to choose if a definition can be provided isn't great, and is just a chain of `if`'s
 - [ ] The range in location is just zero, instead of highlighting the name of the module.


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +298,105 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+                        Nothing       -> pure $ InR $ InR Null
+                        Just location -> pure $ InL $ Definition $ InL location
+    where
+      cursor = Types.lspPositionToCabalPosition (msgParam ^. JL.position)
+      uri = msgParam ^. JL.textDocument . JL.uri
+      isSectionArgName name (Syntax.Section _ sectionArgName _) = name == CabalFields.onelineSectionArgs sectionArgName
+      isSectionArgName _ _ = False
+      isModuleName name (_,  moduleName) = name == moduleName
+
+      lookupBuildTargetPackageDescription :: PackageDescription -> Maybe T.Text -> [BuildInfo]
```
fendor, line 355: Docs please :)

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +298,105 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+              then Just testBuildInfo
+              else Nothing
+          benchmarkNameLookup :: Benchmark -> T.Text -> Maybe BuildInfo
+          benchmarkNameLookup (Benchmark {benchmarkName, benchmarkBuildInfo}) buildTargetName =
+              if T.pack (unUnqualComponentName benchmarkName) == buildTargetName
+              then Just benchmarkBuildInfo
+              else Nothing
+
+      toHaskellFile :: T.Text -> FilePath
+      toHaskellFile moduleName = foldl1 (</>) (map T.unpack $ T.splitOn "." moduleName) ++ ".hs"
```
fendor, line 398: Instead of letting `foldl1` fail, handle this potential error and provide a proper error message for the user!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/CabalFields.hs
```diff
@@ -66,19 +146,149 @@ getOptionalSectionName (x:xs) = case x of
     Syntax.SecArgName _ name -> Just (T.decodeUtf8 name)
     _                        -> getOptionalSectionName xs
 
+type BuildTargetName = T.Text
+type ModuleName      = T.Text
+
+-- | Given a cabal AST returns pairs of all respective target names
+-- and the module name bounded to them. If a target is a main library gives
```
fendor, line 153:
```suggestion
-- and the module name bound to them. If a target is a main library gives
```

fendor (changes requested):
LGTM, couple of nitpicks, then this should be good to go!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -277,6 +298,105 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+                      let locations = map (\pth -> Location (filePathToUri pth) (mkRange 0 0 0 0)) allPaths
+                      case safeHead locations of -- We assume there could be only one source location
+                        Nothing       -> pure $ InR $ InR Null
+                        Just location -> pure $ InL $ Definition $ InL location
+    where
+      cursor = Types.lspPositionToCabalPosition (msgParam ^. JL.position)
+      uri = msgParam ^. JL.textDocument . JL.uri
+      isSectionArgName name (Syntax.Section _ sectionArgName _) = name == CabalFields.onelineSectionArgs sectionArgName
+      isSectionArgName _ _ = False
+      isModuleName name (_,  moduleName) = name == moduleName
```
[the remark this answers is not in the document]
VenInf: I thought that they are very application-specific, and almost don't do what they promise.
But maybe I will be able to rewrite them to something better, I will look into it.

VenInf:
Noteworthy change:
Split one large `gotoDefinition` function to multiple specific ones.
This will hurt the performance a little bit, but hopefully make the addition of other definitions easier.

VeryMilkyJoe (changes requested):
Thank you, this looks great! I just have some small comments.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Definition.hs
```diff
@@ -0,0 +1,205 @@
...
+  -> Syntax.Position -- ^ Cursor position
+  -> [Syntax.Field Syntax.Position] -- ^ Trimmed cabal AST on a cursor
+  -> Maybe Definition
+gotoCommonSectionDefinition uri commonSections cursor fieldsOfInterest =
+  case CabalFields.findTextWord cursor fieldsOfInterest of
+    Nothing -> Nothing
+    Just cursorText -> do
+      case find (isSectionArgName cursorText) commonSections of
+        Just commonSection -> Just $ Definition $ InL $ Location uri $ CabalFields.getFieldLSPRange commonSection
+        Nothing -> Nothing
```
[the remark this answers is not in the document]
VenInf: Oh, that's neat, forgot that it's possible.

fendor (approved):
LGTM, thanks!

PR #4385, opened 2024-08-19
Documentation for build-depends on hover

VenInf:
### Documentation for build-depends on hover.

If you hover over the field under `build-depends`
it will give Documentation with a hackage link to a package.

Video with an example:

https://github.com/user-attachments/assets/2b20354c-3fc9-4bbf-89ec-e0a61e07b47c

### Implementation details

This is a naive solution for this problem, hovered text is filtered using
a regex and compared to found dependencies from `PackageDescription`.
If it was found, then the `Documentation` message is added, otherwise left as is.

This approach assumes, that if a package is mentioned in a cabal file, there should be 
a link with the same name. In this implementation there are no spelling checks.


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -309,6 +317,50 @@ gotoDefinition ideState _ msgParam = do
       isSectionArgName name (Syntax.Section _ sectionArgName _) = name == CabalFields.onelineSectionArgs sectionArgName
       isSectionArgName _ _ = False
 
+-- | CodeActions for hover messages.
+--
+-- Provides a CodeAction for displaying message on hover.
+-- If found that the filtered hover message is a dependency,
+-- adds a Documentation link.
```
fendor, line 324:
```suggestion
-- | Handler for hover messages.
--
-- Provides a Handler for displaying message on hover.
-- If found that the filtered hover message is a dependency,
-- adds a Documentation link.
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -309,6 +317,50 @@ gotoDefinition ideState _ msgParam = do
...
+        Just cursorText -> do
+          (gpd, _) <- runActionE "cabal.GPD" ide $ useWithStaleE ParseCabalFile nfp
+          let depsNames = map dependencyName $ allBuildDepends $ flattenPackageDescription gpd
+              mText = filterVersion cursorText
+          case mText of
+            Nothing -> pure $ foundHover (Nothing, [cursorText])
+            Just txt ->
+              if txt `elem` depsNames
+                then pure $ foundHover (Nothing, [txt <> "\n", documentationText txt])
+                else pure $ foundHover (Nothing, [txt])
```
fendor, line 342: Incosistent indentation, could you indent everything with 2 spaces instead of these 4?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -309,6 +317,50 @@ gotoDefinition ideState _ msgParam = do
...
+      uri = msgParam ^. JL.textDocument . JL.uri
+
+      dependencyName :: Dependency -> T.Text
+      dependencyName dep = T.pack $ unPackageName $ depPkgName dep
+
+      filterVersion :: T.Text -> Maybe T.Text
+      filterVersion msg = getMatch (msg =~ regex)
+        where
+          regex :: T.Text
+          regex = "([a-zA-Z0-9-]*[a-zA-Z0-9]).*"
```
fendor, line 354:
```suggestion
          regex = "([a-zA-Z0-9-]+)"
```

I don't understand what the `.*` is for? You only care about the match group, right?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -309,6 +317,50 @@ gotoDefinition ideState _ msgParam = do
...
+hover ide _ msgParam = do
+      nfp <- getNormalizedFilePathE uri
+      (cabalFields, _) <- runActionE "cabal.cabal-hover" ide $ useWithStaleE ParseCabalFields nfp
+      let mCursorText = CabalFields.findTextWord cursor cabalFields
+      case mCursorText of
+        Nothing ->
+          pure $ InR Null
+        Just cursorText -> do
+          (gpd, _) <- runActionE "cabal.GPD" ide $ useWithStaleE ParseCabalFile nfp
+          let depsNames = map dependencyName $ allBuildDepends $ flattenPackageDescription gpd
```
fendor, line 335: I am not sure whether `flattenPackageDescription` is expensive, perhaps we want to cache this as well... However, fine for now, imo.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor, line 342: Why are we returning anything if this is not a name we know? This should return an empty result or something like that

fendor (changes requested):
Couple of nitpicks, otherwise good to go!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
VenInf: > You only care about the match group, right?
Yes, I missed that, thank you.

Also can't replace `([a-zA-Z0-9-]*[a-zA-Z0-9]` with `[a-zA-Z0-9-]+`,
because the first one checks, it doesn't end with a `-`.

fendor: Ah thanks! Perhaps add a comment then, because this is too easy to overlook!

Bodigrim: I'd suggest to make it `([-a-zA-Z0-9]*[a-zA-Z0-9]).*` then so that the difference stands out.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -291,23 +301,67 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+--
+-- Provides a Handler for displaying message on hover.
+-- If found that the filtered hover message is a dependency,
+-- adds a Documentation link.
+hover :: PluginMethodHandler IdeState LSP.Method_TextDocumentHover
+hover ide _ msgParam = do
+  nfp <- getNormalizedFilePathE uri
+  (cabalFields, _) <- runActionE "cabal.cabal-hover" ide $ useWithStaleE ParseCabalFields nfp
+  let mCursorText = CabalFields.findTextWord cursor cabalFields
+  case mCursorText of
```
fendor, line 332:
```suggestion
  case CabalFields.findTextWord cursor cabalFields of
```

Unless you think this uglier

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -291,23 +301,67 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+
+-- | Handler for hover messages.
+--
+-- Provides a Handler for displaying message on hover.
+-- If found that the filtered hover message is a dependency,
+-- adds a Documentation link.
+hover :: PluginMethodHandler IdeState LSP.Method_TextDocumentHover
+hover ide _ msgParam = do
+  nfp <- getNormalizedFilePathE uri
+  (cabalFields, _) <- runActionE "cabal.cabal-hover" ide $ useWithStaleE ParseCabalFields nfp
```
fendor, line 330: I think `useWithStaleE` doesn't quite work with `cabal` files, and we had some bad UX with it in the past.
One reason it doesn't work that well is that we do not do any Position Mapping iirc, so outdated cabal files will have wrong positions... However, hover is usually used when the language server has settled, so it might not be noticeable at all.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -291,23 +301,67 @@ fieldSuggestCodeAction recorder ide _ (CodeActionParams _ _ (TextDocumentIdentif
...
+  let mCursorText = CabalFields.findTextWord cursor cabalFields
+  case mCursorText of
+    Nothing ->
+      pure $ InR Null
+    Just cursorText -> do
+      (gpd, _) <- runActionE "cabal.GPD" ide $ useWithStaleE ParseCabalFile nfp
+      let depsNames = map dependencyName $ allBuildDepends $ flattenPackageDescription gpd
+          mText = filterVersion cursorText
+      case mText of
```
fendor, line 339:
```suggestion
      case filterVersion cursorText of
```

Maybe?

fendor (approved):
LGTM, just needs to be rebased or merge changes from upstream!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
VenInf: Good call, the `mCursorText` didn't add much readability.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
VenInf: I think it's a good idea to change it to the `useWithStale`, so it will be used more often in the future, but I thought I used it excessively in my other solutions.

Should I change it up there as well? Or it was a good choice earlier?

VenInf: Rewrote to the `useWithStale` for hover.

fendor: `useWithStale` is the same thing but without explicit error handling. I think `useE` is the best function.

VenInf: Ah, I didn't understand that the issue was with `withStale` part and not with error handling. I will change it back then, to `useE`.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
fendor: @VenInf Would it be possible to apply this change?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -328,6 +340,47 @@ cabalAddCodeAction state plId (CodeActionParams _ _ (TextDocumentIdentifier uri)
...
+    cursor = Types.lspPositionToCabalPosition (msgParam ^. JL.position)
+    uri = msgParam ^. JL.textDocument . JL.uri
+
+    dependencyName :: Dependency -> T.Text
+    dependencyName dep = T.pack $ unPackageName $ depPkgName dep
+
+    filterVersion :: T.Text -> Maybe T.Text
+    filterVersion msg = getMatch (msg =~ regex)
+      where
+        regex :: T.Text
```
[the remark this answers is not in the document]
VenInf: Added clarifications.

PR #4393, opened 2024-08-30
Fix for multiple build targets at cabal-add

VenInf:
### Fix for multiple build targets at cabal-add

In my current merged solution, cabal-add `CodeAction` is provided only for the main target.
This was only a temporary solution due to limitations of the `getBuildTargets` function.
For more information, [consult](https://github.com/haskell/haskell-language-server/pull/4360#discussion_r1681352283) this thread.

Now it was fixed, and you can choose to what target add the dependency.
Video with an example:

https://github.com/user-attachments/assets/b75587df-e277-48de-8c3c-3fb705b64f32

Also, I changed the title for the `CodeAction`, to a hopefully better one.

### Implementation details
I changed the functions `getBuildTargets` and  `buildTargetComponentName` to  
the `getBuildTargetComponentNames` that handles `GenericPackageDescription` and gives `ComponentName`s of all targets.

### What isn't implemented
- [x] Previous tests were built assuming that there is need to wait only for one `CodeAction`. Now this leads to a race condition.

### DIFF
--- a/plugins/hls-cabal-plugin/test/CabalAdd.hs
+++ b/plugins/hls-cabal-plugin/test/CabalAdd.hs
@@ -28,7 +28,7 @@ cabalAddTests :: TestTree
 cabalAddTests =
   testGroup
     "CabalAdd Tests"
-    [ runHaskellTestCaseSession "Code Actions - Can add hidden package" ("cabal-add-testdata" </> "cabal-add-exe")
+    [ runHaskellTestCaseSession "Code Actions - Can add hidden packages" ("cabal-add-testdata" </> "cabal-add-exe")
         (generateAddDependencyTestSession "cabal-add-exe.cabal" ("src" </> "Main.hs") "split" [253])
     , runHaskellTestCaseSession "Code Actions - Can add hidden package to a library" ("cabal-add-testdata" </> "cabal-add-lib")
         (generateAddDependencyTestSession "cabal-add-lib.cabal" ("src" </> "MyLib.hs") "split" [348])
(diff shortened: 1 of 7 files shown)


plugins/hls-cabal-plugin/test/CabalAdd.hs
```diff
@@ -117,8 +120,10 @@ cabalAddTests =
         _ <- waitForDiagnosticsFrom hsdoc
         cas <- Maybe.mapMaybe (^? _R) <$> getAllCodeActions hsdoc
         let selectedCas = filter (\ca -> "Add dependency" `T.isPrefixOf` (ca ^. L.title)) cas
-        mapM_ executeCodeAction selectedCas
-        _ <- skipManyTill anyMessage $ getDocumentEdit cabDoc -- Wait for the changes in cabal file
+        let runAvait codeAction = do
+              executeCodeAction codeAction
+              skipManyTill anyMessage $ getDocumentEdit cabDoc
+        mapM_ runAvait selectedCas
```
fendor, line 126:
```suggestion
        let runAvail codeAction = do
              executeCodeAction codeAction
              skipManyTill anyMessage $ getDocumentEdit cabDoc
        mapM_ runAvail selectedCas
```

Maybe?

fendor (commented):
So, do I understand this correctly, we offer to add the dependency to *some* dependency, even though we should be able to statically tell to which component the filepath belongs to?

VenInf:
> So, do I understand this correctly, we offer to add the dependency to _some_ dependency, even though we should be able to statically tell to which component the filepath belongs to?

Yes, this implementation offers to chose, to what target add the dependency. I think that I can narrow down the list of potential targets using the information from the filepath, but it will take some time to implement. I don't know if it's possible to completely resolve to what target the dependency is related to.
Right now, in current merged solution, there are no resolvers at all.

VenInf:
After a better look at the current implementation, it turned out that target selection was better than expected.
Unfortunately, the mechanism isn't properly documented and tested, and I will solve it in a separate PR.

PR #4406, opened 2024-09-18
Rule inputs alternate hasinput class

nlander:
Instead of type level list inclusion for `HasInput`, this enumerates all of the instances and `ValidInputs` is a sum type.

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -135,6 +135,7 @@ library
     Development.IDE.Core.FileStore
     Development.IDE.Core.FileUtils
     Development.IDE.Core.IdeConfiguration
+    Development.IDE.Core.InputPath
     Development.IDE.Core.OfInterest
     Development.IDE.Core.PluginUtils
     Development.IDE.Core.PositionMapping
--- a/ghcide/src/Development/IDE/Core/FileStore.hs
+++ b/ghcide/src/Development/IDE/Core/FileStore.hs
@@ -69,6 +69,7 @@ import           Language.LSP.VFS
 import           System.FilePath
 import           System.IO.Error
 import           System.IO.Unsafe
+import Development.IDE.Core.InputPath (InputPath (InputPath, unInputPath))
 
 
 data Log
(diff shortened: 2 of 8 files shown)


michaelpj:
sketch of what I was saying on the call:
```haskell
type family RuleInput k :: Type

newtype ProjectHaskellFile = ProjectHaskellFile NormalizedFilePath

-- examples
type instance RuleInput GhcSession = ProjectHaskellFile
type instance RuleInput GetClientSettings = ()

-- various stuff might require `RuleInputFromFilePath (RuleInput r)` 
-- which basically says you have a way to convert a filepath into
-- input for the rule, possibly failing
class RuleInputFromFilePath i where
   toRuleInput :: NormalizedFilePath -> Either String i

-- example
someRule = do
   ...
   validFp <- case toRuleInput nfp of
     Right p -> pure p
     Left e -> ... some error ...
   use someRule validFp
```

nlander:
> sketch of what I was saying on the call:
> 
> ```haskell
> type family RuleInput k :: Type
> 
> newtype ProjectHaskellFile = ProjectHaskellFile NormalizedFilePath
> 
> -- examples
> type instance RuleInput GhcSession = ProjectHaskellFile
> type instance RuleInput GetClientSettings = ()
> 
> -- various stuff might require `RuleInputFromFilePath (RuleInput r)` 
> -- which basically says you have a way to convert a filepath into
> -- input for the rule, possibly failing
> class RuleInputFromFilePath i where
>    toRuleInput :: NormalizedFilePath -> Either String i
> 
> -- example
> someRule = do
>    ...
>    validFp <- case toRuleInput nfp of
>      Right p -> pure p
>      Left e -> ... some error ...
>    use someRule validFp
> ```

@michaelpj The problem with that approach is that for some rules, both project haskell files and dependency haskell files are valid inputs. This is why I distinguish between the `ValidInputs` kind and the `InputType` kind.

michaelpj:
> The problem with that approach is that for some rules, both project haskell files and dependency haskell files are valid inputs.

Sure, so then you would need to have `ProjectOrHaskellFile` as the rule input and your `RuleInputFromFilePath` would be a bit more complex. It doesn't solve the "open universe of input file types" problem, but I think it's maybe a little nicer than what you're doing right now, and _does_ solve the "this rule should take an input of type `()`" problem.

Issue #4379, opened 2024-08-10
Action to export pattern (:+) produces malformed export list


Bodigrim:
### Your environment

Which OS do you use? macOS
Which version of GHC do you use and how did you install it? 9.6.5, from ghcup

Which LSP client (editor/plugin) do you use? Sublime Text
Which version of HLS do you use and how did you install it? 2.9.0.1, from ghcup.

### Steps to reproduce

Put

```haskell
{-# LANGUAGE PatternSynonyms #-}
module Lib () where 
pattern x :+ y = (x, y)
```

and run code action "Export :+"

### Expected behaviour

`pattern (:+)` is to be added to the export list.

### Actual behaviour

`pattern :+` is added to the export list, which is a parse error.

pbrinkmeier:
I'll take a stab at this :)

fendor:
Closed by #4420

PR #4419, opened 2024-10-11
parenthise pattern export if needed

ttylec:
Solves #4379 issue. `parenthizeIfNeeds` was missing.


ttylec:
Closing since there is a better one.

PR #4420, opened 2024-10-11
Fix exporting operator pattern synonym

pbrinkmeier:
This PR addresses #4379.

- Adds a few tests for exporting pattern synonyms with operator names.
- Parenthesizes any name to be exported if it starts with an operator character.

The last commit restructures the code such that any name is parenthesized regardless of what it is exported as.


fendor (approved):
LGTM!

pbrinkmeier:
Not sure why the pre-commit check is failing :/

Issue #4400, opened 2024-09-12
Show "Add dependency" Code Action only if there is no `package.yaml` file


fendor:
If the project is a stack project, there might be a `package.yaml` file. Adding a dependency to the generated .cabal file is a bad idea, as it will lead to annoying behaviour: Stack will no longer re-generate the .cabal file, so changes to `package.yaml` are not propagated to the `.cabal` file any more.

solution:

When looking for the `.cabal` file in the filesystem, check there is no `package.yaml` file next to it.

AureliaSnd:
We are trying to work on this issue.

fendor:
Closed by #4422

PR #4422, opened 2024-10-11
Cabal plugin: implement check for package.yaml in a stack project

JMoss-dev:
Solves: #4400

Modifies `findResponsibleCabalFile` function to add `guardAgainstHpack` function which checks if a `package.yaml` file exists in the same directory as the `.cabal` file and handles the response in the cabal plugin.

Prevents stack from not more re-generating .cabal file if package.yaml is changed.

### DIFF
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
@@ -277,7 +277,7 @@ getDependencyEdit recorder env cabalFilePath buildTarget dependency = do
       logWith recorder Debug $ LogCreatedEdit edit
       pure edit
 
--- | Given a path to a haskell file, returns the closest cabal file.
+-- | Given a path to a haskell file, returns the closest cabal file, except if package.yaml is also present.
 --   If cabal file wasn't found, gives Nothing.
 findResponsibleCabalFile :: FilePath -> IO (Maybe FilePath)
 findResponsibleCabalFile haskellFilePath = do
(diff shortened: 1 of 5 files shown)


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -277,7 +277,7 @@ getDependencyEdit recorder env cabalFilePath buildTarget dependency = do
       logWith recorder Debug $ LogCreatedEdit edit
       pure edit
 
--- | Given a path to a haskell file, returns the closest cabal file.
+-- | Given a path to a haskell file, returns the closest cabal file, except if package.yaml is also present.
```
fendor, line 280: Could you please expand a little why we don't return the .cabal file if a `package.yaml` file is found right nex to it?

fendor (changes requested):
Looks great, thanks! One little nitpick about the documentation, then this is good to go :confetti_ball:

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -278,6 +278,9 @@ getDependencyEdit recorder env cabalFilePath buildTarget dependency = do
       pure edit
 
 -- | Given a path to a haskell file, returns the closest cabal file.
+--   If package.yaml is present in same directory as .cabal file, it returns nothing, because adding a dependency to a generated cabal file
+--   will break propagation of changes from package.yaml to cabal file in stack project.
+--
```
VeryMilkyJoe, line 283:
```suggestion
--   If a package.yaml is present in same directory as the .cabal file, returns nothing, because adding a dependency to a generated cabal file
--   will break propagation of changes from package.yaml to cabal files in stack projects.
```

VeryMilkyJoe (changes requested):
Thank you very much for this PR! 
I just have a small nitpick regarding the documentation :)

fendor (approved):

VeryMilkyJoe (approved):

PR #4423, opened 2024-10-12
link executables dynamically to speed up linking

develop7:
Dramatically decreases linking times and shaves 6 of 14G off HLS' `dist-newstyle` directory on my machine by producing way smaller executables.

While `cabal build`ing HLS after switching a branch I've noticed linking test executables takes minutes, plural, _each_. Then I've checked the size of a freshly-built test binary and it was 150-something megabytes, no wonder linker took a while. With `executable-dynamic` enabled, linking is lightning fast and binaries are a megabyte at most, which should probably improve the developer experience and decrease the feedback loop time.

UPD: https://github.com/haskell/haskell-language-server/issues/2659 seems related, not sure how up-to-date though

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -19,7 +19,9 @@ write-ghc-environment-files: never
 
 -- Link executables dynamically so the linker doesn't produce test
 -- executables of ~150MB each and works lightning fast at that too
-executable-dynamic: True
+-- Disabled on Windows
+if(!os(windows))
+  executable-dynamic: True
 
 -- Many of our tests only work single-threaded, and the only way to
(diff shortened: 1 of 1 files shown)


fendor:
I am not sure what ramification this change might have. @wz1000 or @michaelpj, what do you think?
We could also try to enabling it in CI only.

wz1000 (approved):

fendor:
Thank you for the patience! We appreciate the contribution and the improved test execution times!

develop7:
> Thank you for the patience! We appreciate the contribution and the improved test execution times!

Music to my ears! Feedback loop cannot be too short, period.

Issue #4401, opened 2024-09-12
Cabal Plugin is too strict for the newest cabal versions


fendor:
Generating a new package with cabal `3.12` via

```sh
$ mkdir test-hls-cabal
$ cd test-hls-cabal
$ cabal init --lib -n
```

Generates a `.cabal` file with `cabal-version: 3.12`, which `haskell-language-server-2.9.0.1` built with `ghc-9.6.5`, which is linked against `Cabal-3.10.3.0`. Consequentially, HLS highlights the whole `.cabal` file as incorrect, as it doesn't recognise `3.12` as a valid `cabal-version`.

![image](https://github.com/user-attachments/assets/1ac3087b-b220-496f-a721-413734d7db8c)

This is bad user-experience.

I am not sure whether we can fix that for good, as older HLS versions will never get linked against a newer `Cabal` version.
However, what we could do is to be more lenient when it comes to parsing `.cabal` files. 
We can scan the `.cabal` file for the line `cabal-version:` (`scanSpecVersion`) and if the version is not known to us, we parse the cabal file with the newest CabalSpecVersion available to the HLS binary.
The `cabal` features may still break if users actually need new cabal-the-spec-features, but for many future changes, we likely only display warnings and many other IDE features keep working.

The only real disadvantage is that we can't reliably report incorrect `cabal-version` numbers any more. Also, the proposed behaviour might be unexpected to users.

The last resort I can think of, is to simply not provide any IDE features if the cabal-version is seemingly too new for the Cabal version linked into HLS.

fendor:
It seems to be more difficult than hoped to parse a `.cabal` file with a newer `cabal-version` in a lenient way. The `cabal-version` is parsed in the `FieldGrammar`, so we would have to basically copy-paste major parts of `Cabal-syntax` to not verify the provided `CabalSpecVersion` is not equal to the actual parsed ones.

Thus, the only way I can think of now is to basically overwrite the `cabal-version` field in-memory.

Ignoring `.cabal` files with a too new version seems to be the less hacky approach.

michaelpj:
Yeah, I think just emitting a diagnostic saying that we don't know how to parse it is best? So pretty much the current behaviour, but with a nicer diagnostic that doesn't cover the whole file :sweat_smile:

fridewald:
I'll give it a try

fridewald:
[...]

PR #4425, opened 2024-10-12
gracefully handle errors for unsupported cabal version

fridewald:
This PR makes sure that line number 0 returned from cabal is handled correctly.

This caused the  marking of the whole cabal file if the cabal version is not supported as described in #4401

### DIFF
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
@@ -252,12 +252,13 @@ cabalRules recorder plId = do
                 -- We don't support the cabal version, this should not be an error, as the
                 -- user did not do anything wrong. Instead we cast it to a warning
                 regex = "Unsupported cabal-version [0-9]+.[0-9]*"
+                unsupportedCabalHelpText = "\nThe used cabal version is not fully supported by hls. This means that some functionallity might not work as expected.\nIf you face any issues try to downgrade to a supported cabal version."
                 errorDiags =
                   NE.toList $
                     NE.map
                       ( \pe@(PError pos text) ->
                           if text =~ regex
-                            then Diagnostics.warningDiagnostic file (Syntax.PWarning Syntax.PWTOther pos text)
+                            then Diagnostics.warningDiagnostic file (Syntax.PWarning Syntax.PWTOther pos (text <> unsupportedCabalHelpText))
                             else Diagnostics.errorDiagnostic file pe
                       )
                       pErrorNE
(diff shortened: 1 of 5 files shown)


plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
```diff
@@ -63,8 +63,9 @@ positionFromCabalPosition :: Syntax.Position -> Position
 positionFromCabalPosition (Syntax.Position line column) = Position (fromIntegral line') (fromIntegral col')
   where
     -- LSP is zero-based, Cabal is one-based
-    line' = line-1
-    col' = column-1
+    -- Cabal can return line 0 for errors in the first line
+    line' = if line == 0 then 0 else line-1
```
fridewald, line 67: Is this enough or should we handle `line <= 0` too?

fendor: I think it is safer to handle `line <= 0`.

fendor (approved):
LGTM, thanks!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Diagnostics.hs
fridewald: updated

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal.hs
```diff
@@ -252,12 +252,13 @@ cabalRules recorder plId = do
                 -- We don't support the cabal version, this should not be an error, as the
                 -- user did not do anything wrong. Instead we cast it to a warning
                 regex = "Unsupported cabal-version [0-9]+.[0-9]*"
+                unsupportedCabalHelpText = "\nThe used cabal version is not fully supported by hls. This means that some functionallity might not work as expected.\nIf you face any issues try to downgrade to a supported cabal version."
```
fendor, line 255: We can get the set of supported versions from https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Completion/Data.hs#L33 and display it here for the best UX.

You can pull out the lower bound into its own dedicated variable, e.g. 

```haskell
lowestSupportedCabalVersion = CabalSpecV2_2
```

and then use that. Or pull out the list of supported cabal versions... Either way is fine.

fendor (approved):
One little improvement to the much appreciated error message!

Otherwise, LGTM and we can merge.

fridewald:
Sorry for not responding, and thanks for adding the last missing change!

fendor:
No worries, it is usually no hurry, but we likely will have a release soon, I would love to have this change in :)

Issue #4289, opened 2024-06-08
Field and value completion does not work inside conditionals


VeryMilkyJoe:
Cabal completion seems to not be able to complete anything inside if statements.

Here, the default completion for vscode is suggested inside an if conditional, when it should be the library section fields:
![image](https://github.com/haskell/haskell-language-server/assets/22590431/8103a64c-4004-4a53-a11d-aa733d0dc7cc)
Here, the exposed modules values should be completed instead of the default vscode completions:
![image](https://github.com/haskell/haskell-language-server/assets/22590431/3c6102ca-42dd-454b-bb31-a7b2129e6e86)

The fields we are checking for completion are defined [here](https://hackage.haskell.org/package/Cabal-syntax-3.12.0.0/docs/Distribution-Fields.html). 
I think to fix this we can check the parent sections recursively instead of simply the most direct parent to figure out the stanza the cursor is located in.

fendor:
Do we need to change the `Context` then?

VeryMilkyJoe:
> Do we need to change the `Context` then?

I don't think it is necessary to change the type of `Context` for this issue, since we do not care whether we are inside a conditional for the completion.
We can simple set the `StanzaContext` to the nearest parent which is a Stanza.

fendor:
@dyniec is taking a stab at this :confetti_ball:

VeryMilkyJoe:
@dyniec are you still working on this issue?

dyniec:
I couldn't make it work. Please mark it as up for grabs

SamuelLess:
Hi, I'll give it a try :)

fendor:
Implemented by #4427

PR #4427, opened 2024-10-12
Cabal ignore if for completions (#4289)

SamuelLess:
If conditions were previously handled as normal sections, for which we don't have any completion information. Now, we ignore the if section and use the parent section for completion purposes.


fendor (changes requested):
Could we have a test for `else` and `elif` as well?

VeryMilkyJoe (approved):

fendor (approved):

fendor:
Closes #4427 4289

PR #4428, opened 2024-10-13
Show package descriptions in Cabal plugin

pbrinkmeier:
This PR adds a proof of concept for showing the description for a package when it's hovered in a `.cabal` file, e.g.:

image

This uses `cabal info <package name>` under the hood and parses its output.

### DIFF
--- a/docs/contributing/contributing.md
+++ b/docs/contributing/contributing.md
@@ -81,6 +81,14 @@ Running just the wrapper tests
 $ cabal test wrapper-test
 ```
 
+Running just the tests for a specific plugin
+
+```bash
+$ cabal test hls-<plugin-name>-plugin-tests
+# E.g.
+$ cabal test hls-refactor-plugin-tests
+```
+
 Running a subset of tests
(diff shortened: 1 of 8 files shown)


pbrinkmeier:
Huh, apparently `hoistMaybe` was added only recently. It's trivial to add our own impl but probably not worth it for this poc.

fendor:
I discussed this feature with @wz1000 and GHC is aware of all the package databases in the current compiler session. The package database contains the relevant info (e.g. `ghc-pkg --global describe base`), so we should be able to read the package db *somehow* (maybe even via `Cabal`), and then look for the package with the given name.

In short, there should be a way forward for this feature, that avoids calling `cabal info`, which is great :)

pbrinkmeier:
Great news! I'll happily incorporate that. Do you have any specific ideas for a way forward?

fendor:
Not yet, I need to do some research, but Ill write it down and let you know once I have some idea :)

Issue #4417, opened 2024-10-10
Show package description on hover for package dependencies


VeryMilkyJoe:
It would be nice to be able to see a description of a dependency when hovering over that dependency in a cabal file.

Packages already have such a description field, the tricky part of this issue is to extract a given package's description from the correct package database.
We should already have access to the package database, but we would need to be able to decide which package database corresponds to the current cabal file.

pbrinkmeier:
I am currently working on this and #4416. As a first approximation we are going to call out to `cabal info <package>` when hovering a dependency. In the future we might be able to coax `Cabal` to provide this information through its API.

As a reference for the `cabal info` output format, I'll be using the [implementation in `cabal-install`](https://github.com/haskell/cabal/blob/256f85d735310196d437ede86046f2a86efedc6d/cabal-install/src/Distribution/Client/List.hs#L182).

pbrinkmeier:
![image](https://github.com/user-attachments/assets/2a2c422e-cced-4b0e-8c1b-65b8e2bc62cb)

My current impl simply dumps the `cabal info` output into the tooltip

Issue #4416, opened 2024-10-10
Show information on installed package versions


VeryMilkyJoe:
It would be useful to be able to see the installed version for each package dependency in a cabal file.

We could either show this in form of a code lens, or on hover, but we probably want both, seeing that users might turn off code lenses.

For implementing this, we can look at the project's  `plan.json` file and simply find the entry where `name`  corresponds to the package's name and then show the version of this entry.

lucalabs-de:
I'll work on this.

PR #4430, opened 2024-10-14
Show version of installed packages

lucalabs-de:
Implements #4416. Shows installed versions of cabal dependencies as inlay hints, or alternatively as code lenses when inlay hints are not supported.

2024-10-14-152038_1271x1059_scrot

Code Lenses are currently always displayed at the beginning of the line, I'm not sure if this is an error on my part or a VSCode thing. Depending on how the cabal file is formatted, this can look unpleasant:

2024-10-14-152248_867x341_scrot

I don't yet show the version information in the hover overlay as mentioned in the issue, since I was not sure how it should interplay with #4417. However, it should be a quick change to show it there as well.


fendor:
Discussing the feature with @wz1000, HLS knows the exact package versions as it intercepts the `cabal build` requests containing the ghc-options `-package-id base-4... -package-id aeson-2.2...`. 

That means, given the `HscEnv` a particular `.cabal` version belongs to, we should be able to consult the `UnitState` and `UnitInfoMap` in particular to find the exact version dependencies without resorting to reading `plan.json`.

lucalabs-de:
That certainly seems like a more sensible approach. I'm definitely interested in implementing it. You can explain it here or we can pair, whatever works best for you.

PR #4433, opened 2024-10-17
Support structured diagnostics 2

noughtmare:
This is my attempt at finishing the work of @dylan-thinnes on [!4311](https://github.com/haskell/haskell-language-server/pull/4311).

Addresses https://github.com/haskell/haskell-language-server/issues/2014

Should hopefully enable https://github.com/haskell/haskell-language-server/issues/3246

These are two things Dylan listed as work for future PRs:

*    Replacing regexes with structured diagnostic checks
*    Improve typing so that it can be known that an error has a structured diagnostic, instead of defaulting to Maybe

Some future work discovered in reviews of this PR:

* `requireDiagnostic` could take a more structured error code query type as argument. See https://github.com/haskell/haskell-language-server/pull/4433#discussion_r1813073104
* The HTTP work should be un-reverted in a future PR: 1bc221c1d66cb5107b052acbf212198a54fa9ed6
* Check that tests using `requireDiagnostics` and related functions use explicit error codes if possible


noughtmare:
All failures in the jobs of e651d41 seem to be spurious (except for the stylish-haskell formatting failure).

noughtmare:
I think the remaining failures are not caused by this PR. Are these kinds of spurious failures common in HLS?

Also, I would really like to get this merged before the next release, because I want to develop a VS Code extension which uses these diagnostic numbers.

fendor:
> Are these kinds of spurious failures common in HLS?

fendor:
Some errors in ghc `9.4` seem to be genuine, though. Many tests are seemginly time outing for some reason.

noughtmare:
You're right. Can anyone give me some tips on how to profile HLS to see where these timeouts are coming from?

fendor:
@noughtmare Profiling is likely not necessary, I think the tests are waiting for an LSP message that never arrives, and then just wait for 60s and time out.

Use `HLS_TEST_LOG_STDERR=1 cabal test ...` and `TASTY_PATTERN` to run a specific test with all HLS logging enabled.
Printf debugging also works quite well to figure out where the tests gets stuck.
To get the raw `LSP` messages, you can use `LSP_TEST_LOG_MESSAGES=1` or `LSP_TEST_LOG_STDERR=1 ` (they do the same thign)

ghcide/src/Development/IDE/GHC/Compat.hs
```diff
@@ -310,12 +310,25 @@ corePrepExpr _ = GHC.corePrepExpr
...
+  let renderMsgs extractor = (fmap . fmap) GhcPsMessage . getMessages $ extractor msgs
+  in (renderMsgs psWarnings, renderMsgs psErrors)
+#else
   let renderMsgs extractor = (fmap . fmap) renderDiagnosticMessageWithHints . getMessages $ extractor msgs
   in (renderMsgs psWarnings, renderMsgs psErrors)
+#endif
 
+#if MIN_VERSION_ghc(9,6,1)
+pattern PFailedWithErrorMessages :: forall a b. (b -> Bag (MsgEnvelope GhcMessage)) -> ParseResult a
+#elif MIN_VERSION_ghc(9,3,0)
```
noughtmare, line 323: TODO: make this the fallthrough case, we don't support GHC versions below 9.4 any more.

ghcide/src/Development/IDE/GHC/Compat/Driver.hs
```diff
@@ -0,0 +1,144 @@
...
+import           GHC.Utils.Error                 (emptyMessages)
+#endif
+
+hscTypecheckRenameWithDiagnostics :: HscEnv -> ModSummary -> HsParsedModule
+                   -> IO ((TcGblEnv, RenamedStuff), Messages GhcMessage)
+hscTypecheckRenameWithDiagnostics hsc_env mod_summary rdr_module =
+#if MIN_VERSION_ghc(9,6,1)
+    runHsc' hsc_env $ hsc_typecheck True mod_summary (Just rdr_module)
+#else
+    (,emptyMessages) <$> hscTypecheckRename hsc_env mod_summary rdr_module
```
noughtmare, line 54: Through the magic of inserting `false &&` in these CPP blocks, I've figured out this is the problematic one. But it is still mysterious to me that this causes errors; it seems very benign.

noughtmare: @fendor I'm getting stuck debugging this. Every caller of this typechecking code path seems to complete fine all the way up to the shake rules. However, I don't see where the resulting LSP message is supposed to be produced.

fendor: `emptyMessages` looks suspicious to me, maybe we are not propagating diagnostics occurring in the renamer phase, not producing code actions for all messages originating from that phase?

noughtmare: My thoughts were that it returns no messages because diagnostic numbers were only introduced in 9.6, but maybe that is indeed not the whole story. I wonder if 9.4 ever worked with the original PR.

fendor: The messages are presumably the diagnostics, not the error codes themselves, right? 

I can try to take a closer look some time this week, but I am not terribly familiar with how it used to work.

noughtmare: Yes, that does sound fishy. I'll take another look with that in mind.

noughtmare: I think it was just a dummy implementation. Now I've changed it to do the same thing as we do for 9.6 and it seems to work locally for me.

noughtmare:
I've decided to make the tests ignore all error codes on GHC 9.4. I think some errors already have codes in 9.4, so we could do a bit more, but I'm not sure if it is worth the effort.

noughtmare:
The remaining MacOS failure seems like a compiler bug. The linker is throwing an error.

noughtmare:
Oh, I see we already have this workaround in some places. I'll just copy that.

noughtmare:
Yay, it has finally passed CI. Should I squash the commits? I'm not sure what to write in the message as I don't really know the details of what exactly Dylan did.

fendor:
We usually squash the PR when merging, so I don't think it is particularly necessary to squash it yourself, unless you want to add a custom message.

Awesome that the CI of the PR is green!
Let's try to get some reviews in.

fendor:
Thanks! However, you may be slightly unhappy to hear that I haven't reviewed the previous PR at all yet, so I need to look at everything.

I noticed https://github.com/haskell/haskell-language-server/pull/4433/commits/14d66975decfa01187aa1260ea5e3e8a17ef6d1b is still in this PR. In the discussion https://github.com/haskell/haskell-language-server/pull/4311#discussion_r1660206034 it was basically decided to not include this in this PR.
Could you take this out?

ghcide/ghcide.cabal
```diff
@@ -83,8 +83,10 @@ library
     , hiedb                        ^>= 0.6.0.0
     , hls-graph                    == 2.9.0.1
     , hls-plugin-api               == 2.9.0.1
+    , http-conduit
```
fendor, line 86: Let's split out the `http` stuff, it is unrelated (but a welcome addition in general) to the structured diagnostics

hls-test-utils/src/Development/IDE/Test/Diagnostic.hs
```diff
@@ -32,6 +33,17 @@ requireDiagnostic actuals expected@(severity, cursor, expectedMsg, expectedTag)
...
+        && codeMatches d
+
+    codeMatches d =
+#if MIN_VERSION_GLASGOW_HASKELL(9,5,0,0)
+        case (mbExpectedCode, _code d) of
+          (Nothing, _)                         -> True
+          (Just expectedCode, Nothing)         -> False
+          (Just expectedCode, Just actualCode) -> InR expectedCode == actualCode
+#else
+        True
```
fendor, line 45: Could we get away with avoiding CPP here? 
`Compat.hs` defines:
```
data GhcVersion
  = GHC94
  | GHC96
  | GHC98
  | GHC910
  deriving (Eq, Ord, Show, Enum)
  ```
  
 which could turn this CPP into a run-time check like:
 ```suggestion
    codeMatches d
      | ghcVersion >= GHC96 =
          case (mbExpectedCode, _code d) of
            (Nothing, _)                         -> True
            (Just expectedCode, Nothing)         -> False
            (Just expectedCode, Just actualCode) -> InR expectedCode == actualCode
      | otherwise = True
```
 
Unless I am overlooking something

hls-test-utils/src/Development/IDE/Test/Diagnostic.hs
```diff
@@ -17,9 +18,9 @@ type ErrorMsg = String
 requireDiagnostic
     :: (Foldable f, Show (f Diagnostic), HasCallStack)
     => f Diagnostic
-    -> (DiagnosticSeverity, Cursor, T.Text, Maybe DiagnosticTag)
+    -> (DiagnosticSeverity, Cursor, T.Text, Maybe T.Text, Maybe DiagnosticTag)
```
fendor, line 21: I think `Maybe T.Text` is not a good assertion type. It works for now, but it will get annoying once new error types are added, then we need awkward conditionals in the test cases.

In the future, we might want to have something like 

```haskell
data ErrorCode = NoCode | GhcCode Int | GhcCodeSince GhcVersion Int
```

However, this is not a blocker for this PR, I mostly want to know your opinion on this.

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +76,86 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
 -- | an IdeResult with a fingerprint
 type IdeResultNoDiagnosticsEarlyCutoff  v = (Maybe ByteString, Maybe v)
 
-ideErrorText :: NormalizedFilePath -> T.Text -> FileDiagnostic
-ideErrorText = ideErrorWithSource (Just "compiler") (Just DiagnosticSeverity_Error)
+ideErrorText :: Maybe (MsgEnvelope GhcMessage) -> NormalizedFilePath -> T.Text -> FileDiagnostic
+ideErrorText origMsg fdFilePath msg =
```
fendor, line 80:
```suggestion
ideErrorText ::  NormalizedFilePath -> T.Text -> Maybe (MsgEnvelope GhcMessage) -> FileDiagnostic
ideErrorText fdFilePath msg origMsg =
```

Maybe to be consistent with `ideErrorFromLspDiag`?

ghcide/test/exe/FuzzySearch.hs
```diff
@@ -65,6 +65,7 @@ replaceAt t i c =
 dictionaryPath :: FilePath
 dictionaryPath = "/usr/share/dict/words"
 
+{-# ANN dictionary ("HLint: ignore Avoid restricted function" :: String) #-}
```
fendor, line 68: Is this still needed? I think I would prefer this exception to be added to the `.hlint.yaml` file, but this is just my opinion.

fendor (commented):
Start of the review, didn't get to the meat of the PR yet

hls-test-utils/src/Development/IDE/Test/Diagnostic.hs
noughtmare: You're right, this was me being lazy and not wanting to figure out how to import `ghcVersion`. My bad.

hls-test-utils/src/Development/IDE/Test/Diagnostic.hs
noughtmare: It seems like a good idea. But I think we should indeed defer that to a future PR, because it requires parsing the error code, and I think we might not only want GHC codes in the future, for example cabal also has diagnostic codes.

noughtmare:
One thing I don't understand is what the final codes actually are. When we construct them it seems to include the `GHC-` prefix:

https://github.com/haskell/haskell-language-server/pull/4433/files#diff-15ac916db5d1445c16a0cb4a644ecc44854ce21d0f648b07663212035766b1feR101-R121

But when we use `requireDiagnostic` we do not include the prefix:

https://github.com/haskell/haskell-language-server/pull/4433/files#diff-52be86ee0fb96c538fa0bf2d9872b8aa41a8dd1b0f434383ce6f9301d4c39572R269

Why does this pass the tests?

hls-test-utils/src/Development/IDE/Test/Diagnostic.hs
fendor: I want to avoid the parsing by using different constructors for each error code type, e.g. `GhcCode` vs `CabalCode` and so on.

noughtmare: I mean the LSP diagnostics we get are strings/text, so we'd need to parse that.

fendor:
I think the linked code is wrong, the benchmark isn't executed on each PR, see https://github.com/haskell/haskell-language-server/actions/runs/11481138334/job/31951108378?pr=4433, but only nightly or on demand. So good catch :)

ghcide/test/exe/FuzzySearch.hs
noughtmare: I think it is better to keep this annotation. It is warning about the use of `unsafePerformIO` which should be considered per function in my opinion.

hls-test-utils/src/Development/IDE/Test.hs
```diff
@@ -114,25 +114,25 @@ flushMessages = do
 --
 --   Rather than trying to assert the absence of diagnostics, introduce an
 --   expected diagnostic (e.g. a redundant import) and assert the singleton diagnostic.
-expectDiagnostics :: HasCallStack => [(FilePath, [(DiagnosticSeverity, Cursor, T.Text)])] -> Session ()
+expectDiagnostics :: HasCallStack => [(FilePath, [(DiagnosticSeverity, Cursor, T.Text, Maybe T.Text)])] -> Session ()
```
soulomoon, line 117: Probably a good place to add doc about what `Maybe T.Text` is.

plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -1996,7 +1995,7 @@ suggestImportDisambiguationTests = testGroup "suggest import disambiguation acti
     compareHideFunctionTo = compareTwo "HideFunction.hs"
     withTarget file locs k = runWithExtraFiles "hiding" $ \dir -> do
         doc <- openDoc file "haskell"
-        void $ expectDiagnostics [(file, [(DiagnosticSeverity_Error, loc, "Ambiguous occurrence") | loc <- locs])]
+        void $ expectDiagnostics [(file, [(DiagnosticSeverity_Error, loc, "Ambiguous occurrence", Nothing) | loc <- locs])] -- TODO: Give this a proper error
```
soulomoon, line 1998: Are we going to finish this TODO in this PR?

soulomoon (commented):
Good work, a Just a few neatpicks and a question

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -80,13 +172,60 @@ data ShowDiagnostic
...
 --   on either the console or in the IDE at the right source location.
 --
-type FileDiagnostic = (NormalizedFilePath, ShowDiagnostic, Diagnostic)
+--   It also optionally keeps a structured diagnostic message GhcMessage in
+--   StructuredMessage.
+--
+data FileDiagnostic = FileDiagnostic
+  { fdFilePath             :: NormalizedFilePath
+  , fdShouldShowDiagnostic :: ShowDiagnostic
+  , fdLspDiagnostic        :: Diagnostic
```
soulomoon, line 215: I am wondering about these two field `fdLspDiagnostic` and `StructuredMessage`,
I guess we have some duplicated information between them and how would they envolve in the future? Any plan?

noughtmare: I haven't really looked at the contents of this PR in detail. So, I'll take a look, but I'd also love to hear suggestions.

plugins/hls-refactor-plugin/test/Main.hs
noughtmare: I think it should be `GHC-87543`

hls-test-utils/src/Development/IDE/Test.hs
noughtmare: I think I'll just make this four-tuple into its own data type. I'll make another for the expected diagnostic with a tag.

noughtmare: Nooo... That way all the tests need to be rewritten. Ugh...

noughtmare: I guess a type synonym it is then.

plugins/hls-refactor-plugin/test/Main.hs
noughtmare: There are several others with `Nothing` in the place of the code. I don't know why this todo was only put on this particular occurrence. There are probably some errors that don't have codes, but I think many of the current `Nothing`s could be replaced. 

I guess I'll try replacing this one and put a note in the list of future work that we should check the rest.

ghcide/src/Development/IDE/Types/Diagnostics.hs
noughtmare: I don't know what the purpose of the structured message field is. It seems to be unused.

plugins/hls-refactor-plugin/test/Main.hs
noughtmare: Turns out GHC 9.6 doesn't have an error code for this, but GHC 9.8 does have it: `GHC-87110`

noughtmare:
Interesting, it seems the stack build uses more warnings than the cabal build.

plugins/hls-refactor-plugin/test/Main.hs
noughtmare: I don't really want to open the can of worms of changing the tests based on which GHC version is used, so I think leaving it as is might be the best option. Do you agree?

ghcide/src/Development/IDE/Types/Diagnostics.hs
noughtmare: It seems I just need to comment out three places to remove this. I guess I'll just do that, because we can easily add it back if it turns out we do need it for something in the future.

noughtmare:
@wz1000 I think I removed that in 0776c65, because we don't use it anywhere yet and it would be very simple to add back when we do actually need it.

noughtmare:
Thank you for the reviews! I hope I have addressed all feedback appropriately. Is there anything more I can do to speed up the process of merging this?

fendor:
Probably not, except for regularly pinging on this thread to make sure we get enough reviews, soon :) We need two approvals, considering the importance of this change, we shouldn't rush merging it without proper review time.

However, I hope we can merge it within the next two weeks (hopefully a pessimistic timeframe, but I don't really want to give any timeestimate :D ).

plugins/hls-refactor-plugin/test/Main.hs
soulomoon: Yes, probably this is the best, we can wait until the version updated and then finish the todo by then

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -875,10 +876,19 @@ newComponentCache recorder exts _cfp hsc_env old_cis new_cis = do
     hscEnv' <- -- Set up a multi component session with the other units on GHC 9.4
               Compat.initUnits dfs hsc_env
 
-    let closure_errs = checkHomeUnitsClosed' (hsc_unit_env hscEnv') (hsc_all_home_unit_ids hscEnv')
-        multi_errs = map (ideErrorWithSource (Just "cradle") (Just DiagnosticSeverity_Warning) _cfp . T.pack . Compat.printWithoutUniques) closure_errs
+    let closure_errs = maybeToList $ checkHomeUnitsClosed' (hsc_unit_env hscEnv') (hsc_all_home_unit_ids hscEnv')
```
soulomoon, line 879: ` checkHomeUnitsClosed'` changed then adding `maybeToList` looks like redundant changes

soulomoon (commented):
Just a neatpick, other places look good to me.

ghcide/src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -44,26 +75,82 @@ type IdeResult v = ([FileDiagnostic], Maybe v)
...
+  :: LSP.Diagnostic
+  -> NormalizedFilePath
+  -> Maybe (MsgEnvelope GhcMessage)
+  -> FileDiagnostic
+ideErrorFromLspDiag lspDiag fdFilePath mbOrigMsg =
+  let fdShouldShowDiagnostic = ShowDiag
+      fdLspDiagnostic =
+        lspDiag
+          & attachReason (fmap (diagnosticReason . errMsgDiagnostic) mbOrigMsg)
+          & setGhcCode mbOrigMsg
```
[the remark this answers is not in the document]
noughtmare: I think so, yes.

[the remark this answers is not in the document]
noughtmare: I was actually looking at when this function is called with a `Just` and couldn't easily find anything, because there are so many callers that pass along a value for which you then have to transitively search.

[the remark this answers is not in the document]
noughtmare: OK, I found this now:

* https://github.com/haskell/haskell-language-server/pull/4433/files/0776c65490d8840d03ec8740f0c8048ea1d55b35#diff-9ec5d8d4031c00a406a3c0ce65ca9e155f3a05d7014a97469f89b5fd3bb65bb9R71

fendor:
In the last meeting, we discussed on how to continue with this PR, for transparency, these are the notes https://docs.google.com/document/d/1zqRhA3uoiYgA_Q8eDpIEG0OiFcofGqkVP7N4ODDmikM/edit?tab=t.0

In short, I will review soon and try to make use of the structured diagnostics in 1 or 2 of the plugins, to demonstrate how we can use them in the future and to make sure this PR takes the right directions.
Are you fine with me pushing to this branch, or should I open PRs against your branch?

noughtmare:
I'm fine with pushing directly to my branch if that is possible. Do I need to give you special permissions?

fendor:
I don't know, I will just try to push at some point and ping you if I have any trouble :)

fendor:
@noughtmare Is there anythign else you need or want from this PR?

noughtmare:
I personally just wanted access to the diagnostic codes. So I'm more than satisfied!

soulomoon (commented):
LGTM, Just two questions

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
```diff
@@ -80,23 +86,25 @@ addMethodPlaceholders _ state _ param@AddMinimalMethodsParams{..} = do
 -- This implementation is ad-hoc in a sense that the diagnostic detection mechanism is
 -- sensitive to the format of diagnostic messages from GHC.
 codeAction :: Recorder (WithPriority Log) -> PluginMethodHandler IdeState Method_TextDocumentCodeAction
-codeAction recorder state plId (CodeActionParams _ _ docId _ context) = do
+codeAction recorder state plId (CodeActionParams _ _ docId caRange _) = do
```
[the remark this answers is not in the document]
fendor: Indeed, that's exactly what we are doing :)

Re 1., I think `ShakeExtra` also contains Diagnostics which are not displayed to the user, see the docs for `ShowDiagnostic`:
```haskell
-- | Defines whether a particular diagnostic should be reported
--   back to the user.
--
--   One important use case is "missing signature" code lenses,
--   for which we need to enable the corresponding warning during
--   type checking. However, we do not want to show the warning
--   unless the programmer asks for it (#261).
data ShowDiagnostic
    = ShowDiag  -- ^ Report back to the user
    | HideDiag  -- ^ Hide from user
    deriving (Eq, Ord, Show)
 ```
This means, we may have more diagnostics available. 

Re 2., I confidently, but without evidence, claim that the diagnostics sent by the client and the diagnostics we store in `StoreExtras` may be out of sync, but it has no noticeable negative effect for the user experience.
 
 However, why are we doing this? Primarily due to the docs of [`CodeActionContext`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionContext):
 ```javascript
   /**
	 * An array of diagnostics known on the client side overlapping the range
	 * provided to the `textDocument/codeAction` request. They are provided so
	 * that the server knows which errors are currently presented to the user
	 * for the given range. There is no guarantee that these accurately reflect
	 * the error state of the resource. The primary parameter
	 * to compute code actions is the provided range.
	 */
	diagnostics: [Diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic)[];
 ```
 
 Which tells us that we cannot rely on the diagnostics provided by the `CodeActionContext`.
 I will add this crucial bit of information to the documentation of `activeDiagnostics...` function.

File: plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs [hunk 14]
```diff
@@ -112,15 +112,15 @@ instance NFData InstanceBindLensResult where
...
   | LogShake Shake.Log
 
 instance Pretty Log where
   pretty = \case
     LogImplementedMethods cls methods ->
-      pretty ("Detected implemented methods for class" :: String)
+      pretty ("The following methods are missing" :: String)
         <+> pretty (show (getOccString cls) <> ":") -- 'show' is used here to add quotes around the class name
-        <+> pretty methods
+        <+> pretty (showSDocUnsafe $ ppr methods)
```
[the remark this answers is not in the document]
fendor: I think that's fine for logging, `unsafe` doesn't mean it might crash, but rather we use default `DynFlags` which may be missing some information, such as import paths, packages in scope, etc... So, the displayed names may be inaccurate. However, I can double-check what we do in other locations

plugins/hls-class-plugin/src/Ide/Plugin/Class/CodeAction.hs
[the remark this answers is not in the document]
soulomoon: Thanks, adding it to the docs helps for the future reviewers.:)

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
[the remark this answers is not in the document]
soulomoon: Maybe we should get the more exact `DynFlags` from rule like `GetModSummary`, but, I agree, for logging, it should be fine as long as it won't crash.

fendor:
Since the release is stalled due to CI issues, I am inclined to just push the trigger and be done with it. Opinions, @soulomoon, @michaelpj, @wz1000 ?

soulomoon:
AFAIK, there’s no downside, aside from the potential for future development to utilize it. Therefore, I believe it’s fine to go ahead and push it.

soulomoon (approved):

PR #4437, opened 2024-10-21
Enable hls-stan-plugin for GHC 9.10.1

fendor:


fendor:
I also checked other plugins, but floskell, hlint, retrie and stylish-haskell are still blocked.

soulomoon (approved):
LGTM!

PR #4441, opened 2024-10-29
Don't suggest -Wno-deferred-out-of-scope-variables

### LINKED ISSUE #4440: Don't suggest `-Wno-deferred-out-of-scope-variables` as a possible action.

With code

```haskell
f = let x = Doesn'tExist
     in undefined
```
when I execute code action on `Doesn'tExist`, HLS (haskell-language-server version: 2.9.0.0 (GHC: 9.6.6)) adds `{-# OPTIONS_GHC -Wno-deferred-out-of-scope-variables #-}` on the top of my file. It does this, I believe, because this is the only code action available. I don't think this is the solution that user wants in _most_ cases. At least in Emacs it is easy to miss that HLS did that (if only one action is available, it is executed), leading to confusion on as to why, suddenly, HLS "doesn't work". It works, of course, because with that flag there is no longer a compile time error.

Don't suggest `{-# OPTIONS_GHC -Wno-deferred-out-of-scope-variables #-}`.

Restarting HLS... a lot :) I don't think there is an alternative.

jeukshi:
Fixes #4440

This does the trick, I don't get the code action on my local branch.

~I'm just unsure if those tests do what they are supposed to do.~

~If I disable blacklist~

```haskell
warningBlacklist :: [T.Text]
warningBlacklist =
  -- Don't suggest disabling type errors as a solution to all type errors.
  [ --"deferred-type-errors"
  -- Don't suggest disabling out of scope errors as a solution to all out of scope errors.
  --, "deferred-out-of-scope-variables"
  ]
```

~they do pass on my machine. The list is empty anyway, also in the case of deferred-type-errors. But that is beyond my current knowledge of the codebase.~

Edit: fixed.

### DIFF
--- a/plugins/hls-pragmas-plugin/test/Main.hs
+++ b/plugins/hls-pragmas-plugin/test/Main.hs
@@ -109,10 +109,15 @@ codeActionTests' =
         _ -> assertFailure $ "Expected one code action, but got: " <> show cas
       liftIO $ (ca ^. L.title == "Add \"NamedFieldPuns\"") @? "NamedFieldPuns code action"
       executeCodeAction ca
-  , goldenWithPragmas pragmasSuggestPlugin "doesn't suggest disabling type errors" "DeferredTypeErrors" $ \doc -> do
+  , goldenWithPragmas pragmasDisableWarningPlugin "doesn't suggest disabling type errors" "DeferredTypeErrors" $ \doc -> do
       _ <- waitForDiagnosticsFrom doc
       cas <- map fromAction <$> getAllCodeActions doc
       liftIO $ "Disable \"deferred-type-errors\" warnings" `notElem` map (^. L.title) cas @? "Doesn't contain deferred-type-errors code action"
+      liftIO $ length cas == 0 @? "Expected no code actions, but got: " <> show cas
+  , goldenWithPragmas pragmasDisableWarningPlugin "doesn't suggest disabling out of scope variables" "DeferredOutOfScopeVariables" $ \doc -> do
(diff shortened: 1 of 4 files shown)


plugins/hls-pragmas-plugin/test/Main.hs
```diff
@@ -114,6 +114,11 @@ codeActionTests' =
       cas <- map fromAction <$> getAllCodeActions doc
       liftIO $ "Disable \"deferred-type-errors\" warnings" `notElem` map (^. L.title) cas @? "Doesn't contain deferred-type-errors code action"
       liftIO $ length cas == 0 @? "Expected no code actions, but got: " <> show cas
+  , goldenWithPragmas pragmasSuggestPlugin "doesn't suggest disabling out of scope variables" "DeferredOutOfScopeVariables" $ \doc -> do
```
fendor, line 117:
```suggestion
  , goldenWithPragmas pragmasDisableWarningPlugin "doesn't suggest disabling out of scope variables" "DeferredOutOfScopeVariables" $ \doc -> do
```

Wrong plugin :) Same goes for the other test, if you dont mind, could you update that test as well? 

Thanks!

fendor (approved):
Awesome, thanks!
One small issue, otherwise gtg!

plugins/hls-pragmas-plugin/test/Main.hs
jeukshi: Yep, that was it. Both test now actually work.

VeryMilkyJoe (approved):

Issue #694, opened 2020-12-23
Upload new logo


GuillaumeDesforges:
https://github.com/haskell/haskell-ide-engine/issues/267

Any logo for HLS?

-----------------------------------
EDIT (@jneira)

### We are choosing our new logo through an open survey. ~~You can vote in the comment [below](https://github.com/haskell/haskell-language-server/issues/694#issuecomment-762701777).~~

--------------------
EDIT (@Ailrun)

### We decided our new logo. You can check the result in the comment [below](https://github.com/haskell/haskell-language-server/issues/694#issuecomment-767741115) and related [PR](https://github.com/haskell/haskell-language-server/pull/1267).

jneira:
Absolutely! Afair we tried to adapt the hie one (simply changing ie for ls) when it was started but we did not finish anything

alanz:
I propose we open it up for contributions again, and vote, as we did for HIE.

alanz:
[Called for contributions](https://mail.haskell.org/pipermail/haskell-cafe/2020-December/133253.html)

> Hi all,
>
> The haskell language server is coming up for its first birthday in
January.  To celebrate we would like to collectively choose a logo.
>
> So if you have a creative streak and would like to donate a lasting
contribution to the project, please submit for consideration at [1].
>
> We will aim to close the judging on 25 Jan 2021, being the anniversary of
the Bristol Hackathon[2] where HLS was born.
>
> Alan
> 
> [1] https://github.com/haskell/haskell-language-server/issues/694
> [2] https://mpickering.github.io/bristol2020.html

gizmo-mk0:
![hls-logo](https://user-images.githubusercontent.com/8467908/103012469-94618600-4533-11eb-9e18-b521e7d43021.png)

sjcjoosten:
![Logo_animated](https://user-images.githubusercontent.com/7753626/103015547-a1b84980-450e-11eb-8578-bb17cda714e9.png)
[Still files as svg, png, etc](https://github.com/haskell/haskell-language-server/files/5736072/Stills.zip)

sjcjoosten:
![ProposalSJC](https://user-images.githubusercontent.com/7753626/103015736-f360d400-450e-11eb-85a8-b93ed6aaca5e.png)
[Separate Files as pdf and png](https://github.com/haskell/haskell-language-server/files/5736079/SeparateFiles.zip)

alanz:
[...]

PR #4442, opened 2024-10-30
Clean up logo

georgefst:
I might be the only person who really cares about this, but every time I've seen this logo in the last [four years](https://github.com/haskell/haskell-language-server/issues/694#issuecomment-767741115) the fact that nothing is aligned properly has irritated me.

I can match the gaps and thicknesses of the old version more closely if people genuinely prefer it, but I've taken the measurements from the actual Haskell logo instead.


fendor:
Thanks for the PR! 
I don't have any particularly strong opinions, both look good to me :)
Pinging @Ailrun as they designed the logo originally, afaict.

Ailrun:
I am the original designer. I tried the "fully aligned" version when I designed it, but it looked awful for me because the lambda suppressed L and S. Thus, those "misalignments" are somewhat intended in terms of overall balance, like many other logos out there (e.g. Google logo). However, I'm all ears if others think that aligned version is better as I don't think my arrangement was anywhere near perfect.

Ailrun:
If we decide to change the logo: I suggest that making L and S provide more support for the lambda, so that the logo can express that this Language Server (what L and S stand for) provides a stable base for Haskell (what the Haskell logo stands for) development.

Ailrun:
@georgefst By that basically, I mean making the top part of L and S wider enough than the bottom part of lambda. One possibility is that I over-widened the top part to a visually unpleasing level, but I still believe the linearly matching width is not wide enough.

georgefst:
Oh, I think I see. Your issue is that the bottom-left corner of the lambda is not "supported", in the sense that there's nothing underneath it? I think that could only be rectified by dropping the arrow from the logo, i.e. the darker purple part on the left. Then again, that could be cool, because it would bring us down to three glyphs, so we could use one colour for each. I might play around with this, although it's a more radical change than I intended...

georgefst:
So roughly like this:

hls

Ailrun:
@georgefst Not really about the floating leg of lambda. Maybe I should have used "Haskell logo" instead of "lambda character" in the above description. I don't think it as a physical model (if I did, then I would be against that floating leg). It is more about that there is some visual intuition that affects people's view. When something is smaller, people could think that is less important. If something stands on a tight space, it seems unstable. If that tight space stands for something, then that something gives an impression of a weak foundation.

georgefst:
I'd like to move this discussion along. Is anyone actually against merging this in its current form?

Ailrun:
As I said, currently as-is, I am against the merging.

georgefst:
Okay. Sorry if I came across a bit dismissive. I had forgotten some of the thread context when I first posted.

Bodigrim:
I'm not quite sure that a casual viewer understands that thick strokes of L and S, especially their bottoms, are to symbolise a concrete basis for Haskell development. I'd prefer a more aligned version with the same thickness in all lines. 

If we wish to put more emphasis on the foundation, could we add an additional solid line (well, trapezium) under L and S?

PR #4447, opened 2024-11-10
Support record positional construction inlay hints

### LINKED ISSUE #4212: Inlay hints for positional record construction

When using a data constructor that has record fields positionally, it would be nice to know what the field names are, and maybe have the option to convert to record synax.

e.g. this
```
Foo a b c
```
becomes
```
Foo <hint>x=</hint>a <hint>y=</hint>b <hint>z=</hint>c
```

### LINKED ISSUE #4391: Code action: switch to record syntax

My codebase is full of pieces of record that are constructed using "function" syntax. Something such as:

```haskell
data Foo = Foo {
   count :: Int,
   name :: String
}

boz = Foo 10 "hello"
```

I would like a code action available on `Foo` in `boz` which expand the syntax to record field syntax. Said otherwise, after usage, I would like:

```haskell
boz = Foo 10 "hello"
```

to become:

```haskell
boz = Foo { count = 10, name = "hello" }
```

## Additional context

This can also be extended to other cases where record are used in "function" form, example for pattern matching:

```haskell
biz (Foo a b) = ...
```

Could be turned to:

```haskell
biz (Foo {count = a, name = b}
```

If you are interested and if the initial design is ok, I can give a try to the implementation.

jetjinser:
Resolves #4391, resolves #4212.

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -145,7 +145,7 @@ codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
   let actions = map (mkCodeAction enabledExtensions) (RangeMap.filterByRange range crCodeActions)
   pure $ InL actions
   where
-    mkCodeAction :: [Extension] ->  Int -> Command |? CodeAction
+    mkCodeAction :: [Extension] -> Int -> Command |? CodeAction
     mkCodeAction exts uid = InR CodeAction
       { _title = mkTitle exts
       , _kind = Just CodeActionKind_RefactorRewrite
@@ -186,14 +186,14 @@ inlayHintProvider _ state pId InlayHintParams {_textDocument = TextDocumentIdent
                   , uid <- RangeMap.elementsInRange range crCodeActions
                   , Just record <- [IntMap.lookup uid crCodeActionResolve] ]
         -- Get the definition of each dotdot of record
-        locations = [ getDefinition nfp pos
+        locations = [ fmap (,record) (getDefinition nfp pos)
                     | record <- records
                     , pos <- maybeToList $ fmap _start $ recordInfoToDotDotRange record ]
     defnLocsList <- lift $ sequence locations
-    pure $ InL $ mapMaybe (mkInlayHints crr pragma) (zip defnLocsList records)
+    pure $ InL $ mapMaybe (mkInlayHint crr pragma) defnLocsList
    where
-     mkInlayHints :: CollectRecordsResult -> NextPragmaInfo -> (Maybe [(Location, Identifier)], RecordInfo) -> Maybe InlayHint
-     mkInlayHints CRR {enabledExtensions, nameMap} pragma (defnLocs, record) =
+     mkInlayHint :: CollectRecordsResult -> NextPragmaInfo -> (Maybe [(Location, Identifier)], RecordInfo) -> Maybe InlayHint
+     mkInlayHint CRR {enabledExtensions, nameMap} pragma (defnLocs, record) =
        let range = recordInfoToDotDotRange record
            textEdits = maybeToList (renderRecordInfoAsTextEdit nameMap record)
                     <> maybeToList (pragmaEdit enabledExtensions pragma)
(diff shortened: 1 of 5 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -145,9 +154,9 @@ codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
   let actions = map (mkCodeAction enabledExtensions) (RangeMap.filterByRange range crCodeActions)
   pure $ InL actions
   where
-    mkCodeAction :: [Extension] ->  Int -> Command |? CodeAction
+    mkCodeAction :: [Extension] -> Int -> Command |? CodeAction
     mkCodeAction exts uid = InR CodeAction
-      { _title = mkTitle exts
+      { _title = mkTitle exts -- TODO: `Expand positional record` without NamedFieldPuns if RecordInfoApp
```
jetjinser, line 159: The existing `codeActionProvider` only needs "the list of extensions, and a int to allow us to resolve it later", but here `mkTitle` needs to be decided based on the `RecordInfo` how to generate it; If the `RecordInfo` is obtained here, does it not conform to the idea of ​​resolve?

should I adjust the `RecordInfo` to make two `codeActionProvider`? Or is there any better way?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -224,6 +237,40 @@ inlayHintProvider _ state pId InlayHintParams {_textDocument = TextDocumentIdent
...
+       let (name, loc) = ((flSelector . unLoc) &&& (srcSpanToLocation . getLoc)) label
+           fieldDefLoc = srcSpanToLocation (nameSrcSpan name)
+       in do
+         (Location _ recRange) <- loc
+         pure InlayHint { _position = _start recRange
+                        , _label = InR $ pure (mkInlayHintLabelPart name fieldDefLoc)
+                        , _kind = Nothing -- neither a type nor a parameter
+                        , _textEdits = Just (maybeToList te) -- same as CodeAction
+                        , _tooltip = Just $ InL "Expand positional record" -- same as CodeAction
+                        , _paddingLeft = Nothing -- padding after dotdot
```
michaelpj, line 268: comment is not quite right?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -504,6 +567,24 @@ getRecCons e@(unLoc -> RecordCon _ _ flds)
...
+  in (recInfo, not (null recInfo))
+  where
+    mkRecInfo :: RecordAppExpr -> [RecordInfo]
+    mkRecInfo appExpr =
+      [ RecordInfoApp realSpan' appExpr | RealSrcSpan realSpan' _ <- [ getLoc expr ] ]
+
+    getFields :: HsExpr GhcTc -> [LHsExpr GhcTc] -> Maybe RecordAppExpr
+    getFields (HsApp _ (unLoc -> (XExpr (ConLikeTc (conLikeFieldLabels -> fls) _ _))) _) _
+      | null fls = Nothing
+    getFields (HsApp _ constr@(unLoc -> (XExpr (ConLikeTc (conLikeFieldLabels -> fls) _ _))) arg) args
```
michaelpj, line 582: I think you can just guard this case with `| not (null fls)` and then delete the case above, since the fallthrough case will handle it?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
michaelpj: I think we probably don't want to get the `RecordInfo` here if we can avoid it, that's the expensive bit, right? 

I would have thought that the code action doesn't apply it all if it's a `RecordInfoApp`. Isn't the code action only for record wildcards?

michaelpj (approved):
Looking good!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
jetjinser: Yes, I also think we should avoid getting `RecordInfo` before resolve.

Since `RecordInfoApp` is a constructor of `RecordInfo`, so it is also processed in the code action provider that process `RecordInfo`.

When the code action resolves, we can get the `RecordInfo`, so a useless extension `RecordWildCards` will not actually be inserted if `RecordInfo` is `RecordInfoApp`.
The only problem with the current code is that the `title` of the code action will display **Expand record wildcard _(needs extension: NamedFieldPuns)_** (when the extension does not exist in current file), no matter what it is `RecordInfo`.

michaelpj: I guess what I'm saying is that we shouldn't have any _code actions_ where the `RecordInfo` is `RecordInfoApp`, since `RecordInfoApp` can't be a wildcard, and that's what the code action applies to?

michaelpj:
Let's merge this!

PR #4448, opened 2024-11-11
Prepare release 2.10.0.0

fendor:
# Releasing

## Release checklist

- [x] check ghcup supports new GHC releases if any
- [x] check all plugins still work if release includes code changes
- [x] set the supported GHCs in workflow file `.github/generate-ci/gen_ci.hs`
- [x] regenerate the CI via `./.github/generate-ci/generate-jobs`
- [x] bump package versions in all `*.cabal` files (same version as hls)
  - HLS uses lockstep versioning. The core packages and all plugins use the same version number, and only support exactly this version.
    - Exceptions:
      - `hie-compat` requires no automatic version bump.
      - `shake-bench` is an internal testing tool, not exposed to the outside world. Thus, no version bump required for releases.
  - For updating cabal files, the following script can be used:
    - ```sh
      ./release/update_versions.sh <OLD_VERSION> <NEW_VERSION>
      ```
    - It still requires manual verification and review
- [x] generate and update changelog
  - Generate a ChangeLog via `./GenChangelogs.hs <api-key> <tag>`
    - `<tag>` is the git tag you want to generate the ChangeLog from.
    - `<api-key>` is a github access key: https://github.com/settings/tokens
- [x] update https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html#current-ghc-version-support-status
- [x] create release branch as `wip/<version>`
  - `git switch -c wip/<version>`
- [x] create release tag as `<version>`
  - `git tag <version>`
- [x] trigger release pipeline by pushing the tag
  - this creates a draft release
  - `git push <remote> <version>`
- [x] run `sh scripts/release/download-gh-artifacts.sh <version> <your-gpg-email>`
  - downloads artifacts to `gh-release-artifacts/haskell-language-server-<version>/`
  - also downloads FreeBSD bindist from circle CI
  - adds signatures
- [x] upload artifacts to downloads.haskell.org from `gh-release-artifacts/haskell-language-server-<version>/`
  - You require sftp access, contact wz1000, bgamari or chreekat
  - `cd gh-release-artifacts/haskell-language-server-<version>`
  - `SIGNING_KEY=... ../../release/upload.sh upload`
    - Your SIGNING_KEY can be obtained with `gpg --list-secret-keys --keyid-format=long`
  - Afterwards, the artifacts are available at: `https://downloads.haskell.org/~hls/haskell-language-server-<version>/`
  - Run `SIGNING_KEY=... ../../release/upload.sh purge_all` to remove CDN caches
[...]

PR #4452, opened 2024-11-18
Add support for splice plugin with GHC 9.10

georgefst:
I was hoping that we might get this plugin back for the upcoming release (https://github.com/haskell/haskell-language-server/pull/4448), but unfortunately it's not completely straightforward. The code compiles with no changes, but the test outputs in a few cases are slightly wrong, generally around leading and trailing whitespace and parentheses.

I spent a few hours looking in to this last weekend, but I'm over my head. I think someone with more experience with the GHC API and `ghc-exactprint` is going to have to take over. I'm putting my WIP stuff here just in case it's of use.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,8 +155,7 @@ jobs:
         name: Test hls-eval-plugin
         run: cabal test hls-eval-plugin-tests || cabal test hls-eval-plugin-tests
 
-      # TODO enable when it supports 9.10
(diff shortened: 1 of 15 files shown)


georgefst:
I've just seen @alanz on Matrix say:

> There are more changes to ghc-exactprint for 9.12.1, but then it should be stable going forward.

So maybe it's worth waiting one more version before spending too much time on this.

georgefst:
Taking a fresh stab at this in #4915.

PR #4463, opened 2024-12-09
Always handle "completion/resolve" requests

### LINKED ISSUE #4451: issues with CompletionItemResolve
## Your environment 

Which OS do you use? 
**NixOS**

**9.6.6, with nix**

**emacs+lsp-mode**
**2.9.0.0, using nix**
**No**

## What's wrong?

some error about SMethod_CompletionItemResolve keeps popping up as I type anything in the buffer
the error is not an emacs problem too, same behavior occurs on vscode (I just tested)
(see below)
### Debug information

```
Error processing message (error "No plugins are available to handle this SMethod_CompletionItemResolve request.
ghcide-completions does not handle resolve requests for (unable to determine resolve owner)).").
```

fendor:
Resolves the LSP "completion/resolve" requests, even if there is nothing more to resolve.
If there is nothing to resolve, we simply return the initial completion item.

If we reject "completion/resolve" requests, then some LSP clients show our rejection as pop-ups which are hugely distracting.

We change completion tests to always resolve completions, regardless of the existence of `_data_`, as this seems to be the behaviour of VSCode.

Closes #4451

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -127,7 +127,13 @@ dropListFromImportDecl iDecl = let
     in f <$> iDecl
 
 resolveCompletion :: ResolveFunction IdeState CompletionResolveData Method_CompletionItemResolve
-resolveCompletion _ide _pid comp _uri NothingToResolve = pure comp
+resolveCompletion _ide _pid comp _uri NothingToResolve =
+  -- See docs of 'NothingToResolve' for the reasoning of this
+  -- NO-OP handler.
+  --
+  -- Handle "completion/resolve" requests, even when we don't want to add
+  -- any additional information to the 'CompletionItem'.
+  pure comp
 resolveCompletion ide _pid comp@CompletionItem{_detail,_documentation,_data_} uri (CompletionResolveData _ needType (NameDetails mod occ)) =
   do
     file <- getNormalizedFilePathE uri
--- a/ghcide/src/Development/IDE/Plugin/Completions.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions.hs
@@ -127,6 +127,7 @@ dropListFromImportDecl iDecl = let
     in f <$> iDecl
 
 resolveCompletion :: ResolveFunction IdeState CompletionResolveData Method_CompletionItemResolve
+resolveCompletion _ide _pid comp _uri NothingToResolve = pure comp
 resolveCompletion ide _pid comp@CompletionItem{_detail,_documentation,_data_} uri (CompletionResolveData _ needType (NameDetails mod occ)) =
   do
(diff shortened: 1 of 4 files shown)


michaelpj:
Hmmm. So what's happening? We advertise support for resolve but don't actually resolve it? Or we advertise resolve because some plugins support resolve, but not all of them do? And so when we hit that case we get a rejection?

fendor:
We advertise "completion/resolve", and the completion plugin supports it, but for some plugin, there is simply no more info to resolve. But vscode still asks for "completion/resolve" which so far, we simply rejected if it was for a completion that didn't have any more info. E.g., documentation for local variables, such as in `where` blocks.

michaelpj:
Okay, I guess I'm worried that we have the same problem in other plugins for other reasons, but I guess we can tackle that later!

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -127,6 +127,7 @@ dropListFromImportDecl iDecl = let
     in f <$> iDecl
 
 resolveCompletion :: ResolveFunction IdeState CompletionResolveData Method_CompletionItemResolve
+resolveCompletion _ide _pid comp _uri NothingToResolve = pure comp
```
michaelpj, line 130: comment?

michaelpj (approved):

fendor:
We do, not with resolve in particular, but in other instances, e.g. semantic-token plugin rejects requests by default as well, since the plugin is disabled. Cabal plugin also rejects certain requests, such as `Goto Type Definition`, etc...

ghcide/src/Development/IDE/Plugin/Completions.hs
fendor: I thought the comment for `NothingToResolve`suffices, but will add the comment here as well.

soulomoon (approved):
LGTM

soulomoon:
> We do, not with resolve in particular, but in other instances, e.g. semantic-token plugin rejects requests by default as well, since the plugin is disabled. Cabal plugin also rejects certain requests, such as Goto Type Definition, etc...
> 
> I will open an issue tracking this in general.

Emmm, so we might need to respond with a dummy respond to these request as well ?

fendor:
> Emmm, so we might need to respond with a dummy respond to these request as well ?

Yeah, that sounds like a reasonable solution. Maybe not so pretty, but solves the issue as well.

fendor:
We might be better off solving this issue by adding logic to the completion/resolve dispatch logic. For example, we could specify that by convention if the resolve request *doesn't* have any `_data` JSON, we simply return the `CompletionItem`. 

Then the request wouldn't fail, and it would be handled at a central location.

fendor:
Closed in favour of the more principled PR #4478

PR #4472, opened 2024-12-23
Migrate `hls-class-plugin` to use structured diagnostics

fendor:
Split off from #4433, this PR moves the `hls-class-plugin` to use the structured diagnostics. 

It worked well in my tests, but unfortunately it is incompatible with GHC 9.4, because the error message the plugin cares about is not structured in 9.4.
We can either wait until we drop support for 9.4 or maintain both paths until we drop support for 9.4.

Unfortunately, the only way to push GHC into the direction we want, is to do the latter and report any lacking error messages to GHC ASAP.

FYI, only the last commit matters.

### DIFF
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
@@ -112,15 +112,15 @@ instance NFData InstanceBindLensResult where
 type instance RuleResult GetInstanceBindLens = InstanceBindLensResult
 
 data Log
-  = LogImplementedMethods Class [T.Text]
+  = LogImplementedMethods DynFlags Class ClassMinimalDef
   | LogShake Shake.Log
 
 instance Pretty Log where
   pretty = \case
-    LogImplementedMethods cls methods ->
-      pretty ("Detected implemented methods for class" :: String)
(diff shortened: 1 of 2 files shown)


michaelpj (approved):
Looks nice! It's a shame about not everything being structured. I wonder if we should just do an inventory of the places where we use GHC's diagnostics and check manually that they've been made structured? Maybe that's cheaper than doing all the PRs and either letting them get stale or adding a bunch of CPP?

plugins/hls-class-plugin/src/Ide/Plugin/Class/Types.hs
```diff
@@ -112,15 +112,15 @@ instance NFData InstanceBindLensResult where
...
   | LogShake Shake.Log
 
 instance Pretty Log where
   pretty = \case
     LogImplementedMethods cls methods ->
-      pretty ("Detected implemented methods for class" :: String)
+      pretty ("The following methods are missing" :: String)
         <+> pretty (show (getOccString cls) <> ":") -- 'show' is used here to add quotes around the class name
-        <+> pretty methods
+        <+> pretty (showSDocUnsafe $ ppr methods)
```
fendor: Iirc, there is a better way than `showSDocUnsafe`, though it probably doesn't really matter for logs.

Issue #4467, opened 2024-12-17
Can't get FixIt to add import


Aster89:
### Your environment

Which OS do you use?
ArchLinux

Which version of GHC do you use and how did you install it?
9.8.2, managed via GHCup

How is your project built (alternative: link to the project)?
Just created a project via `cabal init --simple`, answering `2) Executable` to the first and only question.

Which LSP client (editor/plugin) do you use?
Terminal Vim + YCM

Which version of HLS do you use and how did you install it?
Should be 2.9.0.1, if I understand correctly (see debug info).

Have you configured HLS in any way (especially: a `hie.yaml` file)?
Not that I'm aware of (and log says definitely I don't have a `hie.yaml` file), but please ask specific question if anything makes you think I have.

### Steps to reproduce
  1. Initialize a project with `cabal init --simple`,
  2. add the line `f = when`,
  3. ask for `FixIt` on `when`,
  4. choose 4, i.e. `import Control.Monad (when)` (any other will do I guess)

### Expected behaviour
`import Control.Monad (when)` should be added in the preamble of the file.

### Actual behaviour
Expected behavior does not happen, and this error is triggered instead:
```
ResponseFailedException: Request failed: -32601: No plugins are available to handle this SMethod_CodeActionResolve request.
 Plugins installed for this method, but not available to handle this request are:
explicit-fields does not handle resolve requests for (unable to determine resolve owner)).
hlint does not handle resolve requests for (unable to determine resolve owner)).
overloaded-record-dot does not handle resolve requests for (unable to determine resolve owner)).
importLens does not handle resolve requests for (unable to determine resolve owner)).
```

### Debug information
This is the log of the HLS:
```
No 'hie.yaml' found. Try to discover the project type!
Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 2.9.0.1 x86_64 ghc-9.10.1
Current directory: /home/enrico/deleteme
Operating system: linux
Arguments: ["--lsp"]
Cradle directory: /home/enrico/deleteme
Cradle type: Cabal

Tool versions found on the $PATH
cabal:          3.12.1.0
stack:          Not found
ghc:            9.8.2


Consulting the cradle to get project GHC version...
2024-12-17T16:50:16.704393Z | Debug | cabal exec -v0 -- ghc --print-libdir
[...]

PR #4478, opened 2025-01-06
Implement fallback handler for `*/resolve` requests

### LINKED ISSUE #4451: issues with CompletionItemResolve

fendor:
We had multiple reports, where `resolve` requests (such as `completion/resolve` and `codeAction/resolve`) are rejected by HLS since the `_data_` field of the respective LSP feature has not been populated by HLS.
This makes sense, as we only support `resolve` for certain kinds of `CodeAction`/`Completions`, when they contain particularly expensive properties, such as documentation or non-local type signatures.

So what to do? We can see two options:

1. Be dumb and permissive: if no plugin wants to resolve a request, then just respond positively with the original item! Potentially this masks real issues, but may not be too bad. If a plugin thinks it can handle the request but it then fails to resolve it, we should still return a failure.
2. Try and be smart: we try to figure out requests that we're "supposed" to resolve (e.g. those with a data field), and fail if no plugin wants to handle those. This is possible since we set data. So as long as we maintain the invariant that only things which need resolving get data, then it could be okay.

In 'fallbackResolveHandler', we implement the option (2).

Supersedes #4463 and presumably fixes #4451 and #4467

### DIFF
--- a/ghcide/test/exe/ResolveTests.hs
+++ b/ghcide/test/exe/ResolveTests.hs
@@ -1,0 +1,199 @@
+{-# LANGUAGE DeriveAnyClass     #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE OverloadedStrings  #-}
+module ResolveTests (tests) where
+
+import           Config
+import           Control.Lens
+import           Data.Aeson
+import qualified Data.Maybe                    as Maybe
+import           Data.Text                     (Text)
+import qualified Data.Text                     as T
+import           GHC.Generics
+import           Ide.Logger
+import           Ide.Types                     (PluginDescriptor (..), PluginId,
+                                                defaultPluginDescriptor,
+                                                mkPluginHandler,
+                                                mkResolveHandler)
+import qualified Language.LSP.Protocol.Lens    as J
+import qualified Language.LSP.Protocol.Lens    as JL
+import           Language.LSP.Protocol.Message (SomeMethod (..))
+import qualified Language.LSP.Protocol.Message as LSP
+import           Language.LSP.Protocol.Types
+import           Language.LSP.Test
+import           Test.Hls                      (IdeState, SMethod (..), liftIO,
+                                                mkPluginTestDescriptor,
+                                                someMethodToMethodString,
+                                                waitForAllProgressDone)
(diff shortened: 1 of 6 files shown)


fendor:
Added a test for all the resolvable requests that we currently support. Notably, our handlers do not support resolve requests for 
* `SMethod_WorkspaceSymbolResolve`
* `SMethod_DocumentLinkResolve`
* `SMethod_InlayHintResolve`

How they are handled is similarly hard-coded in https://github.com/haskell/haskell-language-server/blob/master/hls-plugin-api/src/Ide/Types.hs#L996

I don't want to add more features, so I am just adding tests for the supported request types.

ghcide/test/exe/ResolveTests.hs
```diff
@@ -0,0 +1,199 @@
...
+    { _range = mkRange 0 0 1 0
+    , _command = Just $ Command
+        { _title = lbl
+        , _command = lbl
+        , _arguments = Nothing
+        }
+    , _data_ = Nothing
+    }
+
+-- TODO: expose this from lsp-test
```
soulomoon, line 193: Unfinish todo?

soulomoon (commented):
Look good to me, just a minor problem

ghcide/test/exe/ResolveTests.hs
fendor: Finished todo, just needs an lsp-test release https://github.com/haskell/lsp/pull/603#pullrequestreview-2567905321

fendor:
@soulomoon Could you approve? I can't merge without an approval :)

soulomoon (approved):
Yes, let's merge it

PR #4480, opened 2025-01-13
Remove Unsafe Dynflags deadcode, they don't exist any more!

fendor:
I am currently reading through the ghcide code, and I noticed some supposed dead code, as `unsafeDynFlags` don't exist any more, [`setUnsafeDynFlags`](https://hackage.haskell.org/package/ghc-9.12.1/docs/src/GHC.Driver.Session.html#setUnsafeGlobalDynFlags) (and for [`9.4.8`](https://hackage.haskell.org/package/ghc-9.4.8/docs/src/GHC.Driver.Session.html#setUnsafeGlobalDynFlags)) just sets a couple of `Outputable` `IORef`s, which we likely don't want/need.

Thus, just delete and see whether CI is happy.

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -8,7 +8,7 @@ module Development.IDE.Session
   (SessionLoadingOptions(..)
   ,CacheDirs(..)
   ,loadSessionWithOptions
-  ,setInitialDynFlags
+  ,getInitialGhcLibDirDefault
   ,getHieDbLoc
   ,retryOnSqliteBusy
   ,retryOnException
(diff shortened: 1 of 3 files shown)


soulomoon (commented):
Looking good at CI but, it seems effect https://gitlab.haskell.org/ghc/ghc/-/blob/59202c800f2c97c16906120ab2561f6e1556e4af/compiler/GHC/Utils/Trace.hs#L42 ?
Do we have another place passing in the setting for `dynFlagsForPrinting`?

fendor:
All `SDoc`s we produce, should have a reference to `SDocContext` which are produced from the `DynFlags` respective to each component. GHC was changed in the past to move towards this `SDocContext` instead of `unsafeGlobalDynFlags` to avoid this global state. Thus, not relying on these globals is more correct than before, in my opinion, even though I think we are not even using them at all.
I don't think we need to call `dynFlagsForPrinting` ourselves, as we are not doing anything with the produced `DynFlags`. In `hiedb`, `dynFlagsForPrinting` essentially creates the default `DynFlags`, and then uses these to initialise the `SDocContext` when printing AST nodes. Since it is the default `DynFlags`, we are effectively initialising the global variables from the default `DynFlags`, which I think will be the same values for the global variables any way.

soulomoon (approved):
Thanks for the explaination,  make sense to me !

PR #4490, opened 2025-02-08
Nix tooling (minor changes)

dschrempf:
Hi! This PR contains a few minor changes to the Nix Flake.

b473c731 * Nix: Update lock files (GHC 9.6.5 -> 9.6.6)
a1b5c431 * Nix Flake: Small fixes, make comments a bit more consistent
39df6247 * Nix Flake: Remove `hlint`

I also removed `hlint` from the flake (last commit). However, this is an opinion up for debate:

We do not really need the `hlint` executable because it is integrated in HLS.

However, `hlint` is used by a GitHub Action, which we may want to replicate locally.

Also, the `fmt.sh` script is downloading an (unknown) `hlint` version which will
not work on NixOS anyways. In my opinion, if `fmt.sh` is still used (I can not
find any usages) we should fix the script so it is reproducible.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1696426674,
-        "narHash": "sha256-kvjfFW7WAETZlt09AgDn1MrtKzP7t90Vf7vypd3OL1U=",
+        "lastModified": 1733328505,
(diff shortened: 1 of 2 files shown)


fendor (approved):
LGTM, thank you for updating the nix setup!

fendor:
I am surprised `fmt.sh` isn't used at all, seemingly... Probably time to delete it.

PR #4494, opened 2025-02-11
Documentation: Imrpove "Contributing" (and amend Sphinx builders)

dschrempf:
Contains opinionated changes, please check!

I was compiling the documentation using  `nix build .#docs`. Two observations:
1. The Nix build command uses a different `sphinx-build` command than the one used in `docs/Makefile`.
2. The Nix build command actually failed because warnings were treated as errors.

So I fixed the warnings (duplicate IDs etc.) and thought, why not use the Makefile to build the documentation also with Nix?
However, that was not easily possible because the Makefile passed on all arguments (i.e., even arguments to the `make` command itself) to `sphinx-build`. (I think that is/was actually a bug.)

Anyways, I ended up changing the Makefile in ways that you may not agree with. In particular,
- I added the -n and -W flags to the `sphinx-build` invocation (warnings are errors).
- I removed the `help` target which seemed a bit unnecessary and overcomplicates things.
- I added a single `html` target to build the documentaiton using the HTML builder.

Have a look and let me know what you think!

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,6 +1,7 @@
 version: 2
 
 sphinx:
+  builder: "html"
   configuration: docs/conf.py
 
 build:
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,5 +1,4 @@
 # Minimal makefile for Sphinx documentation
-#
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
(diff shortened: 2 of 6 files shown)


fendor:
Thanks for the PR!

The docs are built for each PR:
image

I was able to look at the docs and verify the links are still working

fendor (approved):
LGTM!

michaelpj (approved):
LGTM!

PR #4498, opened 2025-02-16
Fig warning on startup when `"initializationOptions": null`

dschrempf:
See https://github.com/haskell/lsp/issues/604#issuecomment-2661389404


michaelpj:
I was actually thinking we'd fix this in `lsp`, since I suspect in reality nobody cares about `null` being a possibly-valid config object? What do you think?

PR #4510, opened 2025-03-03
Bugfix: Explicit record fields inlay hints for polymorphic records

### LINKED ISSUE #4475: Polymorphic record does not get inlay hint
### Your environment

Which OS do you use? Linux
Which version of GHC do you use and how did you install it? GHC 9.10.1
How is your project built (alternative: link to the project)? standalone

Which LSP client (editor/plugin) do you use? Neovim+lspconfig
Which version of HLS do you use and how did you install it? I picked latest commit, so https://github.com/haskell/haskell-language-server/commit/f09500b85f0f75acdbf15096ef0c04e0a134f43a because I was excited to test.
Have you configured HLS in any way (especially: a `hie.yaml` file)? standalone files

### Steps to reproduce

The following code:

```haskell
data Foo = Foo { bar :: Int, baz :: Char }

foo = Foo 1 'a'

data FooM m = FooM { barM :: Int, bazM :: Char }

fooM = FooM 1 'a'
```

### Expected behaviour

I would like inlay hint for the implicit fields of constructors `Foo` and `FooM` when defining `foo` and `fooM`.

More generally, it seems that inlay hints are not shown for record field when the record does have type parameters.

### Actual behaviour

I only have inlay hint for `Foo/foo` but not for `Foo/fooM`:

image

I also tried to add type signature and monomorphise `FooM`:

image

wczyz:
After type checking, expressions in the application of polymorphic records are wrapped in `WrapExpr`, which does not happen in the case of regular records

```haskell
HsApp _ _ (XExpr (WrapExpr (HsWrap _ (XExpr (ConLikeTc ...
```
vs
```haskell
HsApp _ _ (XExpr (ConLikeTc ...
```

Fixes #4475


guibou:
As the original author of the linked issues, https://github.com/haskell/haskell-language-server/issues/4475, I can confirm that I tested and it fixs the problem. Tested with ghc 9.10.

image

guibou:
Works also with `pattern`:

image

fendor (approved):

PR #4512, opened 2025-03-06
Enable bench for 9.10

soulomoon:
1. Fix the dependency problem preventing 9.10 from building bench 
2. Enable the bench to build even not for performance label, still need performance label if we need to run it.
3. bump up to actions/checkout-v4 in bench

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -52,8 +52,8 @@ jobs:
         # see discussion https://github.com/haskell/haskell-language-server/pull/4118
         # also possible to add more GHCs if we performs better in the future.
         ghc:
-          - '9.6'
           - '9.8'
+          - '9.10'
         os:
           - ubuntu-latest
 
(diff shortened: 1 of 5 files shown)


fendor:
@soulomoon Since hp2pretty seems to be unmaintained, I am wondering whether it would make sense to migrate our benchmarks to [eventlog2html](https://github.com/mpickering/eventlog2html). I am unfamiliar with the benchmarks, though, so maybe there is a good reason it is done like this. I suspect that the benchmark suite is rather old, and we could move some of the infrastructure to maintained alternatives.

.github/workflows/bench.yml
```diff
@@ -17,7 +17,6 @@ on:
 jobs:
   pre_job:
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'performance')
```
fendor, line 20: Needs to be readded, right?

fendor (approved):
LGTM, this is a great improvement!

.github/workflows/bench.yml
soulomoon: I move it down to the run session, so if performance label is provided, we build and run the bench, if not, we only build the bench.

PR #4528, opened 2025-03-25
Fix cabal check for Hackage release

fendor:
The current `haskell-language-server.cabal` fails when running `cabal check` for `cabal-install  3.12.1.0`. 
I don't know which `cabal` version is used on Hackage, but I know that this `cabal check` was an issue for the last Hackage release.

We need to add a check to the release CI, but it is quite annoying since `cabal check` doesn't accept `all` as a target.


fendor:
This PR is a prerequisite for the release.

Issue #4530, opened 2025-03-26
Add `words` package to ubuntu CI runs


fendor:
In #4527 we noticed the `FuzzySearch` tests get stuck easily. Turns out, there is a typo in the reference implementation causing it to endlessly loop. For all GHC versions. Why have we never noticed this before?
Well, turns out, the tests are only run if the path `/usr/share/dict/words` from the `words` package is installed.
Which it isn't in our CI.

Solution:
We should add this package to our CI, even though these tests add a full minute to our CI time.
Or we delete the test, one of the two 🙃

PR #4531, opened 2025-03-27
Simplify FuzzySearch test (avoid dependency on /usr/share/dict/words)

jhrcek:
Trying to address https://github.com/haskell/haskell-language-server/issues/4530
Saw this in Haskell Weekly newsletter :smile:


fendor:
@jhrcek Thanks for taking care of this! You should only have to add this in `test.yml`. The bindist and release.yaml jobs are only relevant for the release.

jhrcek:
Ok, words file is now present on ubuntu.
What remains now is to fix (stabilize?) those tests :smile: 
image

fendor:
Didn't expect that :hushed: Welp, so sounds like the reference impl is not the same as the FuzzySearch impl...

jhrcek:
I can reproduce that failure locally
`cabal test ghcide-tests --test-options='-p"/match works as expected on the english dictionary/"'`

Seems to fail quite often and every time there seems to be uppercase letters involved.

I don't know how it's supposed to work, but the main implementation seems to expect that the first parameter has lowercase words except for the first letter: https://github.com/haskell/haskell-language-server/blob/fcf8a7fb8a3b27ac62536749ee4bde2a7af55d07/ghcide/src/Text/Fuzzy/Parallel.hs#L32

but there are many words in the words file that have uppercase letters..

Bodigrim:
> Sometimes they assign different scores (reference results on the left, parallel on the right):

That's probably expected, because the "parallel" implementation cuts corners with regards to non-ASCII characters.

> Sometimes reference doesn't return any matches:

These violate a precondition that only the first character might be upper case.

jhrcek:
@Bodigrim you seem to have some context on this :smile_cat: Would you say it would make more sense to
1)  filter out words that don't satisfy the precondition and that contain non-ascii characters (or alternatively filter out the non-ascii chars / map non-first chars to lowercase)?
2) Invest time into making the reference impl. to be have more closely to the parallel impl?
3) Disable this flaky unit test, because it doesn't add too much value / adds too much time to CI?

Bodigrim:
I'd put a few unit tests to check that HLS fuzzy search can order a few alternatives in a reasonable way, but that's it. The reference implementation is apparently fragile as are additional system dependencies. Ditch them, there is bigger fish to fry.

fendor (approved):
LGTM, thanks!

PR #4532, opened 2025-04-01
reinstating ignore-plugins-ghc-bounds

GuillaumedeVolpiano:
Seen from Gentoo, the ignore-plugins-ghc-bounds is useful, this patch tries to make it do something again.

https://github.com/gentoo-haskell/gentoo-haskell/issues/1766

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -174,7 +174,7 @@ flag cabalgild
   manual:      True
 
 common cabalgild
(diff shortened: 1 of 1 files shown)


fendor (commented):
LGTM!

I couldn't 100% tell from the discussion, does the Gentoo project have a particular use-case for this flag?

fendor (approved):

Issue #4479, opened 2025-01-10
Add Package doesn't work with package-imports


fendor:
![image](https://github.com/user-attachments/assets/60039fa0-ca91-4fcf-b2b9-78ddcf1d049a)

Should offer the `Add dependency to cabal file` code action, but doesn't.
That's because the error message looks different. Fixing it should be a matter of improving the regex.

PR #4537, opened 2025-04-02
Support PackageImports in hiddenPackageSuggestion

jian-lin:
Fix: #4479

TODO: Probably I need to add more tests.

CC @VeryMilkyJoe @fendor

## Video demo

https://github.com/user-attachments/assets/692f441e-19be-46ac-8197-a74bfe6c481c

## More info

Here is the string being matched.

With `PackageImports`
```
not found: Could not find module ‘Data.List.Split’
Perhaps you meant
  Data.List.Split (needs flag -package-id split-0.2.5)
  Data.List.Split (needs flag -package-id split-0.2.5)
```

Without `PackageImports`
```
not found: Could not load module ‘Data.List.Split’
It is a member of the hidden package ‘split-0.2.5’.
Perhaps you need to add ‘split’ to the build-depends in your .cabal file.
It is a member of the hidden package ‘split-0.2.5’.
Perhaps you need to add ‘split’ to the build-depends in your .cabal file.
```

I also notice that the "not found" error remains even after the dependency is added by the code action (with and without `PackageImports`).  That may be a bug of my lsp client (`eglot` on Emacs) or HLS.  Needs further investigation.

### DIFF
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
@@ -190,6 +190,12 @@ addDependencySuggestCodeAction plId verTxtDocId suggestions haskellFilePath caba
 -- > It is a member of the hidden package ‘split-0.2.5’.
 -- > Perhaps you need to add ‘split’ to the build-depends in your .cabal file."
 --
+-- or this if PackageImports extension is used:
+--
+-- > "Could not find module ‘Data.List.Split’
+-- > Perhaps you meant
+-- >   Data.List.Split (needs flag -package-id split-0.2.5)"
+--
 -- It extracts mentioned package names and version numbers.
 -- In this example, it will be @[("split", "0.2.5")]@
 --
@@ -204,12 +210,17 @@ hiddenPackageSuggestion diag = getMatch (msg =~ regex)
     msg :: T.Text
     msg = _message diag
     regex :: T.Text -- TODO: Support multiple packages suggestion
-    regex = "It is a member of the hidden package [\8216']([a-zA-Z0-9-]*[a-zA-Z0-9])(-([0-9\\.]*))?[\8217']"
+    regex =
+        let regex' = "([a-zA-Z0-9-]*[a-zA-Z0-9])(-([0-9\\.]*))?"
+        in "It is a member of the hidden package [\8216']" <> regex' <> "[\8217']"
+           <> "|"
+           <> "needs flag -package-id " <> regex'
     -- Have to do this matching because `Regex.TDFA` doesn't(?) support
     -- not-capturing groups like (?:message)
     getMatch :: (T.Text, T.Text, T.Text, [T.Text]) -> [(T.Text, T.Text)]
     getMatch (_, _, _, []) = []
-    getMatch (_, _, _, [dependency, _, cleanVersion]) = [(dependency, cleanVersion)]
+    getMatch (_, _, _, [dependency, _, cleanVersion, "", "", ""]) = [(dependency, cleanVersion)]
+    getMatch (_, _, _, ["", "", "", dependency, _, cleanVersion]) = [(dependency, cleanVersion)]
     getMatch (_, _, _, _) = error "Impossible pattern matching case"
 
 command :: Recorder (WithPriority Log) -> CommandFunction IdeState CabalAddCommandParams
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
@@ -221,7 +221,7 @@ hiddenPackageSuggestion diag = getMatch (msg =~ regex)
     getMatch (_, _, _, []) = []
     getMatch (_, _, _, [dependency, _, cleanVersion, "", "", ""]) = [(dependency, cleanVersion)]
     getMatch (_, _, _, ["", "", "", dependency, _, cleanVersion]) = [(dependency, cleanVersion)]
-    getMatch (_, _, _, _) = error "Impossible pattern matching case"
+    getMatch (_, _, _, _) = []
 
 command :: Recorder (WithPriority Log) -> CommandFunction IdeState CabalAddCommandParams
 command recorder state _ params@(CabalAddCommandParams {cabalPath = path, verTxtDocId = verTxtDocId, buildTarget = target, dependency = dep, version = mbVer}) = do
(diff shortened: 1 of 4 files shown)


plugins/hls-cabal-plugin/test/testdata/cabal-add-testdata/cabal-add-tests/test/MainPackageImports.hs
jian-lin, line 1: Maybe it is better to find another hidden package and merge the test of `PackageImports` into [the old one](https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-cabal-plugin/test/testdata/cabal-add-testdata/cabal-add-tests/test/Main.hs).

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -204,12 +204,16 @@ hiddenPackageSuggestion diag = getMatch (msg =~ regex)
...
+        <> "|"
+        <> "needs flag -package-id ([a-zA-Z0-9-]*[a-zA-Z0-9])(-([0-9\\.]*))?"
     -- Have to do this matching because `Regex.TDFA` doesn't(?) support
     -- not-capturing groups like (?:message)
     getMatch :: (T.Text, T.Text, T.Text, [T.Text]) -> [(T.Text, T.Text)]
     getMatch (_, _, _, []) = []
-    getMatch (_, _, _, [dependency, _, cleanVersion]) = [(dependency, cleanVersion)]
+    getMatch (_, _, _, [dependency, _, cleanVersion, "", "", ""]) = [(dependency, cleanVersion)]
+    getMatch (_, _, _, ["", "", "", dependency, _, cleanVersion]) = [(dependency, cleanVersion)]
     getMatch (_, _, _, _) = error "Impossible pattern matching case"
```
fendor, line 217: I know you didn't implement this, but I would feel much better if this `error` wasn't there :) Could you fix this while you are here?

I think we can simply ignore this case, e.g. `getMatch (_, _, _, _) = []`.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
```diff
@@ -204,12 +204,16 @@ hiddenPackageSuggestion diag = getMatch (msg =~ regex)
     msg :: T.Text
     msg = _message diag
     regex :: T.Text -- TODO: Support multiple packages suggestion
-    regex = "It is a member of the hidden package [\8216']([a-zA-Z0-9-]*[a-zA-Z0-9])(-([0-9\\.]*))?[\8217']"
+    regex =
+        "It is a member of the hidden package [\8216']([a-zA-Z0-9-]*[a-zA-Z0-9])(-([0-9\\.]*))?[\8217']"
+        <> "|"
+        <> "needs flag -package-id ([a-zA-Z0-9-]*[a-zA-Z0-9])(-([0-9\\.]*))?"
```
fendor, line 210: Please also update the docs for this function!

plugins/hls-cabal-plugin/test/testdata/cabal-add-testdata/cabal-add-tests/test/MainPackageImports.hs
fendor: I think it is fine, but you need to actually run this test as well, right?

fendor (changes requested):

plugins/hls-cabal-plugin/test/testdata/cabal-add-testdata/cabal-add-tests/test/MainPackageImports.hs
jian-lin: Nice catch! Fixed.

plugins/hls-cabal-plugin/test/CabalAdd.hs
```diff
@@ -122,6 +124,23 @@ cabalAddTests =
...
+                                   , "(needs flag -package-id AWin32Console-1.19.1)"
+                                   ]
+                                   [ ("base","0.1.0.0")
+                                   , ("Blammo-wai", "0.11.0")
+                                   , ("BlastHTTP", "2.6.4.3")
+                                   , ("CC-delcont-ref-tf", "0.0.0.2")
+                                   , ("3d-graphics-examples", "1.1.6")
+                                   , ("AAI", "0.1")
+                                   , ("AWin32Console", "1.19.1")
+                                   ]
```
jian-lin, line 143: I did not add the "without version" test because I am not sure if the "without version" case exists when `PackageImports` is used.

plugins/hls-cabal-plugin/test/CabalAdd.hs
```diff
@@ -33,6 +33,8 @@ cabalAddTests =
         (generateAddDependencyTestSession "cabal-add-lib.cabal" ("src" </> "MyLib.hs") "split" [348])
     , runHaskellTestCaseSession "Code Actions - Can add hidden package to a test" ("cabal-add-testdata" </> "cabal-add-tests")
         (generateAddDependencyTestSession "cabal-add-tests.cabal" ("test" </> "Main.hs") "split" [478])
+    , runHaskellTestCaseSession "Code Actions - Can add hidden package to a test with PackageImports" ("cabal-add-testdata" </> "cabal-add-tests")
+        (generateAddDependencyTestSession "cabal-add-tests.cabal" ("test" </> "MainPackageImports.hs") "split" [731])
```
jian-lin, line 37: Is this single test case enough?  Should I mirror all similar tests?

jian-lin (commented):
I think I have addressed all previous comments.  Please take another look.

plugins/hls-cabal-plugin/test/CabalAdd.hs
fendor: Yeah, sounds reasonable!

fendor (approved):
LGTM, thank you for your contribution :heart:

plugins/hls-cabal-plugin/test/CabalAdd.hs
fendor: Sorry, I mean this single test case is enough.

jian-lin:
Thanks for your review!  I had a great experience in my first HLS PR.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd.hs
jian-lin, line 1: > I also notice that the "not found" error remains even after the dependency is added by the code action (with and without PackageImports). That may be a bug of my lsp client (eglot on Emacs) or HLS. Needs further investigation.

Oh, it is indeed a known HLS [bug][1].  With HLS 2.10.0.0 and vscode 2.5.3, I need to manually restart HLS or saving `hie.yaml` after adding dependencies to `.cabal` file to let HLS find the newly added dependencies.

In addition, there seems to be a bug in my preferred setting: Emacs with `eglot` as a LSP client.  In that setting, only manual restart works and saving `hie.yaml` has no effect.

[1]: https://github.com/haskell/haskell-language-server/issues/1068#issuecomment-2293383826

fendor: Yes, that sounds correct!

fendor:
@jian-lin It looks like you need to rebase your PR or merge changes from master into it before we can merge :)

jian-lin:
@fendor Rebased.

VeryMilkyJoe (approved):
LGTM, thank you!

Issue #4533, opened 2025-04-01
2.10.0.0: Bindist not working when built with GHC versions ≥ 9.10.1


benz0li:
### Your environment

Which OS do you use? Alpine Linux 3.21.3

Which version of GHC do you use and how did you install it? 9.10.1  
👉 Docker image `quay.io/benz0li/ghc-musl:9.10.1` ([https://github.com/benz0li/ghc-musl](https://github.com/benz0li/ghc-musl/blob/7621695049ae68c6d1271723d404ac82fc43c380/prior/9.10.1.Dockerfile))

Which version of HLS do you use and how did you install it? See below.

### Steps to reproduce

**On a host with docker installed**

```bash
docker run --rm -ti quay.io/benz0li/ghc-musl:9.10.1 bash
```

**Inside the container**

Build HLS bindist:

```shell
export HLS_VERSION=2.10.0.0 \
  && apk add --no-cache patchelf findutils \
  && cd /tmp \
  && curl -sSL "https://github.com/haskell/haskell-language-server/archive/refs/tags/$HLS_VERSION.tar.gz" \
    -o "haskell-language-server-$HLS_VERSION.tar.gz" \
  && tar -xzf "haskell-language-server-$HLS_VERSION.tar.gz" \
  && cd "haskell-language-server-$HLS_VERSION" \
  && . .github/scripts/env.sh \
  && . .github/scripts/common.sh \
  && sed -i.bak -e '/DELETE MARKER FOR CI/,/END DELETE/d' cabal.project \
  && GHCS="$GHC_VERSION" \
  && export GHCS \
  && ARTIFACT="$(uname -m)-linux-alpine" \
  && export ARTIFACT \
  && cabal update \
  && emake hls-ghc \
  && CABAL_VERSION_NUMERIC="$(cabal --numeric-version)" \
  && if dpkg --compare-versions "${CABAL_VERSION_NUMERIC%.*.*}" ge "3.11"; then \
    sed -i 's|"$(STORE_DIR)/ghc-$(GHC_VERSION)"|"$(STORE_DIR)/ghc-$(GHC_VERSION)-inplace"|g' \
      GNUmakefile; \
  fi \
  && emake bindist \
  && strip "out/bindist/$ARTIFACT/haskell-language-server-$HLS_VERSION/lib/$GHC_VERSION"/*.so \
  && emake bindist-tar
```

Install from bindist:

```shell
cd /var/tmp/
[...]

PR #4541, opened 2025-04-03
Revert "link executables dynamically to speed up linking (#4423)"

fendor:
This reverts commit 2df8775fa6062073904e96c48b456045511f05b5.

The PR #4423 seems to have introduced issues for release binaries. Thus, we are reverting the changes as advised in https://github.com/haskell/haskell-language-server/issues/4533#issuecomment-2774883109

We still want to benefit from the reduced test-times, so we might want to enable this option for our test-CI only.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -17,12 +17,6 @@ benchmarks: True
 
 write-ghc-environment-files: never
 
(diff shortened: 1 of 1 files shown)


develop7 (approved):
Agreed, this should be applied more granularly

PR #4542, opened 2025-04-03
add ghcide-bench flag to .cabal file

### LINKED ISSUE #4544: add Cabal flag for building ghcide-bench
I think most users don't need ghcide-bench built and installed by default.
So there should be a Cabal flag to disable it.

juhp:
Closes #4544

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -2172,9 +2172,14 @@ test-suite ghcide-tests
     RecordWildCards
     ViewPatterns
 
+flag ghcide-bench
+  description: Build the ghcide-bench executable
+  default:     True
 
 executable ghcide-bench
     import: defaults
+    if !flag(ghcide-bench)
+      buildable: False
     build-depends:
         aeson,
         bytestring,


fendor (approved):
LGTM. Which issue does this close again?

juhp:
I can create an issue if it is needed for relnotes etc: sorry if I didn't follow the proper workflow.

Also I wish this flag (as well as `test-exe`) would default to `False`: I didn't do this since I would rather not touch the CI, if you are using them in CI??  If not then they should definitely be off IMO.

juhp:
Ugh why didn't github rebase...

fendor:
No, it is fine, we don't follow a strict workflow, I thought we discussed this somewhere before... In #4359 actually!

Funnily enough, I started a similar PR in #4376 but kind of forgot about it. Could you check @soulomoon's comment from there?

soulomoon:
Yes, benchmark is in CI, if you wish to turn it off by default, you need to modified CI as well to explicitily enable it.

fendor:
@soulomoon Looks like it worked? Does this look good then?

soulomoon (approved):
Look good to me

PR #4543, opened 2025-04-03
porting hls-refactor to ghc-9.12

GuillaumedeVolpiano:
These changes allow the hls-refactor plugin to compile and pass the tests with ghc-9.12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -134,7 +134,7 @@ jobs:
           HLS_WRAPPER_TEST_EXE: hls-wrapper
         run: cabal test wrapper-test
 
-      - if: matrix.test && matrix.ghc != '9.12'
+      - if: matrix.test
         name: Test hls-refactor-plugin
         run: cabal test hls-refactor-plugin-tests || cabal test hls-refactor-plugin-tests
 
@@ -185,7 +185,7 @@ jobs:
         name: Test hls-call-hierarchy-plugin test suite
         run: cabal test hls-call-hierarchy-plugin-tests || cabal test hls-call-hierarchy-plugin-tests
 
-      - if: matrix.test && matrix.os != 'windows-latest' && matrix.ghc != '9.12'
+      - if: matrix.test && matrix.os != 'windows-latest'
         name: Test hls-rename-plugin test suite
         run: cabal test hls-rename-plugin-tests || cabal test hls-rename-plugin-tests
 
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -580,13 +580,13 @@ flag rename
   manual:      True
 
 common rename
-  if flag(rename) && (impl(ghc < 9.11) || flag(ignore-plugins-ghc-bounds))
+  if flag(rename)
     build-depends: haskell-language-server:hls-rename-plugin
     cpp-options: -Dhls_rename
 
 library hls-rename-plugin
   import:           defaults, pedantic, warnings
-  if !flag(rename) || (impl(ghc > 9.11) && !flag(ignore-plugins-ghc-bounds))
+  if !flag(rename)
     buildable: False
   exposed-modules:  Ide.Plugin.Rename
   hs-source-dirs:   plugins/hls-rename-plugin/src
@@ -610,7 +610,7 @@ library hls-rename-plugin
 
 test-suite hls-rename-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
-  if !flag(rename) || (impl(ghc > 9.11) && !flag(ignore-plugins-ghc-bounds))
+  if !flag(rename)
     buildable: False
   type:             exitcode-stdio-1.0
   hs-source-dirs:   plugins/hls-rename-plugin/test
@@ -1596,13 +1596,13 @@ flag refactor
   manual:      True
 
 common refactor
-  if flag(refactor) && (impl(ghc < 9.11) || flag(ignore-plugins-ghc-bounds))
+  if flag(refactor) 
     build-depends: haskell-language-server:hls-refactor-plugin
     cpp-options: -Dhls_refactor
 
 library hls-refactor-plugin
   import:           defaults, pedantic, warnings
-  if !flag(refactor) || (impl(ghc > 9.11) && !flag(ignore-plugins-ghc-bounds))
+  if !flag(refactor)
     buildable: False
   exposed-modules:  Development.IDE.GHC.ExactPrint
                     Development.IDE.GHC.Compat.ExactPrint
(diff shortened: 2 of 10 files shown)


haskell-language-server.cabal
```diff
@@ -1596,13 +1596,13 @@ flag refactor
...
 common refactor
-  if flag(refactor) && (impl(ghc < 9.11) || flag(ignore-plugins-ghc-bounds))
+  if flag(refactor) 
     build-depends: haskell-language-server:hls-refactor-plugin
     cpp-options: -Dhls_refactor
 
 library hls-refactor-plugin
   import:           defaults, pedantic, warnings
-  if !flag(refactor) || (impl(ghc > 9.11) && !flag(ignore-plugins-ghc-bounds))
+  if !flag(refactor) -- || (impl(ghc > 9.11) && !flag(ignore-plugins-ghc-bounds))
```
fendor, line 1605:
```suggestion
  if !flag(refactor)
```

fendor (approved):
LGTM, amazing work! 
I have honestly not reviewed in-depth. I will rely on the test-suite to catch any mistakes :)

haskell-language-server.cabal
GuillaumedeVolpiano: Of course!

GuillaumedeVolpiano: Fixed.

fendor:
The tests will not run on CI, you need to change this line: https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/test.yml#L137

Very difficult to find, apologies!

Also, just making sure, you can ignore the `pre-commit` error in CI.

GuillaumedeVolpiano:
Changed. Hadn't even thought to look there. Hopefully they will run smoothly.

GuillaumedeVolpiano:
Let's make sure this is fine, but I just wanted to note that as a consequence, rename and smile built smoothly on my end, they'll just need unmasking afterwards. Unless you want me to add it here.

fendor:
It is fine to enable more plugins immediately, spares us a separate CI run :)

You have to accept the changes to the golden files, this is just there so we are sure to catch the changes!

GuillaumedeVolpiano:
OK. Hold a bit then, because I see that splice is currently <ghc-9.10. I need to check it with ghc-9.10 too

GuillaumedeVolpiano:
Tests pass on my side. Let's see how the CI goes.

GuillaumedeVolpiano:
I might need help understanding these failing ghcide tests. They don't seem to relate to any of my changes…

fendor:
The windows CI failures are spurious and will be resolved over by simply rerunning CI. The 9.12. ubuntu failure is genuine, as you are enabling plugins, you need to accept the test changes.

GuillaumedeVolpiano:
So basically I need to patch the test file as indicated by the results of the test, do I understand this right? Sorry if it's a dumb question, it's my first time running into these.

fendor:
Yes, this should do the trick: https://github.com/haskell/haskell-language-server/blob/master/test/functional/ConfigSchema.hs#L20

Assuming, you run the test with `ghc-9.12.2`.

GuillaumedeVolpiano:
Now that was easy. Thanks for the pointer

GuillaumedeVolpiano:
Well, seems I actually failed a few tests.will get back to work on it

fendor:
Don't worry, we are very happy you are working on it :)

GuillaumedeVolpiano:
Thanks. Maybe worth adding a WIP tag to it for now?

fendor:
Like this?

GuillaumedeVolpiano:
Yes thanks!

GuillaumedeVolpiano:
Only two tests left. Of course, the first four were easy because they're actually just changes in the way ghc report things.

GuillaumedeVolpiano:
Five downs, one to go

GuillaumedeVolpiano:
And we should be good to go

.github/workflows/test.yml
```diff
@@ -134,7 +134,7 @@ jobs:
           HLS_WRAPPER_TEST_EXE: hls-wrapper
         run: cabal test wrapper-test
 
-      - if: matrix.test && matrix.ghc != '9.12'
+      - if: matrix.test
```
fendor, line 137: Similar changes are necessary for `hls-splice-plugin` and `hls-rename-plugin`. Once you do that, you will need to rerun the `schema/ghc912` test, since you are enabling additional plugins.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1937,7 +1984,10 @@ extractQualifiedModuleName x
 extractDoesNotExportModuleName :: T.Text -> Maybe T.Text
 extractDoesNotExportModuleName x
   | Just [m] <-
-#if MIN_VERSION_ghc(9,4,0)
+#if MIN_VERSION_ghc(9,12,0)
+    matchRegexUnifySpaces x "The module ‘([^’]*)’ does not export"
```
fendor, line 1988: This looks like the only difference is the `t|T`?
We could avoid this CPP by changing the regex to `(t|T)`... But that changes the capture groups, which is probably more annoying than the duplication.

However, we can avoid the CPP altogether by using the [`ghcVersion` variable](https://hackage.haskell.org/package/ghcide-2.10.0.0/docs/Development-IDE-GHC-Compat.html#v:ghcVersion) and using pattern matching e.g.
```haskell
  case ghcVersion of
    GHC912 -> matchRegexUnifySpaces x "The module ‘([^’]*)’ does not export"
      <|> matchRegexUnifySpaces x "nor ‘([^’]*)’ export"
    _ ->  matchRegexUnifySpaces x "the module ‘([^’]*)’ does not export"
      <|> matchRegexUnifySpaces x "nor ‘([^’]*)’ export"
```

I think I would prefer that, what do you think?
Note, the oldest GHC version we support is `9.4`, so the `#else` in this case is dead code

fendor (approved):
Amazing work, only two small comments, then we will merge!

Thank you for this great contribution :)

.github/workflows/test.yml
GuillaumedeVolpiano: Oh, that's why the tests didn't run on rename and splice. Stupid me.

fendor (approved):

GuillaumedeVolpiano:
I'm going to re-disable splice. I'll look at it and gadt in a second time.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
GuillaumedeVolpiano: Yes, it's cleaner.

GuillaumedeVolpiano:
@fendor : can you have a look at the failures? I think they are of the spurious kind, but if they're not, I'll also disable rename for now and add it to my todo list ;)

fendor:
Yeah, they look spurious. I will rerun the CI job once the others have finished running.

fendor:
Awesome! There will be likely another release in the next two weeks, our users will be happy to see the `refactor` plugin enabled for GHC 9.12.2 :)

GuillaumedeVolpiano:
Will do my best to add a few other plugins in the meantime :p

PR #4546, opened 2025-04-04
Fix typo of rename plugin config

jian-lin:
related: #2809


fendor (approved):
LGTM! 
We should really try to generate this page instead of manually updating.

PR #4551, opened 2025-04-08
Remove allow-newer for hiedb

jhrcek:
Using GHC 9.12 compatible version of hiedb released yesterday.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,14 +60,14 @@ version: 2
 jobs:
   stackage-lts22:
     docker:
-      - image: haskell:9.6.6-slim-buster
+      - image: haskell:9.6-slim-bullseye
     environment:
       - STACK_FILE: "stack-lts22.yaml"
     <<: *defaults
 
   stackage-nightly:
     docker:
-      - image: haskell:9.8.4-slim-buster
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -77,4 +77,4 @@ workflows:
   multiple-ghcs:
     jobs:
       - stackage-lts22
-      - stackage-nightly
+      - stackage-lts23
```
jhrcek, line 80: the current stackage nightly is now using ghc-9.10.1.

So I renamed the job to lts-23 to reflect reality, but now repo settings is expecting stackage-nightly job that doesn't exist.

To fix this we can:
- bump nightly resolver to use ghc 9.10 (more work - I would prefer not doing that in this PR)
- change repo settings not to expect stackage-nightly job as mandatory, but to expect stackage-lts23 (but I don't have permissions to fix repo settings)

michaelpj: I'll fix the repo settings

michaelpj (approved):

michaelpj:
I've just removed the `stackage-nightly` required check for now.

michaelpj:
Apparently mergify can't update it because it conflicts with master

PR #4552, opened 2025-04-08
Update plugin support table for GHC 9.12.2

fendor:
Demote `hls-refactor-plugin` to tier 2, as we regularly ship the initial GHC support without it.

### DIFF
--- a/docs/support/plugin-support.md
+++ b/docs/support/plugin-support.md
@@ -37,34 +37,34 @@ For example, a plugin to provide a formatter which has itself been abandoned has
 
 ## Current plugin support tiers
 
-| Plugin                              | Tier | Unsupported GHC versions |
-| ----------------------------------- | ---- | ------------------------ |
-| ghcide core plugins                 | 1    |                          |
-| `hls-call-hierarchy-plugin`         | 1    |                          |
-| `hls-code-range-plugin`             | 1    |                          |
-| `hls-explicit-imports-plugin`       | 1    |                          |
-| `hls-pragmas-plugin`                | 1    |                          |
-| `hls-refactor-plugin`               | 1    |                          |
-| `hls-alternate-number-plugin`       | 2    |                          |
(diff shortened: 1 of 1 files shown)


jhrcek (approved):
LGTM

maerten:
I noticed `hls-fourmolu-plugin` is listed as not supported by GHC `9.12.2`.

However in a GHC 9.12.2 project with stack, it is possible to run `stack install fourmolu` and run it from the commandline without any issue.


```
$ stack --version
Version 3.3.1, Git revision 62d1bdf099c8c30634208f76cb4444f9896e4336 aarch64 hpack-0.37.0
$ stack ghc -- --version
The Glorious Glasgow Haskell Compilation System, version 9.12.2

$ which fourmolu
$HOME/.local/bin/fourmolu
$ fourmolu --version
fourmolu 0.18.0.0
using ghc-lib-parser 9.12.2.20250320
```

The only change needed to be able to run `stack install fourmolu` was adding the following to stack.yaml:
```
extra-deps:
  - ghc-lib-parser-9.12.2.20250320
```


From VSCode (actually Windsurf), formatting with fourmolu via HLS indeed didn't work. 
```
2025-04-09T07:17:41.514403Z | Warning | No plugin handles this "textDocument/formatting" request.
[Error - 9:17:41 AM] Request textDocument/formatting failed.
  Message: No plugins are available to handle this SMethod_TextDocumentFormatting request.
cabal-fmt is not the formatting provider (fourmolu) you chose in your config.
  Code: -32601 
```

Onfortunately I using the external formatter in HLS doesn't work.

image

Just thought I would share, since at least fourmolu seems to be compatible with this GHC version :)

fendor:
@maerten That's good to know, thanks!
This table is supposed to show the status quo! Right now, `hls-fourmolu-plugin` isn't supported, so that's why the table reflects it.

PR #4553, opened 2025-04-08
Enable hls-cabal-gild-plugin for GHC 9.12.2

fendor:


haskell-language-server.cabal
```diff
@@ -117,7 +117,7 @@ flag cabalfmt
   manual:      True
 
 common cabalfmt
-  if flag(cabalfmt)
+  if flag(cabalfmt) || !flag(cabal)
```
fendor, line 120:
```suggestion
  if !flag(cabalfmt) || !flag(cabal)
```

PR #4555, opened 2025-04-09
enable hlint for ghc-9.12

GuillaumedeVolpiano:
@fendor : I know this fails a test, but could you check it out? It seems to not be doing what the test is checking that it doesn't do, and the added space is actually what I would expect to happen so before I dive for a bug that might actually be a feature, I'd rather have your advice.

P.S.: Not enabling for ghc-9.10 as apply-refact doesn't work with ghc-9.10.

P.P.S.: apparently I've done something wrong with the flags :( . Happy for some pointers.

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -703,14 +703,14 @@ flag hlint
   manual:      True
 
 common hlint
-  if flag(hlint) && (impl(ghc < 9.10) || flag(ignore-plugins-ghc-bounds))
+  if flag(hlint) && ((impl(ghc < 9.10) || impl(ghc > 9.11)) || flag(ignore-plugins-ghc-bounds))
     build-depends: haskell-language-server:hls-hlint-plugin
     cpp-options: -Dhls_hlint
 
 library hls-hlint-plugin
   import:           defaults, pedantic, warnings
   -- https://github.com/ndmitchell/hlint/pull/1594
-  if !(flag(hlint)) || (impl(ghc > 9.10) && !flag(ignore-plugins-ghc-bounds))
+  if !(flag(hlint)) || ((impl(ghc >= 9.10) && impl(ghc < 9.11)) && !flag(ignore-plugins-ghc-bounds))
     buildable: False
   exposed-modules:  Ide.Plugin.Hlint
   hs-source-dirs:   plugins/hls-hlint-plugin/src
@@ -722,7 +722,7 @@ library hls-hlint-plugin
     , filepath
     , ghcide                == 2.10.0.0
     , hashable
-    , hlint                 >= 3.5 && < 3.9
+    , hlint                 >= 3.5 && < 3.11
     , hls-plugin-api        == 2.10.0.0
     , lens
     , mtl
@@ -753,7 +753,7 @@ library hls-hlint-plugin
 
 test-suite hls-hlint-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
-  if (!flag(hlint)) || (impl(ghc > 9.10) && !flag(ignore-plugins-ghc-bounds))
+  if !flag(hlint) || ((impl(ghc >= 9.10) && impl(ghc < 9.11)) && !flag(ignore-plugins-ghc-bounds))
     buildable: False
   type:             exitcode-stdio-1.0
   hs-source-dirs:   plugins/hls-hlint-plugin/test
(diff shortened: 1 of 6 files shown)


haskell-language-server.cabal
```diff
@@ -703,14 +703,14 @@ flag hlint
...
-  if flag(hlint) && (impl(ghc < 9.10) || flag(ignore-plugins-ghc-bounds))
+  if flag(hlint) && ((impl(ghc < 9.10) || impl(ghc > 9.10)) || flag(ignore-plugins-ghc-bounds))
     build-depends: haskell-language-server:hls-hlint-plugin
     cpp-options: -Dhls_hlint
 
 library hls-hlint-plugin
   import:           defaults, pedantic, warnings
   -- https://github.com/ndmitchell/hlint/pull/1594
-  if !(flag(hlint)) || (impl(ghc > 9.10) && !flag(ignore-plugins-ghc-bounds))
+  if !(flag(hlint)) || ((impl(ghc < 9.10) || impl(ghc > 9.10)) && !flag(ignore-plugins-ghc-bounds))
```
fendor, line 713:
```suggestion
  if !(flag(hlint)) || ((impl(ghc >= 9.10) || impl(ghc < 9.11)) && !flag(ignore-plugins-ghc-bounds))
```

haskell-language-server.cabal
```diff
@@ -753,7 +753,7 @@ library hls-hlint-plugin
 
 test-suite hls-hlint-plugin-tests
   import:           defaults, pedantic, test-defaults, warnings
-  if (!flag(hlint)) || (impl(ghc > 9.10) && !flag(ignore-plugins-ghc-bounds))
+  if !flag(hlint) || (!(impl(ghc < 9.10) || impl(ghc >9.10)) && !flag(ignore-plugins-ghc-bounds))
```
fendor, line 756:
```suggestion
  if !(flag(hlint)) || ((impl(ghc >= 9.10) && impl(ghc < 9.11)) && !flag(ignore-plugins-ghc-bounds))
```

haskell-language-server.cabal
```diff
@@ -703,14 +703,14 @@ flag hlint
   manual:      True
 
 common hlint
-  if flag(hlint) && (impl(ghc < 9.10) || flag(ignore-plugins-ghc-bounds))
+  if flag(hlint) && ((impl(ghc < 9.10) || impl(ghc > 9.10)) || flag(ignore-plugins-ghc-bounds))
```
fendor, line 706:
```suggestion
  if flag(hlint) && ((impl(ghc < 9.10) || impl(ghc > 9.11)) || flag(ignore-plugins-ghc-bounds))
```

GuillaumedeVolpiano:
> Too many negations 🙃

The strange case of the misplaced parenthesis. Thank you, that did the trick.

GuillaumedeVolpiano:
@fendor Now I can ask for your opinion on the failing test.

fendor:
Ill have a look, but it is not my area of expertise. Maybe @jhrcek has some pointers, too?

michaelpj:
Looking at the test failure, it looks like the newer `apply-refact` produces very slightly different (but okay) output. Sadly that means we probably need to have the test output be conditional on the GHC version, as we do in various places.

michaelpj:
Why does `apply-refact` not work with 9.10? I'm a bit confused by that if it works for 9.12?

GuillaumedeVolpiano:
> Why does `apply-refact` not work with 9.10? I'm a bit confused by that if it works for 9.12?

First, thanks for the invite. It's an honour.

For apply-refact, the owner (and @jhrcek ) have had difficulties with ghc-exactprint.

Looking into it is on my todo list, but if jhrcek is stumped, I guess I'm not going to do any better ;)

GuillaumedeVolpiano:
Anyhow, we should be good now.

fendor:
Could you update the plugin support table? In particular https://github.com/haskell/haskell-language-server/blob/master/docs/support/plugin-support.md?plain=1#L58

I know `hls-refactor-plugin` already supports GHC 9.12.2, I just forgot to update the table myself :grimacing:

fendor:
To master? No, I don't think you have the permissions to do that (no one should have these permissions, I believe), I meant in this PR

GuillaumedeVolpiano:
> To master? No, I don't think you have the permissions to do that (no one should have these permissions, I believe), I meant in this PR

Oh right. I thought that @michaelpj 's invitation gave me write access (which seemed a lot, to be honest). Anyhow, I made a separate PR and you merged it,  so we're good.

Sadly, I think some tests need to be rerun, there was a pipe problem.

michaelpj:
You've got write access (you've contributed a lot already! :) ), but even people with write access can't commit directly to master, we require everything to go via PRs. You can merge things though!

fendor (approved):
Awesome, thank you!

PR #4559, opened 2025-04-14
Enable stylish-haskell for ghc-9.10 and ghc-9.12

GuillaumedeVolpiano:
With the release of stylish-haskell-0.15.0.0, it now builds with ghc-lib-parser-9.10, so we can enable the plugin for ghc-9.10.
Note that a ghc-lib-parser-9.12 version should be released soon.

### DIFF
--- a/plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs
+++ b/plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs
@@ -79,10 +79,15 @@ provider recorder ide _token typ contents fp _opts = do
 -- If no such file has been found, return default config.
 loadConfigFrom :: FilePath -> IO Config
 loadConfigFrom file = do
+#if MIN_VERSION_stylish_haskell(0,15,0)
+  let configSearchStrategy = SearchFromDirectory (takeDirectory file)
+  config <- loadConfig (makeVerbose False) configSearchStrategy
+#else
   currDir <- getCurrentDirectory
   setCurrentDirectory (takeDirectory file)
   config <- loadConfig (makeVerbose False) Nothing
   setCurrentDirectory currDir
+#endif
   pure config
 
 -- | Run stylish-haskell on the given text with the given configuration.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,8 +164,7 @@ jobs:
         name: Test hls-stan-plugin
         run: cabal test hls-stan-plugin-tests || cabal test hls-stan-plugin-tests
 
-      # TODO enable when it supports 9.10
-      - if: matrix.test && matrix.ghc != '9.10' && matrix.ghc != '9.12'
+      - if: matrix.test
         name: Test hls-stylish-haskell-plugin
         run: cabal test hls-stylish-haskell-plugin-tests || cabal test hls-stylish-haskell-plugin-tests
 
--- a/cabal.project
+++ b/cabal.project
@@ -8,7 +8,7 @@ packages:
          ./hls-test-utils
 
 
-index-state:  2025-04-08T01:30:37Z
+index-state: 2025-04-19T07:34:07Z
 
 tests: True
 test-show-details: direct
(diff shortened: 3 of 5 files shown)


GuillaumedeVolpiano:
@fendor : I guess the tests will need to be rerun later, cabal is failing to find stylish-haskell-0.15.0.0, which was added to hackage yesterday…

plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs
```diff
@@ -81,7 +81,11 @@ loadConfigFrom :: FilePath -> IO Config
 loadConfigFrom file = do
   currDir <- getCurrentDirectory
   setCurrentDirectory (takeDirectory file)
+#if MIN_VERSION_stylish_haskell(0,15,0)
+  config <- loadConfig (makeVerbose False) SearchFromCurrentDirectory
```
jhrcek, line 85: Thanks for taking care of this update.

I think here you should reshape the code to look like in [this PR](https://github.com/haskell/haskell-language-server/pull/4338/files) (which was the reason why I did the api change in stylish haskell - so we could stop calling `setCurrentDirectory` just search for config)

fendor:
@GuillaumedeVolpiano You have to update the `index-state` in the `cabal.project` file. You can find your own `index-state` by running `cabal update`. You can just set that.
The CI will never find the newer version until you update the `index-state` in `cabal.project`, since we are intentionally freezing our dependencies to avoid build failures.

plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs
GuillaumedeVolpiano: Of course, and sorry, I hadn't seen your draft.

jhrcek: Don't be sorry, it's quite non-obvious thing that we should be doing something like that :smile_cat:

jhrcek: Oh, and also move the surrounding get/setCurrentDirectory to the else branch, please, so that there's no CWD manipulation with stylish >=0.15

GuillaumedeVolpiano:
@jhrcek : stylish-haskell now needs Cabal-3.14, but this conflicts with a lot of things here. I can get stylish-haskell to compile nicely with Cabal-3.12 with the following changes to lib/Language/Haskell/Stylish/Config/Cabal.hs
`+#if MIN_VERSION_Cabal(3,14,0)
     let projectRoot = Just $ Cabal.makeSymbolicPath p
     potentialCabalFile <- Cabal.findPackageDesc projectRoot
     case potentialCabalFile of
       Right cabalFile -> pure $ Just $
         Cabal.interpretSymbolicPath projectRoot cabalFile
       _ -> go (p : searched) ps
+#else
+    let projectRoot = p
+    potentialCabalFile <- Cabal.findPackageDesc projectRoot
+    case potentialCabalFile of
+      Right cabalFile -> pure $ Just cabalFile
+      _ -> go (p : searched) ps
+#endif
`
But I guess this kindof defeats your purpose. If not, I'll submit a PR there, hopefully we get a 0.15.0.1 version and everything will be simple then.

@fendor : I guess we need a Draft tag here for now.

plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs
GuillaumedeVolpiano: Done

jhrcek:
Oh, that's not good. I guess that Jasper proactively bumped to the latest version of cabal syntax in https://github.com/haskell/stylish-haskell/pull/489 before merging my 9.10 support PR.

I wonder if we could broaden the accepted version range of stylish-haskell Cabal dependency (not really sure how Cabal stuff is used in stylish-haskell)?

GuillaumedeVolpiano:
@fendor @michaelpj @jhrcek : this is now ready for review

fendor (approved):
Lgtm! When can we drop support for stylish-haskell 0.14? Like, which ghc version is only compatible with the older stylish-haskell version?

GuillaumedeVolpiano:
9.4

jhrcek (approved):
Great work, including pushing the work on upstream dependencies :+1:

PR #4567, opened 2025-04-21
Cleanup CPPs, begin deprecation of GHC 9.4

### LINKED ISSUE #4529: Cleanup CPP in `GHC/Orphans.hs`
The CPP in `GHC/Orphans.hs` needs to be cleaned up, as it doesn't follow our best practices when it comes to CPP:

* A CPP `#if` should span the minimal range possible
* No nested CPP statements

Especially after #4527 is merged and support for GHC 9.4 series is dropped, we should revisit that module.

GHC 9.4 will likely be dropped after the next release.

dschrempf:
Closes #4529.

Also,
- update flake lock
- some girl scout changes while trying to understand some code


fendor:
Fyi, in the biweekly HLS meeting, we have discussed to *not* drop support for `9.4` until the next release 2.11.0.0 which will happen once GHC 9.10.2 is released.

dschrempf:
Thanks for the information! I leave this PR in draft mode for now, but feel free to close it, and I can reopen in due time!

michaelpj:
btw you might find this useful: https://www.michaelpj.com/blog/2023/11/20/cpp-partial-evaluation.html

dschrempf:
@fendor I have removed `#if MIN_VERSION_ghc(9,4,0)` statements, but there are a lot of `#if MIN_VERSION_ghc(9,5,0)` statement left. Also, this is a bit of a nightmare because `stylish-haskell` fails to parse some of the amended files having many CPP statements. I had to rearrange two already.

Also, the Nix CI fails for MacOS for a reasons I don't understand. I have to investigate. I do not have more time now, maybe I get to it tomorrow. If you take over, please let me know!

Cheers!

EDIT: Ad Nix failure for MacOS. I am not sure if we should test this in CI.

dschrempf:
I tackled all `MIN_VERSION_ghc(9,5,...)` macros, as well as GLASGOW_HASKELL macros related to GHC 9.4. I also removed GHC 9.4 from the tests.

There is one question left in the tests. At a point we expect "ghc94.expected" (I didn't undertand this, maybe it was wrong in the first place?). I will highlight the code change with a comment.

Also there is a lot more macros a'la `MIN_VERSION_ghc(9,6,...)` which we also can remove by now, but I don't have any energy left ^^. This PR is HUGE already.

Depending on how the CI will go: I think we should merge this soon, conflicts will be really ugly.

plugins/hls-eval-plugin/test/Main.hs
```diff
@@ -84,8 +84,8 @@ tests =
       evalInFile "T8.hs" "-- >>> 3 `div` 0" "-- divide by zero" -- The default for marking exceptions is False
   , goldenWithEval "Applies file LANGUAGE extensions" "T9" "hs"
   , goldenWithEval "Evaluate a type with :kind!" "T10" "hs"
-  , goldenWithEval' "Reports an error for an incorrect type with :kind!" "T11" "hs"
-        (if ghcVersion >= GHC94 then "ghc94.expected" else "expected")
+  -- TODO: I do not understand why this is 'ghc94.expected'.
```
dschrempf: Why `ghc94.expected`?

.github/workflows/nix.yml
```diff
@@ -44,7 +44,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        # TODO: Fix compilation problems on macOS.
```
dschrempf, line 47: NOTE: I remove `macOS` from the Nix CI.

dschrempf:
I just realized this is PR 4567.

dschrempf:
Some notes:

- I think `stylish-haskell` and CPPs do not play well togehter. `stylish-haskell` seems to ignore conditional CPPs entirely and fails in many cases. How did you deal with this in the past? I had to copy some code into both branches of the conditionals for `stylish-haskell` to parse the files.

- It is difficult to know which warnings lead to CI errors, and which warnings don't. Do you think it is possible to use the same warning/error configuration for local builds and the CI?

- In general, I think somebody has to clean build HLS and fix warnings. I think some warnings are induced by changes between GHC and base libraries, but even those should be fixed with CPP statements, shouldn't we?

plugins/hls-eval-plugin/test/Main.hs
dschrempf: @fendor pulling you in on this.

There are still some independent CI errors, but I think I fixed most of them. I will not have to time to work on this before tomorrow.

fendor (approved):
Amazing work, I didn't anticipate this would be so much work! Thank you for taking care of this, it is much appreciated!

fendor:
> I think stylish-haskell and CPPs do not play well togehter.

formatters don't work with CPP in general, we usually just ignore these kinds of errors in the pre-commit job, it is not blocking :speak_no_evil: Sorry for not telling you sooner, but I think you improved the overall code base considerably!

> It is difficult to know which warnings lead to CI errors

Good point, I think ci runs with `-fpedantic`, but maybe we should borrow some lingo from javscript and have a flag such as `-fci` and document that in the user docs?

> In general, I think somebody has to clean build HLS and fix warnings. 

Making ghcide compile without warnings is a big amount of work with little payoff... Not sure anyone wants to tackle that.

plugins/hls-eval-plugin/test/Main.hs
fendor: I presume this is a result of code churn, initially the test was only failing for GHC94, then we added support for GHC96... Would be fine to just use `goldenWithEval`.

fendor: I changed it to use `goldenWithEval`

PR #4568, opened 2025-04-22
Enable hls-plugin-gadt for ghc-9.12

GuillaumedeVolpiano:
Unless I'm mistaken, that's the last tier 2 plugin for ghc-9.12

### DIFF
--- a/plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
+++ b/plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
@@ -7,16 +7,20 @@
 {-# OPTIONS_GHC -Wno-overlapping-patterns #-}
 module Ide.Plugin.GHC where
 
+#if !MIN_VERSION_ghc(9,11,0)
 import           Data.Functor                            ((<&>))
+#endif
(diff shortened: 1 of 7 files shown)


plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
```diff
@@ -229,21 +248,38 @@ prettyGADTDecl df decl =
...
+#endif
         -- Remove the first extra line if exist
         removeExtraEmptyLine s = case stripInfix "\n\n" s of
             Just (x, xs) -> x <> "\n" <> xs
             Nothing      -> s
 
+#if MIN_VERSION_ghc(9,11,0)
+
+#endif
+
```
fendor, line 282:
```suggestion
```

Right?

fendor (approved):
LGTM, great work as always!

plugins/hls-gadt-plugin/src/Ide/Plugin/GHC.hs
GuillaumedeVolpiano: Right. There was also a lost indentation a little higher, which I've also corrected.

PR #3655, opened 2023-06-12
Rewrite plugin tutorial

cgeorgii:
This is an almost-complete rewrite of the outdated plugin tutorial. The section about the code lens provider is still unchanged and probably incorrect. I'll try to get to it at some point.

To view the new docs locally follow the instructions from the [docs](https://haskell-language-server.readthedocs.io/en/latest/contributing/contributing.html#building-the-docs). TLDR: `cd docs; make html`

It also contains some minor refactors to the files mentioned in the plugin since it aided in my understanding of what was going on.

PR #4570, opened 2025-04-23
Plugin tutorial, more changes

dschrempf:
Some changes are mine, but many have been cherry picked and amended from PR #3655 by Christian Georgii <cgeorgii@gmail.com>. @cgeorgii 

This PR now includes most changes of #3655. I did not include the last sections which are behind a "what shall we do with those" comment in the aforementioned PR. I am not sure how to deal with those.

Another issue is that the link to the gist is outdated (and still links to the original one by Pepe Iborra). So I think we either have to remove the link or write a new gist and replace the link. What are your opinions?

I did not implement the literate file stuff, it seems too complicated to achieve in one PR.

Supersedes #3655.

### DIFF
--- a/docs/contributing/plugin-tutorial.md
+++ b/docs/contributing/plugin-tutorial.md
@@ -1,4 +1,5 @@
 # Let’s write a Haskell Language Server plugin
+
 Originally written by Pepe Iborra, maintained by the Haskell community.
 
 Haskell Language Server (HLS) is a Language Server Protocol (LSP) server for the Haskell programming language. It builds on several previous efforts to create a Haskell IDE.
@@ -22,6 +23,49 @@ While writing them, I didn't have to worry about performance, UI, or distributio
 
 The plugins also make these tools much more accessible to all users of HLS.
 
+## Preamble
+
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+
+import Ide.Types
+import Ide.Logger
+import Ide.Plugin.Error
+
+import Development.IDE.Core.RuleTypes
+import Development.IDE.Core.Service hiding (Log)
+import Development.IDE.Core.Shake hiding (Log)
+import Development.IDE.GHC.Compat
+import Development.IDE.GHC.Compat.Core
+import Development.IDE.GHC.Error
+import Development.IDE.Types.HscEnvEq
+import Development.IDE.Core.PluginUtils
+
+import qualified Language.LSP.Server as LSP
+import Language.LSP.Protocol.Types as JL
+import Language.LSP.Protocol.Message
+
+import Data.Aeson as Aeson
+import Data.Map (Map)
+import Data.IORef
+import Data.Maybe (fromMaybe, catMaybes)
+import qualified Data.Map as Map
+import qualified Data.HashMap.Strict as HashMap
+import qualified Data.Text as T
+import Control.Monad (forM)
+import Control.Monad.IO.Class (liftIO)
+import Control.Monad.Trans.Class
+import GHC.Generics (Generic)
+```
+
 ## Plugins in the HLS codebase
 
 The HLS codebase includes several plugins (found in `./plugins`). For example:
@@ -37,7 +81,9 @@ I recommend looking at the existing plugins for inspiration and reference. A few
 - Folders containing the plugin follow the `hls-pluginname-plugin` naming convention
 - Plugins are "linked" in `src/HlsPlugins.hs#idePlugins`. New plugin descriptors
   must be added there.
-  ```haskell -- src/HlsPlugins.hs
+
+  ```haskell ignore
+  -- Defined in src/HlsPlugins.**hs**
 
   idePlugins = pluginDescToIdePlugins allPlugins
     where
@@ -53,6 +99,7 @@ I recommend looking at the existing plugins for inspiration and reference. A few
         , NewPlugin.descriptor "new-plugin" -- Add new plugins here.
         ]
   ```
+
 To add a new plugin, extend the list of `allPlugins` and rebuild.
 
 ## The goal of the plugin we will write
@@ -80,7 +127,7 @@ Once the build is done, you can find the location of the HLS binary with `cabal 
 This way you can simply test your changes by reloading your editor after rebuilding the binary.
 
 > **Note:** In VSCode, edit the "Haskell Server Executable Path" setting.
-
+>
 > **Note:** In Emacs, edit the `lsp-haskell-server-path` variable.
 
 ![Settings](settings-vscode.png)
@@ -90,6 +137,7 @@ This way you can simply test your changes by reloading your editor after rebuild
 ## Digression about the Language Server Protocol
 
 There are two main types of communication in the Language Server Protocol:
+
 - A **request-response interaction** type where one party sends a message that requires a response from the other party.
 - A **notification** is a one-way interaction where one party sends a message without expecting any response.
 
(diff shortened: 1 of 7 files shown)


docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,153 @@ Here is a visual statement of what we want to accomplish:
...
-                    -> IdeState
-                    -> RenameParams
-                    -> IO (Either ResponseError WorkspaceEdit)
-```
+To achieve our plugin goals, we need to define:
+- a command handler (`importLensCommand`),
+- a code lens request handler (`lensProvider`).
 
-Providers are functions that receive some inputs and produce an IO computation that returns either an error or some result.
+These will be assembled together in the `descriptor` function of the plugin, which contains all the information wrapped in the `PluginDescriptor` datatype mentioned above.
```
VeryMilkyJoe, line 116:
```suggestion
These will be assembled in the `descriptor` function of the plugin, which contains all the information wrapped in the `PluginDescriptor` datatype mentioned above.
```

docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,153 @@ Here is a visual statement of what we want to accomplish:
...
-```
+We'll start with the command, since it's the simplest of the two.
 
-There are three flavours of `use` combinators:
+### The command handler
 
-1. `use*` combinators block and propagate errors,
-2. `useWithStale*` combinators block and switch to stale data in case of an error,
-3. `useWithStaleFast*` combinators return immediately with stale data if any, or block otherwise.
+In short, commands work like this:
```
VeryMilkyJoe, line 138:
```suggestion
In short, LSP commands work like this:
```

docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,153 @@ Here is a visual statement of what we want to accomplish:
...
+- The LSP server (HLS) initially sends a command descriptor to the client, in this case as part of a code lens.
+- Whenever the client decides to execute the command on behalf of a user (in this case a click on the code lens), it sends this same descriptor back to the LSP server. The server then handles and executes the command; this latter part is implemented by the `commandFunc` field of our `PluginCommand` value.
```
VeryMilkyJoe, line 140:
```suggestion
- When the user clicks on the code lens, the client asks HLS to execute the command with the given descriptor. The server then handles and executes the command; this latter part is implemented by the `commandFunc` field of our `PluginCommand` value.
```

docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,153 @@ Here is a visual statement of what we want to accomplish:
...
-    return (Right Null, Just (WorkspaceApplyEdit, ApplyWorkspaceEditParams edit))
-
+runImportCommand :: CommandFunction IdeState ImportCommandParams
+runImportCommand _ (ImportCommandParams edit) = do
+  -- This command simply triggers a workspace edit!
+  _ <- sendRequest SWorkspaceApplyEdit (ApplyWorkspaceEditParams Nothing edit) (\_ -> pure ())
+  return (Right Null)
 ```
 
+`runImportCommand` [sends a request](https://hackage.haskell.org/package/lsp/docs/Language-LSP-Server.html#v:sendRequest) to the client using the method `SWorkspaceApplyEdit` and the parameters `ApplyWorkspaceEditParams Nothing edit`, providing a response handler that does nothing. It then returns `Right Null`, which is an empty `Aeson.Value` wrapped in `Right`.
```
VeryMilkyJoe, line 202: Maybe explain why we have an `Either` type here, and that `Left` is used for failures?

docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,153 @@ Here is a visual statement of what we want to accomplish:
...
->  3. Ask GHC to produce the minimal import lists for this AST,
->  4. For every import statement without an explicit import list, find out the minimal import list, and produce a code lens to display it together with a command to graft it on.
+>  1. Request the type checking artifacts.
+>  2. Extract the actual import lists from the type-checked AST.
+>  3. Ask GHC to produce the minimal import lists for this AST.
+>  4. For each import statement lacking an explicit list, determine its minimal import list and generate a code lens displaying this list along with a command to insert it.
 
-The provider takes the usual `LspFuncs` and `IdeState` argument, as well as a `CodeLensParams` value containing the URI
-for a file, and returns an IO action producing either an error or a list of code lenses for that file.
+The provider takes the usual `LspFuncs` and `IdeState` arguments, as well as a `CodeLensParams` value containing a file URI. It returns an IO action that produces either an error or a list of code lenses for that file.
```
VeryMilkyJoe, line 213: `LspFuncs` does not exist anymore, they're now part of the plugin monad.

docs/contributing/plugin-tutorial.md
```diff
@@ -381,14 +315,16 @@ generateLens pId uri minImports (L src imp)
...
-- `./cabal*.project` and `./stack*.yaml`: add the plugin package in the `packages` field,
-- `./haskell-language-server.cabal`: add a conditional block with the plugin package dependency,
-- `./.github/workflows/test.yml`: add a block to run the test suite of the plugin,
-- `./.github/workflows/hackage.yml`: add the plugin to the component list to release the plugin package to Hackage,
-- `./*.nix`: add the plugin to Nix builds.
 
-The full code as used in this tutorial, including imports, can be found in [this Gist](https://gist.github.com/pepeiborra/49b872b2e9ad112f61a3220cdb7db967) as well as in this [branch](https://github.com/pepeiborra/ide/blob/imports-lens/src/Ide/Plugin/ImportLens.hs)
+Integrating the plugin into HLS itself requires changes to several configuration files.
+
+A good approach is to use the ID of an existing plugin (e.g., `hls-class-plugin`) as a guide:
```
VeryMilkyJoe, line 321:
```suggestion
A good approach is looking for the ID (f.e. `hls-class-plugin`) of an existing plugin as a reference:
```

VeryMilkyJoe (changes requested):
Thanks for picking up this PR! 
I have some comments and suggestions.

dschrempf:
Thanks for your detailed review @VeryMilkyJoe , I have this on my radar, just didn't get to it yet!

docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,153 @@ Here is a visual statement of what we want to accomplish:
...
+A plugin has a unique ID, command handlers, request handlers, notification handlers and rules:
 
-- The `ghcide` plugin, which embeds `ghcide` as a plugin (`ghcide` is also the engine under HLS),
-- The `ormolu`, `fourmolu`, `floskell` and `stylish-haskell` plugins, a testament to the code formatting wars of our community,
-- The `eval` plugin, a code lens provider to evaluate code in comments,
-- The `retrie` plugin, a code actions provider to execute retrie commands.
+* Request handlers respond to requests from an LSP client. They must fulfill these requests as quickly as possible.
+* Notification handlers receive notifications from code not directly triggered by a user or client.
+* Rules add new targets to the Shake build graph. Most plugins do not need to define new rules.
+* Commands are an LSP abstraction for user-initiated actions that the server handles. These actions can be long-running and involve multiple modules.
```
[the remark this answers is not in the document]
dschrempf: This is a good summary of the LSP communication scheme. I moved part of the summary before "The anatomy of a plugin" into a subsection called "Notes about the Language Server Protocol" and tried to consolidate the rest. I think the text has improved a lot.

docs/contributing/plugin-tutorial.md
dschrempf: I have moved the description of the Either up to when we introduce `CommandFunction`. Please have a look.

docs/contributing/plugin-tutorial.md
dschrempf: Here I do not know what to do. This seems like an integral part of the tutorial. Should we just leave it like that at the moment and add an issue to fix the tutorial later?

docs/contributing/plugin-tutorial.md
dschrempf: I have changed this to

```
+A good approach is to search for the ID of an existing plugin (e.g., `hls-class-plugin`):
```

dschrempf:
Hi @VeryMilkyJoe , I have implemented your proposed changes, thank you. Please have a look if the result is to your liking. One more involved task is still left: The `LspFuncs` problem (`CodeLensProvider` is also gone). I suggest merging these changes and creating an issue, but it is a suboptimal solution.

VeryMilkyJoe (approved):
Looks great thank you, I just spotted a spelling mistake!

docs/contributing/plugin-tutorial.md
fendor: We are updating the tutorial right now, I don't think it makes sense to update the tutorial to still contain outdated info.

docs/contributing/plugin-tutorial.md
VeryMilkyJoe: That is much clearer, good call!

docs/contributing/plugin-tutorial.md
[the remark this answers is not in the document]
VeryMilkyJoe: Thanks, I agree :)

docs/contributing/plugin-tutorial.md
fendor: Also, there are exactly two mentions of `LspFuncs`, why is it integral? Let's just delete the reference here and in the `provider` function just below it. The type `CodeLensProvider` doesn't exist any more, so that needs to be updated anyway, right?

docs/contributing/plugin-tutorial.md
```diff
@@ -29,254 +64,167 @@ Here is a visual statement of what we want to accomplish:
...
 
-For our command, we are going to have a very simple handler that receives a diff (`WorkspaceEdit`) and returns it to the client. The diff will be generated by our code lens provider and sent as part
-of the code lens to the LSP client, who will send it back to our command handler when the user activates
-the code lens:
-```haskell
-importCommandId :: CommandId
-importCommandId = "ImportLensCommand"
+`CommandFunction` takes an `ideState` and a JSON-encodable argument. `LspM` is a monad transformer with access to IO, and having access to a language context environment `Config`. The action evaluates to an `Either` value. `Left` indicates failure with a `ResponseError`, `Right` indicates sucess with a `Value`.
```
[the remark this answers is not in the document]
dschrempf: Thank you!

fendor:
I pushed an update where we are compiling the `plugin-tutorial.md` using HLS head. The implementation uses `markdown-unlit` and requires a symlink to `plugin-tutorial.lhs`.

In the process, I had to re-implement large parts of the code in the tutorial, which I think is a net-positive.

dschrempf:
That's great, thank you! I focussed on the text, and mostly copied the code over, so it's great if we can fix that too!

fendor (approved):
LGTM, thanks @dschrempf for pushing this over the finish line!

PR #4572, opened 2025-04-27
Set the XDG_CACHE_HOME for tests

fendor:
This makes sure each test run is independent.

Makes the tests
* iface-error-test-1
* iface-th-test

less flaky locally.
Should not have any effect on the CI flakiness issue.


soulomoon (commented):
LGTM. But ghcide test is not running for this PR, do you know why it is skipped

fendor:
https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/test.yml#L51

Outdated skip in CI. Gonna fix that!

michaelpj (approved):

fendor:
Test failures all look genuine, so more investigation is required.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -1111,7 +1111,7 @@ getModSummaryFromImports env fp _modTime mContents = do
     modLoc <- liftIO $ if mod == mAIN_NAME
         -- specially in tests it's common to have lots of nameless modules
         -- mkHomeModLocation will map them to the same hi/hie locations
-        then mkHomeModLocation dflags (pathToModuleName fp) fp
+        then mkHomeModLocation dflags mod fp
```
fendor, line 1114: This seems to be causing long-path issues on windows.
I am unsure how relevant this still is... Anybody an idea?

ghcide-test/exe/FindDefinitionAndHoverTests.hs
```diff
@@ -187,7 +187,7 @@ tests = let
   holeL65 = Position 65 8  ;  hleInfo2 = [ExpectHoverText ["_ :: a -> Maybe a"]]
   cccL17 = Position 17 16  ;  docLink = [ExpectHoverTextRegex "\\*Defined in 'GHC.Types'\\* \\*\\(ghc-prim-[0-9.]+\\)\\*\n\n"]
   imported = Position 56 13 ; importedSig = getDocUri "Foo.hs" >>= \foo -> return [ExpectHoverText ["foo", "Foo", "Haddock"], mkL foo 5 0 5 3]
-  reexported = Position 55 14 ; reexportedSig = getDocUri "Bar.hs" >>= \bar -> return [ExpectHoverText ["Bar", "Bar", "Haddock"], if ghcVersion >= GHC94 && ghcVersion < GHC910 then mkL bar 3 5 3 8 else mkL bar 3 0 3 14]
+  reexported = Position 55 14 ; reexportedSig = getDocUri "Bar.hs" >>= \bar -> return [ExpectHoverText ["Bar", "Bar", "Haddock"], mkL bar 3 0 3 14]
```
fendor, line 190: This seems to be caused by a race/inconsistency between serving a hover request based on `hiedb` and getting the information from an interface file.

hls-test-utils/src/Test/Hls.hs
```diff
@@ -516,23 +516,45 @@ runWithLockInTempDir tree act = withLock lockForTempDirs $ do
...
         act fs
+    where
+        cache_home_var = "XDG_CACHE_HOME"
+        -- Set the dir for "XDG_CACHE_HOME".
+        -- When the operation finished, make sure the old value is restored.
+        withTempCacheHome tempCacheHomeDir act =
+            bracket
+              (do
+                old_cache_home <- lookupEnv cache_home_var
+                setEnv cache_home_var tempCacheHomeDir
```
jian-lin, line 544: `XDG_CACHE_HOME` is also set in [`setupTestEnvironment`][1].  That should be a no-op after this PR.  What about removing that to reduce confusion?

[1]: https://github.com/haskell/haskell-language-server/blob/664931bbd62e8d797e28d3dec1ca63e1620b45be/hls-test-utils/src/Test/Hls.hs#L578

ghcide/src/Development/IDE/Core/Compile.hs
fendor: In a test cabal project, it seems we always create separate `.hi`/`.hie` files

image

fendor: But not, if there are multiple "main" scripts: 
image

fendor: This is unfortunately a case we have to support, so this change can't be applied as is.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -1120,11 +1120,7 @@ getModSummaryFromImports env fp _modTime mContents = do
     liftIO $ evaluate $ rnf textualImports
 
 
-    modLoc <- liftIO $ if mod == mAIN_NAME
-        -- specially in tests it's common to have lots of nameless modules
-        -- mkHomeModLocation will map them to the same hi/hie locations
-        then mkHomeModLocation dflags (pathToModuleName fp) fp
-        else mkHomeModLocation dflags mod fp
+    modLoc <- liftIO $ mkHomeModLocation dflags mod fp
```
fendor:
```suggestion
    modLoc <- liftIO $ if mod == mAIN_NAME
        -- specially in tests it's common to have lots of nameless modules
        -- mkHomeModLocation will map them to the same hi/hie locations
        then mkHomeModLocation dflags (pathToModuleName fp) fp
        else mkHomeModLocation dflags mod fp
```

soulomoon:
Is it blocking by the long path issue for windows? Can we shrink the test cache dir path?

fendor:
@soulomoon Yes, it is blocking :( 

I have no idea how to further shrink down the test cache dir path, without either reimplementing `createTemporaryDirectory`, or mangling test-case names.
I am further confused why this even is an issue, as the CI runners allegedly enable long file paths.

soulomoon:
That seems to be ghc's issue ? see https://gitlab.haskell.org/ghc/ghc/-/issues/13373 @fendor 

So to workaround this, we might have to customize our `createTemporaryDirectory` ?

fendor:
I am not sure if the issue is still relevant

fendor:
The lazy solution to set a shorter test directory on windows seems to be working. Finally!

@guibou, @soulomoon perhaps a final review? Otherwise, Ill just merge it.

PR #4576, opened 2025-05-04
Make cradle dependencies absolute paths

### LINKED ISSUE #1068: Reload required when adding new dependencies to package.yaml/.cabal file

jian-lin:
This patch fixes #1068 when the cabal plugin is not used.

---

   
The code change in Session.hs is written by @fendor.  Tests and comments are added by me.

### DIFF
--- a/ghcide-test/exe/WatchedFileTests.hs
+++ b/ghcide-test/exe/WatchedFileTests.hs
@@ -3,11 +3,14 @@
 
 module WatchedFileTests (tests) where
 
-import           Config                          (testWithDummyPluginEmpty')
+import           Config                          (mkIdeTestFs,
+                                                  testWithDummyPlugin',
+                                                  testWithDummyPluginEmpty')
 import           Control.Applicative.Combinators
 import           Control.Monad.IO.Class          (liftIO)
 import qualified Data.Aeson                      as A
 import qualified Data.Text                       as T
+import qualified Data.Text.IO                    as T
 import           Development.IDE.Test            (expectDiagnostics)
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types     hiding
@@ -18,6 +21,7 @@ import           Language.LSP.Protocol.Types     hiding
 import           Language.LSP.Test
 import           System.Directory
 import           System.FilePath
+import           Test.Hls.FileSystem
 import           Test.Tasty
 import           Test.Tasty.HUnit
 
@@ -69,6 +73,17 @@ tests = testGroup "watched files"
         sendNotification SMethod_WorkspaceDidChangeWatchedFiles $ DidChangeWatchedFilesParams
                [FileEvent (filePathToUri $ sessionDir </> "B.hs") FileChangeType_Changed ]
         expectDiagnostics [("A.hs", [(DiagnosticSeverity_Error, (3, 4), "Couldn't match expected type '()' with actual type 'Int'", Just "GHC-83865")])]
+      , testWithDummyPlugin' "reload HLS after .cabal file changes" (mkIdeTestFs [copyDir ("watched-files" </> "reload")]) $ \sessionDir -> do
+          let hsFile = "src" </> "MyLib.hs"
+          _ <- openDoc hsFile "haskell"
+          expectDiagnostics [(hsFile, [(DiagnosticSeverity_Error, (2, 7), "Could not load module \8216Data.List.Split\8217", Nothing)])]
+          let cabalFile = "reload.cabal"
+          cabalContent <- liftIO $ T.readFile cabalFile
+          let fix = T.replace "build-depends:    base" "build-depends:    base, split"
+          liftIO $ T.writeFile cabalFile (fix cabalContent)
+          sendNotification SMethod_WorkspaceDidChangeWatchedFiles $ DidChangeWatchedFilesParams
+            [ FileEvent (filePathToUri $ sessionDir </> cabalFile) FileChangeType_Changed ]
+          expectDiagnostics [(hsFile, [])]
     ]
   ]
 
--- a/ghcide-test/data/watched-files/reload/reload.cabal
+++ b/ghcide-test/data/watched-files/reload/reload.cabal
@@ -1,0 +1,12 @@
+cabal-version:      2.4
+name:               reload
+version:            0.1.0.0
+author:             Lin Jian
+maintainer:         me@linj.tech
+build-type:         Simple
+
+library
+  exposed-modules:  MyLib
+  build-depends:    base
+  hs-source-dirs:   src
+  default-language: Haskell2010
--- a/ghcide-test/data/watched-files/reload/src/MyLib.hs
+++ b/ghcide-test/data/watched-files/reload/src/MyLib.hs
@@ -1,0 +1,6 @@
+module MyLib (someFunc) where
+
+import Data.List.Split
+
+someFunc :: IO ()
+someFunc = putStrLn "someFunc"
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -698,7 +698,15 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
             let ncfp = toNormalizedFilePath' (toAbsolutePath file)
             cachedHieYamlLocation <- HM.lookup ncfp <$> readVar filesMap
             hieYaml <- cradleLoc file
-            sessionOpts (join cachedHieYamlLocation <|> hieYaml, file) `Safe.catch` \e ->
+            let
+              -- Each one of deps will be registered as a FileSystemWatcher in the GhcSession action
+              -- so that we can get a workspace/didChangeWatchedFiles notification when a dep changes.
+              -- The GlobPattern of a FileSystemWatcher can be absolute or relative.
+              -- We use the absolute one because it is supported by more LSP clients.
+              -- Here we make sure deps are absolute and later we use those absolute deps as GlobPattern.
+              absolutePathsCradleDeps (eq, deps)
+                = (eq, fmap toAbsolutePath deps)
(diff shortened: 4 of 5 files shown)


plugins/hls-cabal-plugin/test/Utils.hs
```diff
@@ -61,6 +65,14 @@ runHaskellAndCabalSession :: FilePath -> Session a -> IO a
 runHaskellAndCabalSession subdir =
     failIfSessionTimeout . runSessionWithServer def (cabalPlugin <> cabalHaskellPlugin) (testDataDir </> subdir)
 
+runShiftedRootInTmpDirSession :: FilePath -> (FilePath -> Session a) -> IO a
+runShiftedRootInTmpDirSession subdir =
+    failIfSessionTimeout . runSessionWithTestConfig def
+        { testDirLocation = Right $ mkVirtualFileTree testDataDir [copyDir subdir]
+        , testShiftRoot = True
```
jian-lin, line 72: It takes me some time to figure out that `testShiftRoot = True` is necessary for tests to work.

When `testShiftRoot` is the default value `False`, cradle dependencies except `hie.yaml` are relative paths, which can be wrong under the working directory of HLS.  When the path is wrong, its mtime is also wrong.  Wrong mtime makes it [always be considered up-to-date][1], which will never [invalidate][2] the cache of `GhcSession` action.

Maybe it is better to make cradle dependencies always absolute?

[1]: https://github.com/haskell/haskell-language-server/blob/173b5a7fbfb92c3c6a3375be100b662226aad852/ghcide/session-loader/Development/IDE/Session.hs#L679
[2]: https://github.com/haskell/haskell-language-server/blob/173b5a7fbfb92c3c6a3375be100b662226aad852/ghcide/src/Development/IDE/Core/Rules.hs#L719

plugins/hls-cabal-plugin/test/CabalAdd.hs
```diff
@@ -183,3 +195,17 @@ cabalAddTests =
...
+        hsdoc <- openDoc haskellFile "haskell"
+        _ <- waitForDiagnosticsFrom hsdoc
+        cabalContent <- liftIO $ T.readFile cabalFile
+        let fix = T.replace "build-depends:    base" "build-depends:    base, split"
+        liftIO $ T.writeFile cabalFile (fix cabalContent)
+        sendNotification SMethod_WorkspaceDidChangeWatchedFiles $ DidChangeWatchedFilesParams
+            [ FileEvent (filePathToUri $ root </> cabalFile) FileChangeType_Changed ]
+        diagnostics <- waitForDiagnosticsFrom hsdoc
+        case diagnostics of
+            [diagnostic] -> liftIO $ assertEqual "Wrong diagnostic after HLS restarts" (Just $ InR "GHC-66111") (diagnostic ^. L.code)
```
jian-lin, line 210: This fails in GHC 9.4 and 9.6.  Should we use CPP or should we match something else for all GHC versions?

fendor: Wait, this test can only succeed if the hls-cabal-plugin is enabled...? But this patch only works if the hls-cabal-plugin is disabled? Or is this test run without the `hls-cabal-plugin`? If the test is run without the plugin, I think it should be part of the `ghcide-test`s.

Regarding `InR "GHC-66111"` (which I assume to be the reason for the test case to fail) The `expectDiagnostics` test helper ignores the test code for the unsupported GHC versions 9.4 and 9.6, maybe use that if possible?

The windows failure is flaky, just ignore :)

plugins/hls-cabal-plugin/test/Utils.hs
fendor: Yeah, cradle dependencies are supposed to be always absolute. Perhaps this is even a bug in `hie-bios`...

fendor (approved):
LGTM, small nitpick

plugins/hls-cabal-plugin/test/CabalAdd.hs
jian-lin: > this test can only succeed if the hls-cabal-plugin is enabled...?

`hls-cabal-plugin` contains two plugins: [`cabal`][1] and [`cabalHaskellIntegration`][2]. This test works if and only if HLS does not handle `.cabal` files, i.e., the `cabal` plugin is not used.

> If the test is run without the plugin, I think it should be part of the `ghcide-test`s.

I was not quite sure where to put this test.  I will move this test to `ghcide-test`.

When a fix for #1068 with `cabal` plugin **enabled** is proposed in the future, its test probably should be put here.  I think it makes (some) sense to put tests for these two cases together so I put this test (for #1068 with `cabal` plugin **disabled**) here.

[1]: https://github.com/haskell/haskell-language-server/blob/173b5a7fbfb92c3c6a3375be100b662226aad852/src/HlsPlugins.hs#L154
[2]: https://github.com/haskell/haskell-language-server/blob/173b5a7fbfb92c3c6a3375be100b662226aad852/src/HlsPlugins.hs#L155



> The `expectDiagnostics` test helper ignores the test code for the unsupported GHC versions 9.4 and 9.6, maybe use that if possible?

`expectDiagnostics` seems not suitable here since it only ignores [GHC < 9.6][3].  Here, the actual diagnostic code is `Nothing` for GHC 9.6 and 9.4 and the expected diagnostic is as follows:

```
  [ Diagnostic
      { _range = Range {_start = Position {_line = 2, _character = 0}, _end = Position {_line = 2, _character = 22}},
        _severity = Just DiagnosticSeverity_Warning,
        _code = Just (InR "GHC-66111"),
        _codeDescription = Nothing,
        _source = Just "typecheck",
        _message = "The import of \8216Data.List.Split\8217 is redundant\n  except perhaps to import instances from \8216Data.List.Split\8217\nTo import instances alone, use: import Data.List.Split()",
        _tags = Just [DiagnosticTag_Unnecessary],
        _relatedInformation =
          Just
            [ DiagnosticRelatedInformation
                { _location =
                    Location
                      { _uri = Uri {getUri = "file:///tmp/hls-test-root/extra-dir-44362888928157/src/MyLib.hs"},
                        _range = Range {_start = Position {_line = 2, _character = 0}, _end = Position {_line = 2, _character = 22}}
                      },
                  _message = "TypeCheck"
                }
            ],
        _data_ = Just (Object (fromList [("attachedReason", Array [String "-Wunused-imports"])]))
      }
  ]
```



[3]: https://github.com/haskell/haskell-language-server/blob/173b5a7fbfb92c3c6a3375be100b662226aad852/hls-test-utils/src/Development/IDE/Test/Diagnostic.hs#L69

fendor: Ok, then you can use the `ghcVersion` variable to change the assertion based on the GHC version. This avoids CPP which much more annoying.

plugins/hls-cabal-plugin/test/Utils.hs
jian-lin: > cradle dependencies are supposed to be always absolute

That is good to know.  When I have time, I will check hie-bios to see if I can do something there.

ghcide-test/exe/WatchedFileTests.hs
```diff
@@ -69,6 +73,17 @@ tests = testGroup "watched files"
...
+          let hsFile = "src" </> "MyLib.hs"
+          _ <- openDoc hsFile "haskell"
+          expectDiagnostics [(hsFile, [(DiagnosticSeverity_Error, (2, 7), "Could not load module \8216Data.List.Split\8217.\nIt is a member of the hidden package \8216split", Nothing)])]
+          let cabalFile = "reload.cabal"
+          cabalContent <- liftIO $ T.readFile cabalFile
+          let fix = T.replace "build-depends:    base" "build-depends:    base, split"
+          liftIO $ T.writeFile cabalFile (fix cabalContent)
+          sendNotification SMethod_WorkspaceDidChangeWatchedFiles $ DidChangeWatchedFilesParams
+            [ FileEvent (filePathToUri $ sessionDir </> cabalFile) FileChangeType_Changed ]
+          expectDiagnostics [(hsFile, [])]
```
jian-lin, line 86: When this test is in `hls-cabal-plugin-tests`, there [is][1] a diagnostic after HLS restarts.

```
The import of \8216Data.List.Split\8217 is redundant\n  except perhaps to import instances from \8216Data.List.Split\8217\nTo import instances alone, use: import Data.List.Split()
```

Moving the tests into `ghcide-tests` causes that diagnostic to disappear.  I have not tried to find the cause yet.  I guess it is because we enable fewer plugins in the test now and the plugin providing this diagnostic is not enabled.

I have verified that HLS does reload after the `SMethod_WorkspaceDidChangeWatchedFiles` notification.

[1]: https://github.com/haskell/haskell-language-server/pull/4576#discussion_r2072602661

jian-lin (commented):
I think previous comments are addressed.

ghcide-test/exe/WatchedFileTests.hs
fendor: Hm, it is weird, the diagnostic is generated by GHC, not any plugin... In any way, still a clear improvement, thank you!

fendor (approved):

fendor:
@jian-lin Please rebase, then we can get this in :)

jian-lin:
Now all windows `/Diagnostics in .hs files from invalid .cabal file/` tests fail.  I vaguely remember that all tests pass before the rebase.  Any ideas?

fendor:
Hm, that's indeed weird... No, unfortunately nothing comes immediately to mind.

jian-lin:
Thanks for all the help!

PR #4582, opened 2025-05-06
Use hie-bios 0.15.0

### LINKED ISSUE #4562: Speed up startup times: `hie-bios` should use `cabal path`
The startup times of HLS can be quite slow, especially on `cabal` projects.

One of the reasons is that we essentially call `cabal` four times:

* What's the path of the ghc binary?
* What's the `libdir` of the ghc binary?
* What's the `ghc-pkg` location of the ghc binary?
* What are the compilation options to compile the package?

The first three commands boil down to roughly this:

```sh
cabal exec -- ...
```

`cabal exec` can be quite slow though, as it sets up `source-repository`s, runs the solver, etc...

On a perfectly set up HLS code base, `cabal exec` needs quite some time:

```sh
$ time cabal exec -- ghc --version
Configuration is affected by the following files:
- cabal.project
- cabal.project.local
The Glorious Glasgow Haskell Compilation System, version 9.6.7

________________________________________________________
Executed in    1.10 secs      fish           external
   usr time  842.53 millis    0.00 millis  842.53 millis
   sys time  219.74 millis    1.28 millis  218.46 millis
```

So around 3 seconds are spent on figuring out the correct GHC binary, before we do anything.

The newly introduced `cabal path` command on the other hand:

```sh
$ time cabal path
Configuration is affected by the following files:
- cabal.project
- cabal.project.local
compiler-flavour: ghc
compiler-id: ghc-9.6.7
compiler-path: /home/hugin/.ghcup/bin/ghc-9.6.7
cache-home: /home/hugin/.cache/cabal
remote-repo-cache: /home/hugin/.cache/cabal/packages
logs-dir: /home/hugin/.cache/cabal/logs
store-dir: /home/hugin/.local/state/cabal/store
config-file: /home/hugin/.config/cabal/config
installdir: /home/hugin/.local/bin

________________________________________________________
Executed in   34.27 millis    fish           external
   usr time   16.10 millis    0.00 micros   16.10 millis
   sys time   14.06 millis  965.00 micros   13.10 millis
```

and contains exactly the information we need: `compiler-path`.
Likely, that's the only thing we need.

So, the idea is to update `hie-bios` to use `cabal path` if `cabal` is recent enough (`>= 3.14`) to shave off a couple of seconds of start up time on big Haskell projects.

Changes are exclusively to be done in `hie-bios`, in particular https://github.com/haskell/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L578

fendor:
This allows us to benefit from the `cabal path` command in hie-bios for improved start-up time.

Also, update CI to use `cabal 3.14` so that we actually test the `cabal path` code path.

Closes #4562

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -7,7 +7,7 @@ inputs:
   cabal:
     description: "Cabal version"
     required: false
-    default: "3.10.2.0"
+    default: "3.14.2.0"
   os:
     description: "Operating system: Linux, Windows or macOS"
     required: true
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -123,7 +123,7 @@ jobs:
       matrix:
         ghc: ['9.8', '9.10']
         os: [ubuntu-latest]
-        cabal: ['3.10']
+        cabal: ['3.14']
         example: ['cabal', 'lsp-types']
 
     steps:
--- a/cabal.project
+++ b/cabal.project
@@ -8,7 +8,7 @@ packages:
          ./hls-test-utils
 
 
-index-state: 2025-04-19T07:34:07Z
+index-state: 2025-05-06T13:26:29Z
 
 tests: True
 test-show-details: direct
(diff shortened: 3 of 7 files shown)


fendor:
cc @alexfmpe

michaelpj (approved):

PR #4586, opened 2025-05-10
Only expand positional records if the DataCon application is fully saturated

### LINKED ISSUE #4564: Code action "expand positional record" produces invalid code.
### Your environment

HLS version is `Version 2.9.0.1 x86_64 ghc-9.10.1` under VSCode/Linux

### Steps to reproduce

I've failed to reproduce this outside of a modified version of GHC.

### Expected behaviour

Code actions on valid code should result in valid code.

### Actual behaviour

Given this snippet:

```
data T = MkT { fa :: Int, fb :: Char}

foo :: Int -> Char -> T
foo x = (MkT x)
```

HLS suggests replacing `(MkT x)` with `MkT { fa = x }`.

However this fails to typecheck. As the first expression has type `Char -> T` while the later has type `T`

ozkutuk:
Fixes #4564.

"Positional record expansion" only makes sense if the constructor application has all the required arguments provided, otherwise invalid code is generated as demonstrated in the linked issue. This PR augments the collected records with "saturation" information, so that unsatured applications can be filtered out during the code action generation.

(cc: @jetjinser, do you mind taking a look into this, as you were the author of #4447)

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -151,10 +151,17 @@ descriptor recorder plId =
 codeActionProvider :: PluginMethodHandler IdeState 'Method_TextDocumentCodeAction
 codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
   nfp <- getNormalizedFilePathE (docId ^. L.uri)
-  CRR {crCodeActions, enabledExtensions} <- runActionE "ExplicitFields.CollectRecords" ideState $ useE CollectRecords nfp
+  CRR {crCodeActions, crCodeActionResolve, enabledExtensions} <- runActionE "ExplicitFields.CollectRecords" ideState $ useE CollectRecords nfp
   -- All we need to build a code action is the list of extensions, and a int to
   -- allow us to resolve it later.
-  let actions = map (mkCodeAction enabledExtensions) (RangeMap.filterByRange range crCodeActions)
+  let recordUids = [ uid
+                      | uid <- RangeMap.filterByRange range crCodeActions
+                      , Just record <- [IntMap.lookup uid crCodeActionResolve]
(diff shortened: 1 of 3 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -169,6 +176,11 @@ codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
       , _data_ = Just $ toJSON uid
       }
 
+    isConvertible :: RecordInfo -> Bool
+    isConvertible = \case
+      RecordInfoApp _ (RecordAppExpr Unsaturated _ _) -> False
+      _ -> True
```
fendor, line 182: Since it is only two cases, maybe we should be exhaustive?

fendor (approved):
LGTM, thanks for taking care of this!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
ozkutuk: Maybe I misunderstood you, but I can't see how it is only two cases. This is what I come up with if I try to make this fully exhaustive:

```haskell
    isConvertible :: RecordInfo -> Bool
    isConvertible = \case
      RecordInfoApp _ (RecordAppExpr Unsaturated _ _) -> False
      RecordInfoApp _ (RecordAppExpr Saturated _ _) -> True
      RecordInfoPat {} -> True
      RecordInfoCon {} -> True
```

I personally find this a bit too verbose, but if that's the style the codebase favors, I would be happy to change.

fendor: It is just my preference, but since you are maintaining the plugin right now, your opinion matters more than mine in style preferences. 

Also, thought it would be two, not 4, so yeah, that's more verbose than anticipated.

PR #4591, opened 2025-05-13
Use shorter test names in ghcide-tests

fendor:
When running the ghcide-tests, the test names are overflowing my terminal, leading to duplicated lines, making it more annoying to see the test progress.

Admittedly, this only happens when the terminal isn't using the full desktop width:

image

By shortening some of the longer test names, the output looks like this:

image


soulomoon (approved):
LGTM

Issue #4443, opened 2024-11-02
Performance: adding an `import` retriggers a long "Processing


guibou:
### Your environment
Which OS do you use? Linux / NixOS
Which version of GHC do you use and how did you install it? GHC 9.10.1, installed using nix/nixpkgs.
How is your project built (alternative: link to the project)? A custom build system based on nix.

Which LSP client (editor/plugin) do you use? Neovim builtin lsp
Which version of HLS do you use and how did you install it? 9.10.1
Have you configured HLS in any way (especially: a `hie.yaml` file)?

My `hie.yaml` contains:

```
cradle:
  bios:
    shell: "nix run .#haskell_hie_bios"
```

And the nix target `haskell_hie_bios` lists all the required build flags used by GHC to build one monolithic packages composed of 1089 modules. It also contains the listing of the 1089 modules, eg:

```
-XFoo
-XBar
-ipath/to/src
-ipath/to/src2
-ffoo
path/to/src/Module/Module.hs
path/to/src2/Module2/Module2.hs
```

### Steps to reproduce

I don't know how to reproduce that outside of our large codebase, but what do I observe:

- I've turned `--debug --log /path/to/a/file.log`
- I start my editor (neovim, note that my collegues using vscode observes the same problem), load a file (for the example, I loaded a file which is at the end of the dependency tree, it contains a `main` function and no other module depends on that module).
- After a bit of time, I can see:
```
2024-11-02T08:45:19.194755Z | Debug | Finished: C:GetHieAst:/home/guillaume/jinko/jinko/simwork/core-webservice/app/Local/Main.hs Took: 2m34s
```

In the log.
- I can edit my file, do some change, that's super reactive. I can add function, hover for documentation, go to definition, everything is smooth and fast. Code action works fine, everything works great.
- If I go to the top of the module, `import` section, I can add/remove symbols in the currently existing imported module, but if I add (or remove) an `import` statment, it starts `Processing` for a long time. For example, after adding `import Data.Functor` on top of my file, I can see `Processing` and `Indexing` (and actually, a few, sometime 2 or 3 `Processing`) and for a few seconds, HLS is not reactive anymore. Here is the message in the log:

```
2024-11-02T08:48:35.124980Z | Debug | Finished: C:GetHieAst:/home/guillaume/jinko/jinko/simwork/core-webservice/app/Local/Main.hs Took: 36.24s
[...]

PR #4594, opened 2025-05-19
compute deps fingerprint only

soulomoon:


guibou:
Thank you for working on that topic.

I applied the patch and did not observe much improvement in my work codebase (e.g. any added `import` triggers a long, 30s-1minute rebuild).

I will have a look a bit later during the day with the "simple" examples that I had provided in #4443, maybe this actually fix / improves them and maybe there is another issue with my codebase.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1145,7 +1145,7 @@ needsCompilationRule file
   | "boot" `isSuffixOf` fromNormalizedFilePath file =
     pure (Just $ encodeLinkableType Nothing, Just Nothing)
 needsCompilationRule file = do
-  graph <- useNoFile GetModuleGraph
+  graph <- use GetFileModuleGraph file
```
soulomoon, line 1148: I forget to change this rule previously.

wz1000 (approved):
These changes look very promising, thank you!

soulomoon:
> I think for correctness reasons we might need to restrict di to only include keys in the downward dependency closure. 

Since we are not restrict it before, it should fine if we are not doing it right away. But indeed it might be a good idea to implement the restriction to ensure the correctness.  Also I am a bit worried about the performance impact doing the restriction. Givem the above, I'll open another PR to do it.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1247,6 +1251,19 @@ mainRule recorder RulesConfig{..} = do
...
+        let finger = lookupFingerprint file di (depTransDepsFingerprints di)
+        return (fingerprintToBS <$> finger, ([], finger))
+    defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GetModuleGraphTransReverseDepsFingerprints file -> do
+        di <- useNoFile_ GetModuleGraph
+        let finger = lookupFingerprint file di (depTransReverseDepsFingerprints di)
+        return (fingerprintToBS <$> finger, ([], finger))
+    defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \GetModuleGraphImmediateReverseDepsFingerprints file -> do
+        di <- useNoFile_ GetModuleGraph
+        let finger = lookupFingerprint file di (depImmediateReverseDepsFingerprints di)
+        return (fingerprintToBS <$> finger, ([], finger))
```
fendor, line 1265: For the future, I think this should be a function `moduleGraphRules`

PR #4596, opened 2025-05-21
Avoid unnecessary recompilation due to -haddock

### LINKED ISSUE #4511: All dependent files are recompiled on a save
When a save a file, recompilation is triggered for all the files which depend on the file I saved.

I tracked down the bug to the fact that

1. Interfaces are created with `-haddock` option set.
2. Recompilation checking is run in a context where `-haddock` is not set.

This seems to be complicated with all the logic with turning -haddock on in some situations and not in others.

The following diff "fixes" the immediate problem for me, but I don't know if there are some situations where interfaces are created without `-haddock` being enabled.

```
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -1275,7 +1275,7 @@ loadInterface
   -> RecompilationInfo m
   -> m ([FileDiagnostic], Maybe HiFileResult)
 loadInterface session ms linkableNeeded RecompilationInfo{..} = do
-    let sessionWithMsDynFlags = hscSetFlags (ms_hspp_opts ms) session
+    let sessionWithMsDynFlags = hscSetFlags (gopt_set (ms_hspp_opts ms) Opt_Haddock) session
         mb_old_iface = hirModIface . fst <$> old_value
         mb_old_version = snd <$> old_value

```

It would be simpler if the options chosen to compile a file were fixed at the start of compilation and not modified in ad-hoc places later on.

In my opinion this is a serious bug.

fendor:
Due to unprincipled adding and removing the `-haddock` flag during compilation and recompilation checking, we were performing more work than necessary.
We avoid this by compiling everything with `-haddock` by default. This is safe nowadays, we have essentially been doing this for many releases, and know this is fine.
For the occasion where we actually want to parse without the `-haddock` flag, we keep explicitly disabling it.

We enable `-haddock` flag during session loading, since we already perform a number of DynFlags tweaks.
This behaviour is dependent on the `OptHaddockParse` opton, which can, currently, only be modified at compile-time.

Closes #4511


fendor:
I would love to implement a regression test for this, but I am currently unsure how.

fendor:
I manually tested it on GHC and *think* I can confirm this fixes the recompilation issue but I struggle to come up with a test case in HLS itself.

cassandracomar:
possible issue with this PR -- the haddocks for definitions in the local module don't appear to be available in the completion details. I instead see:

```
2025-06-02T17:03:10.743161Z | Warning | No plugin handles this "completionItem/resolve" request.
```

in the lsp server's log output. non-local details DO appear to be resolvable, however, and that warning isn't printed. if I just use master instead of applying this PR, this warning is never printed for either local or non-local bindings.

fendor:
Thank you for testing this change! I will look into it.

fendor:
Can't reproduce, with GHC 9.6.7, I get: 
image

and the logs are clear.

fendor:
I am unsure how to test this, I manually verified that the behaviour reported in #4511 doesn't seem to be occurring in GHC.

Are we fine with merging this as is?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -973,7 +971,7 @@ regenerateHiFile sess f ms compNeeded = do
     opt <- getIdeOptions
 
     -- Embed haddocks in the interface file
-    (diags, mb_pm) <- liftIO $ getParsedModuleDefinition hsc opt f (withOptHaddock ms)
+    (diags, mb_pm) <- liftIO $ getParsedModuleDefinition hsc opt f ms
```
soulomoon: We do still need the comment above ?

ghcide/src/Development/IDE/Types/Options.hs
```diff
@@ -69,9 +69,7 @@ data IdeOptions = IdeOptions
     -- ^ When to typecheck reverse dependencies of a file
   , optHaddockParse       :: OptHaddockParse
     -- ^ Whether to return result of parsing module with Opt_Haddock.
-    --   Otherwise, return the result of parsing without Opt_Haddock, so
-    --   that the parsed module contains the result of Opt_KeepRawTokenStream,
-    --   which might be necessary for hlint.
+    --   Otherwise, return the result of parsing without Opt_Haddock.
```
soulomoon: OptHaddockParse is not a boolean so we do not need `    --   Otherwise, return the result of parsing without Opt_Haddock.`, I think.

soulomoon (approved):
Good work @fendor .
A few nitpicks for comments.

Before this commit, we do not repect optHaddockParse, and now we correctly repect it in the session loader. Is my understanding correct?

ghcide/src/Development/IDE/Types/Options.hs
fendor: I changed the phrasing, thanks!

ghcide/src/Development/IDE/Core/Rules.hs
fendor: I rephrased, but don't want to remove the comment affirming that we parse with '-haddock' here (usually)

PR #4608, opened 2025-06-01
allow Diff 1.x

peterbecich:
https://hackage.haskell.org/package/Diff

Compilation succeeds with `Diff ^>= 1.0`

```
cabal test --enable-tests --constraint 'Diff>=1.0'
```

1 test fails:
```
1 out of 376 tests failed (535.93s)
```

```
FAIL (1.74s)
        plugins/hls-refactor-plugin/test/Main.hs:3966:
        CodeAction with title "add signature: hello :: GHC.Types.Any -> IO ()" not found in ["add signature: hello :: GHC.Types.ZonkAny 0 -> IO ()"]
        Use -p '/hello = print/' to rerun this test only.

```

### DIFF
--- a/ghcide/ghcide.cabal
+++ b/ghcide/ghcide.cabal
@@ -57,7 +57,7 @@ library
     , deepseq
     , dependent-map
     , dependent-sum
-    , Diff                         ^>=0.5
+    , Diff                         ^>=0.5 || ^>=1.0.0
     , directory
     , dlist
     , enummapset
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -473,7 +473,7 @@ library hls-eval-plugin
     , bytestring
     , containers
     , deepseq
-    , Diff                  ^>=0.5
+    , Diff                  ^>=0.5 || ^>=1.0.0
     , dlist
     , extra
     , filepath
--- a/hls-plugin-api/hls-plugin-api.cabal
+++ b/hls-plugin-api/hls-plugin-api.cabal
@@ -60,7 +60,7 @@ library
     , data-default
     , dependent-map
     , dependent-sum         >=0.7
-    , Diff                  ^>=0.5
+    , Diff                  ^>=0.5 || ^>=1.0.0
     , dlist
     , extra
     , filepath


michaelpj:
I'm confused - does this work or not? If there's a test failure I guess we shouldn't merge it?

fendor:
The test failure is for GHC 9.10.2, which we don't test in CI right now. So, can be ignored. However, I think we should only support `Diff ^>= 1.0`, no reason to support both, right?

ghcide/ghcide.cabal
```diff
@@ -57,7 +57,7 @@ library
     , deepseq
     , dependent-map
     , dependent-sum
-    , Diff                         ^>=0.5
+    , Diff                         ^>=0.5 || ^>=1.0.0
```
fendor, line 60:
```suggestion
    , Diff                         ^>=1.0.0
```

haskell-language-server.cabal
```diff
@@ -473,7 +473,7 @@ library hls-eval-plugin
     , bytestring
     , containers
     , deepseq
-    , Diff                  ^>=0.5
+    , Diff                  ^>=0.5 || ^>=1.0.0
```
fendor, line 476:
```suggestion
    , Diff                  ^>=1.0.0
```

hls-plugin-api/hls-plugin-api.cabal
```diff
@@ -60,7 +60,7 @@ library
     , data-default
     , dependent-map
     , dependent-sum         >=0.7
-    , Diff                  ^>=0.5
+    , Diff                  ^>=0.5 || ^>=1.0.0
```
fendor, line 63:
```suggestion
    , Diff                  ^>=1.0.0
```

peterbecich:
> should only support Diff ^>= 1.0, no reason to support both, right?

NixPkgs `haskell-updates` branch has two versions of `Diff`:
* `Diff (0.5)`
* `Diff_1_0_2 (1.0.2)`

The existence of the old version 0.5 (pinned here https://github.com/NixOS/nixpkgs/blob/c8fe7786a1c661dc76f4398dbf952939f4aed378/pkgs/development/haskell-modules/configuration-hackage2nix/stackage.yaml#L631) indicates many other packages still require 0.5. By removing the support for `0.5` we are taking away some flexibility in building `haskell-language-server` in Nix.

cassandracomar:
we're able to loosen the version bounds in nixpkgs via jailbreaking. so this will take a little work to add jailbreaks for any reverse dependencies that restrict the version of Diff to `< 0.6`, but it's no more unusual than many of the other version upgrades we do regularly. as these changes are gated for awhile in the `haskell-updates` branch for testing, there's also plenty of time for reverse dependencies to pick up the new Diff version as well.

of note, I haven't yet found a package that builds with Diff-0.5.X that doesn't also build with Diff-1.0.X.

fendor (approved):
LGTM

fendor:
Looks like we do need to allow both `Diff` versions.

peterbecich:
https://github.com/haskell/haskell-language-server/pull/4608#issuecomment-2939109788 reset to the first commit

michaelpj (approved):

PR #4610, opened 2025-06-02
Local binding hints (WIP)

cassandracomar:
continuation of #4368.

so far, I've merged in changes in master and updated things so they compile with ghc-9.12. 

TODO:
1. get tests passing (the 9.12 flakes seem flaky)
2. allow both code lenses and inlay hints for rendering local sig hints, via a configurable toggle.

update: this now allows either code lenses or inlay hints for local bindings. as [we discussed](https://github.com/haskell/haskell-language-server/pull/4368#issuecomment-2932683893), I've defaulted this to code lenses, with an optional toggle to flip on inlay hints instead. but the code lenses don't have their tests restored as yet. will reintroduce those tomorrow.


ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -91,28 +101,35 @@ instance Pretty Log where
...
+  '[ 'PropertyKey "localBindingInlayHintOn" 'TBoolean,
+     'PropertyKey "mode" ('TEnum Mode)]
 properties = emptyProperties
   & defineEnumProperty #mode "Control how type lenses are shown"
     [ (Always, "Always displays type lenses of global bindings")
     , (Exported, "Only display type lenses of exported global bindings")
     , (Diagnostics, "Follows error messages produced by GHC about missing signatures")
     ] Always
+  & defineBooleanProperty #localBindingInlayHintOn
+    "Display inlay hints of local bindings"
```
cassandracomar, line 139: this needs a better explanation for documentation

test/testdata/schema/ghc910/default-config.golden.json
```diff
@@ -86,10 +86,12 @@
             "symbolsOn": true
         },
         "ghcide-type-lenses": {
+            "codeLensOn": true,
             "config": {
+                "localBindingInlayHintOn": true,
                 "mode": "always"
             },
-            "globalOn": true
+            "inlayHintsOn": true
```
cassandracomar, line 94: regenerate these

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
cassandracomar: this may also be clearer as a three-way enum called `localBindingHintsMode` with the options `NoLocalHints`, `UseInlayHints`, `UseCodeLenses`.

happy to let others bikeshed this, though.

michaelpj:
I wonder what to do about inlay hints vs code lenses. As I commented in the other issue, I do think we should prefer them if we can. But maybe we need a consistent approach. e.g. something like:
- a global "preferred in-buffer display mode: inlay hints/code lenses", default inlay-hints
- if it's set to inlay hints then use inlay hints if the client supports it otherwise fall back to code lenses

Out of scope for this PR though.

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
michaelpj: Yeah, sounds good, although I'm somewhat tempted towards a global option for the user preference here.

michaelpj:
Give a shout if you need help/want attention on any bit of it in particular!

cassandracomar:
> Give a shout if you need help/want attention on any bit of it in particular!

will do! haven't had a chance to go through the test failures.

PR #4613, opened 2025-06-04
Remove `hie-compat`

dschrempf:
After removing support for GHC 9.4, I was in the mood of cleaning up a bit more. While removing references to GHC 9.4, I realized that some parts of HLS are referring to even older versions of GHC.

Now, some people may hate me for that, but: I completely removed the now obsolete `hie-compat` library.

`hie-compat` is a compatibility library backporting support of Haskell IDE Engine (HIE) features to older versions of GHC. Since GHC 9.2, `hie-compat` only re-exported definitions already present in the `ghc` library, and so, is essentially obsolete.

FYI: We still have `hie-compat` in the dependency graph, because some libraries (e.g., `hiedb`) are using it.

This may be a bit harsh, but let me know what you think!


michaelpj:
Yeah, I think this is a good idea. In principle, I think we'd like to have some kind of compatibility layer library, but in practice I think it's just spread out everywhere.

dschrempf:
There is `Development.IDE.GHC.Compat`, which we could extract into a separate package. That would probably make sense.

dschrempf:
Any other opinions? (I don't feel like dragging this along for too long, but maybe there is a way forward?)

michaelpj:
I think @wz1000 is most likely to have a real opinion?

dschrempf:
It seems there are no more opinions. I sgguest reviewing/merging or closing this PR.

michaelpj:
I think we just need an opinion from @wz1000

dschrempf:
I think we should go ahead and rebase/review/merge or close.

wz1000:
I think this is fine, but we do lose a bit of flexibility by losing the ability to backport fixes and improvements to hie file generation. The format is more stable now, so perhaps that isn't such a big concern and we just require upgrading the ghc version to get access to newer hie file fixes.

dschrempf:
Thank you @wz1000 for your opinion. I have rebased the PR.

I think it is good to remove unnecessary compability layers. The only concern I have is that on the Hackage page of `hie-compat` (https://hackage.haskell.org/package/hie-compat) we link to this repository, and after merging this PR the relevant files will be gone.

michaelpj:
Nothing stops us reviving the package and publishing new versions in the future if we need, so let's do this for now.

dschrempf:
I am unsure what the two failing checks mean. Let me know, if I can do something about them! @michaelpj

PR #4616, opened 2025-06-07
Support hlint on 9.10 apart from apply-refact

michaelpj:
This enables the hlint plugin on GHC 9.10, at the cost of disabling refactoring actions. `apply-refact` is not even buildable on 9.10, so we have to push this all the way to the cabal file and use CPP, alas.

We have two lines of defense: we don't consider hints applicable if we don't have `apply-refact`, and if we somehow do get to trying to apply a hint, we fail.


fendor (approved):
LGTM

PR #4619, opened 2025-06-07
Fix completion for record dot syntax when record isn't known

samuel-williams-da:
The issue was that the prefix info extractor didn't check that the "modName" it pulls out is a valid module name, so would include cases when using record dot syntax.
This is only an issue when the record type beforehand's type isn't known , i.e. `undefined.myField`, as otherwise the incorrect completions were replaced with the field names directly.

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -912,7 +912,7 @@ getCompletionPrefixFromRope pos@(Position l c) ropetext =
           [] -> Nothing
           (x:xs) -> do
             let modParts = reverse $ filter (not .T.null) xs
-                modName = T.intercalate "." modParts
+                modName = if all (isUpper . T.head) modParts then T.intercalate "." modParts else ""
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -912,7 +912,7 @@ getCompletionPrefixFromRope pos@(Position l c) ropetext =
           [] -> Nothing
           (x:xs) -> do
             let modParts = reverse $ filter (not .T.null) xs
-                modName = T.intercalate "." modParts
+                modName = if all (isUpper . T.head) modParts then T.intercalate "." modParts else ""
```
michaelpj, line 883: Can you include some explanation for why this is the right thing to do? You explain it nicely in the PR description but it'll get lost if we don't write it in!

michaelpj (approved):
nice!

ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
samuel-williams-da: Ah yes of course, will do!
Currently fighting CI :sweat_smile:

fendor (approved):

PR #4621, opened 2025-06-08
Change tracking of file types to language kinds

### LINKED ISSUE #3567: HLS not working for file without .hs extension
Heya,

I've searched a bit but couldn't find this mentioned before, but I found it pretty hard to search for terms like "extension", ".hs", etc., so please forgive me if this has already been brought up.

I'm using haskell to write a [./go script](https://www.thoughtworks.com/insights/blog/praise-go-script-part-i). In a nutshell, it fills the same role as makefiles in projects that don't use it's only-rebuild-what's-neccessary functionality, or npm's package.json `scripts` stuff.

I wanted to call it literally `do`, but then I realized that HSL doesn't work.
I don't have a cabal file or anything since this is not even a haskell project, I'm using a nix devshell to get ghc with the packages I need.

When I rename the file to `do.hs`, everything works as it should.

While my use case might be better solved by a simpler approach anyways, this should work, right? Looking into :LspLog, it seems like HLS is running but doesn't care about the file. It didn't even print anything about hie-bios or sth.

### Your environment

nvim + lspconfig 
`haskell-language-server version: 1.10.0.0 (GHC: 9.2.7) (PATH: /nix/store/cqcm8dqw7zjj363q8z0mphg17bnxrrn1-haskell-language-server-1.10.0.0/bin/haskell-language-server-wrapper)`

Which OS do you use?
Arch linux, but everything is done via nix

I tried adding a hie.yaml with different content but nothing changed

If this is actually debugging-worthy, I'd be happy to trim down my stuff and upload it somewhere, reproducing should be straightforward with nix.

VeryMilkyJoe:
The plugin descriptor now tracks the language kinds it is responsible for instead of the file endings.
This will allow files without ".hs" or other relevant file endings to still be handled by HLS if their language kind is set properly.

We get the language kinds of any file from the VFS.

Currently we are using a source repository to be able to use the lsp changes needed, but once lsp is released this can be removed.

Closes #3567

image

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -6,8 +6,7 @@ packages:
          ./hls-plugin-api
          ./hls-test-utils
 
-
-index-state: 2026-02-02T09:03:26Z
+index-state: 2026-02-24T00:00:00Z
 
 tests: True
 test-show-details: direct
@@ -41,6 +40,7 @@ constraints:
   -- We want to be able to benefit from the performance optimisations
   -- in the future, thus: TODO: remove this flag.
   bitvec -simd,
+  monad-control >=1.0.3,  
 
 
 -- Some of the formatters need the latest Cabal-syntax version,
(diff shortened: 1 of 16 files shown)


Bodigrim:
@VeryMilkyJoe please upgrade to the latest commit of `cabal-add`, https://github.com/Bodigrim/cabal-add/commit/d831980e1d74af4c6a98d057cb980dad526dcfe2. Once you confirm that it works well, I'll make a release.

PR #4307, opened 2024-06-09
Switch ghcide tests to sequential execution

dyniec:
Should address  #3245

PR #4623, opened 2025-06-08
Revert "Switch ghcide tests to sequential execution (#4307)"

fendor:
This reverts commit 349ff6e18cbd35a502e05780f0bc6bca4bbc1074.

I don't know why, but this change breaks the usage of `TASTY_PATTERN` and `-p` for filtering by pattern.


michaelpj:
Ah yes, I've also seen this. I think tasty interprets this as "the tests absolutely must be run in this order", which also implies "you can't skip any"

michaelpj:
Although I made an issue on tasty to check if it's intentional.

PR #4625, opened 2025-06-08
Start using structured diagnostics for missing signatures

janvogt:


flake.nix
```diff
@@ -67,6 +69,7 @@
             # Compiler toolchain
             hpkgs.ghc
             pkgs.haskellPackages.cabal-install
+            pkgs.haskellPackages.haskell-language-server
```
jian-lin: Not sure if changes to flake should be merged as part of this PR, but HLS has to match the GHC version

```suggestion
            hpkgs.haskell-language-server
```

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -200,22 +208,28 @@ commandHandler _ideState _ wedit = do
...
 
-isGlobalDiagnostic :: Diagnostic -> Bool
-isGlobalDiagnostic Diagnostic{_message} = _message =~ ("(Top-level binding|Pattern synonym) with no type signature" :: T.Text)
+isGlobalDiagnostic :: FileDiagnostic -> Bool
+isGlobalDiagnostic diag = diag ^? fdStructuredMessageL
+                                  . _SomeStructuredMessage
+                                  . msgEnvelopeErrorL
+                                  .  _TcRnMessage
+                                  . to stripTcRnMessageContext
+                                  . _TcRnMissingSignature
```
michaelpj, line 230: nice

michaelpj (approved):

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -200,22 +208,28 @@ commandHandler _ideState _ wedit = do
...
   | otherwise = Nothing
 
-isGlobalDiagnostic :: Diagnostic -> Bool
-isGlobalDiagnostic Diagnostic{_message} = _message =~ ("(Top-level binding|Pattern synonym) with no type signature" :: T.Text)
+isGlobalDiagnostic :: FileDiagnostic -> Bool
+isGlobalDiagnostic diag = diag ^? fdStructuredMessageL
+                                  . _SomeStructuredMessage
+                                  . msgEnvelopeErrorL
+                                  .  _TcRnMessage
+                                  . to stripTcRnMessageContext
```
fendor:
```suggestion
```

Already done by `_TcRnMessage`, right?

flake.nix
janvogt: Agreed! And whether to merge or not is a subjective decision. I put it in its own commit to easily take it out if it's not wanted. Who does have final say regarding this?

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
janvogt: Yes! 🙈

flake.nix
jian-lin: I personally also add hls to the dev shell.

I guess one may not want to add it to avoid potential confusion with the locally built HLS.

janvogt: I thought that's the reason for not including it, then I realised to that you usually wan't the hls you actively working on to run in an unrelated test project and probably not on his itself. However, there IS probably a reason why it's not included. The question to me is WHO knows it?

janvogt:
Ok, I guess, the CI is now as green as it gets 😅

PR #4626, opened 2025-06-08
Implement signature help

### LINKED ISSUE #3598: Show function signature while providing the arguments
**Is your feature request related to a problem? Please describe.**

When adding a function call, the extension suggests a list of functions with signatures, but when a function is selected and you start providing the parameters, the signature is not visible anymore and you need to bring the signature back sometimes for more complex functions.

It'd be nice to keep the tooltip with the function signature open till you finish the call and make the current parameter bold there. It seems that the VS code extension for Idris has such feature, so maybe it'd be possible to add it here also?

Please see the Idris solution here - https://github.com/zjhmale/vscode-idris/commit/fd424291c73a03b61822603df65bf08bc945f00b

jian-lin:
## Introduction

Haskell language server (HLS) implements the server part of [Language Server Protocol][ilsp] (LSP) to improve user experience of reading and writing Haskell. [Signature help][ish] is a LSP [language feature][ilf] which shows function signatures and documentation when the cursor is inside a function application. It can also highlight part of the signature related to the current parameter the cursor is at. This is a useful feature and users [want][iw] it.

This PR aims to implement the signature help feature for HLS.

[ilsp]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/
[ish]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp
[ilf]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#languageFeatures
[iw]: https://github.com/haskell/haskell-language-server/issues/3598

## My work

I have implemented most parts of the signature help feature for HLS in this PR.  The last commit during GSoC is d0200a4c38b457c25e40e75c99dc63589090409c.

The signature help feature works well in most cases, especially when a user is reading code.

<details> <summary>Click for a video demo</summary>

[hls-signature-help-gsoc-demo.webm](https://github.com/user-attachments/assets/014d278d-2565-4c4c-b417-b6263bbe6ebb)

</details>

When I worked on this PR, a bug of other parts of HLS was found and I created a PR to fix it.

- https://github.com/haskell/haskell-language-server/pull/4717

Besides HLS proper, I also created 2 PRs for `lsp-test` to add a needed function for tests.

- https://github.com/haskell/lsp/pull/621
- https://github.com/haskell/lsp/pull/623

<details> <summary>Click for more detailed work</summary>

### 2025-06-02 - 2025-06-08

- Add basic boilerplate for signature help plugin
  - commit: 7a54a1dc5d5c360eb57a7c0f6e84f60349cf49ba
  - <details> <summary>click for screenshot</summary>
  
    image
    
    </details>
  
### 2025-06-09 - 2025-07-13

- Finish signature help plugin MVP: show function signature and highlight the current parameter
  - commit: 9168b740867611439c0b030d1c07e209e8489ace
  - <details> <summary>click for video demo</summary>
  
    [Screencast From 2025-07-13 02-11-44.webm](https://github.com/user-attachments/assets/95329f0d-f5a2-475c-87a6-628c12532056)

    </details>
  
### 2025-07-14 - 2025-07-20

- Add basic tests.  Some of them are not passed for now.
  - commit: 62fbccf409c01e79ba96dd1dc853cf6d02ca5606

### 2025-07-21 - 2025-07-27

- Change expected test results considering the cursor shape (a6635caca8328fa3b7e523e9e1a97a1241f57b86)
- Replace `maybe` with case for better readability (bf0b4d529e9e4e616ed03255bf90b7a43bb636ab)
- Call extractInfoFromSmallestContainingFunctionApplicationAst once (c95d6e4cf364afd087c7672f077634ce8145e5c6)

### 2025-07-28 - 2025-08-03

- Show more types: each type as one signature help (d603ec4a3008983bf5f70e6ff63a30d9d8387c06)
- Add tests for kind signatures and higher-order function (dca1311cf345d7ba9d79bc499fe906f0afad0a89) (471958ffe18d278e9c76117cf32c4da6f3e2b6f3)

### 2025-08-04 - 2025-08-10

- Use structured type and type string to generate signature help (9bea0e37a7a32c091d1f49a926ab65c69bee64f8)
- Add more tests
  - middle => (fdd5acd4c1e28dfb29c7bfea0f356548c7e7a118)
  - => in argument (c6ece42acc555eeba7fcd000684e51bca8852821)
  - RankNTypes(forall in middle) (fe2d61822df27e364e86374c228fc5c75c0e6147 and 081ea8fb4c0dc5f40d415eae6ad033c1bc8bdaee)
  - LinearTypes (c3224d7b644ca54cb4e9ae532dbdda97d4cfe72f)

### 2025-08-11 - 2025-08-17

- Show function documentation in signature help (a522e88591b63ea296d5ded9f2cc5936429177ea)
- Show function **argument** documentation in signature help (d826d062c4234abc41c271950204904b9aab5755)
- Do not error if doc is not available (35399e71f8c986b53d7ef83f4c0a1f6309aa634a)
- <details> <summary>click for screenshots</summary>
  
  image

  image
  </details>

### 2025-08-18 - 2025-08-24

- Upstream `getSignatureHelp` to `lsp-test`: https://github.com/haskell/lsp/pull/621
- Make signature helps reproducible in tests (eeeb28386a664859b897b0744f25e92b620257ac)
- Do not show uris in the argument documentation (f1d19ba58abe022b587f6f82354ed9d573fbd290)
- Add tests
  - imported function with argument doc (433a8ad47cceaf6ff326cdcb09eaf8abaa1bfc96)
  - imported function with no doc (d3d7d12757ac5fd0707e7d0d5581b2a240154760)
- Remember previous active signature (e6702b780c57e5e013797b761bc2195bd6f27341)
- Show signature help even when there are type applications (800f908e42e819a65f64192f1457da021cf572d0)
- Fix tests for ghc > 9.8
  - changed doc (cb844809db0f14ee3a8014608c3093385f0ff33f)
  - `Any` becomes `ZonkAny 0` (c7931a284933c566256e6eb96ea7ff311ccc82a2)
- Add myself as codeowner of hls-signature-help-plugin (b6d4617b867f0b7a4aee968f3354f64f864fca06)
- Add hls-signature-help-plugin to some documentation files (9a1de4e990a6365b423bb2efe7c464edc2c2271f)

### 2025-08-25 - 2025-08-31

- https://github.com/haskell/lsp/pull/623
- Use T.isInfixOf as equality check for doc string (5925c1c0bd27a4a90b249f47b9d4c074d468bd9e)
- Replace string interpolation lib in signature help tests (9c972384eeb584757f131edfd436004a46e167b2)
- Mark broken tests as expected failures (195db1855382935c9c3a3d943aafd523e671a8ac)
- Mention signature help plugin in configuration.md (8d6b1bb0d53558479d08bc9e3f7e8b53527344b1)
- Cleanup
  - Remove outdated TODO items (f5a9858562b7100b384d9cb7f039c20b50e03c5b)
  - Format with ormolu (9c87ef4bab9fec2f002eb1202c695e1e006a04bf)
  - Unify argument and parameter (757d4a585d18aa5ca16e92c28579b0e75beb75a2)
- #4717

### 2025-09-01 - 2025-09-02

- Use record syntax for data with many fields (b5c13a3a0af45d722bdc169d9658976627a24d52)
</details>

## Future work

- Fix [a known bug][todo1] of wrong highlight in the signature in some cases.
- Show signature and highlight the related part of the next parameter when a user is [writing code][todo2].
- [Highlight][todo3] the type variable of `forall` when the cursor is at a type application such as `@Int`.
- Support `OpApp` (operator application) of `HsApp` such as `f x $ y|`.
- Trigger a signature help request after a space char is typed.

[todo1]: https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2261133076
[todo2]: https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2298808232
[todo3]: https://github.com/haskell/haskell-language-server/pull/4626#issuecomment-3172723975

## Acknowledgements

This is a Google Summer of Code (GSoC) project.  Thanks to all the people involved.  Special thanks to my mentors, @michaelpj and @fendor, for their help.

---

Closes #3598

Related to #2348 since we make [mkDocMap](https://github.com/haskell/haskell-language-server/blob/e3d38b0c4666681dca39901aa9d47b04422c82ff/ghcide/src/Development/IDE/Spans/Documentation.hs#L43) expose the argument doc map


hls-plugin-api/src/Ide/Types.hs
```diff
@@ -764,6 +770,10 @@ instance PluginRequestMethod Method_TextDocumentDocumentSymbol where
             si = SymbolInformation name' (ds ^. L.kind) Nothing parent (ds ^. L.deprecated) loc
         in [si] <> children'
 
+-- TODO(@linj) is this correct?
```
michaelpj: I think this is reasonable. Really `combineResponses` should have a way to return an error. There are a bunch of methods where it really only makes sense if we have one handler.

We _could_ try to combine responses: we would combine the signatures, and so long as only one of them indicated an active signature it would be okay. But that's a bit sketchy and I doubt we'll have several anyway!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+signatureHelpProvider :: PluginMethodHandler IdeState Method_TextDocumentSignatureHelp
+signatureHelpProvider ideState _pluginId (SignatureHelpParams (TextDocumentIdentifier uri) position _mProgreeToken _mContext) = do
+    nfp <- getNormalizedFilePathE uri
+    mResult <- runIdeActionE "signatureHelp" (shakeExtras ideState) $ do
+        -- TODO(@linj) why HAR {hieAst} may have more than one AST?
+        (HAR {hieAst, hieKind}, positionMapping) <- useWithStaleFastE GetHieAst nfp
+        case fromCurrentPosition positionMapping position of
+            Nothing -> pure Nothing
+            Just oldPosition -> do
+                let functionName =
```
michaelpj: you probably want to do all these in one go, this looks like it will traverse the AST three times! I think you can just do one call to `extractInfoFromSmallestContainingFunctionApplicationAst` and get all the pieces together though?

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+    where
+        extractName = rightToMaybe
+
+-- TODO(@linj) share code with getNodeName
+getNodeType :: HieKind a -> RealSrcSpan -> HieAST a -> Maybe Text
+getNodeType (hieKind :: HieKind a) _span hieAst =
+    if nodeHasAnnotation ("HsVar", "HsExpr") hieAst
+       then
+           case M.elems $ M.filter isUse $ getSourceNodeIds hieAst of
+               [identifierDetails] -> identType identifierDetails >>= (prettyType >>> Just)
```
michaelpj: I suggest not printing the type here and instead returning the type itself. That would mean that when you're doing `mkArguments` you can pattern match on the type itself, not on the string representation. Then you don't need to worry about e.g. whether it prints on multiple lines or whatever.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+                  (functionNameLabelPrefix <> functionType)
+                  Nothing
+                  (Just $ mkArguments (fromIntegral $ T.length functionNameLabelPrefix) functionType)
+                  (Just $ InL argumentNumber)
+            ]
+            (Just 0)
+            (Just $ InL argumentNumber)
+
+-- TODO(@linj) can type string be a multi-line string?
+mkArguments :: UInt -> Text -> [ParameterInformation]
```
michaelpj: I think this would be a lot simpler if you were just looking at the type itself!

michaelpj: Although hmm, the offsets are into the string :scream: So maybe you would need to incrementally build up the `ParameterInformation`s and the overall `label`, but that would probably still be nicer?

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+    where
+        positionToSpan hiePath position =
+            let loc = mkLoc hiePath position in mkRealSrcSpan loc loc
+        mkLoc (LexicalFastString hiePath) (Position line character) =
+            mkRealSrcLoc hiePath (fromIntegral line + 1) (fromIntegral character + 1)
+
+type Annotation = (FastStringCompat, FastStringCompat)
+
+nodeHasAnnotation :: Annotation -> HieAST a -> Bool
+nodeHasAnnotation annotation = sourceNodeInfo >>> maybe False (isAnnotationInNodeInfo annotation)
```
michaelpj: nit: I think it tends to be easier to read a simple case expression than one of the pattern-matching functions like `maybe`. I usually only use `maybe` if it's partially-applied.

You can still avoid naming the `HieAst` parameter if you want using `LambdaCase`.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+
+-- TODO(@linj): the left most node may not be the function node. example: (if True then f else g) x
```
michaelpj: write a test!

michaelpj: this is a great example btw. In this case I think we would have to give no signature help, I think, since it's only dynamically known which function we're getting.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+    case nodeChildren thisNode of
+        []           -> thisNode
+        leftChild: _ -> getLeftMostNode leftChild
+
+getNodeName :: RealSrcSpan -> HieAST a -> Maybe Name
+getNodeName _span hieAst =
+    if nodeHasAnnotation ("HsVar", "HsExpr") hieAst
+        then
+            case mapMaybe extractName $ M.keys $ M.filter isUse $ getSourceNodeIds hieAst of
+                [name] -> Just name -- TODO(@linj) will there be more than one name?
```
michaelpj: I'm not 100% sure when/if this can happen. I think it might happen if we're looking at a typeclass method, in which case we could get the generic typeclass method and the specific one perhaps? That would be a good use-case for multilple signatures, e.g. if we have 

```
foldMap @[] f ^
```
then we might want to see both the generic `Foldable` signature and the one for the list instance.

michaelpj: I think my general suggestion would be: let's treat possibly-multiple node annotations as producing possibly-multiple signatures.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,208 @@
...
+    let functionNameLabelPrefix = printOutputable (ppr functionName) <> " :: "
+     in SignatureHelp
+            [ SignatureInformation
+                  (functionNameLabelPrefix <> functionType)
+                  Nothing
+                  (Just $ mkArguments (fromIntegral $ T.length functionNameLabelPrefix) functionType)
+                  (Just $ InL argumentNumber)
+            ]
+            (Just 0)
+            (Just $ InL argumentNumber)
```
michaelpj: do we know if this is 0 or 1-indexed? specification doesn't say (of course :roll_eyes: ) so I guess we'll have to test it in vscode to see

michaelpj (commented):
I think really worth trying to start getting some tests in place!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: When I developed this, I did experiments using Emacs eglot (LSP client) and rust-analyzer (LSP server) and the result is 0-indexed.  I did not verify that on Vscode since there is no negotiation about 0-indexed and 1-indexed in the LSP spec.

I just verified that Vscode also uses 0-indexed.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Good idea.  Fixed in c95d6e4cf364afd087c7672f077634ce8145e5c6.

In addition, I merge `getNodeName` and `getNodeType` into one function `getNodeNameAndTypes` in d603ec4a3008983bf5f70e6ff63a30d9d8387c06, which should improve the performance a bit.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Sounds good.  Changed in bf0b4d529e9e4e616ed03255bf90b7a43bb636ab.  I think my old preference for `maybe` over `case` is taught by hlint's "Replace case with maybe" [rule](https://github.com/ndmitchell/hlint/blob/master/hints.md).

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Implemented in d603ec4a3008983bf5f70e6ff63a30d9d8387c06.

Related tests:
- type constraint
- type constraint with kind signatures
- multi-line type with type constraint

---

> I'm not 100% sure when/if this can happen.

I do not have a principled conclusion.  But from my experiments, the summary is that
- There are multiple types iif there are type variables.
- There seems to be only one `Name`.

jian-lin:
All tests on Linux and darwin passed! Not sure what happens on windows.

---

Added a few more tests and now `3 out of 87 tests failed`.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin, line 1: a bit offtopic: Is 2-space indentation instead of 4 allowed for new code?

The [.editorconfig][1] says 4.

However, some newly added plugins use 2:
- [hls-cabal-gild-plugin][2] by @fendor
- [hls-stan-plugin][3]

[1]: https://github.com/haskell/haskell-language-server/blob/cfeced8fa3088be7af03e8794ff6504ed0fed0a0/.editorconfig#L13
[2]: https://github.com/haskell/haskell-language-server/pull/4101
[3]: https://github.com/haskell/haskell-language-server/pull/3851

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,263 @@
...
+  g :: forall a b. Eq a => a -> Num b => b -> b
+- ppr the first argument type of '(a -> b) -> a -> b' is 'a -> b' (no parentheses)
+- 'forall' is not always shown
+
+Alternative method 2: use only unstructured type string
+This method is hard to implement because we need to parse the type string.
+Some tricky cases are as follows:
+- h :: forall a (m :: Type -> Type). Monad m => a -> m a
+-}
+findArgumentStringRanges :: UInt -> Text -> [Text] -> [(UInt, UInt)]
```
jian-lin: Context:
- https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2203250834
- https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2203251467
- https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2203252268

@michaelpj 

I implemented a mixed way which uses both structured type and type string.  See the comment for the reason of doing it this way instead of using only structured type or only type string.

It seems to work pretty well.  "3 out of 87 tests failed".

---

One failed test case is `f :: Integer -> Num Integer => Integer -> Integer`.  We matched `Num Integer` as the first argument.  This probably can be fixed by using regex.

Another similar failed test case is `f :: forall l. l -> forall a. a -> a`.  When we should highlight the argument `l`, we highlight the `l` for the second `forall`.

---

Here is another failed test case.
```haskell
f :: a -> forall a. a -> a
f = _
```
The printed type string for `f` is `f :: forall a. a -> forall a1. a1 -> a1`.  This seems tricky to fix in the current implementation because it needs us to duplicate the renaming logic (the second `forall a` becomes `forall a1`) of `ppr`.

This case only happens when `RankNTypes` is used so I do not think it is very common.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: We have one related test case for this now: the `dynamic function` test.

This test is not a unit test for this `getLeftMostNode` function.  This test is an integration testing for the signature help plugin.

```
f :: Int -> Int -> Int
f = _
g :: Int -> Int -> Int
g = _
x = (if _ then f else g) 1 2
^^ ^^^ ^  ^^^ ^  ^^^^^^^^
```

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: See https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2261133076

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: See https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2261133076

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
fendor: Follow the `.editorconfig` :)
`hls-cabal-gild-plugin` was copy-pasted from the `hls-cabal-fmt-plugin`

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
michaelpj: I'm wondering how bad it would be to just use the structured type and re-implement some pretty-printing logic. I think the only bit we'd need to duplicate would be the printing of `->`, `=>`, and `forall`? Maybe that's okay?

michaelpj:
Another thing we should think about: type arguments.
```
id :: forall a . a -> a
id x = x

id @...
      ^^ --- we probably want to consider the "a" a parameter so we can highlight it here when the user is providing a type

-- with RequiredTypeArguments
id :: forall a -> a -> a
id _ x = x

id 
    ^^ -- here again we need to highlight the forall
```

Type arguments are kind of annoying because they might be optional (with normal foralls), or they might be required (with required type arguments).

plugins/hls-signature-help-plugin/test/Main.hs
```diff
@@ -0,0 +1,358 @@
...
+                  ]
+            ]
+
+mkTest :: TestName -> Text -> [Maybe SignatureHelp] -> TestTree
+mkTest name sourceCode expectedSignatureHelps =
+    parameterisedCursorTest
+        name
+        sourceCode
+        expectedSignatureHelps
+        getSignatureHelpFromSession
```
jian-lin: @michaelpj suggests adding tests where the function is not defined in this module but is imported from elsewhere.

We already have one test case for that: `type constraint with kind signatures`.  The problem is that there are local file paths in the doc string, which makes the test fragile.

For example, the doc string of `pure` on my system is:
```
"\n\nLift a value.\n\n[Documentation](file:///nix/store/1cyav1d9cq6dgffj7y6v6zvr1qcy78pf-ghc-9.8.4-doc/share/doc/ghc/html/libraries/base-4.19.2.0-9ef4/GHC-Base.html#v:pure)\n\n[Source](file:///nix/store/1cyav1d9cq6dgffj7y6v6zvr1qcy78pf-ghc-9.8.4-doc/share/doc/ghc/html/libraries/base-4.19.2.0-9ef4/src/GHC.Base.html#pure)\n\n"
```

One solution is to let `parameterisedCursorTest` support custom "equality" which is passed as a function `(a -> a -> Bool)`.  Any ideas @fendor?

ghcide/src/Development/IDE/Spans/Common.hs
```diff
@@ -29,13 +30,15 @@ import           GHC.Generics
...
 import           Control.Lens
+import           Data.IntMap                  (IntMap)
 import           Development.IDE.GHC.Compat
 import           Development.IDE.GHC.Orphans  ()
 import qualified Language.LSP.Protocol.Lens   as JL
 import           Language.LSP.Protocol.Types
 
 type DocMap = NameEnv SpanDoc
 type TyThingMap = NameEnv TyThing
+type ArgDocMap = NameEnv (IntMap SpanDoc)
```
jian-lin, line 41: Instead of creating a new `ArgDocMap`, an alternative implementation is to extend `DocMap` to `NameEnv (SpanDoc, IntMap SpanDoc)`.

michaelpj: That sounds sensible to me? Any reason not to do that?

jian-lin: I do not have a strong opinion on this.  I will change the code to `type DocMap = NameEnv (SpanDoc, IntMap SpanDoc)`.

michaelpj: Looking at the code it seems like it would simplifiy things quite a lot.

plugins/hls-signature-help-plugin/test/Main.hs
michaelpj: We could also pre-process the signature helps to get rid of such things. e.g. write a function `tidy :: SignatureHelp -> SignatureHelp` that goes through all the documentation strings and replaces any `file://` URLs with the empty string, or similar?

ghcide/src/Development/IDE/Spans/Common.hs
jian-lin: Now I have second thoughts about this because the alternative implementation `type DocMap = NameEnv (SpanDoc, IntMap SpanDoc)` is semantically incorrect.  The semantically correct version should be `type DocMap = NameEnv (Maybe SpanDoc, IntMap SpanDoc)` or `type DocMap = NameEnv (Maybe SpanDoc, Maybe (IntMap SpanDoc))` because a `Name` can have only function doc or arg doc.

I prefer to keep the old implementation, i.e., introducing a new `ArgDocMap`, since the semantically correct version of the alternative implementation seems more complicated.

plugins/hls-signature-help-plugin/test/Main.hs
jian-lin: Good idea!  Implemented in eeeb28386a664859b897b0744f25e92b620257ac

plugins/hls-signature-help-plugin/test/Main.hs
jian-lin, line 1: I am a bit confused about this failed test in CI which I cannot reproduce locally.

In CI, the function doc of `pure` is `"\n\nLift a value.\n\n"`.  Locally, it is `"\n\nLift a value.\n\n\\[Documentation\\]\\(file://.*\\)\n\n\\[Source\\]\\(file://.*\\)\n\n"`.  So somehow in CI HLS fails to find uris of the **imported** function `pure`.

```
type constraint with kind signatures
    type constraint with kind signatures 2:          FAIL (0.15s)
      src/Test/Hls.hs:400:
      Full Line:       "x = pure True"
      Cursor Column:   "         ^"
      Prefix Text:     ""
      
      expected: Just (SignatureHelp {_signatures = [SignatureInformation {_label = "pure :: forall (f :: Type -> Type) a. Applicative f => a -> f a", _documentation = Just (InR (MarkupContent {_kind = MarkupKind_Markdown, _value = "\n\nLift a value.\n\n\\[Documentation\\]\\(file://.*\\)\n\n\\[Source\\]\\(file://.*\\)\n\n"})), _parameters = Just [ParameterInformation {_label = InR (55,56), _documentation = Nothing}], _activeParameter = Just (InL 0)},SignatureInformation {_label = "pure :: Bool -> IO Bool", _documentation = Just (InR (MarkupContent {_kind = MarkupKind_Markdown, _value = "\n\nLift a value.\n\n\\[Documentation\\]\\(file://.*\\)\n\n\\[Source\\]\\(file://.*\\)\n\n"})), _parameters = Just [ParameterInformation {_label = InR (8,12), _documentation = Nothing}], _activeParameter = Just (InL 0)},SignatureInformation {_label = "pure :: forall a. a -> IO a", _documentation = Just (InR (MarkupContent {_kind = MarkupKind_Markdown, _value = "\n\nLift a value.\n\n\\[Documentation\\]\\(file://.*\\)\n\n\\[Source\\]\\(file://.*\\)\n\n"})), _parameters = Just [ParameterInformation {_label = InR (18,19), _documentation = Nothing}], _activeParameter = Just (InL 0)}], _activeSignature = Just 0, _activeParameter = Just (InL 0)})
       but got: Just (SignatureHelp {_signatures = [SignatureInformation {_label = "pure :: forall (f :: Type -> Type) a. Applicative f => a -> f a", _documentation = Just (InR (MarkupContent {_kind = MarkupKind_Markdown, _value = "\n\nLift a value.\n\n"})), _parameters = Just [ParameterInformation {_label = InR (55,56), _documentation = Nothing}], _activeParameter = Just (InL 0)},SignatureInformation {_label = "pure :: Bool -> IO Bool", _documentation = Just (InR (MarkupContent {_kind = MarkupKind_Markdown, _value = "\n\nLift a value.\n\n"})), _parameters = Just [ParameterInformation {_label = InR (8,12), _documentation = Nothing}], _activeParameter = Just (InL 0)},SignatureInformation {_label = "pure :: forall a. a -> IO a", _documentation = Just (InR (MarkupContent {_kind = MarkupKind_Markdown, _value = "\n\nLift a value.\n\n"})), _parameters = Just [ParameterInformation {_label = InR (18,19), _documentation = Nothing}], _activeParameter = Just (InL 0)}], _activeSignature = Just 0, _activeParameter = Just (InL 0)})
      Use -p '/type constraint with kind signatures 2/' to rerun this test only.
```

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin, line 1: TODO: different hightlighting strategies for "reading code" and "writing code".

By "writing code", I mean the cursor is at the end of the function application and we intend to write more args to that function application.

Other cases are "reading code".

---

Currently, the highlight strategy for "reading code" works well.

For example, `f x | y` will highlight `y`.  Note that in this case the cursor `|` is in the subtree of a `HsApp` AST node.

---

Currently, the highlight strategy for "writing code" is problematic.

Case 1: only the function itself is written `f |`.  We want to highlight the first arg of `f`.  Note that in this case there is no `HsApp` node and so the cursor `|` is not in the subtree of any `HsApp` AST node.

Case 2: the function itself and at least 1 arg have been written `f x |` (`f :: a -> b -> c`).  We want to highlight the second arg of `f`.  Note that in this case there are `HsApp` nodes but the cursor `|` is not in the subtree of any `HsApp` AST node.

plugins/hls-signature-help-plugin/test/Main.hs
jian-lin: Fixed in 5925c1c0bd27a4a90b249f47b9d4c074d468bd9e

plugins/hls-signature-help-plugin/test/Main.hs
jian-lin: I decided to use `T.isInfixOf` as equality check following fendor's suggestion.  Implemented in 5925c1c0bd27a4a90b249f47b9d4c074d468bd9e.

See also https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2292563051

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
michaelpj: It sounds to me like your two strategies are the same? In both cases the strategy is "highlight the next argument after the cursor position (which may not exist)"?

Presumably in this situation we will highlight the second argument, though:

```
f one t|wo
```

plugins/hls-signature-help-plugin/test/Main.hs
jian-lin, line 1: A lot of tests fail on Windows due to different newlines.  This is found by @fendor, thank you!

Let me explain this in a bit more detail.  By default, git files uses `\r\n` as newline on Windows.  The [string interpolation lib][1] we use [follows the newline of the file][2], so it also uses `\r\n` on Windows.  However, `T.lines` always uses `\n` as newline.

I think there are 3 possible fixes, from most preferred to least preferred.

1. Add a `.gitattributes` file with this content `*.hs text eol=lf` in the repo to make git files always use `\n` as newline.
2. Use another string interpolation lib.  I have tested [string-interpolate][3] and it works well.  This lib uses `#` to interpolate so we have to escape `{-# LANGUAGE LinearTypes #-}` to `{-\# LANGUAGE LinearTypes \#-}`, which is a bit unfortunate.  So maybe we should choose yet another lib?
3. Add code to handle the newline difference.  I think the most correct patch is replacing `\r\n` with `\n` before passing the string to `T.lines`.

<details> <summary>My patch</summary>

```patch
 extractCursorPositions :: T.Text -> (T.Text, [PosPrefixInfo])
 extractCursorPositions t =
     let
-        textLines = T.lines t
+        textLines = handleWindowsNewline <$> T.lines t
         foldState = List.foldl' go emptyFoldState textLines
         finalText = foldStateToText foldState
         reconstructCompletionPrefix pos = getCompletionPrefixFromRope pos (Rope.fromText finalText)
@@ -412,6 +412,9 @@ extractCursorPositions t =
         go foldState l = case T.indices "^" l of
             [] -> addTextLine foldState l
             xs -> List.foldl' addTextCursor foldState xs
+        handleWindowsNewline l = case T.stripSuffix "\r" l of
+            Nothing -> l
+            Just l' -> l'
```
</details>

<details> <summary>@fendor's patch</summary>

```patch
 -- | Produce the final file contents, without any lines containing cursor positions.
 foldStateToText :: FoldState -> T.Text
-foldStateToText state = T.unlines $ reverse $ foldStateFinalText state
+foldStateToText state = T.unlines $ fmap (T.filter (/= '\r')) $ reverse $ foldStateFinalText state
```
</details>

WDYT?

[1]: https://hackage.haskell.org/package/neat-interpolation
[2]: https://github.com/nikita-volkov/neat-interpolation/issues/14
[3]: https://hackage.haskell.org/package/string-interpolate

fendor: if the `__i` quasi quoter works just like `trimming`, then let's use `string-interpolate`!

jian-lin: Method 2 has been implemented in 9c972384eeb584757f131edfd436004a46e167b2.

To keep this PR simple and focused, neat-interpolation usage outside
of the signature help plugin is not changed.  I plan to change that in
a future PR.

jian-lin: When working on a followup PR to replace `neat-interpolation` treewide, I notice that the string has to be indented if it was not indented before.

For example, we have to change https://github.com/haskell/haskell-language-server/blob/bd586063b29c67111e227f66251e339b5db6ec91/plugins/hls-cabal-plugin/test/Context.hs#L219-L228

to 

```haskell
libraryStanzaData :: T.Text
libraryStanzaData = [__i|
  cabal-version:      3.0
  name:               simple-cabal
  library
      default-language: Haskell98
      build-depends:
  
  ma
|]
```

I think this change is acceptable.  WDYT?

---

The cause can be seen more clearly using this example in the repl:

```
λ> :{
λ| x1 :: Text
λ| x1 = [__i|
λ| line1
λ|   line2
λ| |]
λ| :}
λ> :{
λ| x2 :: Text
λ| x2 = [__i|
λ|   line1
λ|     line2
λ| |]
λ| :}
λ> x1
"line1\nline2"
λ> x2
"line1\n  line2"
```

jian-lin: I replace `neat-interpolation` outside of this plugin in https://github.com/haskell/haskell-language-server/pull/4717.

jian-lin:
I think this PR is in a good shape now.  All tests have passed. Please review.

ghcide/src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -249,9 +249,9 @@ type instance RuleResult GetHieAst = HieAstResult
 -- | A IntervalMap telling us what is in scope at each point
 type instance RuleResult GetBindings = Bindings
 
-data DocAndTyThingMap = DKMap {getDocMap :: !DocMap, getTyThingMap :: !TyThingMap}
+data DocAndTyThingMap = DKMap {getDocMap :: !DocMap, getTyThingMap :: !TyThingMap, getArgDocMap :: !ArgDocMap}
```
michaelpj: Let's add some haddock explaining what these things are, and that e.g. only functions (?) will have argument documentation.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+signatureHelpProvider :: PluginMethodHandler IdeState Method_TextDocumentSignatureHelp
+signatureHelpProvider ideState _pluginId (SignatureHelpParams (TextDocumentIdentifier uri) position _mProgreeToken mSignatureHelpContext) = do
+  nfp <- getNormalizedFilePathE uri
+  results <- runIdeActionE "signatureHelp.ast" (shakeExtras ideState) $ do
+    (HAR {hieAst, hieKind}, positionMapping) <- useWithStaleFastE GetHieAst nfp
+    case fromCurrentPosition positionMapping position of
+      Nothing -> pure []
+      Just oldPosition -> do
+        pure $
+          extractInfoFromSmallestContainingFunctionApplicationAst
```
michaelpj, line 119: We have a nice explanation of the approach for coming up with the parameter strings, but I think this part is also important and worth explaining? Especially since it's something we might want to try and improve later on.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+descriptor _recorder pluginId =
+  (defaultPluginDescriptor pluginId "Provides signature help of something callable")
+    { Ide.Types.pluginHandlers = mkPluginHandler SMethod_TextDocumentSignatureHelp signatureHelpProvider
+    }
+
+signatureHelpProvider :: PluginMethodHandler IdeState Method_TextDocumentSignatureHelp
+signatureHelpProvider ideState _pluginId (SignatureHelpParams (TextDocumentIdentifier uri) position _mProgreeToken mSignatureHelpContext) = do
+  nfp <- getNormalizedFilePathE uri
+  results <- runIdeActionE "signatureHelp.ast" (shakeExtras ideState) $ do
+    (HAR {hieAst, hieKind}, positionMapping) <- useWithStaleFastE GetHieAst nfp
```
michaelpj, line 114: Have you tried using this from an editor? I wonder if we can actually get away with using a stale HIE AST here. Because we're likely to be looking at a bit of the AST that has _just_ been edited, so if we get a stale one we could be missing e.g. the last argument that got typed. So maybe we actually want to insist on an up-to-date HIE AST?

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+                Just (functionName, functionTypes, parameterIndex)
+            )
+  (docMap, argDocMap) <- runIdeActionE "signatureHelp.docMap" (shakeExtras ideState) $ do
+    mResult <- ExceptT $ Right <$> useWithStaleFast GetDocMap nfp
+    case mResult of
+      Just (DKMap docMap _tyThingMap argDocMap, _positionMapping) -> pure (docMap, argDocMap)
+      Nothing -> pure (mempty, mempty)
+  case results of
+    [(_functionName, [], _parameterIndex)] -> pure $ InR Null
+    [(functionName, functionTypes, parameterIndex)] ->
```
michaelpj, line 136: It's a little unclear to me at what point we exclude things that don't look like functions?

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+    activeSignature
+    (Just $ InL parameterIndex)
+  where
+    activeSignature = case mSignatureHelpContext of
+      Just
+        ( SignatureHelpContext
+            _triggerKind
+            _triggerCharacter
+            True
+            (Just (SignatureHelp _signatures oldActivateSignature _activeParameter))
```
michaelpj: I guess it's fine for us to ignore the active parameter. I don't know if this is something that the user can control? Maybe we should try it in VSCode? Perhaps the user can switch between the parameters, and if they've done that we should leave the active parameter as what they selected? Unclear.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+mkSignatureInformation docMap argDocMap parameterIndex functionName functionType =
+  let functionNameLabelPrefix = printOutputableOneLine (ppr functionName) <> " :: "
+      mFunctionDoc = case lookupNameEnv docMap functionName of
+        Nothing      -> Nothing
+        Just spanDoc -> Just $ InR $ mkMarkdownDoc spanDoc
+      thisArgDocMap = case lookupNameEnv argDocMap functionName of
+        Nothing             -> mempty
+        Just thisArgDocMap' -> thisArgDocMap'
+   in SignatureInformation
+        (functionNameLabelPrefix <> printOutputableOneLine functionType)
```
michaelpj: Do we definitely want to do this on one line? Did we experiment with splitting it? some function signatures can be very long!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+      Nothing -> pure (mempty, mempty)
+  case results of
+    [(_functionName, [], _parameterIndex)] -> pure $ InR Null
+    [(functionName, functionTypes, parameterIndex)] ->
+      pure $ InL $ mkSignatureHelp mSignatureHelpContext docMap argDocMap (fromIntegral parameterIndex - 1) functionName functionTypes
+    _ -> pure $ InR Null
+
+mkSignatureHelp :: Maybe SignatureHelpContext -> DocMap -> ArgDocMap -> UInt -> Name -> [Type] -> SignatureHelp
+mkSignatureHelp mSignatureHelpContext docMap argDocMap parameterIndex functionName functionTypes =
+  SignatureHelp
```
michaelpj, line 142: Probably worth using record constructor syntax here: it's not always clear what these fields are!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+    (mkSignatureInformation docMap argDocMap parameterIndex functionName <$> functionTypes)
+    activeSignature
+    (Just $ InL parameterIndex)
+  where
+    activeSignature = case mSignatureHelpContext of
+      Just
+        ( SignatureHelpContext
```
michaelpj, line 150: here also

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+mkSignatureInformation :: DocMap -> ArgDocMap -> UInt -> Name -> Type -> SignatureInformation
+mkSignatureInformation docMap argDocMap parameterIndex functionName functionType =
+  let functionNameLabelPrefix = printOutputableOneLine (ppr functionName) <> " :: "
+      mFunctionDoc = case lookupNameEnv docMap functionName of
+        Nothing      -> Nothing
+        Just spanDoc -> Just $ InR $ mkMarkdownDoc spanDoc
+      thisArgDocMap = case lookupNameEnv argDocMap functionName of
+        Nothing             -> mempty
+        Just thisArgDocMap' -> thisArgDocMap'
+   in SignatureInformation
```
michaelpj, line 168: also here

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+
+Alternative method 2: use only unstructured type string
+This method is hard to implement because we need to parse the type string.
+Some tricky cases are as follows:
+- h :: forall a (m :: Type -> Type). Monad m => a -> m a
+-}
+findParameterStringRanges :: UInt -> Text -> [Text] -> [(UInt, UInt)]
+findParameterStringRanges _totalPrefixLength _functionTypeString [] = []
+findParameterStringRanges totalPrefixLength functionTypeString (parameterTypeString : restParameterTypeStrings) =
+  let (prefix, match) = T.breakOn parameterTypeString functionTypeString
```
michaelpj, line 240: This isn't quite what I was suggesting, and I'm not sure it will work properly if we do have renaming inside types? e.g. does this work for `forall a . a -> a`? I would expect that we would get something like
- `functionTypeString = "forall a1. a1 -> a1`
- `parameterTypeStrings = ["a", "a"]`

And then the parameter type strings won't appear in the full type string? But you have a test for this, so I guess it does work? :thinking: 

Anyway, perhaps worth explaining if there are cases where you know this won't work, as I'm a bit confused!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -0,0 +1,308 @@
...
+        case extractName identifier of
+          Nothing -> Nothing
+          Just name ->
+            let mTypeOfName = identType identifierDetails
+                typesOfNode = case sourceNodeInfo hieAst of
+                  Nothing       -> []
+                  Just nodeInfo -> nodeType nodeInfo
+                allTypes = case mTypeOfName of
+                  Nothing -> typesOfNode
+                  Just typeOfName -> typeOfName : filter (isDifferentType typeOfName) typesOfNode
```
michaelpj, line 307: Why do we need to filter here? worth explaining!

michaelpj (approved):
Looks good! Could maybe do with a few more comments but I think we should merge this and make some issues for later improvements.

ghcide/src/Development/IDE/Core/RuleTypes.hs
jian-lin: > what these things are

You mean `DocMap`, `TyThingMap` and `ArgDocMap`?  Or just `ArgDocMap`?

> explaining what these things are

I am not sure what to explain here.  For example, it is clear that `DocMap` is a map from name to doc.

What about this doc: `'DocMap' stores docs for declarations: functions, data types, instances, methods etc.  'ArgDocMap' stores docs for arguments, e.g., function arguments, method arguments.`

> only functions (?) will have argument documentation

Probably.  Here is its [docstring][1] from GHC:

> Docs for arguments. E.g. function arguments, method arguments.

[1]: https://hackage-content.haskell.org/package/ghc-9.12.2/docs/GHC-Hs-Doc.html#v:docs_args

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: The core of this function is these [two lines][1].  I think the function name itself already explains the core pretty well.  Not sure what to add.

```
smallestContainingSatisfying (positionToSpan hiePath position) (nodeHasAnnotation ("HsApp", "HsExpr")) hieAst
  >>= extractInfo (positionToSpan hiePath position)
```

[1]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L235-L236

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: I have done many tests using Emacs eglot as the client.  I also have done only a few tests in vscode.  It seems to work well in both cases.

When experimenting the feature of automatically showing signature help when a space char in inserted in vscode, which is a future work and not implemented in this PR, I do notice that the stale value is not enough, I have to use `useE` to get the up-to-date value.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: This is by nature of our algorithm.

Input: `cursorPosition`
1. `hsAppNode` = get the smallest `HsApp` AST node which contains `cursorPostion` with the help of [extractInfoFromSmallestContainingFunctionApplicationAst][1]
2. `functionNode` = find the [left-most node][2] of `hsAppNode`
3. [try to get][3] `(functionName, functionTypes)` from `functionNode`.  `Nothing` means we fail to get the info.
4. count `parameterIndex` by [traversing][4] the `hsAppNode` tree from root to `cursorPosition`.  `Nothing` means either `cursorPosition` is at `functionNode` or we encounter some AST nodes we do not yet know how to continue our traversal.

So step 1, 3 and 4 can exclude things that don't look like functions.

[1]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L235-L236
[2]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L99
[3]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L100
[4]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L101

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: I think what you refer to as `active parameter` is `activateSignature` or `oldActivateSignature`.

> Perhaps the user can switch between the parameters, and if they've done that we should leave the active parameter as what they selected?

Users cannot select active parameter.  The active parameter is defined by the cursor position.

Users can select active signature and the feature of "remembering the previously selected signature when the cursor moves to another parameter" is implemented by reusing `oldActivateSignature`.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Vscode will ignore newlines sent from LSP server and does its own line wrapping.

Emacs eglot will respect newlines sent from LSP server.

Sending one line works well with vscode.  For Emacs eglot, I also think one line is better than multi-line, but this is just a personal taste or maybe my signature is not long enough.

Sending multi-line probably will have issues on vscode due to its own line wrapping.  For example, parameter type ranges in the signature may become wrong after vscode's own line wrapping.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: You should use the inlay hints of HLS :)

Just kidding, will use record constructor.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: The last type of `typesOfNode` may (always?) be the same as `typeOfName`.  To avoid generating two identical signature helps, we do a filtering here.

This filtering is similar to this [dropEnd1][1], I think.

Do you think it is worth adding a comment for it?  I personally think the code already says this clearly.

[1]: https://github.com/haskell/haskell-language-server/blob/60a6c486aa3d7d108022fa9059d09c7ade5e716e/ghcide/src/Development/IDE/Spans/AtPoint.hs#L270

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: You mean `forall a . a -> a` is as part of a larger type, right?  We have [a test case][1] for it.  That test case is marked as `mkTestExpectFail` and is [linked][2] to a future work.

If `forall a . a -> a` is not part of a larger type, it works.  [Many test cases][3] can verify this.

> This isn't quite what I was suggesting, and I'm not sure it will work properly if we do have renaming inside types

Renaming currently does not work.  We have [a test case][1] to track it.  It is also listed in the future work.

As we discussed in the meeting, I was trying to work on fixing it this week but have not finished it.  I plan to continue working on fixing it after the deadline.

[1]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/test/Main.hs#L295-L312
[2]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/test/Main.hs#L293-L294
[3]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/test/Main.hs#L139-L171

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Oh, I though you meant the `extractInfoFromSmallestContainingFunctionApplicationAst` function.  But maybe you mean [the whole algorithm][1].  I am not sure.

The algorithm is briefly described [here][2].  I can expand it a bit and add it to comments if that is what you mean.

[1]: https://github.com/linj-fork/haskell-language-server/blob/757d4a585d18aa5ca16e92c28579b0e75beb75a2/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L95-L103
[2]: https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2312567354

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Since both `SignatureHelp` and `SignatureInformation` have a field called `_activeParameter`, using record constructor syntax for both of them causes this error.

```
typecheck [GHC-87543]: Ambiguous occurrence ‘_activeParameter’.
It could refer to
   either the field ‘_activeParameter’ of record ‘SignatureHelp’,
          imported from ‘Language.LSP.Protocol.Types’ at /home/linj/code/fork/haskell-language-server/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs:89:5-22
          (and originally defined in ‘lsp-types-2.3.0.1:Language.LSP.Protocol.Internal.Types.SignatureHelp’),
       or the field ‘_activeParameter’ of record ‘SignatureInformation’,
          imported from ‘Language.LSP.Protocol.Types’ at /home/linj/code/fork/haskell-language-server/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs:92:5-29
          (and originally defined in ‘lsp-types-2.3.0.1:Language.LSP.Protocol.Internal.Types.SignatureInformation’).
```

fendor: Should work fine with the `{-# LANGUAGE DuplicateRecordFields #-}` extension

jian-lin: Thanks @fendor!  Since we only use (not define) the duplicated field here, `DisambiguateRecordFields` is enough.

Implemented in b5c13a3a0af45d722bdc169d9658976627a24d52.

---

I have a vague memory that there is an extension for this.  However, since GHC 9.8 does not gives suggestions in the error message, I thought my vague memory was wrong.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Implemented in b5c13a3a0af45d722bdc169d9658976627a24d52

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Implemented in b5c13a3a0af45d722bdc169d9658976627a24d52

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
fendor: What it does may be clear, but the reason is less clear. For example, in which cases are we generating the same types?

Since this filtering is similar to the one in `atPoint`, could we perhaps extract a common function?
Otherwise, just adding what you are saying in #4691 sounds like a good thing to write down!

jian-lin: > in which cases are we generating the same types

I think we always generate the same types but I am not sure. I need to check how HIE files are generated to figure this out if it is necessary.

> the reason is less clear

You mean why `typesOfNode` may (always?) contain `typeOfName`?  Currently, I do not know.  I need to check how HIE files are generated to figure this out if it is necessary.

> Since this filtering is similar to the one in atPoint, could we perhaps extract a common function?

Sounds good.

I did not do so because I am not sure 1) if there always are duplicate types and 2) if the duplicate type is the last one of `typesOfNode`.  If the answers of both questions are true, then we can use a more efficient implementation, i.e., `dropEnd1`, than filtering.

If you think filtering is efficient enough, I can extract a common function using filtering.

> Otherwise, just adding what you are saying in #4691 sounds like a good thing to write down!

I think what I said in #4691 is two things.

One thing is that only doing `dropEnd1` in some cases leads to a [bug][1].

Another thing can be summarized well using my reply [here][0]:

> The last type of typesOfNode may (always?) be the same as typeOfName. To avoid generating two identical signature helps, we do a filtering here.

So I guess you mean adding this sentence as a comment?  I can do that if that is what you mean.


[0]: https://github.com/haskell/haskell-language-server/pull/4626#discussion_r2312573327
[1]: https://github.com/haskell/haskell-language-server/pull/4691#discussion_r2319532259

fendor: > I did not do so because I am not sure 1) if there always are duplicate types and 2) if the duplicate type is the last one of typesOfNode. If the answers of both questions are true, then we can use a more efficient implementation, i.e., dropEnd1, than filtering.

I am not 100% sure, but we can try it out, and if a test fails, then we know whether it is the same :)
Otherwise, someone will report a bug at some point in which case we can write a regression test and fix and document it for all eternity!

> So I guess you mean adding this sentence as a comment? I can do that if that is what you mean.

Yes, please add this sentence, this gives us more context already.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
michaelpj: Yes, something like that. I mean the general approach to figuring out what bit of the AST is relevant. It's worth summarising!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
michaelpj: Worth a comment!

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
michaelpj: I mean the `activeParameter` that we get sent by the client. Maybe that doesn't matter and we only have to look at the `activeSignature` they send us?

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
michaelpj: One issue is that we can probably do better line wrapping than the client, because we know something about the text, e.g. we can break on function arrows. But happy to just leave a comment here, especially since the behaviour is inconsistent  between clients.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: We calculate the new `activeParameter` using the new cursor position.  I do not think the old `activeParameter` is useful.

fendor (approved):
LGTM, let's get this merged for the next HLS release :)

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: > I guess it's fine for us to ignore the active parameter. I don't know if this is something that the user can control? Maybe we should try it in VSCode? Perhaps the user can switch between the parameters, and if they've done that we should leave the active parameter as what they selected? Unclear.

OK, now I understand what you mean.  I am not a vscode user, but from my tests in vscode, users cannot control `activeParameter`.  `activeParameter` is decided by the cursor position.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Comments are added in 01ff52a84de4f32c94f7fb08443d85e9cf8c7963

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: comments are added in 01ff52a84de4f32c94f7fb08443d85e9cf8c7963

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: comments are added in 01ff52a84de4f32c94f7fb08443d85e9cf8c7963

ghcide/src/Development/IDE/Core/RuleTypes.hs
jian-lin: comments are added in 01ff52a84de4f32c94f7fb08443d85e9cf8c7963

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
jian-lin: Comments are added in 01ff52a84de4f32c94f7fb08443d85e9cf8c7963.

Extracting the common function is added as a TODO item in the comment.

PR #4630, opened 2025-06-09
Reload .cabal files when they are modified

fendor:


ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -586,7 +586,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
           unless (null new_deps || not checkProject) $ do
                 cfps' <- liftIO $ filterM (IO.doesFileExist . fromNormalizedFilePath) (concatMap targetLocations all_targets)
                 void $ shakeEnqueue extras $ mkDelayedAction "InitialLoad" Debug $ void $ do
-                    mmt <- uses GetModificationTime cfps'
+                    mmt <- uses GetPhysicalModificationTime cfps'
```
soulomoon: I don't understand how ignoring version in the VFS help us here🤔

fendor: Because cabal only works with the physical file system. If the VFS changes, then cabal simply can't perform a reload, as `cabal repl` will still observe the same `.cabal` file. That's why we only want to invalidate the shake session, if the `.cabal` file on disk has been modified.

Further, let's have a look at `getModificationTimeImpl`:

```haskell
getModificationTimeImpl
  :: Bool
  -> NormalizedFilePath
  -> Action (Maybe BS.ByteString, ([FileDiagnostic], Maybe FileVersion))
getModificationTimeImpl missingFileDiags file = do
    let file' = fromNormalizedFilePath file
    let wrap time = (Just $ LBS.toStrict $ B.encode $ toRational time, ([], Just $ ModificationTime time))
    mbVf <- getVirtualFile file
    case mbVf of
        Just (virtualFileVersion -> ver) -> do
            alwaysRerun
            pure (Just $ LBS.toStrict $ B.encode ver, ([], Just $ VFSVersion ver))
        Nothing -> do
            isWF <- use_ AddWatchedFile file
            if isWF
                then -- the file is watched so we can rely on FileWatched notifications,
                        -- but also need a dependency on IsFileOfInterest to reinstall
                        -- alwaysRerun when the file becomes VFS
                    void (use_ IsFileOfInterest file)
                else if isInterface file
                    then -- interface files are tracked specially using the closed world assumption
                        pure ()
                    else -- in all other cases we will need to freshly check the file system
                        alwaysRerun
                        
           ...
``` 

First, we don't care about the `isInterface file` check. Then, if the file isn't open in the VFS, we will never rerun this rule, unless the `IsFileOfInterest` rule is marked dirty, which it never will for `.cabal` files.

In my understanding, `isWF` will be `True` for `.cabal` files.

soulomoon: Thanks for the explanation, much clear to me now.

soulomoon:
Just some more interesting thoughts, do you think we can make GetModificationTime rule use GetPhysicalModificationTime rule in the end to remove some duplicated file time query?

fendor:
I am not sure, maybe. The whole file watcher architecture requires a thorough overhaul, it feels very brittle.

soulomoon:
> The whole file watcher architecture requires a thorough overhaul, it feels very brittle.

Indeed, are you planning to do it in this PR or open a seperate one for it?

fendor:
I am not sure :sweat_smile: 

Probably in a follow-up, maybe we should merge this as is and hope for the best.

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -586,7 +586,7 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
           unless (null new_deps || not checkProject) $ do
                 cfps' <- liftIO $ filterM (IO.doesFileExist . fromNormalizedFilePath) (concatMap targetLocations all_targets)
                 void $ shakeEnqueue extras $ mkDelayedAction "InitialLoad" Debug $ void $ do
-                    mmt <- uses GetModificationTime cfps'
```
soulomoon: If I am not mistaken, `GetPhysicalModificationTime ` is for `.cabal` files?
and `cfps'` are mostly `.hs` file locations, and we are type checking those files here.

But I do not really know how we handle `.cabal` file in HLS and where we call `GetModificationTime` for `.cabal` files.

fendor: Good catch, this is wrong, it should be `GetModificationTime`

fendor:
Finally with a regression test! This was a pain to write and I think I discovered another bug. Should be good to merge now.

EDIT: strictly speaking, I still need to add a test that the reload-logic works even when the `.cabal` file is not open, and we are not using the `hls-cabal-plugin`... I manually tested that it works.

PR #4641, opened 2025-06-23
Add tests for references to record fields and update symbol retrieval to only include source nodes

### LINKED ISSUE #4412: `textDocument/references` on a record field in a definition returns way too many things with ghc 9.6.4
This is essentially a duplicate of #3731, because I am now running 9.6.4, and it is still not fixed. I've copied over the original text below.

---

Currently, calling `textDocument/references` on the field of a record in a definition returns way too many things, including all of the other record fields, as well as the data type definition line, as well as some seemingly duplicate entries. Example:

For this file
```haskell
data Foo = MkFoo
  {
    bar :: Int,
    baz :: String
  }

fooUse0 :: Foo -> Int
fooUse0 MkFoo{bar} = 5

fooUse1 :: Foo -> Int
fooUse1 MkFoo{..} = 6

fooUse2 :: Int -> String -> Foo
fooUse2 bar baz =
  MkFoo{..}
```

Calling `textDocument/references` while my cursor is on `bar` on line 4 returns the following results:
2023-07-26-165942_636x284_scrot
A lot of these seem redundant:
1. The constructor definition (the last result)
2. Some of the lines are seemingly duplicated - `fooUse1` and `fooUse2` are both seemingly mentioned twice, which seems like a bug?
3. (my biggest issue) `baz :: String` also comes up as a result, which also seems like a bug, and is a huge pain when dealing with Real World :tm: data types with loads of fields, as this happens for each field.

I would like for only lines that mention the field in question to show up.

Maybe I'm just confused regarding what "references" means?

soulomoon:
Fix #4412
Very small change, Change `getNamesAtPoint` to only include names in the source.

### DIFF
--- a/ghcide-test/data/references/Fields.hs
+++ b/ghcide-test/data/references/Fields.hs
@@ -1,0 +1,18 @@
+{-# LANGUAGE RecordWildCards #-}
+module Fields where
+
+data Foo = MkFoo
+  {
+    barr :: String,
+    bazz :: String
+  }
+
+fooUse0 :: Foo -> String
+fooUse0 MkFoo{barr} = "5"
+
+fooUse1 :: Foo -> String
+fooUse1 MkFoo{..} = "6"
+
+fooUse2 :: String -> String -> Foo
+fooUse2 bar baz =
+  MkFoo{..}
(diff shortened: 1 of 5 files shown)


soulomoon:
@mergifyio requeue

PR #4651, opened 2025-07-08
4416 - show versions of installed packages 

### LINKED ISSUE #4416: Show information on installed package versions

batkot:
Took my chances on implementing @fendor suggestions from @lucalabs-de #4430 MR.

Inlay Hints
Screenshot 2025-07-08 at 23 23 16

CodeLenses got hard to read when there are multiple packages in single line, so I added package name then to make it easier to understand. 
Screenshot 2025-07-08 at 23 21 49

As noticed in previous MR, because of limited positioning options lenses can look a bit wonky depending on cabal file formatting. Hover could be a nice (and easy to add) fallback as well, but I'm not sure about status of hover ticket (#4417) and for now I just decided to avoid unnecessary conflicts.

Closes #4416 (based on #4430)


michaelpj:
Personally I'd be somewhat inclined to do this in a hover rather than an inlay hint, but debatable.

michaelpj:
Not sure what a coherent argument would look like, but I would argue something like:
- Inlay hints are good for things you always want to see that provide missing context
- Inlay hints are often supposed to be something you can "accept", which adds the content into the buffer. But the inlay hint here isn't valid syntax!
- Hovers are good for extra information that you want when you are focussing your attention on a specific thing. I think I mostly want to know package versions when I'm thinking about a specific package dependency, so I think that suits hovers a bit more?

batkot:
I added installed package version info to dependency hover and adjusted documentation link to point to the right version as well:

https://github.com/user-attachments/assets/f80f1d91-c636-4723-b453-cd59d30d1bad

Fair points and in principle I share your views on inlays. However I still like their look and feel for package versions. You immediately get a list of what's exactly installed. It didn't feel like cluttering the UI, since there's not a lot of LSP related info in `.cabal` files 🤔 

Anyway, I'm not married to the idea, so I can as well just drop it (inlays/code lenses) 😉

Issue #4653, opened 2025-07-09
`lookupKeyValue` in `Development.IDE.Graph.Internal.Key` sometimes errors


MangoIV:
while trouble shooting something else, I encountered HLS erroring with the following message: 

```
(GhcSessionDeps,NormalizedUri (-1539086352248807771) "inmemory://model/1")IntMap.!: key 15 is not
```

This comes from `lookupKeyValue` in `Development.IDE.Graph.Internal.Key`. The `IntMap` in that module's type `GlobalKeyValueMap` which is where the missing key was tried be looked up from, stores the keys to the values in another HashMap, i.e. the type looks as follows:

```hs
newtype Key = UnsafeMkKey Int

pattern Key :: () => (Typeable a, Hashable a, Show a) => a -> Key
pattern Key a <- (lookupKeyValue -> KeyValue a _)
{-# COMPLETE Key #-}

data KeyValue = forall a . (Typeable a, Hashable a, Show a) => KeyValue a Text

instance Eq KeyValue where
    KeyValue a _ == KeyValue b _ = Just a == cast b
instance Hashable KeyValue where
    hashWithSalt i (KeyValue x _) = hashWithSalt i (typeOf x, x)
instance Show KeyValue where
    show (KeyValue _ t) = T.unpack t

data GlobalKeyValueMap = GlobalKeyValueMap !(Map.HashMap KeyValue Key) !(IntMap KeyValue) {-# UNPACK #-} !Int

keyMap :: IORef GlobalKeyValueMap
keyMap = unsafePerformIO $ newIORef (GlobalKeyValueMap Map.empty IM.empty 0)

{-# NOINLINE keyMap #-}
```

The `Int` and the IntMap are related in the sense that the `Int` is supposed to indicate the *next* KeyValue to use. 

The only way to update the `IntMap` is by the use of `newKey` which returns an opaque wrapper around the `Int`. To my knowledge it is also the only way of obtaining such a Key, since its constructor is not exported: 

```haskell
newKey :: (Typeable a, Hashable a, Show a) => a -> Key
newKey k = unsafePerformIO $ do
  traceM $ "inserting new key: " <> show k
  let !newKey = KeyValue k (T.pack (show k)) -- FIXME: this is a spaceleak in its purest form (no use for the thing, purely space leak)
  atomicModifyIORef' keyMap $ \km@(GlobalKeyValueMap hm im n) ->
    let new_key = Map.lookup newKey hm
    in case new_key of
[...]

PR #4654, opened 2025-07-10
Mangoiv/improve key business

MangoIV:
Fixes issue #4653

### DIFF
--- a/hls-graph/src/Development/IDE/Graph/Internal/Key.hs
+++ b/hls-graph/src/Development/IDE/Graph/Internal/Key.hs
@@ -34,6 +34,7 @@ module Development.IDE.Graph.Internal.Key
     ) where
 
 --import Control.Monad.IO.Class ()
+import           Control.Exception             (evaluate)
 import           Data.Coerce
 import           Data.Dynamic
 import qualified Data.HashMap.Strict           as Map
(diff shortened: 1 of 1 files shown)


wz1000:
`renderKey` is pretty essential in large projects as it exists to ensure interning of `Text` strings that end up in diagnostics. When I last measured this we could be wasting tens to hundreds of megabytes on duplicated strings.

hls-graph/src/Development/IDE/Graph/Internal/Key.hs
```diff
@@ -83,8 +83,15 @@ newKey k = unsafePerformIO $ do
...
+-- NOTE:
+-- The argument to this function has a bang pattern. This bang pattern is a load
+-- bearing bang pattern. The reason is that the x, if not forced yet, is a thunk
+-- that forces the atomicModifyIORef' in the creation of the new key. If it
+-- isn't forced *before* reading the keyMap, the keyMap will only obtain the new
+-- key (x) *after* the IntMap is already copied out of the keyMap reference,
+-- i.e. when it is forced for the lookup in the IntMap.
 lookupKeyValue :: Key -> KeyValue
-lookupKeyValue (UnsafeMkKey x) = unsafePerformIO $ do
+lookupKeyValue (UnsafeMkKey !x) = unsafePerformIO $ do
```
wz1000: Perhaps an explicit `evaluate` might be better than a bang?

MangoIV:
Got it, thank you for the review @wz1000

fendor (approved):

Issue #201, opened 2020-07-14
Documentation popup: \mathcal not resolved/formatted?


mouse07410:
VS Code with the current `haskell-language-server` master, GHC-8.10.1.

For some reason, when the Documentation popup is displayed - `\mathcal{}` content is not formatted. Here's what the popup shows:
![Screen Shot 2020-07-14 at 11 29 56 ](https://user-images.githubusercontent.com/5923577/87445002-817e7900-c5c5-11ea-8b82-2089bb63fb55.png)

Here's what I see after clicking on "Documentation":
![Screen Shot 2020-07-14 at 10 59 32 ](https://user-images.githubusercontent.com/5923577/87442070-d7512200-c5c1-11ea-98d9-13402fe44f18.png)

Is it normal, or is there a way to make it formatted properly? How can I change my setup/configuration to enable that? Do I need to install some extra package...?

ndmitchell:
The source text actually says mathcal, but then Javascript goes in and replaces it. In general, Javascript won't run on docs being served up from local disk (as its a security risk), which I suspect is what happens. I personally think that relying on Javascript to format a slightly slopey O instead of just doing italics is a dumb case of making everything worse so pedants can have a fractionally more correct O. In Hoogle I just drop the entire mathcal bit, since I'm splicing in snippets of HTML, and don't want some Javascript going on too, which trashes page reflow time etc. See https://github.com/ndmitchell/hoogle/issues/353

mouse07410:
> Javascript won't run on docs being served up **from local disk** (as its a security risk), which I suspect is what happens

As far as I'm concerned, the **real** security risk is a ton of crappy JavaScript pulled from all over the world that a lot of web sites link to. 

I'd much prefer to have a configuration option to allow JavaScript run on docs served from the local disk, as it's an acceptable security risk (after all, I'm running this Cabal executable that came from the 'Net, ain't I? ;). 

However, if it's possible to avoid JavaScript altogether - I'd love to embrace that solution. Is it feasible here?

ndmitchell:
It would be feasible for upstream not to use Javascript, if they wanted faster pages and an irrelevantly less precise font. But upstream seems to have introduced it deliberately and foolishly...

mouse07410:
Ping?

mouse07410:
Ping?

ndmitchell:
[...]

PR #4658, opened 2025-07-13
Show LaTeX math expressions in haddockToMarkdown

webdevred:
- Replace fallback messages with raw LaTeX math expressions using `$...$` and `$$...$$`.
- This lets editors display the original math content even if they don't render LaTeX.
- No sanitization is performed, raw LaTeX is output as-is.
- Related to #201

### DIFF
--- a/ghcide/src/Development/IDE/Spans/Common.hs
+++ b/ghcide/src/Development/IDE/Spans/Common.hs
@@ -191,9 +191,9 @@ haddockToMarkdown (H.DocDefList things)
   = '\n' : (unlines $ map (\(term, defn) -> "+ **" ++ haddockToMarkdown term ++ "**: " ++ haddockToMarkdown defn) things)
 
 haddockToMarkdown (H.DocMathInline s)
-  = "$" ++ s ++ "$"
+  = "`" ++ s ++ "`"
 haddockToMarkdown (H.DocMathDisplay s)
-  = "\n$$\n" ++ s ++ "\n$$\n"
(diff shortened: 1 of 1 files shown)


fendor (approved):
LGTM, thanks!

michaelpj:
Should we be putting these inside a code formatting? They are, if anything, code, and I think it will read less weirdly if it's formatted that way? or are we hoping that some markdown renderers might actually try to render the latex?

webdevred:
@michaelpj Makes sense. I can update the patch to use code formatting instead if that’s preferred.

Edit:
Quick question, if we don’t go with `$...$` or `$$...$$` for math formatting, should we use plain backticks (\`...\`) for DocMathInline, or would it make sense to use fenced code with a latex language tag even for inline math content?

The idea behind using a latex tag or math-style blocks is to make it clearer that it’s LaTeX content. I know fenced blocks are unconventional for inline content, but I wanted to ask before finalizing.

Either way, this change is something I will really appreciate since it means I won’t have to open the documentation in a browser to see important details such as time complexity, all the important information will be in my text editor.

michaelpj:
I think it probably makes sense to use inline-code-formatting (single backticks) for inline math, and a fenced code block (triple backticks) for display math? so that would mean we could only include the language tag for display math, but that's a downside that all users of markdown face I guess.

webdevred:
I updated the patch to use single backticks for inline math and fenced code blocks with the latex tag for display math, as discussed.

michaelpj (approved):

PR #4663, opened 2025-07-16
concurrency bug fixes/ improvements

MangoIV:
Hi. I have two concurrency bug fixes/ improvements that pave the way towards multi client haskell-language-server. 

1. [[fix] don't bake ide state mvar into setup and getIdeState](https://github.com/haskell/haskell-language-server/commit/804169f008bf6baba79e6c5e230eefef1d03a1b9)
This is the right thing to do because othewise it is not possible to
create new ideStates in a single instance of the executable. This will
be useful if the hls executable is supposed to talk to multiple clients
and lives beyond a single client disconnecting

2. [[fix] don't throw hard errors when no shutdown message is handled](https://github.com/haskell/haskell-language-server/commit/dc9b00850206b6230e74b6f75d7a7abc87774800)
Previously, when there was no shutdown message by a client and the
client disconnected, resulting in the handlers to be GC'd the race that
was supposed to free resources for the HieDB & co. would throw a hard
error talking about the MVar being unreachable. We would like to instead
finish gracefully because finishing the race as soon as the MVar was
GC'd is the right thing to do anyway.


ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -265,11 +266,13 @@ runWithWorkerThreads recorder dbLoc f = evalContT $ do
...
--- | Runs the action until it ends or until the given MVar is put.
+-- | Runs the action until it ends or until the given MVar is put or the thread to fill the mvar is dropped, in which case the MVar will never be filled.
+--   This happens when the thread that handles the shutdown notification dies. Ideally, this should not rely on the RTS detecting the blocked MVar
+--   and instead *also* run the shutdown inf a finally block enclosing the handlers. In which case the BlockedIndefinitelyOnMVar Exception also wouldn't
+--   be thrown.
 --   Rethrows any exceptions.
 untilMVar :: MonadUnliftIO m => MVar () -> m () -> m ()
-untilMVar mvar io = void $
-    waitAnyCancel =<< traverse async [ io , readMVar mvar ]
+untilMVar mvar io = race_ (readMVar mvar `catch` \BlockedIndefinitelyOnMVar -> pure ()) io
```
fendor: This seems a little bit round-about, is this really preferable over

```suggestion
untilMVar mvar io = race_ (readMVar mvar) (io `finally` putMVar mvar ())
```

Also, I am not quite sure if I understand whether the `io` thread dying without putting the MVar is a bug on its own that needs fixing? What does dying mean here, does the thread crash for some reason?

fendor:
The second commit seems to be neither a bug fix nor a concurrency improvement.
However, as the change is benign, it seems fine to me.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
MangoIV: The io is not the thing putting the MVar. The MVar is put my the shutdown notification. If none is sent but the connection dies anyway then the MVar gets GC'd and the thread that tries to read the MVar gets a blocked indefinitely on MVar exception. Thats why your proposed change wouldn't work.

MangoIV: The "real" fix is to put the MVar as a bracket around the server dying, too.

MangoIV:
It doesn't appear to be one as of now ;) 

Later when we create multiple clients per run of the executable, it's important that we can create multiple ide states, too.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
fendor: Maybe the server shouldn't crash but gracefully shutdown if the connection is dropped?

MangoIV: Im not sure it crashes. It may just be that the thread dies before it receives a shutdown notification. That's very well possible if the client doesn't/ can't implement graceful shutdown

fendor:
Right, as of now it is a random change :P

As you know, I don't think the complexity that might be introduced by handling multiple clients at the same time should be handled within HLS. Perhaps if the complexity was encapsulated in a separate module / executable.
In any way, we can discuss this in detail once you created a ticket :)

MangoIV:
Well *this* complexity is necessary if we ever want a single executable with multiple clients which I think is actually *required* to make it feasible at all, mainly wrt memory footprint.

MangoIV:
I'll create a ticket.

MangoIV:
@fendor so whether or not I create the ticket, are these changes controversial?

ghcide/src/Development/IDE/LSP/LanguageServer.hs
fendor: But the server can handle the connection termination gracefully, right?

fendor:
No, these changes are not controversial

ghcide/src/Development/IDE/LSP/LanguageServer.hs
MangoIV: Yea. The thread just drops. Which makes the putMVar drop, too. Which triggers a threadBlockedIndefinitelyOnMVar exception

fendor: Right, but can we avoid relying on the rts for this case?

MangoIV: Well I'm assuming that the io will also finish so it's not a problem to not rely on the RTS, perhaps it just finishes later most of the time. But that's what I outlined above. The ideal fix is to hand the MVar somewhere else where it can be put as part of some bracketing operation. But since that ideal fix doesn't give us any but conceptual advantages, idk if it's necessary for now

MangoIV:
@wz1000 @fendor can I get another review pass so that I can rebase my uri branch on top of this? I would like to get that in a reviewable shape. :)

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -94,14 +95,14 @@ runLanguageServer
     -> (MVar ()
         -> IO (LSP.LanguageContextEnv config -> TRequestMessage Method_Initialize -> IO (Either (TResponseError Method_Initialize) (LSP.LanguageContextEnv config, a)),
                LSP.Handlers (m config),
-               (LanguageContextEnv config, a) -> m config <~> IO))
+               (LanguageContextEnv config, a) -> m config <~> IO, [IO ()]))
```
fendor:
```suggestion
               (LanguageContextEnv config, a) -> m config <~> IO, 
               [IO ()]))
```

And since you are already here, you could document the type :)

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -118,13 +119,14 @@ runLanguageServer recorder options inH outH defaultConfig parseConfig onConfigCh
...
-    void $ untilMVar clientMsgVar $
-          void $ LSP.runServerWithHandles
+    untilMVar clientMsgVar $
+          LSP.runServerWithHandles
             lspCologAction
             lspCologAction
             inH
             outH
             serverDefinition
+          `finally` sequence_ onExit
```
fendor: Comment with the explanation what `onExit` does, please :)

Perhaps introducing an intermediate variable for the `runServerWithHabdles` action?

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -265,11 +270,13 @@ runWithWorkerThreads recorder dbLoc f = evalContT $ do
             (WithHieDbShield hiedb, threadQueue) <- runWithDb recorder dbLoc
             liftIO $ f hiedb (ThreadQueue threadQueue sessionRestartTQueue sessionLoaderTQueue)
 
--- | Runs the action until it ends or until the given MVar is put.
+-- | Runs the action until it ends or until the given MVar is put or the thread to fill the mvar is dropped, in which case the MVar will never be filled.
+--   This happens when the thread that handles the shutdown notification dies. Ideally, this should not rely on the RTS detecting the blocked MVar
+--   and instead *also* run the shutdown inf a finally block enclosing the handlers. In which case the BlockedIndefinitelyOnMVar Exception also wouldn't
+--   be thrown.
```
fendor: Outdated comment? Or still kind of accurate?

fendor (approved):
LGTM, modulo nitpicks

MangoIV:
@fendor I think I applied all of your suggestions

MangoIV:
Ah shoot forgot to adjust the wrapper here.

PR #4670, opened 2025-07-22
 Add action for ignoring HLint hint for a single definition 

georgefst:
Based on #4669. While working on that, I realised that this should be fairly easy. Six hours later...

I've only done some light manual testing, but it seems to generally work. The one issue I've found is that the ignore comment is inserted after a value definition's type signature, when there is one, which is probably not what we want.

There's also some general code cleanup to do. See TODO comments. I'll get to that once #4669 is merged. As well as adding tests.

It doesn't actually matter where the ignore comment is placed, but I figured that at the top of the relevant definition is conventional.

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -675,6 +675,7 @@ library hls-hlint-plugin
     , bytestring
     , containers
     , deepseq
+    , extra
     , filepath
     , ghcide                == 2.14.0.0
     , hashable
--- a/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
+++ b/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
@@ -34,7 +34,7 @@ import           Control.Monad.IO.Class                             (MonadIO (li
 import           Control.Monad.Trans                                (lift)
 import           Control.Monad.Trans.Except                         (ExceptT (..),
                                                                      runExceptT)
-import           Control.Monad.Trans.Maybe                          (MaybeT (MaybeT),
+import           Control.Monad.Trans.Maybe                          (hoistMaybe,
                                                                      runMaybeT)
 import           Data.Aeson.Types                                   (FromJSON (..),
                                                                      ToJSON (..),
@@ -476,7 +476,7 @@ diagnosticToCodeActions nfp ideState verTxtDocId diagnostic
       -- When `declName` returns `Nothing`, that means HLint has no way of naming the declaration.
       -- And therefore we can't offer an action to ignore the hint locally at all.
       -- For example, we might be in an instance declaration.
-      name <- maybe (MaybeT $ pure Nothing) pure $ T.pack <$> declName containingDecl
+      name <- hoistMaybe $ T.pack <$> declName containingDecl
       line <- case srcSpanStart $ getLoc containingDecl of
         RealSrcLoc sl _ -> pure (srcLocLine sl - 1)
         UnhelpfulLoc _  -> err "unhelpful loc"
--- a/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
+++ b/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
@@ -693,7 +693,7 @@ applyRefactorings  =
 #endif
 #endif
 
--- TODO expose from HLint: https://github.com/ndmitchell/hlint/blob/086bf4430362ccced79ba2363840952eec67382a/src/GHC/Util/HsDecl.hs#L16
+-- TODO expose from HLint: https://github.com/ndmitchell/hlint/issues/1697
 declName :: LHsDecl GhcPs -> Maybe String
 declName (L _ x) = occNameString . occName <$> case x of
   TyClD _ FamDecl{tcdFam=FamilyDecl{fdLName}} -> Just $ unLoc fdLName
--- a/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
+++ b/plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE MultiWayIf            #-}
+{-# LANGUAGE NoFieldSelectors      #-}
 {-# LANGUAGE OverloadedLabels      #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE PatternSynonyms       #-}
@@ -30,13 +31,15 @@ import           Control.Exception
 import           Control.Lens                                       ((?~), (^.))
 import           Control.Monad.Error.Class                          (MonadError (throwError))
 import           Control.Monad.IO.Class                             (MonadIO (liftIO))
+import           Control.Monad.Trans                                (lift)
 import           Control.Monad.Trans.Except                         (ExceptT (..),
                                                                      runExceptT)
+import           Control.Monad.Trans.Maybe                          (MaybeT (MaybeT),
+                                                                     runMaybeT)
 import           Data.Aeson.Types                                   (FromJSON (..),
                                                                      ToJSON (..),
                                                                      Value (..))
 import qualified Data.ByteString                                    as BS
-import           Data.Either.Extra                                  (maybeToEither)
 import           Data.Hashable
 import qualified Data.HashMap.Strict                                as Map
 import           Data.List                                          (find)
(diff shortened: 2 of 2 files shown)


plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -64,19 +64,30 @@ import           System.Environment                                 (setEnv,
...
+                                                                     pattern UnhelpfulLoc,
                                                                      extensionFlags,
+                                                                     getLoc,
+                                                                     hsmodDecls,
                                                                      ms_hspp_opts,
+                                                                     pm_parsed_source,
+                                                                     srcLocLine,
+                                                                     srcSpanStart,
                                                                      topDir,
-                                                                     wopt)
```
georgefst: These imports are a bit of a mess, due to this file defining its own `RealSrcSpan` and others, because of the option to use `ghc-lib` rather than `ghc`. I'm not certain I've got it right for all versions. Hopefully CI will let us know.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -508,6 +530,7 @@ data HlintResolveCommands =
   | IgnoreHint
       { verTxtDocId     :: VersionedTextDocumentIdentifier
       , ignoreHintTitle :: HintTitle
+      , startPosition   :: IgnoreHintScope
```
georgefst: Whoops, I should have renamed this field.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -458,27 +473,45 @@ mkCodeAction title diagnostic data_  isPreferred =
...
+            (pm, _) <- runActionE "Hlint.GetParsedModule" ideState $ useWithStaleE GetParsedModule nfp
+            let defInfo = do
+                  containingDecl <- find (maybe False (positionInRange pos) . srcSpanToRange . getLoc)
+                    $ hsmodDecls $ unLoc $ pm_parsed_source pm
+                  defStartLine <- case srcSpanStart $ getLoc containingDecl of
+                    -- TODO `srcLocLine` can apparently raise an error, but it's not clear what the safe version is
+                    RealSrcLoc sl _ -> Just (srcLocLine sl - 1)
+                    UnhelpfulLoc _  -> Nothing
+                  -- TODO `documentSymbolForDecl` wasn't intended to be exported, and computes more than we need
+                  -- (although laziness should save us there)
```
georgefst: @fendor Do you know a better way to go `LHsDecl GhcPs -> Maybe Text`?

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -458,27 +473,45 @@ mkCodeAction title diagnostic data_  isPreferred =
...
+          IgnoreInModule ->
+            let NextPragmaInfo{nextPragmaLine, lineSplitTextEdits} = getNextPragmaInfo dynFlags (Just contents)
+            in pure $ mkSuppressHintTextEdits nextPragmaLine ignoreHintTitle lineSplitTextEdits Nothing
+          IgnoreInDefinition pos -> do
+            (pm, _) <- runActionE "Hlint.GetParsedModule" ideState $ useWithStaleE GetParsedModule nfp
+            let defInfo = do
+                  containingDecl <- find (maybe False (positionInRange pos) . srcSpanToRange . getLoc)
+                    $ hsmodDecls $ unLoc $ pm_parsed_source pm
+                  defStartLine <- case srcSpanStart $ getLoc containingDecl of
+                    -- TODO `srcLocLine` can apparently raise an error, but it's not clear what the safe version is
```
georgefst: I need to look in to this...

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
georgefst: Otherwise we can factor out the required part of `documentSymbolForDecl`, and use it there and here.

fendor: Sorry for the delay, I can't think of a better way, and think factoring out `documentSymbolForDecl` is reasonable!

georgefst: Aha, we can re-use HLint's logic: https://github.com/ndmitchell/hlint/blob/086bf4430362ccced79ba2363840952eec67382a/src/GHC/Util/HsDecl.hs#L16.

georgefst: Oh, wait, it's not exported...

georgefst: See https://github.com/ndmitchell/hlint/issues/1697. I've also updated the comment in the source to point there.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
georgefst: Someone at Zurihac did look in to this, and it seems to only actually apply to some ancient GHC versions. So we're fine, but should probably make a GHC PR to remove the docs.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
georgefst: Seems fine.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -64,18 +68,44 @@ import           System.Environment                                 (setEnv,
...
-import           GHC.Types.SrcLoc                                   hiding
-                                                                    (RealSrcSpan)
-import qualified GHC.Types.SrcLoc                                   as GHC
+import           GHC.Types.SrcLoc                                   hiding (L,
+                                                                     RealSrcSpan,
+                                                                     SrcLoc (..),
+                                                                     getLoc,
+                                                                     srcLocLine,
+                                                                     srcSpanStart,
+                                                                     unLoc)
```
georgefst, line 106: Could "Stylish Haskell" be more of a misnomer?

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -426,24 +456,44 @@ applyRefactAvailable = False
...
+    defIgnoreAction <- runMaybeT $ do
+      let err = throwError . PluginInternalError
+      (pm, _) <- lift $ runActionE "Hlint.GetParsedModule" ideState $ useWithStaleE GetParsedModule nfp
+      containingDecl <- maybe (err "no decl found covering expected source position") pure
+        $ find (maybe False (positionInRange start) . srcSpanToRange . getLoc)
+        $ hsmodDecls $ unLoc $ pm_parsed_source pm
+      -- When `declName` returns `Nothing`, that means HLint has no way of naming the declaration.
+      -- And therefore we can't offer an action to ignore the hint locally at all.
+      -- For example, we might be in an instance declaration.
+      name <- maybe (MaybeT $ pure Nothing) pure $ T.pack <$> declName containingDecl
```
georgefst: Oh, I meant to go back and check whether there's a shorthand for `maybe (MaybeT $ pure Nothing) pure`. I was surprised not to get any good hole suggestions.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
fendor: It is basically formatting the imports only :shrug: For everything else, you need to do it yourself :P

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
georgefst: It's `hoistMaybe`. Changed and pushed.

haskell-language-server.cabal
```diff
@@ -675,6 +675,7 @@ library hls-hlint-plugin
     , bytestring
     , containers
     , deepseq
+    , extra
```
georgefst, line 678: Ah, this was used during some intermediate state (for `liftEither`, I think), but no longer is.

But, honestly, it doesn't matter. We already have `extra` as an HLS dependency via many different routes.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
```diff
@@ -426,24 +456,44 @@ applyRefactAvailable = False
...
+  = do
+    defIgnoreAction <- runMaybeT $ do
+      let err = throwError . PluginInternalError
+      (pm, _) <- lift $ runActionE "Hlint.GetParsedModule" ideState $ useWithStaleE GetParsedModule nfp
+      containingDecl <- maybe (err "no decl found covering expected source position") pure
+        $ find (maybe False (positionInRange start) . srcSpanToRange . getLoc)
+        $ hsmodDecls $ unLoc $ pm_parsed_source pm
+      -- When `declName` returns `Nothing`, that means HLint has no way of naming the declaration.
+      -- And therefore we can't offer an action to ignore the hint locally at all.
+      -- For example, we might be in an instance declaration.
```
georgefst, line 478: I think this comment is pretty much correct.

In an ideal world, we'd have a pragma for just ignoring the following definition, without needing to name it. And one for acting on a particular line. But I've got better things to do than work on major upstream HLint changes...

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
georgefst: I know, and I hate the way it does imports.

georgefst:
Review request here was slightly premature. I forgot I haven't actually added any tests yet. Although from manual testing, I am fairly confident there are no major issues.

plugins/hls-hlint-plugin/src/Ide/Plugin/Hlint.hs
georgefst: > Someone at Zurihac

I should try to remember who this was. Probably either @dyniec or @xsebek?

xsebek: @georgefst you are right! 👍 

This changed some version ago when the SrcSpan types were split. Now this function is total, because the RealSrcSpan type does not have an empty `UnhelpfulSpan` alternative.

PR #4672, opened 2025-07-22
[feat] syntactic semantic tokens

### LINKED ISSUE #4501: the syntax highlighting of a field name used as a function and a field used with `-XOverloadedRecordDot` should look the same

- for some reason, fields when used as functions get highlighted as fields, but when used as accessors using overloaded record dot, get highlighted as functions

- both get highlighted as record accessors (distinct from function highlighting) 

n/a

## Additional context

- I'm using semantic highlighting

MangoIV:
Use the GHC AST and lsp semantic tokens to convince the language server to give highlighting even without any editor highlighting plugins.

Fixes #4501 

- [x] check that this works on all ghc versions 
- [ ] check which AST nodes I might have overlooked that are important. (add a tests that checks that all source ranges are covered (except space)


soulomoon:
Some keypoints beyond the tokens retrieval:
1. UTF8 vs UTF16 differences for the token range between haskell and LSP spec. see `focusTokenAt`.
2. Consider position mappings so we have correct highlighting even the file could not even parse, see `toCurrentRange`.
3. we need to consider delta between each result, see `semanticTokensFullDelta`. This is to minimize the bandwith between LSP server and clients. 

Post it here as a reminder.

MangoIV:
Thank you. I will go through all of these.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -75,8 +97,17 @@ computeSemanticTokens recorder pid _ nfp = do
   config <- lift $ useSemanticConfigAction pid
   logWith recorder Debug (LogConfig config)
   semanticId <- lift getAndIncreaseSemanticTokensId
-  (RangeHsSemanticTokenTypes {rangeSemanticList}, mapping) <- useWithStaleE GetSemanticTokens nfp
-  withExceptT PluginInternalError $ liftEither $ rangeSemanticsSemanticTokens semanticId config mapping rangeSemanticList
+
+  (sortOn fst -> tokenList, First mapping) <- do
+    rangesyntacticTypes <- lift $ useWithStale GetSyntacticTokens nuri
+    rangesemanticTypes <- lift $ useWithStale GetSemanticTokens nuri
```
soulomoon: These two rules might have different mappings. Need to consider both mappings.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -75,8 +97,17 @@ computeSemanticTokens recorder pid _ nfp = do
...
+    let mk w u (toks, mapping) = (map (fmap w) $ u toks, First mapping)
+    maybeToExceptT (PluginRuleFailed "no syntactic nor semantic tokens") $ hoistMaybe $
+      (mk HsSyntacticTokenType rangeSyntacticList <$> rangesyntacticTypes)
+      <>  (mk HsSemanticTokenType rangeSemanticList <$> rangesemanticTypes)
+
+  -- NOTE: rangeSemanticsSemanticTokens actually assumes that the tokesn are in order. that means they have to be sorted by position
+  withExceptT PluginInternalError $ liftEither $ rangeSemanticsSemanticTokens semanticId config mapping tokenList
```
soulomoon: Possible optimization: Inside rangeSemanticsSemanticTokens, we can skip some checks that might not needed for syntax tokens.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
MangoIV: why do we have to consider both mappings? isn't the one that we got most recently enough?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
MangoIV: which ones exactly? afaiu `rangeSemanticsSemanticTokens` applies the position mapping (`toCurrentRange`) and then actually converts the semantic tokens into their `lsp-types` counterpart.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: mappings are diffs from last successful computation, hence two rules may contain different mappings

MangoIV: So ideally we'd want to get a single mapping from both of the rules? Is it possible that the mapping changes after the first one? I e if we keep the first, is it possible that the two rules are interleaved with communication? Probably right?

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
soulomoon: Sorry, I confused it with computeRangeHsSemanticTokenTypeList. You’re right, rangeSemanticsSemanticTokens does exactly what you described and nothing more.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
MangoIV: So how about this: make a single rule that does both the token sets and build in the failover logic there. Then call a single rule here that succeeds if at least one set of tokens exist and use that with the position mapping that was generated.

soulomoon: >  succeeds if at least one set of tokens exist

That might not work for `semanticTokensFullDelta`.

One possible solution, adjust positions of each result by its own mappings first and then combine them.

plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Internal.hs
```diff
@@ -75,8 +98,17 @@ computeSemanticTokens recorder pid _ nfp = do
...
-  (RangeHsSemanticTokenTypes {rangeSemanticList}, mapping) <- useWithStaleE GetSemanticTokens nfp
-  withExceptT PluginInternalError $ liftEither $ rangeSemanticsSemanticTokens semanticId config mapping rangeSemanticList
+
+  (sortOn fst -> tokenList, First mapping) <- do
+    rangesyntacticTypes <- lift $ useWithStale GetSyntacticTokens nfp
+    rangesemanticTypes <- lift $ useWithStale GetSemanticTokens nfp
+    let mk w u (toks, mapping) = (map (fmap w) $ u toks, First mapping)
+    maybeToExceptT (PluginRuleFailed "no syntactic nor semantic tokens") $ hoistMaybe $
+      (mk HsSyntacticTokenType rangeSyntacticList <$> rangesyntacticTypes)
+      <>  (mk HsSemanticTokenType rangeSemanticList <$> rangesemanticTypes)
```
soulomoon: So we apply the each mappings first, then combine and sort the result.

MangoIV: Alright I'll do something like that then

MangoIV:
@soulomoon I'm currently struggling a bit with the tests - the ASTs are different enough that making the syntax highlighting consistent between ghc versions is probably a big pain. Should I introduce more goldens for all supported ghc versions?

MangoIV:
also `stylish-haskell`'s parser seems to be confused about the CPP :/

soulomoon:
> @soulomoon I'm currently struggling a bit with the tests - the ASTs are different enough that making the syntax highlighting consistent between ghc versions is probably a big pain. Should I introduce more goldens for all supported ghc versions? 

The only different should be that module name in the head is included since 9.10?

PR #4691, opened 2025-08-13
Add "Go to type" hyperlinks in the hover popup.

dnikolovv:
go-to-types

### DIFF
--- a/ghcide/src/Development/IDE/Core/LookupMod.hs
+++ b/ghcide/src/Development/IDE/Core/LookupMod.hs
@@ -1,0 +1,24 @@
+module Development.IDE.Core.LookupMod (lookupMod, LookupModule) where
+
+import           Control.Monad.Trans.Maybe       (MaybeT (MaybeT))
+import           Development.IDE.Core.Shake      (HieDbWriter, IdeAction)
+import           Development.IDE.GHC.Compat.Core (ModuleName, Unit)
+import           Development.IDE.Types.Location  (Uri)
+
+-- | Gives a Uri for the module, given the .hie file location and the the module info
+-- The Bool denotes if it is a boot module
+type LookupModule m = FilePath -> ModuleName -> Unit -> Bool -> MaybeT m Uri
+
+-- | Eventually this will lookup/generate URIs for files in dependencies, but not in the
+-- project. Right now, this is just a stub.
+lookupMod ::
+  -- | access the database
+  HieDbWriter ->
+  -- | The `.hie` file we got from the database
+  FilePath ->
+  ModuleName ->
+  Unit ->
+  -- | Is this file a boot file?
+  Bool ->
+  MaybeT IdeAction Uri
+lookupMod _dbchan _hie_f _mod _uid _boot = MaybeT $ pure Nothing
--- a/ghcide/src/Development/IDE/Spans/AtPoint.hs
+++ b/ghcide/src/Development/IDE/Spans/AtPoint.hs
@@ -45,7 +45,6 @@ import           Development.IDE.GHC.Util             (printOutputable)
 import           Development.IDE.Spans.Common
 import           Development.IDE.Types.Options
 
-import           Control.Applicative
 import           Control.Monad.Extra
 import           Control.Monad.IO.Class
 import           Control.Monad.Trans.Class
@@ -61,17 +60,25 @@ import           Data.Either
 import           Data.List.Extra                      (dropEnd1, nubOrd)
 
 
+import           Control.Lens                         ((^.))
 import           Data.Either.Extra                    (eitherToMaybe)
 import           Data.List                            (isSuffixOf, sortOn)
+import           Data.Set                             (Set)
+import qualified Data.Set                             as S
 import           Data.Tree
 import qualified Data.Tree                            as T
 import           Data.Version                         (showVersion)
+import           Development.IDE.Core.LookupMod       (LookupModule, lookupMod)
+import           Development.IDE.Core.Shake           (ShakeExtras (..),
+                                                       runIdeAction)
 import           Development.IDE.Types.Shake          (WithHieDb)
 import           GHC.Iface.Ext.Types                  (EvVarSource (..),
                                                        HieAST (..),
                                                        HieASTs (..),
                                                        HieArgs (..),
-                                                       HieType (..), Identifier,
+                                                       HieType (..),
+                                                       HieTypeFix (..),
+                                                       Identifier,
                                                        IdentifierDetails (..),
                                                        NodeInfo (..), Scope,
                                                        Span)
@@ -86,11 +93,8 @@ import           GHC.Iface.Ext.Utils                  (EvidenceInfo (..),
                                                        selectSmallestContaining)
 import           HieDb                                hiding (pointCommand,
                                                        withHieDb)
+import qualified Language.LSP.Protocol.Lens           as L
 import           System.Directory                     (doesFileExist)
-
--- | Gives a Uri for the module, given the .hie file location and the the module info
--- The Bool denotes if it is a boot module
-type LookupModule m = FilePath -> ModuleName -> Unit -> Bool -> MaybeT m Uri
 
 -- | HieFileResult for files of interest, along with the position mappings
 newtype FOIReferences = FOIReferences (HM.HashMap NormalizedFilePath (HieAstResult, PositionMapping))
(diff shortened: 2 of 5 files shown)


dschrempf:
Thanks! I had a quick look at the popups in my editor, and the "Go to ..." renders well and looks great. A few questions/suggestions:

- Can we sort the list of types in the same order they appear in the type signature of the highlighted definition?
- Does it make sense to write "Go to type ..." because we are only listing types afterwards.
- The hint only shows up when symbols on the value-level are highlighted. (See attached screenshots).

image
image

dnikolovv:
Hi @dschrempf 

> * Can we sort the list of types in the same order they appear in the type signature of the highlighted definition?

I need to check if the order they come in is deterministic when we query the locations.

> * Does it make sense to write "Go to type ..." because we are only listing types afterwards.

Yes, I just copied the Rust one.

> * The hint only shows up when symbols on the value-level are highlighted. (See attached screenshots).

Yeah. For some reason the locations query does not return anything when you're on the type level. This is how `Go To Definition` works right now. It might be a bug though?

dschrempf:
Ad "Go to type ... not showing for symbols on the type level": Actually, it makes sense, but only somewhat.

When the cursor is on `X` (see picture) 
image

- Go to definition goes to the definition of the data type `X`. That's good.
- Go to type does nothing (error: `Not found for: X`) -> Makes sense, the type of `X` is `Type`, or `Star`, we could go there.

However, when the cursor is on `x` on the type level (see picture)
image
- Go to definition goes to `x` on the value level (one line below). Does this make sense? I guess it does.
- Go to type gives the same error as above (error: `Not found for: x`). This should actually go to `X`, or suggest a list of types involved in defining `x`, shouldn't it? And this is also why you do not get a list of types @dnikolovv

dnikolovv:
@dschrempf 

> Go to type gives the same error as above (error: Not found for: x). This should actually go to X, or suggest a list of types involved in defining x, shouldn't it? And this is also why you do not get a list of types @dnikolovv

Exactly - the definitions in the hover use the same functionality underneath. I don't know if this is a bug or intended behavior. Maybe @fendor knows?

fendor:
I don't know exactly, I presume it is an artefact of HIE file structure or how GHC defines the AST and its source locations. So, probably a "bug".

I don't think it is something worth investing too much time into right now (i.e., not for this PR in particular), and we should rather focus on the presentation in the hover.

dnikolovv:
Not sure why this pipeline fails. Running `cabal test ghcide-tests` locally with GHC 9.12.2 gives me `All 398 tests passed (287.79s)`.

fendor:
@dnikolovv The testsuite in HLS is unfortunately quite flaky. You can likely ignore the testcase, and wait for an admin to restart the job. If the failing test case cannot be reproduced locally, then you can ignore it for now.

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -251,31 +255,41 @@ gotoImplementation withHieDb getHieFile ideOpts srcSpans pos
...
       where
-        pTypes :: [T.Text]
-        pTypes
-          | Prelude.length names == 1 = dropEnd1 $ map wrapHaskell prettyTypes
-          | otherwise = map wrapHaskell prettyTypes
+        pTypes :: M.Map Name Location -> [T.Text]
+        pTypes locationsMap =
+          case names of
+            [_singleName] -> dropEnd1 $ prettyTypes Nothing locationsMap
+            _             -> prettyTypes Nothing locationsMap
```
jian-lin, line 288: a bit offtopic:
- Why do we `dropEnd1` when `length names == 1`?
- Why don't we `dropEnd1` when `length names /= 1`?

dnikolovv: I'm tempted to remove this (actually I did but then reverted it) because it seems weird and unneeded to me as well. I tested a build without it and it seemed fine.

However, this is 4 year old code introduced by 2fef04193cfe2ac08372a008a5b5cfb6275d5bea so I can't really comment.

jian-lin: My guess for the motivation behind `dropEnd1` is that the author wants to avoid showing the same type/signature in `pTypes` as the one shown in [prettyName](https://github.com/haskell/haskell-language-server/blob/e3d38b0c4666681dca39901aa9d47b04422c82ff/ghcide/src/Development/IDE/Spans/AtPoint.hs#L304).  If that is the case, I think we should **also** do `dropEnd1` when `length names /= 1` (for example, `names` has 1 "actual" name and an evidence name), at least in today's GHC versions.

I also need to deal with this duplication when implementing [signature help](https://github.com/haskell/haskell-language-server/pull/4626).  I decide to [filter out the same type](https://github.com/linj-fork/haskell-language-server/blob/c0f784455ecf2026f37f8dd18dacecb7f104ff59/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs#L257) instead of `dropEnd1`.

jian-lin: Actually `dropEnd1` was introduced in 5dfee4c8c90a5ac66d5a5a66b869e5596e52c1db.  Maybe @wz1000 has some ideas to share?

soulomoon: ~Why just change it to use intelecate.~

ghcide/src/Development/IDE/Core/LookupMod.hs
```diff
@@ -0,0 +1,24 @@
...
+  -- | access the database
+  HieDbWriter ->
+  -- | The `.hie` file we got from the database
+  FilePath ->
+  ModuleName ->
+  Unit ->
+  -- | Is this file a boot file?
+  Bool ->
+  MaybeT IdeAction Uri
+lookupMod _dbchan _hie_f _mod _uid _boot = MaybeT $ pure Nothing
```
fendor, line 24: Is this still needed?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -251,31 +255,41 @@ gotoImplementation withHieDb getHieFile ideOpts srcSpans pos
...
+atPoint opts@IdeOptions{} shakeExtras@ShakeExtras{ withHieDb, hiedbWriter } har@(HAR _ (hf :: HieASTs a) rf _ (kind :: HieKind hietype)) (DKMap dm km) env pos =
     listToMaybe <$> sequence (pointCommand hf pos hoverInfo)
   where
     -- Hover info for values/data
     hoverInfo :: HieAST hietype -> IO (Maybe Range, [T.Text])
     hoverInfo ast = do
-        prettyNames <- mapM prettyName names
-        pure (Just range, prettyNames ++ pTypes)
+        locationsWithIdentifier <- runIdeAction "TypeCheck" shakeExtras $ do
+          runMaybeT $ gotoTypeDefinition withHieDb (lookupMod hiedbWriter) opts har pos
```
fendor, line 271: I think we might have to worry about start-up time until a hover request can be served.
But this should mostly pull data from hiedb, iirc, so maybe it is completely fine.

fendor (changes requested):
Some more review

ghcide/src/Development/IDE/Spans/AtPoint.hs
wz1000: @jian-lin your explanation is correct. If we have the expression `x` where `x :: Int`, then in the hover, we would get both the lines `x :: Int` for the variable at this point and `_ :: Int` for the entire expression. The point of the code is to remove the duplication.

jian-lin: Thanks @wz1000.  However, my main question is that we should always (not just when `length names == 1`) filter out the same type.  For example, the current hover info for `==` in `x = (==) True` contains duplicated type `forall a. Eq a => a -> a -> Bool`.

```
== :: forall a. Eq a => a -> a -> Bool

Defined in ‘GHC.Classes’ (ghc-prim-0.11.0)

Documentation

Source


* * *
Evidence of constraint Eq Bool
  using an external instance
  Defined in ‘GHC.Classes’
* * *

_ :: Bool -> Bool -> Bool
* * *

_ :: forall a. Eq a => a -> a -> Bool
  
* * *
infix 4 ==
```

ghcide/src/Development/IDE/Spans/AtPoint.hs
fendor: @wz1000 said this should be fine, so Ill resolve this comment

ghcide/src/Development/IDE/Core/LookupMod.hs
dnikolovv: ~~Nope, removed.~~

That was a bit too hasty. This was old code that I had to move around to avoid circular references. It doesn't do anything functionally but maybe the author wants to go back to it at some point.

dnikolovv:
> Thank you for implementing this nice feature!
> 
> Could we add a testcase? Otherwise, this looks good to me and we can merge after a test has been added :)

@fendor

Added tests.

ghcide/src/Development/IDE/Core/LookupMod.hs
dschrempf: In my opinion, and if it is dead code, it should go. We can revive it later if need be.

fendor:
There was one concern that this might explode the Hover message if maaaanny types are involved.

Do we perhaps want to impose a hard upper-limit on the number of elements we display, such as 10? What do you think about this?

ghcide/src/Development/IDE/Core/LookupMod.hs
fendor: We want to get back to it eventually, but I also think this is not needed as is and barely gives any advantage right now. So, let's just get rid of it, and then merge :)

fendor (approved):
LGTM, modulo my last comment, thank you for taking care of this!

dnikolovv:
> There was one concern that this might explode the Hover message if maaaanny types are involved.
> 
> Do we perhaps want to impose a hard upper-limit on the number of elements we display, such as 10? What do you think about this?

@fendor 

Yeah, that makes a lot of sense, I'll get that done in the following days.

ghcide/src/Development/IDE/Core/LookupMod.hs
dnikolovv: Perfect, I'll remove it then.

fendor:
@dnikolovv Are you blocked on anything or just a matter of time? I would like to merge this before the hls release, which I expect to be in a month at the latest.

fendor:
@dnikolovv If you don't mind, I can give a hand pushing this PR over the finish line.

dnikolovv:
> @dnikolovv If you don't mind, I can give a hand pushing this PR over the finish line. 

@fendor 

Oh, sorry! I was bogged down with work and forgot about this PR. I wouldn't mind if you want to push it over the finish line. I'll be able to do so next week if you can't.

dnikolovv:
@fendor Set the type count cap to 10.

fendor:
Thank you for taking your time to address the final PR comments! Much appreciated!

I am looking forward to seeing this feature in the next release :confetti_ball:

PR #4693, opened 2025-08-19
Support fourmolu 0.19.0.0

brandonchinn178:


brandonchinn178:
Yes, it should be compatible.

PR #4695, opened 2025-08-19
WIP: Use Structured Diagnostics for Typed Holes

sgillespie:
This adds structured diagnostics to the hole fit suggestions. Unfortunately, the suggestions still come through via `ErrorInfo`, so we still need to use regex matchers to extract them.

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
@@ -3,28 +3,30 @@ module Development.IDE.Plugin.Plugins.FillHole
   ) where
 
 import           Control.Lens
-import           Control.Monad                     (guard)
+import           Control.Monad                             (guard)
 import           Data.Char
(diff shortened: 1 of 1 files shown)


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -103,3 +133,18 @@ getIndentedGroupsBy pred inp = case dropWhile (not.pred) inp of
 indentation :: T.Text -> Int
 indentation = T.length . T.takeWhile isSpace
 
+-- TODO: This doesn't seem to handle qualified imports properly:
+--
+-- plugins/hls-refactor-plugin/test/Main.hs:4011:
+-- CodeAction with title "Replace _toException with E.toException" not found in
+-- [..., "Replace _toException with toException", ...]
+printErr :: SDoc -> T.Text
```
sgillespie, line 141: I can't seem to find any combination of `SDocContext` options that will result in `E.toException` (that is, with `import qualified Control.Exception as E`.

PR #4710, opened 2025-08-26
Replace writeFile with atomicFileWrite

### LINKED ISSUE #4709: Flasky test: addDependentFile
Fail result:
```
FAIL
        Exception: dep-file.txt: withFile: resource busy (file is locked)
```

We can switch our test to use atomic write to prevent this situation from happenning

soulomoon:
Fix #4709

### DIFF
--- a/hls-test-utils/src/Test/Hls/FileSystem.hs
+++ b/hls-test-utils/src/Test/Hls/FileSystem.hs
@@ -29,8 +29,12 @@ module Test.Hls.FileSystem
   , directProjectMulti
   , simpleCabalProject
   , simpleCabalProject'
+  , atomicFileWriteString
+  , atomicFileWriteStringUTF8
+  , atomicFileWriteText
   ) where
 
+import           Control.Exception           (onException)
 import           Data.Foldable               (traverse_)
 import qualified Data.Text                   as T
 import qualified Data.Text.IO                as T
@@ -38,6 +42,7 @@ import           Development.IDE             (NormalizedFilePath)
 import           Language.LSP.Protocol.Types (toNormalizedFilePath)
 import           System.Directory
 import           System.FilePath             as FP
+import           System.IO.Extra             (newTempFileWithin, writeFileUTF8)
 import           System.Process.Extra        (readProcess)
 
 -- ----------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


hls-test-utils/src/Test/Hls/FileSystem.hs
```diff
@@ -244,3 +249,26 @@ simpleCabalProject' :: [FileTree] -> [FileTree]
 simpleCabalProject' fps =
   [ simpleCabalCradle
   ] <> fps
+
+
+-- | Also resets the interface store
```
soulomoon:
```suggestion
```

hls-test-utils/src/Test/Hls/FileSystem.hs
```diff
@@ -244,3 +249,25 @@ simpleCabalProject' :: [FileTree] -> [FileTree]
...
   ] <> fps
+
+
+atomicFileWrite :: FilePath -> (FilePath -> IO a) -> IO a
+atomicFileWrite targetPath write = do
+  let dir = takeDirectory targetPath
+  createDirectoryIfMissing True dir
+  (tempFilePath, cleanUp) <- newTempFileWithin dir
+  (write tempFilePath >>= \x -> renameFile tempFilePath targetPath >> pure x)
+    `onException` cleanUp
```
fendor, line 260: While I am fine with writing a custom method, I can't believe there isn't already a library which does this for us, right?

soulomoon: Yes, perhaps we can introduce https://hackage.haskell.org/package/atomic-write as our dep ?

fendor: Yeah, perhaps a good idea... Or we just keep it as is, both is probably fine.

soulomoon: I take a look at [atomic-write](https://hackage.haskell.org/package/atomic-write) api, I do not see atomicFileWriteStringUTF8. So we need a bit customization even with that lib.
 
Let's keep as it is since it is very simple and just for test.

soulomoon:
Yes, perhaps we can enforce it in a follow up MR.

PR #4719, opened 2025-08-31
4718 hls graph session restart kills threads belong to later sesion session

### LINKED ISSUE #4718: hls-graph: thread belong to later session killed by previous session restart
From flaky test: `hls-hlint-plugin-tests::adding hlint flags to plugin configuration removes hlint diagnostics`
We can observe this incorrect cancellation by writing related event log.

soulomoon:
WordAround #4718

Fix the two problems that make hlint plugin flaky
1. uninterruptibleMask_ to avoid argsSettings capture async exception
2. bracket around `runWithSignal` to avoid sending the start signal without end signal, which would make hlint's tests flasky


fendor:
I just found this again, is this ready for merge? At least the `bracket_` seems like an improvement for the tests.

soulomoon:
> I just found this again, is this ready for merge? At least the `bracket_` seems like an improvement for the tests.

Thanks for digging it up.

This patch is not enough. and I traced it down, one of the thing we still need to do is that we should wait for the debouncer before we send the done signal.

PR #4722, opened 2025-09-05
Use edit-distance to fix typos in cabal fields

### LINKED ISSUE #4357: Don't use completions for cabal file codeactions
At the moment, codeactions (PR: #3273) in cabal files put a cursor at the end of the misspelled word and then suggest any possible completions.
This is not ideal, since words that have a typo early on will not trigger the correct codeaction.
We should use a different algorithm for the fuzzy matching than completions to better support the semantics of the codeactions.

crtschin:
Closes #4357.

This adds a pretty simple function for calculating the edit distance between two strings and uses it to score suggestions in completion lists. Some 
- I increased the cost of substitutions, otherwise the suggestions would very often lean towards nonsense.
- Kind of hackily injected the `Matcher` that's used in scoring texts into `CompleterData`. Couldn't really massage that into the `weightedConstantCompleter` that's used for licenses.
- This new scoring function doesn't do any filtering, so I had to adjust the cabal codeaction golden test to pick the top action per diagnostic.

### DIFF
--- a/ghcide/src/Text/Fuzzy/Levenshtein.hs
+++ b/ghcide/src/Text/Fuzzy/Levenshtein.hs
@@ -1,36 +1,16 @@
 module Text.Fuzzy.Levenshtein where
 
-import           Data.Function       (fix)
 import           Data.List           (sortOn)
-import           Data.MemoTrie
+import           Data.Text           (Text)
 import qualified Data.Text           as T
-import qualified Data.Text.Array     as T
-import           Data.Text.Internal  (Text (..))
+import           Text.EditDistance
 import           Text.Fuzzy.Parallel
-
--- | Same caveats apply w.r.t. ASCII as in 'Text.Fuzzy.Parallel'.
--- Might be worth optimizing this at some point, but it's good enoughᵗᵐ for now
-levenshtein :: Text -> Text -> Int
-levenshtein a b | T.null a = T.length b
-levenshtein a b | T.null b = T.length a
-levenshtein (Text aBuf aOff aLen) (Text bBuf bOff bLen) = do
-  let aTot = aOff + aLen
-      bTot = bOff + bLen
-      go' _ (!aIx, !bIx) | aIx >= aTot || bIx >= bTot = max (aTot - aIx) (bTot - bIx)
-      go' f (!aIx, !bIx) | T.unsafeIndex aBuf aIx == T.unsafeIndex bBuf bIx = f (aIx + 1, bIx + 1)
-      go' f (!aIx, !bIx) =
-        minimum
-          [ 2 + f (aIx + 1, bIx + 1), -- Give substitutions a heavier cost, so multiple typos cost more
-            1 + f (aIx + 1, bIx),
-            1 + f (aIx, bIx + 1)
-          ]
-      go = fix (memo . go')
-  go (aOff, bOff)
 
 -- | Sort the given list according to it's levenshtein distance relative to the
 -- given string.
 levenshteinScored :: Int -> Text -> [Text] -> [Scored Text]
-levenshteinScored chunkSize needle haystack =
+levenshteinScored chunkSize needle haystack = do
+  let levenshtein = levenshteinDistance $ defaultEditCosts {substitutionCosts=ConstantCost 2}
   sortOn score $
     matchPar chunkSize needle haystack id $
-      \a b -> Just $ levenshtein a b
+      \a b -> Just $ levenshtein (T.unpack a) (T.unpack b)
--- a/ghcide/src/Text/Fuzzy/Levenshtein.hs
+++ b/ghcide/src/Text/Fuzzy/Levenshtein.hs
@@ -1,0 +1,36 @@
+module Text.Fuzzy.Levenshtein where
+
+import           Data.Function       (fix)
+import           Data.List           (sortOn)
+import           Data.MemoTrie
+import qualified Data.Text           as T
+import qualified Data.Text.Array     as T
+import           Data.Text.Internal  (Text (..))
+import           Text.Fuzzy.Parallel
+
+-- | Same caveats apply w.r.t. ASCII as in 'Text.Fuzzy.Parallel'.
+-- Might be worth optimizing this at some point, but it's good enoughᵗᵐ for now
+levenshtein :: Text -> Text -> Int
+levenshtein a b | T.null a = T.length b
+levenshtein a b | T.null b = T.length a
+levenshtein (Text aBuf aOff aLen) (Text bBuf bOff bLen) = do
+  let aTot = aOff + aLen
+      bTot = bOff + bLen
+      go' _ (!aIx, !bIx) | aIx >= aTot || bIx >= bTot = max (aTot - aIx) (bTot - bIx)
+      go' f (!aIx, !bIx) | T.unsafeIndex aBuf aIx == T.unsafeIndex bBuf bIx = f (aIx + 1, bIx + 1)
+      go' f (!aIx, !bIx) =
+        minimum
+          [ 2 + f (aIx + 1, bIx + 1), -- Give substitutions a heavier cost, so multiple typos cost more
+            1 + f (aIx + 1, bIx),
+            1 + f (aIx, bIx + 1)
+          ]
+      go = fix (memo . go')
+  go (aOff, bOff)
+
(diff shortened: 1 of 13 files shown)


ghcide/src/Text/Fuzzy/Levenshtein.hs
```diff
@@ -0,0 +1,36 @@
+module Text.Fuzzy.Levenshtein where
```
Bodigrim, line 1: Can we use https://hackage.haskell.org/package/text-metrics-0.3.2/docs/Data-Text-Metrics.html#v:levenshtein ?

crtschin: Definitely.

Wanted to have an adventure in optimizing this, but that's probably better suited in a scratch file than HLS. I'll use [`edit-distance`](https://hackage.haskell.org/package/edit-distance-0.2.2.1/docs/Text-EditDistance.html) which allows me to vary the cost of substitions, which make the suggestions feel more logical (as evident by the golden test being reduced to `select top suggestion` instead of `find and select appropriate suggestion`, though that is biased).

I ran some quick benchmarks and `edit-distance` doesn't differ much in performance from `text-metrics`. Looks to be quicker even for most inputs I ran benches with.

<details><summary>The numbers Mason</summary>

```
differs = the texts being compared have no characters in common
same = the texts are identical

benchmarking edit-distance/empty
time                 13.03 ns   (13.02 ns .. 13.04 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 13.03 ns   (13.02 ns .. 13.07 ns)
std dev              50.10 ps   (19.90 ps .. 110.9 ps)

benchmarking edit-distance/differs 10
time                 193.4 ns   (192.2 ns .. 194.8 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 192.4 ns   (191.6 ns .. 193.3 ns)
std dev              2.959 ns   (2.310 ns .. 3.586 ns)
variance introduced by outliers: 17% (moderately inflated)

benchmarking edit-distance/differs 100
time                 9.553 μs   (9.525 μs .. 9.575 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 9.547 μs   (9.536 μs .. 9.557 μs)
std dev              35.09 ns   (27.82 ns .. 50.36 ns)

benchmarking edit-distance/differs 1000
time                 134.4 μs   (134.1 μs .. 134.5 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 132.8 μs   (132.4 μs .. 133.3 μs)
std dev              1.343 μs   (1.198 μs .. 1.535 μs)

benchmarking edit-distance/same 10
time                 267.5 ns   (265.3 ns .. 269.5 ns)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 263.9 ns   (262.5 ns .. 265.6 ns)
std dev              5.134 ns   (3.862 ns .. 6.415 ns)
variance introduced by outliers: 24% (moderately inflated)

benchmarking edit-distance/same 100
time                 22.24 μs   (22.17 μs .. 22.29 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 21.96 μs   (21.84 μs .. 22.05 μs)
std dev              342.6 ns   (277.3 ns .. 416.1 ns)
variance introduced by outliers: 12% (moderately inflated)

benchmarking edit-distance/same 1000
time                 405.5 μs   (404.4 μs .. 406.7 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 404.7 μs   (404.2 μs .. 405.3 μs)
std dev              1.769 μs   (1.425 μs .. 2.215 μs)

benchmarking text-metrics/empty
time                 7.633 ns   (7.555 ns .. 7.696 ns)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 7.532 ns   (7.484 ns .. 7.582 ns)
std dev              160.2 ps   (137.2 ps .. 198.4 ps)
variance introduced by outliers: 34% (moderately inflated)

benchmarking text-metrics/differs 10
time                 239.0 ns   (238.4 ns .. 239.8 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 240.0 ns   (239.6 ns .. 240.4 ns)
std dev              1.441 ns   (1.239 ns .. 1.697 ns)

benchmarking text-metrics/differs 100
time                 20.31 μs   (20.29 μs .. 20.33 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 20.29 μs   (20.27 μs .. 20.31 μs)
std dev              75.42 ns   (64.33 ns .. 88.95 ns)

benchmarking text-metrics/differs 1000
time                 1.906 ms   (1.902 ms .. 1.910 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.910 ms   (1.905 ms .. 1.923 ms)
std dev              22.69 μs   (9.892 μs .. 47.35 μs)

benchmarking text-metrics/same 10
time                 202.8 ns   (202.6 ns .. 203.2 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 204.0 ns   (203.4 ns .. 204.8 ns)
std dev              2.205 ns   (1.825 ns .. 2.698 ns)

benchmarking text-metrics/same 100
time                 16.09 μs   (16.08 μs .. 16.10 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 16.10 μs   (16.08 μs .. 16.14 μs)
std dev              70.10 ns   (11.60 ns .. 122.4 ns)

benchmarking text-metrics/same 1000
time                 1.562 ms   (1.561 ms .. 1.566 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.561 ms   (1.561 ms .. 1.563 ms)
std dev              2.427 μs   (469.5 ns .. 5.499 μs)
```
</details>

plugins/hls-cabal-plugin/test/testdata/code-actions/FieldSuggestionsTypos.cabal
```diff
@@ -0,0 +1,36 @@
...
+               -Wno-name-shadowing
+               -Wno-unticked-promoted-constructors
+
+source-repository head
+  type:     git
+  locqt: fake
+
+library
+    qqjfault-lang: Haskell2010
+    -- Import isn't supported right now.
```
fendor: Outdated comment

ghcide/src/Text/Fuzzy/Parallel.hs
```diff
@@ -18,6 +18,8 @@ import           Prelude                     hiding (filter)
 data Scored a = Scored {score :: !Int, original:: !a}
   deriving (Functor, Show)
 
+newtype Matcher a = Matcher { runMatcher :: T.Text -> [T.Text] -> [Scored a] }
```
fendor: Perhaps a comment would not hurt for why we have this :) 

Perhaps all fuzzy filters should return this interface? As is, this type is only used in the cabal-plugin, right?

crtschin: > Perhaps a comment would not hurt for why we have this :)

Will do.

> Perhaps all fuzzy filters should return this interface? As is, this type is only used in the cabal-plugin, right?

Interesting thought. It currently is indeed yes. I'll try changing the fuzzy functions to have it and check if it's better, otherwise I'll move it to the `cabal-plugin`, as I just added it to not have a bare function type in `CompleterData`.

crtschin: I ended up deciding against it, as this PR doesn't warrant touching the other plugins I think. Instead, I moved the `Matcher` type to `cabal-plugin`'s sources next to `CompleterData` where it's currently only used.

fendor:
@mergifyio requeue

crtschin:
I'm not sure I understand why the test is failing. It doesn't look related to my changes. I can reproduce this locally, but the same test also fails on master on my machine. 

I tried checking [the failing test](https://github.com/haskell/haskell-language-server/blob/master/ghcide-test/exe/DiagnosticTests.hs?plain=1#L368), but also don't see why it could be flaky though (if flaky tests are a thing in this repository). I might need some pointers.

fendor:
@crtschin Don't worry about it, most of our tests are unfortunately flaky one way or another :) I restarted the job, should be fine.

PR #4723, opened 2025-09-06
Enhance testing progress reporting

### LINKED ISSUE #4714: Flaky test: progress report should always send begin and end message if we start a progress
Ensure progress begin and end to be send during test if we start a progress.
A lot of test rely on it. Example #4627

Setting optProgressStartDelay = 0 and optProgressUpdateDelay = 0 is not enough. The progress report might still ends before it get the chance to sen progress begin and end message.

soulomoon:
- Introduce TestReporting style for progress reporting in IDE options.

fix #4714


soulomoon:
circleci failure is not caused by this patch :(

soulomoon:
might be this should be done in lsp too.

Issue #4390, opened 2024-08-25
enable eval plugin to run in its own terminal and enble terminal to be piped to a log file


theGhostJW:
## Is your enhancement request related to a problem? Please describe.

I just discovered that in VSCode the `eval` plugin logs to the Haskell Output window in addition adding the result to the code editor.

#### Code to Eval
```haskell
-- >>> helloWorldNonInteractive
-- "DONE"
helloWorldNonInteractive :: IO Text
helloWorldNonInteractive = putStrLn "Hello, World!" >> pure "DONE"
```

#### VSCode Haskell Output
```bash
2024-08-24T23:53:08.298464Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
2024-08-24T23:54:08.351720Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
toIfaceIdDetails [DataConWrapper]
toIfaceIdDetails [DataConWrapper]
toIfaceIdDetails [DataConWrapper]
Hello, World!
2024-08-24T23:55:08.383369Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
```

It is also apparent that the output that is interleaved with the HLS logs are not copied to file when that option is selected in VSCode settings.
 
#### haskell.log
```
2024-08-24T23:54:08.351135Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
2024-08-24T23:55:08.382573Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
2024-08-24T23:56:08.436285Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
2024-08-24T23:57:08.481933Z | Info | Live bytes: 95.98MB Heap size: 718.27MB
```

## Describe the solution you'd like

It would be awesome if there could be an option with the `eval` plugin to output to a separate output window, free of the noise of HLS logs and optionally log the same to a file. Better still if an interactive terminal could be used instead of a an output window. Interactive code seems to bork the current version hls `eval` (which is understandable given there is no way to interact with a program invoked via `eval`).

```haskell
-- >>> helloWorld
-- "DONE"
helloWorld :: IO Text
helloWorld = do
  putStrLn "What is your name?"
  name <- T.getLine
  putStrLn $ "Hello, " <> toS name <> "!"
  pure "DONE"
```

## Describe alternatives you've considered

[...]

Issue #4410, opened 2024-09-28
Eval plugin: output difff when there's trailing newline


jakio6:
### Your environment

Which OS do you use?
Linux

Which version of GHC do you use and how did you install it?
9.8.2, ghcup

How is your project built (alternative: link to the project)?
single file

Which LSP client (editor/plugin) do you use?
neovim builtin

Which version of HLS do you use and how did you install it?
2.7.0.0, ghcup

Have you configured HLS in any way (especially: a `hie.yaml` file)?
no
### Steps to reproduce

Evaluate twice
```hs
-- |
-- >>> import Prettyprinter
--
-- >>> pretty "foo"
--
-- >>> pretty "foo\n"
--
-- >>> pretty "foo\n\n"
--
```

### Expected behaviour

No diff for the last two test.

### Actual behaviour
First evaluate:
```hs
-- |
-- >>> import Prettyprinter
--
-- >>> pretty "foo"
-- foo
--
-- >>> pretty "foo\n"
-- foo
--
-- >>> pretty "foo\n\n"
-- foo
-- <BLANKLINE>
--
```
Second evaluate:
```hs
-- |
-- >>> import Prettyprinter
--
-- >>> pretty "foo"
-- foo
--
-- >>> pretty "foo\n"
-- WAS foo
-- NOW foo
--
-- >>> pretty "foo\n\n"
-- WAS foo
-- WAS <BLANKLINE>
-- NOW foo
-- <BLANKLINE>
--
```

fendor:
Hi, thank you for the bug report!

I feel like the reason why this is difficult to do correctly is that it is impossible to tell when the expected output ends... We would need a terminating character, or another identifier for multi line outputs. Or, we might want to forbid multi-line outputs altogether.

Maybe we can take inspiration from `cabal-docspec` and see how it behaves in this situation?

jakio6:
Append an extra `<BLANKLINE>` to the last two case would be enough to pass `doctest`.

jott8:
I'll try to work on this issue

xsebek:
This seems to have been fixed already. 🙂

<details><summary>Example without dependencies and output of HLS 2.14.0.0 (GHC: 9.12.4)</summary>

```Haskell
module MyLib (someFunc) where

data D = X | Y | Z

instance Show D where
    show = \case
      X -> "X\nX\n"
      Y -> "Y\n\nY\n\n"
      Z -> "Z\n\n\nZ\n\n\n"

[...]

Issue #2913, opened 2022-05-20
getLine in eval code lens causes connection got disposed


michaelmesser:
### Your environment

Which OS do you use:

MacOS
Which LSP client (editor/plugin) do you use:

VS Code+vscode-haskell

Describe your project (alternative: link to the project):

```
resolver:
  url: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/19/7.yaml

packages:
- .
```

### Steps to reproduce

```
{-
>>> getLine
-}
```

Run the Evaluate code lens

### Expected behaviour

Not sure

### Actual behaviour

```
{-
>>> getLine
"Content-Length: 191\r"
-}
```

VSCode shows error message: Connection got disposed.

xsebek:
Hi @michaelmesser, this is a known limitation of the **`hls-eval-plugin`**.


My understanding of the plugin is that it runs in the same process as HLS, so you can not easily redirect its `stdout`/`stderr`.
Redirecting `stdin` seems even trickier than output because it's not clear how you should interact with it from the editor.

If you want, we could use this Issue to discuss what should be the **Expected behaviour**. :slightly_smiling_face: 

- See also #1977

---

<sup>If you are looking for a workaround, you can get `IO String` with simple `pure "My Input"` or you can write your tested functions to accept `String` and separate the input/output handling. I hope that's helpful. :wink:</sup>

michaelpj:
Maybe we could capture this limitation in the docs? We have a few "Known limitations" sections in the Feature spage.

xsebek:
> Actual behaviour
> ```
> >>> getLine
> "Content-Length: 191\r"
> ```

Btw. is the string stolen from the input sent to HLS and making it close the connection? :thinking: 

@anka-213 pointed out in #1977 that HLS moves its `stdout`, maybe a similar thing is necessary for `stdin` here?

pepeiborra:
Yes, this is right. The redirection js done in Development.IDE.Main

PR #4726, opened 2025-09-12
Capture output of `eval` plugin and some girls scout changes

### LINKED ISSUE #1977: Capture stdout and stderr in eval plugin

dschrempf:
This PR contains some girl-scout changes as well as changes to how the `eval` plugin treats `stdout` and `stderr`.

#### Eval plugin

We now capture output to `stdout` and `stderr` induced as a possible side effect by the statement being evaluated. This is fragile because the output may be scrambled in a concurrent setting when HLS is writing to one of these file handles from a different thread.

Fixes #1977, most likely also #4390 (although this is debatable, we could do something more sophisticated).

```haskell
-- >>> putStrLn "Hello"
-- Hello

-- >>> hPutStrLn stderr "Hello"
-- Hello

-- >>> putStrLn "Hello," >> pure "World!"
-- Hello,
-- "World!"
```

#### Girl scout changes

- Docs: Fix some links in `eval` plugin in-source documentation
- Clarify documentation of expected test failure
- Update Nix Flake (Darwin issues seem to be solved)
- Update some tests and avoid mentioning GHC 9.2
- Remove comment referring to GHC 9.2 and fix code

### DIFF
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
@@ -68,12 +68,20 @@ showDiff (Second w) = "NOW " <> w
 showDiff (Both w _) = w
 
 -- | Compare the expected output recorded in the 'EvalExpr' with the actual
--- output @out@. When diffing is enabled and there is a recorded output, return
--- a line-by-line diff (see 'showDiffs'); otherwise return @out@ unchanged.
+-- output @out@, returning the result as one 'T.Text' per line. When diffing is
+-- enabled and there is a recorded output, return a line-by-line diff (see
+-- 'showDiffs'); otherwise return @out@ unchanged.
+--
+-- @out@ is normalised to one element per line first: a multi-line result
+-- arrives here as a single element with embedded newlines, whereas the recorded
+-- output is already split per line, and 'getDiff' compares element-wise. Without
+-- this, identical multi-line results would be reported as entirely changed.
 evalExprCheck :: Bool -> (Section, EvalExpr) -> [T.Text] -> [T.Text]
 evalExprCheck diff (section, evalExpr) out
-    | not diff || null (evalExprOutput evalExpr) || sectionLanguage section == Plain = out
-    | otherwise = showDiffs $ getDiff (map T.pack $ evalExprOutput evalExpr) out
+    | not diff || null (evalExprOutput evalExpr) || sectionLanguage section == Plain = outLines
+    | otherwise = showDiffs $ getDiff (map T.pack $ evalExprOutput evalExpr) outLines
+  where
+    outLines = concatMap T.lines out
 
 -- | The number of (expression lines, result lines) an 'EvalExpr' occupies.
 evalExprLengths :: EvalExpr -> (Int, Int)
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
@@ -72,11 +72,15 @@ showDiff (First w)  = "WAS " <> w
 showDiff (Second w) = "NOW " <> w
 showDiff (Both w _) = w
 
+-- | Compare the expected output recorded in the test with the actual output
+-- @out@. When diffing is enabled and there is a recorded output, return a
+-- line-by-line diff (see 'showDiffs'); otherwise return @out@ unchanged.
 testCheck :: Bool -> (Section, Test) -> [T.Text] -> [T.Text]
 testCheck diff (section, test) out
     | not diff || null (testOutput test) || sectionLanguage section == Plain = out
     | otherwise = showDiffs $ getDiff (map T.pack $ testOutput test) out
 
+-- | The number of (expression lines, result lines) a test occupies.
 testLengths :: Test -> (Int, Int)
 testLengths (Example e r _)  = (NE.length e, length r)
 testLengths (Property _ r _) = (1, length r)
@@ -84,9 +88,13 @@ testLengths (Property _ r _) = (1, length r)
 -- |A one-line Haskell statement
 type Statement = Loc String
 
+-- | The Haskell statements to feed to GHCi for a test, each tagged with its
+-- source line so evaluation errors can be located.
 asStatements :: Test -> [Statement]
 asStatements lt = locate $ Located (fromIntegral $ testRange lt ^. L.start . L.line) (asStmts lt)
 
+-- | The raw statement lines of a test. A 'Property' is wrapped so its result
+-- is evaluated through 'propEvaluation' (see 'propSetup').
 asStmts :: Test -> [Txt]
 asStmts (Example e _ _) = NE.toList e
 asStmts (Property t _ _) =
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs
@@ -17,6 +17,7 @@ import           Control.Lens                ((^.))
 import           Control.Monad.Catch         (MonadMask, bracket)
 import           Control.Monad.IO.Class
 import           Data.Algorithm.Diff         (Diff, PolyDiff (..), getDiff)
+import           Data.List                   (dropWhileEnd)
 import qualified Data.List.NonEmpty          as NE
 import           Data.Maybe                  (listToMaybe)
 import           Data.String                 (IsString)
@@ -45,7 +46,6 @@ evalExprRanges tst =
         resLine = startLine + exprLines
      in ( Range
             (Position startLine 0)
-            --(Position (startLine + exprLines + resultLines) 0),
             (Position resLine 0)
         , Range (Position resLine 0) (Position (resLine + resultLines) 0)
         )
@@ -54,16 +54,15 @@ evalExprRanges tst =
 resultRange :: EvalExpr -> Range
 resultRange = snd . evalExprRanges
 
--- TODO: handle BLANKLINE
-{-
->>> showDiffs $  getDiff ["abc","def","ghi","end"] ["abc","def","Z","ZZ","end"]
+{- |
+>>> showDiffs $ getDiff ["abc" :: String,"def","ghi","end"] ["abc","def","Z","ZZ","end"]
 ["abc","def","WAS ghi","NOW Z","NOW ZZ","end"]
 -}
 showDiffs :: (Semigroup a, IsString a) => [Diff a] -> [a]
 showDiffs = map showDiff
 
 showDiff :: (Semigroup a, IsString a) => Diff a -> a
-showDiff (First w)  = "WAS " <> w
+showDiff (First  w) = "WAS " <> w
 showDiff (Second w) = "NOW " <> w
 showDiff (Both w _) = w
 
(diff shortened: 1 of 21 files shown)


plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
```diff
@@ -426,11 +428,18 @@ A, possibly multi line, error is returned for a wrong declaration, directive or
...
-Variable not in scope: cls :: t0 -> t
-Data constructor not in scope: C
+WAS Variable not in scope: cls :: t0 -> t
+WAS Data constructor not in scope: C
+NOW Illegal term-level use of the class `C'
+NOW   defined at <interactive>:1:2
+NOW In the first argument of `cls', namely `C'
+NOW In the expression: cls C
+NOW In an equation for `it_a5Zks': it_a5Zks = cls C
+NOW Variable not in scope: cls :: t0_a5ZlS[tau:1] -> t1_a5ZlU[tau:1]
```
fendor: Slightly weird, probably needs to be fixed one way or another?

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
```diff
@@ -426,11 +428,18 @@ A, possibly multi line, error is returned for a wrong declaration, directive or
...
 
 >>> "A
-lexical error in string/character literal at end of input
+WAS lexical error in string/character literal at end of input
+NOW lexical error at end of input
```
fendor: Same here

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
```diff
@@ -446,7 +455,9 @@ bad times
 Or for a value that does not have a Show instance and can therefore not be displayed:
 >>> data V = V
 >>> V
-No instance for (Show V) arising from a use of ‘evalPrint’
+WAS No instance for (Show V) arising from a use of ‘evalPrint’
+NOW No instance for `Show V' arising from a use of `evalPrint'
+NOW In a stmt of an interactive GHCi command: evalPrint it_a5ZwT
 -}
```
fendor: and here :)

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
```diff
@@ -7,11 +7,10 @@
...
 
-{- |
-A plugin inspired by the REPLoid feature of <https://github.com/jyp/dante Dante>, <https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810775744 Haddock>'s Examples and Properties and <https://hackage.haskell.org/package/doctest Doctest>.
-
-For a full example see the "Ide.Plugin.Eval.Tutorial" module.
--}
+-- | -- A plugin inspired by the REPLoid feature of
+-- [Dante](https://github.com/jyp/dante),
+-- [Haddock examples and properties](https://haskell-haddock.readthedocs.io/latest/markup.html#examples),
+-- and [Doctest](https://hackage.haskell.org/package/doctest).
```
fendor:
```suggestion
-- | A plugin inspired by the REPLoid feature of
-- [Dante](https://github.com/jyp/dante),
-- [Haddock examples and properties](https://haskell-haddock.readthedocs.io/latest/markup.html#examples),
-- and [Doctest](https://hackage.haskell.org/package/doctest).
```

fendor (approved):
Needs a proper test in `hls-eval-plugin`, but otherwise looks good to me, thank you for the clean up! Feel free to merge once you've added a test :)

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
dschrempf: I didn't realize the `eval` plugin changed some other results. Thanks for picking that up. I will investigate.

dschrempf: I fixed those `WAS` `NOW` errors, one question though:

Does the plugin also evaluate all actions in the same comment block in your case?

If I evaluate one action in this comment block, all of them get evaluated and a lot of WAS NOW messages get added. Some of them are wrong. I am not sure if this PR added this discrepancy, or if the problem was already there before. I will check.

theGhostJW:
This could be an interesting surprise for some users. I often eval expressions that produce > 1000 lines of IO. In my case detailed pretty printed logs. If this was to suddenly be rendered in the code editor I'd need to change my workflow. If i wasn't expecting it, it would be quite confusing.

dschrempf:
Yes, if this is your workflow. In theory, we could limit the amount of output. The underlying library redirects output to a temporary file and then reads this temporary file. We could amend the function so that it only takes the first N lines of this file (or bytes, or whatever).

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
fendor: After locally running, I can confirm I also get these errors, independent of your changes.
This is how this file looks like after running eval: 

```patch
diff --git a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
index 1f19b5b47..0ae407bcb 100644
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
@@ -426,8 +426,14 @@ A, possibly multi line, error is returned for a wrong declaration, directive or
 Some flags have not been recognized: -XNonExistent
 
 >>> cls C
-Variable not in scope: cls :: t0 -> t
-Data constructor not in scope: C
+WAS Variable not in scope: cls :: t0 -> t
+WAS Data constructor not in scope: C
+NOW Illegal term-level use of the class `C'
+NOW   defined at <interactive>:1:2
+NOW In the first argument of `cls', namely `C'
+NOW In the expression: cls C
+NOW In an equation for `it_a2KTO': it_a2KTO = cls C
+NOW Variable not in scope: cls :: t0_a2KVe[tau:1] -> t1_a2KVg[tau:1]
 
 >>> "A
 lexical error in string/character literal at end of input
@@ -437,16 +443,20 @@ in GHCi or doctest. This allows it to be used as a hack to simulate print until
 get proper IO support. See #1977
 
 >>> 3 `div` 0
-divide by zero
+WAS divide by zero
+NOW *** Exception: divide by zero
 
 >>> error "Something went wrong\nbad times" :: E.SomeException
-Something went wrong
+WAS Something went wrong
+NOW *** Exception: Something went wrong
 bad times
 
 Or for a value that does not have a Show instance and can therefore not be displayed:
 >>> data V = V
 >>> V
-No instance for (Show V) arising from a use of ‘evalPrint’
+WAS No instance for (Show V) arising from a use of ‘evalPrint’
+NOW No instance for `Show V' arising from a use of `evalPrint'
+NOW In a stmt of an interactive GHCi command: evalPrint it_a2LvA
 -}
 evals :: Recorder (WithPriority Log) -> Bool -> TEnv -> DynFlags -> [Statement] -> Ghc [Text]
 evals recorder mark_exception fp df stmts = do
 ```

Comment (fendor) [reply plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs]: I think, we probably need to ignore these.

Comment (dschrempf) [reply plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs]: Alright! Thanks for confirming. I will make some room this week ;-).

Did we conclude on whether we should limit the maximum output? I think we decided not to have a limit, did we?

Comment (theGhostJW) [reply plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs]: who are you asking?

Comment (dschrempf) [reply plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs]: @fendor 

But also everybody else if they have an opinion or a fact to add, please go ahead!

**Comment by dschrempf** [comment]
Okay, finally some updates on this one.

- I had to revamp the method how we capture `stdout` and `stderr`. We now inject a setup and teardown function into the GHCi call. These calls setup the `stdout` and `stderr` capture, and also make sure the handles are closed afterwards. This is further ensured using `bracket` functions (new dependency on `exceptions`).
- I have changed the existing IO test to exercise the capture of `stdout` (instead of testing that it is _not_ captured).
- I have added another test that exercises capture of `stderr`.

I think we can merge this now, but waiting for your OK @fendor @xsebek.

**Comment by dschrempf** [comment]
(I still have to refresh the in-source `eval` examples).

File: plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Code.hs [hunk 5]
```diff
@@ -79,20 +92,115 @@ asStmts (Example e _ _) = NE.toList e
...
+-- can restore them. Bound to a tuple (rather than evaluated as a bare
+-- expression) so GHCi does not pass it to the interactive print function.
+captureSetup :: FilePath -> String
+-- Squeeze into one line (executed by GHCi).
+captureSetup outputTemp = unwords
+    [ "(__hls_captureHandle, __hls_savedStdout, __hls_savedStderr) <- do {"
+    , "  __hls_h <- System.IO.openFile", show outputTemp, "System.IO.WriteMode;"
+    , "  System.IO.hSetBuffering __hls_h System.IO.LineBuffering;"
+    , "  __hls_o <- GHC.IO.Handle.hDuplicate System.IO.stdout;"
+    , "  __hls_e <- GHC.IO.Handle.hDuplicate System.IO.stderr;"
```
xsebek, line 211: Could you please check if it is also possible to redirect standard input? The [docs](https://haskell-language-server.readthedocs.io/en/latest/features.html#evaluation-code-snippets-in-comments) mention it is currently broken:
> Standard input is shared with HLS, so e.g. [getLine breaks the connection to server](https://github.com/haskell/haskell-language-server/issues/2913).

But I guess fewer people run into that issue, while (`pretty`)`print` is very common. 🙂

xsebek (approved):
I tested it on the example from https://github.com/haskell/haskell-language-server/issues/4410 and it now works with just `putStrLn`, great! 🚀

dschrempf:
I retract my statement about LGTM; there are still minor issues to fix (<BLANKLINE>?).

Also, some lenses get out of date (not sure if this is a problem of my editor setup, or a more general problem with HLS).

dschrempf:
OK, I gave it another go.

I fixed one more bug, improved large parts of the code (it is piece of art, but still a mess), and updated the Readme.

I think it is good to merge now. @xsebek, I didn't implement your feature request (I ran out of time); I think it can easily go in a separate PR.

Let's discuss tomorrow.

PR #4727, opened 2025-09-15
Fix warning for `-Werror` build in Circle CI

fendor:
Circle CI runs its builds with `-Werror`. Thus, we need to fix the warning that was accidentally introduced.

See for example the logs of the job https://circleci.com/gh/haskell/haskell-language-server/70398 in https://github.com/haskell/haskell-language-server/pull/4726

### DIFF
--- a/plugins/hls-signature-help-plugin/test/Main.hs
+++ b/plugins/hls-signature-help-plugin/test/Main.hs
@@ -5,7 +5,6 @@ import           Control.Arrow                            ((>>>))
 import           Control.Exception                        (throw)
 import           Control.Lens                             ((^.))
 import           Data.Maybe                               (fromJust)
-import           Data.String.Interpolate                  (__i)
 import           Data.Text                                (Text)
 import qualified Data.Text                                as T
 import           Development.IDE.Plugin.Completions.Types (PosPrefixInfo (PosPrefixInfo))


plugins/hls-signature-help-plugin/test/Main.hs
```diff
@@ -5,7 +5,6 @@ import           Control.Arrow                            ((>>>))
 import           Control.Exception                        (throw)
 import           Control.Lens                             ((^.))
 import           Data.Maybe                               (fromJust)
-import           Data.String.Interpolate                  (__i)
```
jian-lin, line 8: Thanks for the fix and sorry for the extra work.

---

I am aware that this `import` line is redundant after #4717 is merged.  However, I am surprised that this causes CI failures since the [circleci check][1] did pass before I merged #4626.

I have done some digging and found the cause.  The default behavior of GitHub's `actions/checkout` is merging the PR branch and the target branch.  On the contrary, the default behavior of circleci's `checkout` does not do the merge.  Both of them can be configured to change to another behavior.  Here is GitHub's [doc][2].  I do not find the doc of circleci, but it should be doable.

It is debatable whether we should do the merge in the checkout action or not.  However, having different behaviors in different CI checks is just wrong, I think.  We should choose one single checkout behavior for all CI checks.

related: https://www.github.com/actions/checkout/issues/504

[1]: https://app.circleci.com/pipelines/github/haskell/haskell-language-server/14295/workflows/ff2a4ad3-81e8-4b57-b145-c39f34876a1b/jobs/70349
[2]: https://github.com/actions/checkout?tab=readme-ov-file#checkout-pull-request-head-commit-instead-of-merge-commit

fendor: I would like to migrate away from CircleCI as there is no reason to have CircleCI and GHA. The main issue so far was that we are limited by our caches, we only have 10GB, and our current jobs fill this quota by far. In fact, we are usually not hitting caches for at least one job because they were already evicted.
In the release CI, there is a cabal-cache process that backs up CI artefacts to a s3 buckets, perhaps we want to do something similar for regular CI jobs. We can't use the same existing s3 buckets, as we are only borrowing them.

PR #4736, opened 2025-10-04
Rely on gitignore to exclude listed files in ghcide (#4665)

### LINKED ISSUE #4665: `typecheck` should ignore `.gitignore`-ed files

Running `haskell-language-server` with no argument runs the `typecheck` command on all haskell files from the working directory recursively. In a repo using Bazel it can find Haskell files from the cache directory, which fails to typecheck.

`haskell-language-server` should not look into the cache directories. To do so it could use `.gitignore` if it finds one.

An alternative is to specify all relevant sub-directories manually like this:
```
haskell-language-server typecheck a b c d
```

## Additional context

A similar issue was originally reported in https://github.com/haskell/hie-bios/issues/299

blackheaven:
This aimes to fix #4665.

Note that I directly calling `git`, as th `gitignore` format is non-trivial.

I can roll a partial parsing implementation if requested.


blackheaven:
I don't think the CI failure is related to my PR.

sgillespie:
> I don't think the CI failure is related to my PR.

Looks like a flaky thing, it fixed itself

blackheaven:
Thanks, let's try again

ghcide/src/Development/IDE/Main.hs
```diff
@@ -446,15 +447,29 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
...
 
 expandFiles :: [FilePath] -> IO [FilePath]
-expandFiles = concatMapM $ \x -> do
+expandFiles paths = do
+  let haskellFind x =
+        let recurse "." = True
+            recurse y | "." `isPrefixOf` takeFileName y = False -- skip .git etc
+            recurse y = takeFileName y `notElem` ["dist", "dist-newstyle"] -- cabal directories
+        in filter (\y -> takeExtension y `elem` [".hs", ".lhs"]) <$> IO.listFilesInside (return . recurse) x
+  (testGitExitCode, _, _) <- readProcessWithExitCode "git" ["status"] ""
```
fendor: What happens if `git` is not installed?

ghcide/src/Development/IDE/Main.hs
```diff
@@ -446,15 +447,29 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
             c ide
 
 expandFiles :: [FilePath] -> IO [FilePath]
-expandFiles = concatMapM $ \x -> do
+expandFiles paths = do
```
fendor: As the behaviour got more complicated, could you expand the documentation of what this function does?

fendor (changes requested):
Small comments, otherwise LGTM!

Unfortunately, there are not automated tests for this cli mode, we'd need to add some, but that's some fair amount of work...

ghcide/src/Development/IDE/Main.hs
blackheaven: it should fallback to the current implementation

fendor: `readProcessWithExitCode` will throw an exception if the program can't be found:
```
ghci> import System.Process
ghci> readProcessWithExitCode "gi" ["status"] ""
*** Exception: gi: readCreateProcessWithExitCode: posix_spawnp: does not exist (No such file or directory)
```

we need to catch this exception to correctly fall back to the haskell implementation

ghcide/src/Development/IDE/Main.hs
```diff
@@ -445,16 +446,33 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
...
+            recurse y = takeFileName y `notElem` ["dist", "dist-newstyle"] -- cabal directories
+        in filter (\y -> takeExtension y `elem` [".hs", ".lhs"]) <$> IO.listFilesInside (return . recurse) x
+  (testGitExitCode, _, _) <- readProcessWithExitCode "git" ["status"] ""
+  let findFiles =
+        case testGitExitCode of
+          ExitSuccess -> \path -> do
+            let lookups = [path, path </> "*.hs", path </> "*.lhs"]
+            (trackedExitCode, trackedStdout, _) <- readProcessWithExitCode "git" ("ls-files":lookups) ""
+            (untrackedExitCode, untrackedStdout, _) <- readProcessWithExitCode "git" ("ls-files":"-o":lookups) ""
+            if trackedExitCode == ExitSuccess && untrackedExitCode == ExitSuccess
```
fendor: If I run hls via the rather common `haskell-language-server .` to load all files in the current directory, `git ls-files .` will probably list all files in the project, right?

fendor (changes requested):
Couple of issues when trying this patch on the HLS codebase, but otherwise I like the direction!

ghcide/src/Development/IDE/Main.hs
blackheaven: I rewrote my CLI calls to catch this.

ghcide/src/Development/IDE/Main.hs
blackheaven: I've just check, yes, I have added a check to distinguish file/directory.

blackheaven:
I won't lie, I have no clue about the broken pipe on windows error is related to my change.

fendor:
@blackheaven it is not caused by you. As already said, iirc there is not a single test case for the code path you are changing, so you can safely ignore any CI failures except pre-commit and `-Werror` :D

blackheaven:
Thanks a lot

ghcide/src/Development/IDE/Main.hs
```diff
@@ -445,16 +447,43 @@ defaultMain recorder Arguments{..} = withHeapStats (cmapWithPrio LogHeapStats re
...
+            case mResult of
+              Just (ExitSuccess, gitStdout, _) -> Just gitStdout
+              _                                -> Nothing
+  mHasGit <- git ["status"]
+  let findFiles =
+        case mHasGit of
+          Just _ -> \path -> do
+            let lookups =
+                  if takeExtension path `elem` [".hs", ".lhs"]
+                      then [path]
```
fendor, line 475: Question, what should happen if a user explicitly asks for a file to load, but the file is `.gitignore`d?

Should we perhaps still load it? I feel like, that might be more natural behaviour, but I am admittedly not sure.
If we do want to ignore ignored files, then I would still suggest that we should warn the user in this case, that we ingored a file they requested.

blackheaven: We could add an option in the CLI to by-pass gitignore, but I guess it'll be more complex.

Any idea which form the warning would take?

fendor: The warning should probably just be printed via the `Recorder`, so nothing fancy.

blackheaven: Here we are!

fendor (approved):
LGTM, thank you!

fendor:
Our CI is broken for good right now, @wz1000 is looking into it, but we don't know how long it is going to be broken...

Issue #4579, opened 2025-05-04
Documentation URLs are broken


svenpanne:
### Your environment

Which OS do you use?
Windows

Which version of GHC do you use and how did you install it?
GHC 9.10.1 from `ghcup`

Which LSP client (editor/plugin) do you use?
VS Code + Haskell plugin

Which version of HLS do you use and how did you install it?
HLS 2.10.0.0 from `ghcup`

### Steps to reproduce

Install HLS 2.10.0.0 and GHC 9.10.0 via `ghcup`. Install Haskell extension in VS Code. Write tiny "Hello, world!" program. Hover over e.g. `putStrLn` and click on the "Documentation" link. 

### Expected behaviour

The link should point to https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/GHC-Internal-System-IO.html#v:putStrLn.

### Actual behaviour

The broken link points to https://hackage.haskell.org/package/ghc-internal-9.1001.0-9d27/docs/GHC-Internal-System-IO.html#v:putStrLn. Note the `-9d27` in the URL, which is wrong. Basically the same happens for the "Source" link in the hover window.

I've got no idea if this is a problem with the VS Code plugin, the HLS, Haddock or something else, so I'll try here first. If this is wrong, it would be great to know where to create this issue.

fendor:
Thanks for pointing this out!

I think the initial link is generated by HLS, so this repo sounds like the right location.

VeryMilkyJoe:
I could not reproduce this on linux.

svenpanne:
I just tried it again with HLS 2.11.0.0 and GHC 9.12.2 (Windows, VS Code + Haskell plugin), but the strange suffixes are still there.

LCamel:
I encounter the same issue.

```
$ haskell-language-server-wrapper --version
haskell-language-server version: 2.10.0.0 (GHC: 9.10.1) (PATH: /home/vscode/.ghcup/hls/2.10.0.0/lib/haskell-language-server-2.10.0.0/bin/haskell-language-server-wrapper)
$ stack exec -- ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.8.4
$ uname -a
Linux codespaces-ccb522 6.8.0-1027-azure #32~22.04.1-Ubuntu SMP Thu Apr  3 20:26:27 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

Here is my configuration:
[...]

PR #4746, opened 2025-10-25
Use Hackage for Documentation and Source links

Saizan:
This is enabled by the new `linkToHackage` configuration option.

This is useful if you don't have local documentation or sources built, or other minor downsides of using local files. The `vscode-haskell` provides similar functionality but by trying to mangle the local file links, which is error prone because the path structure has changed between ghc versions.

If local documentation/source exists it gets used as a hint for which module to link to, which generally gives better results in case of re-exports. Otherwise we pick the module where the name is defined, and generate the link using its package name and version.

I am quite unsure if how I wired in the `linkToHackage` config option is appropriate. But I didn't want to do any refactor without some feedback first. 

Initially I figured `linkToHackage` should be a plugin config for `ghcide-hover-and-symbols`, but the flag affects the `GetDocMap` rule which structurally appears to be more part of the core? The first way I wrote it I got import cycles so I backed off.

I am happy to have the flag where it is or move it if desired. And in any case I will split it into linkDocToHackage and linkSrcToHackage for flexibility.

### DIFF
--- a/ghcide-test/exe/FindDefinitionAndHoverTests.hs
+++ b/ghcide-test/exe/FindDefinitionAndHoverTests.hs
@@ -18,6 +18,7 @@ import           Control.Lens               ((^.))
 import           Development.IDE.Test       (expectDiagnostics,
                                              standardizeQuotes)
 import           Hover
+import           Ide.Types                  (Config (..), OptLinkTo (..))
 import           Test.Hls
 import           Test.Hls.FileSystem        (copyDir)
 import           Text.Regex.TDFA            ((=~))
@@ -104,7 +105,9 @@ tests = let
           , ( "GotoHover.hs", [(DiagnosticSeverity_Error, (65, 8), "Found hole: _", Just "GHC-88464")])
           ]]
     , testGroup "type-definition" typeDefinitionTests
-    , testGroup "hover-record-dot-syntax" recordDotSyntaxTests ]
+    , testGroup "hover-record-dot-syntax" recordDotSyntaxTests
+    , testGroup "source-and-doc-links" linkToTests
+    ]
 
   typeDefinitionTests = [ tst (getTypeDefinitions, checkDefs) aaaL14 sourceFilePath (pure tcData) "Saturated data con"
                         , tst (getTypeDefinitions, checkDefs) aL20 sourceFilePath (pure [ExpectNoDefinitions]) "Polymorphic variable"]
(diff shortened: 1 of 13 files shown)


ghcide/src/Development/IDE/Types/Options.hs
```diff
@@ -85,6 +85,8 @@ data IdeOptions = IdeOptions
       -- ^ Experimental feature to re-run only the subset of the Shake graph that has changed
   , optVerifyCoreFile     :: Bool
     -- ^ Verify core files after serialization
+  , optLinkToHackage      :: Bool
+    -- ^ `Documentation` and `Source` link to Hackage, rather than local docs.
```
fendor: This could be a type `OptLinkToHackage` to avoid some boolean blindness, what do you think about that?

fendor (approved):
One suggestion and can we add a regression test case, somehow to make sure this doesn't accidentally break?

Otherwise, LGTM :)

Saizan:
@fendor I got a bit lost trying to add regression tests, would it look something like the ones in [FindDefinitionAndHoverTests](https://github.com/haskell/haskell-language-server/blob/master/ghcide-test/exe/FindDefinitionAndHoverTests.hs) ?

ghcide/src/Development/IDE/Spans/Documentation.hs
```diff
@@ -95,18 +101,46 @@ getDocumentationsTryGhc env names = do
...
-    toFileUriText = (fmap . fmap) (getUri . filePathToUri)
+    toFileUriText = fmap (getUri . filePathToUri)
+    toHackageUriText subdir sep env mod hint = do
+     ui <- lookupUnit env (moduleUnit mod)
+     let htmlFile = case hint of
+           Nothing -> T.intercalate sep (map T.pack $ moduleNameChunks mod) <> ".html"
+           Just foundFile -> T.replace "-" sep $ T.pack foundFile
+     pure $!
+      mconcat $
+      [ "http://hackage.haskell.org/package/"
```
vidit-od: We can use `https` here

ghcide/src/Development/IDE/Spans/Documentation.hs
```diff
@@ -95,18 +101,46 @@ getDocumentationsTryGhc env names = do
...
+              doc_link = case linkDoc of
+                LinkToHackage ->
+                  toHackageDocUriText env mod (takeFileName <$> doc)
+                LinkToLocal ->
+                  toFileUriText doc
+              src_link = case linkSource of
+                LinkToHackage ->
+                  toHackageSrcUriText env mod (takeFileName <$> src)
+                LinkToLocal ->
+                  toFileUriText src
```
vidit-od: ( personal nitpick , this is also fine)
can make these one lines 
i.e 
```haskell
    LinkToHackage -> toHackageDocUriText env mod (takeFileName <$> doc)
    LinkToLocal -> toFileUriText doc
src_link = case linkSource of
    LinkToHackage ->  toHackageSrcUriText env mod (takeFileName <$> src)
    LinkToLocal -> toFileUriText src
```

vidit-od (changes requested):

vidit-od:
@Saizan  Overall looks good, 
regarding regression tests, 

what you linked is a good pattern to follow, but from what i could see, 
`FindDefinitionAndHoverTests` will use `LinkToLocal` by default. You'll need to write a separate test function that uses `testithconfig` with a custom `Testconfig` that sets to `LinkToHackage` maybe

ghcide-test/exe/FindDefinitionAndHoverTests.hs
```diff
@@ -104,7 +105,9 @@ tests = let
           , ( "GotoHover.hs", [(DiagnosticSeverity_Error, (65, 8), "Found hole: _", Just "GHC-88464")])
           ]]
     , testGroup "type-definition" typeDefinitionTests
-    , testGroup "hover-record-dot-syntax" recordDotSyntaxTests ]
+    , testGroup "hover-record-dot-syntax" recordDotSyntaxTests
+    , testGroup "linkSourceTo,linkDocTo" linkToTests
```
fendor:
```suggestion
    , testGroup "source-and-doc-links" linkToTests
```

I am uneasy with `,` in test names, can tasty pattern handle these appropriately?

fendor (approved):
LGTM, thank you!

I am wondering what the default should be and we need to add it as a vscode config as well, so the users can change it...

Should the default be `LinkToHackage` or do you prefer `LinkToSource`?

Saizan:
I have been using LinkToHackage because I often don't have the local versions, however I figured we'd leave the old behavior as default

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -584,8 +584,8 @@ getDocMapRule recorder =
       (tmrTypechecked -> tc, _) <- useWithStale_ TypeCheck file
       (hscEnv -> hsc, _)        <- useWithStale_ GhcSessionDeps file
       (HAR{refMap=rf}, _)       <- useWithStale_ GetHieAst file
-
-      dkMap <- liftIO $ mkDocMap hsc rf tc
+      linkTgts <- linkTargets <$> getIdeOptions
```
vidit-od: i am not sure, but using `getClientConfigAction` would be better here right ?

vidit-od (approved):
Looks good to me as well !! LGTM !!

also, `LinktoSource` seems like a good default to me.

ghcide/src/Development/IDE/Core/Rules.hs
Saizan: Seems so, according to the note, thanks!

fendor:
> also, LinktoSource seems like a good default to me.

In VSCode, this links to an HTML page which is not rendered, so I don't think that's a good default.

[Screencast_20260515_111823.webm](https://github.com/user-attachments/assets/6ab784ef-5f99-4955-aadb-bd18afec7e96)


> however I figured we'd leave the old behavior as default

In VSCode, we rewrote this to hackage links, so it is not really the "default".
As the local sources aren't always available, I think we should use Hackage as the default.

---

Thus, I request to change the default to `LinkToHackage`, but I'll leave it up to @Saizan

Saizan:
It is properly rendered for me, but the links within it are opened in my actual browser, where they do not not resolve properly.

I'll make LinkToHackage the default for both documentation and source then.

dawkrish:
Can someone explain how does this ticket solve issue
https://github.com/haskell/haskell-language-server/issues/4579?

PR #4759, opened 2025-11-11
getLocatedImportsRule: "Simplify" some bogus logic

sjakobi:
This PR could have been an issue, but I wanted to use your CI to check whether my understanding of this code is correct.

* `ttmap` is a `HashMap` that maps each `NormalizedFilePath` key to itself.
* So if `nfp` is a member of `ttmap`, we find the same `nfp` in the map, and have `nfp' = nfp`.
* If `nfp` isn't a member, we fall back to the default, which is also `nfp`.
* So in both cases we end up with `nfp' = nfp`.

I don't actually know what this code was meant to do! It was introduced in https://github.com/haskell/haskell-language-server/commit/0e642b3a35d150b8e4ca743a78f7048db15d1562.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -328,8 +328,7 @@ getLocatedImportsRule recorder =
                     return $ if itExists then Just nfp' else Nothing
                 | Just tt <- HM.lookup (TargetModule modName) targets = do
                     -- reuse the existing NormalizedFilePath in order to maximize sharing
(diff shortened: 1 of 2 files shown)


sjakobi:
Ugh, now I get it. It really is only about the sharing.

Seems like the cold I have is also affecting my brain function. :/

sjakobi:
I wonder whether `unordered-containers` could make this much simpler though:

https://github.com/haskell-unordered-containers/unordered-containers/issues/546

I would love to get your feedback on that idea!

PR #4766, opened 2025-11-25
Add default-extension completions in .cabal files

### LINKED ISSUE #4758: hls-cabal-plugin: Missing completions for `default-extensions`
Image

Seems to be missing regardless of the common stanza.

VeryMilkyJoe:
Suggest all known language extensions for the `default-extensions` field in .cabal files.

Also add a new variation, `mkParameterisedTestCaseM`, of `mkParameterisedTestCase` which takes an Assertion instead checking for equality of the given values.

Closes #4758


fendor (commented):
LGTM, thank you for updating the plugin!

PR #4768, opened 2025-11-26
Use 9.10.2 instead of 9.10.3 for CI on windows

wz1000:
9.10.2 is currently broken with TH on windows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -146,7 +146,7 @@ jobs:
         run: cabal test hls-refactor-plugin-tests || cabal test hls-refactor-plugin-tests
 
       # TODO enable when it supports 9.10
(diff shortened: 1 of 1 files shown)


fendor:
GHC commit that seems to be the root of the issue: https://gitlab.haskell.org/ghc/ghc/-/commit/b9f49013061efcb2ae1d0986fdad3baf39994067

Commit is present in GHC 9.14.1 rc and GHC 9.12.3 rc, but the commit is likely to be reverted.

wz1000:
See https://gitlab.haskell.org/ghc/ghc/-/issues/26613

wz1000:
I have updated CI to add exceptions for 9.10.2 wherever there were exceptions to 9.10. Hopefully this passes now.

PR #4773, opened 2025-12-06
Fix notes plugin: avoid crashes on missing note definitions and improve lookup handling

### LINKED ISSUE #4763: `hls-notes-plugin` crashes instead of reporting no locations
The `hls-notes-plugin` reports an error instead of no locations when it can't find the note in question.

Rather, we should just show the same thing like when "Goto definition" fails.

Image

~~Slightly unrelated, I don't know why it fails to find the note, as it does exist, but that's another question.~~ The note `Note [Shutdown]` is ill-formatted.

SATVIKsynopsis:
Fixes #4763 

This PR updates the Notes plugin to safely handle cases where note definitions or references are missing.
So, Instead of throwing internal errors, the plugin now returns `Null` results as expected by the LSP specs.

Changes include:

1. Graceful handling of missing notes in jump-to-definition and references.

2. Replacing unsafe lookups with `fromMaybe`.

3. Cleaning up unused imports.

Functionality tested with a simple example;   jump-to-definition works correctly and missing notes no longer cause errors.

### DIFF
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -1,7 +1,8 @@
 module Ide.Plugin.Notes (descriptor, Log) where
 
 import           Control.Lens                     ((^.))
-import           Control.Monad.Except             (ExceptT)
+import           Control.Monad.Except             (ExceptT, MonadError,
+                                                   throwError)
 import           Control.Monad.IO.Class           (liftIO)
 import qualified Data.Array                       as A
 import           Data.Foldable                    (foldl')
@@ -10,13 +11,13 @@ import qualified Data.HashMap.Strict              as HM
 import qualified Data.HashSet                     as HS
 import           Data.List                        (uncons)
 import           Data.Maybe                       (catMaybes, listToMaybe,
-                                                   mapMaybe, fromMaybe)
+                                                   mapMaybe)
 import           Data.Text                        (Text, intercalate)
 import qualified Data.Text                        as T
 import qualified Data.Text.Utf16.Rope.Mixed       as Rope
 import           Data.Traversable                 (for)
 import           Development.IDE                  hiding (line)
-
+import           Development.IDE.Core.PluginUtils (runActionE, useE)
 import           Development.IDE.Core.Shake       (toKnownFiles)
 import qualified Development.IDE.Core.Shake       as Shake
 import           Development.IDE.Graph.Classes    (Hashable, NFData)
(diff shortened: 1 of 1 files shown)


plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -100,25 +100,36 @@ findNotesRules recorder = do
...
+                Nothing -> pure Nothing
+                Just ln ->
+                    
+                    pure $
+                        listToMaybe $
+                            mapMaybe (atPos (fromIntegral c)) $
+                                matchAllText noteRefRegex ln
   where
-    atPos c arr = case arr A.! 0 of
-        -- We check if the line we are currently at contains a note
```
jvanbruegge, line 116: Why delete this comment?

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -100,25 +100,36 @@ findNotesRules recorder = do
...
-    contents <-
-        err "Error getting file contents"
-        =<< liftIO (runAction "notes.getfileContents" state (getFileContents nfp))
-    line <- err "Line not found in file" (listToMaybe $ Rope.lines $ fst
-        (Rope.splitAtLine 1 $ snd $ Rope.splitAtLine (fromIntegral l) contents))
-    pure $ listToMaybe $ mapMaybe (atPos $ fromIntegral c) $ matchAllText noteRefRegex line
+    mContents <- liftIO (runAction "notes.getfileContents" state (getFileContents nfp))
+    case mContents of
+        Nothing ->
+
```
jvanbruegge: Why the extra newlines?

SATVIKsynopsis:
Thanks for the feedback

1. I have updated the `listReferences` and `jumpToNote` handlers as requested

2. Removed the previous err-based handling for missing notes and references.

3. Added explicit case `HM.lookup` branches to return `Null` when the note or reference is not found.

4. Ensured both handlers now match the behavior of `“Goto Definition”` when no match exists.

5. No unrelated logic or formatting was changed.

Please let me know if anything else should be adjusted.

mati865:
@jvanbruegge a heads-up, it might be hard to review this PR because this is at least largely a chatbot's work. You can see similar pull requests to other projects on author's profile.

SATVIKsynopsis:
To clarify, I reviewed the changes myself before submitting the PR.
If anything looks off or needs adjustment I am m happy to revise it. Please let me know what should be improved.

jvanbruegge:
The code looks now fine, but the pre-commit hook needs to pass and it should have a test for the new empty responses

SATVIKsynopsis:
Thanks for the review

I am currently focusing on Rust contributions and I no longer have the Haskell setup locally.
Because of this, I won’t be able to continue this PR right now.

Feel free to close it or let someone else pick it up.
I learned from the review and appreciate your time.

fendor:
Superseded by #4828

PR #4780, opened 2026-01-01
9.14 support

wz1000:
- **wip: Build ghcide with 9.14**
- **plugins build with 9.14**

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -318,7 +318,6 @@ test-suite hls-cabal-plugin-tests
     , haskell-language-server:hls-cabal-plugin
     , hls-test-utils    == 2.12.0.0
     , lens
-    , lsp
     , lsp-types
     , text
 
--- a/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Definition.hs
+++ b/plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/Definition.hs
@@ -198,9 +198,10 @@ lookupBuildTargetPackageDescription (PackageDescription {..}) (Just buildTargetN
 -- better ways to get the path.
 --
 -- Examples: (output is system dependent)
---   >>> toHaskellFile "My.Module.Lib"
---   "My/Module/Lib.hs"
---   >>> toHaskellFile "Main"
---   "Main.hs"
+--
+-- >>> toHaskellFile "My.Module.Lib"
+-- "My/Module/Lib.hs"
+-- >>> toHaskellFile "Main"
+-- "Main.hs"
 toHaskellFile :: T.Text -> FilePath
 toHaskellFile moduleName = joinPath (map T.unpack $ T.splitOn "." moduleName) <.> ".hs"
--- a/plugins/hls-cabal-plugin/test/Definition.hs
+++ b/plugins/hls-cabal-plugin/test/Definition.hs
@@ -6,8 +6,6 @@ module Definition (
 
 import           Control.Lens                ((^.))
 import           Data.List.Extra             (isSuffixOf)
-import qualified Data.Text                   as T
-import           Ide.Plugin.Cabal.Definition (toHaskellFile)
 import qualified Language.LSP.Protocol.Lens  as L
 import qualified Language.LSP.Protocol.Types as LSP
 import           System.FilePath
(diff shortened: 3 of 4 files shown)


cabal.project
```diff
@@ -7,7 +7,12 @@ packages:
          ./hls-test-utils
 
 
-index-state: 2025-08-08T12:31:54Z
+index-state: 2026-01-01T00:00:00Z
+
+source-repository-package
+  type: git
+  location: https://github.com/snowleopard/alga
+  tag: d4e43fb42db05413459fb2df493361d5a666588a
```
fendor, line 15: Do we need this? Or is this just an allow-newer?

dschrempf: I advice against using `source-repository-package`. I think it is a workaround when the package is not yet available on another repository such as Hackage. Can we update `alga` on Hackage instead?

.github/actions/setup-build/action.yml
```diff
@@ -7,7 +7,7 @@ inputs:
   cabal:
     description: "Cabal version"
     required: false
-    default: "3.14.2.0"
+    default: "3.16.0.0"
```
fendor: This should be bumped to 3.16.1.0

cabal.project
fendor: Unfortunately, `alga` requires changes and can't be revised on hackage. While this is rather unfortunate, we will have to release with this `source-repository-package` stanza, as we are otherwise blocked for an unknown amount of time.

Kleidukos: @fendor It's perfectly fine to vendor the patched `alga` in the HLS tree for the time being and remove it after the fix has been released. This allows people to install HLS from its Hackage sdist without trouble. :)

benz0li:
FYI Building HLS 2.12.0.0 using GHC 9.12.3 fails, too:

- x86_64: https://gitlab.b-data.ch/haskell/hlssi/-/jobs/229672
- AArch64: https://gitlab.b-data.ch/haskell/hlssi/-/jobs/229673

Will this PR also fix this?

fendor:
@benz0li I don't think so, the failed CI jobs point to `ghc-lib-parser` being incompatible with ghc 9.12.3.

dschrempf:
According to https://discourse.haskell.org/t/critical-code-generation-bug-with-ghc-9-12-3/13505 we should avoid using GHC 9.12.3 altogether. (Maybe irrelevant).

benz0li:
> @benz0li I don't think so, the failed CI jobs point to `ghc-lib-parser` being incompatible with ghc 9.12.3.

True. Then, I wonder why it picks [`ghc-lib-parser-9.12.2.20250421`](https://hackage.haskell.org/package/ghc-lib-parser-9.12.2.20250421) when there is [`ghc-lib-parser-9.12.3.20251228`](https://hackage.haskell.org/package/ghc-lib-parser-9.12.3.20251228) available.

fendor:
@benz0li Ah, perhaps the index-state needs to be updated? That might be fixed by this PR.

benz0li:
> According to https://discourse.haskell.org/t/critical-code-generation-bug-with-ghc-9-12-3/13505 we should avoid using GHC 9.12.3 altogether. (Maybe irrelevant).

Cross reference:

* https://discourse.haskell.org/t/ghc-9-12-3-is-now-available/13454/10

.github/workflows/test.yml
```diff
@@ -105,162 +107,162 @@ jobs:
...
-        run: cabal build all
+        run: cabal build ${CABAL_ARGS} all
 
       - name: Set test options
         # See https://github.com/ocharles/tasty-rerun/issues/22 for why we need
         # to include 'new' in the filters, since many of our test suites are in the
         # same package.
         run: |
-          cabal configure --test-options="--rerun-update --rerun-filter failures,exceptions,new"
+          cabal configure ${CABAL_ARGS} --test-options="--rerun-update --rerun-filter failures,exceptions,new"
```
jhrcek: @fendor Instead of repeating cabal args in all the command (seems like too much boilerplate)
you can just set it once in this `cabal configure` step - it saves those args in cabal.project.local, so it's used by all subsequent cabal invocations.

If you then move the cabal configure step to run before `cabal build all`, we could have all the cabal options setting done in one place and then reused by all the following cabal invocations..

fendor: Thanks, totally forgot about that!

fendor: I think we can't easily specify the builddir in the `cabal.project` file due to a missing cabal feature.

Issue #4788, opened 2026-01-16
Cli mode resolves files relative to git root instead of current working directory


Aster89:
Originally reported on element. @fendor's comment: 

> https://github.com/haskell/haskell-language-server/pull/4736 caused unexpected behaviour for a user; initial file paths should not be interpreted relative to the git root, but to the current working directory.

### Your environment

Which OS do you use?
up-to-date ArchLinux

Which version of GHC do you use and how did you install it?
9.10.3 from GHCup

How is your project built (alternative: link to the project)?
https://github.com/sellout/pathway

Which LSP client (editor/plugin) do you use?
Terminal Vim + YCM

Which version of HLS do you use and how did you install it?
2.12.0.0 from GHCup

Have you configured HLS in any way (especially: a `hie.yaml` file)?
Not that I know of.

### Steps to reproduce

```bash
cd ~
git clone https://github.com/sellout/pathway.git
cd pathway # into the repo/project
cd pathway # into the homonymous package
cabal build -w ghc-9.10.3 # not sure if needed, but why not, so you see it does build successfully
haskell-language-server-wrapper --debug src/Data/Path.hs 
```

### Expected behaviour

Output should end successfully like this:
```none
Completed (1 file worked, 0 files failed)
```

### Actual behaviour

Error:
```
No 'hie.yaml' found. Try to discover the project type!
Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 2.12.0.0 x86_64 ghc-9.10.3
Current directory: /home/enrico/pathway/pathway
Operating system: linux
Arguments: ["--debug","src/Data/Path.hs"]
Cradle directory: /home/enrico/pathway
Cradle type: Cabal

Tool versions found on the $PATH
cabal:          3.14.2.0
stack:          3.7.1
ghc:            9.10.3
1                                                                                                                                                                                              


Consulting the cradle to get project GHC version...
2026-01-16T13:04:37.994574Z | Debug | cabal --numeric-version
2026-01-16T13:04:38.008109Z | Debug | cabal path --output-format=json
2026-01-16T13:04:38.035360Z | Debug | /home/enrico/.ghcup/bin/ghc-9.10.3 --numeric-version
Project GHC version: 9.10.3
haskell-language-server exe candidates: ["haskell-language-server-9.10.3","haskell-language-server"]
Launching haskell-language-server exe at:/home/enrico/.ghcup/bin/haskell-language-server-9.10.3
[...]

PR #4790, opened 2026-01-17
fix: git sub-repository files listing (#4788)

blackheaven:
Tested locally.

### DIFF
--- a/exe/Wrapper.hs
+++ b/exe/Wrapper.hs
@@ -132,7 +132,13 @@ launchHaskellLanguageServer recorder parsedArgs = do
   hPutStrLn stderr $ "Current directory: " ++ d
   hPutStrLn stderr $ "Operating system: " ++ os
   args <- getArgs
-  hPutStrLn stderr $ "Arguments: " ++ show args
+  let args' = case parsedArgs of
+        Ghcide GhcideArguments{..}
+            -- If no --cwd is given, assume the current working directory as the --cwd
+           | isNothing argsCwd -> args ++ ["--cwd", argsInitialWorkingDirectory]
+        _ -> args
+
+  hPutStrLn stderr $ "Arguments: " ++ show args'
   hPutStrLn stderr $ "Cradle directory: " ++ cradleRootDir cradle
   hPutStrLn stderr $ "Cradle type: " ++ show (actionName (cradleOptsProg cradle))
   programsOfInterest <- findProgramVersions
@@ -162,7 +168,7 @@ launchHaskellLanguageServer recorder parsedArgs = do
           liftIO $ hPutStrLn stderr $ "Launching haskell-language-server exe at:" ++ e
 
 #ifdef mingw32_HOST_OS
-          liftIO $ callProcess e args
+          liftIO $ callProcess e args'
 #else
 
           let Cradle { cradleOptsProg = CradleAction { runGhcCmd } } = cradle
@@ -177,7 +183,7 @@ launchHaskellLanguageServer recorder parsedArgs = do
 
           env <- Map.fromList <$> liftIO getEnvironment
           let newEnv = Map.insert "GHC_BIN" ghcBinary $ Map.insert "GHC_LIBDIR" libdir env
-          liftIO $ executeFile e True args (Just (Map.toList newEnv))
+          liftIO $ executeFile e True args' (Just (Map.toList newEnv))
 #endif
 
 
--- a/src/Ide/Arguments.hs
+++ b/src/Ide/Arguments.hs
@@ -22,6 +22,7 @@ import           Ide.Logger                    (Priority (..))
 import           Ide.Types                     (IdePlugins)
 import           Options.Applicative
 import           Paths_haskell_language_server
+import           System.Directory              (getCurrentDirectory)
 import           System.Environment
 
 -- ---------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


blackheaven:
@fendor @Aster89 I have fixed it, but it seems to be an issue existing before my change.

src/Ide/Arguments.hs
```diff
@@ -38,18 +39,19 @@ data Arguments
...
     -- ^ the minimum log level to show
-    , argsLogStderr         :: Bool
-    , argsLogClient         :: Bool
-    , argsThreads           :: Int
-    , argsProjectGhcVersion :: Bool
+    , argsLogStderr               :: Bool
+    , argsLogClient               :: Bool
+    , argsThreads                 :: Int
+    , argsProjectGhcVersion       :: Bool
+    , argsInitialWorkingDirectory :: FilePath
```
Aster89, line 54: Out of curiosity, am I correct that all these `args*` prefixes could be dropped, at the cost of `import qualified ...... as Args` in the files that need these and prepending `Args.` to the usages?

exe/Wrapper.hs
```diff
@@ -132,7 +132,12 @@ launchHaskellLanguageServer recorder parsedArgs = do
   hPutStrLn stderr $ "Current directory: " ++ d
   hPutStrLn stderr $ "Operating system: " ++ os
   args <- getArgs
-  hPutStrLn stderr $ "Arguments: " ++ show args
+  let args' = case parsedArgs of
+        Ghcide GhcideArguments{..}
```
Aster89: Just noticing that this pattern of `case parsedArgs of Ghcide GhcideArguments{..}` is already used a couple of times in this function in the lines above. I wonder if there's a way to avoid doing it multiple times?

src/Ide/Arguments.hs
fendor: I think so, but I think such a refactoring only makes sense if we employ some kind of code-style for all records in HLS or at least ghcide. Changing it only here feels like it would serve no purpose.

fendor:
@blackheaven Feel free to ignore the test - failures, they are caused by hard-to-reproduce issues on windows.

exe/Wrapper.hs
blackheaven: We could but it would made the logic even more brittle, currently we have:

* A first `case` changing (eventually) the cwd
* Some logic fetching the `cradle` and moving to root
* A second `case` checking GHC version depending on `cradle` (previous step)
* Some logic which get `args`/`progName`
* A third `case` (mine) which "fix" `args` (previous step)

I can merge my `case` in one of the two others, but it'll increase temporal coupling.

Let me know what you think.

Aster89: Sorry for not being clear the first time. Re-reading my message, I see I've worded it very poorly.

I didn't mean to suggest _merging_ the 3 `case`s, but actually to pinpoint some abtraction that would allow expressing those 3 cases in a more concise way.

If you look at the first two usages, adjusting the whitespace:
```haskell
{-1-}  case parsedArgs of
{-2-}    Ghcide GhcideArguments{..} ->
{-M-}      whenJust argsCwd setCurrentDirectory
{-3-}    _ -> pure ()

{-1-}  case parsedArgs of
{-2-}    Ghcide GhcideArguments{..} ->
{-M-}      when argsProjectGhcVersion $ do
{-M-}        runExceptT (getRuntimeGhcVersion' cradle) >>= \case
{-M-}          Right ghcVersion -> putStrLn ghcVersion >> exitSuccess
{-M-}          Left err -> T.putStrLn (prettyError err NoShorten) >> exitFailure
{-3-}    _ -> pure ()
```
the lines I marked as 1, 2, and 3 are the same in the 3 cases, so I was wondering if one could write just
```haskell
parsedArgs >={-someoperator-}=> do {-MMM-}
```

Maybe `>={-someoperator-}=>` is actually some operator that already exists?

Anyway, only now I realize the 3rd usage you added doesn't exactly map to those 2, as the default case doesn't return `pure ()` but `args`, which is captured.

So... I guess you can just drop this suggestion? :smile:

blackheaven: I don't mind creating a `whenGhcide`, but it is not really relted, WDYT @fendor ?

fendor: Personally, I dont think this amount of code duplication warrants a refactor, unless already included in a refactor.
So, I would keep it as is, but Ill leave it up to you. It is a bug fix after all, and bug fixes should preferably not be refactorings :)

The operator in question is likely a Prism traversal, but such a refactoring doesn't *feel* like a huge readability improvement 😆

blackheaven: LGTM, thanks.

exe/Wrapper.hs
```diff
@@ -132,7 +132,12 @@ launchHaskellLanguageServer recorder parsedArgs = do
...
+           | isNothing argsCwd -> args ++ ["--cwd", argsInitialWorkingDirectory]
+        _ -> args
```
fendor:
```suggestion
  let argsWithCwd = case parsedArgs of
        Ghcide GhcideArguments{..}
           -- If no --cwd is given, assume the current working directory as the --cwd
           | isNothing argsCwd -> args ++ ["--cwd", argsInitialWorkingDirectory]
        _ -> args
```

I am a little confused why we add `argsInitialWorkingDirectory` to `Arguments` if we only need it here and in `Arguments` it is not really an argument but a constant. Wouldn't it make more sense to query here the `getCurrentDirectory` and provide it only if `argsCwd` is `Nothing`?

blackheaven: It would, but the directory has been changed in the meantime.

fendor: It should only be changed if `--cwd` has been given in which case we are not changing it any way, right?

blackheaven: Yes but, the directory would already has been changed to `cradle`/`cabal.project` regardless `--cwd`.

fendor: Ah, I see, you mean we change the `cwd` here:

https://github.com/haskell/haskell-language-server/pull/4790/changes#diff-c439338faccfcf78b6713535ed01a419e4c01852e7c94004e1bec2d536098642R119

I guess I still don't 100% understand why we couldn't query the `cwd` here: https://github.com/haskell/haskell-language-server/pull/4790/changes#diff-c439338faccfcf78b6713535ed01a419e4c01852e7c94004e1bec2d536098642R109 and then just use that in https://github.com/haskell/haskell-language-server/pull/4790/changes#diff-c439338faccfcf78b6713535ed01a419e4c01852e7c94004e1bec2d536098642R138 ?

Again, this mostly avoids a somewhat unused `Arguments` field, so not really a big deal.
If you think this is better as is, we'll merge as is once the compile time errors are fixed.

blackheaven: I agree that it's confusing, the end goal is to act on https://github.com/haskell/haskell-language-server/blob/b3e29e1f63f24ce739773dd4d142adc592199ace/src/Ide/Main.hs#L129 so, with the correct cwd, the given file path is resolved correctly.

blackheaven:
Windows build is still broken.

PR #4808, opened 2026-01-24
Bump hlint action, move options to .hlint.yaml

### LINKED ISSUE #4806: Move `--ignore-glob` to `.hlint.yaml`
Can we put these `--ignore-glob` exclusions into `.hlint.yaml` so that running `hlint -j .` is the same as the hlint CI action?

https://github.com/haskell/haskell-language-server/blob/b3e29e1f63f24ce739773dd4d142adc592199ace/.github/workflows/hlint.yml#L23

philderbeast:
Fixes #4806. Bumps hlint to `3.10`, moves options to `.hlint.yaml`. Rerun `hlint --default` and merge.

Kept warning and not ignoring this suggestion:

```yaml
- warn: {name: "Use module export list"} # 86 hints
```

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -6,29 +6,56 @@
 # $ hlint --git -j4
 
 # Warnings currently triggered by our code
-- ignore: {name: "Use <$>"}
-- ignore: {name: "Use :"}
-- ignore: {name: "Redundant do"}
-- ignore: {name: "Avoid lambda"}
-- ignore: {name: "Use newtype instead of data"}
-- ignore: {name: "Use unless"}
-- ignore: {name: "Move brackets to avoid $"}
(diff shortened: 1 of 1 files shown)


.hlint.yaml
```diff
@@ -236,3 +263,11 @@
 # We really do not want novel usages of restricted functions, and mere
 # Warning is not enough to prevent those consistently; you need a build failure.
 - error: {name: Avoid restricted function}
+
+- group: {name: extra, enabled: true}
```
philderbeast, line 267: This is the `--with-group` option from the gh-action invocation that I've moved to `.hlint.yaml`:

https://github.com/haskell/haskell-language-server/blob/b3e29e1f63f24ce739773dd4d142adc592199ace/.github/workflows/hlint.yml#L23

Issue #4782, opened 2026-01-05
Expanding Record Wildcards expands the outermost record


fendor:
<img width="1369" height="585" alt="Image" src="https://github.com/user-attachments/assets/5ec00cc1-195b-4ff0-8de9-deb871a9f866" />

Turns the record into 

<img width="1423" height="324" alt="Image" src="https://github.com/user-attachments/assets/2b9ec739-e67f-4a7e-a2e2-3e8a526e721f" />

Which is not what I expected! I wanted to expand the record I was clicking on. 
Applying the code action again has the behaviour I want, but now I need to undo the inner expansion!

It would be better if the code action expands exactly the record under the cursor, e.g., in this case this is what I would have wanted: (applied manually, so ignore the formatting)

<img width="1262" height="702" alt="Image" src="https://github.com/user-attachments/assets/2b74163e-4b2f-4857-a33c-b1ebd557aa96" />

GHC 9.10.3
HLS 2.12

agam263:
Hi, can I please be assigned this issue? I’m a first-time contributor and would like to work on it.
​

fendor:
Excited to hear that!

I have assigned this issue to you, please give us a shout here or in our [matrix channel](https://matrix.to/#/#haskell-language-server:matrix.org) if you get stuck or need help.

agam263:
Thank you for assigning this issue! I’ll start working on it and will reach out here or on the Matrix channel if I need any help.

fendor:
Closed by #4813

PR #4813, opened 2026-01-26
Cursor aware record expansion

vidit-od:
Another PR to fix issue #4782. Previously PR #4807 was raised but later discarded due to the following main reason. 
- Getting Multiple suggestions increases cognitive load to identify what actually to click 
- The name at the beginning do not help in improving user experience. 

explicitRecord

This is a Small PR aiming to introduce Cursor aware record expansion.
- users just need to trigger code action 
- gets just one suggestion closest to cursor position. 
- does not require to expand any other records to expand current record as experienced in original issue.

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -19,6 +19,7 @@ import           Control.Monad.IO.Class               (MonadIO (liftIO))
 import           Control.Monad.Trans.Class            (lift)
 import           Control.Monad.Trans.Maybe
 import           Data.Aeson                           (ToJSON (toJSON))
+import           Data.Function                        (on)
 import           Data.Generics                        (GenericQ, everything,
                                                        everythingBut, extQ, mkQ)
 import qualified Data.IntMap.Strict                   as IntMap
@@ -27,7 +28,6 @@ import           Data.List                            (find, intersperse,
 import qualified Data.Map                             as Map
 import           Data.Maybe                           (fromMaybe, isJust,
                                                        mapMaybe, maybeToList)
-import           Data.Ord                             (Down (..))
 import           Data.Text                            (Text)
 import qualified Data.Text                            as T
 import           Data.Unique                          (hashUnique, newUnique)
@@ -165,12 +165,13 @@ codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
                       -- converted to the record syntax through the code action
                       , isConvertible record
                       ]
-      sortedRecords = sortOn (Down . recordDepth recordsWithUid . snd) recordsWithUid
+      recordsOnly = map snd recordsWithUid
+      sortedRecords = sortOn (recordDepth recordsOnly . snd) recordsWithUid
   pure $ InL $ case sortedRecords of
-    (top : _) -> [mkCodeAction enabledExtensions top]
+    (top : _) -> [mkCodeAction enabledExtensions (fst top)]
     []        -> []
   where
-    mkCodeAction exts (uid, _record) = InR CodeAction
+    mkCodeAction exts uid = InR CodeAction
       { _title = mkTitle exts -- TODO: `Expand positional record` without NamedFieldPuns if RecordInfoApp
       , _kind = Just CodeActionKind_RefactorRewrite
       , _diagnostics = Nothing
@@ -303,10 +304,12 @@ mkTitle exts = "Expand record wildcard"
                    then mempty
                    else " (needs extension: NamedFieldPuns)"
 
-recordDepth :: [(Int, RecordInfo)] -> RecordInfo -> Int
+-- Calculate the nesting depth of a record by counting how many other records
+-- contain it. Used to prioritize more deeply nested records in code actions.
+recordDepth :: [RecordInfo] -> RecordInfo -> Int
 recordDepth allRecords record =
-  let r = recordInfoToRange record
-  in length [ ()| (_, other) <- allRecords, let r' = recordInfoToRange other, subRange r' r]
+  let isSubrangeOf = subRange `on` recordInfoToRange
+  in length $ filter (`isSubrangeOf` record) allRecords
 
 pragmaEdit :: [Extension] -> NextPragmaInfo -> Maybe TextEdit
 pragmaEdit exts pragma = if NamedFieldPuns `elem` exts
--- a/plugins/hls-explicit-record-fields-plugin/test/Main.hs
+++ b/plugins/hls-explicit-record-fields-plugin/test/Main.hs
@@ -38,6 +38,8 @@ test = testGroup "explicit-fields"
     , mkTestNoAction "Prefix" "Prefix" 10 11 10 28
     , mkTestNoAction "PartiallyAppliedCon" "PartiallyAppliedCon" 7 8 7 12
     , mkTest "PolymorphicRecordConstruction" "PolymorphicRecordConstruction" 15 5 15 15
+    , mkTest "CursorAwarePositional" "CursorPositional" 15 26 15 34
+    , mkTest "CursorAwareRecords" "CursorRecords" 9 40 9 40
     ]
   , testGroup "inlay hints"
     [ mkInlayHintsTest "Construction" Nothing 16 $ \ih -> do
--- a/plugins/hls-explicit-record-fields-plugin/test/testdata/CursorPositional.expected.hs
+++ b/plugins/hls-explicit-record-fields-plugin/test/testdata/CursorPositional.expected.hs
@@ -1,0 +1,16 @@
+module CursorPositional where
+data RecOuter = RecOuter{
+  middle :: RecMiddle
+}
+
+data RecMiddle = RecMiddle {
+  inner :: RecInner
+}
+
+data RecInner = RecInner{
+  foo   :: Char
+  , bar :: Int
+}
+
+ex :: RecOuter
+ex = RecOuter (RecMiddle (RecInner { foo = 'c', bar = 42 }))
--- a/plugins/hls-explicit-record-fields-plugin/test/testdata/CursorRecords.expected.hs
+++ b/plugins/hls-explicit-record-fields-plugin/test/testdata/CursorRecords.expected.hs
@@ -1,0 +1,11 @@
+{-# LANGUAGE RecordWildCards #-}
+module CursorRecords where
+
+data L4 = L4{ a0 :: Int}
+data L3 = L3{ a1 :: Int, a11 :: Int, l4 :: L4}
+data L2 = L2{ a2 :: Int, l3 :: L3}
+data L1 = L1{ a3 :: Int, l2 :: L2}
+
+test :: L1 -> Int
+test L1 {l2 = L2{ l3 = L3 {l4 = L4 {..}, a1, a11}, ..}, ..} =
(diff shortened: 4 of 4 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -296,6 +299,13 @@ mkTitle exts = "Expand record wildcard"
                    then mempty
                    else " (needs extension: NamedFieldPuns)"
 
+containsRange :: Range -> Range -> Bool
+containsRange (Range s1 e1) (Range s2 e2) = s1 <= s2 && e2 <= e1 && (s1 /= s2 || e1 /= e2)
```
ozkutuk: This seems to be identical to [`Ide.PluginUtils.subRange`](https://github.com/haskell/haskell-language-server/blob/4eefbff4e48765dc4f2c4904dde80b0875b6cbf3/hls-plugin-api/src/Ide/PluginUtils.hs#L278-L279).

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
vidit-od: This Line is the culprit, This returns True whenever a positional constructor is found and does not go inside. 

if this is changed to `in (recInfo, False)` i.e hardcode false. 

Then traversal will continue for inside expressions and we get expected wildcard expansions for expected inner constructor !! 

I'm unclear why `not (null recInfo)` was used originally ? Is there some edge case i am not considering ?

ozkutuk: Changing the right component of the tuple to be `False` seems to break a few inlay hint tests, so looking into why they break might be a good starting point. Apart from that, maybe @jetjinser could explain why the implementation is the way it is, since IIRC they are the one who implemented the inlay hints feature.

vidit-od: Yes i also encountered this same problem. I digged deeper and found 
-  the tests fail because of crestion of multiple unnecessary inhaly hints that are unsaturated in nature. 
-  i.e we are creating hints for positional constructor that are kind of not Complete yet (wrt cursor ) 

Example : 
`
Recouter ( a1, Recinner (a2,a3,a4))
`
Here we expect 2 hints (1 outer and 1 inner)
But we will get 7. 

Solution : 
We already maintain a saturated flag in our code, we just require a minor change to tackle this case. This will pass all the testcases. 

I have already experimented with this changes and wrote a few new extra testcases to check this nested behaviour. Will update the PR in some time !

P.s. : hopefully i did a decent job explaining the reasoning , also thanks a lot for the review !

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -266,9 +270,12 @@ inlayHintPosRecProvider _ state _pId InlayHintParams {_textDocument = TextDocume
...
      mkInlayHints :: UniqFM Name [Name] -> PositionMapping -> RecordInfo -> [InlayHint]
-     mkInlayHints nameMap pm record@(RecordInfoApp _ (RecordAppExpr _ _ fla)) =
-       let textEdits = renderRecordInfoAsTextEdit nameMap record
-       in mapMaybe (mkInlayHint textEdits pm) fla
+     mkInlayHints nameMap pm record@(RecordInfoApp _ (RecordAppExpr sat _ fla)) =
+       -- Only create inlay hints for fully saturated constructors
+       case sat of
+         Saturated -> let textEdits = renderRecordInfoAsTextEdit nameMap record
+                      in mapMaybe (mkInlayHint textEdits pm) fla
+         Unsaturated -> []
```
vidit-od, line 279: This is the Saturated Flag i was talking about in the Previous Conversation. With this change, we get the expected Cursor Oriented Expansions behavior and All the original tests Pass. 

On top of that 2 new type of Testcases were added to check Nested behavior on both Records and Positional syntax. 

Requesting a Re-Review on the Updated PR @ozkutuk @fendor

plugins/hls-explicit-record-fields-plugin/test/Main.hs
```diff
@@ -292,6 +292,89 @@ test = testGroup "explicit-fields"
...
+              , _data_ = Nothing
+              }
+          ]
+    , mkInlayHintsTest "CursorPositional" Nothing 15 $ \ih -> do
+        let mkLabelPart' = mkLabelPartOffsetLengthSub1 "CursorPositional"
+        middle <- mkLabelPart' 2 2 "middle="
+        inner  <- mkLabelPart' 6 2 "inner="
+        foo    <- mkLabelPart' 10 2 "foo="
+        bar    <- mkLabelPart' 11 4 "bar="
+        (@?=) ih
```
fendor, line 348:
```suggestion
         ih @?=
```

Infix notation

plugins/hls-explicit-record-fields-plugin/test/Main.hs
```diff
@@ -292,6 +292,89 @@ test = testGroup "explicit-fields"
...
                          }
           ]
+    , mkInlayHintsTest "CursorRecords" Nothing 9 $ \ih -> do
+        let mkLabelPart' = mkLabelPartOffsetLength "CursorRecords"
+        a0  <- mkLabelPart' 3 14 "a0"
+        a1  <- mkLabelPart' 4 14 "a1"
+        a11 <- mkLabelPart' 4 25 "a11"
+        a2  <- mkLabelPart' 5 14 "a2"
+        a3  <- mkLabelPart' 6 14 "a3"
+        (@?=) ih
```
fendor, line 304:
```suggestion
        ih @?=
```

Infix notation again

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -155,18 +158,19 @@ codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
...
                       , Just record <- [IntMap.lookup uid crCodeActionResolve]
                       -- Only fully saturated constructor applications can be
                       -- converted to the record syntax through the code action
                       , isConvertible record
                       ]
-  let actions = map (mkCodeAction enabledExtensions) recordUids
-  pure $ InL actions
+      sortedRecords = sortOn (Down . recordDepth recordsWithUid . snd) recordsWithUid
+  pure $ InL $ case sortedRecords of
+    (top : _) -> [mkCodeAction enabledExtensions top]
```
fendor:
```suggestion
    (top : _) -> [mkCodeAction enabledExtensions (fst top)]
```

Rather than ignoring the parameter in `mkCodeAction`

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -296,6 +303,10 @@ mkTitle exts = "Expand record wildcard"
                    then mempty
                    else " (needs extension: NamedFieldPuns)"
 
+recordDepth :: [(Int, RecordInfo)] -> RecordInfo -> Int
```
fendor: Docs please :)

One sentence about its purpose is sufficient!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -155,18 +158,19 @@ codeActionProvider ideState _ (CodeActionParams _ _ docId range _) = do
...
+  let recordsWithUid = [ (uid, record)
                       | uid <- RangeMap.filterByRange range crCodeActions
                       , Just record <- [IntMap.lookup uid crCodeActionResolve]
                       -- Only fully saturated constructor applications can be
                       -- converted to the record syntax through the code action
                       , isConvertible record
                       ]
-  let actions = map (mkCodeAction enabledExtensions) recordUids
-  pure $ InL actions
+      sortedRecords = sortOn (Down . recordDepth recordsWithUid . snd) recordsWithUid
```
fendor: This `Down` seems to be incorrect, I had to remove it to get the record expansion behaviour as the issue suggested.

I think there explicitly needs to be a test for `Expand record wildcard` code action, so this doesn't accidentally regress.

fendor (changes requested):
Looks pretty good to me, one more test please and a couple of nitpicks!

plugins/hls-explicit-record-fields-plugin/test/Main.hs
vidit-od: All of the Previous Testcases also do not use infix notation. 
should i change them as well ?

fendor: Nope, sorry, didn't see that, please continue doing this then!

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
jetjinser: > Changing the right component of the tuple to be `False` seems to break a few inlay hint tests, so looking into why they break might be a good starting point. Apart from that, maybe @jetjinser could explain why the implementation is the way it is, since IIRC they are the one who implemented the inlay hints feature.

Apologies for the late reply.

Thank you to @vidit-od for the explanation—it's clear that the code is now better understood from that perspective. Having written it quite some time ago, I found the explanation helpful for recalling the context.

In any case, feel free to adjust as considered appropriate; there are no deeper reasons behind it.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -296,6 +302,12 @@ mkTitle exts = "Expand record wildcard"
                    then mempty
                    else " (needs extension: NamedFieldPuns)"
 
+-- Calculate the nesting depth of a record by counting how many other records
+-- contain it. Used to prioritize more deeply nested records in code actions.
+recordDepth :: [(Int, RecordInfo)] -> RecordInfo -> Int
+recordDepth allRecords record =
+  let r = recordInfoToRange record
+  in length [ ()| (_, other) <- allRecords, let r' = recordInfoToRange other, subRange r' r]
```
ozkutuk: The `uid` information is not used within function, so it might be clearer to eliminate that from the parameter. This is probably how I would write this:

```haskell
import Data.Function (on)

recordDepth :: [RecordInfo] -> RecordInfo -> Int
recordDepth allRecords record =
  let isSubrangeOf = subRange `on` recordInfoToRange
  in length $ filter (`isSubrangeOf` record) allRecords

vidit-od:
Done, Hopefully everything is covered now. 

Thanks a lot for reviewing and helping me on this PR @ozkutuk @fendor . 

Also a thing to consider for PR #4819 . This PR introduces a few extra testcases that still follows "Record wildcard expansion" format. They will require change.

ozkutuk:
@vidit-od Looking good! I'll leave it to @fendor to push the merge button.

> Also a thing to consider for PR https://github.com/haskell/haskell-language-server/pull/4819 . This PR introduces a few extra testcases that still follows "Record wildcard expansion" format. They will require change.

Yes, I will rebase that PR on top of your work once it gets merged, but thanks for the heads-up.

PR #4819, opened 2026-01-31
Distinguish between wildcard expansion and traditional record syntax

ozkutuk:
While reviewing #4813, I realized that the code action for the expansion of positional records was producing the message "Expand record wildcard". This must be confusing for users on multiple levels, since:
1. the actual effect is not at all related to wildcards
2. depending on the set of enabled extensions the message may have "needs extension: NamedFieldPuns" suffix, which is not correct

This PR aims to fix this by producing different titles for the code action depending on whether it is a wildcard expansion or "positional record expansion". I have opted to rephrase the latter as "Convert to traditional record syntax", both to be consistent with the GADT conversion code action, and also with [`-XTraditionalRecordSyntax`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/traditional_record_syntax.html#traditional-record-syntax).

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -114,7 +114,7 @@ import           Language.LSP.Protocol.Message        (Method (..),
 import           Language.LSP.Protocol.Types          (CodeAction (..),
                                                        CodeActionKind (CodeActionKind_RefactorRewrite),
                                                        CodeActionParams (CodeActionParams),
-                                                       InlayHint (..),
+                                                       Command, InlayHint (..),
                                                        InlayHintLabelPart (InlayHintLabelPart),
                                                        InlayHintParams (InlayHintParams, _range, _textDocument),
                                                        TextDocumentIdentifier (TextDocumentIdentifier),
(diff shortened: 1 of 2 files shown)


ozkutuk:
@fendor I have rebased the PR on top of `master`, and it is ready to merge from my perspective.

fendor:
Thanks!

In the future, you should also be able to merge, right?
We should probably document our policies, I think you should have the autonomy to merge plugin changes at your own responsibility... Perhaps a topic for the next HLS meeting.

ozkutuk:
> In the future, you should also be able to merge, right?

@fendor Yes, there were no technical blockers for me to merge, I was just being cautious (perhaps overly so) by leaving the final decision to a more "authoritative" maintainer.

> We should probably document our policies, I think you should have the autonomy to merge plugin changes at your own responsibility... Perhaps a topic for the next HLS meeting.

Sounds like a good idea!

fendor:
> I was just being cautious (perhaps overly so) by leaving the final decision to a more "authoritative" maintainer.

I understand that and it makes sense to me! I am a little worried, we have few "authoritative" maintainers left at the moment, rendering us kind of reliant on these few maintainers, waiting for their approval...
So we need to empower the maintainers we have to at least act with authority over the parts they "own". Hence, making sure that you do have the permissions to actually merge if no one else was around.

PR #4824, opened 2026-02-03
docs: add WSL build location note for contributors

Siya-05:
This adds a short note to the contributing guide for users building HLS under WSL.

Building from Windows-mounted paths (e.g. /mnt/c) can cause permission issues, performance problems, or memory-related build failures. Recommending cloning into the Linux filesystem (e.g. ~/dev) helps avoid these issues, especially for new contributors.

This was discussed and confirmed in the related issue.

### DIFF
--- a/docs/contributing/contributing.md
+++ b/docs/contributing/contributing.md
@@ -15,6 +15,18 @@ Clone the repository:
 $ git clone https://github.com/haskell/haskell-language-server
 ```
 
+#### Note for contributors using WSL (Windows Subsystem for Linux)
+
+When building HLS in WSL, clone and build the repository from the Linux filesystem (e.g. `~/dev`), not from Windows-mounted paths like `/mnt/c`, as this can cause permission or build issues.
+
+```bash
+cd ~
+mkdir -p dev
+cd dev
+git clone https://github.com/haskell/haskell-language-server
+cd haskell-language-server
+```
+
 The project can then be built with both `cabal build` and `stack build`.
 
 ### Building with Cabal
@@ -25,19 +37,7 @@ $ cabal update
 # Then
 $ cabal build
 ```
-#### Note for contributors using WSL (Windows Subsystem for Linux)
 
-If you are building Haskell Language Server inside WSL, it is recommended to clone and build the repository from the Linux filesystem (for example, `~/dev`) rather than from Windows-mounted paths such as `/mnt/c` or `/mnt/p`.
-
-Building from Windows-mounted directories can cause permission issues, degraded performance, or unexpected build failures—including builds being terminated—especially on low-memory systems.
-
-```bash
-cd ~
-mkdir -p dev
-cd dev
-git clone https://github.com/haskell/haskell-language-server
-cd haskell-language-server
-```
 If you encounter memory-related build failures, try reducing peak memory usage by running:
 ```
 cabal build -j1


docs/contributing/contributing.md
```diff
@@ -25,6 +25,24 @@ $ cabal update
 # Then
 $ cabal build
 ```
+#### Note for contributors using WSL (Windows Subsystem for Linux)
+
+If you are building **Haskell Language Server** inside WSL, it is strongly recommended to clone and build the repository from the Linux filesystem (for example, `~/dev`) rather than from Windows-mounted paths such as `/mnt/c` or `/mnt/p`.
+
+Building from Windows-mounted directories can cause permission issues, degraded performance, or unexpected build failures—including builds being terminated—especially on low-memory systems.
+
+##### Recommended setup
```
fendor:
```suggestion
```

docs/contributing/contributing.md
```diff
@@ -25,6 +25,24 @@ $ cabal update
...
+```bash
+cd ~
+mkdir -p dev
+cd dev
+git clone https://github.com/haskell/haskell-language-server
+cd haskell-language-server
+```
+If you encounter memory-related build failures, try reducing peak memory usage by running:
+```
+cabal clean && cabal build --disable-tests -j1
```
fendor:
```suggestion
cabal build -j1
```

clean seems optional to me.

Also, disabling tests shouldn't really matter, right?

fendor: This part can stay for cabal.

docs/contributing/contributing.md
```diff
@@ -25,6 +25,24 @@ $ cabal update
 # Then
 $ cabal build
 ```
+#### Note for contributors using WSL (Windows Subsystem for Linux)
+
+If you are building **Haskell Language Server** inside WSL, it is strongly recommended to clone and build the repository from the Linux filesystem (for example, `~/dev`) rather than from Windows-mounted paths such as `/mnt/c` or `/mnt/p`.
```
fendor:
```suggestion
If you are building Haskell Language Server inside WSL, it is recommended to clone and build the repository from the Linux filesystem (for example, `~/dev`) rather than from Windows-mounted paths such as `/mnt/c` or `/mnt/p`.
```

nowhere else highlighted

docs/contributing/contributing.md
```diff
@@ -25,6 +25,24 @@ $ cabal update
...
+
+Building from Windows-mounted directories can cause permission issues, degraded performance, or unexpected build failures—including builds being terminated—especially on low-memory systems.
```
fendor: Do you have an idea why this is the case? I am mostly building HLS on windows directly, not via the WSL, so I don't know what's going on here? Is WSL broken?

fendor (changes requested):
Thanks for the contribution! 

I think this section should be right after cloning the repo, e.g. here https://github.com/SivapriyaVenkateswarar/haskell-language-server/blob/7a1ce5114576162ef6d748476eedee23c3f9476f/docs/contributing/contributing.md#building

Since this is not cabal specific, afaict?

Also, it is important to keep it brief, so a couple of suggestions to reduce the size

docs/contributing/contributing.md
Siya-05: Hey, thanks for all the suggestions, and extremely sorry for the delayed reply!

from what I found, this doesn’t look like a WSL bug itself. It’s a known issue with building on Windows-mounted paths (/mnt/*) under WSL.

Cabal does a lot of small file operations, and on /mnt/* those operations go through the Windows filesystem, making it much slower and can increase memory usage. On low-RAM systems, that can even cause builds to get killed.

Cloning the repo into the Linux home directory (like ~/dev) uses WSL’s native filesystem, which is why builds there are usually faster and more reliable.

docs/contributing/contributing.md
Siya-05: Sure, I changed it.
Thank you!

docs/contributing/contributing.md
Siya-05: Thanks for pointing this out as well.

I added --disable-tests because I was personally running into OOM/SIGKILL issues during the build, and skipping tests reduced peak memory usage on my machine. That said, you’re right, this is cabal-specific and probably unnecessary here.

fendor (approved):
LGTM, thank you!

PR #4828, opened 2026-02-05
fix notes crashing and leaking single comments notes

vidit-od:
This PR fixes : 
1) Issue #4763 
2) Leaking Single Comment Notes Declaration 
singleLine Leak

here the definatination declaration is : 
singleLine Leak Def

### DIFF
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -132,9 +132,14 @@ listReferences state _ param
             Nothing -> pure (InR Null)
             Just note -> do
                 notes <- runActionE "notes.definedNoteReferencess" state $ useE MkGetNoteReferences nfp
-                poss <- err ("Note reference (a comment of the form `{- Note [" <> note <> "] -}`) not found") (HM.lookup note notes)
-                pure $ InL (mapMaybe (\(noteFp, pos@(Position l' _)) -> if l' == l then Nothing else Just (
-                        Location (fromNormalizedUri $ normalizedFilePathToUri noteFp) (Range pos pos))) poss)
+                case HM.lookup note notes of
+                  Nothing -> pure (InL [])
+                  Just poss -> pure $ InL $ mapMaybe (\(noteFp, pos@(Position l' _)) ->
+                      if l' == l
+                        then Nothing
+                        else Just (Location (fromNormalizedUri $ normalizedFilePathToUri noteFp) (Range pos pos))
+                    )
+                    poss
     where
         uriOrig = toNormalizedUri $ param ^. (L.textDocument . L.uri)
 listReferences _ _ _ = throwError $ PluginInternalError "conversion to normalized file path failed"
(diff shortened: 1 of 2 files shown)


plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -227,12 +232,19 @@ extractNoteContent nfp (Position startLine _) = do
...
       afterSeparator =
         case dropWhile (not . isSeparatorLine) afterDecl of
           []     -> []
           (_:xs) -> xs
-
+      isStopLine :: Text -> Bool
+      isStopLine t
+        | isBlockNote = isSeparatorLine t || isNoteHeader t || isEndMarker t
+        | isLineNote  = isSeparatorLine t || isNoteHeader t || isEndMarker t || not ("--" `T.isPrefixOf` T.stripStart t)
+        | otherwise   = isEndMarker t || isNoteHeader t
```
fendor, line 247: What's the alternative to line and block notes?

fendor:
> Leaking Single Comment Notes Declaration

What does this mean?

vidit-od:
My bad, i did not do a good job explaining the issue : 

There are 2 ways to do Note Declaration 
1 ) multi line comments ( or Block Note)
2 ) single line comments ( or Line Note ) Example below : 

singleLine Leak Def

The Note ending condition for 2nd type of declaration was not defined properly. Hence the content outside it was also picked up as content. Below image has `{- Note [Not a valid note]` which should not be there in hover. 

singleLine Leak

To Prevent this we have to Define proper ending conditions, which are different for both the types on Notes. Hence in the above implementation we first check the type of note declaration and extract content accordingly !

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
vidit-od: Ideally none, there are only the above 2 types. There is no alternative.

But i thought having an else condition to gracefully fall off to would be good.

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -132,9 +132,14 @@ listReferences state _ param
...
                 notes <- runActionE "notes.definedNoteReferencess" state $ useE MkGetNoteReferences nfp
-                poss <- err ("Note reference (a comment of the form `{- Note [" <> note <> "] -}`) not found") (HM.lookup note notes)
-                pure $ InL (mapMaybe (\(noteFp, pos@(Position l' _)) -> if l' == l then Nothing else Just (
-                        Location (fromNormalizedUri $ normalizedFilePathToUri noteFp) (Range pos pos))) poss)
+                case HM.lookup note notes of
+                  Nothing -> pure (InL [])
+                  Just poss -> pure $ InL $ mapMaybe (\(noteFp, pos@(Position l' _)) ->
+                      if l' == l
+                        then Nothing
+                        else Just (Location (fromNormalizedUri $ normalizedFilePathToUri noteFp)(Range pos pos))
```
fendor:
```suggestion
                        else Just (Location (fromNormalizedUri $ normalizedFilePathToUri noteFp) (Range pos pos))
```

jvanbruegge (approved):
Aside from @fendor's small nit LGTM

vidit-od:
Done !! 
Thank you for reviewing this PR.

PR #4834, opened 2026-02-06
Use `VirtualFileTree` in `hls-call-hierarchy-plugin-tests`

fendor:
Hopefully makes the test less flaky on windows.


fendor:
cc @vidit-od, perhaps this fixes the flaky call-hierarchy tests

PR #4837, opened 2026-02-07
Ignore records in matching groups that are compiler-generated

### LINKED ISSUE #4656: Avoid aking inlay record field hints for template haskell generated code

In the newest hls version, we get explicit record fields inlay hint.
But somehow this also applies to spliced template haskell code, a very common situation is in something like this:

```haskell
$(generateSomeDecs)

$(logError) $ "error text"

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
....
|]
```

In these senarios, the user cannot see the code or records as they are generated by template haskell splice, but there are a lot of hints produced that occupies considerable amount of screen, blocking the visual and makes editing code tricky.

I would like to know, is there possibility to improve this plugin, so that it *does not generate hints* for *template haskell* generated code (i.e. splices, quasi-quotes).

Alternative would be just turn off this plugin.

## Additional context

Image

Image

Image

Thank you very much for your attention, I deeply appreciate the work involved in this project

ozkutuk:
Fixes #4656.

From what I can tell, there is not a way to tell whether something is compiler-generated or not at the level of expressions or patterns in the typechecked AST. However, I could find such flag to be part of matching groups under `FunBind`s, which is sufficient to come up with an implementation that gets the test I introduced pass. Hopefully this should be enough to cover most scenarios the users might encounter.

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -56,19 +56,22 @@ import           Development.IDE.GHC.Compat           (FieldLabel (flSelector),
                                                        GenLocated (L), GhcPass,
                                                        GhcTc,
                                                        HasSrcSpan (getLoc),
+                                                       HsBindLR (..),
                                                        HsConDetails (RecCon),
                                                        HsExpr (HsApp, HsVar, XExpr),
                                                        HsFieldBind (hfbLHS),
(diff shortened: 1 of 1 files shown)


fendor:
> From what I can tell, there is not a way to tell whether something is compiler-generated or not at the level of expressions or patterns in the typechecked AST

@wz1000 Perhaps you can confirm this?

ozkutuk:
Merging this since it has not received any additional comment in the meanwhile. We can revisit the approach if it turns out that a better way to do this exists.

Issue #4831, opened 2026-02-05
Drop `hls-floskell-plugin`


fendor:
In my understanding, `floskell` is no longer maintained, so we will have to drop the plugin as well

https://github.com/ennocramer/floskell

fendor:
@peterbecich Would you weigh on this? Do you agree?

peterbecich:
> In my understanding, `floskell` is no longer maintained, so we will have to drop the plugin as well
> 
> https://github.com/ennocramer/floskell

Agreed it appears to no longer be maintained, and agreed.

PR #4838, opened 2026-02-07
Remove Floskell support

omarjatoi:
Closes https://github.com/haskell/haskell-language-server/issues/4831

### DIFF
--- a/test/functional/Format.hs
+++ b/test/functional/Format.hs
@@ -46,7 +46,7 @@ providerTests = testGroup "lsp formatting provider"
         documentContents doc >>= liftIO . (@?= formattedOrmolu)
 
         setHlsConfig (formatLspConfig "fourmolu")
-        formatDoc doc (FormattingOptions 2 True Nothing Nothing Nothing)
+        formatDoc doc (FormattingOptions 4 True Nothing Nothing Nothing)
         documentContents doc >>= liftIO . (@?= formattedFourmolu)
 
         setHlsConfig (formatLspConfig "ormolu")


omarjatoi:
Ah I actually just pushed a change up, specifically updating the test in `test/functional/Format.hs` so we _still_ test with the `Ormolu` plugin, I wasn't sure if my change would effectively lower coverage. I pushed the change up right as you hit review, so might be worth another quick look (sorry about that)! I'm happy to also just leave that test case out.

 Also squashed my changed into a single commit.
 
EDIT: though on the other hand there are already tests for `ormolu` in `plugins/hls-ormolu-plugin/test/Main.hs`, so this might be redundant?

test/functional/Format.hs
```diff
@@ -34,25 +32,12 @@ providerTests = testGroup "lsp formatting provider"
...
         formatDoc doc (FormattingOptions 2 True Nothing Nothing Nothing)
         documentContents doc >>= liftIO . (@?= formattedOrmolu)
-
-        setHlsConfig (formatLspConfig "floskell")
-        formatDoc doc (FormattingOptions 2 True Nothing Nothing Nothing)
-        documentContents doc >>= liftIO . (@?= formattedFloskell)
-
-        setHlsConfig (formatLspConfig "ormolu")
-        formatDoc doc (FormattingOptions 2 True Nothing Nothing Nothing)
-        documentContents doc >>= liftIO . (@?= formattedOrmoluPostFloskell)
```
fendor, line 55: Right, I overlooked this in the first review, thank you for pointing this out!

You could change this test to run the formatter with `fourmolu` or `stylish-haskell` instead?

E.g., instead of `requiresOrmoluPlugin . requiresFloskellPlugin`, use `requiresOrmoluPlugin . requiresFourmoluPlugin`.

We shouldn't drop this test as it checks that we can successfully change the formatter on the fly without having to restart. This is not about running the formatter at all.

omarjatoi: Agreed, let me update this to run through with `ormolu` and `fourmolu`. Thanks!

omarjatoi: Alrighty, done. Thank you!

fendor (approved):
LGTM, thank you!

omarjatoi:
I'm working through failing CI, but unsure what the issue with e.g. https://github.com/haskell/haskell-language-server/actions/runs/21804188540/job/62904669400 is. I'll dig into this more and see if it's something related to my change.

fendor:
Please ignore these test failures, we seem to be segfaulting a lot on windows, I restart the jobs :)

PR #4701, opened 2025-08-21
Fix broken pip during test

soulomoon:
gracefully exit the server

PR #4841, opened 2026-02-09
fix spacing issues while refactor-reduce

### LINKED ISSUE #4825: "Remove constraint" refactoring gets spacing wrong
Code action for remove constraint gets sometimes spacing wrong.
```haskell
isInteractionMetaB
  :: forall m. (ReadTCState m, MonadReduce m, MonadPretty m)
  => MetaId
  -> Args
  -> m (Maybe (MetaId, InteractionId, Args))
```
is turned into
```haskell
isInteractionMetaB
  :: forall m. ( MonadPretty m)
  =>MetaId
  -> Args
  -> m (Maybe (MetaId, InteractionId, Args))
```
but correct would be
```haskell
isInteractionMetaB
  :: forall m. (MonadPretty m)
  => MetaId
  -> Args
  -> m (Maybe (MetaId, InteractionId, Args))
```
So the space from behind the `=>` is shifted to behind the first `(`.

### Movie

https://github.com/user-attachments/assets/6c5580ce-bc27-4494-b712-d042d325b03e

### Your environment

- MacOS
- GHC 9.12.2
- HLS 2.12
- VSCode

Related:
- #1578
- #1300

vidit-od:
Fixes #4825 

In the previous implementation `resetEntryDP ` function was triggered when reduction code action removed any leading constraint from the context. This lead to inconsistent spacing for this scenario. 

fixes this by 
1) not using the `resetEndtryDP` function ( fixes removal of space after => )
2) add a normalization function that trims leading spaces

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
@@ -141,6 +141,13 @@ dropHsParTy :: LHsType (GhcPass pass) -> LHsType (GhcPass pass)
 dropHsParTy (L _ (HsParTy _ ty)) = ty
 dropHsParTy other                = other
 
+normalizeConstraintDP :: [LocatedAn AnnListItem (HsType GhcPs)] -> [LocatedAn AnnListItem (HsType GhcPs)]
+normalizeConstraintDP [] = []
+normalizeConstraintDP xs = normalizeFirst xs
+  where
+    normalizeFirst (x:rest) = setEntryDP x (SameLine 0) : rest
+    normalizeFirst [] = []
+
 removeConstraint ::
   -- | Predicate: Which context to drop.
   (LHsType GhcPs -> Bool) ->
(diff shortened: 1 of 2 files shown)


vidit-od:
while working on this issue i also gathered the following observation that 

* `Remove redundant constraint` does not get triggered while dealing with single constrains. i.e  

triggering it on the below example does nothing (But constantly shows warning)

image

while for example 2 (Below) we do observe expected result from 
```haskell
-- Example 2 before remove redundant constraint
j :: (Eq a, Show a) => a -> a
j x = x
```
to 

```haskell
-- Example 2 after remove redundant constraint
j :: a -> a
j x = x
```

i did try and fix this but could not !! 
main possible suspect of this behaviour could be lying at `\hls-refactor-plugin\src\Development\IDE\Plugin\CodeAction.hs`

to be specific `findSigOfDeclRanged` function possibly returning `Nothing` on single constrains. 

Requesting a look at it if worthwhile @fendor

fendor:
~~Looks like the ubuntu 9.12 errors are genuine~~ EDIT: Nope, flaky

vidit-od:
Error logs stated that it failed due to resource vanish ( broken pipe ) 

Should i look into it ? ( any possibility of optimisation like call-hierarchy ? )

fendor:
> Error logs stated that it failed due to resource vanish ( broken pipe )

We try to address this kind of error in #4701

Dont investigate, I don't think it is worth it

fendor (approved):
LGTM, tbh, I am relying on the test suite for correctness, because I can't tell without serious tinkering whether this is correct.

Thank you for fixing this!

PR #4845, opened 2026-02-14
Cross Module renaming

### LINKED ISSUE #4816: `hls-rename-plugin`: Rename across modules fails
### Your environment

HLS 2.13.0.0, GHC 9.10.3

### Steps to reproduce

```
git clone github.com:fendor/rename-hls-bug-test.git
code .
```

Rename `someFunc` identifier.

### Expected behaviour

rename `someFunc` in `MyLib1` and `MyLib2`

### Actual behaviour

Renames only in the current module

vidit-od:
Fixes #4816 

This PR aims at implementing Cross Module Renaming. Recently Cross Module Renaming was made True by default in HLS, but current implementation shows `Renaming of an exported name is unsupported` and `Renaming of an imported name is unsupported` when attempting to do so. 

The main reason for this was the `failWhenImportOrExport` function blocking rename whenever any imported or exported variables are renamed.

The current implementation carefully checks and distinguish bw local , exported, imported variables. If cross Module is disabled. Blocks renaming imported and exported variables renaming. The older constrain of having an Explicit export list is still maintained. 

The below comment ([ link ](https://github.com/haskell/haskell-language-server/pull/4845#issuecomment-3906573813)) also talks about a scenario where relying completely on `HieDb` can lead to incorrect results when local references and imported/ exported references have same name. Hence Ast is also referred for renaming of local variables.

### DIFF
--- a/plugins/hls-rename-plugin/test/Main.hs
+++ b/plugins/hls-rename-plugin/test/Main.hs
@@ -49,7 +49,7 @@ prepareRenameTests = testGroup "PrepareRename"
         void waitForBuildQueue
         result <- prepareRename doc (Position 10 16)
         liftIO $ result @?=
-            InL (PrepareRenameResult (InL (Range (Position 10 14) (Position 10 19))))
+            InL (PrepareRenameResult (InL (Range (Position 10 16) (Position 10 19))))
 
     , testCase "Non-renameable position" $ runRenameSession "" $ do
         doc <- openDoc "PrepareRename.hs" "haskell"


vidit-od:
This PR aims at implementing cross module renaming. 

It is ready for review and most of the functionality has been check by me manually. Raising this PR to see how the tests perform on various environments. 

Will add testcases for this Feature once we can verify old ones work properly.

fendor:
Could you briefly explain the strategy of the solution? I was under the impression, that the rename feature could entirely rely on `hiedb` to find references to a Name and replacing them, but it seems like you are doing quite a lot more with the GHC AST directly?

vidit-od:
## Reason for using AST 

_Implementation with HieDb :_ 
OldCrossNaming-ezgif com-video-to-gif-converter
Here changing the local variable leads to changing of all other variables of same names (Local and Global). The reason for this is HieDB uses OccName for finding references. AST presents absolute truth and should be used to rely on for preventing inconsistencies. 

This similar break can be observed in any and all examples where a local variable has same name as imported variable. Although this is a very specific example of : Importing a variable -> having a same name variable as a local variable -> user triggers rename on that variable. 

_current implementation enables :_ 
NewCrossNaming

**IMP :** HieDB does cover all of the simpler cases. considering that AST traversal takes comparatively longer time we have to choose b/w faster but little inconsistent implementation vs slower but always correct one ; both version of codes are available with me :) 

Hopefully this makes everything clear ! @fendor

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -99,16 +100,43 @@ renameProvider state pluginId (RenameParams _prog (TextDocumentIdentifier uri) p
...
     case oldNames of
         -- There were no Names at given position (e.g. rename triggered within a comment or on a keyword)
         [] -> throwError $ PluginInvalidParams "No symbol to rename at given position"
         _  -> do
-            refs <- HS.fromList . concat <$> mapM (refsAtName state nfp) oldNames
+            refs' <- HS.fromList . concat <$> mapM (refsAtName state nfp) oldNames
+            exportRefs <- exportNameLocs pm oldNames
+            isExported <- or <$> mapM (isNameExplicitExported pm) oldNames
+            let refs = HS.union refs' (HS.fromList exportRefs)
+                currentModule = fmap unLoc $ hsmodName $ unLoc $ pm_parsed_source pm
```
wz1000, line 150: get this from the `ModSummary`, will allow you to get rid of the `Nothing` case.

wz1000:
Can you offer a overview of what the bug was, and your approach towards fixing it please? It would be good to add this to the commit message as well.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -195,23 +249,33 @@ refsAtName ::
...
-    pure $ nameLocs name ast ++ dbRefs
+    let localRefs = nameLocs targetName ast
+
+    -- Query HieDb for global matches (by OccName)
+    ShakeExtras{withHieDb} <- liftIO $ runAction "Rename.HieDb" state getShakeExtras
+    dbCandidates <- liftIO $ withHieDb $ \hieDb ->
+        fmap (mapMaybe rowToLoc) $
+          case nameModule_maybe targetName of
+            Just mod -> findReferences hieDb True (nameOccName targetName) (Just $ moduleName mod) (Just $ moduleUnit mod) []
+            Nothing  -> findReferences hieDb True (nameOccName targetName) Nothing Nothing []
```
wz1000: I'm not sure this does anything, `hiedb` doesn't index local names.

wz1000: You can look for references in the HieAst itself using the cached `RefMap`.

wz1000:
It would also be good to add the example from https://github.com/haskell/haskell-language-server/issues/4816 as a regression test.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
vidit-od: i am not sure if this is logical / why this is happening but i tried doing what the review recommends by replacing the highlighted code with this code 

```
currentModule = ms_mod $ pm_mod_summary pm
                isLocallyDefined name =
                    case nameModule_maybe name of
                        Nothing -> True
                        Just nameModule -> nameModule == currentModule             
```

doing this made a lot of testcases flaky. i have tried running almost all of the testcases numerous times and find no logic to why they fail. below is the list of testcases that showed flaky nature; 

* `"Data constructor with fields"`
* `"Function name"`
* `"Realigns do block indentation"`

i dont know why this would happen but for all of the above cases we randomly get the below error 
`"Cannot rename symbol: module has no explicit export list and the symbol is referenced from other modules."`

reverting this block to the older version made all of the cases work fine ( Confirmed this multiple times as well )
any guesses why this would happen ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -99,46 +100,98 @@ renameProvider state pluginId (RenameParams _prog (TextDocumentIdentifier uri) p
...
+            -- if CrossModule renaming requires Explicit Export list
+            -- if variable is imported somewhere else && No explicit export => ERROR
+            -- if variable is locally used => No ERROR
+            let hasExplicitExportList = isJust (hsmodExports (unLoc $ pm_parsed_source pm))
+            refFiles <- forM (HS.toList refs) $ \loc -> do
+                (file, _) <- locToFilePos loc
+                pure file
+            let hasExternalRefs = any (/= nfp) refFiles
+            when ( crossModuleEnabled && not hasExplicitExportList && hasExternalRefs && renamingLocalDeclaration ) $ throwError $ PluginInvalidParams
+                "Cannot rename symbol: module has no explicit export list and the symbol is referenced from other modules."
```
vidit-od, line 175: These are the conditions that define when to show the mentioned error. 

CurrentModule -> isLocallyDefined -> renamingLocalDeclaration -> causing the error ? 

could this be a possible reason ?

vidit-od:
* i have updated the PR description with an overview explanation, 
* also new regression tests including the ones listed in original issue as well as a few more cases have been added. 
* The one review regarding usage of `Modsummary` has been reverted because of the reasons listed in above conversation. 

On a side note: the tests for rename plugin gave me a very hard time. Tho ,  I dont know if we can do anything about it or not :) 

requesting a re-review @wz1000 @fendor

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -195,29 +271,29 @@ refsAtName ::
...
-                [fromNormalizedFilePath nfp]
-            )
-    pure $ nameLocs name ast ++ dbRefs
-
-nameLocs :: Name -> HieAstResult -> [Location]
-nameLocs name (HAR _ _ rm _ _) =
-    concatMap (map (realSrcSpanToLocation . fst))
-              (M.lookup (Right name) rm)
-
+refsAtName state nfp targetName = do
```
wz1000: I don't understand why this function needs to change, or what the changes do.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -272,19 +348,62 @@ getNamesAtPoint' hf pos =
...
+  where
+    unsupported = throwError $ PluginInternalError "Renaming is unsupported for complex export forms"
+
+    matchWrapped :: SrcSpan -> LIEWrappedName GhcPs -> ExceptT PluginError (HandlerM config) [Location]
+    matchWrapped l ieWrapped =
+        case unwrapIEWrappedName (unLoc ieWrapped) of
+            Just rdr
+              | any (matchesRdr rdr) names
+              -> do
+                loc <- srcSpanToLocE l
```
wz1000, line 458: I don't think this should be throwing

plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -49,7 +49,7 @@ prepareRenameTests = testGroup "PrepareRename"
         void waitForBuildQueue
         result <- prepareRename doc (Position 10 16)
         liftIO $ result @?=
-            InL (PrepareRenameResult (InL (Range (Position 10 14) (Position 10 19))))
+            InL (PrepareRenameResult (InL (Range (Position 10 16) (Position 10 19))))
```
vidit-od, line 52: Image

changed this Regression test. 
we now trigger rename only when cursor name is properly on identifier.

PR #4849, opened 2026-02-23
Migrate HLS-refactor-plugin to use Structured Diagnostic

vidit-od:
A contributing to #4605 

This PR aims at **Partially** Migrating `Hls-refactor-plugin` to use Structured Diagnostics.  

Below is a file wise list of functions that have been successfully migrated by this PR. A complete migration was attempted but could not be achieved due to lack of some information from GHC Api .

FillHole.hs
- `suggestFillHole`

Diagnostic.hs
- `matchFoundHole`
- `matchFoundHoleIncludeUnderscore`
- `matchVariableNotInScope`

CodeAction.hs
- `suggestHideShadow`
- `suggestRemoveRedundantImport`
- `suggestDeleteUnusedBinding`
- `suggestExportUnusedTopBinding`
- `suggestNewDefinition`
- `removeRedundantConstraints`

Inspiration tax : 
Some of the content in this PR was inspired / stolen (👀) from PR #4695 :)

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
@@ -69,42 +69,39 @@ type HsArrow pass = HsMultAnn pass
 --      In this case a new argument would have to add its type between b and c in the signature.
 plugin :: ParsedModule -> FileDiagnostic -> Either PluginError [(T.Text, [TextEdit])]
 plugin parsedModule fd
-  | Just (name, typ) <- matchVariableNotInScope fd = addArgumentAction parsedModule _range name typ
-  | Just (name, typ) <- matchFoundHoleIncludeUnderscore fd = addArgumentAction parsedModule _range name (Just typ)
+  | Just (rdrName, typ) <- matchVariableNotInScope fd = addArgumentAction parsedModule _range rdrName typ
+  | Just (rdrName, typ) <- matchFoundHole fd = addArgumentAction parsedModule _range rdrName (Just typ)
   | otherwise = pure []
   where
-    Diagnostic{_message, _range} = fdLspDiagnostic fd :: Diagnostic
+    Diagnostic{_message, _range} = fdLspDiagnostic fd
 
 -- Given a name for the new binding, add a new pattern to the match in the last position,
 -- returning how many patterns there were in this match prior to the transformation:
 --      addArgToMatch "foo" `bar arg1 arg2 = ...`
 --   => (`bar arg1 arg2 foo = ...`, 2)
-addArgToMatch :: T.Text -> GenLocated l (Match GhcPs (LocatedA (HsExpr GhcPs))) -> (GenLocated l (Match GhcPs (LocatedA (HsExpr GhcPs))), Int)
+addArgToMatch :: RdrName -> GenLocated l (Match GhcPs (LocatedA (HsExpr GhcPs))) -> (GenLocated l (Match GhcPs (LocatedA (HsExpr GhcPs))), Int)
 
 -- NOTE: The code duplication within CPP clauses avoids a parse error with
 -- `stylish-haskell`.
 #if MIN_VERSION_ghc(9,11,0)
-addArgToMatch name (L locMatch (Match xMatch ctxMatch (L l pats) rhs)) =
-  let unqualName = mkRdrUnqual $ mkVarOcc $ T.unpack name
-      newPat = L noAnnSrcSpanDP1 $ VarPat NoExtField $ L noAnn unqualName
+addArgToMatch unqualName (L locMatch (Match xMatch ctxMatch (L l pats) rhs)) =
+  let newPat = L noAnnSrcSpanDP1 $ VarPat NoExtField $ L noAnn unqualName
       -- The intention is to move `= ...` (right-hand side with equals) to the right so there's 1 space between
       -- the newly added pattern and the rest
       indentRhs :: GRHSs GhcPs (LocatedA (HsExpr GhcPs)) -> GRHSs GhcPs (LocatedA (HsExpr GhcPs))
       indentRhs rhs@GRHSs{grhssGRHSs} = rhs {grhssGRHSs = fmap (`setEntryDP` (SameLine 1)) grhssGRHSs }
    in (L locMatch (Match xMatch ctxMatch (L l (pats <> [newPat])) (indentRhs rhs)), Prelude.length pats)
 #elif MIN_VERSION_ghc(9,9,0)
-addArgToMatch name (L locMatch (Match xMatch ctxMatch pats rhs)) =
-  let unqualName = mkRdrUnqual $ mkVarOcc $ T.unpack name
-      newPat = L noAnnSrcSpanDP1 $ VarPat NoExtField $ L noAnn unqualName
+addArgToMatch unqualName (L locMatch (Match xMatch ctxMatch pats rhs)) =
+  let newPat = L noAnnSrcSpanDP1 $ VarPat NoExtField $ L noAnn unqualName
       -- The intention is to move `= ...` (right-hand side with equals) to the right so there's 1 space between
       -- the newly added pattern and the rest
       indentRhs :: GRHSs GhcPs (LocatedA (HsExpr GhcPs)) -> GRHSs GhcPs (LocatedA (HsExpr GhcPs))
       indentRhs rhs@GRHSs{grhssGRHSs} = rhs {grhssGRHSs = fmap (`setEntryDP` (SameLine 1)) grhssGRHSs }
    in (L locMatch (Match xMatch ctxMatch (pats <> [newPat]) (indentRhs rhs)), Prelude.length pats)
 #else
-addArgToMatch name (L locMatch (Match xMatch ctxMatch pats rhs)) =
-  let unqualName = mkRdrUnqual $ mkVarOcc $ T.unpack name
-      newPat = L (noAnnSrcSpanDP1 generatedSrcSpan) $ VarPat NoExtField (noLocA unqualName)
+addArgToMatch unqualName (L locMatch (Match xMatch ctxMatch pats rhs)) =
+  let newPat = L (noAnnSrcSpanDP1 generatedSrcSpan) $ VarPat NoExtField (noLocA unqualName)
       indentRhs = id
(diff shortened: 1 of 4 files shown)


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
```diff
@@ -66,13 +67,13 @@ type HsArrow pass = HsMultAnn pass
...
-  | Just (name, typ) <- matchVariableNotInScope message = addArgumentAction parsedModule _range name typ
-  | Just (name, typ) <- matchFoundHoleIncludeUnderscore message = addArgumentAction parsedModule _range name (Just typ)
+plugin :: ParsedModule -> FileDiagnostic -> Either PluginError [(T.Text, [TextEdit])]
+plugin parsedModule fd
+  | Just (name, typ) <- matchVariableNotInScope fd = addArgumentAction parsedModule _range name typ
+  | Just (name, typ) <- matchFoundHoleIncludeUnderscore fd = addArgumentAction parsedModule _range name (Just typ)
   | otherwise = pure []
   where
-    message = unifySpaces _message
+    Diagnostic{_message, _range} = fdLspDiagnostic fd :: Diagnostic
```
VeryMilkyJoe: Is there a reason the type hint is needed here?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
```diff
@@ -27,33 +43,53 @@ matchRegex message regex = case message =~~ regex of
...
+    solverReport <-
+        diag
+            ^? fdStructuredMessageL
+                . _SomeStructuredMessage
+                . msgEnvelopeErrorL
+                . _TcRnMessage
+                . _TcRnSolverReport
+                . _1
+    (hole, _) <- solverReport ^? reportContentL . _ReportHoleError
+
```
VeryMilkyJoe:
```suggestion
```

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
```diff
@@ -27,33 +43,53 @@ matchRegex message regex = case message =~~ regex of
...
+matchFoundHoleIncludeUnderscore :: FileDiagnostic -> Maybe (T.Text, T.Text)
+matchFoundHoleIncludeUnderscore fd = first ("_" <>) <$> matchFoundHole fd
+
+matchVariableNotInScope :: FileDiagnostic -> Maybe (T.Text, Maybe T.Text)
+matchVariableNotInScope fd = do
+    (rdrName, _) <- diagReportNotInScope fd
+    Just (printOutputable rdrName, Nothing)
+
+-- | Extract the 'Hole' out of a 'FileDiagnostic'
```
VeryMilkyJoe: I think this could use a bit more explanation in the documentation

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -1,43 +1,162 @@
...
+suggestFillHole exportsMap pm diag
+    | Just holeName <- extractHoleName diag
+#if MIN_VERSION_ghc(9,13,0)
+    , Just _errInfo <- extractErrInfo diag
+    , let supplText = _message (diag ^. fdLspDiagnosticL)
+    , let ctxText = supplText
+#else
+    , Just (ErrInfo ctx suppl) <- extractErrInfo diag
+    , let ctxText = printOutputable ctx
+    , let supplText = printOutputable suppl
```
VeryMilkyJoe: Just a small nitpick, but for better readabilty, I would prefer the variables to have the same order as above.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -1,43 +1,162 @@
...
+    , let (holeFits, refFits) = processHoleSuggestions (T.lines supplText)
+    , let isInfixHole = ctxText =~ addBackticks holeName :: Bool =
         map (proposeHoleFit holeName False isInfixHole) holeFits
-        ++ map (proposeHoleFit holeName True isInfixHole) refFits
+        ++
+        map (proposeHoleFit holeName True isInfixHole) refFits
     | otherwise = []
     where
-      extractHoleName = fmap (headOrThrow "impossible") . flip matchRegexUnifySpaces "Found hole: ([^ ]*)"
+      qualify = qualifyFit exportsMap pm
```
VeryMilkyJoe: This is the only function in the where block without a signature, also I feel like extracting it does not really add anything, as it is only used once and the name is not more than seeing it applied directly imo.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -1,43 +1,162 @@
...
+    findQualifier :: Maybe T.Text
+    findQualifier =
+        let exportingSet = exportingModules
+        in fmap snd
+         . safeHead
+         . filter (\(modN, _) -> modN `elem` exportingSet)
+         $ qualifiedImports
+
+    safeHead []    = Nothing
+    safeHead (x:_) = Just x
```
VeryMilkyJoe: indentation here is inconsistent.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -1,43 +1,162 @@
...
+
```
VeryMilkyJoe, line 153:
```suggestion
```

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -1,43 +1,162 @@
...
+    isQualified decl = ideclQualified decl `elem` [QualifiedPre, QualifiedPost]
+
+    modName decl =
+        T.pack . moduleNameString . unLoc . ideclName $ decl
+
+    qualifier decl =
+        case ideclAs decl of
+            Just alias -> T.pack . moduleNameString . unLoc $ alias
+            Nothing    -> modName decl
```
VeryMilkyJoe: There is some name shadowing regarding `qualifier` here. 
Maybe the function could have a name, that more closely indicates what it does instead?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
```diff
@@ -27,33 +43,53 @@ matchRegex message regex = case message =~~ regex of
...
+    hole <- diagReportHoleError fd
+    Just (printOutputable (hole_occ hole), printOutputable (hole_ty hole))
+
+matchFoundHoleIncludeUnderscore :: FileDiagnostic -> Maybe (T.Text, T.Text)
+matchFoundHoleIncludeUnderscore fd = first ("_" <>) <$> matchFoundHole fd
+
+matchVariableNotInScope :: FileDiagnostic -> Maybe (T.Text, Maybe T.Text)
+matchVariableNotInScope fd = do
+    (rdrName, _) <- diagReportNotInScope fd
+    Just (printOutputable rdrName, Nothing)
```
VeryMilkyJoe: It seems strange that you use the new structured data to produce text immediately, instead of working with it directly.
Is there a reason why you transform it to text and do all this string matching and manipulation?

vidit-od: iirc, i did that to temporarily not change the behavior on caller site. i planned to change it later and lost track of it. 

but yes, it makes sense to not produce text immediately. will change it and update PR soon :)

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
vidit-od, line 186: since we get the Types from GHC now, 
we can try and improve this from `add type hole to signature`
to  `add type to signature` maybe ? 

i am not sure tho, i cannot replicate the current expected behavior locally ( examples in comment ).

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/Diagnostic.hs
vidit-od: i am not 100% sure if i fixed this, 

by adding documentation did you mean : 
a ) for why we are directly converting to string ? if so we have removed this behavior
b) for the hole extract function ? if so this was before this PR, this function already existed ( in fillwildcard.hs )
This PR just moved the exact function from fillwildcard to dignostics.hs ( it better suits in this file )

i did expand the comment a little to address this review, but just wanted to clear if i understood the assignment or not :)

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/FillHole.hs
```diff
@@ -1,43 +1,158 @@
...
+import           Development.IDE.Plugin.Plugins.Diagnostic (diagReportHoleError)
+import           Development.IDE.Types.Diagnostics         (_SomeStructuredMessage,
+                                                            fdStructuredMessageL)
+import           Development.IDE.Types.Exports             (ExportsMap (..),
+                                                            mkVarOrDataOcc,
+                                                            moduleNameText)
+import           GHC.Tc.Errors.Types                       (ErrInfo (ErrInfo))
+import           Language.Haskell.Syntax.ImpExp            (ImportDeclQualifiedStyle (..))
+import           Language.LSP.Protocol.Lens                (HasRange (..),
+                                                            message)
```
vidit-od, line 36: These imports are breaking circleci build tests. 
Need some help/direction on how to tackle these failures ?

fendor: Looks like we can't CPP in this situation.

vidit-od: yea, could not find any way to remove this redundant import error :(

PR #4851, opened 2026-02-23
Prefer extending import code actions

### LINKED ISSUE #4750: Put QuickFix "Add ... to import list of ..." first
When I have a not-in-scope identifier, I often get a couple of suggestions in the "Quick Fix" list, but in most cases of the cases I want add this identifier to an existing import list.

My suggestion is to put this fix first in the list.

Maybe even make this fix available by a special short cut (like double-click).

Image

"Define ..." has definitely the lowest prio.

vherrmann:
Fixes #4750.

Before this PR:

screen_2026-02-23_22-32-49

After this PR:

screen_2026-02-23_22-30-57

The PR achieves this by setting isPreferred for the corresponding code actions. I do not find this a completely satisfactory solutions, since ideally we should order the code actions according to a prioritization ("Add … to the import list …" should imo appear before more generic import list changes which should appear before "Replace with …" which should appear before "Define …"). 
On the other hand the LSP Spec afaik does not specify that clients should show the code actions according to the order given by the lsp server (And at least my emacs config does not seem to care about the servers order, in fact it does not even care about the "isPreferred" setting).

### DIFF
--- a/ghcide-test/exe/THTests.hs
+++ b/ghcide-test/exe/THTests.hs
@@ -4,6 +4,7 @@ module THTests (tests) where
 import           Config
 import           Control.Monad.IO.Class      (liftIO)
 import qualified Data.Text                   as T
+import           Development.IDE.GHC.Compat  (GhcVersion (..), ghcVersion)
 import           Development.IDE.GHC.Util
 import           Development.IDE.Test        (expectCurrentDiagnostics,
                                               expectDiagnostics,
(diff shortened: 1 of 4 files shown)


vherrmann:
Btw: the definition `toCodeAction` in ./plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/Args.hs is some very nice code. I guess one could achieve something similar with a reader monad (stack), but I still find the code kinda mind-blowing.

PR #4852, opened 2026-02-24
fix(refactor): do not offer 'Add Argument' for qualified names

### LINKED ISSUE #4842: Incorrect CodeAction to add qualified name as function argument
Image

Clearly, in no sane world can `NE.toList` be a function argument, so we shouldn't offer this code action, imo!

HLS 2.13.0.0
GHC 9.10.3

ijatinydv:
Fixes #4842

**The Problem**
Currently, if a user has an out-of-scope qualified name (like `NE.toList`), the `AddArgument `plugin incorrectly offers a code action to add it as a function parameter (e.g., "Add argument 'NE.toList' to function").

Accepting this action generates invalid Haskell syntax, as function parameters must be unqualified variable names.

**The Solution**

- **Added a guard**: Introduced a `not (T.any (== '.') name)` check in `AddArgument.hs` to explicitly skip generating this code action if the missing identifier contains a module qualifier.
- **Added regression tests**: Created a new test fixture (`QualifiedName.hs`) and a test (`mkNoAddArgForQualifiedNameTest`) to guarantee the action is strictly suppressed for qualified names.

**Verification**
All tests in `hls-refactor-plugin` pass successfully:
```
refactor
  code actions
    add argument
      No add argument for qualified names: OK (0.63s)
```

### DIFF
--- a/.github/workflows/clear-cache.yaml
+++ b/.github/workflows/clear-cache.yaml
@@ -9,6 +9,8 @@ on:
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    permissions:
+      actions: write
     steps:
       - name: Cleanup caches for closed PR
         # remove all cache entries for this PR
@@ -16,7 +18,7 @@ jobs:
         # other PRs / master cannot reuse them
         # so clear them when the PR is closed or merged
         run: |
-          gh cache list --ref ${{ github.ref }} --json id \
+          gh cache list --ref refs/pull/${{ github.event.pull_request.number }}/merge --json id \
             | jq -r '.[].id' \
             | xargs -I{} gh cache delete {}
         env:


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
```diff
@@ -68,7 +68,10 @@ type HsArrow pass = HsMultAnn pass
 --      In this case a new argument would have to add its type between b and c in the signature.
 plugin :: ParsedModule -> Diagnostic -> Either PluginError [(T.Text, [TextEdit])]
 plugin parsedModule Diagnostic {_message, _range}
-  | Just (name, typ) <- matchVariableNotInScope message = addArgumentAction parsedModule _range name typ
+  -- Qualified names (e.g. NE.toList) can never be valid function argument patterns,
+  -- so we must not offer the "Add argument" code action for them.
+  | Just (name, typ) <- matchVariableNotInScope message
+  , not (T.any (== '.') name) = addArgumentAction parsedModule _range name typ
```
VeryMilkyJoe: It would be more readable if you used a actual predicate here, that has a name, representative of what it checks.

Also, could you explain why you chose to check for dots in the name instead of capitalisation?

plugins/hls-refactor-plugin/test/data/add-arg/QualifiedName.hs
```diff
@@ -0,0 +1,6 @@
+module QualifiedName where
+
+import Data.List.NonEmpty (NonEmpty(..))
+
+foo :: NonEmpty a -> [a]
+foo xs = NE.toList xs
```
VeryMilkyJoe, line 6: This is not a golden test file, so please move it to a directory in data.

test_output.log
```diff
@@ -0,0 +1,28 @@
...
+  code actions
+    add argument
+      No add argument for qualified names: OK (0.63s)
+
+All 1 tests passed (0.63s)
+
+Test suite hls-refactor-plugin-tests: PASS
+Test suite logged to:
+C:\Users\HP\Desktop\fortan\haskell-language-server\.\dist-newstyle\build\x86_64-windows\ghc-9.14.1\haskell-language-server-2.13.0.0\t\hls-refactor-plugin-tests\test\haskell-language-server-2.13.0.0-hls-refactor-plugin-tests.log
+1 of 1 test suites (1 of 1 test cases) passed.
```
VeryMilkyJoe: Please remove this file.

plugins/hls-refactor-plugin/test/Test/AddArgument.hs
```diff
@@ -71,3 +74,18 @@ mkGoldenAddArgTest' testFileName range varName = do
...
+mkNoAddArgForQualifiedNameTest =
+    testCase "No add argument for qualified names" $ runSessionWithServerInTmpDir def
+      (mkPluginTestDescriptor Refactor.bindingsPluginDescriptor "ghcide-code-actions-bindings")
+      (FS.mkVirtualFileTree "plugins/hls-refactor-plugin/test/data/golden/add-arg" (FS.directProject "QualifiedName.hs"))
+      $ do
+        doc <- openDoc "QualifiedName.hs" "haskell"
+        _ <- waitForDiagnostics
+        actions <- getCodeActions doc (Range (Position 5 0) (Position 5 50))
+        let addArgAction = find (\case
+              InR CodeAction {_title = t} -> "Add argument" `T.isPrefixOf` t
```
VeryMilkyJoe, line 101: It would be helpful to also verify that we get a suggestion here, namely adding `Data.List.NonEmpty as NE` so we know we're in the right spot.

VeryMilkyJoe (changes requested):
Thank you for the PR! 
After some small changes, I think this will be a nice addition :)

ijatinydv:
> Thank you for the PR! After some small changes, I think this will be a nice addition :)

Thanks for the great feedback! @VeryMilkyJoe  I've just pushed an update addressing all your points:

1. **Predicate Extraction**: Extracted the check into an `isQualifiedName` helper for better readability. I also added a comment explaining the choice of checking for dots instead of capitalization (since qualified operators like `NE.:|` aren't capitalized).
2. **File Location**: Moved `QualifiedName.hs` out of the `golden `directory into `test/data/add-arg`.
3. **Cleanup**: Removed the accidentally committed `test_output.log` file.
4. **Test Robustness**: Updated the test to explicitly verify that the "not in scope" diagnostic actually fires at that location before we check that the code action is suppressed.

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
ijatinydv: > It would be more readable if you used a actual predicate here, that has a name, representative of what it checks.
> 
> Also, could you explain why you chose to check for dots in the name instead of capitalisation?

yes of course, because the dot is the exact syntactic marker for a qualified name in Haskell, and an unqualified local variable can never contain one.

If we relied on capitalization, we would have to parse the string to isolate the module prefix (since the actual function name, like `toList`, is lowercase). Furthermore, capitalization checks fall apart for qualified operators (like `NE.:| `or `Map.!?`).

Using `T.any (== '.')` is not only simpler and more efficient, but it safely covers all of those edge cases! I've added a brief comment in the code to clarify this as well.

plugins/hls-refactor-plugin/test/data/add-arg/QualifiedName.hs
ijatinydv: Done! Moved this out of the `golden` directory to `test/data/add-arg/QualifiedName.hs`

plugins/hls-refactor-plugin/test/Test/AddArgument.hs
ijatinydv: Great point. I've added an assertion to verify that an `Import` action is indeed offered, ensuring the diagnostic is firing correctly at this location.

fendor: In the assertion, you are not mentioning `Import` at all, so I fail to see how this assertion checks a `not in scope` code action exists?

ijatinydv: > In the assertion, you are not mentioning `Import` at all, so I fail to see how this assertion checks a `not in scope` code action exists?

Ah my bad, I completely misread the original request. I was checking the diags list for the raw compiler error instead of actually checking the actions list for the import suggestion.

Just pushed a fix to properly assert against the code actions. Thanks for catching that!

plugins/hls-refactor-plugin/test/Test/AddArgument.hs
```diff
@@ -71,3 +75,31 @@ mkGoldenAddArgTest' testFileName range varName = do
...
       action
+
+-- | Verify that the "Add argument" code action is NOT offered for qualified names (e.g. NE.toList).
+-- We also verify that an import suggestion IS offered, to confirm we're testing the right spot.
+mkNoAddArgForQualifiedNameTest :: TestTree
+mkNoAddArgForQualifiedNameTest =
+    testCase "No add argument for qualified names" $ runSessionWithServerInTmpDir def
+      (  mkPluginTestDescriptor Refactor.iePluginDescriptor "ghcide-code-actions-imports-exports"
+      <> mkPluginTestDescriptor Refactor.bindingsPluginDescriptor "ghcide-code-actions-bindings"
+      )
```
fendor, line 86:
```suggestion
      ( mkPluginTestDescriptor Refactor.bindingsPluginDescriptor "ghcide-code-actions-bindings"
      )
```

Other `add-args` tests don't seem to need `iePluginDescriptor`, is it needed here?

plugins/hls-refactor-plugin/src/Development/IDE/Plugin/Plugins/AddArgument.hs
```diff
@@ -68,11 +68,20 @@ type HsArrow pass = HsMultAnn pass
...
+  , not (isQualifiedName name) = addArgumentAction parsedModule _range name typ
   | Just (name, typ) <- matchFoundHoleIncludeUnderscore message = addArgumentAction parsedModule _range name (Just typ)
   | otherwise = pure []
   where
     message = unifySpaces _message
+    -- Qualified names (e.g. NE.toList) contain a module qualifier separated by
+    -- a dot. They can never be valid function argument patterns, so we skip them.
+    -- We check for dots rather than capitalisation because:
+    --   1. Operators can also be qualified (e.g. NE.:|), where the part after the
+    --      dot is not capitalised.
```
fendor: When talking about capitalisation, the `NE` *is* capitalised. We wouldn't check whether `toList` or `:|` is capitalised, but rather whether the first component is capitalised, since module prefixes are always capitalised.

I think this point should be removed, as it is misleading.

The only reason I can think of to prefer `T.any (== '.')` over `isUpper . head` is to avoid adding `foo.bar`  as an argument, as this is using `RecordDotSyntax` and can also never be added as an argument.
However, this is not the case, as GHC generates a different error message.

Given that, I honestly fail to understand why we would prefer `T.any (== '.')`.

ijatinydv: You're totally right. I was so focused on the operator itself (like `:|`) not being capitalized that I forgot the module prefix obviously is. I've dropped that misleading comment and switched the check to look for an uppercase first letter.

plugins/hls-refactor-plugin/test/Test/AddArgument.hs
ijatinydv: I originally added that to explicitly verify the import suggestion based on the earlier feedback. But I completely agree it just bloats the test and makes it inconsistent with the rest of the suite. I've ripped the descriptor and the import check back out.

fendor: Wait, I am sorry, I didn't see this connection, didn't we add this assertion due to an earlier comment were asked that we are getting some meaningful code action?

ijatinydv: That's completely okay! Yes, an earlier review suggested adding an assertion for the `import qualified ...` code action just to guarantee we were testing the exact right spot. That's why I brought in the `iePluginDescriptor`.

I'm completely fine either way! I can easily revert that last commit and put the import check back in, or if you feel the test is cleaner as a pure `bindings` test without it, we can just leave it as is.

Let me know which version you prefer?

fendor: I prefer having an meaninful code action.

ijatinydv:
> Thank you for taking this on, let's fix the test assertion and then merge this!

I have fixed the test assertion. Please have a look!

PR #4854, opened 2026-02-27
Fixes redundant hash in ghcide cache path (getCacheDirsDefault)

### LINKED ISSUE #4843: Redundant hash in ghcide cache path causes long path
## Description
while working with tests for hls-rename-plugin i observed inability to access Hie files causing failure of almost all tests. The main reason for this was disabled Long paths by default on windows. 

To be specific the path for my cache was 
`C:\\Users\\Asus\\AppData\\Local\\ghcide\\main-f16cf8d0c66360e1b6e279533ab63762122d89cc-f16cf8d0c66360e1b6e279533ab63762122d89cc\\Let.hie`

## Possible solution 
From the path we can observe this pattern `main-<hash>-<hash>`, here both hash are same. thus the Full path can be significantly shortened. 

achieving this can improve HLS window's UX.
## Alternatives 
1. create a temp path which is shorter using commands
```Haskell
$env:TEMP = "C:\temp"
$env:TMP  = "C:\temp"
```
this is temporary solution, have to do everytime
or
2. enable longpaths using `regedit`

## Additional context
OS : windows 11 
GHC : 9.6.7

adpad-13:
Fixes #4843

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -88,8 +88,8 @@ data HomeUnitConfig = HomeUnitConfig
   -- | All the targets for this unit.
   , homeUnitTargets  :: [GHC.Target]
   -- | Optional hash seed to differentiate home units
-  -- with same `-this-unit-id`. Used when `-this-unit-id` is "main"
-  -- , common when loading a single target.
+  -- with same `-this-unit-id`. Used when `-this-unit-id` is "main",
+  -- which is common when loading a single target.
   , homeUnitHash     :: Maybe B.ByteString
   }
 -- This is pristine information about a component


fendor:
Hi, thank you for your PR.

You give no reasoning or justification for your fix. I encourage you to think about it in more detail or to reach out with questions if you do not understand why the code is how it is right now.
Also, commit messages should include reasoning for the change so that we understand better in the future why a change was made.

adpad-13:
Hi,thanks for the feedback. I'm pretty new to open source and this codebase, so I really appreciate the guidance.
You were completely right to ask for the reasoning. From what I found, sometimes the upstream code passes down an ID that already has a hash in it, but getCacheDirsDefault function doesn't check for that and blindly adds the hash again, causing the main-hash-hash duplication.
I added the isSuffixOf check just to stop it from adding a second hash if one is already there. I've also updated the commit to explain this.
I'm not familiar with the upstream cradle code yet, but if you think it's better to fix the root cause up there instead of doing this check here, I'm happy to give it a try, I just guessed that touching the upstream logic might be a much bigger and heavier change.
Let me know what you think is best.

fendor:
If you validate your PR against the cache dirs on your local system, do you think your patch would work?
Iirc, there is no guarantee the hashes would be the same, and I think they aren't always identical. 

Fixing upstream is a far better solution than monkey patching in HLS, imo :)
However, not appending the hash when there is a hash already can be entirely achieved in HLS, I believe.

adpad-13:
Ah , thanks I just realised that now , if the hashes are different it will be appended no matter what .
to handle it entirely in HLS like you mentioned, we can implement a logic to check if a valid Cabal hash is already present in the ID, and just skip appending a new one. I'll have to look up exactly how Cabal formats its hashes .

fendor:
Out of curiosity, how did you figure out this hash is appended by cabal or is a cabal hash?

adpad-13:
I'm still learning how the different layers interact, so i mostly figured this out just by understanding what I could by looking at my local files, I was not really sure that the hash was added by cabal but since i saw the the prefix already had an appended hash
I assumed the hash was being added by Cabal, but after checking the dist-newstyle cache and running a verbose build, I saw that the ids there actually do not have that hash. However, the folders being generated in ~/.cache/ghcide/ definitely have a hash attached to the prefix. This led me to deduce that the hash is being appended in between ,somewhere after the build process but before the id is passed to ghcide, so assumed that cabal was adding those hashes.
Since I wasn't sure exactly where that upstream logic lives or how that works, I focused on adding a logic to the getCacheDirsDefault as discussed earlier ,to handle these pre-hashed IDs.
After looking at my local files  , i noticed most of them  used 40-digit hashes .
So I have updated the logic , i set a lower limit of 32 characters for the hasHashSuffix just to be safe.
It checks if the suffix of the uid given to the function contains valid hex digits.


My local tests confirm this prevents the double-hash.




Let me know if you'd like me to push the commit

fendor:
> My local tests confirm this prevents the double-hash.

The screenshot you are sending never had any double hashes, the double hashes are only added when the unit id is `main`. So I am curious what local tests did you perform?

---

Your reasoning seems very circular, are you saying, you assume it is a cabal hash, although it doesn't look like cabal generated uses this hash, but you have no better explanation than it being a cabal hash?
To cut to the chase, no, this is not a cabal hash, we generate the first hash here: https://github.com/haskell/haskell-language-server/blob/master/ghcide/session-loader/Development/IDE/Session.hs#L1171C43-L1171C44

The idea of stripping the last part of the hash is wrong. Both hashes are present for a reason.
While I fully acknowledge that it is basically impossible for you to know why these hashes are necessary, you need to be careful when proposing changes you do not understand the ramifications of.

I updated the original issue description, which hopefully explains the solution in more detail, instead of simply stripping the last hash away.

adpad-13:
oh , i'm sorry , i meant to share this 
beforefix
afterfix



> Your reasoning seems very circular, are you saying, you assume it is a cabal hash, although it doesn't look like cabal generated uses this hash, but you have no better explanation than it being a cabal hash? To cut to the chase, no, this is not a cabal hash, we generate the first hash here: https://github.com/haskell/haskell-language-server/blob/master/ghcide/session-loader/Development/IDE/Session.hs#L1171C43-L1171C44
> 
You are totally right that my explanation was confusing , to clarify my thought process: I initially assumed it was Cabal adding the hash, but when I checked the dist-newstyle cache and verbose build logs, i realized Cabal wasn't generating it. I deduced the hash was being appended somewhere in between the build and ghcide, but since I did not go through the whole thing as i was fixated on shortening the path by stripping the last hash, I didn't know exactly where that was happening. Thank you for pointing me directly to those lines of code.
> The idea of stripping the last part of the hash is wrong. Both hashes are present for a reason. While I fully acknowledge that it is basically impossible for you to know why these hashes are necessary, you need to be careful when proposing changes you do not understand the ramifications of.
> 
Understood, I will drop this idea of stripping the last part .
> I updated the original issue description, which hopefully explains the solution in more detail, instead of simply stripping the last hash away.
>
Thanks for the update. I see now the goal is to thread that original ByteString hash to getCacheDirsDefault and combine them into a single hash.
I'm going to take a few days to properly trace the data flow through Session.hs and work on this refactor. Thanks for the patience and guidance!

fendor:
Thank you, this looks much better now. I kind of still dislike the tuple, what do you say to replacing the tuple you modified into a proper datatype?

What do you think about this?

Referring to this tuple in particular: `(DynFlags, [GHC.Target],Maybe B.ByteString)`

A potential name could be something again `HomeUnitConfig`, but I am not 100% sure.

Also, please make sure you use the whitespace consistently, like adding a space after a comma in tuples, e.g. `(a, b)` instead of `(a,b)`

Afterwards, I think this is ready to merge :)

Note, you will also have to rebase this PR, as we have merged a quite a number of changes to `Session.hs` recently.

adpad-13:
Hi, thanks for the heads-up, I see all the hashing logic was moved to Ghc.hs , I agree to use a datatype named `HomeUnitConfig` instead of that tuple

adpad-13:
Hey @fendor,
I've rebased the PR and pushed the updates based on your feedback. Specifically, I added a type for the monadic tuple `HomeUnitConfig` and also applied the hashing in `Ghc.hs` (since the hashing logic was moved there).
Kindly have another look whenever you're free

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -682,8 +683,10 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
               , sessionLoadingOptions = newSessionLoadingOptions
               }
 
+
```
fendor:
```suggestion
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -682,8 +683,10 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
...
         writeTQueue que (runReaderT (getOptionsLoop recorder sessionShake sessionState knownTargetsVar) sessionEnv)
 
+
```
fendor, line 687:
```suggestion
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -1072,8 +1075,10 @@ toFlagsMap :: TargetDetails -> [(NormalizedFilePath, (IdeResult HscEnvEq, Depend
 toFlagsMap TargetDetails{..} =
     [ (l, (targetEnv, targetDepends)) | l <-  targetLocations]
 
+
```
fendor:
```suggestion
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -1072,8 +1075,10 @@ toFlagsMap :: TargetDetails -> [(NormalizedFilePath, (IdeResult HscEnvEq, Depend
...
 -- | Mapping from @hie.yaml@ to all components that have been loaded
 -- from this @hie.yaml@ location.
+
```
fendor:
```suggestion
```

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -1102,6 +1107,7 @@ memoIO op = do
                 return (Map.insert k res mp, res)
             Just res -> return (mp, res)
 
+
```
fendor:
```suggestion
```

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -422,14 +431,18 @@ setCacheDirs recorder CacheDirs{..} dflags = do
           & maybe id setHieDir hieCacheDir
           & maybe id setODir oCacheDir
 
-getCacheDirsDefault :: String -> [String] -> IO CacheDirs
-getCacheDirsDefault prefix opts = do
-    dir <- Just <$> getXdgDirectory XdgCache (cacheDir </> prefix ++ "-" ++ opts_hash)
+getCacheDirsDefault :: String -> Maybe B.ByteString -> [String] -> IO CacheDirs
+getCacheDirsDefault prefix mFirstHash opts = do
```
fendor, line 452: As this function got more complicated, let's give it some proper haddock docs!

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -95,6 +101,8 @@ data RawComponentInfo = RawComponentInfo
   -- | Maps cradle dependencies, such as `stack.yaml`, or `.cabal` file
   -- to last modification time. See Note [Multi Cradle Dependency Info].
   , rawComponentDependencyInfo :: DependencyInfo
+  -- | The raw ByteString for the hash generated in setoptions for the uid "main"
```
fendor:
```suggestion
  -- | An optional hash seed generated in 'setOptions' for the unit id "main".
```

If you quote function names and definitions like this, then they are linked in HLS, and you can even `gotoDefinition` for `setOptions` in the docs.

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -80,6 +81,11 @@ instance Pretty Log where
     LogDLLLoadError errorString ->
       "Error dynamically loading libm.so.6:" <+> pretty errorString
 
+data HomeUnitConfig = HomeUnitConfig
+  { homeUnitDynFlags :: DynFlags
+  , homeUnitTargets  :: [GHC.Target]
+  , homeUnitHash     :: Maybe B.ByteString
+  }
```
fendor, line 94:
```suggestion
data HomeUnitConfig = HomeUnitConfig
  { homeUnitDynFlags :: DynFlags
  , homeUnitTargets  :: [GHC.Target]
  , homeUnitHash     :: Maybe B.ByteString
  }

```

Some superficial docs would be nice, especially `homeUnitHash` should have a word or two for its purpose.
Something along the line of ``Optional hash seed needed to tell home units apart that have the same `-this-unit-id`. We use this when the `-this-unit-id` is "main", which is common when loading single targets.``

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -40,6 +40,7 @@ import           Data.List.Extra                     as L
 import qualified Data.Map.Strict                     as Map
 import           Data.Maybe
 import           Data.Proxy
+
```
fendor:
```suggestion
```

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -79,7 +80,18 @@ instance Pretty Log where
...
+-- | Configuration info for a particular home unit.
+data HomeUnitConfig = HomeUnitConfig
+  {
+  -- | The dynamic flags to compile this specific unit.
+    homeUnitDynFlags :: DynFlags
+  -- | All the targets for this unit.
+  , homeUnitTargets  :: [GHC.Target]
+  -- | Optional hash seed to differentiate home units
+  -- with same `-this-unit-id`. Used when `-this-unit-id` is "main"
+  -- , common when loading a single target.
```
fendor:
```suggestion
  -- with same `-this-unit-id`. Used when `-this-unit-id` is "main",
  -- which is common when loading a single target.
```

adpad-13:
One of the checks is failing , can you please re run that

PR #4856, opened 2026-02-28
Use Server side Diagnostics

### LINKED ISSUE #4805: CodeAction Handlers should use Server-side known Diagnostics
In #4056, it was established that HLS should not rely on the Diagnostics provided by the client to compute possible CodeActions.

Since then, we have introduced the functions `activeDiagnosticsInRangeMT` and `activeDiagnosticsInRange` in `ghcide/src/Development/IDE/Core/PluginUtils.hs` and use it in some locations.

The next step is to review all occurrences of `mkPluginHandler SMethod_TextDocumentCodeAction` and make sure we use these instead of the Diagnostics provided by the client.

So the task is to review such handlers, use `activeDiagnosticsInRangeMT` or `activeDiagnosticsInRange` respectively and then make sure the tests still pass, as lsp-test sometimes does something weird with source ranges, leading unexpected changes in the tests.

vidit-od:
Fixes #4805 

This PR aims to deliver server-side diagnostics to codeactions. 
The Comment below linked issue by @fendor  stated 
> Ideally, we come up with some API or API changes to the plugin handlers that allow us to omit the step of manually calling activeDiagnosticsInRange* and automatically do the right thing... But I don't know right away how.

This is the approach i followed, hopefully it aligns with the expectations : 

- exposed a new API named `mkCodeActionHandlerWithDiagnostics`
- previously users called ` mkPluginHandler SMethod_TextDocumentCodeAction`
- now users need to call this API instead, 
- internally `activeDiagnosticsInRange` is used to fetch server-side diagnostics
- if file not published / server-side diagnostics inaccessible then use client side diagnostics 
- call `mkPluginHandler SMethod_TextDocumentCodeAction` with this fresh diagnostics ( if available ). 

Edit : 
the above mentioned approach has been discarded. The new approach follows : 

- making an injectDiagnostics function that uses activeDiagnosticsinRange to fetch server side diagnostics. 
- if no server side diagnostics are found then return an empty list. 
- this function is later used after intercepting codeactions requests and used to inject server diagnostics.

### DIFF
--- a/ghcide/src/Development/IDE/Core/PluginUtils.hs
+++ b/ghcide/src/Development/IDE/Core/PluginUtils.hs
@@ -225,16 +225,14 @@ activeDiagnosticsInRange :: MonadIO m => Shake.ShakeExtras -> NormalizedFilePath
 activeDiagnosticsInRange ide nfp range = runMaybeT (activeDiagnosticsInRangeMT ide nfp range)
 
 -- Prefer server-side diagnostics if available; they are authoritative.
--- Fall back to client-supplied diagnostics when none are found.
 injectServerDiagnostics :: IdeState -> CodeActionParams -> IO CodeActionParams
-injectServerDiagnostics ide params@LSP.CodeActionParams{_textDocument=LSP.TextDocumentIdentifier{_uri}, _range, _context=context} = do
-  let clientDiags = context ^. LSP.diagnostics
+injectServerDiagnostics ide params@LSP.CodeActionParams{_textDocument=LSP.TextDocumentIdentifier{_uri}, _range} = do
   serverDiags <- case LSP.uriToNormalizedFilePath (LSP.toNormalizedUri _uri) of
-    Nothing  -> pure clientDiags
+    Nothing  -> pure []
     Just nfp -> do
       mDiags <- activeDiagnosticsInRange (shakeExtras ide) nfp _range
       case mDiags of
-        Nothing    -> pure clientDiags
+        Nothing    -> pure []
         Just diags -> pure $ diags ^.. traverse . fdLspDiagnosticL
   pure $ params & LSP.context . LSP.diagnostics .~ serverDiags
 
--- a/ghcide/src/Development/IDE/Plugin/HLS.hs
+++ b/ghcide/src/Development/IDE/Plugin/HLS.hs
@@ -48,12 +48,12 @@ import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types
 import qualified Language.LSP.Server              as LSP
 import           Language.LSP.VFS
-import           Prettyprinter.Render.String   (renderString)
-import           Text.Regex.TDFA.Text          ()
-import           UnliftIO                      (MonadUnliftIO, liftIO,
-                                                readTVarIO)
-import           UnliftIO.Async                (forConcurrently)
-import           UnliftIO.Exception            (catchAny)
+import           Prettyprinter.Render.String      (renderString)
+import           Text.Regex.TDFA.Text             ()
+import           UnliftIO                         (MonadUnliftIO, liftIO,
+                                                   readTVarIO)
+import           UnliftIO.Async                   (forConcurrently)
+import           UnliftIO.Exception               (catchAny)
 
 -- ---------------------------------------------------------------------
 --
--- a/plugins/hls-hlint-plugin/test/Main.hs
+++ b/plugins/hls-hlint-plugin/test/Main.hs
@@ -218,8 +218,8 @@ suggestionsTests =
         doc <- openDoc "TwoHints.hs" "haskell"
         _ <- hlintCaptureKick
 
-        firstLine <- map fromAction <$> getCodeActions doc (mkRange 0 0 0 0)
-        secondLine <- map fromAction <$> getCodeActions doc (mkRange 1 0 1 0)
+        firstLine <- map fromAction <$> getCodeActions doc (mkRange 0 0 1 0)
+        secondLine <- map fromAction <$> getCodeActions doc (mkRange 1 0 2 0)
         thirdLine <- map fromAction <$> getCodeActions doc (mkRange 2 0 2 0)
         multiLine <- map fromAction <$> getCodeActions doc (mkRange 0 0 2 0)
 


vidit-od:
Below is a small documentation of a different API Design that i tried exploring. 

## Origin : 
if we notice the use of our new API in `change-type-signature-plugin`. we can observe the use of ActiveDiagnosticinRange as well as our API.  
 
- Our API uses this same function internally as well
- i.e for each call we will be using it 2x 
- This is inefficient, but done just to justify the goal of this issue. 


## attempted solution

TLDR : case (File Not published ) => empty codeaction on first load :)

- To fix the above issue i tried to update our API to accept `[FileDiagnostic]`
- This leads to a conflict in fallback case ( when file is not published and we have to rely on clint side )
- client-supplied diagnostics are of type [LSP.Diagnostic], and since FileDiagnostic is a richer server-side type, we cannot reconstruct it from [LSP.Diagnostic]
- essentially forcing us to return `Nothing` here 

## Pros :
More versatile API ( and efficient )

## Cons : 
Fall back case results to no codeaction

plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
vidit-od, line 99: This usecase is inefficient. Using the old `mkPluginHandler` is better here;  

Refer linked documentation for more context. [[Link]](https://github.com/haskell/haskell-language-server/pull/4856#issuecomment-3976678612)

fendor: Right, this is some very annoying detail we need to figure out, as you point out, we just do the same work twice...

fendor:
Using a separate primitive for writing CodeAction handlers does the job, but can we do better?
Could we, for example, interject in ghcide all `CodeAction` requests and replace the Diagnostics once and for all, without plugin authors having to remember that they should use something else for `CodeAction` handlers only?

I think we perform some similar special logic in ghcide for certain requests?

vidit-od:
I began working with this same logic that you present. 

I tried updating the f' helper that is responsible for intercepting and calling for codeaction in mkpluginhandler. 

Later discarded it because that felt too complicated. 

But yea, i can surly work on it. Just so that we are on the same page, our expectation is :

Basically, update mkpluginhandler such that it does everything our new API does. Helping users to remain as they were before on the caller site.

vidit-od:
I think this does what we were expecting. Surprisingly simpler than what i was trying to do before so thanks for the direction :)

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -252,26 +252,29 @@ extensiblePlugins recorder plugins = mempty { P.pluginHandlers = handlers }
     handlers = mconcat $ do
       (IdeMethod m :=> IdeHandler fs') <- DMap.assocs handlers'
       pure $ requestHandler m $ \ide params -> do
-        vfs <- readTVarIO $ vfsVar $ shakeExtras ide
+        -- Intercept CodeAction requests and inject server-side diagnostics
+        params' <- case m of
+          SMethod_TextDocumentCodeAction -> liftIO $ injectServerDiagnostics ide params
+          _                              -> pure params
```
fendor:
```suggestion
pure $ requestHandler m $ \ide params' -> do
        -- Intercept CodeAction requests and inject server-side diagnostics
        params <- case m of
          SMethod_TextDocumentCodeAction -> liftIO $ injectServerDiagnostics ide params'
          _                              -> pure params'
```

In this case, I prefer changing the input parameter name!

Also, the comment should explain why we do this!

fendor:
Could you also expand the commit message to contain more information? `Server Diagnostic` is really not enough, a proper commit message would be good for such an internal and somewhat tricky change.

fendor:
Are there any call sites of `activeDiagnosticsInRange` we can now get rid of?

Also, you have to rebase this PR as there is a conflict.

ghcide/src/Development/IDE/Plugin/HLS.hs
```diff
@@ -251,8 +252,15 @@ extensiblePlugins recorder plugins = mempty { P.pluginHandlers = handlers }
         PluginHandlers hs = HLS.pluginHandlers pluginDesc
     handlers = mconcat $ do
       (IdeMethod m :=> IdeHandler fs') <- DMap.assocs handlers'
-      pure $ requestHandler m $ \ide params -> do
+      pure $ requestHandler m $ \ide params' -> do
         vfs <- readTVarIO $ vfsVar $ shakeExtras ide
+        -- issue https://github.com/haskell/haskell-language-server/issues/4056
+        -- established that HLS should rely on server-side diagnostics to compute codeaction
+        -- To ensure consistency, we intercept codeAction requests and explicitly inject
+        -- server-side diagnostics before delegating to the plugin handler.
```
fendor:
```suggestion
        -- In issue https://github.com/haskell/haskell-language-server/issues/4056,
        -- we established that HLS should rely on server-side diagnostics to compute 'CodeAction's.
        -- To ensure consistency across plugins, we intercept 'CodeAction' requests and explicitly inject
        -- server-side diagnostics before delegating to the plugin handler.
```

vidit-od:
> Are there any call sites of activeDiagnosticsInRange we can now get rid of?

I don't think we can replace any of the `activeDiagnosticsInRange`. All of the plugins that use it expect [Filediagnostics]. This is later used to extract `fdStructuredMessageL`.

Below is the list of all plugins that uses `activeDiagnosticsInRange` and i investigated : 
- `hls-change-type-signature-plugin`
- `hls-class-plugin`
- `hls-pragmas-plugin`

> Also, you have to rebase this PR as there is a conflict.

I have Rebased it on top of master;

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -223,6 +224,20 @@ activeDiagnosticsInRangeMT ide nfp range = do
 activeDiagnosticsInRange :: MonadIO m => Shake.ShakeExtras -> NormalizedFilePath -> LSP.Range -> m (Maybe [FileDiagnostic])
 activeDiagnosticsInRange ide nfp range = runMaybeT (activeDiagnosticsInRangeMT ide nfp range)
 
+-- Prefer server-side diagnostics if available; they are authoritative.
+-- Fall back to client-supplied diagnostics when none are found.
```
fendor:
```suggestion
-- | Replace client-supplied diagnostics with the server-side computed diagnostics if available; 
-- they are authoritative.
--
-- Falls back to client-supplied diagnostics when none are found.
```
Please always use haddock style comments for top-level functions https://haskell-haddock.readthedocs.io/latest/ :)

The fallback behaviour is also quite interesting, why did you decide to do this?
What if a client sends outdated diagnostics and the server determined all Diagnostics have been resolved since the client last asked for any?

I think this (semi?) accidentally avoids breaking a lot of our tests, as `lsp-test` sends all diagnostics when requesting code actions, allowing us to get away with inaccurate `Range`s in the tests. So, I am slightly torn, but I feel like the fallback behaviour is not the best idea?

vidit-od: Well this is interesting, i did not think of this; 
TBH, I thought availability of client side diagnostics is more guaranteed over server-side diagnostics. ( just a guess ). 

But since you have mentioned, we can remove the fall back part. while we are at it. i can also try and solve this issue [[link]](https://github.com/haskell/haskell-language-server/pull/4856#discussion_r2867248789). The main blocker here was lack of clientside diagnostics in fallback.

> I think this (semi?) accidentally avoids breaking a lot of our tests, as lsp-test sends all diagnostics when requesting code actions, allowing us to get away with inaccurate Ranges in the tests

yea, but first things first. will test this after removing fallback !
Thanks :)

fendor: > TBH, I thought availability of client side diagnostics is more guaranteed over server-side diagnostics. ( just a guess ).

How should that be the case, where does the client get its diagnostics from?

vidit-od: From the server :) 
That was a blunder from my part, mb !

fendor (approved):
LGTM, thank you!

vidit-od:
I have rebased on top of Master; 
should pass all of the previous failing tests now ! 

Thanks a lot for reviewing and co-authoring this PR 🚀

PR #4861, opened 2026-03-03
Draft: Cabal rename module

### LINKED ISSUE #713: Add garbage collect of diagnostics
Currently if you have a file open with errors, then shut it, the errors persist. The reason is that we don't have any sophisticated removal of non-interesting files, unlike what we have with DAML.

VeryMilkyJoe:
Adds handling of renaming of modules.

The plan is for a rename notification to trigger a rename of the respective module in the cabal file, the module's header and all relevant imports. 
Also a complete session reloading is triggered after a rename action has taken place.

All of this works as of now but i need to split up the implementation to the relevant places in HLS, right now it all takes place in the cabal plugin, while ideally only the renaming in the `.cabal` file should.

Closes #713

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -5,7 +5,6 @@ packages:
          ./ghcide
          ./hls-plugin-api
          ./hls-test-utils
-         ../lsp/lsp
 
 index-state: 2026-02-24T00:00:00Z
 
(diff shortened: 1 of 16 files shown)


VeryMilkyJoe:
I have split this into separate PRs, to make it easier to review (See issue #4951).

PR #4862, opened 2026-03-04
Only show linear arrows on data constructors with LinearTypes

### LINKED ISSUE #2904: Type-on-hover for type constructors shows linear arrows even if -XLinearTypes is not enabled
This is probably confusing for newcomers who don't know about `LinearTypes`, even if technically correct according to the user's guide 

> By default, all fields in algebraic data types are linear 
> -- https://downloads.haskell.org/ghc/9.0.2/docs/html/users_guide/exts/linear_types.html#data-types

GHCi gets around this because, also according to the user's guide,

> [...] Given `data T1 a = MkT1 a` [...] [w]hen used as a value, `MkT1` is given a multiplicity-polymorphic type: `MkT1 :: forall {m} a. a %m -> T1 a`. [...] When displaying types, unless -XLinearTypes is enabled, multiplicity polymorphic functions are printed as regular functions [...]; therefore constructors appear to have regular function types.
> -- https://downloads.haskell.org/ghc/9.0.2/docs/html/users_guide/exts/linear_types.html#data-types

The solution, then, would be to use `MkT1`-the-value for the type-on-hover tooltip, instead of `MkT1`-the-constructor. 

### Your environment

Which OS do you use: macOS
Which LSP client (editor/plugin) do you use: VSCode
Describe your project (alternative: link to the project): plain cabal project, ghc-9.0.2

### Steps to reproduce
Define a newtype or data type and hover on one of its constructors.

```haskell
newtype WrappedInt = MkWrapped Int
```

### Expected behaviour
The tooltip shows the type `MkWrapped :: Int -> WrappedInt`

### Actual behaviour
The tooltip shows the type `MkWrapped :: Int %1 -> WrappedInt`, which includes linear arrows.

crtschin:
Closes #2904.

Does what it says on the tin. Mostly follows the investigation @July541 already did [here](https://github.com/haskell/haskell-language-server/issues/2904#issuecomment-1126969727). Some notes:
- I added a usage of parsed module as I couldn't get the extensions active in the current module from the `HscEnv`, and I didn't know how to do so otherwise.
- Turning the `DataCon` into a `Type` used to occur through `dataConWrapId`, which I changed to `dataConDisplayType`, which also displays wrapper types from what I can tell.
- I also touched `ghcide/src/Development/IDE/Plugin/Completions.hs`, because I changed the signature of `safeTyThingType`. I didn't understand what was being completed so I stuck to always emitting linear annotations there.


July541:
Thanks for this fantastic work! ❤️❤️❤️

Unfortunately I can't make the review because I have totally lost the details after 4 years. I didn't get in touch with Haskell for a long time and not familiar with recent ghc versions.

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -282,6 +288,9 @@ atPoint opts@IdeOptions{} shakeExtras@ShakeExtras{ withHieDb, hiedbWriter } har@
         prettyNames <- mapM (prettyName locationsMap) names
         pure (Just range, prettyNames ++ pTypes locationsMap)
       where
+        enabledExtensions :: Util.EnumSet Extension
+        enabledExtensions = getExtensionsSet $ parsedModule
```
fendor: Are the enabled extensions from the `HscEnv`'s `DynFlags` not the correct `DynFlags`, e.g., the `DynFlags` of the module we are calling `atPoint` on?

In other words, I don't think we should need the `ParsedModule` to get the enabled Extensions, but should be able to use the `DynFlags` of the `HscEnv`?

crtschin: I initially tried this, but it doesn't no. I tried using `extensionFlags (hsc_dflags hscEnv)`, which doesn't detect the `{-# LANGUAGE LinearTypes #-}` inline language pragmas ([these](https://github.com/haskell/haskell-language-server/pull/4862/changes#diff-3cb23dd82e42f5d90f1d51c7085c68ba9b9798c13b346281e67de985055ea074R26-R30) tests fail to show the linear arrow).

ghcide/src/Development/IDE/Spans/Common.hs
```diff
@@ -39,22 +38,18 @@ import           Language.LSP.Protocol.Types
 type DocMap = NameEnv SpanDoc
 type TyThingMap = NameEnv TyThing
 type ArgDocMap = NameEnv (IntMap SpanDoc)
+data A = A Word Bool
```
fendor: What's this?

ghcide/src/Development/IDE/Spans/AtPoint.hs
fendor: Make sense, we use `GhcSession` to get the `HscEnv`, but what we would need is the `GhcSessionDeps` result, I think. If you want, you could try it, but it is overkill and may have other unexpected side effects.

As we only care about the Extension set, I think we should only pass `Util.EnumSet Extension` to the `atPoint` function. 

Moreover, instead of `GetParsedModule`, could you try to use `GetModSummary` and use the `ms_hspp_opts`'s `extensions` field?

I think this should succeed more often, as it doesn't require the full module to be parsed.

ghcide/src/Development/IDE/Core/Actions.hs
```diff
@@ -52,12 +52,13 @@ getAtPoint file pos = runMaybeT $ do
   shakeExtras <- lift askShake
 
   env <- hscEnv . fst <$> useWithStaleFastMT GhcSession file
+  parsedModule <- fst <$> useWithStaleFastMT GetParsedModule file
```
fendor: I think here `GetModSummary`, then getting its `DynFlags` and the `Extension` should suffice as well.

ghcide/src/Development/IDE/Spans/Common.hs
crtschin: Ah woopsie, was me debugging if this worked 😅.

ghcide/src/Development/IDE/Core/Actions.hs
crtschin: Nice!

PR #4864, opened 2026-03-09
new snapshot for stack build

### LINKED ISSUE #4696: Upgrade `stack*.yaml` files
They use rather old snapshots, we should upgrade them to more recent ones!

vidit-od:
Fixes #4696 

This PR does : 
- remove `ghc 9.6` stack build option 
- introduces `ghc 9.10` stack build 
- fixes a few of the redundant imports that caused build problems. 

this now enables us with 2 stack builds 
- Ghc 9.8 ( `stack.yaml` )
- Ghc 9.10 ( `stack-lts24.yaml` )

### DIFF
--- a/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
+++ b/plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE CPP #-}
-
 module Ide.Plugin.Notes (descriptor, Log) where
 
 import           Control.Lens                     ((^.))
@@ -10,10 +8,10 @@ import qualified Data.Array                       as A
 import           Data.HashMap.Strict              (HashMap)
 import qualified Data.HashMap.Strict              as HM
 import qualified Data.HashSet                     as HS
-import           Data.List                        (uncons)
+import           Data.List                        as List
 import           Data.Maybe                       (catMaybes, fromMaybe,
                                                    listToMaybe, mapMaybe)
-import           Data.Text                        (Text, intercalate)
+import           Data.Text                        (Text)
 import qualified Data.Text                        as T
 import qualified Data.Text.Utf16.Rope.Mixed       as Rope
 import           Data.Traversable                 (for)
@@ -33,10 +31,6 @@ import           Text.Regex.TDFA                  (Regex, caseSensitive,
                                                    defaultCompOpt,
                                                    defaultExecOpt,
                                                    makeRegexOpts, matchAllText)
-
-#if !MIN_VERSION_base(4,20,0)
-import           Data.Foldable                    (foldl')
-#endif
 
 data Log
     = LogShake Shake.Log
@@ -72,7 +66,7 @@ instance Pretty Log where
             LogNoteReferencesFound file refs -> "Found note references in " <> prettyNotes file refs
             LogNotesFound file notes -> "Found notes in " <> prettyNotes file notes
         where prettyNotes file hm = pretty (show file) <> ": ["
-                <> pretty (intercalate ", " (fmap (\(s, p) -> "\"" <> s <> "\" at " <> intercalate ", " (map (T.pack . show) p)) hm)) <> "]"
+                <> pretty (T.intercalate ", " (fmap (\(s, p) -> "\"" <> s <> "\" at " <> T.intercalate ", " (map (T.pack . show) p)) hm)) <> "]"
 
 {-
 The first time the user requests a jump-to-definition on a note reference, the
@@ -105,7 +99,7 @@ findNotesRules recorder = do
                 references <- fmap snd <$> use MkGetNotesInFile nfp
                 pure $ fmap (HM.map (fmap (nfp,))) references
             )
-        pure $ Just $ foldl' (HM.unionWith (<>)) HM.empty definedReferences
+        pure $ Just $ List.foldl' (HM.unionWith (<>)) HM.empty definedReferences
 
 err :: MonadError PluginError m => Text -> Maybe a -> m a
 err s = maybe (throwError $ PluginInternalError s) pure
@@ -180,7 +174,7 @@ findNotesInFile file recorder = do
     let refMatches = (A.! 1) <$> matchAllText noteRefRegex content
         refs = toPositions refMatches content
     logWith recorder Debug $ LogNoteReferencesFound file (HM.toList refs)
-    pure $ Just (HM.mapMaybe (fmap fst . uncons) notes, refs)
+    pure $ Just (HM.mapMaybe (fmap fst . List.uncons) notes, refs)
     where
         uint = fromIntegral . toInteger
         -- the regex library returns the character index of the match. However
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,11 +58,11 @@ defaults: &defaults
 
 version: 2
 jobs:
-  stackage-lts22:
+  stackage-lts24:
     docker:
-      - image: haskell:9.6.7-slim-bullseye
+      - image: haskell:9.10.3-slim-bullseye
     environment:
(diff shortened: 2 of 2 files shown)


vidit-od:
Edit : I re-ran circleci failure and it passed this time. 
Any reason for this flakyness ? 

Original :
Regarding Stackage-lts23 failure. 
Need some direction here on how to update the Stack version. 

This is wrt ghc 9.8 version ( corresponding to `stack.yaml` )
The CircleCi logs say : 

```
Warning: You are currently using Stack
         version 3.3.1 but version 3.9.3
         is available.
         
         You can try to upgrade by running
         stack upgrade
         
         Tired of seeing this? Add
         recommend-stack-upgrade: false
         to /root/.stack/config.yaml.
```

fendor:
@vidit-od I think we shouldn't bother too much with circle ci, let's get this merged and then migrate the circle ci jobs to github. We just need to be careful not overstretch our cache size.

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -33,6 +34,10 @@ import           Text.Regex.TDFA                  (Regex, caseSensitive,
                                                    defaultExecOpt,
                                                    makeRegexOpts, matchAllText)
 
+#if !MIN_VERSION_base(4,20,0)
+import           Data.Foldable                    (foldl')
+#endif
```
fendor: This feels incorrect, we have been building HLS with GHC <9.10 for a while, why do we need this?

vidit-od:
Edit : 
ok i was wrong here, Flag tests do do a one more thing that makes them important.  They test normal build & build with non default flags for all the ghc variants. There is no way to test the latter part using testing github actions. 

below suggestion can be ignored :) 

original : 
@fendor makes sense; 
i am not sure if this will help but we can also investigate about Flag tests on github maybe ? 
although, i am not sure if it will have a significant ( or any ) form of cache relaxation

from what i understand, those tests are just for build checks only. we can delegate this to our Testing tests on github. The only extra thing these tests provide is : if any one fails all of the other flag tests gets cancelled. 

thank you for reviewing the PR ;

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
```diff
@@ -20,6 +19,10 @@ import           Language.LSP.Protocol.Types             (Range)
 import           Control.Lens                            (_head, over)
 #endif
 
+#if !MIN_VERSION_ghc_exactprint(1,10,0)
+import           Data.Functor.Identity                   (Identity)
+#endif
```
fendor, line 24: Same for this, why do we need this?

plugins/hls-stylish-haskell-plugin/src/Ide/Plugin/StylishHaskell.hs
```diff
@@ -25,9 +25,12 @@ import           Ide.PluginUtils
 import           Ide.Types                        hiding (Config)
 import           Language.Haskell.Stylish
 import           Language.LSP.Protocol.Types      as LSP
-import           System.Directory
 import           System.FilePath
 
+#if !MIN_VERSION_stylish_haskell(0,15,0)
+import           System.Directory
+#endif
```
fendor, line 32: Probably for `-Werror`?  😬 Can we avoid this somehow?

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
vidit-od: when building with our new stack snapshot, these parts give redundant import error and fails on build. 
Error logs 
```
G:\coding\github\haskell-language-server\plugins\hls-notes-plugin\src\Ide\Plugin\Notes.hs:36:1: error: [GHC-66111]
[-Wunused-imports, Werror=unused-imports]
haskell-language-server>     The import of `Data.Foldable' is redundant
haskell-language-server>       except perhaps to import instances from `Data.Foldable'
haskell-language-server>     To import instances alone, use: import Data.Foldable()
haskell-language-server>    |
haskell-language-server> 36 | import           Data.Foldable                    (foldl')
haskell-language-server>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
haskell-language-server>
```

<details>
<summary>Reasoning of why this happens </summary>
a few LLM searchs gave this : 

`foldl'` was deliberately excluded from Prelude in base-4.8 and has lived only in `Data.Foldable` ever since — until base-4.21 / GHC 9.10, where it was finally added to Prelude. That's why the import `Data.Foldable (foldl')` was valid and necessary on all previous GHC versions, but becomes a redundant import on GHC 9.10, causing the `-Wunused-imports` error which then fails the build due to `pedantic: true`.

</details>

below are some of the files following similar pattern : 
- [Profile.hs](https://github.com/haskell/haskell-language-server/blob/cf36960042faa6f07a832c28b1a0135fb5fd0c6b/hls-graph/src/Development/IDE/Graph/Internal/Profile.hs#L36)
- [Types.hs](https://github.com/haskell/haskell-language-server/blob/cf36960042faa6f07a832c28b1a0135fb5fd0c6b/hls-plugin-api/src/Ide/Types.hs#L109)
- [QualifyImportedNames.hs](https://github.com/haskell/haskell-language-server/blob/cf36960042faa6f07a832c28b1a0135fb5fd0c6b/plugins/hls-qualify-imported-names-plugin/src/Ide/Plugin/QualifyImportedNames.hs#L79)
- [Splice.hs](https://github.com/haskell/haskell-language-server/blob/cf36960042faa6f07a832c28b1a0135fb5fd0c6b/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L57)

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
vidit-od: same as 1

vidit-od:
@fendor , i tried to eliminate the conditional imports but i could not find any clean way to do that. 
is there any way to ignore redundant import errors maybe ?

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
fendor: This one we can work around by explicitly using the `foldl'` from https://hackage-content.haskell.org/package/unordered-containers-0.2.21/docs/Data-HashMap-Strict.html#v:foldl-39-

This import is then always used.

fendor:
Thanks, one CPP we can avoid, the others also already contain CPP, so it is fine.

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
vidit-od: `data.foldable.foldl'` and `data.hashmap.strict.foldl'` have a little difference in signatures. 
former works with `lists` where as latter works on `hashsets`. 

Using the latter will require us to do a janky fix like : 
```haskell
pure $ Just $ HM.foldl' (HM.unionWith (<>)) HM.empty
    $ HM.fromList $ zip ([0..] :: [Int]) definedReferences
```

fendor: Wait, I think I misunderstood, `definedReferences :: [HashMap Text [(NormalizedFilePath, Position)]]`, so we are iterating over a list.

Then `Data.List.foldl'` ought to work, right? Simply import `import qualified Data.List as List`, and use it for `List.unconcs` and `List.foldl'`

vidit-od: yea, this works :)

PR #4867, opened 2026-03-16
Improve `prepareRename` responses

### LINKED ISSUE #4866: `hls-rename-plugin`: `prepareRename` returns suboptimal responses

On `textDocument/prepareRename` requests, `hls-rename-plugin` returns either of the following:

- `{defaultBehavior: true}` if the cursor is on a `Name` identifier, which can be renamed
- `{defaultBehavior: false}` if the cursor is anywhere else, to signal that renaming is not supported

There are two disadvantages:

1. HLS depends on the client to support `defaultBehavior`. (This [affected the Zed editor back in 2025](https://github.com/zed-industries/zed/issues/24184#issuecomment-2635252620), when it did not yet support `defaultBehavior`.)
2. HLS does not agree with the LSP specification, which says that if a `rename` request is invalid, the server should return `null`. (This [affects Emacs users to this day](https://github.com/emacs-lsp/lsp-mode/issues/4473).)

`prepareRenameProvider` should simply return the range of the identifier at the cursor position, using the HIE AST (which it already uses to verify that the cursor is on a `Name` identifier).

It should also return `null` for invalid requests.

It might be possible to stop relying on `defaultBehavior` altogether.

At first I tried just replacing `{defaultBehavior: false}` with `null`, but my Emacs setup still occasionally failed to perform renaming.

## Additional context

I think there’s none.

izuzu-izuzu:
Closes #4866
Related to https://github.com/haskell/lsp/issues/636 (upstream)

## Description

`prepareRename` now returns an explicit `Range` of the identifier to rename, or `null` if there is no symbol at the given position.

This follows the LSP 3.17 specification; it also eliminates reliance on client support for `defaultBehavior`.

My notes for this PR can be found in `NOTES.md` in the commit history.

## Details

`prepareRenameProvider` currently finds the smallest HIE AST node at the cursor and checks if it contains a `Name` identifier that can be renamed.

This PR retains this logic, except that it also gets the `RealSrcSpan` of the AST node and returns this span as a `Range` if the node is deemed renameable.


izuzu-izuzu:
Apologies for the force-pushing. I’ll stop doing that.

I only squashed commits that modify `NOTES.md` to reduce noise.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -90,16 +90,20 @@ descriptor recorder pluginId = mkExactprintPluginDescriptor exactPrintRecorder $
...
     --
     -- In particular it allows some cases through (e.g. cross-module renames),
     -- so that the full rename handler can give more informative error about them.
-    let renameValid = not $ null namesUnderCursor
-    pure $ InL $ PrepareRenameResult $ InR $ InR $ PrepareRenameDefaultBehavior renameValid
+    pure $ case namesUnderCursor of
+        [] -> InR Null
+        _ -> case pointCommand hieAst pos nodeSpan of
+            (srcSpan : _) -> InL $ PrepareRenameResult $ InL (realSrcSpanToRange srcSpan)
+            [] -> InR Null
```
fendor: I dont understand why we need a second `pointCommand` here if the only point is to get a `SrcSpan`, can't we get this in one step, combined with `getNamesAtPoint'`?

I might be missing the high level idea, is the workflow roughly:

1. Is there something under the cursor
2. Is there something we can rename, which it only is if the thing under the cursor has a `SrcSpan`?

The `SrcSpan` of the `Name` is no good, I assume?

plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -23,10 +24,55 @@ renamePlugin = mkPluginTestDescriptor Rename.descriptor "rename"
...
+    [ prepareRenameTests
+    , renameTests
     , moduleNameTests
     ]
 
+prepareRenameTests :: TestTree
+prepareRenameTests = testGroup "PrepareRename"
+    [ testCase "Module name (not yet renameable)" $ runRenameSession "" $ do
+        doc <- openDoc "PrepareRename.hs" "haskell"
+        -- REVIEW: The wait is for consistency with 'goldenWithDoc'. Is it necessary?
```
fendor: Yes, it is necessary, as the file needs to be renamed (in other words, the names need to be resolved) before we can actually rename anything. Without this barrier, the request for rename might be received before we finished typechecking.

plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -175,6 +221,14 @@ goldenWithRename title path act =
     goldenWithHaskellDoc (def { plugins = M.fromList [("rename", def { plcConfig = "crossModule" .= True })] })
        renamePlugin title testDataDir path "expected" "hs" act
 
+prepareRename :: TextDocumentIdentifier -> Position -> Session (PrepareRenameResult |? Null)
+prepareRename doc pos = do
+  let params = PrepareRenameParams doc pos Nothing
+  rsp <- request SMethod_TextDocumentPrepareRename params
+  case rsp ^. L.result of
+    Left rspError -> liftIO $ assertFailure $ "prepareRename failed: " <> show rspError
+    Right rspResult -> pure rspResult
```
fendor, line 231: Upstream to `lsp-test`? If yes, please open an issue in `lsp`, and link to it from here.

fendor:
> Apologies for the force-pushing. I’ll stop doing that.

That's fine, you can (and are encouraged) to force push all you want on your PR.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
izuzu-izuzu: **This comment is mostly incorrect. See the comment below.**

> Actually you’re right, we can just use `nameSrcSpan`. I probably didn’t know it existed.
> 
> The high-level flow is:
> 1. Is there a `Name` (or even multiple `Name`s) at the cursor?
> 2. If there is, what’s the smallest `RealSrcSpan` containing the cursor?
> 
> For step 2 we don’t need the full AST.
> 
> I think I got tripped up because `getNamesAtPoint'` might return multiple `Name`s, and so out of caution I consulted the AST one more time in case these `Name`s were generated or something.
> 
> Will fix this.

plugins/hls-rename-plugin/test/Main.hs
izuzu-izuzu: Understood, will remove the review comment.

plugins/hls-rename-plugin/test/Main.hs
izuzu-izuzu: I changed the PR description to mention https://github.com/haskell/lsp/issues/636.

Actually, should I create an upstream PR with the new `prepareRename` and wait for it to get merged first?

fendor: I'd appreciate it if you open a PR, but we won't wait for it to be merged and released, otherwise this PR will be open for another year ;)

PR is great, we will merge it in lsp-test as well, but we will merge this PR without waiting for lsp.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
izuzu-izuzu: Wait, no, I’m sorry. `nameSrcSpan` returns the definition site for a given `Name`, so for an imported name it would return a span outside the current module.

We actually **do** need the AST, though you’re right that we don’t need to call `pointCommand` twice. I can amend `getNamesAtPos` to return `(Name, SrcSpan)` so that we can call it just once.

I forgot to update my own notes. Let me know what I should do to my incorrect comment above this one.

izuzu-izuzu: I wrote a new helper function instead of modifying `getNamesAtPos` so that the actual `rename` logic stays unaffected.

izuzu-izuzu:
I added the high-level approach to the PR description.

What should I do about `NOTES.md` once this PR is accepted? I added it to keep track of notes for myself and also show PR reviewers my thinking process, but I imagine I should delete it?

fendor:
> What should I do about NOTES.md once this PR is accepted? I added it to keep track of notes for myself and also show PR reviewers my thinking process, but I imagine I should delete it?

Yes, remove it from the PR, but you can add the contents to the PR description, or to the commit message if the contents help to understand the commit.

Issue #2646, opened 2022-01-28
Unicode symbols >= 𐀀 are broken


Bodigrim:
### Your environment

Which OS do you use: MacOS
Which LSP client (editor/plugin) do you use: Sublime Code

### Steps to reproduce
```
aaa :: Word 
aaa = 10

𐀀𐀀𐀀 :: Word 
𐀀𐀀𐀀 = 10 
```

### Expected behaviour

Both `aaa` and `𐀀𐀀𐀀` should be underlined as unused bindings. 

### Actual behaviour

`aaa` is underlined correctly, but only the first character of `𐀀𐀀𐀀` is underlined.

<img width="96" alt="изображение" src="https://user-images.githubusercontent.com/2293557/151464453-6749fcbd-18c6-4cdb-b242-696430c50fe9.png">

That's because HLS does not distinguish positions returned from GHC (which are in code points = characters) from positions mandated by LSP (which are UTF-16 code units). Basically, GHC says us that 3 first code points in line 5 are an unused binding. Each `𐀀` is a single character, but 2 UTF-16 code units, so HLS should ask LSP to underline first 6 code units. Instead of this HLS asks LSP to underline only 3, and since the 3rd one is in the middle of the 2nd character, only the 1st character gets underlined.

CC @michaelpj @alanz, this is related to https://github.com/haskell/lsp/pull/392#discussion_r790164996.

Bodigrim:
Code actions are also broken. E. g., choosing "Delete '𐀀𐀀𐀀'" results in the following program
```
aaa :: Word 
aaa = 10 10
```
That's again because HLS meant to ask LSP to delete 8 characters, but asked to delete 8 code points, 6 of which were `𐀀𐀀𐀀`.

michaelpj:
Thanks for the reproduction. To be clear, did you mean that this is related to the issue with splitting in the middle of code units, or is it this one? https://github.com/haskell/lsp/pull/392#discussion_r785379400

That is, do you think this will be fixed by the changes you made to `lsp`, or is there still something broken?

jneira:
In any case a regression test (or several of them) here exercising unicode symbols would be great

Bodigrim:
@michaelpj This is unlikely to be fixed in recent `lsp` patches. Essentially there must be a place, responsible for (fast) conversion between positions in code points and positions in code units. I think I can provide such low-level routine from `text-rope` and then someone will have to migrate `ghcide` to it. 

[...]

PR #4874, opened 2026-03-26
Implement renaming import aliases

### LINKED ISSUE #4873: `hls-rename-plugin` can’t rename import aliases

Sometimes I find it difficult to organize my imports, because `hls-rename-plugin` doesn’t support renaming import aliases.

For example, consider the following code:

```haskell
import qualified Data.List as L

betterTail = L.drop 1
```

Now, suppose I want to rename `L` because I want to import `Control.Lens as L` and refer to `Data.List` as `List` instead.

`hls-rename-plugin` can’t do this yet. Specifically:

- If I place my cursor on `L` in `as L`, the plugin says “No symbol to rename at given location.”
- If I place my cursor on `L` in `L.drop` and attempt to rename `L` to `List`, I get “Renaming of an imported name is unsupported.”

I want to be able to place my cursor on an import alias (whether in an import statement or in a qualified name), and rename the alias.

For example, I should be able go to either `L` below, rename it to `List`, and have all occurrences of `L` changed correctly:

```haskell
-- Before: ----------------------------

import qualified Data.List as L

betterTail = L.drop 1

-- After: -----------------------------

import qualified Data.List as List

betterTail = List.drop 1
```

### Bonus

I also want to be able to rename the following aliases *separately*:

```haskell
-- Before: ----------------------------

import Control.Monad as M
import Data.Maybe as M

isSomething = M.isJust  -- from Maybe
discard = M.void  -- from Monad

-- After: -----------------------------

import Control.Monad as Monad
import Data.Maybe as Maybe

isSomething = Maybe.isJust  -- from Maybe
discard = Monad.void  -- from Monad
```

I tried search-and-replace, but it always felt less convenient and more error prone.

For example, if I have two aliases named `L` and `URL`, doing search-and-replace to rename `L` to `List` might also turn `URL` to `URList`.

## Additional context

I think there’s none.

izuzu-izuzu:
Closes #4873

## Description

This PR adds the ability to rename import aliases.

For example, users can now place their cursor on either `L` below, rename it to `List`, and have all occurrences of `L` changed correctly (and *fast*):

```haskell
-- Before: ----------------------------

import qualified Data.List as L

bar = L.take

-- After: -----------------------------

import qualified Data.List as List

bar = List.take
```

Users can now also rename an alias that is *used by multiple imports*, like this:

```haskell
-- Before: ----------------------------

module User (f, L.drop) where

import Control.Lens as L
import Data.List as L

f = L.take

-- After: -----------------------------

module User (f, List.drop) where

import Control.Lens as L
import Data.List as List

f = List.take
```

My notes for this PR can be found in `NOTES.md` in the commit history.

## Approach

1. Get the parsed AST for the current file.

2. Use the parsed AST to check whether the cursor is on an import alias, either in an import declaration (like `L` in `import Data.List as L`) or in a qualified name (like `L` in `L.take`).

3. Check whether multiple imports use the alias, and perform renaming using appropriate resources:

    - Common case (fast): If only one import uses the alias, perform renaming using only the parsed AST.

    - Special case 1 (slower): If multiple unrelated imports share the alias (like `import Control.Lens as L` and `import Data.List as L` at the same time):

        1. Wait for typechecking (which includes qualified-name resolution) to complete.
        2. Use the `GlobalRdrEnv` of the current file to determine which import the cursor is on.
        3. Find qualified names matching this import and perform renaming using the `GlobalRdrEnv` and parsed AST.

    - Special case 2 (error): If the cursor is on a qualified name exported by multiple modules (like on `L.length` with both `import Data.Foldable as L` and `import Data.List as L`), throw an “ambiguous alias” error and suggest that the user rename the alias in an import declaration instead.

## Considerations

- The plugin now fails early if it can’t find the parsed AST for the current file.

    If no parsed ASTs exist, then it means the file is not yet parsed (and therefore not yet typechecked), and so neither the existing name-renaming logic nor the new alias-renaming logic can work.

- The cursor can be at the start or the end of an alias (or anywhere in between), like the following:

    - `import Data.List as ‸L`
    - `import Data.List as L‸`
    - `‸L.take`
    - `L‸.take`

- The new alias-renaming logic handles Unicode characters robustly.

    It avoids directly converting between GHC source spans (which are in Unicode code points) and LSP ranges (in UTF-16 code units). Instead, it mostly uses `VFS.CodePointPosition` and `VFS.CodePointRange`, as suggested in [HLS issue #2646](https://github.com/haskell/haskell-language-server/issues/2646#issuecomment-1025210155).

    Note that the existing name-renaming logic is still susceptible to this issue (see the `Rename.getNamesAtPos` function).


vidit-od:
Thanks a lot for this PR @izuzu-izuzu 

I still haven't gone through the PR properly but while testing it, found this weird behavior of cursor placement. Any way to fix this maybe ? 

Recording 2026-03-29 024637
 
 it shows cant rename when cursor is on the last position of a symbol. 
 Only happens on the Declaration site alias. 
 on the user site, we can rename properly even on the last index. 
 
 PS : your comments are really descriptive. helped a lot while getting context about the PR. Thanks :)

vidit-od (changes requested):
A few more edgecases to look into; 

major one : when user tries to rename on last index. ( following the observation of previous comment )
user site also behaves weirdly when rename is triggered on last index. 

example 1 : 
renames what ever is after the alias
Recording 2026-03-29 031820

example 2 :
the issue pointed in example 1 cascades in all of the occurrences ( should ideally be fixed when you fix for example 1, but worth testing ig :D )
Recording 2026-03-29 032142

example 3 : 
an unlikely scenario, but should be taken into considerations maybe : 
renaming of alias that are exported 

Recording 2026-03-29 032722

izuzu-izuzu:
@vidit-od Thank you so much for reviewing this PR and pointing out the edge cases!

You’re right that I initially disallowed renaming an alias when the cursor is at the end of the alias. This was on purpose, because GHC source spans are supposed to be inclusive-start, exclusive-end (especially relevant in expressions like `x+1` without spaces).

After review, I relaxed this logic, because an import alias is never immediately followed by another renameable thing (there’s always either a space or a dot in between).

Also, I implemented renaming qualified names in the module export list, as you suggested. This was fortunately easy: I just needed to collect `RdrName` values from the `hsmodExports` list in addition to `hsmodDecls`.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,331 @@
...
+{-| Logic for renaming qualified import aliases.
+
+For example:
+
+> -- Before: ---------------------------
+> import qualified Data.List as L
+> bar = L.take
+> -- After: ----------------------------
+> import qualified Data.List as List
+> bar = List.take
```
vidit-od: P2

we can maybe get rid of this example ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,331 @@
...
+            guard (rangeContainsPositionInclusive qualifiedNameRange pos)
+            let qualifierLength = fromIntegral (moduleNameLength qualifier)
+                qualifierStart = qualifiedNameRange ^. VFS.start
+                qualifierRange = qualifiedNameRange
+                    & VFS.end .~ (qualifierStart & VFS.character +~ qualifierLength)
+            guard (rangeContainsPositionInclusive qualifierRange pos)
+            [(qualifierRange, qualifier)]
+    in case qualifiersAtPos of
+        [] -> Nothing
+        (rangeAtPos, qualifierAtPos) : _ -> Just $ (,) rangeAtPos $ do
```
vidit-od, line 217: P0

qualifiersAtPos calculates a full list, then we flush out everything else except head. 
seems like a wasteful compute no ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,331 @@
...
+            let allAliases = mapMaybe (fmap unLoc . ideclAs . unLoc) imports
+            importDecl <- map unLoc imports
+            Just locatedAlias <- [ideclAs importDecl]
+            let aliasName = unLoc locatedAlias
+            guard (aliasName == qualifierAtPos)
+            RealSrcSpan aliasDeclSpan _ <- [getLoc locatedAlias]
+            let aliasModuleName = unLoc (ideclName importDecl)
+                aliasDeclRange = realSrcSpanToCodePointRange aliasDeclSpan
+                aliasIsShared = length (filter (== aliasName) allAliases) > 1
```
vidit-od, line 226: P2 (minor performance maybe )

we just want to know > 1 right ? 
instead of : find all then count > 1 
we can : find 2nd and stop ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,331 @@
...
+findAliasUseAtPos ::
+    VFS.CodePointPosition ->
+    Maybe (XRec GhcPs [LIE GhcPs]) ->
+    [LImportDecl GhcPs] ->
+    [LHsDecl GhcPs] ->
+    Maybe (VFS.CodePointRange, [ImportAlias])
+findAliasUseAtPos pos exports imports hsDecls =
+    let qualifiersAtPos = do
+            locatedRdrName :: XRec GhcPs RdrName <-
+                listify (const True) exports ++ listify (const True) hsDecls
```
vidit-od: P0 maybe 

from what i can understand, listify is a full traversal and slow 
any alternative to that ? 
@fendor any recommendations ? 

at the very least we should not compute the full list because we flush out everything else other than head later.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,331 @@
...
+aliasUseSiteRanges ::
+    ImportAlias ->
+    Maybe (XRec GhcPs [LIE GhcPs]) ->
+    [LHsDecl GhcPs] ->
+    [VFS.CodePointRange]
+aliasUseSiteRanges importAlias exports hsDecls = nubOrd $ do
+    let ImportAlias{aliasName} = importAlias
+        aliasLength = fromIntegral (moduleNameLength aliasName)
+    locatedRdrName :: XRec GhcPs RdrName <-
+        listify (const True) exports ++ listify (const True) hsDecls
```
vidit-od: P2 

This pattern is duplicated 3 times. 
move this to util section maybe ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,331 @@
...
+            Just (_, []) -> pure Nothing
+            Just (range, [alias]) -> pure $ toLSPRange (range, alias)
+            Just (range, candidates) -> do
+                tcModule <- runActionE "rename.resolveAlias" state $ useE TypeCheck nfp
+                namesAtPos <- getNamesAtPosFn state nfp lspPos
+                case disambiguateAliasUse tcModule namesAtPos candidates of
+                    [] -> pure Nothing
+                    [alias] -> pure $ toLSPRange (range, alias)
+                    aliases -> throwError $ PluginInvalidParams $
+                        ambiguousAliasErrorMessage aliases
```
vidit-od: p1

here we make sure that aliases have at least 2 items in it. 
then later in `ambiguousAliasErrorMessage` we return "" for < 2 alias case. seems weird 

can we do the entire pattern matching in  `ambiguousAliasErrorMessage`  ? 
i.e 
```Haskell
ambiguousAliasErrorMessage :: [ImportAlias] -> T.Text
ambiguousAliasErrorMessage [] =  ...
ambiguousAliasErrorMessage [_] = ...
ambiguousAliasErrorMessage aliases@(alias1 : alias2 : _) = ...
```

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -226,6 +268,7 @@ refsAtName state nfp name = do
         Nothing -> pure []
         Just mod -> liftIO $ mapMaybe rowToLoc <$> withHieDb (\hieDb ->
             -- See Note [Generated references]
+            -- REVIEW: Is this filter supposed to keep or remove generated references?
             filter (\(refRow HieDb.:. _) -> refIsGenerated refRow) <$>
```
vidit-od, line 272: P0 

from what the comments state, it should ideally remove generated reference right ? 
i would recommend adding a NOT and replicate the expected behavior of comments

plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -93,6 +122,26 @@ renameTests = testGroup "Identifier"
...
+    , goldenWithRename "Import alias declaration" "ImportAlias" $ \doc ->
+        rename doc (Position 1 14) "G"
+    , goldenWithRename "Import alias at use site" "ImportAlias" $ \doc ->
+        rename doc (Position 5 10) "G"
+    , goldenWithRename "Import alias declaration (shared by unrelated imports)" "ImportAliasShared" $ \doc ->
+        rename doc (Position 3 31) "Maybe"
+    , goldenWithRename "Import alias at use site (shared by unrelated imports)" "ImportAliasShared" $ \doc ->
+        rename doc (Position 6 6) "Maybe"
+    , goldenWithRename "Import alias declaration (with re-exports)" "ImportAliasReexport" $ \doc -> do
+        rename doc (Position 1 18) "Reexport"
```
vidit-od, line 152: P1 

add 1 or 2 more regression tests for checking renaming alias on last position ? 
or maybe edit previous ones to last index ?

vidit-od (changes requested):
Recommended a few more nit picks 

P0, P1s are atleast important. can ignore P2 if you want 

Other than below listed changes , one more P1 change would be : 

input Validation on Rename. Can we check if the Alias starts with a Capital letter before renaming ?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
izuzu-izuzu: `qualifiersAtPos` is a lazy list, so I’m guessing only the first element (or the first few) will be produced and pattern-matched.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
izuzu-izuzu: Good point, an explicit `case` would be better here.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
izuzu-izuzu: `listify (const True) exports ++ listify (const True) hsDecls` is a lazy list, so elements are computed on demand. The moment we find one element for `qualifiersAtPos`, `listify` stops computing any more elements (I think).

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
izuzu-izuzu: Sure, I can do that.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
izuzu-izuzu: I can inline `ambiguousAliasErrorMessage`.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
izuzu-izuzu: This is an interesting bit.

I did some basic testing and found out (not conclusively) that quite often, for a given location, the HIE DB contains two identical reference entries: one marked as originating from real source code, and the other marked as autogenerated by GHC. (This means filtering with `refIsGenerated refRow` and `not (refIsGenerated refRow)` returns the same references.)

I haven’t managed to write tests to prove that the current behavior is incorrect, so I chose not to address this part in this PR.

If I ever get to the bottom of this I’ll raise a new issue and PR.

plugins/hls-rename-plugin/test/Main.hs
izuzu-izuzu: The test case on line 128 actually has the cursor at the end of the alias:

```haskell
    , goldenWithRename "Import alias at use site" "ImportAlias" $ \doc ->
        rename doc (Position 5 10) "G"
```

I can edit the test name to make it clearer.

Also, the new `prepareRename` test cases do test different cursor positions as well.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
vidit-od: Can you maybe add a Trace and confirm this ? 
if hypothesis is correct ,  we can move forward with this;

plugins/hls-rename-plugin/test/Main.hs
```diff
@@ -130,6 +130,24 @@ renameTests = testGroup "Identifier"
...
+    -- The console shows 2 `listify:` traces.
+    , goldenWithRename "Import alias in export list (proving 'listify' is lazy)" "ImportAliasLazyListify" $ \doc ->
+        rename doc (Position 0 39) "Maybe"
+
+    -- REVIEW: `listify (const True) exports ++ listify (const True) hsDecls`
+    -- produces elements until it produces one whose alias is at the cursor.
+    -- This element is then consumed in the pattern match. `listify` then stops
+    -- producing.
+    -- The console shows traces for all `RdrName`s except for the `M.fromMaybe`
+    -- on the last line.
```
vidit-od: ok seems good, 
we can move forward with this then. 

The trace statements will give build errors so we should remove them now. 
Rest everything Looks fine

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
izuzu-izuzu: Sure!

izuzu-izuzu:
@vidit-od I resolved all comments, including the one about validating the new alias.

For validation I borrowed GHC’s `parseIdentifier` function so that we permit whatever GHC permits, like `𝔽` (Mathematical Double-Struck Capital F).

I also removed the traces and tests for `listify`. Would appreciate your review!

vidit-od (approved):
The error message can be improved for invalid inputs, Rest Everything looks good 

Image

izuzu-izuzu:
@vidit-od How did you produce the error messages?

I don’t think they come from my PR.
1. The first message can’t be found in the HLS codebase.
2. The second one seems to be related to the “Reactor” (I don’t know what it is), and not from `PluginInternalError`.

vidit-od:
> @vidit-od How did you produce the error messages?
> 
> I don’t think they come from my PR.
> 1. The first message can’t be found in the HLS codebase.
> 2. The second one seems to be related to the “Reactor” (I don’t know what it is), and not from `PluginInternalError`.

I just entered an invalid input ( small case first letter ) to get these errors

izuzu-izuzu:
@vidit-od What’s your environment? I built my branch with GHC 9.12.2, and am using Doom Emacs with `lsp-mode`, and I’m getting an informative error message, like this:

```text
lsp-request: rename: Invalid Params: ‘map’ is an invalid import alias.
```

We’re not in a rush to merge this PR, so I can spend some time trying to reproduce your error message and fix it.

izuzu-izuzu:
@fendor What should I do with this PR? I can’t replicate vidit-od’s error message. Should we proceed anyway?

vidit-od:
Apologies for not responding in time ! 
yes i also recommend moving forward in case you dont find any way to replicate it ! 

i also could not find anywhere in the codebase which could be responsible for the error message and hence raised the comment ! most likely some issues in my environments :)

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -88,28 +90,71 @@ descriptor recorder pluginId = mkExactprintPluginDescriptor exactPrintRecorder $
...
-    -- to mean "there is no Name at given position" (in which case
-    -- `spansWithNamesUnderCursor` would be empty).
-    --
-    -- In particular it allows some cases through (e.g. cross-module renames),
-    -- so that the full rename handler can give more informative error about them.
-    pure $ case spansWithNamesUnderCursor of
-        [] -> InR Null
-        srcSpan : _ -> InL $ PrepareRenameResult $ InL (realSrcSpanToRange srcSpan)
+    codePointPos <- getCodePointPosition state nfp lspPos
+    maybeParsed <- ImportAlias.getParsedModuleStale state nfp
```
crtschin, line 96: The previous version used `handleGetHieAst`, which has a comment saying that renames should explicitly use non-stale versions to ensure consistency. Doesn't this have that same problem? This is the `prepare` version of the request so stale results should be fine I think. How about using `useWithStaleFast*` instead of just `stale`, as this is a interactive action from what I can tell?

----------

On a secondary note, isn't the existing `HieAst` enough to get the import module aliases?

Looking at the `HieAst` of a `import Data.Map.Strict as Map` in [this file](https://github.com/haskell/HieDb/blob/master/src/HieDb/Dump.hs#L7), gives me the following nodes.

<details><summary>hiedb dump of HieDb/Dump.hs</summary>

```
  Node@src/HieDb/Dump.hs:7:1-39: Source: From source
                                 {(annotations: {(ImportDecl, ImportDecl)}),  (types: []),
                                  (identifier info: {})}

    Node@src/HieDb/Dump.hs:7:18-32: Source: From source
                                    {(annotations: {}),  (types: []),
                                     (identifier info: {(module Data.Map.Strict,  Details:  Nothing {import})})}

    Node@src/HieDb/Dump.hs:7:37-39: Source: From source
                                    {(annotations: {}),  (types: []),
                                     (identifier info: {(module Map,  Details:  Nothing {import as})})}
```
</details>

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -88,28 +90,71 @@ descriptor recorder pluginId = mkExactprintPluginDescriptor exactPrintRecorder $
...
+    case maybeParsed of
+        Nothing -> throwError $ PluginInternalError
+            "The module hasn’t yet been parsed. Please wait for indexing to complete and try again."
+        Just parsed -> do
+            let hsModule = unLoc $ pm_parsed_source parsed
+                exports = hsmodExports hsModule
+                imports = hsmodImports hsModule
+                hsDecls = hsmodDecls hsModule
+            maybeAlias <- ImportAlias.resolveAliasAtPos
+                getNamesAtPos state nfp lspPos codePointPos exports imports hsDecls
```
crtschin, line 106: It's a bit unobvious what's happening here, so we first check whether the cursor is located at the position of an import alias, if not, we traverse the `HieAst` if token describes a `Name` to determine whether we can rename the symbol? Could you add a comment that documents this?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
```diff
@@ -88,28 +90,71 @@ descriptor recorder pluginId = mkExactprintPluginDescriptor exactPrintRecorder $
...
+                    pure $ case spansWithNamesUnderCursor of
+                        [] -> InR Null
+                        srcSpan : _ -> InL $ PrepareRenameResult $ InL (realSrcSpanToRange srcSpan)
 
 renameProvider :: PluginMethodHandler IdeState Method_TextDocumentRename
-renameProvider state pluginId (RenameParams _prog (TextDocumentIdentifier uri) pos newNameText) = do
+renameProvider state pluginId (RenameParams _prog (TextDocumentIdentifier uri) lspPos newNameText) = do
     nfp <- getNormalizedFilePathE uri
+    codePointPos <- getCodePointPosition state nfp lspPos
+    maybeParsed <- ImportAlias.getParsedModuleStale state nfp
```
crtschin, line 132: There's an argument to be made that the `prepareRename` could be done stale, but the actual `rename` shouldn't be. I can definitely see the comment in `handleGetHieAst` being relevant then.

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,378 @@
...
+        runAction "rename.getParsedModuleStale" state
+            (useWithStale GetParsedModule nfp)
+
+-- | Return the 'ImportAlias' and corresponding text range at the cursor. The
+-- cursor may be on the alias token in an import declaration or on a qualifier
+-- at a use site. If multiple imports share the same alias, falls back to the
+-- typechecked module's 'GlobalRdrEnv' to disambiguate.
+-- Returns @Nothing@ if the cursor is not on an import alias.
+-- HACK: The first argument is `Rename.getNamesAtPos`, parameterized to avoid a
+-- circular dependency.
```
crtschin, line 90: Isn't the fix for this comment, moving `getNamesAtPos` to this module?

plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ImportAlias.hs
```diff
@@ -0,0 +1,378 @@
...
+        throwError (PluginInvalidParams (quote newNameText <> " is an invalid import alias."))
+    let ImportAlias{aliasDeclRange, aliasIsShared} = importAlias
+    virtualFile <- runActionE "rename.getVirtualFile" state
+        $ handleMaybeM (PluginInternalError
+            ("Virtual file not found: " <> T.pack (show nfp)))
+        $ getVirtualFile nfp
+    useSiteRanges <-
+        if aliasIsShared
+        then do
+            tcModule <- runActionE "rename.sharedAliasRanges" state $ useE TypeCheck nfp
```
crtschin, line 156: I think it should be possible to avoid typechecking by using a similar mechanism to how completions get calculated to union symbols from multiple modules, to now disambiguate which module a symbol comes from. The relevant bit being how `qualCompls` get calculated in `CachedCompletions`.

crtschin (commented):
Awesome work! Thanks for the PR, I left behind some comments and ideas.

fendor:
@izuzu-izuzu Do you plan to keep working on this? Otherwise, perhaps someone during ZuriHac can pick it up and push it over the finish line

PR #4881, opened 2026-04-02
Fix deleting unused bindings with Haddock docs

### LINKED ISSUE #4876: Delete unused binding code action doesn't delete haddock
The delete unused binding code action seems to leave the Haddock attached to the binding it was meant to delete.

### Your environment

Which OS do you use? NixOs

Which version of GHC do you use and how did you install it? 9.10.2

How is your project built (alternative: link to the project)? cabal-install

Which LSP client (editor/plugin) do you use? emacs+lsp-mode

Which version of HLS do you use and how did you install it? 2.11.0.0

Have you configured HLS in any way (especially: a `hie.yaml` file)? not particularly

### Steps to reproduce

If I have a definition like:

```
module Foo () where
{- | My haddock goes here 
-}
foo :: Int
foo = ...
```
and `foo` is unused, then I get a code action to delete `foo`.

If I trigger this code action then the definition of `foo` is deleted by the haddock comment is left, even though it is attached to `foo`.

### Expected behaviour

I would expect the Haddock docstring to also get deleted.

### Actual behaviour

The Haddock docstring is not deleted.

agam263:
Closes #4876
## What changed

This fixes the refactor plugin's `Delete ‘…’` code action so it removes attached Haddock documentation when deleting an unused binding.

The change covers:
- line Haddock comments like `-- |` and `-- ^`
- block Haddock comments like `{-|` and `{-^`

It also adds regression tests for both forms.

## Why

When an unused top-level binding had Haddock immediately above it, the delete-unused-binding quick fix removed the binding but left the documentation behind. That produced orphaned docs and invalid or misleading source after applying the code action.

## Root cause

The delete action built text edits from declaration and signature spans, but those spans did not account for attached Haddock comments. As a result, the edit range started too late and excluded the documentation block.

There was also a follow-up issue with overlapping deletion ranges once the Haddock span was included. That is now handled by sorting and merging the related ranges before producing text edits.

## Implementation details

- Extend deletion ranges upward to include attached Haddock blocks
- Preserve the existing behavior for ordinary whitespace/newline trimming
- Merge overlapping ranges before constructing the final workspace edits
- Add regression tests for:
  - line Haddock above an unused top-level binding
  - block Haddock above an unused top-level binding

## Validation

Ran locally with:

```bash
cabal test hls-refactor-plugin-tests --with-compiler=$HOME/.ghcup/bin/ghc-9.10.3

### DIFF
--- a/plugins/hls-refactor-plugin/test/Main.hs
+++ b/plugins/hls-refactor-plugin/test/Main.hs
@@ -2531,6 +2531,43 @@ deleteUnusedDefinitionTests = testGroup "delete unused definition action"
       , ""
       , "some = ()"
       ]
+  , testSession "delete unused top level binding with Haddock comment" $
+    testFor
+      [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+      , "module A (some) where"
+      , ""
+      , "-- | docs for f"
+      , "f :: Int"
+      , "f = 1"
+      , ""
+      , "some = ()"
+      ]
+      (5, 0)
+      "Delete ‘f’"
+      [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+      , "module A (some) where"
+      , ""
+      , "some = ()"
+      ]
+  , testSession "delete unused top level binding with block Haddock comment" $
+    testFor
+      [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+      , "module A (some) where"
+      , ""
+      , "{-| docs for f"
+      , "-}"
+      , "f :: Int"
+      , "f = 1"
+      , ""
+      , "some = ()"
+      ]
+      (6, 0)
+      "Delete ‘f’"
+      [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+      , "module A (some) where"
+      , ""
+      , "some = ()"
+      ]
   , testSession "delete unused binding in where clause" $
     testFor
       [ "{-# OPTIONS_GHC -Wunused-binds #-}"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugins/hls-refactor-plugin/test/Main.hs [hunk 1]
```diff
@@ -2531,6 +2531,43 @@ deleteUnusedDefinitionTests = testGroup "delete unused definition action"
...
+      , ""
+      , "some = ()"
+      ]
+      (6, 0)
+      "Delete ‘f’"
+      [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+      , "module A (some) where"
+      , ""
+      , "some = ()"
+      ]
```
vidit-od, line 2570: https://github.com/user-attachments/assets/2ed81d34-8785-49c0-a395-7238e8bb8da5

It seems weird, i cannot replicate the Behavior
But the tests do Pass.

fendor:
Hi, thank you for your contribution!

It looks to me like you are manually computing the haddock comment range.
Is there no way to compute the range based on the AST where we know the range of the haddock comment attached to the AST node we are removing?

If we do it stringly typed, I feel like there are corner cases we might miss. 

For example:

```haskell
-- | Comment
foo = ()

some = ()
```

If we delete `some`, how do we determine whether `-- | Comment` belongs to `foo` or `some`?

---

What about 

```haskell
{-
-- | Commented out
   -}
some = ()
```

Would that delete `Commented out`? It shouldn't!

---

What about 

```haskell
some = ()
-- ^ Comment
```

Do we perform arbitrary look ahead to find trailing haddock comments?

---

Similar and complicated:

```haskell
some = ()
-- Comment
-- ^ Haddock comment
```

If we remove `some`, should `Comment` be deleted? It is not a haddock comment...

--- 


I imagine there are many more corner cases, and I think we need to look into the AST and figure out the correct haddock comment to compute the Range to delete.

fendor (changes requested):
Just to mark to myself there are open questions

fendor:
@agam263 Are you still planning on working on this?

fendor:
Closing in favour of #4965

Issue #4884, opened 2026-04-04
Flakiness on running ghcide-tests due to ghcide + lsp-tests


crtschin:
I was trying to run tests locally, but I'm confronted with flakiness. Specifically, running ghcide-tests fails quite often on lsp-tests attempting to read from the handle that's been closed by the spawned ghcide. For reference this is on 177ca8e4e5a73d7d36cd8bd8e9c5bcd2f3cc6d61, running the tests that I added in #4862. Example of the same failures on CI: [here](https://github.com/haskell/haskell-language-server/actions/runs/23949973106/job/69854903234#step:7:668), [here](https://github.com/haskell/haskell-language-server/actions/runs/23949973106/job/69854903193#step:20:33).

```bash
$ cabal run ghcide-tests -- -p '/constructor hover (#2904)/'
```

I'm running on 12 cores, which defaults to 6 capabilities in ghcide. If I change it to always spawn ghcide with a single capability instead, the flakiness disappears.

<details><summary>Diff for forcing a single capability</summary>

```diff
diff --git a/hls-test-utils/src/Test/Hls.hs b/hls-test-utils/src/Test/Hls.hs
index 3c0180bef..a1172abf7 100644
--- a/hls-test-utils/src/Test/Hls.hs
+++ b/hls-test-utils/src/Test/Hls.hs
@@ -895,6 +895,7 @@ runSessionWithTestConfig TestConfig{..} session =
                 , argsDefaultHlsConfig = testLspConfig
                 , argsProjectRoot = prjRoot
                 , argsDisableKick = testDisableKick
+                , argsThreads = Just 1
                 }
 
 -- | Wait for the next progress begin step
```
</details>

This is probably caused by ghcide setting the number of capabilities for the entire RTS, which bleeds into lsp-tests which isn't compatible with that(?).

<details><summary>Example failing run using the command above</summary>

```
ghcide
  constructor hover (#2904)
    Constructors.hs
      A: OK (0.11s)
      B: OK (0.10s)
      C: OK (0.08s)
      D: OK (0.09s)
      E: OK (0.09s)
      F: OK (0.09s)
    ConstructorsLinear.hs
      A: OK (0.09s)
      B: OK (0.09s)
      C: OK (0.08s)
      D: OK (0.08s)
      E: FAIL
        Exception: Language server unexpectedly terminated
        HasCallStack backtrace:
[...]

PR #4886, opened 2026-04-05
Cancel outstanding threads prior to closing sqlite connection

crtschin:
From the investigation in https://github.com/haskell/haskell-language-server/issues/4884#issuecomment-4189264091.

I was repeatedly running the `ghcide-tests` to debug flakiness, and next to the EOF closed handle reads, I also encountered segfaults. Investigating the latter further via the coredumps showed some sqlite functions, suggesting use-after-free of the sqlite handle.

The problem here is that the connections are spawned and given to numerous async threads, either to handle the request, or via the shake session. These async threads outlive the workerthread lifetimes, leading to racey use-after-frees. Solve this by both keeping track of live threads and cancelling those, and additionally issuing a shake session shutdown prior to leaving. The segfaults disappear with these changes.

Edit: Removed the cancellation of the threads that were handling the request, and only included the shake shutdown prior to leaving scope. This was enough to avoid segfaults in my case (running a small portion of the testsuite 100x repeatedly), and stops the tests on CI from failing.

### DIFF
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -40,7 +40,8 @@ import           Control.Concurrent.Extra              (newBarrier,
                                                         signalBarrier,
                                                         waitBarrier)
 import           Control.Monad.IO.Unlift               (MonadUnliftIO)
-import           Control.Monad.Trans.Cont              (evalContT)
+import           Control.Monad.Trans.Cont              (ContT (..), evalContT)
+import           Data.Foldable                         (traverse_)
 import           Development.IDE.Core.IdeConfiguration
 import           Development.IDE.Core.Service          (shutdown)
 import           Development.IDE.Core.Shake            hiding (Log)
@@ -54,6 +55,7 @@ import           Language.LSP.Server                   (LanguageContextEnv,
                                                         LspServerLog,
                                                         type (<~>))
 import           System.Timeout                        (timeout)
+
 data Log
   = LogRegisteringIdeConfig !IdeConfiguration
   | LogReactorThreadException !SomeException
@@ -291,9 +293,8 @@ handleInit lifecycleCtx env (TRequestMessage _ _ m params) = otTracedHandler "In
     ideMVar <- newEmptyMVar
 
     let
-      handleServerExceptionOrShutDown me = do
+      loggedTeardown me = do
         -- shutdown shake
-        tryReadMVar ideMVar >>= mapM_ shutdown
         case me of
           Left e -> do
             lifetimeConfirm "due to exception in reactor thread"
@@ -326,8 +327,12 @@ handleInit lifecycleCtx env (TRequestMessage _ _ m params) = otTracedHandler "In
                 $ \(e :: SomeException) -> do
                     exceptionInHandler e
                     k $ TResponseError (InR ErrorCodes_InternalError) (T.pack $ show e) Nothing
-    _ <- flip forkFinally handleServerExceptionOrShutDown $ do
-      runWithWorkerThreads (cmapWithPrio LogSession recorder) dbLoc $ \withHieDb' threadQueue' -> do
+    _ <- flip forkFinally loggedTeardown $ do
+      -- Need to be careful about when the shutdown occurs, it needs to be shut
+      -- down after the session loader and restarting threads, and before the
+      -- hiedb connections are closed.
+      let shutdownSession = tryReadMVar ideMVar >>= traverse_ shutdown
+      runWithWorkerThreads (cmapWithPrio LogSession recorder) dbLoc shutdownSession $ \withHieDb' threadQueue' -> do
         ide <- ctxGetIdeState lifecycleCtx env root withHieDb' threadQueue'
         putMVar ideMVar ide
         -- Keep this after putMVar ideMVar ide; otherwise shutdown during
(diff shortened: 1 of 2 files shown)


ghcide/src/Development/IDE/LSP/LanguageServer.hs
fendor: `handleServerExceptionOrShutDown` also shutdowns, is that redundant?

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -36,6 +36,7 @@ import           UnliftIO.Directory
 import           UnliftIO.Exception
 
 import qualified Colog.Core                            as Colog
+import           Control.Concurrent.Async              (cancelMany)
```
fendor: Unused import

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -332,7 +334,7 @@ handleInit lifecycleCtx env (TRequestMessage _ _ m params) = otTracedHandler "In
         putMVar ideMVar ide
         -- Keep this after putMVar ideMVar ide; otherwise shutdown during
         -- initialization could leave handleInit blocked indefinitely on readMVar.
-        untilReactorStopSignal $ forever $ do
+        untilReactorStopSignal $ flip finally (shutdown ide) $ forever $ do
```
fendor: Perhaps this cancellation should happen in `runWithWorkerThreads`? 

Here it looks a bit ad-hoc and non-trivial to understand the relation with the `hiedb` thread.

fendor (dismissed):
Amazing detective work! CI was green at the first attempt, if this fixes the segmentation faults we have been observing on windows, this would be a huge win!

I think the idea is correct, but the exact location for shutdown feels not quite correct.
Perhaps we need to cancel the worker threads in `runWithWorkerThreads`? Then a comment could state the invariant to avoid use-after-free.

soulomoon:
Great detective work @crtschin ! 

Some Threads(ShakeRestart and Sessionloader) in `runWithWorkerThreads`(3 of them in total now) are the controlling threads over the shake session's threads. And only the DB thread would be used in shake session's threads. 

If we cancel the shake session before leaving the scope of the body of `runWithWorkerThreads`, shakeRestart might still wake up shake session's threads. 

Perhaps a better idea would be, shutdown the shake session right before shutting down the DB threads but after the shutting down of ShakeRestart thread and Sessionloader thread inside the `runWithWorkerThreads` as @fendor suggested.

PS. I was wondering why the problem only happens for windows.
and if we could be more informed to the use-after-free problem(Such as replacing the handle with something like error "xxx is already free").

-------------------------------

I was wrong, Sessionloader would be called inside shake session's threads(Rule GhcSession). and Sessionloader would call ShakeRestart hence it creates cycling depedencies. 0 0 It is tricky. Perhaps, do so in three folds, first stop Sessionloader's worker and then ShakeRestart's woker, then shutdown the shake session, finally close the queues of Sessionloader and ShakeRestart.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
crtschin: Ah missed it, nicely spotted. Somewhat yes. They do the same thing in shutting down the connection, though this one is definitely too late in stopping the users of the connection. I'll take this bit into account as well.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
crtschin: Good idea

crtschin:
> I was wrong, Sessionloader would be called inside shake session's threads(Rule GhcSession). and Sessionloader would call ShakeRestart hence it creates cycling depedencies.

That's fine no? Following your and @fendor's suggestion, I'm thinking shutting down the shake session at the following spot:

```diff
diff --git a/ghcide/src/Development/IDE/LSP/LanguageServer.hs b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
index 72720e302..40d34d2ca 100644
--- a/ghcide/src/Development/IDE/LSP/LanguageServer.hs
+++ b/ghcide/src/Development/IDE/LSP/LanguageServer.hs
@@ -354,6 +354,7 @@ handleInit lifecycleCtx env (TRequestMessage _ _ m params) = otTracedHandler "In
 runWithWorkerThreads :: Recorder (WithPriority Session.Log) -> FilePath -> (WithHieDb -> ThreadQueue -> IO ()) -> IO ()
 runWithWorkerThreads recorder dbLoc f = evalContT $ do
             (WithHieDbShield hiedb, threadQueue) <- runWithDb recorder dbLoc
+            -- Shutdown session here, note that shutdown happens bottom->up
             sessionRestartTQueue <- withWorkerQueueSimple (cmapWithPrio Session.LogSessionWorkerThread recorder) "RestartTQueue"
             sessionLoaderTQueue <- withWorkerQueueSimple (cmapWithPrio Session.LogSessionWorkerThread recorder) "SessionLoaderTQueue"
             liftIO $ f hiedb (ThreadQueue threadQueue sessionRestartTQueue sessionLoaderTQueue)
```

So this'd be after the restart and loader threads have closed down. If these threads are closed down, then they shouldn't be accessing the connection anymore. Restarting also wouldn't occur because the thread that does it is now finished.

crtschin:
> PS. I was wondering why the problem only happens for windows.

Forgot to mention, I'm running into the segfaults locally on my linux machine when running the following command repeatedly: `cabal run ghcide-tests -- -p '/constructor hover (#2904)/'`. 

I assume this occurs more often with the above, because I'm running with 6 capabilities and those tests specifically are very quick leaving a big potential gap between everything being loaded in and the threads being torn down.

soulomoon (commented):
> That's fine no? Following your and @fendor's suggestion, I'm thinking shutting down the shake session at the following spot:

yes, put it there look good to me.

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -291,9 +293,8 @@ handleInit lifecycleCtx env (TRequestMessage _ _ m params) = otTracedHandler "In
     ideMVar <- newEmptyMVar
 
     let
-      handleServerExceptionOrShutDown me = do
+      loggedTeardown me = do
         -- shutdown shake
-        tryReadMVar ideMVar >>= mapM_ shutdown
```
soulomoon, line 296: We might still want it if the inclosing "shutdownSession" did not happens ?

soulomoon: oh right, it will always happens if we did `putMVar ideMVar`

ghcide/src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -349,12 +354,16 @@ handleInit lifecycleCtx env (TRequestMessage _ _ m params) = otTracedHandler "In
...
-            sessionRestartTQueue <- withWorkerQueueSimple (cmapWithPrio Session.LogSessionWorkerThread recorder) "RestartTQueue"
-            sessionLoaderTQueue <- withWorkerQueueSimple (cmapWithPrio Session.LogSessionWorkerThread recorder) "SessionLoaderTQueue"
-            liftIO $ f hiedb (ThreadQueue threadQueue sessionRestartTQueue sessionLoaderTQueue)
+runWithWorkerThreads :: Recorder (WithPriority Session.Log) -> FilePath -> IO () -> (WithHieDb -> ThreadQueue -> IO ()) -> IO ()
+runWithWorkerThreads recorder dbLoc shutdownSession f = evalContT $ do
+  (WithHieDbShield hiedb, threadQueue) <- runWithDb recorder dbLoc
+  -- The shake session needs to be shut down prior to the hiedb connections
+  -- being cleaned up, otherwise shake could be referencing dead connections.
+  -- This is passed in via the callsites.
+  void $ ContT $ \action -> action () `finally` shutdownSession
```
soulomoon: We do not need the extra void here ?

fendor (approved):
LGTM, thank you for fixing this incredible strain on our CI!
@soulomoon could you merge this PR once you are happy with it?

soulomoon (approved):
Good job, it looks good to me too. @crtschin would you like to squash and merge.

crtschin:
Squashed my changes, but I can't merge. I don't have write access.

crtschin:
Rebased, rerunning CI before enabling automerge again

PR #4888, opened 2026-04-07
Sorted explicit import

vidit-od:
Closes #4875 

We follow the following procedure for sorting the imports and maintaining import format. 

1. scan the old import list. If it is a single line list we preserve the behaviour after sorting.  

explicite-singleline

2. if the old import list is multilined. we force a one-line one import item format. 
explicite-multiline

Note : a few of the regression tests have been changed to match this behavior. a few more may need to change. In case the above 2 formats are finalized we can change more if needed.

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction/ExactPrint.hs
@@ -18,9 +18,11 @@ module Development.IDE.Plugin.CodeAction.ExactPrint (
 
 import           Control.Monad
 import           Control.Monad.Trans
+import           Control.Applicative                     ((<|>))
 import           Data.Char                              (isAlphaNum)
 import           Data.Data                              (Data)
 import           Data.Generics                          (listify)
+import           Data.List                              (sortOn)
 import qualified Data.Text                              as T
 import           Development.IDE.GHC.Compat             hiding (Annotation)
 import           Development.IDE.GHC.Error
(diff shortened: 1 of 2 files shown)


fendor:
Just to summarise, we discussed this feature, and I think it is quite tricky to get this right or in a state that is better than the status quo. 

For example, should we sort unsorted explicit imports, should we sort case sensitvely, what about `type` and `pattern` namespaces or operators... 

Personally, I think it is better to leave this to your formatter of choice, instead of complicating HLS.

vidit-od:
Just to add to what is said above, and for the sake of Documentation : 
Following things were explored while working on this PR 

1. Slot Based Sorting : 
- add item at the end of unsorted list 
- traverse through the entire list and note positions of each item
- sort the list 
- use the above positions to then add elements in respective positions. 
- keeps relatively same formatting of the list and sorts it  

2. The approach in this PR 
- have 2 valid formats 
- either all imports are sorted and in 1 line 
- or they are a one element in a new line

3. stylish haskell 
- maybe add this as a stylish haskell thing 
- sorts and places everything properly after stylish haskell is triggered
- would create too much chaos on already existing code adding need of too much extra unnecessary formatting code in entire repos.

juhp:
Ah formatters will sort the imported identifiers?  I see that is interesting

(I guess I wasn't expecting perfection and I get it is complicated, just thought it would be better to try to improve the current behavior in any way)

fendor:
> Ah formatters will sort the imported identifiers?

Only if you invoke the formatter :)

> just thought it would be better to try to improve the current behavior in any way

I agree, but maintenance is also important, and if there isn't a clear "better", I am rather careful with features that can easily expand in complexity.

PR #4891, opened 2026-04-10
WIP: Check flakiness against exceptions in `lsp-test` teardown 

crtschin:
See this PR: https://github.com/haskell/lsp/pull/637, the analysis in that PR still makes sense to me.

I closed it because I couldn't repro the flakiness locally anymore, but there's still quite a bit of flakiness on CI here. This PR uses the commit from that PR for lsp-test, to check if it resolves anything.


crtschin:
Going to create a new PR, looks like I won't have to change anything in `lsp-test`.

PR #4896, opened 2026-04-13
Allow skipping over non-required progress messages

crtschin:
Related https://github.com/haskell/haskell-language-server/issues/4714.

I think there may be more of these `progress`-related assertions, I haven't gone through it with a finetooth comb. These were the ones that were most prominent to me.

What happens at each of these is that some messages don't occur at all in the message stream, the main ones being `Indexing` and `Processing`. I presume that we're just too fast, where ghcide finishes indexing by the time it gets a token to communicate progress. 

Instead of failing the test when we don't see these messages that may not occur, modify the assertion so it better matches what we expect for progress flows.
- If we see a begin, we _must_ see an end.
- If we see a begin, it _must_ contain a token from a preceding create.
- If we see a create, its token _may_ be used by subsequent progressions.

Additoinally, I made the error a bit more informative so it shows which messages it hung on. After this I repeated the tests in this file, and couldn't reproduce the failures anymore.

### DIFF
--- a/test/functional/Progress.hs
+++ b/test/functional/Progress.hs
@@ -13,6 +13,7 @@ import           Data.Functor                       (void)
 import           Data.List                          (delete)
 import           Data.Maybe                         (fromJust)
 import           Data.Text                          (Text, pack)
+import           GHC.Stack                          (withFrozenCallStack)
 import           Ide.Types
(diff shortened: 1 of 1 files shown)


fendor (approved):
LGTM, thank you very much, please merge when you are ready.

Issue #4084, opened 2024-02-18
HLS statically advertises support for functionality that is disabled by config


michaelpj:
For example: if the semantic tokens plugin is disabled, we still see we have a plugin that provides handlers for the semantic tokens methods and so say we support it (https://github.com/haskell/haskell-language-server/issues/4081).

Probably the "correct" thing to do would be to not statically advertise support and instead dynamically register for it when the plugin is turned on. Similarly we ought to unregister if the plugin is turned off...

This would be pretty complicated to arrange, however, and is quite different from what we do today.

michaelpj:
Wow, this is quite annoying and makes the logs pretty noisy. Perhaps I need to look into it more...

michaelpj:
Brainstorming: 
- We probably need plugins to offer both "static handlers" and "dynamic handlers"
- A "dynamic handler" is going to need to provide both a handler and some contribution to the registration options
- We probably need a semigroup instance for the various registration options to "sum up" what we offer
- When we get a config update, we might need to change what is registered. The simplest thing to do would be to unregister everything and re-register it. That also means we can "re-create" our registration options from scratch, which might be the simplest thing to do.

This is going to need a whole bunch of machinery for tracking what we have dynamically registered and un-registering/re-registering it as appropriate.

Really, anything which can be disabled by config needs to be a dynamic handler!

michaelpj:
Okay, so I think the first step would be to support dynamic registration better in `lsp`, see https://github.com/haskell/lsp/issues/583

That would give us basically all the pieces to do our usual "multiple handlers for one method" thing while also providing capabilities, since we would already have a way to union up the capabilities. 

Then we need the aforementioned mechanics for keeping track of what is dynamically registered and what should be, plus ideally a nice way of defining our handlers so they can be either static or dynamic depending on what the client supports (perhaps also something we can make easier in `lsp`).

fendor:
@VenInf Are you working on this? I have also asked @jetjinser to have a look at this

VenInf:
Hello, @fendor, no, sorry, had to stop my work on this some time ago. I would appreciate it if @jetjinser would take this task.

jetjinser:
Hi @fendor @VenInf, I'll take a look :)

PR #4898, opened 2026-04-18
feat: add `hls-document-link-plugin` to support LSP document links

jetjinser:
This PR introduces a new plugin `hls-document-link-plugin` that adds LSP [documentLink](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentLink) support to Haskell Language Server.

This plugin automatically extracts documentation links for symbols in source code. These links are the same ones that power the `[Documentation]` link in the hover information.

### Changes

- **New plugin:** `hls-document-link-plugin` – extracts clickable links and provides them as document links. Editors can then open these links directly from the source code.
- **Configuration:** Added `documentLinkOn` option to enable/disable the feature (default: `True`).

### Usage

In Neovim 0.12 or later, you can place the cursor on the link and press `gx` to open it directly in your default browser.

The feature can be turned off by setting `documentLinkOn: false` in the HLS configuration.

---

Note. The `getDocumentLinks` helper for `lsp-test` is implemented manually here, but a same change is also pending in [haskell/lsp#638](https://github.com/haskell/lsp/pull/638). After that PR is merged and the dependency is updated, this part can be simplified.


plugins/hls-document-link-plugin/test/Main.hs
```diff
@@ -0,0 +1,126 @@
...
+
+tests :: TestTree
+tests = testGroup "documentLink"
+  [ goldenTest "no document links" "NoDocumentLinks" [],
+    goldenTest "links of primitive types" "Definition"
+      [ (SimilarDocumentLink
+          (DocumentLink (Range (Position 0 10) (Position 0 13))
+                               (Just (Uri "GHC-Types.html#t:Int"))
+                               Nothing
+                               Nothing)),
```
fendor, line 32: These could most likely benefit from some helpers. E.g.,:


```suggestion
      [ (SimilarDocumentLink (mkDocLink (mkRange 0 10 0 13) (Uri "GHC-Types.html#t:Int")),
```

Just a suggestion, feel free to ignore

plugins/hls-document-link-plugin/test/Main.hs
```diff
@@ -0,0 +1,126 @@
...
+        (SimilarDocumentLink
+          (DocumentLink (Range (Position 1 15) (Position 1 24))
+                               (Just (Uri "GHC-Internal-Data-Maybe.html#v:fromMaybe"))
+                               Nothing
+                               Nothing))
+      ]
+  ]
+
+goldenTest :: TestName -> FilePath -> [SimilarDocumentLink] -> TestTree
+goldenTest title file expected = testCase title $ runWithDocumentLink filehs $ do
```
fendor, line 69: Not a golden test, right?

plugins/hls-document-link-plugin/test/Main.hs
```diff
@@ -0,0 +1,126 @@
...
+getResponseResult rsp =
+  case rsp ^. L.result of
+    Right x  -> x
+    Left err -> throw $ UnexpectedResponseError (fromJust $ rsp ^. L.id) err
+
+newtype SimilarDocumentLink = SimilarDocumentLink DocumentLink
+  deriving newtype (Show)
+
+-- custom Eq to ignore some details, such as specific URI
+-- not symmetry
```
fendor, line 99:
```suggestion
-- Custom Eq to ignore some details, such as specific URI.
-- This is not a lawful instance as it doesn't obey the `symmetry` law. Use at your own risk and only in tests.
```

plugins/hls-document-link-plugin/test/testdata/ImportModule.hs
```diff
@@ -0,0 +1,2 @@
+import Data.Maybe (fromJust, fromMaybe)
+f = (fromJust, fromMaybe)
```
fendor, line 2: Would this also work with constructors? E.g., `Just` or `Nothing`

plugins/hls-document-link-plugin/src/Ide/Plugin/DocumentLink.hs
```diff
@@ -0,0 +1,126 @@
...
+                                                   DocumentLinkParams (DocumentLinkParams),
+                                                   TextDocumentIdentifier (TextDocumentIdentifier),
+                                                   type (|?) (InL))
+
+newtype Log = LogShake Shake.Log
+
+instance Pretty Log where
+  pretty = \case {
+    LogShake shakeLog -> pretty shakeLog
+  }
```
fendor, line 63:
```suggestion
  pretty = \case
    LogShake shakeLog -> pretty shakeLog
```

plugins/hls-document-link-plugin/src/Ide/Plugin/DocumentLink.hs
```diff
@@ -0,0 +1,126 @@
...
+  deriving (Show, NFData)
+
+type instance RuleResult GetDocumentLinks = DocumentLinks
+
+getDocumentLinkRule :: Recorder (WithPriority Log) -> Rules ()
+getDocumentLinkRule recorder =
+  defineNoDiagnostics (cmapWithPrio LogShake recorder) $ \GetDocumentLinks nfp -> runMaybeT $ do
+    HAR {hieAst} <- useMT GetHieAst nfp
+    DKMap {getDocMap} <- useMT GetDocMap nfp
+    ast <- hoistMaybe $ getAsts hieAst Map.!? (HiePath . mkFastString . fromNormalizedFilePath) nfp
```
fendor, line 104:
```suggestion
    ast <- hoistMaybe $ getAsts hieAst Map.!? HiePath (mkFastString $ fromNormalizedFilePath nfp)
```

Just weird bracketing for my taste, feel free to ignore.

plugins/hls-document-link-plugin/src/Ide/Plugin/DocumentLink.hs
```diff
@@ -0,0 +1,126 @@
...
+    let lookup = lookupDoc getDocMap
+    pure $ DocumentLinks (foldAst lookup ast)
+
+foldAst :: forall a t. Monoid a => ((Identifier, Span) -> a) -> HieAST t -> a
+foldAst lookup ast = case (nodeChildren ast) of
```
fendor, line 109:
```suggestion
foldAst lookup ast = case nodeChildren ast of
```

fendor, line 109: Some docs would be nice, what is this folding over?

fendor (changes requested):
Welcome back and happy to see you tackle this issue!

This looks pretty good to my cursory review. I have mostly nitpicks and a couple of clarification questions.
Other than that, LGTM :)

fendor:
Also, I am wondering whether this has any overlap with https://github.com/haskell/haskell-language-server/pull/4746 and the `Documentation` and `Source` links in the hover message?

fendor:
[Screencast_20260418_150352.webm](https://github.com/user-attachments/assets/824eb480-7030-427d-a67e-cb7d63a2270e)

This is the behaviour on vscode... I suppose that's not the expected behaviour?

jetjinser:
> Also, I am wondering whether this has any overlap with #4746 and the `Documentation` and `Source` links in the hover message?

Yes. The links provided in this PR share the same source as the 'Documentation' links in the hover messages. The goal here is to expose that information through a more native and convenient LSP feature (DocumentLink).

Therefore, documentLink provides a local hackage link, not the online link provided in #4746.

plugins/hls-document-link-plugin/src/Ide/Plugin/DocumentLink.hs
jetjinser: Added documentation.

This implementation is actually adapted from the Semantic Tokens plugin.
https://github.com/haskell/haskell-language-server/blob/afc877997423128c8338f8f2abd1a2e4671bb547/plugins/hls-semantic-tokens-plugin/src/Ide/Plugin/SemanticTokens/Tokenize.hs#L80-L85

jetjinser:
> Screencast_20260418_150352.webm
> 
> This is the behaviour on vscode... I suppose that's not the expected behaviour?

I’ve done some local testing and noticed the same behavior in VSCodium.

1) If I manually run `xdg-open` on the generated URI (e.g., file:///nix/store/.../docs/index.html#v:nodeChildren), it opens correctly in the external browser. This is also why `gx` works as expected in Neovim. It seems VSCodium's default behavior is to capture file:// URIs and render them in an internal webview/tab instead of passing them to the system handler? 

2) Additionally, I found a minor UX conflict: when a DocumentLink is present, VSCodium's Ctrl+Click defaults to following the link, which "shadows" the Go to Definition action. This might cause some confusion?

---

EDIT:
For the first question,
```json
    "workbench.externalUriOpeners": {
        "file:///*": "simpleBrowser.open"
    },
```
Ideally, this might work? However, I still couldn't get it to open in either an internal or external browser.

crtschin:
> However, I still couldn't get it to open in either an internal or external browser.

I think this might be a lsp consumer problem. I'm using Helix and by default all `file://` links get opened by the editor as well, and only other links get the `xdg-open` treatment. I guess from an editor's pov, there is a legitimate usecase for opening html files in the editor.

I wonder, what's the hueristic nvim uses to determine whether to open files with `xdg-open` or locally in the editor?

jetjinser:
> I wonder, what's the hueristic nvim uses to determine whether to open files with `xdg-open` or locally in the editor?

Nvim doesn't attempt any heuristics.

It simply uses `vim.ui.open`, which always uses `xdg-open` (the equivalent command on other platforms).

https://github.com/neovim/neovim/blob/2ca31eddae0748de5d54207c2d34f7405fed4f29/runtime/lua/vim/_core/defaults.lua#L147-L181

It's possible that Neovim doesn't have a particular reason to open HTML, while a GUI editor does, and the former distinguishes between `gx` (opens ... using the system default handler) and gf (Edit the file ...). XX

vidit-od:
> Additionally, I found a minor UX conflict:

We also should take into consideration how to disable the underlines. 
The by default view makes everything look too cluttered due to lots of underlines.  

example : 
image

fendor:
> We also should take into consideration how to disable the underlines. 

Disabling should work, otherwise this is a VSCode / theme setting, so not in the scope of this PR

fendor:
I think disabling the plugin by default is most likely not a great idea, having a couple of disabled-by-default plugins increases maintenance burden at no benefit for the user.
@wz1000 and I talked, and perhaps it might be a good middle ground if the document links were only shown in haddock comments. Otherwise, the documentation and source links currenlty have this feature already covered.
What do you think about this @jetjinser?

jetjinser:
> I think disabling the plugin by default is most likely not a great idea, having a couple of disabled-by-default plugins increases maintenance burden at no benefit for the user.
> @wz1000 and I talked, and perhaps it might be a good middle ground if the document links were only shown in haddock comments. Otherwise, the documentation and source links currenlty have this feature already covered.
> What do you think about this @jetjinser?

Fair point! But I’m not entirely sure what “shown only in haddock comments” would actually look like -- what kind of content would get a documentLink attached?

As a side note, I later wired up documentLink myself in Neovim through config, and indeed the current hover info is already enough for my needs. (Also, considering it doesn’t seem very practical in VSCode, I (offensively) suspect it’s an interaction that wasn’t really thought through... XD)

fendor:
> -- what kind of content would get a documentLink attached?

I imagined everything that Haddock was able to link, but thinking about it, if haddock was able to link it... then source and documentation link would presumably work already?

jetjinser:
After some thought, I’d like to close this PR. The current form of `documentLink` isn't very useful in practice. If another form of `documentLink` turns out to be needed, I’ll open a new PR.

@fendor is there anything else you think I could help with? :eyes:

fendor:
It took a couple of days, but if you have capacity, the issue https://github.com/haskell/haskell-language-server/issues/4084 is quite important to make progress on :) This would solve a real pain point with the cabal plugin.

PR #4900, opened 2026-04-18
Only check for garbage in key garbage collection test

crtschin:
Fixes the flaky error related to gc'ing keys from the state. The test leaves a gap open for a concurrent action to add unrelated keys to the state. Modifies the assertions so they only check the absence of garbage, instead of comparing on unrelated keys. 

Tested by repeatedly running the specific test and I couldn't reproduce the flaky error anymore.

```
ghcide
  garbage collection
    dirty keys
      are deleted from the state: FAIL (0.15s)
```


fendor (approved):
LGTM, let's get this in!

So exciting, it feels like we are almost at a stable test suite.

It's been a years-long endeavour :)

PR #4904, opened 2026-04-19
Fix Signature Help inside `do` Notation in GHC >= 9.10

### LINKED ISSUE #4769: Signature Help should provide signature in the `do` notation
It looks like no signature help is provided when the AST node is in a `do` notation block:

Image

You can see in the logs, the rules were triggered, but no signature help was provided.
If I change the code in question to not use `do`-notation, then signature help works as expected:

Image

Happened on GHC 9.10.3

vidit-od:
Closes : #4769 

As recommended, Type information of Some nodes were stored in GHC generated nodes. 
This PR keeps the old pipeline and add a Fall back to traverse these generated nodes in case source node lack type information. 

Attached image of signature help in cases where we did not get them before;
image

Prerequisite : 
- [x] passes all hls-signature-help-plugin tests on GHC 9.12.2 : 

image

<br>
Major Finding credits : @jian-lin

### DIFF
--- a/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
+++ b/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
@@ -4,6 +4,7 @@
 
 module Ide.Plugin.SignatureHelp (descriptor) where
 
+import           Control.Applicative
 import           Control.Arrow                        ((>>>))
 import           Control.Monad.Trans.Except           (ExceptT (ExceptT))
 import           Data.Bifunctor                       (bimap)
@@ -49,6 +50,7 @@ import           GHC.Iface.Ext.Types                  (ContextInfo (Use),
                                                        HieAST (nodeChildren, nodeSpan),
                                                        HieASTs (getAsts),
                                                        IdentifierDetails (identInfo, identType),
+                                                       NodeInfo (nodeIdentifiers),
                                                        nodeType)
 import           GHC.Iface.Ext.Utils                  (smallestContainingSatisfying)
 import           GHC.Types.Name.Env                   (lookupNameEnv)
@@ -295,7 +297,10 @@ getNodeNameAndTypes hieKind hieAst =
         case extractName identifier of
           Nothing -> Nothing
           Just name ->
-            let mTypeOfName = identType identifierDetails
+            let mTypeOfName = identType identifierDetails <|> do
+                                nodeInfo <- generatedNodeInfo hieAst
+                                details <- M.lookup identifier (nodeIdentifiers nodeInfo)
+                                identType details
                 -- types from the source NodeInfo
                 typesOfNode = case sourceNodeInfo hieAst of
                   Nothing       -> []
--- a/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
+++ b/plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
@@ -29,6 +29,7 @@ import           Development.IDE.Core.PluginUtils     (runIdeActionE,
 import           Development.IDE.Core.PositionMapping (fromCurrentPosition)
 import           Development.IDE.GHC.Compat           (FastStringCompat, Name,
                                                        RealSrcSpan,
+                                                       generatedNodeInfo,
                                                        getSourceNodeIds,
                                                        isAnnotationInNodeInfo,
                                                        mkRealSrcLoc,
@@ -295,11 +296,19 @@ getNodeNameAndTypes hieKind hieAst =
           Nothing -> Nothing
           Just name ->
             let mTypeOfName = identType identifierDetails
+                -- types from the source NodeInfo
                 typesOfNode = case sourceNodeInfo hieAst of
                   Nothing       -> []
                   Just nodeInfo -> nodeType nodeInfo
+                -- fall back to generated NodeInfo when source has no types
+                typesOfGeneratedNode = case generatedNodeInfo hieAst of
+                  Nothing       -> []
+                  Just nodeInfo -> nodeType nodeInfo
                 allTypes = case mTypeOfName of
-                  Nothing -> typesOfNode
+                  Nothing ->
+                    case typesOfNode of
+                      [] -> typesOfGeneratedNode
+                      ts -> ts
                   -- (the last?) one type of 'typesOfNode' may (always?) be the same as 'typeOfName'
                   -- To avoid generating two identical signature helps, we do a filtering here
                   -- This is similar to 'dropEnd1' in Development.IDE.Spans.AtPoint.atPoint
--- a/plugins/hls-signature-help-plugin/test/Main.hs
+++ b/plugins/hls-signature-help-plugin/test/Main.hs
@@ -391,10 +391,12 @@ main =
             test :: IO ()
             test = do
               x <- pure 1
+                        ^
               print (f x 2)
                        ^ ^
           |]
-          [ Just $ SimilarSignatureHelp $ SignatureHelp [SignatureInformation "f :: Int -> Int -> Int" Nothing (Just [ParameterInformation (InR (5, 8)) Nothing, ParameterInformation (InR (12, 15)) Nothing]) (Just (InL 0))] (Just 0) (Just (InL 0)),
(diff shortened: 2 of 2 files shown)


jian-lin (changes requested):
Looks good.  I'll do a more careful review later.

In the meantime, it would be good to add a test case for the issue.

plugins/hls-signature-help-plugin/test/Main.hs
```diff
@@ -381,6 +381,21 @@ main =
...
+          ],
+        -- Prevents https://github.com/haskell/haskell-language-server/issues/4769
+        mkTest
+          "inside do block"
+          [__i|
+            f :: Int -> Int -> Int
+            f = _
+            test :: IO ()
+            test = do
+              x <- pure 1
```
fendor, line 393:
```suggestion
              x <- pure 1
                        ^
```

Could we add this location as a test case as well?
Unless it is already tested elsewhere

fendor:
@jian-lin I think this is ready for a review now

plugins/hls-signature-help-plugin/test/Main.hs
```diff
@@ -460,9 +478,18 @@ instance IsSimilar SignatureInformation where
...
       | otherwise = False
+    where
+      {- Note[Strip documentation for Pure]
+         ~~~
+         For the First test, we will also get Documentation on signature help.
+         This may break the tests. So we specifically strip docs for pure.
+      -}
+      docsMatch
+        | "pure ::" `T.isPrefixOf` actualLabel = True
+        | otherwise = actualDocumentation ~= expectedDocumentation
```
vidit-od: @fendor Can you plz go over this; 

Triggering a signature help on Pure will come with a big documentation. 
i am stripping the Docs for Pure :: before comparison here. 

image


is it ok ?

jian-lin: > i am stripping the Docs for Pure :: before comparison here.

I am not a big fan of this because what is ignored here is not only just doc text diff, but also if there is doc at all (Nothing vs Just).

> Triggering a signature help on Pure will come with a big documentation.

Nice catch!  Indeed, putting too much doc text into the expected test outputs makes tests flaky.

This is exactly the issue that the newtype `SimilarSignatureHelp` and the type class `IsSimilar` aim to handle.  The [comment][1] for `SimilarSignatureHelp` explicitly mentions ignoring doc string details.

> custom Eq to ignore some details, such as added doc string

Specifically, the doc text diff is handled by [this instance][2] of `IsSimilar`.

So please remove all code and comments related to `docsMatch` and put only a few words into the expected doc instead.  There is an existing [example][3] of `pure` that you can follow directly.

[1]: https://github.com/haskell/haskell-language-server/blob/b6f2b746a5f7d77eacc3b724aa72edd21610f1a7/plugins/hls-signature-help-plugin/test/Main.hs#L424
[2]: https://github.com/haskell/haskell-language-server/blob/b6f2b746a5f7d77eacc3b724aa72edd21610f1a7/plugins/hls-signature-help-plugin/test/Main.hs#L481-L482
[3]: https://github.com/haskell/haskell-language-server/blob/b6f2b746a5f7d77eacc3b724aa72edd21610f1a7/plugins/hls-signature-help-plugin/test/Main.hs#L154

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -295,11 +296,19 @@ getNodeNameAndTypes hieKind hieAst =
           Nothing -> Nothing
           Just name ->
             let mTypeOfName = identType identifierDetails
+                -- types from the source NodeInfo
```
jian-lin, line 304: Taking a closer look at [the HIE AST diff][1], we will find that, besides type info in `types:`, type info in `identifier info:` of source node is also moved to generated node: `Nothing` vs `Just 437`.

This PR currently only handles type info change of `types:`, but not `identifier info:`.  Presumably, more change is needed.

[1]: https://github.com/haskell/haskell-language-server/issues/4769#issuecomment-3592270126

jian-lin (changes requested):

plugins/hls-signature-help-plugin/test/Main.hs
vidit-od: True , That is a way better way to write this test ! 
Done ;

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -294,12 +297,23 @@ getNodeNameAndTypes hieKind hieAst =
         case extractName identifier of
           Nothing -> Nothing
           Just name ->
-            let mTypeOfName = identType identifierDetails
+            let mTypeOfName = identType identifierDetails <|> do
+                                nodeInfo <- generatedNodeInfo hieAst
+                                details <- M.lookup identifier (nodeIdentifiers nodeInfo)
+                                identType details
```
vidit-od, line 303: @jian-lin This should take care of lookup to generated nodes on fallback for identifier details.

fendor:
@jian-lin I think this is waiting on a re-review from you

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
fendor, line 303: @vidit-od Can we extract this `do` block into a function with some documentation?

fendor (approved):
LGTM, thank you very much!

jian-lin:
Thanks for the reminder.  I can have a look later today or tomorrow.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -4,6 +4,7 @@
 
 module Ide.Plugin.SignatureHelp (descriptor) where
 
+import           Control.Applicative
```
jian-lin, line 7: nitpicking:

The existing import style of this file is explicit.  I think it's a good idea to follow that.

```suggestion
import           Control.Applicative ((<|>))
```

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -294,12 +297,23 @@ getNodeNameAndTypes hieKind hieAst =
...
+                -- fall back to generated NodeInfo when source has no types
+                typesOfGeneratedNode = case generatedNodeInfo hieAst of
+                  Nothing       -> []
+                  Just nodeInfo -> nodeType nodeInfo
                 allTypes = case mTypeOfName of
-                  Nothing -> typesOfNode
+                  Nothing ->
+                    case typesOfNode of
+                      [] -> typesOfGeneratedNode
+                      ts -> ts
```
jian-lin, line 316: blocking:

This handles the fallback of `typesOfNode`.  The fallback should be done in _all_ cases.  However, currently, it is only done in the `Nothing` case, but not the `Just` case.

I would handle the fallback in _one_ place before `allTypes` and keep the `allTypes` code unchanged.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -294,12 +297,23 @@ getNodeNameAndTypes hieKind hieAst =
...
-            let mTypeOfName = identType identifierDetails
+            let mTypeOfName = identType identifierDetails <|> do
+                                nodeInfo <- generatedNodeInfo hieAst
+                                details <- M.lookup identifier (nodeIdentifiers nodeInfo)
+                                identType details
+                -- types from the source NodeInfo
                 typesOfNode = case sourceNodeInfo hieAst of
                   Nothing       -> []
                   Just nodeInfo -> nodeType nodeInfo
+                -- fall back to generated NodeInfo when source has no types
```
jian-lin, line 308: The comment is kind of confusing: it comes too late because `mTypeOfName` also has the fallback logic.

I think we should move the comment to the top of the `let` block.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -294,12 +297,23 @@ getNodeNameAndTypes hieKind hieAst =
         case extractName identifier of
           Nothing -> Nothing
           Just name ->
-            let mTypeOfName = identType identifierDetails
+            let mTypeOfName = identType identifierDetails <|> do
+                                nodeInfo <- generatedNodeInfo hieAst
+                                details <- M.lookup identifier (nodeIdentifiers nodeInfo)
```
jian-lin, line 302: For some identifier, are `Identifier`s of `SourceInfo` and `GeneratedInfo` the same?  In other words, does this `M.lookup` always return `Nothing`?

Needs to verify.

plugins/hls-signature-help-plugin/src/Ide/Plugin/SignatureHelp.hs
```diff
@@ -294,12 +297,23 @@ getNodeNameAndTypes hieKind hieAst =
...
+                                identType details
+                -- types from the source NodeInfo
                 typesOfNode = case sourceNodeInfo hieAst of
```
jian-lin, line 305: The fallback logic is:

- get `mTypeOfName` from `SourceInfo` or `GeneratedInfo`
- get `typesOfNode` from `SourceInfo` or `GeneratedInfo`

The current implementation uses two separated checks for `mTypeOfName` and `typesOfNode`.  Alternatively, we can use only one check for both `mTypeOfName` and `typesOfNode`.  Which way is better needs further investigation.

jian-lin (changes requested):
I need more time for further investigation, but I think I should post what I have now.

PR #4910, opened 2026-04-21
Support contextual top-level snippets

### LINKED ISSUE #4905: Top-level completions could be better

When the cursor is on the top-level, and I issue a completion request. Almost all of the results in the dropdown don't make sense. From what I can tell, the completions include symbols that are available within expressions, which aren't valid on the top-level.

Image

For the top-level completions, I think it'd be best to only have snippets related to declarations. So showing snippets for:

- ~~Pragmas~~
- Imports
- Functions
- Typeclasses
- Instances
- Types

If one of these snippets is picked and the user is filling in the placeholders, the regular completions can kick in again.

crtschin:
Closes https://github.com/haskell/haskell-language-server/issues/4905.

Ended up going a bit in the rabbit hole. This adds top-level snippets and includes a bit better nested behavior in deducing the context of completion. 

- It won't suggest snippets that aren't valid in the current context. So no import snippets if declarations started. No declaration snippets prior any import.
- The context that's returned is the tightest one around the completion requests' range.
- I tried to minimize the amount of the AST that's traversed with syb.

### DIFF
--- a/ghcide-test/exe/CompletionTests.hs
+++ b/ghcide-test/exe/CompletionTests.hs
@@ -8,7 +8,7 @@
 module CompletionTests (tests) where
 
 import           Config
-import           Control.Lens                   ((^.))
+import           Control.Lens                   (view, (^.))
 import qualified Control.Lens                   as Lens
 import           Control.Monad
 import           Control.Monad.IO.Class         (liftIO)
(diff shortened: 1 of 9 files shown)


crtschin:
So there's this funny thing, if the context ranges are too tight, then when typing it's possible to fall out of the edge of a range as the contexts won't follow invalid parses. 

This causes a lot of the existing tests to fail, as they put the cursor beyond the range of text. To partially get around this, I don't compare columns when determining whether a context range applies to the current range. This gets it to at pass the existing tests, without me having to manually correct the positions at each one.

crtschin:
This PR is a bit much and does multiple things at the same time. If anyone is looking at this, I'm going to split it up so it's easier to review.

fendor:
@crtschin I asked @VeryMilkyJoe to review this PR, please ping them once you've split it up :) Otherwise, they'll look at this PR in the future.

crtschin:
I've split this into https://github.com/haskell/haskell-language-server/pull/4969, which contains the nested context fixes, and a TBD follow-up that'll add the actual snippets.

Please review that PR instead @VeryMilkyJoe whenever you have some time!

PR #4917, opened 2026-04-27
Use pretty printing context when expanding records

### LINKED ISSUE #4903: Record Inlay hint expansion removes qualified names
```haskell
module Foo where

import qualified Data.Foldable as Foldable

data Foo = Foo { bar :: [Int] }

foo :: Foo
foo = Foo (Foldable.toList [1,2,3])
```

Expanding the `Foo` record inlay hint turns this example into:

```haskell
module Foo where

import qualified Data.Foldable as Foldable

data Foo = Foo { bar :: [Int] }

foo :: Foo
foo = Foo { bar = (toList [1, 2, 3]) }
```

which doesn't typecheck.

https://github.com/user-attachments/assets/03072159-62ea-4729-8d0e-36c6b0f28179

### Your environment

HLS: 2.13.0.0
Linux
GHC: 9.10.3

### Expected behaviour

Shouldn't get rid of the qualified `Foldable` part.

### Actual behaviour

Removes the qualified.

crtschin:
Closes #4903.

There were some existing helpers for determining whether to print qualified/unqualified, use those when printing the expanded records. 

An alternate approach would be to get the ranges from the expressions and use those to extract the expressions from the source files. That would have the benefit that record expansions with erroneous expressions would be nicely transplanted, but I think that would also require detaching `CollectRecords` from type checking, so I didn't pursue.

### DIFF
--- a/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
+++ b/plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
@@ -38,8 +38,8 @@ import           Development.IDE                      (IdeState,
                                                        Recorder (..), Rules,
                                                        WithPriority (..),
                                                        defineNoDiagnostics,
-                                                       getDefinition, hsep,
-                                                       printName,
+                                                       getDefinition, hscEnv,
+                                                       hsep, printName,
                                                        realSrcSpanToRange,
                                                        shakeExtras,
                                                        srcSpanToLocation,
@@ -48,8 +48,7 @@ import           Development.IDE.Core.PluginUtils
 import           Development.IDE.Core.PositionMapping (PositionMapping,
                                                        toCurrentPosition,
                                                        toCurrentRange)
-import           Development.IDE.Core.RuleTypes       (TcModuleResult (..),
-                                                       TypeCheck (..))
+import           Development.IDE.Core.RuleTypes
 import qualified Development.IDE.Core.Shake           as Shake
 import           Development.IDE.GHC.Compat           (FieldLabel (flSelector),
                                                        FieldOcc (FieldOcc),
@@ -71,9 +70,13 @@ import           Development.IDE.GHC.Compat           (FieldLabel (flSelector),
                                                        Var (varName),
                                                        XXExprGhcTc (..),
                                                        conLikeFieldLabels,
-                                                       isGenerated, nameSrcSpan,
+                                                       isGenerated,
+                                                       mkPrintUnqualifiedDefault,
+                                                       nameSrcSpan,
                                                        pprNameUnqualified,
-                                                       recDotDot, unLoc)
+                                                       printSDocQualifiedUnsafe,
+                                                       recDotDot, tcg_rdr_env,
+                                                       unLoc)
 import           Development.IDE.GHC.Compat.Core      (Extension (NamedFieldPuns),
                                                        HsExpr (RecordCon, rcon_flds),
                                                        HsRecField, LHsExpr,
@@ -95,6 +98,8 @@ import           Development.IDE.Spans.Pragmas        (NextPragmaInfo (..),
                                                        insertNewPragma)
 import           GHC.Generics                         (Generic)
 import           GHC.Iface.Ext.Types                  (Identifier)
+import           GHC.Utils.Outputable                 (NamePprCtx,
+                                                       Outputable (..))
 import           Ide.Logger                           (Priority (..),
                                                        cmapWithPrio, logWith,
                                                        (<+>))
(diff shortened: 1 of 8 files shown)


plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -445,10 +465,10 @@ data RecordInfo
...
 
 instance Pretty RecordInfo where
-  pretty (RecordInfoPat ss p) = pretty (printFieldName ss) <> ":" <+> pretty (printOutputable p)
-  pretty (RecordInfoCon ss e) = pretty (printFieldName ss) <> ":" <+> pretty (printOutputable e)
-  pretty (RecordInfoApp ss (RecordAppExpr _ _ fla))
-    = pretty (printFieldName ss) <> ":" <+> hsep (map (pretty . printOutputable) fla)
+  pretty (RecordInfoPat _ p) = pretty (printOutputable p)
+  pretty (RecordInfoCon _ e) = pretty (printOutputable e)
+  pretty (RecordInfoApp _ (RecordAppExpr _ _ fla))
+    = hsep (map (pretty . printOutputable) fla)
```
ozkutuk: Is there a fundamental reason to remove these, apart from the fact that `printFieldName` now requires a `NamePprCtx`? I used to find these helpful for debugging, so I think we can preserve them by inlining the old definition of `printFieldName` instead:

```suggestion
  pretty (RecordInfoPat ss p) = pretty (stripOccNamePrefix (printOutputable ss)) <> ":" <+> pretty (printOutputable p)
  pretty (RecordInfoCon ss e) = pretty (stripOccNamePrefix (printOutputable ss)) <> ":" <+> pretty (printOutputable e)
  pretty (RecordInfoApp ss (RecordAppExpr _ _ fla))
    = pretty (stripOccNamePrefix (printOutputable ss)) <> ":" <+> hsep (map (pretty . printOutputable) fla)
```

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -207,12 +212,19 @@ codeActionResolveProvider :: ResolveFunction IdeState Int 'Method_CodeActionReso
 codeActionResolveProvider ideState pId ca uri uid = do
   nfp <- getNormalizedFilePathE uri
   pragma <- getFirstPragma pId ideState nfp
-  CRR {crCodeActionResolve, nameMap, enabledExtensions} <- runActionE "ExplicitFields.CollectRecords" ideState $ useE CollectRecords nfp
+  (CRR {crCodeActionResolve, nameMap, enabledExtensions}, pprCtx) <- runActionE "ExplicitFields.CodeAction" ideState $ do
```
ozkutuk: I admit this is nitpicking at this point, but why are we renaming the debug name of just at this specific use site? (i.e. from "ExplicitFields.CollectRecords" to "ExplicitFields.CodeAction"). If the idea is to differentiate between different such calls, I can see the value in that, but in that case we should rename _all of them_ to unique names IMO. Also, technically there are two code actions registered by this plugin, so renaming just one of them to "ExplicitFields.CodeAction" might cause some confusion down the line, I imagine.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
```diff
@@ -591,12 +610,15 @@ showRecordConFlds (RecordCon _ _ flds) =
...
+showRecordApp :: NamePprCtx -> RecordAppExpr -> Maybe Text
+showRecordApp pprCtx (RecordAppExpr _ recConstr fla)
+  = Just $ formatOutputable pprCtx recConstr <>  " { "
          <> T.intercalate ", " (showFieldWithArg <$> fla)
          <> " }"
-  where showFieldWithArg (field, arg) = printFieldName field <> " = " <> printOutputable arg
+  where showFieldWithArg (field, arg) = printFieldName pprCtx field <> " = " <> formatOutputable pprCtx arg
+
+formatOutputable :: Outputable a => NamePprCtx -> a -> Text
+formatOutputable pprCtx a = T.pack $ printSDocQualifiedUnsafe pprCtx (ppr a)
```
ozkutuk: The `printOutputable` function we used to use was doing some "unescaping" in the presence of printable escape sequences within double quotes. Is this not necessary anymore with `formatOutputable`?

ozkutuk (approved):
Thanks for taking care of this! I have a question and a few nitpicky remarks, but otherwise everything looks good.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
fendor: Perhaps helpful adding a regression test?

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
crtschin: Good one! No this was me being lazy.

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
crtschin: I'll give each one a unique name 👍🏽

plugins/hls-explicit-record-fields-plugin/src/Ide/Plugin/ExplicitFields.hs
crtschin: It appears this is only relevant for pretty printing [type-level literals](https://github.com/haskell/haskell-language-server/blob/master/ghcide/src/Development/IDE/GHC/Util.hs?plain=1#L276-L278). So the current implementation didn't suffer from `show`-polluted strings as it's all value-level expressions.

To be sure, I've added a test-case with some unicode fields and expressions. I've also refactored it a bit and moved this `formatOutputable` next to the functions and changed it so it also did the unescaping, as it's probably more-wildly applicable.

PR #4919, opened 2026-05-01
Avoid logging `AsyncCancelled` stacktraces on session restart

crtschin:
Tiny log-related cleanup. We cancel the worker thread deliberately with `AsyncCancelled`'s when restarting the shake session, which subsequently get logged with their traces. Seeing stack traces of harmless exceptions in logs isn't the best when trying to debug an issue.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Shake.hs
+++ b/ghcide/src/Development/IDE/Core/Shake.hs
@@ -223,10 +223,12 @@ instance Pretty Log where
       hsep
         [ "Finished:" <+> pretty (actionName delayedAct)
         , "Took:" <+> pretty (showDuration seconds) ]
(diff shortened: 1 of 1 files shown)


fendor:
I wonder whether we want to log that the previous run was cancelled... Is that information perhaps still relevant to know?

I agree that the current behaviour is very confusing, so this is an improvement, thank you!

fendor (approved):
LGTM

crtschin:
> I wonder whether we want to log that the previous run was cancelled... Is that information perhaps still relevant to know?

I think the `Finished build session` message in this PR only really occurs when the previous session was cancelled. Otherwise, there's also the `Restarting build session due to` message that signals that.

PR #4928, opened 2026-05-06
Fixbench

### LINKED ISSUE #4929: Keep benchmark CI up to date
The graph generation deps for benchmark is too far behide and not even used in the CI, in order to make the performance CI useful and up to date with the latest ghc and cabal version. We should trade off by removing the graph generation parts in the benchmark and keep only the core benchmark CI utility to save hls' benchmark from stale.

soulomoon:
fix #4929


.github/workflows/bench.yml
```diff
@@ -19,7 +19,16 @@ jobs:
...
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
+      bench_ghc_versions: ${{ steps.set_ghc_versions.outputs.versions }}
     steps:
+      - id: set_ghc_versions
+        run: |
+          # benching the two latest GHCs we support now
+          # since benchmark are expansive.
+          # choosing the two latest are easier to maintain and more forward looking
+          # see discussion https://github.com/haskell/haskell-language-server/pull/4118
+          # also possible to add more GHCs if we performs better in the future.
```
fendor:
```suggestion
          # Benchmarking only the two latest GHCs we support now
          # since benchmark are expensive.
          # Choosing the two latest is easier to maintain and more forward looking.
          # See discussion https://github.com/haskell/haskell-language-server/pull/4118
          # also possible to add more GHCs if we perform better in the future.
```

shake-bench/shake-bench.cabal
```diff
@@ -16,20 +16,14 @@ source-repository head
     location: https://github.com/haskell/haskell-language-server.git
 
 library
-  if impl(ghc > 9.11)
+  if impl(ghc >= 9.15)
     buildable: False
```
fendor: Even needed at all?

Perhaps 

```suggestion
```

is fine?

cabal.project
```diff
@@ -51,8 +51,10 @@ constraints:
 allow-newer:
   cabal-install-parsers:Cabal-syntax,
 
-if impl(ghc >= 9.11)
+if impl(ghc >= 9.15)
   benchmarks: False
```
fendor:
```suggestion
```

perhaps?

haskell-language-server.cabal
```diff
@@ -1979,8 +1979,7 @@ benchmark benchmark
     main-is: Main.hs
     hs-source-dirs: bench
     build-tool-depends:
-        haskell-language-server:ghcide-bench,
-        eventlog2html:eventlog2html,
```
fendor, line 1983: Are we not using `eventlog2html` anymore to generate the graphs?

fendor (approved):
This looks good to me, but can you explain from a high level view what has changed now? Are we no longer generating the visualisations for the memory usage?

haskell-language-server.cabal
soulomoon: profiled runs still emit raw .hp, we are just not generate the graphs so we can decouple from its deps.

soulomoon:
Three things have been done.

1. enhence the checking before benchmark, ensure the target repo actually buildable and the targets file actually exist.
2. we no longer generating the visualisations for the memory usage.
3. bump ghc and cabal version and update the target repos.

The benchmark actually been broken for the 9.12 for a couple of months for the target repo lsp. So instead of leaving the benchmark stale and no one actually use it and no one pay attention to it. We leave just enough functionalities for our CI.

PR #4931, opened 2026-05-08
Hls status and cradle-info on VScode

vidit-od:
Provide support for a PR from Vscode-haskell ([link](https://github.com/haskell/vscode-haskell/pull/1369))

This is an attempt to improve user experience by making HLS status more expressive. 

logic : 

flowchart (1)

### DIFF
--- a/ghcide-test/exe/CradleTests.hs
+++ b/ghcide-test/exe/CradleTests.hs
@@ -17,8 +17,9 @@ import qualified Data.Text                       as T
 import           Development.IDE.GHC.Util
 import           Development.IDE.Plugin.Test     (TestRequest (..),
                                                   WaitForIdeRuleResult (..))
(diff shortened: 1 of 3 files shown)


dschrempf:
When you say "VSCode", do you mean LSP-client? Just asking on behalf of the little crew that uses different editors. And thanks for working on the status messages!

fendor:
@dschrempf The feature uses a lsp notification that the client can pick up to show the current status.

This requires extra logic in the client, since how you display this notification depends on the client capabilities...
I don't think there is any way in lsp show the status of the server in a generic way, so while the method itself is not VSCode specific, this PR would only be visible with a change in the client.

dschrempf:
I see. And is there no protocol-conform way to signal status? Seems weird to me that we tailor HLS to an editor when the LSP is supposed to be an editor-agnostic communication protocol.

PR #4932, opened 2026-05-12
Avoid relying on `OccNames` when generating class methods

crtschin:
Follows [this comment](https://gitlab.haskell.org/ghc/ghc/-/issues/27195#note_674205) by SPJ.

I'm able to simplify a bit of code following the above comment and using the class type directly instead of going through the binder environment.
- Keep a interval map for the instances in a module instead of a flat list of binders.
- Instead of using HieAST to infer the position and finding the instance declaration via AST traversal, use the cached instance map.
- Remove the hueristics on finding the right binders, by instead taking class methods and instantiating them with the instance types instead.


soulomoon:
the idea look good to me but I have not have time to review the code yet

fendor (approved):
LGTM, thank you very much!

Needs to be rebased, then you can enable the auto-merge (squash)

PR #4936, opened 2026-05-15
Never compute the DocMap with stale data

### LINKED ISSUE #3694: Incorrect overlapping instances errors
I am getting a bunch of incorrect overlapping instances errors while running HLS HEAD on GHC 9.4 . I imagine it's something to do with the multi-home unit stuff. An example of these errors are:
```
• Overlapping instances for Ord Priority arising from a use of ‘>=’
  Matching instances:
    instance Ord Priority
      -- Defined at /home/zhangtianci/MySourceCode/haskell-language-server/ghcide/src/Development/IDE/Types/Logger.hs:80:31
    instance Ord Priority
      -- Defined at /home/zhangtianci/MySourceCode/haskell-language-server/ghcide/src/Development/IDE/Types/Logger.hs:80:31
• In the expression: priority msg >= Info
  In the first argument of ‘cfilter’, namely
    ‘(\ msg -> priority msg >= Info)’
  In the second argument of ‘($)’, namely
    ‘cfilter
       (\ msg -> priority msg >= Info)
       (cmapWithPrio LogLspServer recorder)’
```

fendor:
As the regression test 'eps-pollution' demonstrates, changing the
imports of a module with class instances can poison the HLS session.

To fix this, we simply shouldn't use stale results for computing the
`DocMap`.

Frankly, there is no reason to use stale results when we are not even
using the `PositionMapping`, this sounds like a recipe for disaster in
general. Especially, since the source locations matter in this instance.

The test has been contributed by @mniip and claude, thank you very much for identifying the issue!
I simplified the tests slightly, removing irrelevant details, but it is more or less the same.

I also refactored `checkHover` to add an additional assertion to the test.
This gets rid of a couple of lines of code, which is nice.

Should close #3694

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -161,8 +161,7 @@
     - Ide.PluginUtils
     - Ide.Plugin.Eval.Parse.Comments
     - Ide.Plugin.Eval.CodeLens
-    - FindDefinitionAndHoverTests #Previously part of GHCIDE Main tests
-    - FindImplementationAndHoverTests #Previously part of GHCIDE Main tests
+    - Hover #Used for tests that are allowed to crash
 
   - name: [Prelude.init, Data.List.init]
     within:
--- a/ghcide-test/data/multi-unit-eps-pollution/a-1.0.0-inplace
+++ b/ghcide-test/data/multi-unit-eps-pollution/a-1.0.0-inplace
@@ -1,0 +1,8 @@
+-i
+-ia
+-this-unit-id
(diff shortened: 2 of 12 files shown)


fendor:
Interesting data point, the test doesn't fail on GHC 9.14.1, but it fails on GHC 9.10.3 :thinking:

crtschin (approved):
Fix and tests both LGTM. Nice!

Issue #4871, opened 2026-03-21
Misplaced Haddock comment?


Aster89:
Is _Unique identifier of the plugin_ `pluginId` or `pluginDescription`? I think the former, in which case the comment below should me moved one line up.

https://github.com/haskell/haskell-language-server/blob/dd3dbd75691341126cabc76450ab06003af5cb34/hls-plugin-api/src/Ide/Types.hs#L317-L319

I can do that as part of some other change, if anybody confirms.

fendor:
Yes, this looks misplaced!

fendor:
closed by #4937

PR #4937, opened 2026-05-19
Fix hls-class-plugin indentation

### LINKED ISSUE #4935: hls-class-plugin ignores indentation of already existing method and always indents by 2
### Your environment

Irrelevant to the bug (might be relevant to the solution).

Which OS do you use?
ArchLinux
9.14.1 via ghcup
not relevant

YCM in Vim
2.13 via GHCup

### Steps to reproduce
Ask HLS to add missing methods to the `instance Applicative` below
```haskell
instance Functor Foo where
instance Applicative Foo where
    pure = undefined
```
### Expected behaviour
```haskell
instance Functor Foo where
instance Applicative Foo where
    pure = undefined
   (<*>) = _
```

### Actual behaviour

```haskell
instance Functor Foo where
instance Applicative Foo where
    pure = undefined
  (<*>) = _
```
notice the only 2 spaces of indentation on the last line.

### Debug information

irrelevant

Aster89:
The main reason for this PR is to fix #4935 by having `hls-class-plugin` detect the indentation of already implemented methods (if any, otherwise the default of 2 is used as before) in an `instance` and use that for indenting the missing methods to be inserted.

Additionally, I've included some other commits where I've removed some redundant `()`s, changed some non-ASCII space to ASCII space, and one commit moving a misplaced haddock commet where it belongs, thus fixing #4871.

### DIFF
--- a/ghcide/src/Development/IDE/Types/Diagnostics.hs
+++ b/ghcide/src/Development/IDE/Types/Diagnostics.hs
@@ -164,7 +164,7 @@ ideErrorWithSource source sev fdFilePath msg origMsg =
   in
   ideErrorFromLspDiag lspDiagnostic fdFilePath origMsg
 
--- | Defines whether a particular diagnostic should be reported
+-- | Defines whether a particular diagnostic should be reported
 --   back to the user.
 --
 --   One important use case is "missing signature" code lenses,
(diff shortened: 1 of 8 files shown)


fendor (changes requested):
This looks good to me, thank you!

Could we have a regression test for this new behaviour?

fendor (approved):
LGTM, thank you!

PR #4946, opened 2026-06-04
Fix `transitiveReverseDependencies` recursion

crtschin:
`new` was inverted so the recursion never expanded past the first step, leaving `typecheckParents` to refresh only direct importers on save.


crtschin:
I'll drop the first commit once https://github.com/haskell/haskell-language-server/pull/4947 lands

fendor (approved):
LGTM, this is amazing, how many years have we not noticed this 🙃

PR #4947, opened 2026-06-04
Fix accidentally deleting the installed GHC in CI

crtschin:
CI currently fails on the linux setups, when GCing caches it accidentally uninstalled the active GHC version.


soulomoon (approved):
LGTM

PR #4949, opened 2026-06-06
Remove unused build-toold-depends

fendor:


fendor:
Looks like it is still used :)

PR #4950, opened 2026-06-06
use knownBuildTypes to populate completion targets

### LINKED ISSUE #4921: Cabal build-types should use `knownBuildTypes` from Cabal-syntax
https://hackage-content.haskell.org/package/Cabal-syntax-3.16.1.0/docs/Distribution-Types-BuildType.html#v:Hooks

Instead of string constants, we could use some variation `fmap prettyShow knownBuildTypes`

olaberglund:
closes #4921

Instead of hard coding the `BuildType` [constructors](https://hackage-content.haskell.org/package/Cabal-syntax-3.16.1.0/docs/Distribution-Types-BuildType.html#t:BuildType), simply `show` the `knownBuildTypes` which enumerates them.

Didn't add a `Pretty` instance because that would add an orphan instance, and I would've implemented it using `viaShow`.

The flake changes were needed to enter the dev shell (legacy sdk pattern). See [docs](https://nixos.org/manual/nixpkgs/stable/#sec-darwin-legacy-frameworks).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -83,12 +83,7 @@
             ] ++ lib.optionals (!stdenv.isDarwin)
                    [ # tracy has a build problem on macos.
                      tracy
-                   ]
-              ++ lib.optionals stdenv.isDarwin
-              (with darwin.apple_sdk.frameworks; [
-                Cocoa
-                CoreServices
-              ]);
+                   ];
 
           shellHook = ''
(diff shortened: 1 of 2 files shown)


VeryMilkyJoe (approved):
Perfect, thank you for the PR!

Issue #4948, opened 2026-06-05
Support refactoring the export list


crtschin:
## Is your enhancement request related to a problem? Please describe.

HLS doesn't provide that many tools for working with a module's export list. IIRC the only one is when a diagnostic is raised on an unused symbol, the refactor plugin offers to export the symbol. It would be nice if there were more utilities.

## Describe the solution you'd like

There's a few tools I can imagine being useful.
- [X] [Exporting an already used symbol](https://github.com/haskell/haskell-language-server/pull/4952)
- [X] [Unexporting a symbol.](https://github.com/haskell/haskell-language-server/pull/4991)
- [ ] Exporting all symbols in a module.
- [ ] Turning a module with blanket exports into one that minimally exports according to in-project usages.

Of these (4) is the one I really want, where I often manually go over each exported symbol, do a "find all references" request, and trim the export list if there are no usages. It's also the hardest one to implement. First step is to probably limit the code action to non-exposed modules (not listed in `.cabal` `exposed-modules` list).

This subsumes the existing diagnostic-only action in the `refactor` plugin.

fendor:
I think this issue supersedes #550 

Ill close that one in favour of this one. Someone wanted to work on #550 for zurihac as well

crtschin:
For the export list manipulation actions, it would be nice if they composed well with CPP directives, see this wrinkle: https://github.com/haskell/haskell-language-server/pull/4952#issuecomment-4680567476

PR #4952, opened 2026-06-06
Create `hls-export-plugin` with `Export '...'` code action

crtschin:
First part to https://github.com/haskell/haskell-language-server/issues/4948.

This differs from the existing action that `hls-refactor-plugin` provided:
- `hls-refactor-plugin` only worked when the symbol was detected unused.
- This uses structured diagnostics
- Handles constructor/types nicely
- Uses ghc-exactprint for export splicing


noughtmare:
I've tried this PR with GHC for #4967. This PR indeed fixes the invalid syntax problem, but unfortunately it does expand CPP macros in the export list. That file uses this macro in its export list:

```
#include "primop-vector-tys-exports.hs-incl"
```

crtschin:
> This PR indeed fixes the invalid syntax problem, but unfortunately it does expand CPP macros in the export list. That file uses this macro in its export list:

Thanks for testing this out! 

That's an interesting wrinkle, I believe HLS generally works with post-preprocessor sources. So I'd have to build something to specifically work with preprocessor directives. I won't do that here, as the PR is already quite large, but I'll note it in #4948.

crtschin:
> I won't do that here, as the PR is already quite large...

For the curious, I now have a patch locally for this, but I'll refrain from making a PR until this is merged, as there isn't a nice way to do stacked PRs on GH (yet).

plugins/hls-export-plugin/src/Ide/Plugin/Export/ExactPrint.hs
```diff
@@ -0,0 +1,283 @@
+{-# LANGUAGE CPP                 #-}
```
fendor, line 1: Parts of this module might fit well into `hls-exactprint-utils` as well

plugins/hls-export-plugin/src/Ide/Plugin/Export.hs
```diff
@@ -0,0 +1,90 @@
...
+  pure [ fdLspDiagnostic d | d <- diags, fdFilePath d == nfp, isUnusedTopBind d ]
+  where
+#if MIN_VERSION_ghc(9,8,0)
+    isUnusedTopBind d =
+      case d ^? fdStructuredMessageL . _SomeStructuredMessage . msgEnvelopeErrorL . _TcRnMessage of
+        Just (TcRnUnusedName _ UnusedNameTopDecl) -> True
+        _                                         -> False
+#else
+    isUnusedTopBind _ = False
+#endif
```
fendor: This CPP could be avoided (or pushed to another place) by introducing a pattern synonym for `TcRnUnusedName` in `Compat.Error`. In that pattern synonym, you use the CPP to make sure it can't be mached on older GHC versions.

Edit: Or introduce another prism `_TcRnUnusedName` that won't ever match on older GHC versions

plugins/hls-export-plugin/test/Main.hs
```diff
@@ -0,0 +1,280 @@
...
+
+plugin :: PluginTestDescriptor ()
+plugin = mkPluginTestDescriptor' descriptor "export"
+
+testDataDir :: FilePath
+testDataDir = "plugins" </> "hls-export-plugin" </> "test" </> "testdata"
+
+runExport :: (FilePath -> Session a) -> IO a
+runExport act =
+    runSessionWithTestConfig def
```
fendor: Personally, I like pushing the `testCase label` into the plugin specific test runner, e.g.

```suggestion
runExport :: String -> (FilePath -> Session a) -> IO a
runExport label act =
    testCase label $ runSessionWithTestConfig def
```

But that's just my opinion, this is fine as well.

plugins/hls-export-plugin/test/testdata/hie.yaml
```diff
@@ -0,0 +1,15 @@
+cradle:
```
fendor: Such a cradle can lead to flaky tests as the test session compiles all test modules in each test. Waiting for progress done, or waiting for a specific diagnostic, can sometimes then be signalled from the wrong source file.

Not a blocker, but generally we prefer to use the test helpers that copy the sources individually into temporary directories for better isolation.
Also, this layout might have issues with your work on #4364, as they would be sharing build directories.

fendor (approved):
Awesome patch, thank you very much! I have a couple of small notes and suggestions.

plugins/hls-export-plugin/src/Ide/Plugin/Export/ExactPrint.hs
crtschin: I'll defer this until other plugins start touching the export list. These would have only one consumer in the export plugin.

PR #4953, opened 2026-06-07
Use defaultTestRunner in ghcide-test

xsebek:
We found with @fendor that this test does not support flags like `--lsp-timeout`.


fendor (approved):
LGTM, thank you for catching this!

Issue #4951, opened 2026-06-06
Support for renaming of Haskell modules


VeryMilkyJoe:
When I rename a Haskell module, I want the IDE to:
* change the module declaration to the new name
* change all imports of that module to the new name
* edit the cabal file by changing the entry of that file to the new name

I will  implement this in multiple PRs:
1) Add support for the new plugin options to listen to the notifications that we want to listen for (`willRenameFile`, `didRenameFile`, `didDeleteFile`).
2) On `didRenameFile`, change the cabal-file entry from the old module name to the new one.
3) On `didRenameFile`, change the module-declaration and imports of the renamed module from the old name to the new one.
4) On `didDeleteFile`  delete the file from the diagnostics store .
5) On `didRenameFile` restart the shake session to get new diagnostics.

VeryMilkyJoe:
I have implemented the draft for all of this already, and I will link the PRs to this issue!

PR #4957, opened 2026-06-07
Handle file renaming

VeryMilkyJoe:
Adds handling of `WillRenameFile` Notification to the rename plugin.

When receiving this notification, we:
* change the module name to the new file name in the module declaration of the renamed file, and
* change all imports of the renamed file to the new name.

Note that in the example below, there are errors after renaming because we do not edit the cabal file so the project will not build after a rename.

This is part of the implementation of #4951.

https://github.com/user-attachments/assets/56c13638-0f85-469c-84b8-0e8859b08258

### DIFF
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ModuleName.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ModuleName.hs
@@ -15,6 +15,8 @@ module Ide.Plugin.Rename.ModuleName (
     codeLens,
     updateModuleNameCommand,
     command,
+    potentialModuleNames,
+    codeModuleName,
 ) where
 
 import           Control.Monad                        (forM_, void)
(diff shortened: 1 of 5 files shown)


plugins/hls-rename-plugin/src/Ide/Plugin/Rename/ModuleName.hs
```diff
@@ -127,33 +128,34 @@ action recorder state uri = do
...
+    let paths = map (normalise . (<> pure pathSeparator)) srcPaths
+    logWith recorder Debug (NormalisedPaths paths)
 
-      let suffixes = mapMaybe (`stripPrefix` mdlPath) paths
-      pure (map moduleNameFrom suffixes)
-  where
-    firstLetter :: (Char -> Bool) -> FilePath -> Bool
-    firstLetter _ []       = False
-    firstLetter pred (c:_) = pred c
+    -- TODO, this can be avoid if the filePath is already absolute,
```
AndreasPK:
```suggestion
    -- TODO, this can be avoided if the filePath is already absolute,
```

AndreasPK (changes requested):

dyniec (approved):
Looks good.
Could use small test like: Foo imports Bla, Bla is renamed to Bar, and check that now Foo imports Bar, and Bar is Bar.

Now that #4961 is merged those squiggles in demo would disappear?

VeryMilkyJoe:
> Looks good. Could use small test like: Foo imports Bla, Bla is renamed to Bar, and check that now Foo imports Bar, and Bar is Bar.
> 
> Now that #4961 is merged those squiggles in demo would disappear?

Yes, I am planning to add tests still!

The squiggles are due to the session not being reloaded, I plan to open the next PR, which reloads the session after a rename has been completed, after this one.

PR #4959, opened 2026-06-07
Change HLS default to prefer multi-component loading

### LINKED ISSUE #4930: Make `sessionLoading: multipleComponents` the default
`singleComponent` mode is conceptually broken as it can't uphold the home unit closure property.

`multipleComponents` mode has been released for some time now and is for `cabal` the only way it should be used.

With #4381 fixed, there shouldn't be a huge startup cost to pay either.

Requires us to change the default in HLS and `vscode-haskell`

xsebek:
* closes #4930

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -45,7 +45,7 @@ Here is a list of the additional settings currently supported by `haskell-langua
 - Max completions (`haskell.maxCompletions`, default 40): maximum number of completions sent to the LSP client.
 - Check project (`haskell.checkProject`, default true): whether to typecheck the entire project on initial load. As it is activated by default could drive to bad performance in large projects.
 - Check parents (`haskell.checkParents`, default `CheckOnSave`): when to typecheck reverse dependencies of a file; one of `NeverCheck`, `CheckOnSave` (means dependent/parent modules will only be checked when you save), or `AlwaysCheck` (means re-typechecking them on every change).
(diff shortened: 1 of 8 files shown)


docs/configuration.md
```diff
@@ -45,7 +45,7 @@ Here is a list of the additional settings currently supported by `haskell-langua
 - Max completions (`haskell.maxCompletions`, default 40): maximum number of completions sent to the LSP client.
 - Check project (`haskell.checkProject`, default true): whether to typecheck the entire project on initial load. As it is activated by default could drive to bad performance in large projects.
 - Check parents (`haskell.checkParents`, default `CheckOnSave`): when to typecheck reverse dependencies of a file; one of `NeverCheck`, `CheckOnSave` (means dependent/parent modules will only be checked when you save), or `AlwaysCheck` (means re-typechecking them on every change).
-- Session loading preference (`haskell.sessionLoading`, default `singleComponent`): how to load sessions; one of `singleComponent` (means always loading only a single component when a new component is discovered) or `multipleComponents` (means always preferring loading multiple components in the cradle at once). `multipleComponents` might not be always possible, if the tool doesn't support multiple components loading. The cradle can decide how to handle these situations, and whether to honour the preference at all.
```
xsebek, line 48: @fendor from what you wrote in the Issue, I understand that this is no longer a concern for most users.

Am I interpreting that correctly, or should this cautionary text be preserved in some form?

fendor: This is still valid, but it isn't really a problem as the cradles just fallback gracefully. I think it is fine to simply record that the default is now `multipleComponents`

fendor (approved):
LGTM, thank you very much!

PR #4961, opened 2026-06-08
Handle file rename in .cabal file

VeryMilkyJoe:
On a `WillRenameFile` Notification, we find the responsible `.cabal` file for the old filepath and replace the corresponding module path with the module path for the new file name.

This is part of the implementation of #4951.
In the next PR, I will propagate the rename in Haskell files, i.e. rename in module declaration and imports.

https://github.com/user-attachments/assets/eda10b01-5c5f-493e-879e-b8782f09c67a

### DIFF
--- a/hls-test-utils/src/Test/Hls/Util.hs
+++ b/hls-test-utils/src/Test/Hls/Util.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DataKinds             #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE OverloadedStrings     #-}
 module Test.Hls.Util
   (  -- * Test Capabilities
@@ -37,6 +38,7 @@ module Test.Hls.Util
     , inspectCommand
     , inspectDiagnostic
     , inspectDiagnosticAny
+    , renameFile
     , waitForDiagnosticsFrom
     , waitForDiagnosticsFromSource
     , waitForDiagnosticsFromSourceWithTimeout
@@ -52,7 +54,8 @@ module Test.Hls.Util
 where
 
 import           Control.Applicative.Combinators          (skipManyTill, (<|>))
-import           Control.Exception                        (catch, throwIO)
+import           Control.Exception                        (catch, throw,
+                                                           throwIO)
 import           Control.Lens                             (_Just, (&), (.~),
                                                            (?~), (^.))
 import           Control.Monad
@@ -69,7 +72,7 @@ import qualified Language.LSP.Protocol.Lens               as L
 import           Language.LSP.Protocol.Message
 import           Language.LSP.Protocol.Types
 import qualified Language.LSP.Test                        as Test
-import           System.Directory
+import qualified System.Directory                         as Directory
 import           System.FilePath
 import           System.Info.Extra                        (isMac, isWindows)
 import qualified System.IO.Extra
@@ -80,7 +83,10 @@ import           Test.Tasty.ExpectedFailure               (expectFailBecause,
                                                            ignoreTestBecause)
 import           Test.Tasty.HUnit                         (assertFailure)
 
+import           Data.Foldable                            (traverse_)
 import qualified Data.List                                as List
+import qualified Data.Map                                 as Map
+import           Data.Maybe                               (fromJust)
 import           Data.String.Interpolate                  (__i)
 import qualified Data.Text.Internal.Search                as T
 import qualified Data.Text.Utf16.Rope.Mixed               as Rope
@@ -187,7 +193,7 @@ onlyRunForGhcVersions vers =
 withCurrentDirectoryInTmp :: FilePath -> IO a -> IO a
 withCurrentDirectoryInTmp dir f =
   withTempCopy ignored dir $ \newDir ->
-    withCurrentDirectory newDir f
+    Directory.withCurrentDirectory newDir f
   where
     ignored = ["dist", "dist-newstyle", ".stack-work"]
 
@@ -200,7 +206,7 @@ withCurrentDirectoryInTmp dir f =
 withCurrentDirectoryInTmp' :: [FilePath] -> FilePath -> IO a -> IO a
 withCurrentDirectoryInTmp' ignored dir f =
   withTempCopy ignored dir $ \newDir ->
-    withCurrentDirectory newDir f
+    Directory.withCurrentDirectory newDir f
 
 -- | Example call: @withTempCopy ignored src f@
 --
@@ -219,15 +225,15 @@ withTempCopy ignored srcDir f = do
 -- that are listed in 'ignored'.
 copyDir :: [FilePath] -> FilePath -> FilePath -> IO ()
 copyDir ignored src dst = do
-  cnts <- listDirectory src
+  cnts <- Directory.listDirectory src
   forM_ cnts $ \file -> do
     unless (file `elem` ignored) $ do
       let srcFp = src </> file
           dstFp = dst </> file
-      isDir <- doesDirectoryExist srcFp
+      isDir <- Directory.doesDirectoryExist srcFp
       if isDir
-        then createDirectory dstFp >> copyDir ignored srcFp dstFp
-        else copyFile srcFp dstFp
+        then Directory.createDirectory dstFp >> copyDir ignored srcFp dstFp
+        else Directory.copyFile srcFp dstFp
 
 fromAction :: (Command |? CodeAction) -> CodeAction
 fromAction (InR action) = action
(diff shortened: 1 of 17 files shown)


hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +340,66 @@ failIfSessionTimeout action = action `catch` errorHandler
           errorHandler e@(Test.Timeout _) = assertFailure $ show e
           errorHandler e                  = throwIO e
 
+-- ---------------------------------------------------------------------
+
+renameFile :: TextDocumentIdentifier -> TextDocumentIdentifier -> Test.Session ()
+renameFile from to = do
+    res <- willRenameFile from to
+    case res of
```
Aster89: Since `res` is only used in `case res of`, would using `LambdaCase` really hurt:
```haskell
willRenameFile from to >>= \case
```
After all, I had to scroll down the whole file to see if `res` was used elsewhere. If you use `LambdaCase`, there's no room for thinking about it.

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +340,66 @@ failIfSessionTimeout action = action `catch` errorHandler
...
+                (Map.assocs e)
+            pure ()
+        InL (WorkspaceEdit _ (Just docChanges) _) -> do
+            traverse_
+                ( \case
+                    InL edit ->
+                        traverse_
+                            (\e -> Test.applyEdit (TextDocumentIdentifier $ edit ^. L.textDocument . L.uri) $ mkTextEdit e)
+                            (edit ^. L.edits)
+                    _ -> error "todo"
```
Aster89: To be done later in the process, but still in this PR, I assume?

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +340,66 @@ failIfSessionTimeout action = action `catch` errorHandler
...
+                )
+                docChanges
+            pure ()
+        _ -> pure ()
+    didRenameFile from to
+    where
+        mkTextEdit :: TextEdit |? AnnotatedTextEdit -> TextEdit
+        mkTextEdit (InL e)   = e
+        mkTextEdit (InR ate) = TextEdit (ate ^. L.range) (ate ^. L.newText)
```
Aster89: Out of curiosity, is _ate_ the past of _to eat_?

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +340,66 @@ failIfSessionTimeout action = action `catch` errorHandler
...
+didRenameFile from to =
+    Test.sendNotification
+        SMethod_WorkspaceDidRenameFiles
+        ( RenameFilesParams
+            [FileRename (docIdToText from) (docIdToText to)
+            ]
+        )
+  where
+        docIdToText :: TextDocumentIdentifier -> T.Text
+        docIdToText tdi = getUri $ tdi ^. L.uri
```
Aster89: Why not pulling it out of `where` and sharing with the one 12 lines above or so?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
```diff
@@ -0,0 +1,242 @@
...
+
+data Log
+  = LogDidRename FilePath FilePath
+  | CabalRenameLog T.Text T.Text
+  deriving (Show)
+
+instance Pretty Log where
+  pretty = \case
+    LogDidRename oldFp newFp -> "Received rename info from:" <+> pretty oldFp <+> "to:" <+> pretty newFp
+    CabalRenameLog oldModulePath newModulePath -> "Executing rename of module from" <+> pretty oldModulePath <+> "to" <+> pretty newModulePath <+> "in cabal file."
```
Aster89: Given some `import` lists are split on more lines to be short, I'd multi-line these too.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
```diff
@@ -0,0 +1,242 @@
...
+  moduleName = Cabal.fromString (T.unpack modulePath)
+
+  findInOtherModules mods =
+    Add.OtherModules <$ findInModules mods
+
+  findInExposedModules mods =
+    Add.ExposedModules <$ findInModules mods
+
+  findInModules mods =
+    guard $ moduleName `elem` mods
```
Aster89, line 192: Any reason for splitting lines that would be shorter than some current one-liner?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
```diff
@@ -0,0 +1,242 @@
...
+
+  getExposedModulesForLib :: ComponentName -> Library -> Maybe [Cabal.ModuleName]
+  getExposedModulesForLib compName lib =
+    case libName lib of
+      LSubLibName lName ->
+        case componentNameString compName of
+          Just unqualCompName -> if lName == unqualCompName then Just $ exposedModules lib else Nothing
+          _ -> Nothing
+      _ -> Nothing
```
Aster89, line 199: Just exploring whether view patterns can make something really more readable 😄 
```haskell
  getExposedModulesForLib     (componentNameString -> Just unqualCompName)
                          lib@(    libName         -> LSubLibName lName)
  | lName == unqualCompName = Just $ exposedModules lib
  | otherwise = Nothing
```

dyniec (approved):
rename yay!
Looks fine,and tests seem alright. Should be fine to merge when CI succeeds

Aster89 (commented):
Love this!

Some newbie comments below 😆

hls-test-utils/src/Test/Hls/Util.hs
VeryMilkyJoe: No, this is AnnotatedTextEdit, compared to Edit above. 
It's a little unfortunate but I didn't think it was that important to change.

hls-test-utils/src/Test/Hls/Util.hs
VeryMilkyJoe: Yeah I will change, that, good call!

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
VeryMilkyJoe: Do you mean, why did I drag this into the `where`?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
VeryMilkyJoe: I'm not a big fan of view patterns, also I don't really think it improves readability here.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
Aster89: Sorry I wasn't clear.

What I meant was: why doing
```haskell
  findInOtherModules mods =
    Add.OtherModules <$ findInModules mods
```
instead of
```haskell
  findInOtherModules mods = Add.OtherModules <$ findInModules mods
```
when there's anyway plenty of lines longer than the latter?

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
Aster89: You could stil pattern match on the pair, given that the two scrutinies are independent:
```haskell
 getExposedModulesForLib compName lib =
    case (libName lib, componentNameString compName) of
      (LSubLibName lName, Just unqualCompName) | lName == unqualCompName -> Just $ exposedModules lib
      _ -> Nothing
```

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
VeryMilkyJoe: I think I prefer the first one, cause it feels nicer to read.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
VeryMilkyJoe: sure

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +341,86 @@ failIfSessionTimeout action = action `catch` errorHandler
...
+                            )
+                            (edit ^. L.edits)
+                    InR unsupported ->
+                        error $ "renameFile: Unsupported WorkspaceEdit " <> (show unsupported)
+                )
+                docChanges
+            pure ()
+        _ -> pure ()
+    liftIO $ Directory.renameFile (getFilePath from) (getFilePath to)
+    didRenameFile from to
```
AndreasPK, line 385: This can race with renameFile on windows.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
```diff
@@ -0,0 +1,243 @@
...
+  logWith recorder Info $ LogDidRename oldHaskellFilePath newHaskellFilePath
+  (contents, fields, gpd, verTxtDocId) <- runActionE "cabal-plugin.getUriContents" ideState $ do
+    let nuri = toNormalizedUri $ filePathToUri cabalFilePath
+        nfp = toNormalizedFilePath cabalFilePath
+    mContent <- lift $ getUriContents nuri
+    verTxtDocId <-
+      runActionE "cabalAdd.getVersionedTextDoc" ideState $
+        lift $
+          getVersionedTextDoc $
+            TextDocumentIdentifier (filePathToUri cabalFilePath)
```
AndreasPK, line 94: Should this take the normalized uri?

VeryMilkyJoe: No, it takes a `Uri` as defined by the type of `TextDocumentIdentifier`.

plugins/hls-cabal-plugin/src/Ide/Plugin/Cabal/CabalAdd/Rename.hs
```diff
@@ -0,0 +1,243 @@
...
+findFieldForModule modulePath compName pd buildInfo =
+  let
+    exposedMods = case compName of
+      CLibName name ->
+        case name of
+          LMainLibName ->
+            Maybe.maybe [] exposedModules $ library pd
+          LSubLibName _ ->
+            concat $ Maybe.concatMapM (getExposedModulesForLib compName) $ subLibraries pd
+      _ -> []
```
AndreasPK, line 178:
```suggestion
      CNotLibName -> ?
```

I think executables can have other-modules and need to be handled here but not 100% sure.

VeryMilkyJoe: This is handled later on, the first part here checks only whether the module is in exposed modules if it occurs, since there are more limited cases for this. 
If the module still occurs in a field but was not in an exposed module then it has to be in other modules, which would cover the case you proposed.
I changed the code a bit to hopefully improve readability.

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +341,89 @@ failIfSessionTimeout action = action `catch` errorHandler
...
+-- ---------------------------------------------------------------------
+
+-- | Simulate the client action of renaming a file
+-- This consists of sending a `WillRenameFiles` request renaming the file from `from` to `to`,
+-- and applying the returned `WorkspaceEdit` changes and actually renaming the file.
+-- Finally sends a `DidRenameFile` notification to the server.
+renameFile :: TextDocumentIdentifier -> TextDocumentIdentifier -> Test.Session ()
+renameFile from to = do
+    willRenameFile from to >>= \case
+        InL (WorkspaceEdit (Just e) _ _) -> do
```
fendor: The `WorkSpaceEdit` should be sorted first with `reverseSortEdit` from `lsp` before applying

hls-test-utils/src/Test/Hls/Util.hs
```diff
@@ -335,6 +341,89 @@ failIfSessionTimeout action = action `catch` errorHandler
...
+            traverse_
+                ( \(uri, edits) ->
+                    traverse (Test.applyEdit (TextDocumentIdentifier uri)) edits
```
fendor: This should be using `applyEdits`, but I am fine with changing this once we upstream this to `lsp-test`.

PR #4965, opened 2026-06-08
Fix deleting unused bindings with Haddock docs V2

### LINKED ISSUE #4876: Delete unused binding code action doesn't delete haddock

xsebek:
* Add test case from #4881 by @agam263
* Use the AST to get the associated comment range - its on the annotation of the signature `L EpAnn SigD`, not on the definition `ValD`
* Closes #4876
* Closes #4881


plugins/hls-refactor-plugin/test/Main.hs
```diff
@@ -2533,6 +2533,46 @@ deleteUnusedDefinitionTests = testGroup "delete unused definition action"
...
       , "some = ()"
       ]
+  , testSession "delete unused top level binding with Haddock comment" $
+    testFor
+      [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+      , "module A (some) where"
+      , ""
+      , "-- | line docs for f"
+      , "f :: Int"
+      -- TODO: , "-- ^ trailing docs for f"
```
xsebek: @fendor Could you please help me decide what to do? There are a two problems with my current version:

1. The GHC AST changed a lot in 9.10 and the types do not match. It would mean writing the code that finds comments again. Should I dive into type families, or disable (how?) this test for GHC 9.6/.8?
    * Apply.Refact.Compat could help, but that does not in turn work on 9.10
2. Trailing Haddocks might not be in the same place after GHC does the parsing, the test would not pass. Maybe it's connected to this note about GHC parser only adding "prior comments":
  https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Parser/Annotation.hs?ref_type=heads#L406-413
    * Trailing style for function declaration seems niche to me, I only used it on record fields. Is this worth fixing, or is there maybe some better example?

santiweight: @alanz may have some insights - i'm too far away

alanz: The parser emits an AST with the comments preceding only, but ghc-exactprint has [balanceComments](https://github.com/alanz/ghc-exactprint/blob/master/src/Language/Haskell/GHC/ExactPrint/Transform.hs#L393) which spreads them as appropriate.  And if you follow that link the comment talks about how exact print processes haddock comments (TL;DR: it ignores them, they show up as regular comments too).

Also, if you use the [HasDecls](https://github.com/alanz/ghc-exactprint/blob/master/src/Language/Haskell/GHC/ExactPrint/Transform.hs#L816) facilities I think it might balance comments when you ask for the decls. But I may be wrong.  But after balancing comments you should be able to just remove the decl from the ast, and the comments go with it.

xsebek: Thanks a lot @alanz! 👍 That does indeed put the docs on the function signature. Except in this case:
```Haskell
foo :: Int
-- ^ Trailing doc?
foo = 42
```
This seems like a weird edge-case to me, where the exact print decided to leave the doc on `foo` value definition instead of consistently moving it to the function signature. I only found out, because it was the first test I hastily wrote. I am not sure if this is a bug in ghc-exactprint.

It would be annoying to get the comments from value definition as it would require some CPP magic. 😕

alanz: In terms of exact print, it treats haddock comments as normal comments, and ignored the specially haddock-processed ones,

When it balances comments, it assumes a comment stays where it is unless there is a blank line to make it clear.

So

```haskell
foo :: Int
-- ^ Trailing doc?

foo = 42
```

would balance the comment as trailing to the signature.

Perhaps an enhancement to `balanceComments` would be to use the haddock comment prefixes to decide on the comment balancing. PR on ghc-exactprint welcome

PR #4966, opened 2026-06-09
Remove `LD_LIBRARY_PATH` override in nix `devShell`

crtschin:
I get glibc mismatch errors (`version 'GLIBC_2.42' not found`) errors with this override between the nix provided shared libraries and what my host system expects. I don't run nixos. These shared libraries are included in `buildInputs`, so should already be available to nix-provided binaries.

I didn't touch `DYLD_LIBRARY_PATH` as I don't run a mac, and I can't evaluate.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -87,7 +87,6 @@
 
           shellHook = ''
             # @guibou: I'm not sure theses lines are needed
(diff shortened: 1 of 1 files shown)


guibou:
Haha, you see the comment above saying that I have no idea why this is needed ? Well, I still don't know ;)

If the shell works for you this way, go.

guibou (approved):

PR #4969, opened 2026-06-10
Fix completion contexts in nested scopes

crtschin:
The previous completion code took the first context returned through descent into the AST. 

Moved context determination into it's own module and adjusted the descent to keep track of the "tightest" context around the cursor. Same caveat applies as #4910, I don't compare columns so the it has better support for invalid parses which are common while typing.

This doesn't contain the top-level snippet from #4905 yet, that'll be in a follow-up.

### DIFF
--- a/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -14,66 +14,66 @@ module Development.IDE.Plugin.Completions.Logic (
 , getCompletionPrefixFromRope
 ) where
 
-import           Control.Applicative
-import           Control.Lens                             hiding (Context,
-                                                           parts)
-import           Data.Char                                (isAlphaNum, isUpper)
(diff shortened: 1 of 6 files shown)


ghcide/src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -551,16 +470,13 @@ getCompletions
...
-    | Just (ImportListContext moduleName) <- maybeContext
-    = moduleImportListCompletions moduleName
-
-    | Just (ImportHidingContext moduleName) <- maybeContext
+    -- IMPORT MODULENAME (NAM|) and IMPORT MODULENAME hiding (NAM|)
+    | ImportListContext moduleName <- context
     = moduleImportListCompletions moduleName
 
     -- ------------------------------------------------------------------------
     -- IMPORT MODULENAM|
```
VeryMilkyJoe: You could maybe fix this type, while you're in the area :D

PR #4971, opened 2026-06-12
Try getting rid of `withWarnings` in `~Compile.hs`

Aster89:
I am making this change in isolation because it is not working as part of [my change for introducing the case split plugin](https://github.com/haskell/haskell-language-server/compare/master...Aster89:haskell-language-server:master), so with this PR the intent is to see if the change is genuine in and of itself.

By the way, @fendor, you suggested _"getting rid of `fmap` and giving names"_ with reference to the top of `compileModule`. I haven't removed `fmap` because that's just digging into `IO`... what makes things less clear is probably `either`, I think, hence I used a `\case` to give names to the two `Left` and `Right` contents.

### DIFF
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -40,8 +40,7 @@ module Development.IDE.Core.Compile
   ) where
 
 import           Control.Concurrent.STM.Stats                 hiding (orElse)
-import           Control.DeepSeq                              (NFData (..),
-                                                               force, rnf)
+import qualified Control.DeepSeq                              as DeepSeq
 import           Control.Exception                            (evaluate)
 import           Control.Exception.Safe
 import           Control.Lens                                 hiding (List, pre,
@@ -110,12 +109,13 @@ import qualified Data.Set                                     as Set
 import qualified GHC                                          as G
 import           GHC.Core.Lint.Interactive
 import           GHC.Driver.Config.CoreToStg.Prep
+import           GHC.Driver.Env                               (runHsc')
+import           GHC.Driver.Main                              (hscDesugar', hscSimplify')
 import           GHC.Iface.Ext.Types                          (HieASTs)
 import qualified GHC.Runtime.Loader                           as Loader
 import           GHC.Tc.Gen.Splice
 import           GHC.Types.Error
 import           GHC.Types.ForeignStubs
-import           GHC.Types.HpcInfo
 import           GHC.Types.TypeEnv
 
 -- See Note [Guidelines For Using CPP In GHCIDE Import Statements]
@@ -143,6 +143,10 @@ import           Development.IDE.GHC.Compat                   hiding
                                                                writeHieFile)
 #endif
 
+#if !MIN_VERSION_ghc(9,11,0)
+import           GHC.Types.HpcInfo                            (emptyHpcInfo)
+#endif
+
 #if MIN_VERSION_ghc(9,11,0)
 import qualified Data.List.NonEmpty                           as NE
 import           Data.Time                                    (getCurrentTime)
@@ -156,7 +160,6 @@ import           GHC.Unit.Home.Graph                          (UnitEnvGraph(..),
 import           GHC.Unit.Home.PackageTable                   (hptInternalTableRef, hptInternalTableFromRef)
 import           GHC.Unit.Module.ModIface                     (IfaceTopEnv(..))
 import           GHC.Types.Avail                              (emptyDetOrdAvails)
-import           GHC.Types.Basic                              (ImportLevel(..), convImportLevel)
 #endif
 
 #if MIN_VERSION_ghc(9,12,0)
@@ -524,7 +527,7 @@ mkHiFileResultCompile se session' tcm simplified_guts = catchErrs $ do
                                   (tcg_import_decls (tmrTypechecked tcm))
                                   simplified_guts
 #else
-  let !partial_iface = force $ mkPartialIface session
+  let !partial_iface = DeepSeq.force $ mkPartialIface session
                                               (cg_binds guts)
                                               details
                                               ms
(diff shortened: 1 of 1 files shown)


ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -685,13 +690,24 @@ compileModule (RunSimplifier simplify) session ms tcg =
...
                -- MP: the flags in ModSummary should be right, if they are wrong then
                -- the correct place to fix this is when the ModSummary is created.
-               desugar <- hscDesugar session' (ms { ms_hspp_opts = hsc_dflags session' }) tcg
+               (desugar, ms :: Messages GhcMessage) <- do
+                             let dflags = (hsc_dflags session') { maxUncoveredPatterns = 1000 }
+                             aw@(_, w) <- runHsc' session' $ do
+                                    hscDesugar' (ms_location (ms { ms_hspp_opts = hsc_dflags session' })) tcg
+                             let !diag_opts = initDiagOpts dflags
+                                 !print_config = initPrintConfig dflags
+                             printOrThrowDiagnostics (hsc_logger session') print_config diag_opts w
```
fendor:
```suggestion
```

Printing was for your debugging ,right?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -685,13 +690,24 @@ compileModule (RunSimplifier simplify) session ms tcg =
...
                if simplify
                then do
                  plugins <- readIORef (tcg_th_coreplugins tcg)
-                 hscSimplify session' plugins desugar
-               else pure desugar
-            return (map snd warnings, desugared_guts)
+                 (a, w) <- runHsc' session' $ hscSimplify' plugins desugar
+                 let !diag_opts = initDiagOpts dflags
+                     !print_config = initPrintConfig dflags
+                 printOrThrowDiagnostics (hsc_logger session') print_config diag_opts w
```
fendor:
```suggestion
```

Same here

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -685,13 +690,24 @@ compileModule (RunSimplifier simplify) session ms tcg =
                -- TODO: maybe settings ms_hspp_opts is unnecessary?
                -- MP: the flags in ModSummary should be right, if they are wrong then
                -- the correct place to fix this is when the ModSummary is created.
-               desugar <- hscDesugar session' (ms { ms_hspp_opts = hsc_dflags session' }) tcg
+               (desugar, ms :: Messages GhcMessage) <- do
+                             let dflags = (hsc_dflags session') { maxUncoveredPatterns = 1000 }
```
fendor:
```suggestion
                             let dflags = hsc_dflags session'
```

Debugging remnant

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -685,13 +690,24 @@ compileModule (RunSimplifier simplify) session ms tcg =
...
+                             aw@(_, w) <- runHsc' session' $ do
+                                    hscDesugar' (ms_location (ms { ms_hspp_opts = hsc_dflags session' })) tcg
+                             let !diag_opts = initDiagOpts dflags
+                                 !print_config = initPrintConfig dflags
+                             printOrThrowDiagnostics (hsc_logger session') print_config diag_opts w
+                             pure aw
+               let dflags = hsc_dflags session'
```
fendor:
```suggestion
```

Probably redundant?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -685,13 +690,24 @@ compileModule (RunSimplifier simplify) session ms tcg =
...
                  plugins <- readIORef (tcg_th_coreplugins tcg)
-                 hscSimplify session' plugins desugar
-               else pure desugar
-            return (map snd warnings, desugared_guts)
+                 (a, w) <- runHsc' session' $ hscSimplify' plugins desugar
+                 let !diag_opts = initDiagOpts dflags
+                     !print_config = initPrintConfig dflags
+                 printOrThrowDiagnostics (hsc_logger session') print_config diag_opts w
+                 pure (diagFromErrMsgs "compile" dflags $ getWarningMessages w, a)
+               else do pure (diagFromErrMsgs "compile" dflags $ getWarningMessages ms, desugar)
```
fendor: Instead of calling `diagFromErrMsgs` twice, perhaps it makes sense to call it once in the `\case` expression?

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -156,11 +158,14 @@ import           GHC.Unit.Home.Graph                          (UnitEnvGraph(..),
...
 import           GHC.Types.Avail                              (emptyDetOrdAvails)
-import           GHC.Types.Basic                              (ImportLevel(..), convImportLevel)
 #endif
 
 #if MIN_VERSION_ghc(9,12,0)
 import           Development.IDE.Import.FindImports
+import GHC.Driver.Main (hscDesugar', hscSimplify')
+import GHC.Driver.Config.Diagnostic (initPrintConfig)
+import GHC.Driver.Errors (printOrThrowDiagnostics)
+import GHC.Plugins (DynFlags(..))
```
fendor: These imports probably need to be outside of the CPP block, otherwise we will only use them in GHC 9.12+

fendor (changes requested):
Nice improvement! Some code nitpicks.

fendor:
> I haven't removed fmap because that's just digging into IO... what makes things less clear is probably either, I think, hence I used a \case to give names to the two Left and Right contents.

Personally, I feel like the reading direction is what makes this tricky to parse, not the use of `fmap` or `either`.
Now that we both understand the code, it is easy to see what it does, but I think the explicit names for the result of the `IO` action would still be an improvement.
However, Ill leave this up to you, if you think the code is now easier to read, then leave it like this.
Perhaps try the refactoring regardless and see if you like it more?

ghcide/src/Development/IDE/Core/Compile.hs
Aster89: I have absolutely no recollection at all of having written these lines, but I've greped through git and appareantly I haven't even copied them from anywhere, so it must have been me D:

ghcide/src/Development/IDE/Core/Compile.hs
Aster89: This is redundant.

ghcide/src/Development/IDE/Core/Compile.hs
Aster89: Probably more the other one a few lines up.

ghcide/src/Development/IDE/Core/Compile.hs
Aster89: But the last argument is different in the two calls... I'm not sure I follow what you mean.

fendor: If you moved everything to be executed in a single `runHsc'` block, then I think you would collect all warnings once and there should be only a single `Messages GhcMessage`.

ghcide/src/Development/IDE/Core/Compile.hs
```diff
@@ -40,8 +40,11 @@ module Development.IDE.Core.Compile
   ) where
 
 import           Control.Concurrent.STM.Stats                 hiding (orElse)
-import           Control.DeepSeq                              (NFData (..),
-                                                               force, rnf)
+import           Control.DeepSeq                              (NFData (..), rnf
+#if !MIN_VERSION_ghc(9,13,0)
+                                                              , force
+#endif
+                                                              )
```
fendor:
```suggestion
import           Control.DeepSeq                              as DeepSeq
```

And replace the usages with `DeepSeq.rnf` and `DeepSeq.force`. Avoiding CPP sound worths it to me :)

ghcide/src/Development/IDE/Core/Compile.hs
Aster89: But there's still an `if simplify`, and `hscSimplify'` is called in only in the `then` case.

fendor: Right but you could write something to the effect of:

```haskell
aw@(_, w) <- runHsc' session' $ do
  desugar <- hscDesugar' (ms_location (ms { ms_hspp_opts = hsc_dflags session' })) tcg
  if simplify
               then do
                 plugins <- liftIO $ readIORef (tcg_th_coreplugins tcg)
                 hscSimplify' plugins desugar
               else pure desugar           
```

Aster89:
> > I haven't removed fmap because that's just digging into IO... what makes things less clear is probably either, I think, hence I used a \case to give names to the two Left and Right contents.
> 
> Personally, I feel like the reading direction is what makes this tricky to parse, not the use of `fmap` or `either`. Now that we both understand the code, it is easy to see what it does, but I think the explicit names for the result of the `IO` action would still be an improvement. However, Ill leave this up to you, if you think the code is now easier to read, then leave it like this. Perhaps try the refactoring regardless and see if you like it more?

Any better now? I'm not too happy with the names I've given to things, but if they're acceptable..

fendor:
The code looks very good now! We fail to compile everything, I hope that's just due to imports and not something like `runHsc'` not being exported in GHC 9.6 :grimacing: That'd be very annothing.

ghcide/src/Development/IDE/Core/Compile.hs
Aster89: By the way, I'm pretty sure it's HLS that has added these inside an `#if` guard. By the titles, I wouldn't say any of [these](https://github.com/haskell/haskell-language-server/issues?q=is%3Aissue%20%60import%60%20%22guard%22) is about it. Shall I open an issue?

Aster89:
> The code looks very good now! We fail to compile everything, I hope that's just due to imports and not something like `runHsc'` not being exported in GHC 9.6 😬 That'd be very annothing.

Nope, looks like what's left is [a Win-only test failure with GHC 9.12](https://github.com/haskell/haskell-language-server/actions/runs/27570443632/job/81505331414?pr=4971).

And now [a Linux-only test failure with GHC 9.10](https://github.com/haskell/haskell-language-server/actions/runs/27579191473/job/81535129207?pr=4971). Is this just test flakiness?

fendor:
> And now [a Linux-only test failure with GHC 9.10](https://github.com/haskell/haskell-language-server/actions/runs/27579191473/job/81535129207?pr=4971). Is this just test flakiness?

Yep, flaky, good job!

fendor (approved):
LGTM, thank you for the improvement!

PR #4972, opened 2026-06-14
Extend delete-unused-bindings to let-expression bindings

olaberglund:
This is an extension to the code action that deletes all unused bindings. It will now also delete all unused local let expression bindings.

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
@@ -30,6 +30,7 @@ import           Data.Char
 import qualified Data.DList                                        as DL
 import           Data.Function
 import           Data.Functor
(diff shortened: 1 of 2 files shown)


santiweight:
there's one interesting case here i can think of: `let !my_strict = expensive in foo`. but in that case user should probably just be underscoring

lgtm except for you need to add some CPP stuff sadly for later GHC versions

also: i don't haskell much anymore, and the fact you got this working in like 5 lines (that i don't really understand too well any more lol) is crazy

olaberglund:
@santiweight thank you! A colleague had hinted that SYB would be perfect here, so I read the paper and applied the design pattern here. Super cool stuff.

Edit: I just tried it with a unused strict binding, and it gets deleted as well, as expected. Interesting point though, I'm open to try to exclude strict bindings in the future.

santiweight (approved):

santiweight:
@olaberglund thanks for the contribution :)

PR #4978, opened 2026-06-17
Support symbolic record fields when converting to record syntax

### LINKED ISSUE #4975: Convert to record syntax wrong for operator fields
If a record constructor has a field named as an operator, "convert to traditional record syntax" inserts the operator unparenthesised. If a record constructor is itself an operator, there is no suggestion (but this seems very unlikely).

Image

### Steps to reproduce

```Haskell
data Example a = Example { (><) :: a -> a -> a }
f a = Example a
```

### Expected behaviour

Suggests 
```Haskell
f a = Example { (><) = a }
```

### Actual behaviour

Suggests 
```Haskell
f a = Example { >< = a }
```

olaberglund:
Closes #4975 

If the field name consists of symbolic characters, add surrounding parentheses before rendering inlay hint or field with argument when converting to record syntax.

Extended relevant test case to feature a record field made up of symbolic characters.


fendor (approved):
LGTM, except for the test failures ;)

PR #4981, opened 2026-06-19
Splice export modifications precisely when using CPP

crtschin:
Builds on top of https://github.com/haskell/haskell-language-server/pull/4952 and handles CPP somewhat gracefully.

If possible, we stick with `ghc-exactprint`. When we encounter a CPP directive in the export list, which is invisible to `ghc-exactprint`, we fallback to using direct text splicing. In this case we only follow the active branch in the CPP conditional.

A lot of this is tests... something about combinatorics and edgecases.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -23,6 +23,7 @@
 /plugins/hls-explicit-fixity-plugin
 /plugins/hls-explicit-imports-plugin
 /plugins/hls-explicit-record-fields-plugin @ozkutuk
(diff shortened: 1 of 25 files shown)


fendor (approved):
I reviewed the changes outside the `hls-export-plugin`, the tests look sane, please merge once you are ready or ping me on specific parts you want to discuss

crtschin:
Thanks! I think it's alright, none of this is complicated.

PR #4449, opened 2024-11-13
Typed rule inputs

nlander:

PR #4983, opened 2026-06-20
Typed rules

vidit-od:
This PR contains work for Part-1 for [Goto dependency definition](https://summerofcode.withgoogle.com/programs/2026/projects/WKEV8qIK), part of GSoC' 26.

Aims to introduce Typed Rules. This will allow us to selectively make sure which rules can be triggered on Dependency files. 
Essentially helps prevent triggering of Type-checking on Dependency files  

Credits :

Borrows from Pr https://github.com/haskell/haskell-language-server/pull/4449 by @nlander
Mentors : @fendor and @wz1000
Parent PR : #4973 ( This is a cherry-picked subsection of parent )

Thank you !

### DIFF
--- a/ghcide/src/Development/IDE/Core/FileExists.hs
+++ b/ghcide/src/Development/IDE/Core/FileExists.hs
@@ -19,6 +19,7 @@ import           Control.Monad.IO.Class
 import qualified Data.ByteString                       as BS
 import           Data.List                             (partition)
 import           Data.Maybe
+import           Development.IDE.Core.InputPath
 import           Development.IDE.Core.FileStore        hiding (Log, LogShake)
 import qualified Development.IDE.Core.FileStore        as FileStore
 import           Development.IDE.Core.IdeConfiguration
@@ -133,7 +134,7 @@ fromChange FileChangeType_Changed = Nothing
 -------------------------------------------------------------------------------------
 
 -- | Returns True if the file exists
-getFileExists :: NormalizedFilePath -> Action Bool
+getFileExists :: InputPath AllHaskellFiles -> Action Bool
 getFileExists fp = use_ GetFileExists fp
 
 {- Note [Which files should we watch?]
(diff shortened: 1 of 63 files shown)


ghcide/src/Development/IDE/Core/InputPath.hs
```diff
@@ -0,0 +1,115 @@
...
+    deriving newtype (Eq, Hashable, NFData, Show)
+
+-- | Construct an InputPath without checking whether the path belongs to the
+-- requested input class.
+--
+-- This is only for trusted internals that are rehydrating already-typed rule
+-- keys from the Shake database. Normal call sites should use the smart
+-- constructors below.
+unsafeMkInputPath :: NormalizedFilePath -> InputPath i
+unsafeMkInputPath = InputPath
```
vidit-od, line 46: This is only used in Shakes. 
ideally we should make it visible to only Shakes !! 

Plugins should not have access to this.

hls-graph/src/Development/IDE/Graph/Internal/Rules.hs
```diff
@@ -23,6 +23,16 @@ import           Development.IDE.Graph.Internal.Types
 -- | The type mapping between the @key@ or a rule and the resulting @value@.
 type family RuleResult key -- = value
 
+-- | The broad class of input paths a rule is allowed to run on
+data InputClass
```
wz1000, line 27: Let's keep this in `ghcide` and out of `hls-graph` if possible

ghcide/src/Development/IDE/Core/FileExists.hs
```diff
@@ -133,7 +134,7 @@ fromChange FileChangeType_Changed = Nothing
 -------------------------------------------------------------------------------------
 
 -- | Returns True if the file exists
-getFileExists :: NormalizedFilePath -> Action Bool
+getFileExists :: InputPath AllHaskellFiles -> Action Bool
```
wz1000, line 137: rules like filecontents, file existence, file modification time should probably work for all files, not just haskell files.

ghcide/src/Development/IDE/Core/InputPath.hs
```diff
@@ -0,0 +1,115 @@
...
+
+-- | Classify a Stack project configuration file.
+toStackYamlInput :: NormalizedFilePath -> Maybe (InputPath StackYaml)
+toStackYamlInput nfp
+    | takeFileName (fromNormalizedFilePath nfp) == "stack.yaml" = Just (InputPath nfp)
+    | otherwise = Nothing
+
+-- | The sentinel input for rules that do not operate on a real file.
+toNoFileInput :: InputPath NoFile
+toNoFileInput = InputPath emptyFilePath
```
wz1000, line 69: While we are in the area, can we figure out a way to get rid of this hack and not have `emptyFilePath` here?

PR #4984, opened 2026-06-20
Remove @santiweight from CODEOWNERS for hls-refactor-plugin

santiweight:
Heya guys - been too many years since i reviewed code and i'm now mostly not in the haskell community. I could be swayed to just reviewing hlr-refactor-plugin code, but would need to lose notifications for other PRs, as the emails are too noisy and i can't keep track.

let me know what i can do here.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -32,7 +32,7 @@
 /plugins/hls-overloaded-record-dot-plugin @joyfulmantis
 /plugins/hls-pragmas-plugin @eddiemundo
 /plugins/hls-qualify-imported-names-plugin @eddiemundo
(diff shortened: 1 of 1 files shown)


fendor:
Hi @santiweight! That makes sense, thank you for your work so far!
Perhaps our ways will meet again in the future!

If it is alright with you, we can ping you on `hls-refactor-plugin` PR explicitly, then you get only Emails for these pings.

However, it is very unlikely that you will be pinged a lot :)

fendor (approved):

santiweight:
perfect :) happily will review the refactor plugin!

santi

PR #4991, opened 2026-06-26
Add `Unexport ...` to `hls-export-plugin`

crtschin:
Allow unexporting symbols using approximately the same machinery as the `Export ...` action.

- Currently omits the action when CPP is present. 
    This is a bit more finicky than adding exports, so I've not tackled this.
- Uses ghc-exactprint to reprint the export list.

### DIFF
--- a/plugins/hls-export-plugin/test/Main.hs
+++ b/plugins/hls-export-plugin/test/Main.hs
@@ -37,18 +37,31 @@ runExportWith extra hsFile act =
             waitForKickDone
             act doc
 
-executeExportAction :: TextDocumentIdentifier -> Range -> Session ()
-executeExportAction doc range = do
-    actions <- rights . map toEither <$> getCodeActions doc range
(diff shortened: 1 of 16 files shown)


plugins/hls-export-plugin/test/Main.hs
```diff
@@ -318,4 +348,107 @@ main = defaultTestRunner $ testGroup "Export"
                 []     -> liftIO $ assertFailure $
                             "Export `unused` not offered; saw: " <> show (map (^. L.title) actions)
         ]
+
+    , testGroup "Remove: value bindings"
```
fendor: I recommend not using `:` in test names as tasty patterns treat `:` special and it can be tricky to run only tests with `:` in it.

fendor (approved):
LGTM, just a quick sanity check not an in-depth review. Please merge once you are happy.

PR #4997, opened 2026-07-01
Docs: update Kakoune configuration instructions

ficd0:
`kak-lsp` no longer uses `kak-lsp.toml`, instead, all configuration is done inside the Kakoune configuration via the buffer-scoped `lsp_servers` option. The included snippet is taken from `kak-lsp`'s default settings.


fendor (approved):
LGTM, thank you very much!

Issue #3715, opened 2023-07-13
HLS proritization and product development


hasufell:
Since we have a contract with WT and a number of contributors, we should start being a little more product oriented and create a priority list.

E.g. my opinion on the top 3 features missing are:

1.  https://github.com/haskell/haskell-language-server/issues/708
2. https://github.com/haskell/haskell-language-server/issues/709
3. a plugin for adding new modules to a cabal file (briefly discussed with @fendor ... seems possible and would make e.g. hpack obsolete, as well as many odd requests to cabal)

For non-functional priorities, I'd say:

1. performance of HLS (including memory footprint)
2. better/swifter releases (lock-step with GHC)
3. cross-compiler support (new ghcjs... was this discussed? Does it even work?)

So far, I propose to discuss priorities and write them down in the readme or contribution document. Then figure out what the status of each of those is, what blocks them and if we have enough volunteers or funds to execute some of it.

----

@wz1000 @michaelpj @fendor @pepeiborra @david-christiansen @Kleidukos

michaelpj:
I think let's discuss this in our next meeting, can you add it to the agenda?

Rather than the readme or whatever, I think we could either make use of the wiki or a github project to track priority issues.

Realistically, apart from WT I think our contribution bandwidth is still somewhat limited, so I think this may mostly prove relevant for them.

hasufell:
> I think our contribution bandwidth is still somewhat limited

Surely, but excited contributors would probably be most interested in tickets that get them the most fame... so there's additional incentive to tackle these tickets.

fendor:
Re 3., @VeryMilkyJoe is working on it, they have a somewhat working prototype. The first iteration will be up, soonish.

wz1000:
These are certainly some possible directions of work, some like @fendor points out are already underway.

For 1, we already have a summer student with a WIP PR here: https://github.com/haskell/haskell-language-server/pull/3704

I would suggest prioritising more mundane directions of work, such as making the testsuite reliable, GHC compatibility, release management when no other volunteers are available, and performance/memory usage work.

hasufell:
> I would suggest prioritising more mundane directions of work

Well, the idea is not so much about what we can reasonably or easily achieve... but what needs to be done for HLS to be a better LSP tool, **product** wise. And then we figure out IF we can do something about it. And if not, we may need more funding.

Kleidukos:
I fully support this initiative of making the priority list public, it certainly helps build trust in the development process.

[...]

Issue #709, opened 2020-12-29
Type information for arbitrary portions of code


jacg:
`ghcide` gives me information on most alphanumeric symbols (and numeric literals) but there are two features that I'm used to from `dante` and/or `itnero`:

+ type information on operators
+ type information on arbitrary selected expressions

Do these features exist, and have I simply not found them?

If not, any hints on how to implement them?

jneira:
@alanz commented in the issue originally in hls:

> I guess someone needs to actually make a PR for microsoft/language-server-protocol#377

jamesdbrock:
“Type information on arbitrary selected expressions” is the killer feature of any Haskell IDE.

[vim-hdevtools](https://github.com/bitc/vim-hdevtools) has that. Actually, what it has works a little bit differently:

1. Press the <kbd>F1</kbd> key and it will select the token which the cursor is on, and tell you the type of the token.
2. Press the <kbd>F1</kbd> key again, and it will *expand the selection to the next outer subexpression* and tell you the type of that.

... and you can repeat step 2 as many times as you like.

It would be great if we could have that feature again with HLS.

jneira:
As @jamesdbrock gently pointed in #1973, this feature used to work in haskell-ide-engine (see the pr in the vscode extension which added the feature [here](https://github.com/haskell/vscode-haskell/pull/40))
Maybe it worths investigate how was done in hie, althoug it seems it used ghcmod underneath, a deprecated library that should not be used.

Martinsos:
I found this comment explaining a bit on how they implemented this in rust-analyzer: https://github.com/microsoft/language-server-protocol/issues/377#issuecomment-889898650 -> maybe this is useful to guide the possible implementation efforts?

hasufell:
@alanz @mpickering does anyone have knowledge whether this is difficult in ghcide itself (getting type of arbitrary expression), ignoring the issues with LSP and clients?

Maybe we can make some step by step progress.

fendor:
For type information, I bet we can hack something together using HIE files and hiedb.

PR #4999, opened 2026-07-03
ghcide: type of selected expression via haskell/hoverRange custom method (#709)

HCHogan:
Implements #709 (type information for arbitrary selected expressions), one of the two features called out in #3715.

## What

A custom LSP request **`haskell/hoverRange`**, with the same semantics as [rust-analyzer's "Hover Range" extension](https://rust-analyzer.github.io/book/contributing/lsp-extensions.html#hover-range):

```typescript
method: "haskell/hoverRange"
params: { textDocument: TextDocumentIdentifier, range: Range }  // range = current selection
```

The response is a standard `Hover | null` for the **smallest expression that fully contains the range** (the returned `Hover.range` is that expression's span, so clients can highlight it). An empty range behaves exactly like `textDocument/hover`; selections that align with a single identifier get the existing rich hover (docs, definition site, evidence).

A custom method rather than rust-analyzer's wire format (a `Range` in `HoverParams.position`), because `lsp-types` parses hover params strictly — a `Range` there is rejected before reaching HLS. Upstream LSP support (microsoft/language-server-protocol#377) is still open. Clients should treat `MethodNotFound` as "not supported"; advertising `capabilities.experimental` like rust-analyzer does would need a small hook in `lsp` first (possible follow-up).

## How

Two layers:

1. **HieAST**: `pointCommand` is generalised to `rangeCommand` — `selectSmallestContaining` with a non-zero-width span. Position hover is the degenerate case; its behaviour is unchanged (all existing hover tests pass).

2. **Typechecked-source fallback**: the HieAST deliberately omits types of intermediate expression nodes (`skipDesugaring` in `GHC.Iface.Ext.Ast` skips `HsApp`/`OpApp`/if/case/...), e.g. for `combined = negate 3 + 7` the `negate 3` node has `types=[]`. So when the selected node yields nothing, `exprTypeAtSpan` finds the smallest `LHsExpr GhcTc` containing the span (SYB over `tcg_binds`, pruning subtrees whose spans don't contain the target) and computes its type with `lhsExprType`. Same `tcg_binds` traversal pattern as TypeLenses / explicit-record-fields; cost is per-request and proportional to one expression.

## Trade-offs (stated upfront)

- The fallback needs a `TypeCheck` result, i.e. works for files of interest only; for `HieFromDisk` you get whatever the AST has. If GHC ever records these types in the HieAST, the fallback can be deleted with no interface change.
- Custom method ⇒ per-client adoption (I'm happy to do the vscode-haskell side; tested from neovim with ~30 lines of Lua).
- Inside TH quotes there are no typechecked expressions; the nearest typed enclosing expression is the bracket itself (`Q Exp`). Splice interiors report the expansion's type.

## Testing

- New `ghcide-test/exe/HoverRangeTests.hs`: exact selection, multi-line selection, misaligned selection, empty range behaving like hover, null for spans without an enclosing expression / spanning declarations.
- All 82 existing hover-related ghcide tests pass unchanged.
- Docs for client implementors added to `docs/features.md`.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -7,6 +7,7 @@ Many of these are standard LSP features, but a lot of special features are provi
 | --------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
 | [Diagnostics](#diagnostics)                         | `textDocument/publishDiagnostics`                                                                 |
 | [Hovers](#hovers)                                   | `textDocument/hover`                                                                              |
+| [Hover over a selection](#hover-over-a-selection)   | `haskell/hoverRange` (custom method)                                                              |
 | [Signature help](#signature-help)                   | `textDocument/signatureHelp`                                                                      |
 | [Jump to definition](#jump-to-definition)           | `textDocument/definition`                                                                         |
 | [Jump to type definition](#jump-to-type-definition) | `textDocument/typeDefinition`                                                                     |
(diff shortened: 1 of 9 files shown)


fendor:
Hi, thank you for PR!

I think it would have been a good idea to discuss the implementation approach before working on the implementation itself.
The approach you are taking uses `syb` heavily, which is quite inefficient and can take a long time to run on big ASTs.
It would be much better, if we added the information required for point of selection to .hie files directly.

See for example this PR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3344 and https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6078 which aims to bring this information into .hie files.

The main advantage would be to have a single source of truth for all IDE related information.

> The fallback needs a TypeCheck result, i.e. works for files of interest only;

This is quite a big disadvantage, and it might destroy our capability to provide hover information before the session finished initialising. 

> i.e. works for files of interest only; for HieFromDisk you get whatever the AST has.

this trade off seems weird, the editor will not request the type of a range for a non-FOI file, right?

PR #5007, opened 2026-07-08
fix(hls-eval-plugin): hardcode QckChck to eval property tests

### LINKED ISSUE #5003: `prop>` fails to find QuickCheck
### Your environment

- HLS: `2.14.0.0` (from GHCup)
- GHC: `9.14.1`
- cabal-install: `3.16.1.0`
- QuickCheck: `2.18.0.0`
- OS: macOS aarch64
- Editor: VS Code+vscode-haskell

- No

### Steps to reproduce

A Cabal component with:

```cabal
build-depends:
    base,
    QuickCheck
```

And a Haskell file with a property test in an eval-style comment:

```haskell
-- prop> \(x :: Int) -> x == x
```

### Expected behavior

`prop>` should run and return "+++ OK, ..."

### Actual behaviour

Clicking `Evaluate...` (or `Refresh...`) fails to find QuickCheck:

```text
-- prop> \(x :: Int) -> x == x
-- Add QuickCheck to your cabal dependencies to run this property.
```

### Debug information

Inspecting `dist-newstyle/cache/plan.json` shows that QuickCheck is indeed available. However, note the package id. 

```json
{
  ...,
  "install-plan": [
    ...,
    {"type":"configured","id":"QckChck-2.18.0.0-46310d07","pkg-name":"QuickCheck","pkg-version":"2.18.0.0", ...},
  ],
}

```

Eval-plugin comments with QuickCheck also works:

```haskell
-- >>> import Test.QuickCheck
-- >>> output <$> quickCheckWithResult stdArgs (\(x :: Int) -> x == x)
-- "+++ OK, passed 100 tests.\n"
```

### Suspected cause

In `hls-eval-plugin`, `prop>` checks:

https://github.com/haskell/haskell-language-server/blob/3f50b22ecabcdde3a47e820d1250ab94932d50de/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs#L419-L423

where `hasQuickCheck` is defined as:

https://github.com/haskell/haskell-language-server/blob/3f50b22ecabcdde3a47e820d1250ab94932d50de/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs#L596-L597

and `hasPackage` inspects `packageFlags` like this:

https://github.com/haskell/haskell-language-server/blob/3f50b22ecabcdde3a47e820d1250ab94932d50de/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/GHC.hs#L36-L49

This works for `PackageArg "QuickCheck"`, but fails for Cabal/GHC unit ids like `QckChck-2.18.0.0-46310d07`.

### Possible fix

A simple solution would be to hardcode `QckChck` and look it up in the package ids (`hasPackage` is only reference once in the project: for checking QuickCheck). For a more robust solution, perhaps it would be possible to use [lookupPackageName](https://hackage-content.haskell.org/package/ghc-9.14.1/docs/GHC-Unit-State.html#v:lookupPackageName).

### Workaround

Use normal eval-plugin comments with `quickCheckWithResult`:

```haskell
-- >>> import Test.QuickCheck
-- >>> output <$> quickCheckWithResult stdArgs (\(x :: Int) -> x == x)
```

abdivasiyev:
Due to problem with [cabal store](https://github.com/haskell/cabal/issues/7209), currently on MacOS property test evaluation not working. And this pr fixes https://github.com/haskell/haskell-language-server/issues/5003 issue

### DIFF
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
@@ -594,7 +594,7 @@ needsQuickCheck :: [(Section, EvalExpr)] -> Bool
 needsQuickCheck = any (isProperty . snd)
 
 hasQuickCheck :: DynFlags -> Bool
-hasQuickCheck df = hasPackage df "QuickCheck"
+hasQuickCheck df = any (hasPackage df) ["QuickCheck", "QckChck"]
 
 singleLine :: String -> [Text]
 singleLine s = [T.pack s]


fendor:
Thank you for your contribution! Have you tested this fixes #5003 by compiling HLS and running it in your editor?

Since our testsuite isn't run for MacOs, it would be great if you could show a brief video or something showcasing it works now.

abdivasiyev:
Here is the requested video demonstration:

with nix shell installed hls:

https://github.com/user-attachments/assets/3aaae031-b384-4a6f-9f4b-72ace81182b5

with patched hls:

https://github.com/user-attachments/assets/08a51ac3-4f1e-4f47-8371-e6f7c6eb0599

test results:

[test results for hls-eval-plugin](https://asciinema.org/a/VgMAoiRFHLx0Mpi2)

dschrempf:
Thanks for fixing this! Should we add tests for macOS @fendor? It seems like a severe limitation that we do not test on macs (I am not a mac user, but many people are).

fendor:
We could enable the tests on macos, so far, it was determined that macos and linux are close enough that it doesn't matter too much... I think this is the first macos specific bug in HLS that isn't a GHC build issue.

fendor (approved):

PR #5008, opened 2026-07-10
bugfix(ghc optimization option): set optimization level to 0 to eval properly

### LINKED ISSUE #4968: Weird bug that happens with code snippets in comments when a GHC Pragma
https://github.com/user-attachments/assets/68d1c282-a77d-4e97-b170-0f53cd396e7a

### Your environment

Which OS do you use?

macOS

9.14.1 from ghcup
vscode-haskell
2.13.0.0 from ghcup
I dont think so?
### Steps to reproduce

Create a new haskell file, put `{-# OPTIONS_GHC -O1 #-}` at the top, and put any code snippet in a comment, and try to run it

### Expected behaviour

It should show the result of the code in a comment

### Actual behaviour

It shows a weird bug

### Debug information

idk

abdivasiyev:
I couldn't fully understand what was root cause, but setting `updOptLevel 0` dynamic flag is fixed the problem with ghc optimization level

Fixes #4968

### DIFF
--- a/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
+++ b/plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
@@ -81,6 +81,7 @@ import qualified Development.IDE.GHC.Compat.Core              as Compat (Interac
 import qualified Development.IDE.GHC.Compat.Core              as SrcLoc (unLoc)
 import           Development.IDE.Types.HscEnvEq               (HscEnvEq (hscEnv))
 import qualified GHC.LanguageExtensions.Type                  as LangExt (Extension (..))
+import           Development.IDE.Session.Ghc                  (disableOptimisation)
 
 import           Data.List.Extra                              (unsnoc)
 import           Development.IDE.Core.PluginUtils
@@ -303,6 +304,7 @@ initialiseSessionForEval needs_quickcheck st nfp = do
                    . flip xopt_unset  LangExt.MonomorphismRestriction
                    . flip gopt_set    Opt_ImplicitImportQualified
                    . flip gopt_unset  Opt_DiagnosticsShowCaret
+                   . disableOptimisation
                    . setBackend ghciBackend
                    $ (ms_hspp_opts ms) {
                         useColor = Never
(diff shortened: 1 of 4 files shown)


plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
```diff
@@ -303,6 +303,7 @@ initialiseSessionForEval needs_quickcheck st nfp = do
                    . flip xopt_unset  LangExt.MonomorphismRestriction
                    . flip gopt_set    Opt_ImplicitImportQualified
                    . flip gopt_unset  Opt_DiagnosticsShowCaret
+                   . updOptLevel 0
```
fendor:
```suggestion
                   . disableOptimisation
```

from `Development.IDE.Session.Ghc`

Same implementation, but if we use the same function everywhere, then it is easier to update if necessary.

fendor (approved):

fendor:
Thank you for the PR!

Please rebase your PR on top of master.

We do set the optimisation level in https://github.com/haskell/haskell-language-server/blob/master/ghcide/session-loader/Development/IDE/Session/Ghc.hs#L293.
However, the issue is that we apply the optimisation level while summarising the module, in `getModSummaryFromImports`. In theory, that should be the better location to update the `DynFlags`, but at the same time, we do want to honour user choices. 
In this very particular instance, it might be warranted to overwrite any user using `{#- OPTIONS_GHC -O1-#}` to disable the optimisations, since we are not generating code, and if we do for TH, we still disable optimisation. A bit unclear what to do.

plugins/hls-eval-plugin/src/Ide/Plugin/Eval/Handlers.hs
fendor: @abdivasiyev This is still open, I think this would be good to address before merging. What do you say?

abdivasiyev: Sorry for that. I couldnt find time to fix yet. Today or tomorrow I will try to complete

abdivasiyev: @fendor Resolved your comment. Can you review please again?

fendor (approved):
LGTM, thank you!

PR #5011, opened 2026-07-13
Fix stale completions after changing an import alias

BurningLutz:
## Description

Refresh qualified completions when an import alias is changed without restarting HLS.

Previously, changing:

```haskell
import A as CM
```

to:
```haskell
import A as C
```

could leave non-local completions associated with the old CM qualifier. As a result, completions for C did not prioritize the members imported from A.

## Root cause

`GetModSummaryWithoutTimestamps` used `msrFingerprint` as its early-cutoff fingerprint. That fingerprint represents dependency-relevant module information, such as imported
module and package names, but does not include full import declarations.

Changing only an import alias therefore left the fingerprint unchanged, preventing downstream `NonLocalCompletions` from being recomputed.

## Changes

- Add `computeImportsFingerprint` for full parsed import declarations, including aliases and import lists.
- Combine the imports fingerprint with `msrFingerprint` in the `GetModSummaryWithoutTimestamps` early cutoff.
- Add a regression test that changes an alias from CM to C in an open document, and verify that the new alias receives the imported module completions and the previous alias is no longer treated as active.

## Testing

Ran the completion and completion-resolve test selection:
```
cabal test ghcide-tests \
  --test-option=-p \
  --test-option='/completion/' \
  --test-option=--hide-progress
```
All 49 selected tests passed.

### DIFF
--- a/ghcide-test/exe/CompletionTests.hs
+++ b/ghcide-test/exe/CompletionTests.hs
@@ -513,7 +513,37 @@ projectCompletionTests =
         compls <- getCompletions doc (Position 2 12)
         let item = head compls
         liftIO $ do
-          item ^. L.label @?= "anidentifier"
+          item ^. L.label @?= "anidentifier",
+      testSessionEmptyWithCradle "refresh completions after changing an import alias"
+                  "cradle: {direct: {arguments: [\"-Wmissing-signatures\", \"A\", \"B\"]}}" $ do
+        _ <- createDoc "A.hs" "haskell" $ T.unlines
+            [ "module A (anidentifier) where"
+            , "anidentifier = ()"
+            ]
+        _ <- waitForDiagnostics
+        doc <- createDoc "B.hs" "haskell" $ T.unlines
+            [ "module B where"
+            , "import A as CM"
+            , "before = CM."
+            ]
+        _ <- waitForDiagnostics
+        before <- getCompletions doc (Position 2 12)
+
+        changeDoc doc [TextDocumentContentChangeEvent . InR . TextDocumentContentChangeWholeDocument $ T.unlines
+            [ "module B where"
+            , "import A as C"
+            , "after = C."
+            , "stale = CM."
+            ]]
+        void $ waitForTypecheck doc
+        after <- getCompletions doc (Position 2 10)
+        stale <- getCompletions doc (Position 3 11)
+
+        liftIO $ do
+          listToMaybe (map (^. L.label) before) @?= Just "anidentifier"
+          listToMaybe (map (^. L.label) after) @?= Just "anidentifier"
+          assertBool "the old alias is still treated as an active import alias" $
+            listToMaybe (map (^. L.label) stale) /= Just "anidentifier"
     ]
 
 completionDocTests :: [TestTree]
--- a/ghcide/src/Development/IDE/Core/Compile.hs
+++ b/ghcide/src/Development/IDE/Core/Compile.hs
@@ -1215,7 +1215,8 @@ getModSummaryFromImports env fp mContents = do
 
         msrImports = implicit_imports ++ imps
 
-        rn_pkg_qual = renameRawPkgQual (hsc_unit_env ppEnv)
+        unitEnv = hsc_unit_env ppEnv
+        rn_pkg_qual = renameRawPkgQual unitEnv
         rn_imps = fmap (\(pk, lmn@(L _ mn)) -> (rn_pkg_qual mn pk, lmn))
 #if MIN_VERSION_ghc(9,13,0)
         -- In GHC 9.13+, ms_srcimps is just [Located ModuleName] and ms_textual_imps includes ImportLevel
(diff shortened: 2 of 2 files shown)


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -953,7 +953,7 @@ getModSummaryRule displayTHWarning recorder = do
                 let ms = msrModSummary {
                     ms_hspp_buf = error "use GetModSummary instead of GetModSummaryWithoutTimestamps"
                     }
-                    fp = fingerprintToBS msrFingerprint
+                    fp = fingerprintToBS $ Util.fingerprintFingerprints [msrFingerprint, computeImportsFingerprint msrImports]
```
fendor: Why are we only using the `computeImportsFingerprint` here and not in the `GetModSummary` rule?

BurningLutz: > Why are we only using the `computeImportsFingerprint` here and not in the `GetModSummary` rule?

@fendor I've considered that, but the imports fingerprint is a derived value from `msrImports` and only used here, in order not to break anything I chose to compute here. :)

Is it expected to compute in the `GetModSummary` rule?

fendor: It feels non-trivial to see why this works, initially I thought `GetModSummary` might produce the same fingerprint and thus short-circuit the calculation, rendering this fingerprint unsued. However, now I can tell, this is not the case, `GetModSummary` will be recomputed in this situation.

I just don't see why the fingerprinting logic should be different here, I think `computeImportsFingerprint msrImports` should be part of `msrFingerprint` from the start.
If an import changes, more than only completions need to be recomputed, like we need to typecheck the whole module again!

Actually, looking at the implementation of `msrFingerprint`, it looks like the imports are already part of the fingerprint. Perhaps there is a bug, or missing value in there?

BurningLutz: > Perhaps there is a bug, or missing value in there?

AFAIK, import aliases are missing in the `fingerPrintImports` of `msrFingerprint` so changing aliases won't result in a different fingerprint value.

fendor: Right, as we are fingerprinting imports, it makes some sense to me to additionally fingerprint the import aliases

BurningLutz: > it makes some sense to me to additionally fingerprint the import aliases

I agree, with a little concern about the usage of `msrFingerprint`.
At [line 684](https://github.com/haskell/haskell-language-server/blob/765b727b51c0a6f840fd8b7e0538b0d77cc3e612/ghcide/src/Development/IDE/Core/Rules.hs#L684), `msrFingerprint` is used to compute `shallowFingers` which is used to compute a `DependencyInformation`.

So I do think It is a good reason to exclude import aliases from `msrFingerprint`: changing aliases won't result in a different depenceny info.

This is why I wasn't touching `msrFingerprint` in the first place.

fendor: If the module header changes, there is a good expectation that we need to recompile a module. That's not to say dependencies should be recompiled, there should be rules that short-circuit if the public interface doesn't change.

I think `shallowFingers` may benefit from using more fine-grained rules that can short circuit earlier if the public interface of a ModSummary doesn't change in any meaningful way. For example, using `GetLocatedImports` instead of `GetModSummaryWithoutTimestamps` (ignoring some implementation details).

BurningLutz: Just to be sure I understand correctly, are you suggesting:
1. we include import aliases when computing `msrFingerprint` to build a more complete semantic, but then it will lead to `DependencyInformation` invalidation when an import alias changes, so:
2. we make `shallowFingers` take advantage of more fine-grained rules such as `GetLocatedImports` to short circuit more precisely.

If so, I'm happy to work it out.

fendor: Yes, that's what I am suggesting, but I am not 100% sure (2) is going to work as trivially as I am making it out to be. Might need some work to get the `msKey` or `NodeKey_Module`. Maybe return it from `GetLocatedImports`? :thinking:

BurningLutz: OK, let me try it out.

BurningLutz:
I've done the first part, the new `fingerPrintImports` now compute from parsed imports which contains alias information.

The trickier part is the `dependencyInfoForFiles`. I think we can directly compute `shallowFingers` from `nodeKeys` since we [already have it](https://github.com/haskell/haskell-language-server/blob/be99b71e60fce572daa8903a20e66ade8cc1bdfa/ghcide/src/Development/IDE/Core/Rules.hs#L667), but what about the global fingerprint of `dependencyInfoForFiles`? It's currently computed from `msrFingerprint`, after we change `msrFingerprint`'s implementation, `dependencyInfoForFiles` will be cache invalidated by a single alias change. If we still want to deal with it, then things get complicated.

What do you think? @fendor

fendor (approved):
LGTM!

fendor:
Yeah, this looks right to me!

>  but what about the global fingerprint of `dependencyInfoForFiles`?

As mentioned in the discussion above, if we changed `dependencyInfoForFiles` to use `GetLocatedImports`, we could use the `Located ModuleName` for fingerprinting.
Then the rule will short-circuit if the imports themselves haven't changed.
Or we move the early cut-off to `GetLocatedImports` altogether :thinking: 

However, it is also matter of how often this scenario really occurs... I don't think it would be a catastrophe to be slightly pessimistic and dirty rules when it is not needed.

BurningLutz:
> As mentioned in the discussion above, if we changed `dependencyInfoForFiles` to use `GetLocatedImports`, we could use the `Located ModuleName` for fingerprinting. Then the rule will short-circuit if the imports themselves haven't changed. Or we move the early cut-off to `GetLocatedImports` altogether 🤔

That's what I've been struggling with for the past few days. I'm so unfamiliar with the `DependencyInformation` that I cannot easily figure out what should be considered when computing the global fingerprint.

From my naive understanding, the global fingerprint should correctly reflect changes in `DependencyInformation`: if some "essential information" inside it changes, the fingerprint should also change. But there are lots of information involved in computing a `DependencyInformation`: the raw deps info, the shallow fingers, the module graph, the boot id map, etc.

I've been trying very hard to figure out what are the true determinants of a `DependencyInformation`, but so far with no luck.

fendor:
Perhaps it is worth thinking about what we are trying to achieve:

* We want to rerun the `NonLocalCompletions` rule for a module if its imports (including the `as Mod` part) change.
* We have a global modulegraph which should only changes if the module imports change (the `as Mod` part is irrelevant)

As such, we have two ways to fingerprint an import: One time with the `as Mod` part, one time without.

However, if we always fingerprint with the `as Mod` part, we are merely over-approximating, never under-approximating.

The issue arises as the call graph is as follows: `NonLocalCompletions > GetModSummaryWithoutTimestamps > GetModSummary`, where we fail to recompile `NonLocalCompletions` if we under-approximate the import fingerprints.
As `NonLocalCompletions` don't change when the `ModSummary` contents change, we prefer `GetModSummaryWithoutTimestamps` over `GetModSummary` in this instance.

To fix this, we switched to the over-approximation of imports by recording the `as Mod` part of an import in the `msrFingerprint`.
This means now, that essentially we invalidate more rules in the module graph if we change the `as Mod` part of an import.
However, that's more work, and ideally we always perform the minimal amount of work necessary.

Given this summary, I think it would be best if we computed two fingerprints, one fingerprint with the `as Mod` part and one without, so we can choose what we want for imports. I think it is evident, that we will need both variants.

We could introduce a new rule, called `GetImportsForModuleGraph` (name bad on purpose), which defines an early cutoff like `msrFingerprint` for only the import module names and pkg qualifier.

To implement `rawDependencyInformation` and `dependencyInfoForFiles` we might need more information, so 

```haskell
type instance RuleResult GetImportsForModuleGraph = (Module, HscSource, [(Located ModuleName, Maybe ArtifactsLocation)], BS.ByteString)
```

since that's quite a lot of fields, a proper data type might be warranted.

All of that to say, I think this situation does not arise too often, so I think the current situation is fine. I would happy to merge this PR as is and open an issue for follow-up discussion.

BurningLutz:
> All of that to say, I think this situation does not arise too often, so I think the current situation is fine. I would happy to merge this PR as is and open an issue for follow-up discussion.

Yes, I agree. The PR is complete as a fix, and the impact of more work is yet unknown, maybe not too bad.

A follow-up discussion would be useful.

fendor:
@BurningLutz Once you rebase the PR, we will merge then :)

BurningLutz:
> Once you rebase the PR, we will merge then :)

@fendor Done!

PR #5017, opened 2026-07-17
Restrict batch session loads to files owned by the same cradle

### LINKED ISSUE #4431: Haskell Language Server passes file target for `direct` cradle to the `cabal` cradle
More generally, if a project has a multi cradle setup, we need to make sure we respect cradle type resolution.

Reproducer:
* cabal 3.12.1.0
* sessionLoading: multiComponent
* HLS: master 1ec65ee176aa34066abec700966900ee3a18b405

Logs:
```
2024-10-15T14:55:15.565298Z | Debug | Session loading result: Left [CradleError {cradleErrorDependencies = ["haskell-language-server.cabal","cabal.project","cabal.project.local"], cradleErrorExitCode = ExitFailure 1, cradleErrorStderr = ["Failed to run [\"cabal\",\"v2-repl\",\"--keep-temp-files\",\"--enable-multi-repl\",\"/home/hugin/Documents/haskell/hls/ghcide/src/Development/IDE/Spans/AtPoint.hs\",\"/home/hugin/Documents/haskell/hls/ghcide/test/data/hover/GotoImplementation.hs\",\"/home/hugin/Documents/haskell/hls/ghcide/test/exe/FindImplementationAndHoverTests.hs\"] in directory \"/home/hugin/Documents/haskell/hls\". Consult the logs for full command and error.","Failed command: cabal --builddir=/home/hugin/.cache/hie-bios/dist-hls-036ebb56a4cc135651b7510a90f3136d v2-repl --with-compiler /home/hugin/.cache/hie-bios/wrapper-b54f81dea4c0e6d1626911c526bc4e36 --with-hc-pkg /home/hugin/.cache/hie-bios/ghc-pkg-99f55999495b863a61d4f15d989eaab4 --keep-temp-files --enable-multi-repl /home/hugin/Documents/haskell/hls/ghcide/src/Development/IDE/Spans/AtPoint.hs /home/hugin/Documents/haskell/hls/ghcide/test/data/hover/GotoImplementation.hs /home/hugin/Documents/haskell/hls/ghcide/test/exe/FindImplementationAndHoverTests.hs","","Error: [Cabal-7121]\nFailed extracting script block: `{- cabal:` start marker not found\n\n","","Process Environment:","HIE_BIOS_GHC: /home/hugin/.ghcup/ghc/9.6.5/lib/ghc-9.6.5/bin/ghc-9.6.5","HIE_BIOS_GHC_ARGS: -B/home/hugin/.ghcup/ghc/9.6.5/lib/ghc-9.6.5/lib"]}]
```

the issue boils down to this cabal error:
```sh
$ cabal v2-repl --enable-multi-repl /home/hugin/Documents/haskell/hls/ghcide/src/Development/IDE/Spans/AtPoint.hs /home/hugin/Documents/haskell/hls/ghcide/test/data/hover/GotoHover.hs
Error: [Cabal-7121]
Failed extracting script block: `{- cabal:` start marker not found
```

`GotoHover` is part of a direct cradle with the contents:
```yaml
cradle: {direct: {arguments: ["Foo", "Bar", "GotoHover", "RecordDotSyntax"]}}
```

simonhorlick:
I keep running into this issue so I've attempted to fix it. 

Essentially:

* Launch VSCode and open the haskell-language-server repo
* Open ghcide-test/data/hover/GotoHover.hs
* Wait for everything to settle
* Open ghcide/src/Development/Ide/Spans/AtPoint.hs

expect: AtPoint.hs should load without error
actual: Loading the module '/Users/simon/src/github.com/haskell/haskell-language-server/ghcide/src/Development/IDE/Spans/AtPoint.hs' failed.

It may not be listed in your .cabal file!
Perhaps you need to add `AtPoint` to other-modules or exposed-modules.
...

Please take a look and let me know if the patch is suitable.

AI disclaimer: This patch was made with assistance from Claude.

Fixes #4431

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session.hs
+++ b/ghcide/session-loader/Development/IDE/Session.hs
@@ -539,6 +539,11 @@ insertFileMapping :: SessionState -> Maybe FilePath -> NormalizedFilePath -> STM
 insertFileMapping state hieYaml ncfp =
   STM.insert hieYaml ncfp (filesMap state)
 
+-- | Same as 'insertFileMapping', but never overwrites an existing value.
+insertFileMappingIfMissing :: SessionState -> Maybe FilePath -> NormalizedFilePath -> STM ()
+insertFileMappingIfMissing state hieYaml ncfp =
+  STM.focus (Focus.alter (<|> Just hieYaml)) ncfp (filesMap state)
+
 -- | Remove a file from the pending file set
 removeFromPending :: SessionState -> FilePath -> STM ()
 removeFromPending state file =
@@ -711,14 +716,14 @@ loadSessionWithOptions recorder SessionLoadingOptions{..} rootDir que = do
     let absolutePathsCradleDeps (eq, deps) = (eq, fmap toAbsolutePath $ Map.keys deps)
     returnWithVersion $ \file -> do
       let absFile = toAbsolutePath file
-      absolutePathsCradleDeps <$> lookupOrWaitCache recorder sessionState absFile
+      absolutePathsCradleDeps <$> lookupOrWaitCache recorder sessionState cradleLoc absFile
 
 -- | Given a file, this function will return the HscEnv and the dependencies
 -- it would look up the cache first, if the cache is not available, it would
 -- submit a request to the getOptionsLoop to get the options for the file
 -- and wait until the options are available
-lookupOrWaitCache :: Recorder (WithPriority Log) -> SessionState -> FilePath -> IO (IdeResult HscEnvEq, DependencyInfo)
-lookupOrWaitCache recorder sessionState absFile = do
+lookupOrWaitCache :: Recorder (WithPriority Log) -> SessionState -> (FilePath -> IO (Maybe FilePath)) -> FilePath -> IO (IdeResult HscEnvEq, DependencyInfo)
+lookupOrWaitCache recorder sessionState cradleLoc absFile = do
   let ncfp = toNormalizedFilePath' absFile
   cacheResult <- maybeM
     (return Nothing)
@@ -736,8 +741,14 @@ lookupOrWaitCache recorder sessionState absFile = do
     Just r -> return r
     Nothing -> do
       -- if not ok, we need to reload the session
-      atomically $ addToPending sessionState absFile
-      lookupOrWaitCache recorder sessionState absFile
+      hieYaml <- cradleLoc absFile
+      atomically $ do
+        -- Insert the mapping into filesMap so the cradle is known up-front.
+        -- This ensures we are able to batch requests belonging to the same
+        -- cradle.
+        insertFileMappingIfMissing sessionState hieYaml ncfp
+        addToPending sessionState absFile
+      lookupOrWaitCache recorder sessionState cradleLoc absFile
 
 checkInCache :: SessionState -> NormalizedFilePath -> STM (Maybe (IdeResult HscEnvEq, DependencyInfo))
 checkInCache sessionState ncfp = runMaybeT $ do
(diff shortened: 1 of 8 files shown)


ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -573,18 +575,24 @@ handleSingleFileProcessingError state hieYaml file diags extraDepFiles = liftIO
...
-      if cfp `Set.member` errorFiles
-        then Set.empty
-        -- remove error files from pending files since error loading need to load one by one
-        else (Set.delete cfp $ pendingFiles `Set.difference` errorFiles) <> old_files
+  let candidates =
+        if cfp `Set.member` errorFiles
+          then Set.empty
+          -- remove error files from pending files since error loading need to load one by one
+          else (Set.delete cfp $ pendingFiles `Set.difference` errorFiles) <> old_files
+  filterM (fmap (== hieYaml) . findHieYamlForTarget (filesMap state)) (Set.toList candidates)
```
fendor: I think checking `filesMap` only is the correct thing to do. This way, we look up the `hie.yaml` on disk every time via `findCradle sessionLoadingOptions file`. Perhaps not a huge issue, but we should avoid repeated disk lookups whenever possible.

For reference
```haskell
findHieYamlForTarget :: FilesMap -> FilePath -> SessionM (Maybe FilePath)
findHieYamlForTarget filesMapping file = do
  let ncfp = toNormalizedFilePath' file
  cachedHieYamlLocation <- join <$> liftIO (atomically (STM.lookup ncfp filesMapping))
  sessionLoadingOptions <- asks sessionLoadingOptions
  hieYaml <- liftIO $ findCradle sessionLoadingOptions file
  pure $ cachedHieYamlLocation <|> hieYaml
```

fendor (changes requested):
This looks reasonable to me, also thank you for the test!

Waiting for @soulomoon to confirm this is the right idea.

ghcide/session-loader/Development/IDE/Session.hs
fendor: Actually it might be necessary to use `findHieYamlForTarget` here, but I would still like to avoid this disk access :thinking:

simonhorlick: I think it should be fine to only consult the filesMap. Anything not in the filesMap would just be excluded from the batch and loaded separately, which seems like a better way to do it.

fendor: I think that could deteriorate initial startup time immensely, if you open a project and have 10 files open, will they now be loaded 1 by 1? @soulomoon might correct me that this is not the location where batching of initial files is happening

soulomoon: Yes, getExtraFilesToLoad is the key path here. And filesMap is empty in the beginning. Only checking filesMap might not be enough.
Maybe we should add a regression test covering the initial multi-file loading scenario to make sure future changes don't break this behavior.

soulomoon:
Another idea to speed up the initial startup slowdown. Along with this patch, we can cache the correct hie.yaml for the file into filesMap right before we put it into pendings files. Then no same hie.yaml pending files would be incorrectly filtered out just because we do not cache its hie.yaml yet.

simonhorlick:
@fendor @soulomoon Appreciate if you could take another look. I've addressed the comments you raised, let me know if you have any more thoughts. I added `waitForTypeChecksCountingCradleLoads` for testing which introduces a small amount of duplication, but I thought it was clearer than modifying `waitForTypeChecksBatched`.

fendor (approved):
LGTM, thank you!

I reviewed it and now I understand how this works :)

PR #4977, opened 2026-06-16
WIP: Run `ghcide-tests` in parallel

crtschin:
Trying something out.

PR #5018, opened 2026-07-18
Isolate per-test caches via explicit args instead of `XDG_CACHE_HOME`

crtschin:
Split off from: https://github.com/haskell/haskell-language-server/pull/4977. Makes the following changes:

- Adds `getCacheDirsIn` and `getHieDbLocIn` that allows using different cache directories than `XDG_CACHE_HOME`.
- Gives each test its own cache directory and removes the need for `lockForTempDirs`.

Contains no user-visible change.

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -450,9 +450,20 @@ setCacheDirs recorder CacheDirs{..} dflags = do
 -- keeping the path short and clean.
 getCacheDirsDefault :: String -> Maybe B.ByteString -> [String] -> IO CacheDirs
 getCacheDirsDefault prefix mFirstHash opts = do
-    dir <- Just <$> getXdgDirectory XdgCache (cacheDir </> prefix' ++ "-" ++ opts_hash)
-    return $ CacheDirs dir dir dir
+    base <- getXdgDirectory XdgCache cacheDir
+    pure $ cacheDirsUnder base prefix mFirstHash opts
+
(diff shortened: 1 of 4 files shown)


ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -450,9 +450,20 @@ setCacheDirs recorder CacheDirs{..} dflags = do
...
 getCacheDirsDefault :: String -> Maybe B.ByteString -> [String] -> IO CacheDirs
 getCacheDirsDefault prefix mFirstHash opts = do
-    dir <- Just <$> getXdgDirectory XdgCache (cacheDir </> prefix' ++ "-" ++ opts_hash)
-    return $ CacheDirs dir dir dir
+    base <- getXdgDirectory XdgCache cacheDir
+    pure $ cacheDirsUnder base prefix mFirstHash opts
+
+-- | Like 'getCacheDirsDefault', but roots the cache under @base@ instead of
+-- 'XdgCache', so callers can isolate a cache without touching @XDG_CACHE_HOME@.
+getCacheDirsIn :: FilePath -> String -> Maybe B.ByteString -> [String] -> IO CacheDirs
```
fendor: Any reason to keep this in `IO`?

fendor (approved):
LGTM, one nitpick and then good to go imo.

ghcide/session-loader/Development/IDE/Session/Ghc.hs
crtschin: Nope, nice catch

PR #5020, opened 2026-07-20
Update cabal-version to 3.14

Aster89:
Attempt at updating `cabal-version` to 3.14.

The real motivation is to have the `extra-files` stanza in which to move most files currently listed under `extra-source-files`; doing so would improve the developer experience by preventing `cabal run`/`build` from rebuilding the project when only `*/testdata/*` files are changed.

### DIFF
--- a/haskell-language-server.cabal
+++ b/haskell-language-server.cabal
@@ -1,4 +1,4 @@
-cabal-version:      3.4
+cabal-version:      3.14
 category:           Development
 name:               haskell-language-server
 version:            2.14.0.0
@@ -15,9 +15,10 @@ license:            Apache-2.0
 license-file:       LICENSE
 build-type:         Simple
 tested-with:        GHC == {9.14.1, 9.12.2, 9.10.3, 9.8.4, 9.6.7}
-extra-source-files:
+extra-doc-files:
   README.md
   ChangeLog.md
+extra-files:
   test/testdata/**/*.project
   test/testdata/**/*.cabal
   test/testdata/**/*.yaml


fendor:
Looks like stackage lts23 doesn't support building with Cabal-3.14.

The lts23 is quite old, so I think we can argue that we should upgrade the snapshot. We can probably drop that lts version or upgrade it to https://www.stackage.org/lts-23.28, to bump to GHC 9.8. GHC 9.6 will be dropped after the next release, so I think it is fine to not test it in CI with stack any more.

The failures with cabal are caused by old ormolu versions. I don't think we can simply force a newer `ormolu` version. In https://github.com/haskell/haskell-language-server/blob/master/haskell-language-server.cabal#L1454C1-L1454C63, we see 

```
    , ormolu          ^>=0.5.3 || ^>= 0.6 || ^>= 0.7 || ^>=0.8
```

I doubt we actually support `^>=0.5.3 || ^>= 0.6` but https://hackage.haskell.org/package/ormolu-0.7.7.0 only works with [Cabal-syntax](https://hackage.haskell.org/package/Cabal-syntax) (>=3.12 && <3.13), and [ghc-lib-parser](https://hackage.haskell.org/package/ghc-lib-parser) (>=9.10 && <9.11). So that sounds like an issue.
To be sure, I am unsure, but it looks like the GHC 9.10 CI might have succeeded, so maybe there is some more wiggle room.

TL;DR, this PR can't be merged right away until we manage to upgrade the ormolu version to 0.8 which only works for GHC >=9.12.

Aster89:
Can't be done yet as Debian ships with Cabal 3.10, and we don't want to break Debian.

PR #5025, opened 2026-07-23
Typed rules

vidit-od:
This PR introduces typed inputs for HLS/ghcide rules.

HLS already associates every rule key with its result through `RuleResult`, but all rules previously accepted the same input type: `NormalizedFilePath`.  Theis PR adds an open type family that associates each rule with the kind of input it supports:

```haskell
type family RuleInput k

type instance RuleInput TypeCheck = ProjectHaskellInput
type instance RuleResult TypeCheck = TcModuleResult
```

Rule inputs are now part of the rule's contract, just like rule results.

## Why this is needed

A raw `NormalizedFilePath` does not tell us what a path actually represent. These distinctions matter. For example, `TypeCheck` and `GhcSession` require a project component and session, while file-content rules can work on many file
types.  Typed inputs make these assumptions explicit, reject incorrect rule/input pairings earlier, and provide the foundation needed to support dependency sources without treating them as project modules.

## How it works

The new `Development.IDE.Core.RuleInput` module defines this input hierarchy:

```text
SomeInput
|-- NoInput
`-- SomeFileInput
    |-- SomeHaskellInput
    |   |-- ProjectHaskellInput
    |   `-- NonProjectHaskellInput
    |-- CabalInput
    `-- arbitrary NormalizedFilePath
```
Internally, Shake still needs one heterogeneous key representation. `SomeInput`
therefore existentially stores typed inputs inside `Q`, while `InputFingerprint`
provides consistent equality and hashing.

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
@@ -9,6 +9,7 @@ import           Data.List
 import           Data.List.Extra                   (split)
 import           Data.Maybe
 import qualified Data.Text                         as T
+import           Development.IDE.Core.RuleInput
 import           Development.IDE.Types.Diagnostics
 import           Development.IDE.Types.Location
 import           GHC.Generics
@@ -40,8 +41,8 @@ data UnknownModuleDetails =
   the cradle error occurred (of the file we attempted to load).
   Depicts the cradle error in a user-friendly way.
 -}
-renderCradleError :: CradleError -> Cradle a -> NormalizedFilePath -> FileDiagnostic
-renderCradleError cradleError cradle nfp =
+renderCradleError :: CradleError -> Cradle a -> ProjectHaskellInput -> FileDiagnostic
+renderCradleError cradleError cradle input =
   let noDetails =
         ideErrorWithSource (Just "cradle") (Just DiagnosticSeverity_Error) nfp (T.unlines $ map T.pack userFriendlyMessage) Nothing
   in
@@ -54,6 +55,7 @@ renderCradleError cradleError cradle nfp =
             }
      else noDetails
   where
+    nfp = inputFilePath input
     ms = cradleErrorStderr cradleError
 
     absDeps = fmap (cradleRootDir cradle </>) (cradleErrorDependencies cradleError)
--- a/ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
@@ -9,8 +9,7 @@ import           Data.List
 import           Data.List.Extra                   (split)
 import           Data.Maybe
 import qualified Data.Text                         as T
-import           Development.IDE.Core.RuleInput (IsFileInput (inputFilePath),
-                                       ProjectHaskellInput)
+import           Development.IDE.Core.RuleInput
 import           Development.IDE.Types.Diagnostics
 import           Development.IDE.Types.Location
 import           GHC.Generics
--- a/ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
@@ -9,6 +9,8 @@ import           Data.List
 import           Data.List.Extra                   (split)
 import           Data.Maybe
 import qualified Data.Text                         as T
+import           Development.IDE.Core.RuleInput (IsFileInput (inputFilePath),
+                                       ProjectHaskellInput)
 import           Development.IDE.Types.Diagnostics
 import           Development.IDE.Types.Location
 import           GHC.Generics
@@ -40,8 +42,8 @@ data UnknownModuleDetails =
   the cradle error occurred (of the file we attempted to load).
   Depicts the cradle error in a user-friendly way.
 -}
-renderCradleError :: CradleError -> Cradle a -> NormalizedFilePath -> FileDiagnostic
-renderCradleError cradleError cradle nfp =
+renderCradleError :: CradleError -> Cradle a -> ProjectHaskellInput -> FileDiagnostic
+renderCradleError cradleError cradle input =
   let noDetails =
         ideErrorWithSource (Just "cradle") (Just DiagnosticSeverity_Error) nfp (T.unlines $ map T.pack userFriendlyMessage) Nothing
   in
@@ -54,6 +56,7 @@ renderCradleError cradleError cradle nfp =
             }
      else noDetails
   where
+    nfp = inputFilePath input
     ms = cradleErrorStderr cradleError
 
     absDeps = fmap (cradleRootDir cradle </>) (cradleErrorDependencies cradleError)
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -512,7 +512,7 @@ fromTargetId is exts (GHC.TargetModule modName) env dep = do
               , i <- is
               , boot <- ["", "-boot"]
               ]
-    let locs = fmap (ProjectHaskellInput . toNormalizedFilePath') fps
+    let locs = mapMaybe (toProjectHaskellInput . toNormalizedFilePath') fps
     return [TargetDetails (TargetModule modName) env dep locs]
 -- For a 'TargetFile' we consider all the possible module names
 fromTargetId _ _ (GHC.TargetFile f _) env deps = do
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -15,8 +15,7 @@ import qualified Data.List.NonEmpty                 as NE
 import qualified Data.Map.Strict                    as Map
 import           Data.Maybe
 import qualified Data.Text                          as T
-import           Development.IDE.Core.RuleInput      (IsFileInput (inputFilePath),
-                                                   ProjectHaskellInput (..))
+import           Development.IDE.Core.RuleInput
 import           Development.IDE.Core.Shake         hiding (Log, knownTargets,
                                                      withHieDb)
 import qualified Development.IDE.GHC.Compat         as Compat
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -15,6 +15,7 @@ import qualified Data.List.NonEmpty                 as NE
 import qualified Data.Map.Strict                    as Map
 import           Data.Maybe
 import qualified Data.Text                          as T
+import           Development.IDE.Core.RuleInput
 import           Development.IDE.Core.Shake         hiding (Log, knownTargets,
                                                      withHieDb)
 import qualified Development.IDE.GHC.Compat         as Compat
@@ -101,7 +102,7 @@ data RawComponentInfo = RawComponentInfo
   -- | All targets of this components.
   , rawComponentTargets        :: [GHC.Target]
   -- | Filepath which caused the creation of this component
-  , rawComponentFP             :: NormalizedFilePath
+  , rawComponentFP             :: ProjectHaskellInput
   -- | Component Options used to load the component.
   , rawComponentCOptions       :: ComponentOptions
   -- | Maps cradle dependencies, such as `stack.yaml`, or `.cabal` file
@@ -120,7 +121,7 @@ data ComponentInfo = ComponentInfo
   -- | All targets of this components.
   , componentTargets        :: [GHC.Target]
   -- | Filepath which caused the creation of this component
-  , componentFP             :: NormalizedFilePath
+  , componentFP             :: ProjectHaskellInput
   -- | Component Options used to load the component.
   , componentCOptions       :: ComponentOptions
   -- | Maps cradle dependencies, such as `stack.yaml`, or `.cabal` file
@@ -145,7 +146,7 @@ addUnit unit_str = liftEwM $ do
 newComponentCache
          :: Recorder (WithPriority Log)
          -> [String]           -- ^ File extensions to consider
-         -> NormalizedFilePath -- ^ Path to file that caused the creation of this component
+         -> ProjectHaskellInput -- ^ Path to file that caused the creation of this component
          -> HscEnv             -- ^ An empty HscEnv
          -> [ComponentInfo]    -- ^ New components to be loaded
          -> [ComponentInfo]    -- ^ old, already existing components
@@ -172,7 +173,7 @@ newComponentCache recorder exts cfp hsc_env old_cis new_cis = do
 #endif
         closure_err_to_multi_err err =
             ideErrorWithSource
-                (Just "cradle") (Just DiagnosticSeverity_Warning) cfp
+                (Just "cradle") (Just DiagnosticSeverity_Warning) (inputFilePath cfp)
                 (T.pack (Compat.printWithoutUniques (singleMessage err)))
                 (Just (fmap GhcDriverMessage err))
         multi_errs = map closure_err_to_multi_err closure_errs
@@ -200,7 +201,7 @@ newComponentCache recorder exts cfp hsc_env old_cis new_cis = do
 
     -- A session representative file of this load. See Note [Session representatives]
     -- Not cfp which may be an error target.
-    let repr = fromMaybe cfp $ listToMaybe
+    let repr = fromMaybe (inputFilePath cfp) $ listToMaybe
           [ loc
           | ci <- Map.elems cis
           , t <- componentTargets ci
@@ -228,7 +229,7 @@ newComponentCache recorder exts cfp hsc_env old_cis new_cis = do
 -- | Throws if package flags are unsatisfiable
 setOptions :: GhcMonad m
     => OptHaddockParse
-    -> NormalizedFilePath
+    -> ProjectHaskellInput
     -> ComponentOptions
     -> DynFlags
     -> FilePath -- ^ root dir, see Note [Root Directory]
@@ -256,7 +257,7 @@ setOptions haddockOpt cfp (ComponentOptions theOpts compRoot _) dflags rootDir =
         --
         -- If we don't end up with a target for the current file in the end, then
         -- we will report it as an error for that file
-        let abs_fp = toAbsolute rootDir (fromNormalizedFilePath cfp)
+        let abs_fp = toAbsolute rootDir (fromNormalizedFilePath (inputFilePath cfp))
         let special_target = Compat.mkSimpleTarget df abs_fp
         pure $ HomeUnitConfig df (special_target : targets) mHash :| []
     where
@@ -315,7 +316,7 @@ addComponentInfo ::
   (String -> Maybe B.ByteString -> [String] -> IO CacheDirs) ->
   DependencyInfo ->
   NonEmpty HomeUnitConfig->
-  (Maybe FilePath, NormalizedFilePath, ComponentOptions) ->
+  (Maybe FilePath, ProjectHaskellInput, ComponentOptions) ->
   Map.Map (Maybe FilePath) [RawComponentInfo] ->
   m (Map.Map (Maybe FilePath) [RawComponentInfo], ([ComponentInfo], [ComponentInfo]))
 addComponentInfo recorder getCacheDirs dep_info newDynFlags (hieYaml, cfp, opts) m = do
@@ -508,7 +509,7 @@ data TargetDetails = TargetDetails
       targetTarget    :: !Target,
       targetEnv       :: !(IdeResult HscEnvEq),
       targetDepends   :: !DependencyInfo,
-      targetLocations :: ![NormalizedFilePath]
+      targetLocations :: ![ProjectHaskellInput]
   }
 
 -- | Candidate locations of a target, in search order.
@@ -537,12 +538,22 @@ fromTargetId :: [FilePath]          -- ^ import paths
              -> IdeResult HscEnvEq
              -> DependencyInfo
              -> IO [TargetDetails]
-fromTargetId is exts tid env dep =
-    return [TargetDetails target env dep (targetIdLocations is exts tid)]
-  where
-    target = case tid of
-      GHC.TargetModule modName -> TargetModule modName
-      GHC.TargetFile f _       -> TargetFile (toNormalizedFilePath' f)
+-- For a target module we consider all the import paths
+fromTargetId is exts (GHC.TargetModule modName) env dep = do
+    let fps = [i </> moduleNameSlashes modName -<.> ext <> boot
+              | ext <- exts
+              , i <- is
+              , boot <- ["", "-boot"]
+              ]
+    let locs = mapMaybe (toProjectHaskellInput . toNormalizedFilePath') fps
+    return [TargetDetails (TargetModule modName) env dep locs]
+-- For a 'TargetFile' we consider all the possible module names
+fromTargetId _ _ (GHC.TargetFile f _) env deps = do
+    let nf = ProjectHaskellInput (toNormalizedFilePath' f)
+    let other
+          | isSuffixOf "-boot" f = ProjectHaskellInput (toNormalizedFilePath' (L.dropEnd 5 (fromNormalizedFilePath (inputFilePath nf))))
+          | otherwise = ProjectHaskellInput (toNormalizedFilePath' (fromNormalizedFilePath (inputFilePath nf) ++ "-boot"))
+    return [TargetDetails (TargetFile nf) env deps [nf, other]]
 
 -- ----------------------------------------------------------------------------
 -- Backwards compatibility
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -15,6 +15,8 @@ import qualified Data.List.NonEmpty                 as NE
 import qualified Data.Map.Strict                    as Map
 import           Data.Maybe
 import qualified Data.Text                          as T
+import           Development.IDE.Core.RuleInput      (IsFileInput (inputFilePath),
+                                                   ProjectHaskellInput (..))
 import           Development.IDE.Core.Shake         hiding (Log, knownTargets,
                                                      withHieDb)
 import qualified Development.IDE.GHC.Compat         as Compat
@@ -101,7 +103,7 @@ data RawComponentInfo = RawComponentInfo
   -- | All targets of this components.
   , rawComponentTargets        :: [GHC.Target]
   -- | Filepath which caused the creation of this component
-  , rawComponentFP             :: NormalizedFilePath
+  , rawComponentFP             :: ProjectHaskellInput
   -- | Component Options used to load the component.
   , rawComponentCOptions       :: ComponentOptions
   -- | Maps cradle dependencies, such as `stack.yaml`, or `.cabal` file
@@ -120,7 +122,7 @@ data ComponentInfo = ComponentInfo
   -- | All targets of this components.
   , componentTargets        :: [GHC.Target]
   -- | Filepath which caused the creation of this component
-  , componentFP             :: NormalizedFilePath
+  , componentFP             :: ProjectHaskellInput
   -- | Component Options used to load the component.
   , componentCOptions       :: ComponentOptions
   -- | Maps cradle dependencies, such as `stack.yaml`, or `.cabal` file
@@ -145,7 +147,7 @@ addUnit unit_str = liftEwM $ do
 newComponentCache
          :: Recorder (WithPriority Log)
          -> [String]           -- ^ File extensions to consider
-         -> NormalizedFilePath -- ^ Path to file that caused the creation of this component
+         -> ProjectHaskellInput -- ^ Path to file that caused the creation of this component
          -> HscEnv             -- ^ An empty HscEnv
          -> [ComponentInfo]    -- ^ New components to be loaded
          -> [ComponentInfo]    -- ^ old, already existing components
@@ -172,7 +174,7 @@ newComponentCache recorder exts _cfp hsc_env old_cis new_cis = do
 #endif
         closure_err_to_multi_err err =
             ideErrorWithSource
-                (Just "cradle") (Just DiagnosticSeverity_Warning) _cfp
+                (Just "cradle") (Just DiagnosticSeverity_Warning) (inputFilePath _cfp)
                 (T.pack (Compat.printWithoutUniques (singleMessage err)))
                 (Just (fmap GhcDriverMessage err))
         multi_errs = map closure_err_to_multi_err closure_errs
@@ -219,7 +221,7 @@ newComponentCache recorder exts _cfp hsc_env old_cis new_cis = do
 -- | Throws if package flags are unsatisfiable
 setOptions :: GhcMonad m
     => OptHaddockParse
-    -> NormalizedFilePath
+    -> ProjectHaskellInput
     -> ComponentOptions
     -> DynFlags
     -> FilePath -- ^ root dir, see Note [Root Directory]
@@ -247,7 +249,7 @@ setOptions haddockOpt cfp (ComponentOptions theOpts compRoot _) dflags rootDir =
         --
         -- If we don't end up with a target for the current file in the end, then
         -- we will report it as an error for that file
-        let abs_fp = toAbsolute rootDir (fromNormalizedFilePath cfp)
+        let abs_fp = toAbsolute rootDir (fromNormalizedFilePath (inputFilePath cfp))
         let special_target = Compat.mkSimpleTarget df abs_fp
         pure $ HomeUnitConfig df (special_target : targets) mHash :| []
     where
@@ -302,7 +304,7 @@ addComponentInfo ::
   (String -> Maybe B.ByteString -> [String] -> IO CacheDirs) ->
   DependencyInfo ->
   NonEmpty HomeUnitConfig->
-  (Maybe FilePath, NormalizedFilePath, ComponentOptions) ->
+  (Maybe FilePath, ProjectHaskellInput, ComponentOptions) ->
   Map.Map (Maybe FilePath) [RawComponentInfo] ->
   m (Map.Map (Maybe FilePath) [RawComponentInfo], ([ComponentInfo], [ComponentInfo]))
 addComponentInfo recorder getCacheDirs dep_info newDynFlags (hieYaml, cfp, opts) m = do
@@ -495,7 +497,7 @@ data TargetDetails = TargetDetails
       targetTarget    :: !Target,
       targetEnv       :: !(IdeResult HscEnvEq),
       targetDepends   :: !DependencyInfo,
-      targetLocations :: ![NormalizedFilePath]
+      targetLocations :: ![ProjectHaskellInput]
   }
 
 fromTargetId :: [FilePath]          -- ^ import paths
@@ -511,14 +513,14 @@ fromTargetId is exts (GHC.TargetModule modName) env dep = do
               , i <- is
               , boot <- ["", "-boot"]
               ]
-    let locs = fmap toNormalizedFilePath' fps
+    let locs = fmap (ProjectHaskellInput . toNormalizedFilePath') fps
     return [TargetDetails (TargetModule modName) env dep locs]
 -- For a 'TargetFile' we consider all the possible module names
 fromTargetId _ _ (GHC.TargetFile f _) env deps = do
-    let nf = toNormalizedFilePath' f
+    let nf = ProjectHaskellInput (toNormalizedFilePath' f)
     let other
-          | "-boot" `isSuffixOf` f = toNormalizedFilePath' (L.dropEnd 5 $ fromNormalizedFilePath nf)
-          | otherwise = toNormalizedFilePath' (fromNormalizedFilePath nf ++ "-boot")
+          | isSuffixOf "-boot" f = ProjectHaskellInput (toNormalizedFilePath' (L.dropEnd 5 (fromNormalizedFilePath (inputFilePath nf))))
+          | otherwise = ProjectHaskellInput (toNormalizedFilePath' (fromNormalizedFilePath (inputFilePath nf) ++ "-boot"))
     return [TargetDetails (TargetFile nf) env deps [nf, other]]
 
 -- ----------------------------------------------------------------------------
--- a/ghcide/src/Development/IDE/Core/Actions.hs
+++ b/ghcide/src/Development/IDE/Core/Actions.hs
@@ -21,6 +21,11 @@ import           Development.IDE.Core.LookupMod       (lookupMod)
 import           Development.IDE.Core.OfInterest
 import           Development.IDE.Core.PluginUtils
 import           Development.IDE.Core.PositionMapping
+import           Development.IDE.Core.RuleInput       (IsFileInput (inputFilePath),
+                                                       SomeFileInput,
+                                                       SomeHaskellInput,
+                                                       reclassifyInput,
+                                                       toSomeHaskellInput)
 import           Development.IDE.Core.RuleTypes
 import           Development.IDE.Core.Service
 import           Development.IDE.Core.Shake
@@ -45,17 +50,17 @@ import           Language.LSP.Protocol.Types          (DocumentHighlight (..),
 -- block waiting for the rule to be properly computed.
 
 -- | Try to get hover text for the name under point.
-getAtPoint :: NormalizedFilePath -> Position -> IdeAction (Maybe (Maybe Range, [T.Text]))
+getAtPoint :: SomeHaskellInput -> Position -> IdeAction (Maybe (Maybe Range, [T.Text]))
 getAtPoint file pos = runMaybeT $ do
   ide <- ask
   opts <- liftIO $ getIdeOptionsIO ide
 
   (hf, mapping) <- useWithStaleFastMT GetHieAst file
   shakeExtras <- lift askShake
-
-  env <- hscEnv . fst <$> useWithStaleFastMT GhcSession file
-  modSummary <- fst <$> useWithStaleFastMT GetModSummary file
-  dkMap <- lift $ maybe (DKMap mempty mempty mempty) fst <$> runMaybeT (useWithStaleFastMT GetDocMap file)
+  projectFile <- MaybeT $ pure $ reclassifyInput file
+  env <- hscEnv . fst <$> useWithStaleFastMT GhcSession projectFile
+  modSummary <- fst <$> useWithStaleFastMT GetModSummary projectFile
+  dkMap <- lift $ maybe (DKMap mempty mempty mempty) fst <$> runMaybeT (useWithStaleFastMT GetDocMap projectFile)
   let enabledExtensions = extensionFlags (ms_hspp_opts (msrModSummary modSummary))
 
   !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
@@ -67,7 +72,7 @@ getAtPoint file pos = runMaybeT $ do
 -- taking into account changes that may have occurred due to edits.
 toCurrentLocation
   :: PositionMapping
-  -> NormalizedFilePath
+  -> SomeFileInput
   -> Location
   -> IdeAction (Maybe Location)
 toCurrentLocation mapping file (Location uri range) =
@@ -75,7 +80,7 @@ toCurrentLocation mapping file (Location uri range) =
   -- file than the one we are calling gotoDefinition from.
   -- So we check that the location file matches the file
   -- we are in.
-  if nUri == normalizedFilePathToUri file
+  if nUri == normalizedFilePathToUri (inputFilePath file)
   -- The Location matches the file, so use the PositionMapping
   -- we have.
   then pure $ Location uri <$> toCurrentRange mapping range
@@ -84,19 +89,24 @@ toCurrentLocation mapping file (Location uri range) =
   else do
     otherLocationMapping <- fmap (fmap snd) $ runMaybeT $ do
       otherLocationFile <- MaybeT $ pure $ uriToNormalizedFilePath nUri
-      useWithStaleFastMT GetHieAst otherLocationFile
+      otherHaskellFile <- MaybeT $ pure $ toSomeHaskellInput otherLocationFile
+      useWithStaleFastMT GetHieAst otherHaskellFile
     pure $ Location uri <$> (flip toCurrentRange range =<< otherLocationMapping)
   where
     nUri :: NormalizedUri
     nUri = toNormalizedUri uri
 
 -- | Goto Definition.
-getDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [(Location, Identifier)])
+getDefinition :: SomeFileInput -> Position -> IdeAction (Maybe [(Location, Identifier)])
 getDefinition file pos = runMaybeT $ do
     ide@ShakeExtras{ withHieDb, hiedbWriter } <- ask
     opts <- liftIO $ getIdeOptionsIO ide
-    (hf, mapping) <- useWithStaleFastMT GetHieAst file
-    (ImportMap imports, _) <- useWithStaleFastMT GetImportMap file
+
+    haskellFile <- MaybeT $ pure $ reclassifyInput file
+    projectFile <- MaybeT $ pure $ reclassifyInput file
+
+    (hf, mapping) <- useWithStaleFastMT GetHieAst haskellFile
+    (ImportMap imports, _) <- useWithStaleFastMT GetImportMap projectFile
     !pos' <- MaybeT (pure $ fromCurrentPosition mapping pos)
     locationsWithIdentifier <- AtPoint.gotoDefinition withHieDb (lookupMod hiedbWriter) opts imports hf pos'
     mapMaybeM (\(location, identifier) -> do
@@ -105,11 +115,12 @@ getDefinition file pos = runMaybeT $ do
       ) locationsWithIdentifier
 
 
-getTypeDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [(Location, Identifier)])
+getTypeDefinition :: SomeFileInput -> Position -> IdeAction (Maybe [(Location, Identifier)])
 getTypeDefinition file pos = runMaybeT $ do
     ide@ShakeExtras{ withHieDb, hiedbWriter } <- ask
     opts <- liftIO $ getIdeOptionsIO ide
-    (hf, mapping) <- useWithStaleFastMT GetHieAst file
+    haskellFile <- MaybeT $ pure $ reclassifyInput file
+    (hf, mapping) <- useWithStaleFastMT GetHieAst haskellFile
     !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
     locationsWithIdentifier <- AtPoint.gotoTypeDefinition withHieDb (lookupMod hiedbWriter) opts hf pos'
     mapMaybeM (\(location, identifier) -> do
@@ -117,16 +128,17 @@ getTypeDefinition file pos = runMaybeT $ do
       pure $ Just (fixedLocation, identifier)
       ) locationsWithIdentifier
 
-getImplementationDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [Location])
+getImplementationDefinition :: SomeFileInput -> Position -> IdeAction (Maybe [Location])
 getImplementationDefinition file pos = runMaybeT $ do
     ide@ShakeExtras{ withHieDb, hiedbWriter } <- ask
     opts <- liftIO $ getIdeOptionsIO ide
-    (hf, mapping) <- useWithStaleFastMT GetHieAst file
+    haskellFile <- MaybeT $ pure $ reclassifyInput file
+    (hf, mapping) <- useWithStaleFastMT GetHieAst haskellFile
     !pos' <- MaybeT (pure $ fromCurrentPosition mapping pos)
     locs <- AtPoint.gotoImplementation withHieDb (lookupMod hiedbWriter) opts hf pos'
     traverse (MaybeT . toCurrentLocation mapping file) locs
 
-highlightAtPoint :: NormalizedFilePath -> Position -> IdeAction (Maybe [DocumentHighlight])
+highlightAtPoint :: SomeHaskellInput -> Position -> IdeAction (Maybe [DocumentHighlight])
 highlightAtPoint file pos = runMaybeT $ do
     (HAR _ hf rf _ _,mapping) <- useWithStaleFastMT GetHieAst file
     !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
@@ -137,8 +149,8 @@ highlightAtPoint file pos = runMaybeT $ do
 refsAtPoint :: NormalizedFilePath -> Position -> Action [Location]
 refsAtPoint file pos = do
     ShakeExtras{withHieDb} <- getShakeExtras
-    fs <- HM.keys <$> getFilesOfInterestUntracked
-    asts <- HM.fromList . mapMaybe sequence . zip fs <$> usesWithStale GetHieAst fs
+    fs <- mapMaybe (toSomeHaskellInput . inputFilePath) . HM.keys <$> getFilesOfInterestUntracked
+    asts <- HM.fromList . mapMaybe sequence . zip (map inputFilePath fs) <$> usesWithStale GetHieAst fs
     AtPoint.referencesAtPoint withHieDb file pos (AtPoint.FOIReferences asts)
 
 workspaceSymbols :: T.Text -> IdeAction (Maybe [SymbolInformation])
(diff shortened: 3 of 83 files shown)


ghcide/src/Development/IDE/Core/Actions.hs
```diff
@@ -84,19 +89,24 @@ toCurrentLocation mapping file (Location uri range) =
...
+      otherHaskellFile <- MaybeT $ pure $ toSomeHaskellInput otherLocationFile
+      useWithStaleFastMT GetHieAst otherHaskellFile
     pure $ Location uri <$> (flip toCurrentRange range =<< otherLocationMapping)
   where
     nUri :: NormalizedUri
     nUri = toNormalizedUri uri
 
 -- | Goto Definition.
-getDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [(Location, Identifier)])
+getDefinition :: SomeFileInput -> Position -> IdeAction (Maybe [(Location, Identifier)])
```
wz1000: Should take SomeHaskellInput?

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,329 @@
...
+someInputFilePathMaybe :: SomeInput -> Maybe NormalizedFilePath
+someInputFilePathMaybe input =
+    case inputFingerprint input of
+        InputFile path -> Just path
+        _ -> Nothing
+
+-- | Returns the underlying Normalised File Path of a Typed Rules.
+-- Returns Empty nfp if no such path found.
+-- Not reccomended to use !
+someInputFilePath' :: SomeInput -> NormalizedFilePath
```
wz1000: Remove this

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -307,14 +317,14 @@ getModifyDynFlags f = do
...
+    -> SomeInput
     -> ModSummary -> IO ([FileDiagnostic], Maybe ParsedModule)
-getParsedModuleDefinition packageState opt file ms = do
-    let fp = fromNormalizedFilePath file
-    (diag, res) <- parseModule opt packageState fp ms
-    case res of
-        Nothing   -> pure (diag, Nothing)
-        Just modu -> pure (diag, Just modu)
+getParsedModuleDefinition packageState opt input ms = do
+    case inputFingerprint input of
```
wz1000: `getParsedModuleDefinition` should accept `SomeFileInput` and not use `inputFingerprint`. In general we should avoid using `inputFingerprint` outside of Shake.hs

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -507,16 +518,20 @@ reportImportCyclesRule recorder =
...
            pure (moduleNameString . moduleName . ms_mod $ ms)
           showCycle mods  = T.intercalate ", " (map T.pack mods)
 
 getHieAstsRule :: Recorder (WithPriority Log) -> Rules ()
 getHieAstsRule recorder =
     define (cmapWithPrio LogShake recorder) $ \GetHieAst f -> do
-      tmr <- use_ TypeCheck f
-      hsc <- hscEnv <$> use_ GhcSessionDeps f
-      getHieAstRuleDefinition f hsc tmr
+      case toProjectHaskellInput (inputFilePath f) of
```
wz1000:
```hs
case (f :: SomeHaskellInput) of
  NonProject -> ...
  Project -> ...
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+    | isNonProjectHaskellInput nfp = Just (NonProjectHaskellInput nfp)
+    | otherwise = Nothing
+
+toCabalInput :: NormalizedFilePath -> Maybe CabalInput
+toCabalInput nfp
+    | isCabalInput nfp = Just (CabalInput nfp)
+    | otherwise = Nothing
+
+toSomeHaskellInput :: NormalizedFilePath -> Maybe SomeHaskellInput
+toSomeHaskellInput nfp
```
wz1000: All of these function can use `toSomeFileInput`, i.e

```
toSomeHaskellInput nfp = case toSomeFileInput of
	SomeFileHaskellInput x -> Just x
  _ -> Nothing

File: ghcide/src/Development/IDE/Core/Actions.hs [hunk 6]
```diff
@@ -84,60 +90,72 @@ toCurrentLocation mapping file (Location uri range) =
...
     locationsWithIdentifier <- AtPoint.gotoDefinition withHieDb (lookupMod hiedbWriter) opts imports hf pos'
     mapMaybeM (\(location, identifier) -> do
-      fixedLocation <- MaybeT $ toCurrentLocation mapping file location
+      fixedLocation <- MaybeT $ toCurrentLocation mapping (SomeFileHaskellInput file) location
       pure $ Just (fixedLocation, identifier)
       ) locationsWithIdentifier
 
 
-getTypeDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [(Location, Identifier)])
+getTypeDefinition :: SomeFileInput -> Position -> IdeAction (Maybe [(Location, Identifier)])
```
wz1000: Should be `SomeHaskellInput`? Why is this different from `getDefinition`?

ghcide/src/Development/IDE/Core/Actions.hs
```diff
@@ -84,60 +90,72 @@ toCurrentLocation mapping file (Location uri range) =
...
+    (hf, mapping) <- useWithStaleFastMT GetHieAst haskellFile
     !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
     locationsWithIdentifier <- AtPoint.gotoTypeDefinition withHieDb (lookupMod hiedbWriter) opts hf pos'
     mapMaybeM (\(location, identifier) -> do
       fixedLocation <- MaybeT $ toCurrentLocation mapping file location
       pure $ Just (fixedLocation, identifier)
       ) locationsWithIdentifier
 
-getImplementationDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [Location])
+getImplementationDefinition :: SomeFileInput -> Position -> IdeAction (Maybe [Location])
```
wz1000: `SomeHaskellInput`?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -307,14 +310,11 @@ getModifyDynFlags f = do
 getParsedModuleDefinition
     :: HscEnv
     -> IdeOptions
-    -> NormalizedFilePath
+    -> SomeFileInput
```
wz1000: `SomeHaskellInput` or even `ProjectHaskellInput`?

ghcide/src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -123,8 +138,9 @@ instance Hashable GetImportMap
 instance NFData   GetImportMap
 
 type instance RuleResult GetImportMap = ImportMap
+type instance RuleInput GetImportMap = ProjectHaskellInput
 newtype ImportMap = ImportMap
-  { importMap :: M.Map ModuleName NormalizedFilePath -- ^ Where are the modules imported by this file located?
+  { importMap :: M.Map ModuleName SomeHaskellInput -- ^ Where are the modules imported by this file located?
```
wz1000: Probably `ProjectHaskellInput`?

wz1000 (commented):
Broadly looks good, some inputs need to be tightened up

ghcide/session-loader/Development/IDE/Session/Diagnostics.hs
```diff
@@ -9,6 +9,8 @@ import           Data.List
 import           Data.List.Extra                   (split)
 import           Data.Maybe
 import qualified Data.Text                         as T
+import           Development.IDE.Core.RuleInput (IsFileInput (inputFilePath),
+                                       ProjectHaskellInput)
```
fendor: I feel like it is generally acceptable to import `RuleInput` without an explicit import list.

E.g.

```suggestion
import           Development.IDE.Core.RuleInput 
```

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -15,6 +15,8 @@ import qualified Data.List.NonEmpty                 as NE
 import qualified Data.Map.Strict                    as Map
 import           Data.Maybe
 import qualified Data.Text                          as T
+import           Development.IDE.Core.RuleInput      (IsFileInput (inputFilePath),
+                                                   ProjectHaskellInput (..))
```
fendor: I am not going to point it out everywhere, but here as well. The explicit imports list feels excessive

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -511,14 +513,14 @@ fromTargetId is exts (GHC.TargetModule modName) env dep = do
               , i <- is
               , boot <- ["", "-boot"]
               ]
-    let locs = fmap toNormalizedFilePath' fps
+    let locs = fmap (ProjectHaskellInput . toNormalizedFilePath') fps
```
fendor: This is an example that we need to expose the constructors for the rule inputs. We can't really do anything else than promise this is going to be `ProjectHaskellInput`

ghcide/src/Development/IDE/Import/DependencyInformation.hs
```diff
@@ -94,29 +95,31 @@ emptyPathIdMap = PathIdMap IntMap.empty HMS.empty 0
...
 insertImport :: FilePathId -> Either ModuleParseError ModuleImports -> RawDependencyInformation -> RawDependencyInformation
 insertImport (FilePathId k) v rawDepInfo = rawDepInfo { rawImports = IntMap.insert k v (rawImports rawDepInfo) }
 
-pathToId :: PathIdMap -> NormalizedFilePath -> Maybe FilePathId
+pathToId :: PathIdMap -> ProjectHaskellInput -> Maybe FilePathId
 pathToId PathIdMap{pathToIdMap} path = pathToIdMap HMS.!? path
 
-lookupPathToId :: PathIdMap -> NormalizedFilePath -> Maybe FilePathId
+lookupPathToId :: PathIdMap -> ProjectHaskellInput -> Maybe FilePathId
 lookupPathToId PathIdMap{pathToIdMap} path = HMS.lookup path pathToIdMap
```
fendor: Is this literally the same as `pathToId`? Looks like it, then let's get rid of it.

ghcide/src/Development/IDE/Import/FindImports.hs
```diff
@@ -56,33 +59,36 @@ instance NFData Import where
...
              -> ModuleName
              -> m LocateResult
 locateModuleFile import_dirss exts targetFor isSource modName = do
   let candidates import_dirs =
-        [ toNormalizedFilePath' (prefix </> moduleNameSlashes modName <.> maybeBoot ext)
-           | prefix <- import_dirs , ext <- exts]
+        mapMaybe
+          (toProjectHaskellInput . toNormalizedFilePath')
+          [ prefix </> moduleNameSlashes modName <.> maybeBoot ext
+          | prefix <- import_dirs , ext <- exts]
```
fendor: This feels like a very important change, this needs a comment why we use `toProjectHaskellInput` here instead of `ProjectHaskellInput` directly.

ghcide/src/Development/IDE/Core/OfInterest.hs
```diff
@@ -57,7 +61,7 @@ instance Pretty Log where
   pretty = \case
     LogShake msg -> pretty msg
 
-newtype OfInterestVar = OfInterestVar (Var (HashMap NormalizedFilePath FileOfInterestStatus))
+newtype OfInterestVar = OfInterestVar (Var (HashMap SomeFileInput FileOfInterestStatus))
```
fendor: This `OfInterestVar` only stores `SomeHaskellInput`. Cabal has its own `OfInterestVar`

ghcide/src/Development/IDE/Core/OfInterest.hs
```diff
@@ -86,24 +90,24 @@ instance IsIdeGlobal GarbageCollectVar
 ------------------------------------------------------------
 -- Exposed API
 
-getFilesOfInterest :: IdeState -> IO( HashMap NormalizedFilePath FileOfInterestStatus)
+getFilesOfInterest :: IdeState -> IO( HashMap SomeFileInput FileOfInterestStatus)
```
fendor: All functions here are too general then.

ghcide/src/Development/IDE/Core/OfInterest.hs
```diff
@@ -134,21 +138,24 @@ kick :: Action ()
 kick = do
     files <- HashMap.keys <$> getFilesOfInterestUntracked
     ShakeExtras{exportsMap, ideTesting = IdeTesting testing, lspEnv, progress} <- getShakeExtras
+    let normalizedFiles = map inputFilePath files
+        haskellFiles = mapMaybe (toSomeHaskellInput . inputFilePath) files
+        projectHaskellFiles = mapMaybe (toProjectHaskellInput . inputFilePath) files
```
fendor, line 140: Comment please!

I think I know what's happening, but we are intentionally not running all the rules for everything here.

ghcide/src/Development/IDE/Core/OfInterest.hs
```diff
@@ -161,3 +168,9 @@ kick = do
         liftIO $ writeVar var False
 
     signal (Proxy @"kick/done")
+
+-- | Convert the files-of-interest map to a list keyed by normalized file path.
+filesOfInterestToNormalizedList :: HashMap SomeFileInput FileOfInterestStatus -> [(NormalizedFilePath, FileOfInterestStatus)]
+filesOfInterestToNormalizedList = fmap firstFilePath . HashMap.toList
+  where
+    firstFilePath (file, status) = (inputFilePath file, status)
```
fendor:
```suggestion
listFilesOfInterestStatus :: HashMap SomeFileInput FileOfInterestStatus -> [(NormalizedFilePath, FileOfInterestStatus)]
listFilesOfInterestStatus = fmap firstFilePath . HashMap.toList
  where
    firstFilePath (file, status) = (inputFilePath file, status)
```

I think that's a better name.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+import           Language.LSP.Protocol.Types           (NormalizedFilePath,
+                                                        Uri,
+                                                        fromNormalizedFilePath,
+                                                        toNormalizedUri,
+                                                        uriToNormalizedFilePath)
+import           System.FilePath                       (normalise,
+                                                        takeExtension)
+
+-- | Map each rule with an input type consumed by that rule
+type family RuleInput k
```
fendor: This needs much more documentation.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+
+-- | Describes identity which is used to compare and hash rule inputs
+data InputFingerprint
```
fendor:
```suggestion
-- | Identity of RuleInputs.
-- 
-- Used to efficiently compare and hash rule inputs.
data InputFingerprint
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+      Just b' -> a == b'
+      Nothing -> False
+  _ == _ = False
+
+instance Hashable InputFingerprint where
+  hashWithSalt s InputNoFile = hashWithSalt s (0 :: Int)
+  hashWithSalt s (InputFile p) = hashWithSalt s (1 :: Int, p)
+  hashWithSalt s (InputValue a) = hashWithSalt s (2 :: Int, hash a)
+
+-- RuleInput : IsInput and SomeInput
```
fendor:
```suggestion
```

Or make it clearer what this is.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+    fromInput :: SomeInput -> Maybe i
+    fromInput (SomeInput i) = cast i
+
+
+    -- should not exist
+    inputFingerprint :: i -> InputFingerprint
+    inputFingerprint i = InputValue i
+
+-- | Stores value that implements 'IsInput'.
+data SomeInput = forall i. IsInput i => SomeInput i
```
fendor:
```suggestion
-- | Any value usable that can be used as a `RuleInput`
data SomeInput = forall i. IsInput i => SomeInput i
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+    show (SomeInput i) = show i
+instance NFData SomeInput where
+    rnf (SomeInput i) = rnf i
+
+instance IsInput SomeInput where
+    toInput = id
+    fromInput = Just
+    inputFingerprint (SomeInput i) = inputFingerprint i
+
+-- RuleInput : NoInput
```
fendor: Pls fix all of these comments. I think it would make sense to document the hierarchy in a Note.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+-- | A valid Rule Input that has no file associated with it.
```
fendor: All of these are "valid". I don't think we need to mention that.

However, we should mention that any rule that takes `NoInput` should be treated as a global rule.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
+-- Copyright (c) 2019 The DAML Authors. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
```
fendor:
```suggestion
```

Not their copyright here :D

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+instance IsHaskellInput NonProjectHaskellInput
+
+-- Helpers to classify NormalizedFilePaths to appropriate Typed Rule.
+isCabalInput :: NormalizedFilePath -> Bool
+isCabalInput = (== ".cabal") . takeExtension . fromNormalizedFilePath
+
+isHaskellFilePath :: NormalizedFilePath -> Bool
+isHaskellFilePath fp = takeExtension (fromNormalizedFilePath fp) `elem`
+    [".hs", ".lhs", ".hs-boot", ".lhs-boot"]
```
fendor: This is strictly speaking depending on `optExtensions`. I don't think we should fix it now, but we should add a comment saying this needs to be unified with `optExtensions` in the future.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Development.IDE.Core.RuleInput
```
fendor, line 7: In this module, there are a lot of formatting issues.

* HLS uses indentation 2
  * Pls update this file to use 2 spaces consistently here
* Use newlines consistently, e.g. after each instance/class/function, add a newline. Avoid double newlines.
* Use sections to help to navigate this file.

For example:
```
-- ----------------------------------------------------------------------------
-- Classify FilePaths as RuleInputs
-- ----------------------------------------------------------------------------
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,312 @@
...
+-- | Defines a valid rule input with a fingerprint
+class (Typeable i, Hashable i, Eq i, Show i, NFData i) => IsInput i where
+    toInput :: i -> SomeInput
+    toInput = SomeInput
+
+    fromInput :: SomeInput -> Maybe i
+    fromInput (SomeInput i) = cast i
+
+
+    -- should not exist
```
fendor: It should, shouldn't it?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -332,19 +332,21 @@ getLocatedImportsRule recorder =
...
         let dflags = hsc_dflags env
         opt <- getIdeOptions
-        let getTargetFor modName nfp
-                | Just (TargetFile nfp') <- HM.lookupKey (TargetFile nfp) targets = do
+        let getTargetFor modName input
+                | Just (TargetFile nfp') <- HM.lookupKey (TargetFile input) targets = do
                     -- reuse the existing NormalizedFilePath in order to maximize sharing
-                    itExists <- getFileExists nfp'
-                    return $ if itExists then Just nfp' else Nothing
+                    itExists <- getFileExists (SomeFileHaskellInput ( SomeProjectHaskellInput nfp'))
```
fendor: this pattern is kind of common, perhaps introduce a function `someHaskellInput :: ProjectHaskellInput -> SomeFileInput`?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -373,28 +375,31 @@ getLocatedImportsRule recorder =
         let moduleImports = catMaybes $ bootArtifact : imports'
         pure (concat diags, Just moduleImports)
 
-type RawDepM a = StateT (RawDependencyInformation, IntMap ArtifactsLocation) Action a
+type RawDepState = (RawDependencyInformation, IntMap ArtifactsLocation, IntMap ProjectHaskellInput)
```
fendor: Why do we need this new IntMap? Feels like a code smell.

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -481,11 +488,19 @@ rawDependencyInformation fs = do
...
     dropBootSuffix hs_src = reverse . drop (length @[] "-boot") . reverse $ hs_src
 
+immediateReverseDependencyInputs :: ProjectHaskellInput -> DependencyInformation -> Maybe [ProjectHaskellInput]
+immediateReverseDependencyInputs file DependencyInformation{..} = do
+  FilePathId cur_id <- lookupPathToId depPathIdMap file
+  pure $
+    map (idToPath depPathIdMap . FilePathId) $
+      maybe mempty IntSet.toList $
+        IntMap.lookup cur_id depReverseModuleDeps
+
```
fendor: Duplication with `immediateReverseDependencies`? Do we really need this?

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -481,11 +488,19 @@ rawDependencyInformation fs = do
...
 reportImportCyclesRule :: Recorder (WithPriority Log) -> Rules ()
 reportImportCyclesRule recorder =
-    defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \ReportImportCycles file -> fmap (\errs -> if null errs then (Just "1",([], Just ())) else (Nothing, (errs, Nothing))) $ do
-        DependencyInformation{..} <- useWithSeparateFingerprintRule_ GetModuleGraphTransDepsFingerprints GetModuleGraph file
-        case pathToId depPathIdMap file of
+    defineEarlyCutoff (cmapWithPrio LogShake recorder) $ Rule $ \ReportImportCycles input -> fmap (\errs -> if null errs then (Just "1",([], Just ())) else (Nothing, (errs, Nothing))) $ do
+        DependencyInformation{..} <- useWithSeparateFingerprintRule_ GetModuleGraphTransDepsFingerprints GetModuleGraph input
+        case lookupPathToId depPathIdMap input of
```
fendor: It feels kind of funny, that you seem to just randomly change `lookupPathToId` to `pathToId` in one location, but here you change to `lookupPathToId` from `pathToId` 🤔

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -612,16 +636,16 @@ readHieFileFromDisk recorder hie_loc = do
...
 -- | Typechecks a module.
 typeCheckRule :: Recorder (WithPriority Log) -> Rules ()
-typeCheckRule recorder = define (cmapWithPrio LogShake recorder) $ \TypeCheck file -> do
-    pm <- use_ GetParsedModule file
-    hsc  <- hscEnv <$> use_ GhcSessionDeps file
-    foi <- use_ IsFileOfInterest file
+typeCheckRule recorder = define (cmapWithPrio LogShake recorder) $ \TypeCheck input -> do
+    pm <- use_ GetParsedModule input
+    hsc  <- hscEnv <$> use_ GhcSessionDeps input
+    foi <- use_ IsFileOfInterest (toSomeFileInput (inputFilePath input))
```
fendor: Anything more direct? This feels round-about

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -937,8 +969,8 @@ getModSummaryRule displayTHWarning recorder = do
         session' <- hscEnv <$> use_ GhcSession f
         modify_dflags <- getModifyDynFlags dynFlagsModifyGlobal
         let session = setNonHomeFCHook $ hscSetFlags (modify_dflags $ hsc_dflags session') session' -- TODO wz1000
-        mFileContent <- getFileContents f
-        let fp = fromNormalizedFilePath f
+        mFileContent <- getFileContents (SomeFileHaskellInput(SomeProjectHaskellInput f))
```
fendor, line 1055: This pattern again

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1325,13 +1366,19 @@ mainRule recorder RulesConfig{..} = do
...
 -- | Get HieFile for haskell file on NormalizedFilePath
-getHieFile :: NormalizedFilePath -> Action (Maybe HieFile)
-getHieFile nfp = runMaybeT $ do
-  HAR {hieAst} <- MaybeT $ use GetHieAst nfp
-  tmr <- MaybeT $ use TypeCheck nfp
-  ghc <- MaybeT $ use GhcSession nfp
-  msr <- MaybeT $ use GetModSummaryWithoutTimestamps nfp
-  source <- lift $ getSourceFileSource nfp
+getHieFile :: SomeFileInput -> Action (Maybe HieFile)
+getHieFile input = runMaybeT $ do
```
fendor: Should be `SomeHaskellInput`, right?

ghcide/src/Development/IDE/LSP/HoverDefinition.hs
```diff
@@ -74,21 +79,22 @@ foundHover (mbRange, contents) =
...
     mbResult <- case uriToFilePath' uri of
-        Just path -> logAndRunRequest recorder label getResults ide pos path
+        Just path -> logAndRunRequest recorder label classify getResults ide pos path
         Nothing   -> pure Nothing
     pure $ maybe notFound found mbResult
 
-logAndRunRequest :: Recorder (WithPriority Log) -> T.Text -> (NormalizedFilePath -> Position -> IdeAction b) -> IdeState -> Position -> String -> IO b
-logAndRunRequest recorder label getResults ide pos path = do
+logAndRunRequest :: Recorder (WithPriority Log) -> T.Text -> (NormalizedFilePath -> Maybe input) -> (input -> Position -> IdeAction (Maybe a)) -> IdeState -> Position -> String -> IO (Maybe a)
+logAndRunRequest recorder label classify getResults ide pos path = do
```
fendor: I am wondering whether this function shouldn't rather take 


```suggestion
logAndRunRequest :: Recorder (WithPriority Log) -> T.Text -> (input -> Position -> IdeAction (Maybe a)) -> IdeState -> Position -> input -> IO (Maybe a)
logAndRunRequest recorder label classify getResults ide pos path = do
```

fendor: `path` being a `String` feels wrong on multiple levels.

ghcide/src/Development/IDE/LSP/Notifications.hs
```diff
@@ -60,34 +64,43 @@ instance Pretty Log where
     LogWatchedFileEvents msg -> "Watched file events:" <+> pretty msg
     LogWarnNoWatchedFilesSupport -> "Client does not support watched files. Falling back to OS polling"
 
-whenUriFile :: Uri -> (NormalizedFilePath -> IO ()) -> IO ()
-whenUriFile uri act = whenJust (LSP.uriToFilePath uri) $ act . toNormalizedFilePath'
+whenUriFile :: Uri -> (SomeFileInput -> IO ()) -> IO ()
+whenUriFile uri act = whenJust (LSP.uriToFilePath uri) $ act . toSomeFileInput . toNormalizedFilePath'
```
fendor: This should be `SomeHaskellInput` as in here, we only should be processing haskell files.

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -170,49 +176,54 @@ getCompletionsLSP ide plId
...
-            knownTargets <- liftIO $ runAction  "Completion" ide $ useNoFile GetKnownTargets
-            let localModules = maybe [] (Map.keys . targetMap) knownTargets
-            let lModules = mempty{importableModules = map toModueNameText localModules}
-            -- set up the exports map including both package and project-level identifiers
-            packageExportsMapIO <- fmap(envPackageExports . fst) <$> useWithStaleFast GhcSession npath
-            packageExportsMap <- mapM liftIO packageExportsMapIO
-            projectExportsMap <- liftIO $ readTVarIO (exportsMap $ shakeExtras ide)
-            let exportsMap = fromMaybe mempty packageExportsMap <> projectExportsMap
+        case toProjectHaskellInput npath of
+          Nothing -> return (InL [])
```
fendor: Comment please!

ghcide/src/Development/IDE/Plugin/TypeLenses.hs
```diff
@@ -177,9 +183,12 @@ codeLensProvider ideState pId CodeLensParams{_textDocument = TextDocumentIdentif
 codeLensResolveProvider :: ResolveFunction IdeState TypeLensesResolve Method_CodeLensResolve
 codeLensResolveProvider ideState pId lens@CodeLens{_range} uri TypeLensesResolve = do
   nfp <- getNormalizedFilePathE uri
+  projectFile <- handleMaybe
+    (PluginInvalidParams $ "Expected project Haskell file: " <> T.pack (show uri))
+    (toProjectHaskellInput nfp)
```
fendor: This should be a common utility function, perhaps defined in PluginUtils or something like that?

ghcide/src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -567,15 +569,15 @@ locationsAtPoint
...
+  -> M.Map ModuleName SomeHaskellInput
   -> Position
   -> HieAstResult
   -> m [(Location, Identifier)]
 locationsAtPoint withHieDb lookupModule _ideOptions imports pos (HAR _ ast _rm _ _) =
   let ns = concat $ pointCommand ast pos (M.keys . getNodeIds)
       zeroPos = Position 0 0
       zeroRange = Range zeroPos zeroPos
-      modToLocation m = fmap (\fs -> pure (Location (fromNormalizedUri $ filePathToUri' fs) zeroRange)) $ M.lookup m imports
+      modToLocation m = fmap (\fs -> pure (Location (fromNormalizedUri $ filePathToUri' $ inputFilePath fs) zeroRange)) $ M.lookup m imports
```
fendor: Utility function, perhaps `inputUri`?

ghcide/src/Development/IDE/Spans/Pragmas.hs
```diff
@@ -50,9 +52,13 @@ insertNewPragma (NextPragmaInfo nextPragmaLine _) newPragma =  LSP.TextEdit prag
         pragmaInsertPosition = LSP.Position (fromIntegral nextPragmaLine) 0
         pragmaInsertRange = LSP.Range pragmaInsertPosition pragmaInsertPosition
 
-getFirstPragma :: MonadIO m => PluginId -> IdeState -> NormalizedFilePath -> ExceptT PluginError m NextPragmaInfo
+getFirstPragma :: MonadIO m => PluginId -> IdeState -> SomeFileInput -> ExceptT PluginError m NextPragmaInfo
```
fendor: `ProjectHaskellInput`?

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -66,6 +66,9 @@ data PluginError
     -- It will be logged with Warning and takes medium precedence (2) in being
     -- returned as a response to the client.
   | PluginInvalidParams T.Text
+    -- |PluginUnsupportedUriType should be used if a request URI cannot be
+    -- classified as a file URI.
```
fendor:
```suggestion
    -- | Plugin received a request for an URI it doesn't support.
    -- For example, in a haskell plugin, code actions were requested for non-haskell files.
```

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -66,6 +66,9 @@ data PluginError
...
+  | PluginUnsupportedUriType Uri
```
fendor, line 73: Perhaps we can also record what the supported uri types were? Can be done in a follow-up PR, but add a TODO

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -109,13 +112,16 @@ instance Pretty PluginError where
...
+      PluginUnsupportedUriType uri -> "Unsupported URI type:" <+>
+                                      pretty (show uri)
       PluginInvalidUserState text -> "Invalid User State:" <+> pretty text
       PluginRequestRefused msg    -> "Request Refused: "   <+> pretty msg
 
 -- |Converts to ErrorCode used in LSP ResponseErrors
 toErrorCode :: PluginError -> (LSPErrorCodes |? ErrorCodes)
 toErrorCode (PluginInternalError _)    = InR ErrorCodes_InternalError
 toErrorCode (PluginInvalidParams _)    = InR ErrorCodes_InvalidParams
+toErrorCode (PluginUnsupportedUriType _) = InR ErrorCodes_InvalidParams
```
fendor, line 127: This feels like the wrong error code, I think we should be using `InL LSPErrorCodes_RequestFailed`, or a custom error. I think, we should add a TODO that this should be a custom error code.

hls-plugin-api/src/Ide/Plugin/Error.hs
```diff
@@ -130,6 +136,7 @@ toErrorCode PluginStaleResolve         = InL LSPErrorCodes_ContentModified
 toPriority :: PluginError -> Priority
 toPriority (PluginInternalError _)    = Error
 toPriority (PluginInvalidParams _)    = Warning
+toPriority (PluginUnsupportedUriType _) = Warning
```
fendor, line 142: We might want to demote this to `Debug`. IIRC, a warning is displayed in the editor, but these errors will be perfectly sensible in the future, right? 

@wz1000 opinions?

fendor (changes requested):
Thanks, this looks very nice! This is a nitpick review, so it is the intention to be done after the review is addressed. I have reviewed the ghcide changes, will review plugins soon.

Further, we agreed the following documentation is still missing and will be added:

        * What are the existing rules?
        * What is RuleInput?
        * Why do we have RuleInput
        * What is the design of RuleInput
            Note [Hierarchical Inputs]
            Note [RuleInput]
            
            Reference these notes from various type classes and RuleInput type
        * Helper functions need documentation

ghcide/session-loader/Development/IDE/Session/Ghc.hs
vidit-od: This is a miss from my part, we want to use toProjectHaskellInput everywhere; i.e : 

```haskell
    let locs = mapMaybe (toProjectHaskellInput . toNormalizedFilePath') fps
```

ghcide/src/Development/IDE/Core/RuleInput.hs
vidit-od: These were for me to navigate, in case i want to show something during our calls 
missed removing them :)

ghcide/src/Development/IDE/Core/Rules.hs
vidit-od: i would recommend keeping it as it is without any helpers, 
we are also using this pattern for pattern matching 

this keeps the same usage for both;

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+import           Ide.Plugin.Error            (PluginError (..))
+import           Language.LSP.Protocol.Types (NormalizedFilePath, Uri,
+                                              filePathToUri,
+                                              fromNormalizedFilePath,
+                                              toNormalizedUri,
+                                              uriToNormalizedFilePath)
+import           System.FilePath             (normalise, takeExtension)
+
+{- Note [RuleInputs]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
fendor: This should be a haddock comment, rather than a note. 

A note is standalone, e.g., not associated to a specific type. You reference notes from multiple locations.

I think this works well as a haddock comment.


```suggestion
-- |
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+-- Every rule key must define an instance of this open type family. For
+-- example:
+--
+-- @
+-- data GetParsedModule = GetParsedModule
+-- type instance RuleInput GetParsedModule = ProjectHaskellInput
+-- @
+--
+-- This makes both the rule definition and its callers use a
+-- 'ProjectHaskellInput' when working with @GetParsedModule@.
```
fendor: Why do we need this?

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+-}
```
fendor:
```suggestion

```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+'RuleInput' can be a node from this AST:
+
+@
+'SomeInput'                          ->  any rule input
++-- 'NoInput'                        ->  global
+`-- 'SomeFileInput'                  ->  has a 'NormalizedFilePath'
+    +-- 'CabalInput'                 ->  all .cabal files
+    `-- 'SomeHaskellInput'           ->  all Haskell files
+        +-- 'NonProjectHaskellInput' ->  files in .hls/dependencies
+        `-- 'ProjectHaskellInput'    ->  all other Haskell files
```
fendor:
```suggestion
'SomeInput'                          --  any rule input
├── 'NoInput'                        --  global
└── 'SomeFileInput'                  --  has a 'NormalizedFilePath'
    ├── 'SomeHaskellInput'           --  all Haskell files
    │   ├── 'NonProjectHaskellInput' --  files in .hls/dependencies
    │   └── 'ProjectHaskellInput'    --  haskell files of your project 
    └── 'CabalInput'                 --  all .cabal files
```

Just looks nicer, tbh 🙈

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+@
+
+Within this hierarchy, a parent accepts all descendants. Upcasting wraps a child
+in each parent constructor, then uses 'toInput' for 'SomeInput'; for example,
+'ProjectHaskellInput' -> 'SomeHaskellInput' -> 'SomeFileInput' -> 'SomeInput'.
+'fromInput' performs the checked downcast and succeeds only within its branch.
```
fendor: What does `a parent accepts all descendants` mean?

fendor:
```suggestion
We can downcast a rule input via 'fromInput'.
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+      Just b' -> a == b'
+      Nothing -> False
+  _ == _ = False
+
+instance Hashable InputFingerprint where
+  hashWithSalt s InputNoFile    = hashWithSalt s (0 :: Int)
+  hashWithSalt s (InputFile p)  = hashWithSalt s (1 :: Int, p)
+  hashWithSalt s (InputValue a) = hashWithSalt s (2 :: Int, hash a)
+
+{- Note [Rule input hierarchy]
```
fendor, line 84: Similar to above, a `Note`, shouldn't be the documentation for a type. I like this `Note`, so I think we should keep it as is.
But the type class `IsInput` should have separate documentation that describes what it does, ignoring the actual rule hierarchy.
The haddock comment should reference this note, though.

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+  toInput :: i -> SomeInput
+  toInput = SomeInput
+
+  fromInput :: SomeInput -> Maybe i
+  fromInput (SomeInput i) = cast i
+
+  inputFingerprint :: i -> InputFingerprint
+  inputFingerprint i = InputValue i
+
+-- | Stores value that implements 'IsInput'.
```
fendor:
```suggestion
-- | Stores any value that implements 'IsInput'.
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+instance NFData SomeInput where
+  rnf (SomeInput i) = rnf i
+
+instance IsInput SomeInput where
+  toInput = id
+  fromInput = Just
+  inputFingerprint (SomeInput i) = inputFingerprint i
+
+-- | A Rule Input that has no file associated with it.
+-- Rules with 'NoInput' ruletype must be treated as global rules.
```
fendor:
```suggestion
-- Rules with 'RuleInput' 'NoInput' must be treated as global rules.
```

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,372 @@
...
+isNonProjectHaskellInput :: NormalizedFilePath -> Bool
+isNonProjectHaskellInput fp = isHaskellFilePath fp && isDependencyHaskellPath fp
+
+isProjectHaskellInput :: NormalizedFilePath -> Bool
+isProjectHaskellInput fp = isHaskellFilePath fp && not (isDependencyHaskellPath fp)
+
+isDependencyHaskellPath :: NormalizedFilePath -> Bool
+isDependencyHaskellPath = (".hls/dependencies" `isInfixOf`) . normalise . fromNormalizedFilePath
+
+-- | Returns the underlying Normalised File Path of a Typed Rule ONLY if it exists.
```
fendor, line 307:
```suggestion
-- | Returns the underlying 'NormalizedFilePath' of 'SomeInput` if it exists.
```

fendor (changes requested):
Some more comments :)

ghcide/src/Development/IDE/LSP/Notifications.hs
```diff
@@ -60,34 +61,43 @@ instance Pretty Log where
...
       \ide vfs _ (DidChangeTextDocumentParams identifier@VersionedTextDocumentIdentifier{_uri} changes) -> liftIO $ do
         atomically $ updatePositionMapping ide identifier changes
         whenUriFile _uri $ \file -> do
-          setFileModified (cmapWithPrio LogFileStore recorder) (VFSModified vfs) ide False file $
-            addFileOfInterest ide file Modified{firstOpen=False}
+          let action = addFileOfInterest ide file Modified{firstOpen=False}
+          case file of
+            SomeProjectHaskellInput projectFile ->
+              setFileModified (cmapWithPrio LogFileStore recorder) (VFSModified vfs) ide False projectFile action
+            _ -> setSomethingModified (VFSModified vfs) ide (fromNormalizedFilePath (inputFilePath file) ++ " (modified)") action
```
fendor: We shouldn't do anything here, rather log this case.

ghcide/src/Development/IDE/LSP/Notifications.hs
```diff
@@ -60,34 +61,43 @@ instance Pretty Log where
...
       whenUriFile _uri $ \file -> do
+          let action = addFileOfInterest ide file Modified{firstOpen=True}
           -- We don't know if the file actually exists, or if the contents match those on disk
           -- For example, vscode restores previously unsaved contents on open
-          setFileModified (cmapWithPrio LogFileStore recorder) (VFSModified vfs) ide False file $
-            addFileOfInterest ide file Modified{firstOpen=True}
+          case file of
+            SomeProjectHaskellInput projectFile ->
+              setFileModified (cmapWithPrio LogFileStore recorder) (VFSModified vfs) ide False projectFile action
+            _ -> setSomethingModified (VFSModified vfs) ide (fromNormalizedFilePath (inputFilePath file) ++ " (modified)") action
```
fendor: Should we set something modified here? What does this do?

ghcide/src/Development/IDE/LSP/Notifications.hs
```diff
@@ -60,34 +61,43 @@ instance Pretty Log where
...
   , mkPluginNotificationHandler LSP.SMethod_TextDocumentDidSave $
       \ide vfs _ (DidSaveTextDocumentParams TextDocumentIdentifier{_uri} _) -> liftIO $ do
         whenUriFile _uri $ \file -> do
-            setFileModified (cmapWithPrio LogFileStore recorder) (VFSModified vfs) ide True file $
-                addFileOfInterest ide file OnDisk
+            let action = addFileOfInterest ide file OnDisk
+            case file of
+              SomeProjectHaskellInput projectFile ->
+                setFileModified (cmapWithPrio LogFileStore recorder) (VFSModified vfs) ide True projectFile action
+              _ -> setSomethingModified (VFSModified vfs) ide (fromNormalizedFilePath (inputFilePath file) ++ " (modified)") action
```
fendor: Same here

ghcide/src/Development/IDE/LSP/Notifications.hs
fendor: @wz1000 What should we do in this case? Call `setSomethingModified` for completeness? We need to add it to the set of files of interest, but it might not need to record that it isn't the first time this non project haskell file was opened, right? :thinking:

ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -494,8 +496,9 @@ reportImportCyclesRule recorder =
                   let cycles = mapMaybe (cycleErrorInFile fileId) (toList errs)
                   -- Convert cycles of files into cycles of module names
                   forM cycles $ \(imp, files) -> do
-                      modNames <- forM files $
-                          getModuleName . idToPath depPathIdMap
+                      modNames <- forM files $ \cycleFileId -> do
+                          let cycleInput = idToPath depPathIdMap cycleFileId
```
fendor: please revert this change, doesn't seem to be related.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -467,56 +471,61 @@ getIdeOptionsIO ide = do
...
-      Nothing -> readPersistent
-      Just (ValueWithDiagnostics value _) -> case value of
-        Succeeded ver (fromDynamic -> Just v) ->
-            atomicallyNamed "lastValueIO 5"  $ Just . (v,) <$> mappingForVersion positionMapping file ver
-        Stale del ver (fromDynamic -> Just v) ->
-            atomicallyNamed "lastValueIO 6"  $ Just . (v,) . maybe id addOldDelta del <$> mappingForVersion positionMapping file ver
-        Failed p | not p -> readPersistent
-        _ -> pure Nothing
+lastValueIO :: forall k v. IdeRule k v => ShakeExtras -> k -> RuleInput k -> IO (Maybe (v, PositionMapping))
+lastValueIO s@ShakeExtras{positionMapping,persistentKeys,state} k input = do
```
fendor: Perhaps we should minimise the diff here. Don't change the `lastValueIO` function at all, but introduce an intermediate function, e.g. `lastValueForInput :: forall k v. IdeRule k v => ShakeExtras -> k -> RuleInput k -> IO (Maybe (v, PositionMapping))` which calls `lastValueIO`. This avoids having this whole function in the diff, especially since you accidentally changed indentation of the guards in `readPersistent`.

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1278,17 +1320,17 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+                staleV <- liftIO $ atomicallyNamed "define -read 3" $ getValues state key input <&> \case
                     Nothing                   -> Failed False
                     Just (Succeeded ver v, _) -> Stale Nothing ver v
                     Just (Stale d ver v, _)   -> Stale d ver v
                     Just (Failed b, _)        -> Failed b
                 (mbBs, (diags, mbRes)) <- actionCatch
                     (do v <- action staleV; liftIO $ evaluate $ force v) $
                     \(e :: SomeException) -> do
-                        pure (Nothing, ([ideErrorText file (prettyRuleAbortedByException key file e) | not $ isBadDependency e],Nothing))
+                        pure (Nothing, ([ideErrorText file (prettyRuleAbortedByException key file e) | Just file <- [mbFile], not $ isBadDependency e], Nothing))
```
fendor: please show the rule input in `prettyRuleAbortedByException` instead of filtering the error out.

This would swallow exceptions

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1097,177 +1107,209 @@ data FastResult a = FastResult { stale :: Maybe (a,PositionMapping), uptoDate ::
...
+    -> RuleInput k
     -> Maybe BS.ByteString
     -> RunMode
     -> (Value v -> Action (Maybe BS.ByteString, IdeResult v))
     -> Action (RunResult (A (RuleResult k)))
-defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
+defineEarlyCutoff' doDiagnostics cmp key input mbOld mode action = do
+    let mbFile = case inputFingerprint input of
+          InputFile file -> Just file
+          _              -> Nothing
```
fendor: Move this into `inProgress` notification change

ghcide/src/Development/IDE/Core/Shake.hs
```diff
@@ -1316,20 +1358,22 @@ defineEarlyCutoff' doDiagnostics cmp key file mbOld mode action = do
...
+            -- GetModificationTime depends on these rules, so avoid creating a cycle
+            | Just Refl <- eqT @k @AddWatchedFile -> pure Nothing
+            | Just Refl <- eqT @k @IsFileOfInterest -> pure Nothing
+            -- GetFileExists gets called for missing files
+            | Just Refl <- eqT @k @GetFileExists -> pure Nothing
+            -- For all other rules - compute the version properly without:
+            --  * creating a dependency: If everything depends on GetModificationTime, we lose early cutoff
+            --  * creating bogus "file does not exists" diagnostics
+            | otherwise -> useWithoutDependency (GetModificationTime_ False) (toSomeFileInput file)
+          _ -> pure Nothing
```
fendor: Please list out all the possible cases here. I think that's better, even though the rhs is going to be identical for each

ghcide/src/Development/IDE/Import/FindImports.hs
```diff
@@ -56,33 +57,38 @@ instance NFData Import where
...
              -> Bool
              -> ModuleName
              -> m LocateResult
 locateModuleFile import_dirss exts targetFor isSource modName = do
   let candidates import_dirs =
-        [ toNormalizedFilePath' (prefix </> moduleNameSlashes modName <.> maybeBoot ext)
-           | prefix <- import_dirs , ext <- exts]
+        mapMaybe
+          -- GHC import paths are untyped and may include generated dependency roots;
+          -- retain only project Haskell inputs.
```
fendor: Which generated dependency roots? Where would this come from?

ghcide/src/Development/IDE/Plugin/Completions.hs
```diff
@@ -126,13 +129,13 @@ dropListFromImportDecl iDecl = let
 resolveCompletion :: ResolveFunction IdeState CompletionResolveData Method_CompletionItemResolve
 resolveCompletion ide _pid comp@CompletionItem{_detail,_documentation,_data_} uri (CompletionResolveData _ needType (NameDetails mod occ)) =
   do
-    file <- getNormalizedFilePathE uri
+    projectInput <- withExceptT (const PluginStaleResolve) $ classifyAsHaskell uri
```
fendor: Wrong classifier if completions only work with ProjectHaskell files, right?

ghcide/src/Development/IDE/Plugin/Test.hs
```diff
@@ -163,29 +167,35 @@ getDatabaseKeys field db = do
...
-parseActions other _ = return $ Left $ "Cannot parse ide rule: " <> pack (original other)
+parseActions :: CI String -> [SomeFileInput] -> Action (Either Text [Bool])
+parseActions action fps
+    | action == fromString "typecheck"
+    , Just pFiles <- traverse projectFile fps =
+        fmap (Right . map isJust) (uses TypeCheck pFiles)
+  where
+    projectFile (SomeFileHaskellInput (SomeProjectHaskellInput pFile)) = Just pFile
+    projectFile _ = Nothing
+parseActions action fps = sequence <$> traverse (parseAction action) fps
```
fendor: What is going on here?

ghcide/src/Development/IDE/Core/RuleInput.hs
```diff
@@ -0,0 +1,373 @@
...
+  [".hs", ".lhs", ".hs-boot", ".lhs-boot"]
+
+isNonProjectHaskellInput :: NormalizedFilePath -> Bool
+isNonProjectHaskellInput fp = isHaskellFilePath fp && isDependencyHaskellPath fp
+
+isProjectHaskellInput :: NormalizedFilePath -> Bool
+isProjectHaskellInput fp = isHaskellFilePath fp && not (isDependencyHaskellPath fp)
+
+isDependencyHaskellPath :: NormalizedFilePath -> Bool
+isDependencyHaskellPath = (".hls/dependencies" `isInfixOf`) . normalise . fromNormalizedFilePath
```
wz1000: Let's use something like

```
isDependencyHaskellPath =
    isInfixOf [".hls", "dependencies"]
      . splitDirectories
      . fromNormalizedFilePath
```

also `normalise` is redundant after `fromNormalizedFilePath`

This let's us avoid issues with path seperators on windows etc.

Perhaps we could even use `System.FilePath.Glob`, we already have that dependency.

ghcide/src/Development/IDE/LSP/Notifications.hs
wz1000: `setSomethingModified` is not correct here, that will trash the entire session. I think `setFileModified` should accept `SomeHaskellInput` instead of `ProjectHaskellInput`

It should always restart with the new VFS and dirty the file’s GetModificationTime, then run typecheckParents only in the SomeProjectHaskellInput branch

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -45,20 +46,23 @@ data GetNotesInFile = MkGetNotesInFile
...
 -- (note name -> [position]).
+type instance RuleInput GetNotesInFile = SomeFileInput
 type instance RuleResult GetNotesInFile = (HM.HashMap Text Position, HM.HashMap Text [Position])
 
 data GetNotes = MkGetNotes
     deriving (Show, Generic, Eq, Ord)
     deriving anyclass (Hashable, NFData)
 -- GetNotes collects all note definition across all files in the
 -- project. It returns a map from note name to pair of (filepath, position).
+type instance RuleInput GetNotes = SomeFileInput
```
wz1000: The comment claims it collects note definitions across all files, then why does it have a `SomeFileInput` instead of `NoInput`?

plugins/hls-notes-plugin/src/Ide/Plugin/Notes.hs
```diff
@@ -45,20 +46,23 @@ data GetNotesInFile = MkGetNotesInFile
...
 type instance RuleResult GetNotes = HashMap Text (NormalizedFilePath, Position)
 
 data GetNoteReferences = MkGetNoteReferences
     deriving (Show, Generic, Eq, Ord)
     deriving anyclass (Hashable, NFData)
 -- GetNoteReferences collects all note references across all files in the
 -- project. It returns a map from note name to list of (filepath, position).
+type instance RuleInput GetNoteReferences = SomeFileInput
```
wz1000: same as above.

wz1000 (changes requested):
plugins and input types broadly look correct. I need to do another more careful pass though.

PR #5014, opened 2026-07-14
Case split plugin

Aster89:
This PR is for (re-¹)introducing the so called _case-split plugin_, as requested in https://github.com/haskell/haskell-language-server/issues/5013, and it is part of my project for GSoC 2026.

To see the full summary of the project, [click here](https://gist.github.com/Aster89/69a57c364ec2d0cc89f474cc3d194371). The summary also contains a list of stretch goals, some of which have been addressed already.

The work I've done and that I plan to submit via the present pull request, has consisted mainly of:

  - implementing the _case-split_ plugin by incrementally focusing on several use-cases from less to more complex,
  - making use of [`ghc-exactprint`](https://hackage-content.haskell.org/package/ghc-exactprint-1.14.0.0) for correctly layout out the patterns to be inserted,
  - contextually writing tests for guarding such use-cases as I enabled them,
  - finally, striving to address some stretch goals.

The result of such work is a plugin providing a code action for adding missing patterns to a `case`/`\case` expression that has a non-exhaustive list of patterns.

For instance, the plugin turns this
```haskell
data X = A
       | B
       | C Int
       | D Int Int
       | E Int Int Int Int
       | F

foo :: X -> Int
foo x = case x of
          A -> 3;
                B -> 4;
              C _ -> 5
```
into this
```haskell
data X = -- same as above

foo :: X -> Int
foo x = case x of
          A -> 3;
                B -> 4;
              C _ -> 5
          D _ _ -> _
          E {} -> _
          F -> _
```

Fix #5013.

---
 
### ¹ Earlier attempts

Indeed, there were some earlier attempts. See the [summary](https://gist.github.com/Aster89/69a57c364ec2d0cc89f474cc3d194371) for some pointers.

PR #5026, opened 2026-07-24
Some improvements of comments, of code, and return->pure

Aster89:
Some changes, hopefully improvements, I made while working on #5014:

  - change `return` to `pure` in a file I passed by,
  - split a very long line in two,
  - update a comment according to what I did in #4937,
  - be more explicit on a pattern match on an `EpToken`, considering it only has 2 ctors,
  - made `addMethodDecls`'s signature and a piece of the implemention simpler for GHC ≥ 9.10.

### DIFF
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
@@ -50,14 +50,10 @@ makeMethodDecl df (mName, sig) = do
     sig' <- eitherToMaybe $ parseDecl df (T.unpack sig) $ T.unpack sig
     pure (name, sig')
 
-#if !MIN_VERSION_ghc(9,10,0)
 #if MIN_VERSION_ghc_exactprint(1,10,0)
-addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> Located (HsModule GhcPs)
+addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> ParsedSource
 #else
 addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> TransformT Identity (Located (HsModule GhcPs))
-#endif
-#else
-addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> ParsedSource
 #endif
 addMethodDecls ps mDecls range withSig
     | withSig = go (concatMap (\(decl, sig) -> [sig, decl]) mDecls)
@@ -71,6 +67,9 @@ addMethodDecls ps mDecls range withSig
 #endif
         case break (inRange range . getLoc) allDecls of
 #if MIN_VERSION_ghc(9,10,0)
+            -- With new enough GHC, `hsDecls` followed by
+            -- `replaceDecls` is idempotent, see
+            -- https://github.com/alanz/ghc-exactprint/blob/e2224c5e3d335e5a2c50a44749367337e278a8c0/src/Language/Haskell/GHC/ExactPrint/Transform.hs#L817
             (_, []) -> ps
 #else
             (before, []) -> replaceDecls ps before
--- a/plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
+++ b/plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
@@ -50,10 +50,14 @@ makeMethodDecl df (mName, sig) = do
     sig' <- eitherToMaybe $ parseDecl df (T.unpack sig) $ T.unpack sig
     pure (name, sig')
 
+#if !MIN_VERSION_ghc(9,10,0)
 #if MIN_VERSION_ghc_exactprint(1,10,0)
 addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> Located (HsModule GhcPs)
 #else
 addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> TransformT Identity (Located (HsModule GhcPs))
+#endif
+#else
+addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> ParsedSource
 #endif
 addMethodDecls ps mDecls range withSig
     | withSig = go (concatMap (\(decl, sig) -> [sig, decl]) mDecls)
@@ -66,6 +70,11 @@ addMethodDecls ps mDecls range withSig
         allDecls <- hsDecls ps
 #endif
         case break (inRange range . getLoc) allDecls of
+#if MIN_VERSION_ghc(9,10,0)
+            (_, []) -> ps
+#else
+            (before, []) -> replaceDecls ps before
+#endif
             (before, L l inst : after) ->
                 let
                     indent = case inst of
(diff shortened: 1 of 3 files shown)


plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
```diff
@@ -50,11 +50,15 @@ makeMethodDecl df (mName, sig) = do
...
 
+#if !MIN_VERSION_ghc(9,10,0)
 #if MIN_VERSION_ghc_exactprint(1,10,0)
 addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> Located (HsModule GhcPs)
 #else
 addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> TransformT Identity (Located (HsModule GhcPs))
 #endif
+#else
+addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> ParsedSource
+#endif
```
fendor: I don't quite understand why we need to add a new signature, was this previously throwing a warning for lacking a signature?

Also, see `Note [Guidelines For Using CPP In GHCIDE Import Statements]`, I think we should follow these CPP guidelines in the class plugin as wlel.

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
```diff
@@ -66,6 +70,11 @@ addMethodDecls ps mDecls range withSig
         allDecls <- hsDecls ps
 #endif
         case break (inRange range . getLoc) allDecls of
+#if MIN_VERSION_ghc(9,10,0)
+            (_, []) -> ps
+#else
+            (before, []) -> replaceDecls ps before
+#endif
```
fendor, line 76: That feels surprising, again, was this just warning with GHC 9.8 or what is going on here?

fendor (changes requested):
Some small comments.

Generally, we don't really do these kinds of refactorings of `return -> pure` unless there is a real incentive.
We don't do this because occasionally, the code in question is directly copied from GHC, changing it just diverges from the original sources, making it harder to see that.
This time, I think it is fine.

Perhaps we should create a style guide for HLS 🤔

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
Aster89: Nothing in this change is _**need**_, but as new contributor, I think the new signature makes the code more welcoming towards other contributors.

Knowing that for recent GHC versions this function takes a `ParsedSource`, other arguments, and returns a `ParsedSource` makes it easier to digest the signature and to remember what the function does, i.e. _**turn a `ParsedSource` into another one**_, so the reader can mentally take not of this signature and archive the others as "the old way".

When I worked on https://github.com/haskell/haskell-language-server/pull/4937, I lost the count of how many times I had to go back and forth between hackage and the IDE to remind myself that `Located (HsModule GhcPs)` was just `ParsedSource`, so I think there's value in encoding it here.

I also think it adds value because this new signature doesn't need to also be excplicitly specific to a version of exactprint. Ideally, when support for GHC pre-9.10 will be dropped, we will be able to retain the new signature without thinking about exactprint versions.

I see I'm adding a nested `#if`, but I think it adds value, not detract it.

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
Aster89: Like above I think there's value in telling the user "if you're on recent enough GHC, the code is _this_ simple, forget about the other one".

Furthermore, the `(before, []) -> replaceDecls ps before` represents an example of _`hsDecls` followed by `replaceDecls`_, which is supposed to be idempotent, according to [Alan's comment in ghc-exactprint codebase](https://github.com/alanz/ghc-exactprint/blob/e2224c5e3d335e5a2c50a44749367337e278a8c0/src/Language/Haskell/GHC/ExactPrint/Transform.hs#L817):

> -- ++AZ++: TODO: add tests to confirm that hsDecls followed by replaceDecls is idempotent

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
fendor: Ok, thank you for the explanation. I agree that trees that grow make it *very* difficult to know with which type you are actually working with. However, adding a type signature that *isn't* used is just unmaintainable in my opinion.

[`ghc-exactprint-1.10.0.0`](https://hackage.haskell.org/package/ghc-exactprint-1.10.0.0) has the constraint: `ghc (>=9.10 && <9.11)`. This means that the `MIN_VERSION_ghc(9,10,0)` is effectively ensuring that `addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> Located (HsModule GhcPs)` is never used and never type checked (not even parsed).

Thus, this patch adds cruft and dead code. I am not sure how to improve the discoverability of type synonyms (e.g., discovering/reminding that `Located (HsModule GhcPs)` == `ParsedSource`) but this will just lead to more confusion down the way.

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
fendor: I see, I think this warrants a comment then :) 

Also, in this case, also consider avoiding CPP and using `ghcVersion` from `ghcide/src/Development/IDE/GHC/Compat.hs`. Then we have IDE support on both branches.

Either way, this definitely needs a comment, like linking to the link you just posted here.

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
Aster89: I hadn't seen that constraint.

Ok, but this simply means I can remove _that_ dead code and keep the new signature:
```haskell
#if !MIN_VERSION_ghc(9,10,0)
addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> TransformT Identity (Located (HsModule GhcPs))
#else
addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> ParsedSource
#endif
```
No?

fendor: Rather keep it as and use `ParsedSource`, even though I am not sure if that makes it better the next time you see some function that takes a `Located (HsModule GhcPs)`

```haskell
#if MIN_VERSION_ghc_exactprint(1,10,0)
addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> ParsedSource
#else
addMethodDecls :: ParsedSource -> [(LHsDecl GhcPs, LHsDecl GhcPs)] -> Range -> Bool -> TransformT Identity (Located (HsModule GhcPs))
#endif
```

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
```diff
@@ -66,6 +66,13 @@ addMethodDecls ps mDecls range withSig
         allDecls <- hsDecls ps
 #endif
         case break (inRange range . getLoc) allDecls of
+            (before, []) ->  if ghcVersion /= GHC910
```
fendor:
```suggestion
            (before, []) ->  if ghcVersion < GHC910
```

plugins/hls-class-plugin/src/Ide/Plugin/Class/ExactPrint.hs
Aster89: Wait a sec, maybe we rushed too much into `ghcVersion`.

Using it to branch can't work because it branches at run time, and the types of `then` and `else` are different!

We'd need something that branches at compile time.

Do I fall back to using the `#if` approach, or have I misunderstood your suggestion?

fendor: I think I misunderstood the code changes, if the types are different, then it needs to be `#if`, sorry for the confusion!

fendor (approved):
LGTM, thank you!

PR #5027, opened 2026-07-24
Guard against leaking threads while restarting shake

crtschin:
Alternative variant of https://github.com/haskell/haskell-language-server/pull/5021.

Third time's the charm.

Instead of keeping track and killing threads via the shake database, do this locally in the `AIO` abstraction that does cleanup. Once shutdown, the `AIO` is closed, so threads that are in the process of being spawned get killed instead.


hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -300,27 +302,57 @@ transitiveDirtySet database = flip State.execStateT mempty . traverse_ loop
...
 asyncWithCleanUp :: AIO a -> AIO (IO a)
 asyncWithCleanUp act = do
     st <- AIO ask
     io <- unliftAIO act
     -- mask to make sure we keep track of the spawned async
     liftIO $ uninterruptibleMask $ \restore -> do
         a <- async $ restore io
-        atomicModifyIORef'_ st (void a :)
-        return $ wait a
+        registered <- registerAsyncs st [void a]
```
soulomoon: should only run on succ registeration instead of finding out the scope have ended and killing it ?

soulomoon: e.g. async thread should know if itself is registered, it only run the body if it is registered or do nothing otherwise.

soulomoon:
`waitConcurrently_` should apply the same `run the body only if registered` pattern.
And we should not spread the throw everywhere and trigger the expection handler multiple times. I've made some ajustments. @crtschin feel free to revert any unsensible changes.

soulomoon:
I am seeing some unintended failure in the benchmark result, there must be something I am missing.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -368,12 +422,18 @@ waitConcurrently_ many = do
...
-    -- work on the sync computations
-    liftIO $ sequence_ syncs
-    -- wait for the async computations before returning
-    liftIO $ traverse_ wait asyncs
+        registered <- liftIO $ registerAsyncs ref asyncs
+        putMVar gate registered
+        return (asyncs, syncs, registered)
+    -- A closed scope means our asyncs were cancelled, so abort the superseded work.
+    if not registered
+      then liftIO $ (forever sleep 10) -- wait for the asyncs to finish, but don't block the main thread
```
crtschin: There's a misunderstanding here I think, there is no main thread. Calling `wait` only suspends the green thread, leaving the GHC capability to continue on with other work. 

Calling `forever (sleep x)` will have the green thread keep making the green thread available to be picked up by a capability every x seconds, only to subsequently yield again. 

Why wait on a different thread to kill this thread, when this thread already locally knows it should be interrupted.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -300,27 +302,77 @@ transitiveDirtySet database = flip State.execStateT mempty . traverse_ loop
...
+    gate <- newEmptyMVar
+    a <- runAsyncIfRegistered gate io
+    registered <- registerAsyncs st [void a]
+    putMVar gate registered
+    return (registered, a)
+  return $ if registered
+              then wait a
+              -- we don't want to throw an exception here because the async is already running and will be cancelled by the scope,
+              -- so we just wait for it to finish
+              else (forever $ sleep 10)
```
crtschin: Same applies here as [here](https://github.com/haskell/haskell-language-server/pull/5027/changes#diff-be9e854b0711afcb729de4ed6c3df56fe1aa6f358af7e750c028cc41ede1c462R434).

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -300,27 +302,77 @@ transitiveDirtySet database = flip State.execStateT mempty . traverse_ loop
...
+
+runAsyncIfRegistered :: MVar Bool -> IO a -> IO (Async a)
+runAsyncIfRegistered gate io =
+    asyncWithUnmask $ \unmask -> unmask $ do
+      b <- readMVar gate
+      -- Only if the thread was successfully registered do we execute
+      if b then io else (error "asyncWithCleanUp: scope closed before thread could be spawned")
```
crtschin: IIRC, calling `error` makes GHC throw an IOError, which is less easy to work with than `AsyncCancelled` which is specifically intended to indicate that a async green thread didn't finish the work it intended.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -300,27 +302,77 @@ transitiveDirtySet database = flip State.execStateT mempty . traverse_ loop
...
-        return $ wait a
+  st <- AIO ask
+  io <- unliftAIO act
+  -- mask so the spawn and registration can't be split by interrupt
+  (registered, a) <- liftIO $ uninterruptibleMask_ $ do
+    -- Use a signal to indicate whether the thread is being spawned into a
+    -- open/closed scope.
+    gate <- newEmptyMVar
+    a <- runAsyncIfRegistered gate io
+    registered <- registerAsyncs st [void a]
```
crtschin: I believe there's a logical gap here now. If the AIO is interrupted between these lines, the async spawned isn't killed. 

Notice that `runAsyncIfRegistered` has a check, and `registerAsyncs` has a subsequent use. If cleanup fires after the `runAsync*` call, the lock passes and spawns the async, but the registration fails because the scope is closed, orphaning the thread.

crtschin (commented):
> waitConcurrently_ should apply the same run the body only if registered pattern.

For individual threads I understand, as it's arbitrary side-effecting user code, but why for `waitConcurrently_` as well? Is there an error condition you see here? 

I've kept a variant of your changes in the PR, but I don't understand the need yet.

> And we should not spread the throw everywhere and trigger the ..exception handler multiple times.

Exception handlers are quite cheap in Haskell. If there's a broken precondition and a thread needs to halt, it's more efficient to have the thread immediately die on its own than wait for a different thread to kill it, which can also be quite tricky if that supervisor thread already died.

> I've made some adjustments. 

I gave this a review, and pushed some small edits.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
soulomoon: Yes, the comment is wrong(accidentally ai completion I forget to delete). 

The main idea is there is a single place we `cancel` the thread(`cleanupAsync`).
And if the thread can not be registered, just let the thread finish without doing anything instead of killing it.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
soulomoon: If cleanup fires after the runAsync* call, doesn't it mean we would cleanup the async during `cleanup` ?

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -392,25 +389,19 @@ cleanupAsync ref = uninterruptibleMask $ \unmask -> do
     -- Close the scope so no later force can register, and take the live asyncs
     -- to interrupt. See Note [Closing escaped rule computations].
     asyncs <- atomicModifyIORef' ref $ \m -> (Nothing, fromMaybe [] m)
-    -- interrupt all the asyncs without waiting
-    mapM_ (\a -> throwTo (asyncThreadId a) AsyncCancelled) asyncs
```
soulomoon, line 342: `AsyncCancelled ` should still happen at once.  The normal fine scoped threads should be ended immediatly here once we call `cleanupAsync`.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
soulomoon: perhaps just use `undefined` ? Since we should never inspect the result after the gate shows that it leave its' scope. And the thread should simply do nothing then finish if it knows it is out of scope.

soulomoon:
> > waitConcurrently_ should apply the same run the body only if registered pattern.

> For individual threads I understand, as it's arbitrary side-effecting user code, but why for waitConcurrently_ as well? Is there an error condition you see here?

Notice we register them into the AIO later. So they are AIO async too and should be managed by AIO.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
crtschin: `undefined` is the same as `error` in this case. `AsyncCancelled` still makes more sense to me here. The thread hit a invariant, so should terminate (safely). Things we definitely want to avoid are hangs, and exceptions that would otherwise cause HLS to crash hard, which things like "undefined" or "error" open the app up to.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
crtschin: Yes, I meant if we didn't cancel these unregistered threads that land in between closing the scope in the cleanup, and the registration.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
crtschin: It still does this, just a bit lower in the function. 

The comment above this snippet. `interrupt all the asyncs without waiting` was not true as `throwTo` is a synchronous blocking call. I fixed this to use `mapConcurrently`, which does cancel all threads at the same time.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
soulomoon: we should only create two types of threads, 
the ones that are registered and canceled by cleanup.
the others that are not registered, should ended itself without doing anything.

we can choose not to kill the second ones, only let it finish.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
soulomoon: If it hangs or exceptions are invoked, we must be doing something wrong in our model, a hard crash is what we wanted so we can fix them.

soulomoon:
necessary invariant:
Only registered threads execute ~force or~ rule bodies.
Registered threads are cancelled exclusively by cleanupAsync.
Rejected threads perform no work and terminate themselves.

crtschin:
I tweaked the abstractions a bit over the last few weeks.

- Made a new Scope type that encapsulates the "scope" threads are spawned in via AIO.
- The above uses a lock instead of `IORef` to lock registration instead of trying to fix it adhoc using atomic `IORef` interactions.
- Make it possible to distinguish between a thread close due to an exception vs the scope being closed.
- It'll now dirty the key automatically if the scope is closed while the key was being processed.

I think the current implementation matches your invariants as well.

soulomoon:
## Bug reproduction

The test `rapid edits then save does not strand a stale diagnostic` was run 1,000 times on each exact implementation:

| Commit | Implementation diff | Passed | Failed | Result |
|---|---|---:|---:|---|
| [`b7e5d2526`](https://github.com/haskell/haskell-language-server/commit/b7e5d25260f7a8ff5df4590bfb2b3a683ac07d42) | Baseline | 993 | **7** | Bug reproduced |
| [`5a250886c`](https://github.com/haskell/haskell-language-server/commit/5a250886c216471bd401bf2ce70f74512953f4d1) | [View diff](https://github.com/haskell/haskell-language-server/pull/5027/files) | **1,000** | 0 | No reproduction |
| [`48f297100`](https://github.com/soulomoon/haskell-language-server/commit/48f297100360281fa97d1f9220b23bf76589c85d) | [View diff](https://github.com/soulomoon/haskell-language-server/pull/38/files) | **1,000** | 0 | No reproduction |

Both fix commits prevent the behavior exercised by this regression test.

> `b7e5d2526` and `48f297100` were tested with only the unchanged regression test added; their implementation code was otherwise unchanged.

## Benchmark comparison

Commit `b7e5d2526` is the benchmark baseline.

- **[`5a250886c`](https://github.com/haskell/haskell-language-server/commit/5a250886c216471bd401bf2ce70f74512953f4d1)** ([diff](https://github.com/haskell/haskell-language-server/pull/5027/files), [benchmark run](https://github.com/haskell/haskell-language-server/actions/runs/31282731150)):
  - Generally close to baseline.
  - `code actions after cradle edit` is approximately **2.3–2.6× slower**:
    - GHC 9.12: **+126.6%**
    - GHC 9.14: **+164.3%**
  - Overall geometric-mean `totalT` change: **+1.02%**

- **[`48f297100`](https://github.com/soulomoon/haskell-language-server/commit/48f297100360281fa97d1f9220b23bf76589c85d)** ([diff](https://github.com/soulomoon/haskell-language-server/pull/38/files), [benchmark run](https://github.com/soulomoon/haskell-language-server/actions/runs/31268638655)):
  - Remains close to baseline and avoids the cradle-edit regression:
    - GHC 9.12: **+2.5%**
    - GHC 9.14: **−0.9%**
  - Overall geometric-mean `totalT` change: **−0.57%**

Excluding the cradle-edit outlier, the two fix commits are effectively tied. Overall, both fix the tested bug, but **`48f297100` has the better benchmark result**.

---------above comparison generated by codex -------

@crtschin The newest head introduce some slowdown, I think it is due to the extra complexity from database repairing and a few places which blurred the ownership model, while the old method stayed minimalized and avoid performance regression.

Do you mind use [View diff](https://github.com/soulomoon/haskell-language-server/pull/38/files) instead.

Also I think we should shape the test `rapid edits then save does not strand a stale diagnostic` better, so we can have a more deterministic failure result.

crtschin:
Thanks for the benchmark run @soulomoon!, I removed an optimization while I was refactoring for the sake of uniformity (the setup runs singular actions inline, instead of separately spawning a thread). I re-added it and now get similar performance to the baseline again now, when running the benchmark locally.

> Do you mind use [View diff](https://github.com/soulomoon/haskell-language-server/pull/38/files) instead. 

I'm hesitant, the main difference is in how a thread who's scope is closed, terminates. The branch does a [loop](https://github.com/soulomoon/haskell-language-server/pull/38/changes/357d0644f0955c71c2855857a42ad5e4910fad04#diff-be9e854b0711afcb729de4ed6c3df56fe1aa6f358af7e750c028cc41ede1c462R384-R389) expecting the thread that initialized the scope to land the `cancel` cleanly. This is cumbersome and tricky, as nothing inherently guarantees that. The risk here being zombie threads that are orphaned and endlessly take up resources.

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -300,27 +337,90 @@ transitiveDirtySet database = flip State.execStateT mempty . traverse_ loop
...
+  3. A second build at step S opens scope-2, sees that entry, waits on the
+     thunk, and forces it on scope-2's thread.
+
+'spawnInScope' decides whether to spawn while holding the scope open,
+so a refused computation is never started and the forcing thread raises
+'ScopeClosed'.
+  * 'repairRefusal' demotes the key to 'Dirty' before re-throwing.
+  * 'build' retries once, so the demoted key recomputes in a live scope.
+  * 'isAsyncException' classes it async, so it isn't swallowed.
+-}
```
soulomoon, line 391: if there is an exception that is not caused by session restart, it should be a bug for the hls-graph, we should not repair it. see `defineEarlyCutoff'` and `actionCatch`, there we catch all the rule's errors and leave only the `isAsyncException` to surface to the hls-graph to handle.

soulomoon:
> The risk here being zombie threads that are orphaned and endlessly take up resources.

If there are zombie threads, there must be some other bugs hidden in other part of the codebase which we should also rule out. It is better to do just enough to reveal the problem than over do it and cover an unstable problem up.

crtschin:
> there must be some other bugs hidden in other part of the codebase which we should also rule out.

Proving the absence of bugs isn't possible.

> It is better to do just enough to reveal the problem than over do it and cover an unstable problem up.

I don't understand how having a thread hang is a better solution than having the thread throw. The throw is more immediate and visible (and reportable by a user). The hang is the one that invisibly worsens the state (where the main symptom would be increased memory usage and performance degradation).

soulomoon:
> I don't understand how having a thread hang is a better solution than having the thread throw. The throw is more immediate and visible (and reportable by a user). The hang is the one that invisibly worsens the state (where the main symptom would be increased memory usage and performance degradation).

My point is not about a hang is better than a throw, I am sorry for giving you that impression. What I am trying to say is that to be able to observe the fact that the scope miss the kills, the parent should not throw immediatly once it sees a failed spawning due to scope closed since it cover up the problems(We do not know wether the scope kill is just late or won't happen). I agree it is not very ideal to let it hang, perhaps a better way is, wait a bit for the scope kills and if that did not happen, log something out and throw.

soulomoon:
Here is the newest perf comparison after the sync to the newest master
I realized can further optimize away the `asyncWithCleanUp`.  I opened pr #5039 to integrate that part. Then we do not have to fix the asyncwithcleanup for the scope issue. And focus on `waitConcurrently_`

The old fix approach spawnning remain lockless, but the new one use a MVar as a lock to wait for the spawn, might be  a bit slower. But usualy only one child for the same scope would be spawned in the same time, so perf degradation might not be obivious.

> 
> ## Updated benchmark comparison
> 
> Both implementations are based on the same master commit, [`16bf046629`](https://github.com/haskell/haskell-language-server/commit/16bf0466293daabac6f216d2af7e5bccbb91fcc1).
> 
> | Exact head | Diff | Geometric-mean `totalT` vs master | Summed `totalT` |
> |---|---|---:|---:|
> | [`3108a99827`](https://github.com/haskell/haskell-language-server/commit/3108a99827c95707eb06fc03f8dda96d4aa91d7d) | [View diff](https://github.com/haskell/haskell-language-server/pull/5027/files) | −0.53% | −1.04% |
> | [`480fcf8bcc`](https://github.com/soulomoon/haskell-language-server/commit/480fcf8bcc2c2c5370f2ffb7a33d830caf459707) | [View diff](https://github.com/soulomoon/haskell-language-server/pull/38/files) | **−4.83%** | **−4.42%** |
> 
> Across 83 common successful benchmark cases, `480fcf8bcc` is:
> 
> - **4.21% faster geometrically**
> - **2.83% faster by median**
> - Faster in **57 of 83** cases
> 
> ### `code actions after cradle edit`
> 
> | Configuration | `3108a99827` | `480fcf8bcc` |
> |---|---:|---:|
> | GHC 9.12 | +1.71% | **−6.36%** |
> | GHC 9.14 | −1.25% | **−5.79%** |
> 
> Benchmark runs:
> 
> - [`3108a99827` benchmark](https://github.com/haskell/haskell-language-server/actions/runs/31498975069)
> - [`480fcf8bcc` benchmark](https://github.com/soulomoon/haskell-language-server/actions/runs/31499046206)
> 
> **Conclusion:** both implementations are slightly faster than master, but `480fcf8bcc` has the better overall performance.
>

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
```diff
@@ -300,27 +337,90 @@ transitiveDirtySet database = flip State.execStateT mempty . traverse_ loop
...
+-- The wait is returned rather than run, so the lock is free before anyone
+-- blocks on a result.
+--
+-- See Note [Closing escaped rule computations].
+spawnInScope :: Scope -> IO a -> IO (IO a)
+spawnInScope Scope{scopeClosing, scopeAsyncs} io =
+    mask_ $ modifyMVar scopeAsyncs $ \case
+        Nothing -> pure (Nothing, refuse)
+        Just as -> do
+            closing <- readIORef scopeClosing
```
soulomoon: do we still need scopeClosing if we have Mvar scopeAsyncs ?

crtschin: Nice, I don't think so no

hls-graph/src/Development/IDE/Graph/Internal/Database.hs
soulomoon: That means if build does not compelete, it must be a session restart, we should not retry it.

crtschin:
@soulomoon I like the simplification in https://github.com/haskell/haskell-language-server/pull/5039! I'll wait with continuing with this PR until that ones resolved. I think I can drop some code here once that lands.

soulomoon:
ready to push it forward ?  @crtschin

PR #5028, opened 2026-07-25
Fix invalid non-singleton constraint suggestion edits

### LINKED ISSUE #3486: Syntax is messed up when adding a constraint to quantified constraints
### Your environment

Which OS do you use?

Nixos unstable

9.2.4 installed via nix

cabal
https://github.com/turion/essence-of-live-coding

vscodium + Haskell 2.2.2

1.9.0.0 via nix

The editor plugin is configured to use fourmolu

### Steps to reproduce

Write a function with quantified constraints in the type signature

image

A constraint is missing, open menu and ask to add the constraint

image

The constraint is added, but the syntax of the constraints is broken: Some important brackets are removed.

image

### Expected behaviour

The missing constraint is added without breaking the syntax

### Actual behaviour

The syntax is broken, brackets are removed in places where nothing should have changed

crtschin:
Ref: https://github.com/haskell/haskell-language-server/issues/4605
Closes: https://github.com/haskell/haskell-language-server/issues/3486

Makes the following changes
- Fixes the incorrect constraint edit in constraint lists with extra parentheses.
- Adds a lens to pull missing constraints from diagnostics.
- Use the diagnostic ranges to deduce where to add the constraints, instead of regex matching.

### DIFF
--- a/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
+++ b/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
@@ -50,8 +50,11 @@ import           Development.IDE.Core.Shake                        hiding (Log)
 import           Development.IDE.GHC.Compat                        hiding
                                                                    (ImplicitPrelude)
 import           Development.IDE.GHC.Compat.Error                  (TcRnMessage (..),
+                                                                    _CouldNotDeducePred,
                                                                     _TcRnMessage,
-                                                                    msgEnvelopeErrorL)
+                                                                    _TcRnSolverReport,
+                                                                    msgEnvelopeErrorL,
+                                                                    reportContentL)
 import           Development.IDE.GHC.Compat.Util
 import           Development.IDE.GHC.Error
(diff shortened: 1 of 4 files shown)


plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -147,6 +151,11 @@ import           GHC.Tc.Types.CtLoc                                (ctl_env,
 #endif
 
 import           GHC.Tc.Types.Origin                               (CtOrigin (LiteralOrigin))
+import           Language.LSP.Protocol.Lens                        hiding
+                                                                   (codeAction,
+                                                                    error,
+                                                                    length,
+                                                                    parent)
```
fendor: Usually we prefer importing the Lens stuff qualified e.g.

```suggestion
import qualified Language.LSP.Protocol.Lens                        as L
```

Or `as J`, I never remember which one it is.

fendor (approved):
LGTM, I have one suggestion, please merge once you are happy.

PR #5029, opened 2026-07-30
Improve error messages in HLS

fendor:
Use more lines to make exceptions more readable once they occur.

For debugging, also attach IPEBacktraces if HLS is compiled with `-finfo-table-map` on GHC >=9.10.

Turns the error:

image

Into 

image

And with `-finfo-table-map` we get something actually useful:

image


crtschin (approved):
Very nice!

PR #5031, opened 2026-08-02
cross-module renames

### LINKED ISSUE #4816: `hls-rename-plugin`: Rename across modules fails

Esgariot:
Fixes #4816, partially

This PR could hopefully serve as a stopgap until #4845 is merged.
Additionally it fixes an edge-case where there are multiple rename terms in the same line,
but still suffers from the existing global vs local term rename problems that the other PR fixes.

### DIFF
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
@@ -20,7 +20,8 @@ import           Data.Generics
 import           Data.Hashable
 import           Data.HashSet                          (HashSet)
 import qualified Data.HashSet                          as HS
-import           Data.List.NonEmpty                    (NonEmpty ((:|)), groupAllWith)
+import           Data.List.NonEmpty                    (NonEmpty ((:|)),
+                                                        groupAllWith)
 import qualified Data.Map                              as M
 import           Data.Maybe
 import           Data.Mod.Word
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
@@ -226,7 +226,7 @@ refsAtName state nfp name = do
         Nothing -> pure []
         Just mod -> liftIO $ mapMaybe rowToLoc <$> withHieDb (\hieDb ->
             -- See Note [Generated references]
-            filter (\(refRow HieDb.:. _) -> refIsGenerated refRow) <$>
+            filter (\(refRow HieDb.:. _) -> not $ refIsGenerated refRow) <$>
             findReferences
                 hieDb
                 True
--- a/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
+++ b/plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs
@@ -20,8 +20,7 @@ import           Data.Generics
 import           Data.Hashable
 import           Data.HashSet                          (HashSet)
 import qualified Data.HashSet                          as HS
-import           Data.List.NonEmpty                    (NonEmpty ((:|)),
-                                                        groupWith)
+import           Data.List.NonEmpty                    (NonEmpty ((:|)), groupAllWith)
 import qualified Data.Map                              as M
 import           Data.Maybe
 import           Data.Mod.Word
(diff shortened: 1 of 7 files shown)


fendor:
Hi, thank you for your PR!

Can you explain how it fixes the issue and what is still missing? Also, any benefit to merge this before https://github.com/haskell/haskell-language-server/pull/4845 or is it the hope that we can review this PR quicker?

Esgariot:
```diff
- filter (\(refRow HieDb.:. _) -> refIsGenerated refRow) <$>
+ filter (\(refRow HieDb.:. _) -> not $ refIsGenerated refRow)
```

condition flip, the intention there was to filter out the generated references, at least that's what the note says in 

```
{- Note [Generated references]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
GHC inserts `Use`s of record constructor everywhere where its record selectors are used,
which leads to record fields being renamed whenever corresponding constructor is renamed.
see https://github.com/haskell/haskell-language-server/issues/2915
To work around this, we filter out compiler-generated references.
-}
```

but instead only the generated refs made it in.
(#4845 also has this fix)

---

```diff
- collectWith :: (Hashable a, Eq b) => (a -> b) -> HashSet a -> [(b, HashSet a)]
- collectWith f = map (\(a :| as) -> (f a, HS.fromList (a:as))) . groupWith f . HS.toList
+ collectWith :: (Hashable a, Ord b) => (a -> b) -> HashSet a -> [(b, HashSet a)]
+ collectWith f = map (\(a :| as) -> (f a, HS.fromList (a:as))) . groupAllWith f . HS.toList
```

the `Location` is "uri + range", so 
1. when `filesRefs = collectWith locToUri refs` was called, 
2. now that more than one file was considered (`refs <- ... refsAtName ...` now is adding correct `dbRefs`, see first diff at the top)
3. when trying to group them by uri in `collectWith`, they were producing disjointed results because they weren't sorted by uri
4. so there were duplicate uri's in `fileRefs`
5. so only the last file edit for a group applied, but only in cases when there were multiple refs to be renamed in single line 

#4845 doesn't have this fix, so this pr is not strictly a subset of that other pr

---

I should probably also try the test cases against the former PR, and try to check if the `collectWith` edit fixes anything, or whether it's handled differently, maybe by more direct AST parsing that PR is doing.

If there is a value in `collectWith` fix by itself then I think it makes sense to merge this one, and I'll keep an eye on #4845 to adapt my test cases so they can use the `goldenWithCrossModuleRename`  

--- 

disclaimer: used AI for analysis, so I don't really know e.g. how `replaceRefs`, `diffText` etc ultimately apply the edits (grouping edits by line, etc), but these two minimal diffs above I am pretty sure about, and the test cases are really simple. Also, sorry if I'm being too chaotic in my comments and it's unreadable, I tend to go into edit spiral instead of just making new separate comment -_-

fendor:
Thank you for the summary, this seems to make sense to me.

Sure, try the other test cases as well, perhaps it already fixes most of them.

fendor:
It would be interesting to see whether this fixes this issue as well: https://github.com/haskell/haskell-language-server/issues/5032

Since we are filtering out the generated notes now

Esgariot:
Hmm, the tests from #4845 are not deterministic, up to 3 of 6 cross-module tests failing on my machine :<

but limiting it to the single test introduced in here, when applying the patch that adds `CrossModule(Definition|Usage).hs` (`--pattern "/CrossModule"`) to #4845, it always fails there unless I also apply the `collectWith` fix, then it always passes. 

So yeah, now I feel more confident that this PR can be merged on its own before #4845, and I should only rewrite my test to use [`goldenWithCrossModuleRename`](https://github.com/vidit-od/haskell-language-server/blob/7ff135093c4d733f1318972cd80fb6e88b8af2ad/plugins/hls-rename-plugin/test/Main.hs#L246-L266) in the future, independently of this PR.

I can take a look at repro of #5032 in a spare moment

fendor:
Could you also format the file using `stylish-haskell -i plugins/hls-rename-plugin/src/Ide/Plugin/Rename.hs`?

That would fix the pre-commit hook.

fendor (approved):
LGTM, thank you! 

I will wait for your confirmation you are happy with the patch as is before merging.

Esgariot:
@fendor I think it's ready to be merged, thx for reviewing!

PR #5033, opened 2026-08-05
Stop unloading all bytecode whenever we generate new bytecode

wz1000:
We were attempting to keep the bytecode, but we only made dummy object
linkables to keep, which is pointless since we don't generate object code in
HLS and always use bytecode.

Also create dummy bytecode linkables so that `unload` knows not to unload
things that are up to date.


fendor (approved):
LGTM!

PR #5035, opened 2026-08-06
Unload linkables if any transitive dependency changes

wz1000:
Take transitive dependencies into account when unloading

In 7d9ccb70be66a9118eee88505b0455174752d184 we finally stopped unloading
all bytecode on every call to GetLinkable. However, this exposed the fact
that our unloading strategy was not sound

In particular, if you have
```haskell
-- A.hs
f = 1
```

```haskell
-- B.hs
import A
g = f + 1
```

If you change A (say f=2), you must unload/reload B as well, otherwise
its bytecode/linkable will still be pointing to the old version of A
with f=1

We do this with two changes:

- GetLinkable now depends on GetLinkable for direct depdencies
  - So it is recomputed whenever a transtive dependency changes (by induction)
  - It's hash includes the hash of all its direct dependencies (transitive by induction)
- When GetLinkable is recomputed, we unload the old version of the linkable
  - Now we identify linkables by hash of the linkable + all the (transitive) dependencies,
    so we can tell GHC exactly what it needs to keep/vs what it needs to unload
  - This involves a horrid hack, see versionToUTCTime

### DIFF
--- a/ghcide/src/Development/IDE/Core/Rules.hs
+++ b/ghcide/src/Development/IDE/Core/Rules.hs
@@ -93,9 +93,7 @@ import           Data.Proxy
 import qualified Data.Text                                    as T
 import qualified Data.Text.Encoding                           as T
 import qualified Data.Text.Utf16.Rope.Mixed                   as Rope
-import           Data.Time                                    (Day (ModifiedJulianDay),
-                                                               UTCTime (..),
-                                                               picosecondsToDiffTime)
+import           Data.Time                                    (UTCTime (..))
 import           Data.Tuple.Extra
 import           Data.Typeable                                (cast)
 import           Development.IDE.Core.Compile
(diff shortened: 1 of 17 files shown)


ghcide/src/Development/IDE/Core/Rules.hs
```diff
@@ -1123,21 +1124,34 @@ usePropertyByPathAction path plId p = do
...
+-- | A horrible hack
+-- UTCTime is an unbounded integer
+-- We can smuggle an arbitary 128 bit fingerprint
+-- in the integer
+--
+-- This is fine because GHC only uses the time for identity
+versionToUTCTime :: Fingerprint -> UTCTime
+versionToUTCTime (Fingerprint a b) =
+  UTCTime (ModifiedJulianDay 0) $ picosecondsToDiffTime $
+    toInteger a * 2 ^ (64 :: Int) + toInteger b
```
fendor: For documentation reasons, I think it would be nice if this was a newtype, e.g.

```
newtype LinkableFingerprint = LinkableFingerprint UTCTime
```

And then pass `LinkableFingerprint` to `generateObjectCode`

wz1000: done.

fendor (approved):
LGTM :)

PR #5041, opened 2026-08-13
Simplify `activeDiagnosticsInRange` by returning `[FileDiagnostic]` instead of `Maybe [FileDiagnostic]`

Aster89:
The function `activeDiagnosticsInRange` returns `MonadIO m => m (Maybe [FileDiagnostic])`, but no callers appear to make any distinction between `Just []` and `Nothing`, so we can probably just drop the `Maybe` layer alltogether? Tests will tell.

### DIFF
--- a/plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
+++ b/plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
@@ -92,11 +92,9 @@ codeActionHandler recorder plId ideState _ CodeActionParams{_textDocument, _rang
     nfp <- getNormalizedFilePathE uri
     decls <- getDecls plId ideState nfp
 
-    activeDiagnosticsInRange (shakeExtras ideState) nfp _range >>= \case
-        Nothing -> pure (InL [])
-        Just fileDiags -> do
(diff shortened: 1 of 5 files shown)


plugins/hls-change-type-signature-plugin/src/Ide/Plugin/ChangeTypeSignature.hs
```diff
@@ -92,9 +92,7 @@ codeActionHandler recorder plId ideState _ CodeActionParams{_textDocument, _rang
     nfp <- getNormalizedFilePathE uri
     decls <- getDecls plId ideState nfp
 
-    activeDiagnosticsInRange (shakeExtras ideState) nfp _range >>= \case
-        Nothing -> pure (InL [])
-        Just fileDiags -> do
+    activeDiagnosticsInRange (shakeExtras ideState) nfp _range >>= \fileDiags -> do
```
MangoIV:
```suggestion
     fileDiags <- activeDiagnosticsInRange (shakeExtras ideState) nfp _range
```

MangoIV (approved):
Seems like none of the use sites use the distinction between [] and Maybe . [] here, so looks good to me.

fendor (approved):
Lgtm, thank you!

PR #5044, opened 2026-08-14
Put the cwd shifts in tests behind an opt-in

crtschin:
Part of https://github.com/haskell/haskell-language-server/pull/4977.

This PR makes it explicit which testsuites depend on being able to shift the current working directory, related https://github.com/haskell/haskell-language-server/issues/4234.

### DIFF
--- a/ghcide/session-loader/Development/IDE/Session/Ghc.hs
+++ b/ghcide/session-loader/Development/IDE/Session/Ghc.hs
@@ -262,7 +262,7 @@ setOptions haddockOpt cfp (ComponentOptions theOpts compRoot _) dflags rootDir =
     where
       initMulti unitArgFiles =
         forM unitArgFiles $ \f -> do
-          args <- liftIO $ expandResponse [f]
+          args <- liftIO $ expandResponse [rebaseResponseFile compRoot f]
           -- The reponse files may contain arguments like "+RTS",
           -- and hie-bios doesn't expand the response files of @-unit@ arguments.
           -- Thus, we need to do the stripping here.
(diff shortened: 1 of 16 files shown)


0rphee (approved):
the stan changes lgtm as it seems to only be change from Bool to the new CwdHandling

ghcide/session-loader/Development/IDE/Session/Ghc.hs
```diff
@@ -309,6 +309,13 @@ setOptions haddockOpt cfp (ComponentOptions theOpts compRoot _) dflags rootDir =
 normaliseImportsPaths :: DynFlags -> DynFlags
 normaliseImportsPaths dflags = dflags { importPaths = fmap normalise (importPaths dflags)}
 
+-- | Rebase a relative response-file path onto the component root, so
+-- 'expandResponse' does not resolve it against the process CWD.
+rebaseResponseFile :: FilePath -> String -> String
+rebaseResponseFile root arg = case arg of
+  '@' : path -> '@' : toAbsolute root path
+  _          -> arg
```
fendor, line 323: I am wondering whether this is something we ought to do in hie-bios 🤔

hls-test-utils/src/Test/Hls.hs
```diff
@@ -879,6 +887,7 @@ runSessionWithTestConfig TestConfig{..} session =
                 , argsDefaultHlsConfig = testLspConfig
                 , argsProjectRoot = prjRoot
                 , argsDisableKick = testDisableKick
+                , argsDisableInitialCwdShift = True
```
fendor, line 890: What does this do in this context, we might shift the directory but ignore any `--cwd` arguments to the HLS server?

We need this because we want to lock the test until we can modify the global `$PWD` variable, right?

fendor (approved):
LGTM, two questions

hls-test-utils/src/Test/Hls.hs
crtschin: > What does this do in this context, we might shift the directory but ignore any --cwd arguments to the HLS server?

`--cwd` only affects running HLS for non-testing. This specifically stops the cwd change to [the workspace root](https://github.com/haskell/haskell-language-server/blob/838a51f7612478dfa458724449b12e5adb301393/ghcide/src/Development/IDE/Core/Shake.hs?plain=1#L535) that's done on initialization, which is necessary because these effect each other when running multiple servers concurrently in the same process.

> We need this because we want to lock the test until we can modify the global $PWD variable, right?

Yes, that's correct. Both the lock and cwd change are still necessary for some plugins.

ghcide/session-loader/Development/IDE/Session/Ghc.hs
crtschin: Makes sense! I'll make an issue.

Edit: Issue created [here](https://github.com/haskell/hie-bios/issues/529).

PR #5045, opened 2026-08-16
WIP: support for GHC 9.14.2 (rc)

guibou:
This is a work in progress in order to get something which runs with 9.14.2. For now, the correct CPP guard are not implemented (because I don't know how to be robust here) and also because the API can still change.

### DIFF
--- a/ghcide/src/Development/IDE/Import/FindImports.hs
+++ b/ghcide/src/Development/IDE/Import/FindImports.hs
@@ -36,6 +36,10 @@ import           GHC.Generics
 import           GHC.Types.PkgQual
 import           GHC.Unit
 
+#if MIN_VERSION_ghc(9,14,1)
+import qualified GHC.Types.Unique.Set              as UniqSet
+#endif
+
 
 #if MIN_VERSION_ghc(9,11,0)
 import           GHC.Driver.DynFlags               (ReexportedModule (..),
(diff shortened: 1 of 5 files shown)


Bodigrim:
> because I don't know how to be robust here

Could you use [`__GLASGOW_HASKELL_PATCHLEVEL{1,2}__`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/phases.html#standard-cpp-macros)?

ghcide/src/Development/IDE/Import/FindImports.hs
```diff
@@ -229,12 +237,16 @@ locateModule moduleMaps env unit_visibility modName mbPkgName isSource = do
     -- several provide the module.
     -- The current unit's own reexports and hidden modules do not apply to it,
     -- which also stops the reexport search from looping.
+
+    -- TODO: here we convert a set into a list and then do n lookup (in
+    -- log(k), hence nlog(k).). However we may be able to do something smarter with
+    -- "restrictKeys" which runs in nlog(n/k + 1).
```
alexfmpe, line 243: Welp, this nerd sniped me big time.
I thought "_why not simply convert both the map and set to lists and then work through the heads of both lists similarly to the merge part of mergesort_" and fell down quite a rabbit hole: https://github.com/haskell/containers/pull/1256

guibou: Haha, sorry ;) Super interesting discussion and research ! Thank you.

PR #5047, opened 2026-08-19
Change `rangesOverlap` to honor the fact that `Range`s are inclusive to the left but exclusive to the right

### LINKED ISSUE #5049: `rangesOverlap` is not LSP-compliant
`rangesOverlap` is currently implemented like follows, as per #4433 by @noughtmare:

https://github.com/haskell/haskell-language-server/blob/a4cfaa80ca94beded6f01547a161b37be7b33558/hls-plugin-api/src/Ide/PluginUtils.hs#L284-L294

The fact that it uses `<=` instead of `<` means that
```haskell
λ> a = Position 1 1
λ> b = Position 1 3
λ> c = Position 1 5
λ> ab = Range a b
λ> bc = Range b c
λ> rangesOverlap ab bc
True
```
which is in contrast with [the LSP definition of `Range` as inclusive to the left and exclusive to the right](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#range) (my emphasis):

> ### Range
>
> A range in a text document expressed as (zero-based) start and end positions. A range is comparable to a selection in an editor. Therefore, ***the end position is exclusive***. If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line. For example.

In other words, the code above should return `False`.

One consequence of not doing so is the this:

```haskell
--                  Here's the "non-exhaustive patterns" diagnostic,
--        vvvvvvvvv and the "disable warning" action is available here, as I expect,
foo x =   case x of
--       ^          but it is available here too, where it should not.
```

I gave a try at changing `<=` to `<` in #5047, but several tests break; most would probably be fixed by simply changing some hard-coded number, and I've indeed verified this is the case in one instance, but I wouldn't be too optimistic, so before proceeding with fixing the broken tests, I'd like some feedback.

One important aspect to consider, I believe, is that despite this change would be a behavioral one, most mouse users, I think, wouldn't even notice it, as it just means that the "disable warning" action (and many other "disable" actions?) would be available one character "later" than they are now. _**I think**_.

@mango agrees with the fix I'm proposing.

Any more feedback is welcome.

Aster89:
I brainstormed what follows, and therefore the intended solution, with @MangoIV

### The setting

Consider this code
```haskell
    foo = bar + baz
```
and imagine that there's two adjacent diagnostics on it¹:
```haskell
--  vvvvvv <---------- diagnostic 1
    foo = bar + baz
--        ^^^^^^^^^ <- diagnostic 2
```
As per the LSP spec, the `Range`s of these diagnostics are (`l` is the 0-based line number):
```haskell
Range (Position l  4) (Position l 10)
Range (Position l 10) (Position l 19)
```
Since all happens on line `l`, I'll use the mathematical notation for right-open intervals:
```none
r1 = [ 4, 10)
r2 = [10, 19)
```
Now imagine the following two scenarî.

### Scenario 1 - the cursor is between the diagnostics

***The cursor***, intended as the thin vertical bar that most GUI-based editors (as opposed to, say, Vim) show, ***is between the `b` of `bar` and the space before it***², and the only way the client can hope to tell this to the server is to send the range `[4, 4)`, which is precisely what VSCode does, for instance.

In this case, the cursor is in a pefectly symmetrical position with respect to the two diagnostics, so there's absolutely no reason to consider it overlapping with only one diagnostic or only with the other.

We either claim it's overlapping with both, or we claim it's overlapping with none.

For that, the inherently asymmetrical interpretation of `Range`s as half-open comes short, and needs changing:

  1. if we wanted the cursor to overlap with neither diagnostic, we'd have to consider their `Range`s to be open on both sides,

  2. if we wanted the cursor to overlap with both diagnostics, we'd have to consider their `Range`s to be losed on both sides.

### Scenario 2 - the second diagnostic is selected

***The user has selected the second diagnostic***, in which case the client sends HLS the range `[10, 19)`, verified on VSCode.

In this case, I think it's only reasonable that we provide interaction with that diagnostic, and not with the other one.

For that, the interpretation of `Range`s as half-open comes handy, because adjacent half-open ranges have no overlap.

### The solution

The solution supports both scenarî, and for the first one, I decided to allow interaction with both diagnostics, as opposed to neither.

---

¹ I'm not aware of how to make HLS generate such diagnostics, but that's irrelevant, as adjacent diagnostics are not forbidden by the LSP spec.

² This is what the LSP spec refers to as "insert cursor", which is probably reminiscent of Vim's insert mode as opposed to normal mode?

---

Fixes #5049.

### DIFF
--- a/ghcide/src/Development/IDE/Core/PluginUtils.hs
+++ b/ghcide/src/Development/IDE/Core/PluginUtils.hs
@@ -56,7 +56,7 @@ import           Development.IDE.Types.Location       (NormalizedFilePath)
 import qualified Development.IDE.Types.Location       as Location
 import qualified Ide.Logger                           as Logger
 import           Ide.Plugin.Error
-import           Ide.PluginUtils                      (rangesOverlap)
+import           Ide.PluginUtils                      (asPosition, rangesOverlap)
 import           Ide.Types
 import qualified Language.LSP.Protocol.Lens           as LSP
 import           Language.LSP.Protocol.Message        (SMethod (..))
@@ -215,10 +215,22 @@ activeDiagnosticsInRangeMT ide nfp range = do
         case mDiags of
             Nothing -> pure Nothing
             Just fileDiags -> do
-                pure $ Just $ filter diagRangeOverlaps fileDiags
+                pure $ Just $ filter (diagRangeOverlaps range) fileDiags
     where
-        diagRangeOverlaps = \fileDiag ->
-            rangesOverlap range (fileDiag ^. fdLspDiagnosticL . LSP.range)
+        -- The rationale of how to decide whether a 'Range' overlap with the
+        -- 'Range' of a diagnostic is explained at
+        -- https://github.com/haskell/haskell-language-server/pull/5047
+        diagRangeOverlaps range fileDiag
+          | Just c <- asPosition range
+            -- The client sent us the cursor position, so we check if it
+            -- overlaps with the **closed** 'Range' of the diagnostic.
+            = LSP.positionInRange c diagRange || c == diagEnd
+          | otherwise
+            -- The client sent us a selection 'Range', so we check if it
+            -- overlaps with the 'Range' of the diagnostic.
+            = rangesOverlap range diagRange
+          where
+            diagRange@(LSP.Range _ diagEnd) = fileDiag ^. fdLspDiagnosticL . LSP.range
 
 -- | Just like 'activeDiagnosticsInRangeMT'. See the docs of 'activeDiagnosticsInRangeMT' for details.
 activeDiagnosticsInRange :: MonadIO m => Shake.ShakeExtras -> NormalizedFilePath -> LSP.Range -> m [FileDiagnostic]
--- a/hls-plugin-api/src/Ide/PluginUtils.hs
+++ b/hls-plugin-api/src/Ide/PluginUtils.hs
@@ -29,6 +29,7 @@ module Ide.PluginUtils
     installSigUsr1Handler,
     subRange,
     rangesOverlap,
+    asPosition,
     positionInRange,
     usePropertyLsp,
     -- * Escape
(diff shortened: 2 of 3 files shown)


ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -215,10 +215,13 @@ activeDiagnosticsInRangeMT ide nfp range = do
...
             Nothing -> pure Nothing
             Just fileDiags -> do
-                pure $ Just $ filter diagRangeOverlaps fileDiags
+                pure $ Just $ filter (diagRangeOverlaps range) fileDiags
     where
-        diagRangeOverlaps = \fileDiag ->
-            rangesOverlap range (fileDiag ^. fdLspDiagnosticL . LSP.range)
+        -- TODO: document this
+        diagRangeOverlaps range@(LSP.Range b e) fileDiag
+          | b /= e = rangesOverlap range diagRange
```
fendor:
```suggestion
          | isPointRange range = rangesOverlap range diagRange
```

where `isPointRange` checks exactly that.

fendor (approved):
LGTM, please squash the commits, and add the comment in your TODO :)

ghcide/src/Development/IDE/Core/PluginUtils.hs
```diff
@@ -215,10 +215,27 @@ activeDiagnosticsInRangeMT ide nfp range = do
...
+            = LSP.positionInRange c diagRange || c == diagEnd
+          | otherwise
+            -- The client sent us a selection 'Range', so we check if it
+            -- overlaps with the 'Range' of the diagnostic.
+            = rangesOverlap range diagRange
+          where
+            diagRange@(LSP.Range _ diagEnd) = fileDiag ^. fdLspDiagnosticL . LSP.range
+
+        asCursorPosition :: LSP.Range -> Maybe LSP.Position
+        asCursorPosition (LSP.Range b e)
```
fendor: it doesn't need to be a cursor position, so I think `asPosition` could be better

fendor (approved):

PR #5048, opened 2026-08-19
Keep only the core file hash in `HiFileResult`

crtschin:
A large part of `CoreFile` was kept explicitly lazy and evaluated only when needed for linkables. These thunks retained buffers the size of the core files, increasing resident memory. 

Consider [`readBinCoreFile`](https://github.com/haskell/haskell-language-server/blob/master/ghcide/src/Development/IDE/GHC/CoreFile.hs?plain=1#L50-L58), which uses [`readBinMem`](https://hackage-content.haskell.org/package/ghc-9.14.1/docs/src/GHC.Utils.Binary.html#readBinMem) and allocates a buffer the same size as the file. The subsequent `getWithUserData` uses [`lazyGet`](https://github.com/haskell/haskell-language-server/blob/master/ghcide/src/Development/IDE/GHC/CoreFile.hs?plain=1#L48) to defer parsing the binary into data, though keeping the buffer resident.

This PR changes the dependency load results to only contain the hash, and loads the core files on demand when needed in `GetLinkable`. The following is over a private codebase, using ghc-debug-brick after opening a module with a large transitive closure and waiting for HLS to settle.

```
Without:
	Maximum residency 1860.7 MB
	Label ARR_WORDS
	Count 642627
	Size 556.76 MB

With fix:
	Maximum residency 1633.4 MB
	Label ARR_WORDS
	Count 641547  
	Size 263.46 MB
```


wz1000 (approved):
This broadly makes sense, I didn't expect that keeping serialised core files in memory would be meaningfully expensive.

PR #5050, opened 2026-08-21
Delete unused `CodeAction.RuleTypes` module

crtschin:
Found this module while grepping for `HscEnvEq`. It isn't built anywhere, so deleting it.


fendor (approved):
LGTM, thank you!

PR #5058, opened 2026-08-27
Appropriately deal with files that don't end with a line terminator

### LINKED ISSUE #2998: Wrong alignment while adding placeholder for class instances
### Steps to reproduce

With the following code
```haskell
data A
instance Eq A where
```
Execute code action to add placeholder for `==` with signature

### Expected behaviour

```haskell
instance Eq A where
  (==) :: A -> A -> Bool
  (==) = _
```

### Actual behaviour

```haskell
instance Eq A where  (==) :: A -> A -> Bool
  (==) = _
```

---

It **only** happens while the instance definition is at the last line and there is no extra line later. It doesn't matter before we have `add instance signature code action` but will result in errors after #2920.

### LINKED ISSUE #5059: Class plugin breaks code when used at last line of a file that does not end with a line terminator
### Your environment

Which OS do you use? up-to-date ArchLinux

Which version of GHC do you use and how did you install it? 9.14.1 from GHCup

How is your project built (alternative: link to the project)? N/A

Which LSP client (editor/plugin) do you use? VSCode + the standard Haskell plugin, I think

Which version of HLS do you use and how did you install it? 2.14.0.0 from GHCup

Have you configured HLS in any way (especially: a `hie.yaml` file)? No

### Steps to reproduce

Create this file, say `ModuleA.hs`:
```haskell
module ModuleA where
data Foo = Foo
instance Show Foo where
```
**_making sure that there's no line terminator at the end of it_**, i.e. making sure that
```bash
git diff <(tail -1 WrongModuleA.hs | od -a) <(tail -1 ModuleA.hs | od -a)
```
```diff
 0000000   i   n   s   t   a   n   c   e  sp   S   h   o   w  sp   F   o
-0000020   o  sp   w   h   e   r   e  nl
-0000030
+0000020   o  sp   w   h   e   r   e
+0000027
```
You can create one by simply saving the file in VSCode witout leaving an empty line at the end of the file.

Then tell HLS to add `show` and its signature

### Expected behaviour

The result should be
```haskell
-- like above
instance Show Foo where
  show :: Foo -> String
  show = _
```

### Actual behaviour

[… but it is this (see GIF)](https://ibb.co/MDkFZgps):
```haskell
-- like above
instance Show Foo where  show :: Foo -> String
  show = _
```

### Debug information

I think this bug affects/might affect and maybe _is_ affecting other existing plugins.

Surely [it _is_ affecting, as proved by the tests](https://github.com/haskell/haskell-language-server/actions/runs/33015704178/job/98333302415?pr=5014) the [case split plugin](https://github.com/haskell/haskell-language-server/pull/5014) I'm working on.

I don't understand why a similar test I wrote for the class split plugin [is not failing](https://github.com/haskell/haskell-language-server/pull/5058/changes), though!

Aster89:
Both the given and expected files don't have a line terminator at EOF.

The new test does fail, demonstrating there _is_ a bug, just like I observe in VSCode, see GIF attached to #5059.

However, Vim+YCM and Neovim seem to be immune to it. (As far as Vim+YCM goes, I know why it's immune because I fixed another bug, https://github.com/ycm-core/YouCompleteMe/pull/4311.)

I think that the test failing shows that the bug is in HLS, not in VSCode. Vim+YCM and Neovim are probably just being smart and sidestepping the bug entirely.

---

Fixes #5059.

### DIFF
--- a/hls-plugin-api/hls-plugin-api.cabal
+++ b/hls-plugin-api/hls-plugin-api.cabal
@@ -53,6 +53,7 @@ library
 
   hs-source-dirs:     src
   build-depends:
+    , split ^>=0.2.5
     , aeson
     , base                  >=4.12    && <5
     , co-log-core
--- a/hls-plugin-api/src/Ide/PluginUtils.hs
+++ b/hls-plugin-api/src/Ide/PluginUtils.hs
@@ -13,6 +13,7 @@ module Ide.PluginUtils
     makeDiffTextEditAdditive,
     diffText,
     diffText',
+    diffTextEdit,
     pluginDescToIdePlugins,
     idePluginsToPluginDesc,
     getClientConfig,
@@ -45,8 +46,10 @@ import           Data.Algorithm.Diff
 import           Data.Algorithm.DiffOutput
 import           Data.Char                   (isPrint, showLitChar)
 import           Data.Functor                (void)
+import           Data.List.Split
 import qualified Data.Map                    as M
 import qualified Data.Text                   as T
+import           Data.Tuple.Extra            (both)
 import           Data.Void                   (Void)
 import           Ide.Plugin.Config
 import           Ide.Plugin.Properties
@@ -119,24 +122,28 @@ makeDiffTextEditAdditive :: T.Text -> T.Text -> [TextEdit]
 makeDiffTextEditAdditive f1 f2 = diffTextEdit f1 f2 SkipDeletions
 
 diffTextEdit :: T.Text -> T.Text -> WithDeletions -> [TextEdit]
-diffTextEdit fText f2Text withDeletions = r
+diffTextEdit fText f2Text withDeletions
+  = map diffOperationToTextEdit diffOps
   where
-    r = map diffOperationToTextEdit diffOps
-    d = getGroupedDiff (lines $ T.unpack fText) (lines $ T.unpack f2Text)
+    -- Lossless versions of lines/unlines (they keep the line breaks)
+    lines = split (dropFinalBlank $ keepDelimsR $ whenElt (== '\n'))
+    unlines = concat
+
+    (linesL, linesR) = both (lines . T.unpack) (fText, f2Text)
 
     diffOps =
       filter
         (\x -> (withDeletions == IncludeDeletions) || not (isDeletion x))
-        (diffToLineRanges d)
-
-    isDeletion (Deletion _ _) = True
-    isDeletion _              = False
+        (diffToLineRanges $ getGroupedDiff linesL linesR)
+      where
+        isDeletion (Deletion _ _) = True
+        isDeletion _              = False
 
     diffOperationToTextEdit :: DiffOperation LineRange -> TextEdit
     diffOperationToTextEdit (Change fm to) = TextEdit range nt
       where
         range = calcRange fm
-        nt = T.pack $ init $ unlines $ lrContents to
+        nt = T.pack $ unlines $ lrContents to
 
     {-
       In order to replace everything including newline characters,
--- a/hls-plugin-api/test/Ide/PluginUtilsTest.hs
+++ b/hls-plugin-api/test/Ide/PluginUtilsTest.hs
@@ -20,9 +20,9 @@ import           Ide.Plugin.Properties       (KeyNamePath (..),
                                               usePropertyByPath,
                                               usePropertyByPathEither)
 import qualified Ide.Plugin.RangeMap         as RangeMap
-import           Ide.PluginUtils             (extractTextInRange, unescape)
+import           Ide.PluginUtils             (extractTextInRange, unescape, diffTextEdit, WithDeletions (IncludeDeletions))
 import           Language.LSP.Protocol.Types (Position (..), Range (Range),
-                                              UInt, isSubrangeOf)
+                                              UInt, isSubrangeOf, TextEdit(TextEdit), mkRange)
 import           Test.Tasty
 import           Test.Tasty.Golden           (goldenVsStringDiff)
 import           Test.Tasty.HUnit
@@ -31,6 +31,7 @@ import           Test.Tasty.QuickCheck
 tests :: TestTree
 tests = testGroup "PluginUtils"
     [ unescapeTest
+    , diffTextEditTest
     , extractTextInRangeTest
     , localOption (QuickCheckMaxSize 10000) $
         testProperty "RangeMap-List filtering identical" $
@@ -55,6 +56,56 @@ unescapeTest = testGroup "unescape"
     , testCase "control characters should not be escaped" $
         unescape "\"\\n\\t\"" @?= "\"\\n\\t\""
     ]
+
+diffTextEditTest :: TestTree
+diffTextEditTest = testGroup "diffTextEdit"
+    [ testGroup "inserting line at EOF"
+      [ testCase "both newline-terminated (linux-style vs linux-style)"
+          $ diffTextEditComplete "foo\n"
+                                 "foo\nbar\n"
+                        @?= [textEdit "bar\n"
+                                      (mkRange 1 0 1 0)]
+      , testCase "neither newline-terminated (win-style vs win-style)"
+          $ diffTextEditComplete "foo"
+                                 "foo\nbar"
+                   @?= [textEdit "foo\nbar"
+                                 (mkRange 0 0 0 3)]
+      , testCase "only left newline-terminated"
+          $ diffTextEditComplete "foo\n"
+                                 "foo\nbar"
+                        @?= [textEdit "bar"
+                                      (mkRange 1 0 1 0)]
+      , testCase "only right newline-terminated"
+          $ diffTextEditComplete "foo"
+                                 "foo\nbar\n"
+                   @?= [textEdit "foo\nbar\n"
+                                 (mkRange 0 0 0 3)]
+      ]
+    , testGroup "deleting line at EOF"
+      [ testCase "both newline-terminated (linux-style vs linux-style)"
+          $ diffTextEditComplete "foo\nbar\n"
+                                 "foo\n"
+                        @?= [textEdit ""
+                                      (mkRange 1 0 2 0)]
+      , testCase "neither newline-terminated (win-style vs win-style)"
+          $ diffTextEditComplete "foo\nbar"
+                                 "foo"
+                   @?= [textEdit "foo"
+                                 (mkRange 0 0 1 3)]
+      , testCase "only left newline-terminated"
+          $ diffTextEditComplete "foo\nbar"
+                                 "foo\n"
+                        @?= [textEdit ""
+                                      (mkRange 1 0 2 0)]
+      , testCase "only right newline-terminated"
+          $ diffTextEditComplete "foo\nbar\n"
+                                 "foo"
+                      @?= [textEdit "foo"
+                                    (mkRange 0 0 1 4)]
+      ]
+    ]
+    where diffTextEditComplete from to = diffTextEdit from to IncludeDeletions
+          textEdit = flip TextEdit
 
 extractTextInRangeTest :: TestTree
 extractTextInRangeTest = testGroup "extractTextInRange"
--- a/hls-plugin-api/test/Ide/PluginUtilsTest.hs
+++ b/hls-plugin-api/test/Ide/PluginUtilsTest.hs
@@ -20,9 +20,12 @@ import           Ide.Plugin.Properties       (KeyNamePath (..),
                                               usePropertyByPath,
                                               usePropertyByPathEither)
 import qualified Ide.Plugin.RangeMap         as RangeMap
-import           Ide.PluginUtils             (extractTextInRange, unescape, diffTextEdit, WithDeletions (IncludeDeletions))
+import           Ide.PluginUtils             (WithDeletions (IncludeDeletions),
+                                              diffTextEdit, extractTextInRange,
+                                              unescape)
 import           Language.LSP.Protocol.Types (Position (..), Range (Range),
-                                              UInt, isSubrangeOf, TextEdit(TextEdit), mkRange)
+                                              TextEdit (TextEdit), UInt,
+                                              isSubrangeOf, mkRange)
 import           Test.Tasty
 import           Test.Tasty.Golden           (goldenVsStringDiff)
 import           Test.Tasty.HUnit
--- a/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
+++ b/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
@@ -17,7 +17,7 @@ import           Control.Arrow                         (Arrow (first))
 import           Control.Exception                     (SomeException)
 import qualified Control.Foldl                         as L
 import           Control.Lens                          (Identity (..), ix, view,
-                                                        (%~), (<&>), (^.), (.~))
+                                                        (%~), (.~), (<&>), (^.))
 import           Control.Monad                         (forM, guard, unless)
 import           Control.Monad.Error.Class             (MonadError (throwError))
 import           Control.Monad.Extra                   (eitherM)
@@ -61,7 +61,8 @@ import           Data.Foldable                         (Foldable (foldl'))
 import           GHC.Data.Bag                          (Bag)
 
 #if MIN_VERSION_ghc(9,13,0)
-import           GHC.Parser.Annotation                 (EpAnn (..), EpToken (..))
+import           GHC.Parser.Annotation                 (EpAnn (..),
+                                                        EpToken (..))
 #elif MIN_VERSION_ghc(9,9,0)
 import           GHC.Parser.Annotation                 (EpAnn (..))
 #else
--- a/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
+++ b/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
@@ -252,10 +252,11 @@ adjustToRange uri ran wsEdit@WorkspaceEdit{..} =
         adjustATextEdits :: Traversable f => f (TextEdit |? AnnotatedTextEdit) -> f (TextEdit |? AnnotatedTextEdit)
         adjustATextEdits = fmap $ \case
           InL t -> InL $ runIdentity $ adjustTextEdits (Identity t)
-          InR AnnotatedTextEdit{_range, _newText, _annotationId} ->
-            let oldTE = TextEdit{_range,_newText}
-              in let TextEdit{_range,_newText} = runIdentity $ adjustTextEdits (Identity oldTE)
-                in InR $ AnnotatedTextEdit{_range,_newText,_annotationId}
+          InR ate@AnnotatedTextEdit{ _annotationId } ->
+               InR $ annotate (runIdentity $ adjustTextEdits $ Identity $ unannotate ate) _annotationId
+          where
+            unannotate AnnotatedTextEdit{..} = TextEdit _range _newText
+            annotate TextEdit{..} anno = AnnotatedTextEdit _range _newText anno
 
         adjustWS = ix uri %~ adjustTextEdits
 
--- a/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
+++ b/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
@@ -237,8 +237,9 @@ setupDynFlagsForGHCiLike env dflags = do
     Loader.initializePlugins (hscSetFlags dflags4 env)
 
 adjustToRange :: Uri -> Range -> WorkspaceEdit -> WorkspaceEdit
-adjustToRange uri ran (WorkspaceEdit mhult mlt x) =
-    WorkspaceEdit (adjustWS <$> mhult) (fmap adjustDoc <$> mlt) x
+adjustToRange uri ran wsEdit@WorkspaceEdit{..} =
+    wsEdit { _changes = adjustWS <$> _changes
+           , _documentChanges = fmap adjustDoc <$> _documentChanges }
     where
         adjustTextEdits :: Traversable f => f TextEdit -> f TextEdit
         adjustTextEdits eds =
@@ -257,6 +258,7 @@ adjustToRange uri ran (WorkspaceEdit mhult mlt x) =
                 in InR $ AnnotatedTextEdit{_range,_newText,_annotationId}
 
         adjustWS = ix uri %~ adjustTextEdits
+
         adjustDoc :: DocumentChange -> DocumentChange
         adjustDoc (InR es) = InR es
         adjustDoc (InL es)
--- a/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
+++ b/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
@@ -243,11 +243,11 @@ adjustToRange uri ran wsEdit@WorkspaceEdit{..} =
     where
         adjustTextEdits :: Traversable f => f TextEdit -> f TextEdit
         adjustTextEdits eds =
-            let minStart =
-                    case L.fold (L.premap (view J.range) L.minimum) eds of
+            let minStart :: Range -- leftmost-starting range or, to break ties, leftmost-ending range
+                 = case L.fold (L.premap (view J.range) L.minimum) eds of
                         Nothing -> error "impossible"
                         Just v  -> v
-            in adjustLine minStart <$> eds
+            in adjustLine ran minStart <$> eds
 
         adjustATextEdits :: Traversable f => f (TextEdit |? AnnotatedTextEdit) -> f (TextEdit |? AnnotatedTextEdit)
         adjustATextEdits = fmap $ \case
@@ -267,10 +267,10 @@ adjustToRange uri ran wsEdit@WorkspaceEdit{..} =
                 InL $ es & J.edits %~ adjustATextEdits
             | otherwise = InL es
 
-        adjustLine :: Range -> TextEdit -> TextEdit
-        adjustLine bad =
-            J.range %~ \r ->
-                if r == bad then ran else bad
+adjustLine :: Range -> Range -> TextEdit -> TextEdit
+adjustLine Range{ _start = s } bad@Range{ _end = e } =
+   J.range %~ \r ->
+       if r == bad then Range s e else bad
 
 -- Define a pattern to get hold of a `SrcSpan` from the location part of a
 -- `GenLocated`. In GHC >= 9.2 this will be a SrcSpanAnn', with annotations;
--- a/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
+++ b/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
@@ -17,7 +17,7 @@ import           Control.Arrow                         (Arrow (first))
 import           Control.Exception                     (SomeException)
 import qualified Control.Foldl                         as L
 import           Control.Lens                          (Identity (..), ix, view,
-                                                        (%~), (<&>), (^.))
+                                                        (%~), (<&>), (^.), (.~))
 import           Control.Monad                         (forM, guard, unless)
 import           Control.Monad.Error.Class             (MonadError (throwError))
 import           Control.Monad.Extra                   (eitherM)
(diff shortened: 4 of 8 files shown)


Aster89:
A small experiment that I should probably turn into a test:
```haskell
$ cabal  repl /home/enrico/haskell-language-server/hls-plugin-api/src/Ide/PluginUtils.hs
λ> import Language.LSP.Protocol.Types
λ> import Ide.PluginUtils
λ> :set -XOverloadedStrings
λ> uri = Uri {getUri = "file:///home/enrico/haskell-language-server/plugins/hls-class-plugin/test/testdata/T1W.hs"}
λ> verTxtDocId = VersionedTextDocumentIdentifier uri 0
λ> old = "module T1 where\n\ndata X = X\n\ninstance Eq X where\n"
λ> new = "module T1 where\n\ndata X = X\n\ninstance Eq X where\n  (==) = _\n"
λ> e1 = diffText' True (verTxtDocId, old) new IncludeDeletions
λ> e1
WorkspaceEdit {_changes = Nothing, _documentChanges = Just [InL (TextDocumentEdit {_textDocument = OptionalVersionedTextDocumentIdentifier {_uri = Uri {getUri = "file:///home/enrico/haskell-language-server/plugins/hls-class-plugin/test/tes
tdata/T1W.hs"}, _version = InL 0}, _edits = [InL (TextEdit {_range = Range {_start = Position {_line = 5, _character = 0}, _end = Position {_line = 5, _character = 0}}, _newText = "  (==) = _\n"})]})], _changeAnnotations = Nothing}
```
See that the  `WorkspaceEdit` contains `_newText = "  (==) = _\n"`.

This is correct. But look what happens if we remove the trailing `\n` to both `old` and `new`:
```haskell
λ> old = "module T1 where\n\ndata X = X\n\ninstance Eq X where"
λ> new = "module T1 where\n\ndata X = X\n\ninstance Eq X where\n  (==) = _"
λ> e2 = diffText' True (verTxtDocId, old) new IncludeDeletions
λ> e1 == e2
True
```
which is wrong! The `_newText` should be `"\n (==) = _"`, not `"  (==) = _\n"`.

These are probably the shortest reproduction steps (but `diffTextEdit` is not currently exported):
```haskell
λ> d1 = diffTextEdit "foo" "foo\nbar" IncludeDeletions 
λ> d2 = diffTextEdit "foo\n" "foo\nbar\n" IncludeDeletions 
λ> d1 == d2
True
λ> d1
[TextEdit {_range = Range {_start = Position {_line = 1, _character = 0}, _end = Position {_line = 1, _character = 0}}, _newText = "bar\n"}]
```

Aster89:
Well, the bug is clearly on this line:
https://github.com/haskell/haskell-language-server/blob/d3464b6361dd84b8aa6ae9426ff9814b7809d23c/hls-plugin-api/src/Ide/PluginUtils.hs#L125

I mean, once you've done `lines` on the input texts, linebreaks are long gone. Unless you re-inspect the input texts to see whether they ended with a line terminator, but that's not done, as `fText` and `f2Text` are only used on this line.

Aster89:
Building on my earlier experience with text diff tools, the solution I've attempted consisted of

  1. swapping `Prelude`'s `lines`/`unlines` for these
     ```haskell
     lines = split (dropFinalBlank $ keepDelimsR $ whenElt (== '\n'))
     unlines = concat
     ```
  2. swapping `getGroupedDiff` for ``getGroupedDiffBy ((==) `on` takeWhile (/= '\n'))``

The idea is that point 1 allows us not to throw away the line terminators, and point 2 preserves the equality used so far.

The drawback is that the resulting `[Diff [String]]` for `foo` vs `foo\nbar` is the following
```haskell
[Both ["foo"] ["foo\n"],Second ["bar"]]
```
from which the current code deduces that `"bar"` is the only thing to be added to the `Second` side, forgetting entirely that there was a `\n` difference between the two first lines.

In the context of a full-fledged text diff tool, the direction I'd take is to perform a sub-comparison between the 2 sides of the `Both`s; eventually we'd get something like this,
```haskell
[(Both ["foo"] ["foo\n"], Just (NonEmpty [Both "foo" "foo", Second "\n"])]),(Second ["bar"], Nothing)]
```
where the `Maybe` wraps the possibly absent/empty subcomparison.


If we were to diff `foo\n` vs `foo\nbar\n` then the above "diff+subdiff" would look like this:
```haskell
[(Both ["foo\n"] ["foo\n"], Nothing]),(Second ["bar\n"], Nothing)]
```

---

Anyway, I've taken note of the above to avoid forgetting, but it sounds too much of a complication considering that the only time that `Just` would ever materialize is when we're adding a line at the end of a Windows/VSCode/windows-like-thingy--generated file.

And maybe it would break several tests.

Probably a simple hack is a better approach. Looking into it. But I also have to check what happens in case an action _removes_ the last line of a file.

Aster89:
This PR, in its current state, is

  - a partial solution to the bug that affects `diffTextEdit` intended as a standalone function,
  - but it's an actual solution to #5059.

What do I mean?

  1. From the perspective of text comparison, the solution is wrong, as demonstrated by the two tests that fail;

      - incidentally, it doesn't even update the the [`TextEdit`](https://hackage-content.haskell.org/package/lsp-types-2.4.0.0/docs/Language-LSP-Protocol-Types.html#t:TextEdit)'s `_range` field, only the `_newText` field;

  3. from the perspective of _our usage of it_, which is from call sites that guarantee (or don't they?) that we'll never get those inputs like in the files that cause those failures, it's good enough.

The point is that we never use this `diffTextEdit` function on two independent `Text` inputs. Those two inputs are

  - the content of the source file on which HLS wants to do the change,
  - the content after the change as computed via GHC's API.

As long as GHC (well, and `ghc-exactprint` after it) guarantees to honor the line-ending policy of a file, the scenario of the tests at point 1 above should never materialize.

Aster89:
All tests that fail are failing the same way.

Here's an example failure:
```
  TSimpleDecl (golden):          FAIL (6.45s)
    Test output was different from 'plugins/hls-splice-plugin/test/testdata/TSimpleDecl.expected.hs'. Output of ["git","-c","core.fileMode=false","diff","--no-index","--text","--exit-code","plugins/hls-splice-plugin/test/testdata/TSimpleDecl.expected.hs","/tmp/TSimpleDecl.expected1941492-18.actual"]:
    diff --git a/plugins/hls-splice-plugin/test/testdata/TSimpleDecl.expected.hs b/tmp/TSimpleDecl.expected1941492-18.actual
    index 90c2bf1b0..d0178c7b0 100644
    --- a/plugins/hls-splice-plugin/test/testdata/TSimpleDecl.expected.hs
    +++ b/tmp/TSimpleDecl.expected1941492-18.actual
    @@ -7,6 +7,7 @@ import Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )
     --  Bar
     foo :: Int
     foo = 42
    +
     -- Bar
     -- ee
     -- dddd

    Use -p '/TSimpleDecl (golden)/' to rerun this test only.
```
The given file is this
```haskell
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE QuasiQuotes #-}
module TSimpleDecl where
import Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )

-- Foo
--  Bar
$(sequence
    [sigD (mkName "foo") [t|Int|]
    ,funD (mkName "foo") [clause [] (normalB [|42|]) []]
    ]
    )
-- Bar
-- ee
-- dddd
```
and the expected is this
```haskell
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE QuasiQuotes #-}
module TSimpleDecl where
import Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )

-- Foo
--  Bar
foo :: Int
foo = 42
-- Bar
-- ee
-- dddd
```
I've logged the `TextEdit` inside the `WorkspaceEdit`:
```haskell
TextEdit {_range = Range {_start = Position {_line = 7, _character = 0}, _end = Position {_line = 11, _character = 5}}, _newText = "foo :: Int\nfoo = 42\n"}
```
Its `_range` goes from the beginning of the line containing the `$(`, to right after the last character of the line containing the matching `)`.

Since `Range` represets a half-open range, that `_range` is representing the text from `$(` to the matching `)`, _without_ the line break character after it. In other words, that range is representing the following bytes

  - if the file was saved on Linux:
    ```
    $(sequence\n    [sigD (mkName "foo") [t|Int|]\n    ,funD (mkName "foo") [clause [] (normalB [|42|]) []]\n    ]\n    )
    ```
  - if the file was saved on Windows:
    ```
    $(sequence\r\n    [sigD (mkName "foo") [t|Int|]\r\n    ,funD (mkName "foo") [clause [] (normalB [|42|]) []]\r\n    ]\r\n    )
    ```

Notice that there isn't a line break at the end, because a `Range` is half open (and, to be precise, the LSP doesn't even talk about line breaks as _proper_ characters/bytes).

If we change that text for one that does have a line break at the end (see the `_newText` from the `TextEdit` above), we'll get an empty line in excess.

I'm now trying to understand where that `\n` comes from, and what breaks if I change the code that inserts it not to insert it.

Aster89:
This snippet,

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L166-L174

is where the `WorkspaceEdit` is generated by feeding the document and a function of type `Graft (Either String) ParsedSource` to the `transform` function below:

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-refactor-plugin/src/Development/IDE/GHC/ExactPrint.hs#L216-L227

But `transform` seems to do a pretty simple job, and it uses the very `diffText` of which I'm alterning the implementation, so I would first investigate `graftDecls`, as maybe that's the one that is doing something weird with line breaks.

---

For the very example in [a previous message](https://github.com/haskell/haskell-language-server/pull/5058#issuecomment-5524147239), of which I copy-and-paste the input file,
```haskell
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE QuasiQuotes #-}
module TSimpleDecl where
import Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )

-- Foo
--  Bar
$(sequence
    [sigD (mkName "foo") [t|Int|]
    ,funD (mkName "foo") [clause [] (normalB [|42|]) []]
    ]
    )
-- Bar
-- ee
-- dddd
```
I've printed the `src` and `res` passed to `diffText`, together with a comment marking the start of each 0-based line:
```haskell
   "\n\nmodule TSimpleDecl where\nimport Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )\n\n\n\n$(sequence\n    [sigD (mkName \"foo\") [t|Int|]\n    ,funD (mkName \"foo\") [clause [] (normalB [|42|]) []]\n    ]\n    )\n\n\n\n"
--  0 1 2                         3                                                                   4 5 6 7           8                                    9                                                           10     11     12131415
   "\n\nmodule TSimpleDecl where\nimport Language.Haskell.TH ( mkName, clause, normalB, funD, sigD )\nfoo :: Int\nfoo = 42\n\n\n\n"
--  0 1 2                         3                                                                   4           5         6 7 8 9
```
A couple of observations:

  - I see that the the leading `\n\n`, i.e. the two leading empty lines, are what remains of the two pragmas `{-# LANGUAGE TemplateHaskell #-}` and `{-# LANGUAGE QuasiQuotes #-}` taken away by some pre-processing phase;
  - I see that the `\n\n\n\n` in the first string correspond to the fact that the line where `$(` is 4 lines after the `import` line, so 3 intercurring lines, and again, they are what remains of (one line that was already empty, and) two lines with just comments (`-- Foo`, and `--  Bar`);
  - I don't see why in the result there's only one `\n` between the `import` line and the signature of `foo`, i.e. no empty line in between, considering that the generated file still has the correct 3 lines in between.
  - After all, the `TextEdit`
    ```haskell
    TextEdit {_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 11, _character = 6}}, _newText = "foo :: Int\nfoo = 42\n"}
    ```
    is correct (notice that `_end` is _**after**_ the `\n` that's after the `)` corresponding to `$(`, which is consistent with `_newText` providing its own trailing `\n`)…
  - … but despite that `TextEdit` has the same `_newText` as that in [a previous message](https://github.com/haskell/haskell-language-server/pull/5058#issuecomment-5524147239), it has, crucially, a different `_start` and `_end`:
    ```haskell
    TextEdit {_range = Range {_start = Position {_line = 7, _character = 0}, _end = Position {_line = 11, _character = 5}}, _newText = "foo :: Int\nfoo = 42\n"}
    ```
    This above is the value of `edits` (wrongly called with plural) of the following line:

    https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L192

  - Maybe, the `_start = Position {_line = 7, _character = 0}` that replaces the `_start = Position {_line = 4, _character = 0}` is precisely to take into account those 3 lines (1 empty + 2 just comments) that have to be preserved in the output;

  - however replacing `_end = Position {_line = 11, _character = 6}` with `_end = Position {_line = 11, _character = 5}` is what causes the printing of the empty line, because it doesn't match the `\n` that the other `TextEdit` matches, but it substitutes this range with the same `_newText`, which comes with a trailing `\n`.

Maybe I'm getting somewhere.

Aster89:
Oh, that's the culprit!

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L172-L174

Eh... Not sure what to do yet. Surely removing it does not good.

Next to investigate: `adjustToRange`.

plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
```diff
@@ -62,6 +62,7 @@ import           GHC.Data.Bag                          (Bag)
 
 #if MIN_VERSION_ghc(9,13,0)
 import           GHC.Parser.Annotation                 (EpAnn (..), EpToken (..))
+import Debug.Trace
```
crtschin: Leftover debug statements.

plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs
```diff
@@ -189,7 +190,8 @@ expandTHSplice _eStyle ideState _ params@ExpandSpliceParams {..} = ExceptT $ do
                         MessageType_Error
                         [T.pack $ "Error during expanding splice: " <> show (pretty err)]
                     pure (Left err)
-                Right edits ->
+                Right edits -> do
+                    liftIO $ print edits
```
crtschin: I assume this is leftover as well.

haskell-language-server.cabal
```diff
@@ -30,7 +30,6 @@ extra-source-files:
   plugins/**/*.cabal
   plugins/**/*.yaml
   plugins/**/*.txt
-  plugins/**/*.hs
```
crtschin: Can you expand the above comment with a short `, because`? I assume you deleted this because something was misbehaving (probably recompilation avoidance, but double checking)?

Aster89:
In the usecase I'm referring to (see https://github.com/haskell/haskell-language-server/pull/5058#issuecomment-5524147239), the relevant line of `adjustToRange` is this

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L249

but what `adjustLine` does

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L267-L270

is simply picking one between `bad` and `ran`, where `ran` is actually an input to `adjustToRange`, which is passed at this line:

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L174

`range` is defined here:

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L202

and `spliceSpan` is brought into scope by `..` here

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L97

and it's one of `ExpandSpliceParams`'s fields:

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice/Types.hs#L16-L23

Aster89:
A short-ish recap:

1. The `spliceSpan` passed to the plugin through `ExpandSpliceParams`'s:
   ```haskell
   SrcSpanMultiLine "/home/enrico/haskell-language-server/plugins/hls-splice-plugin/test/testdata/TSimpleDecl.hs" 8 1 12 6
   ```
   represents the sequence of bytes
   ```none
   $(sequence\n    [sigD (mkName "foo") [t|Int|]\n    ,funD (mkName "foo") [clause [] (normalB [|42|]) []]\n    ]\n    )
   ```
   which does _**not**_ include a trailing `\n`,

2. a `Graft (Either String) ParsedSource` is created via `graftDecls`

   https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L170

3. the whole call to `transform`

   https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L166-L171

   results in this `TextEdit`:
   ```haskell
   TextEdit {_range = Range {_start = Position {_line = 4, _character = 0}, _end = Position {_line = 11, _character = 6}}, _newText = "foo :: Int\nfoo = 42\n"}
   ```

   which is:

     - good, because it is _**self-consistent**_ in that

       - it covers the `\n` after the `)` corresponding to the opening `$(`,

       - has a `_newText` coming with its own trailing `\n`;

     - bad, because it starts too early, covering the lines with the comments.

4. That `TextEdit` is then "post-processed" here:

   https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L172-L174

   which is:

     - good, because it works around the bad point above;

     - bad, because it exchanges the computed `TextEdit`'s `_range` with the `Range` corresponding to `spliceSpan`, which does _**not**_ cover the trailing `\n`, and is therefore inconsistent with the computed `TextEdit`'s `_newText`.

---

Now, I think that the call to `adjustToRange` is a hack, in line with the comment accompanying it, to workaround a buggy (or not well understood) behavior of ghc-exactprint. As a hack, it is _ad-hoc_, and it's falling apart upon my attempt to fix `diffTextEdit`.

So what's `adjustToRange` really doing?

I've rewritten it below, but with some slight change that is really just cosmetic
```haskell
adjustToRange :: Uri -> Range -> WorkspaceEdit -> WorkspaceEdit
adjustToRange uri ran wsEdit@WorkspaceEdit{..} =
    wsEdit { _changes = adjustWS <$> _changes
           , _documentChanges = fmap adjustDoc <$> _documentChanges }
    where
        adjustTextEdits :: Traversable f => f TextEdit -> f TextEdit
        adjustTextEdits eds =
            let minStart =
                    case L.fold (L.premap (view J.range) L.minimum) eds of
                        Nothing -> error "impossible"
                        Just v  -> v
            in adjustLine ran minStart <$> eds

        adjustATextEdits :: Traversable f => f (TextEdit |? AnnotatedTextEdit) -> f (TextEdit |? AnnotatedTextEdit)
        adjustATextEdits = fmap $ \case
          InL t -> InL $ runIdentity $ adjustTextEdits (Identity t)
          InR ate@AnnotatedTextEdit{ _annotationId } ->
               InR $ annotate (runIdentity $ adjustTextEdits $ Identity $ unannotate ate) _annotationId
          where
            unannotate AnnotatedTextEdit{..} = TextEdit _range _newText
            annotate TextEdit{..} anno = AnnotatedTextEdit _range _newText anno

        adjustWS = ix uri %~ adjustTextEdits

        adjustDoc :: DocumentChange -> DocumentChange
        adjustDoc (InR es) = InR es
        adjustDoc (InL es)
            | es ^. J.textDocument . J.uri == uri =
                InL $ es & J.edits %~ adjustATextEdits
            | otherwise = InL es

adjustLine :: Range -> Range -> TextEdit -> TextEdit
adjustLine good bad =
    J.range %~ \r ->
        if r == bad then good else bad
```
`adjustWS` and `adjustDoc` do very similar things, one using `adjustTextEdits` and the other using `adjustATextEdits`, but in very similar ways; furthermore, `adjustATextEdits` simply offloads the job to `adjustTextEdits`, and the crucial bit is what this latter function does:

1. it extracts the `Range` of the left-most (or shortest when ties) `TextEdit` from the `Traversable` argument;

2. then for each `TextEdit` in the `Traversable`, it calls `adjustLine` with

   - the `ran :: Range` that turns out to correspond to `spliceSpan`

   - the `minStart :: Range` computed at step 1

   -  the `TextEdit`.


<s>I need to understand why `adjustTextEdits` and `adjustLine` do what they do.</s>

I haven't investigated why `adjustLine`, and the whole `adjustToRange`, really, do what they do, but they were part of a hack in the first place, according to the comment accompanying the _invocation_ of `adjustToRange`:

https://github.com/haskell/haskell-language-server/blob/410747c50f3eb6e53d7891ffbae59050670eb382/plugins/hls-splice-plugin/src/Ide/Plugin/Splice.hs#L172-L174

With my change, I'm really just changing `adjustLine`:

  - before my change, it swaps `bad :: Range` for `ran :: Range`;

  - after my change, it limits such a replacement to the `_start` of
    `bad`, leaving `_end` unchanged.

So mine is just another hack on top of that existing hack.

hls-plugin-api/hls-plugin-api.cabal
```diff
@@ -53,6 +53,7 @@ library
 
   hs-source-dirs:     src
   build-depends:
+    , split ^>=0.2.5
```
crtschin: It doesn't appear like you're using anything `>=0.2.5` specific in `split`, and looking at the API changelog, it hasn't changed much in the past decade. To maximize compatibility with different snapshot in the ecosystem, it might be better to drop the bounds here.

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -119,24 +122,28 @@ makeDiffTextEditAdditive :: T.Text -> T.Text -> [TextEdit]
...
 
 diffTextEdit :: T.Text -> T.Text -> WithDeletions -> [TextEdit]
-diffTextEdit fText f2Text withDeletions = r
+diffTextEdit fText f2Text withDeletions
+  = map diffOperationToTextEdit diffOps
   where
-    r = map diffOperationToTextEdit diffOps
-    d = getGroupedDiff (lines $ T.unpack fText) (lines $ T.unpack f2Text)
+    -- Lossless versions of lines/unlines (they keep the line breaks)
+    lines = split (dropFinalBlank $ keepDelimsR $ whenElt (== '\n'))
```
crtschin, line 129: Is this alright on windows where newlines are denoted with `\r\n`?

hls-plugin-api/src/Ide/PluginUtils.hs
```diff
@@ -119,24 +122,28 @@ makeDiffTextEditAdditive :: T.Text -> T.Text -> [TextEdit]
...
+    unlines = concat
```
crtschin, line 130: Bit of a nit: you're overloading existing Prelude functions in `lines/unlines`. I would prefer different names, as it kinda confuses what the intention is. I'll leave it to you whether you want to change this or not though.

hls-plugin-api/test/Ide/PluginUtilsTest.hs
```diff
@@ -56,6 +60,56 @@ unescapeTest = testGroup "unescape"
...
+      , testCase "only left newline-terminated"
+          $ diffTextEditComplete "foo\nbar"
+                                 "foo\n"
+                        @?= [textEdit ""
+                                      (mkRange 1 0 2 0)]
+      , testCase "only right newline-terminated"
+          $ diffTextEditComplete "foo\nbar\n"
+                                 "foo"
+                      @?= [textEdit "foo"
+                                    (mkRange 0 0 1 4)]
```
crtschin: Isn't this wrong? I understand it like the following
```
<start>     <start>
foo         foo<end>
bar     -> 

<end>
```
The expected edit rewrite from (0, 0) to (1, 4). The documentation for [`Range`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range) states that newlines on lines should be denoted using the column 0 of the next line. So the current expected diff should be interpreted as
```
<start>    <start>
foo        foo
bar     -> <end>

<end>
```
where the newline after `bar` is kept. This does not match the after value. From reading the LSP spec, I think the range should be `mkRange 0 0 2 0`.

crtschin (commented):
Nice investigative work! I think there might still be a positional bug here.

hls-plugin-api/src/Ide/PluginUtils.hs
Aster89: I think so, here's how it works:
```haskell
λ> lines = split (dropFinalBlank $ keepDelimsR $ whenElt (== '\n'))
λ> lines "one\ntwo\nthree"
["one\n","two\n","three"]
λ> lines "one\r\ntwo\r\nthree"
["one\r\n","two\r\n","three"]
λ> lines "one\ntwo\nthree\n"
["one\n","two\n","three\n"]
λ> lines "one\r\ntwo\r\nthree\r\n"
["one\r\n","two\r\n","three\r\n"]
```

hls-plugin-api/src/Ide/PluginUtils.hs
Aster89: The way I see it is that the fact I'm overriding two `Prelude` functions communicates that the overriding functions are not entirely different beasts than the overridden ones: `Prelude`'s `lines`/`unlines` is what I'd use, but they don't quite fit the bill and I'm changing them just a little bit.

Indeed, if I had to choose a different name, I'd have a hard time to make it a single-word (other than via appending a `'`). `linesWithNewline`? `losslessLines`? `llLines`? `posixLines`¹?

Since I'm limiting the overriding to a local binding, I think it's ok. Unless anybody feels strong about it.

---

¹ Based on [the POSIX definition of _line_](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).

hls-plugin-api/hls-plugin-api.cabal
Aster89: I'm not expert of this, but I'm pretty sure somewhere I've read that having `>=` and `<` bounds for versions is sort of a standard. I'll research a bit more.

hls-plugin-api/test/Ide/PluginUtilsTest.hs
Aster89: Thank you very much for this!

First, whether `"foo\nbar\n"` is rendered with the empty line at the end is an editor's choice, which is historically different in Windows vs Linux. I have [my opinion](https://stackoverflow.com/a/67337745/5825294) which is that [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) is right.

But you are correct that the LSP states what it states about `Range` as regards line breaks.

In this light, I'm not even sure why the following test passes, **despite the presence of the bug** you highlighted:
```
TASTY_PATTERN='/TSimpleDecl/' cabal run hls-splice-plugin-tests
```
Understanding that would be inviting, but… I'm not sure when I'll be willing to look at that, as there's probably nothing really interesting in it.

On the other hand, I have first-hand experience (on Vim+YCM, can't try VSCode comfortably from remote, as I've never set it up as I don't use it) that _clients_ can be confused by a `mkRange _ _ m n :: Range`, where `n` is the number of characters **_including the trailing line break character(s)_** on the 0-based `m`th line, as opposed to `mkRange _ _ (m+1) 0`.

So surely I need to address this as you suggest!

hls-plugin-api/hls-plugin-api.cabal
crtschin: Did a quick check on hackage and if you want a more lax'ed bound, you can use `>= 0.2.4` which will cover our current GHC supported set (`>=9.6`).

crtschin (approved):
Nice job fixing this! 🚀 

I'm not going to be insistent about the version bounds, so approving! Feel free to loosen or delete the version bound on split if you still feel like it. 

Keeping as-is is also okay, split is stable enough where it probably doesn't matter.

hls-plugin-api/hls-plugin-api.cabal
Aster89: Ok, never mind. I got lost in https://discourse.haskell.org/t/the-operator-in-cabal-files/6938 and decided that right now I don't care, if `split` as stable as you suggest :rofl: 

Removing the bound and merging.

PR #5056, opened 2026-08-26
Prepare release 2.15.0.0

crtschin:
## Release-related fixes

- Support both zoned and non-zoned timestamps in `GenChangelogs.hs`.
- Extended release instructions with which token permissions to use.
- Fix calling `GenChangelogs.hs` via nix.
- Bumped `GHC 9.14` to `full` support as fourmolu and ormolu got fixed.
- Bump `--max-backjumps` on Windows, matching the existing `setup-build` action.
- Use `--isolate` when installing via ghcup for testing bindists
- Drop historic cabal flag overrides in cabal.project

## Release checklist

This release does not cover more/different GHC versions than `2.14.0.0`, I've marked all related tasks to do with changing GHC set.

- [X] check ghcup supports new GHC releases if any
- [X] check all plugins still work if release includes code changes
- [X] set the supported GHCs in workflow file `.github/generate-ci/gen_ci.hs`
- [X] regenerate the CI via `./.github/generate-ci/generate-jobs`
- [X] bump package versions in all `*.cabal` files (same version as hls)
  - HLS uses lockstep versioning. The core packages and all plugins use the same version number, and only support exactly this version.
    - Exceptions:
      - `shake-bench` is an internal testing tool, not exposed to the outside world. Thus, no version bump required for releases.
  - For updating cabal files, the following script can be used:
    - ```sh
      ./release/update_versions.sh <OLD_VERSION> <NEW_VERSION>
      ```
    - It still requires manual verification and review
- [X] generate and update changelog
  - Generate a ChangeLog via `./GenChangelogs.hs <api-key> <tag>`
    - `<tag>` is the git tag you want to generate the ChangeLog from.
    - `<api-key>` is a github access key: https://github.com/settings/tokens
- [X] update https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html#current-ghc-version-support-status
- [X] create release branch as `wip/<version>`
  - `git switch -c wip/<version>`
- [x] create release tag as `<version>`
  - `git tag <version>`
- [x] trigger release pipeline by pushing the tag
  - this creates a draft release
  - `git push <remote> <version>`
- [x] run `sh scripts/release/download-gh-artifacts.sh <version> <your-gpg-email>`
  - downloads artifacts to `gh-release-artifacts/haskell-language-server-<version>/`
  - also downloads FreeBSD bindist from circle CI
  - adds signatures
- [x] upload artifacts to downloads.haskell.org from `gh-release-artifacts/haskell-language-server-<version>/`
[...]

PR #5062, opened 2026-08-27
Drop historic cabal flag overrides in cabal.project

### LINKED ISSUE #5061: Revisit `constraints` in `cabal.project`
There's two things here that caught my eye:

https://github.com/haskell/haskell-language-server/blob/e79a030d7872664da711f979c48d699cc94d1bd7/cabal.project?plain=1#L58-L60

This was done as part of release `2.4.0.0`. Reading the [original issue](https://github.com/haskell/haskell-language-server/issues/3822) back, this problem seems specific to GHC 9.0 and 9.2. [This comment reads like it is fixed on 9.4](https://github.com/haskell/haskell-language-server/issues/3822#issuecomment-1744615988) where text starts relying on `system-cxx-lib` instead.

The minimum GHC version we support is 9.6, so we should be in the clear.

https://github.com/haskell/haskell-language-server/blob/3009a45c74dfbc5c3d6cc584b41b01c3616995ec/cabal.project?plain=1#L39-L43

Centos support was dropped in HLS 2.11.0.0, PR and discussion can be found [here](https://github.com/haskell/haskell-language-server/pull/4481).

Should try dropping these flag overrides and see if there are any other blockers that still need these.

crtschin:
Closes #5061.

* `text -simdutf` was necessary on GHC 9.0 and 9.2, outside of our supported GHC set.
* `bitvec -simd` was necessary for Centos 7, which we've stopped supporting since the 2.11.0.0 release.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -26,20 +26,12 @@ package *
   ghc-options: -haddock
 
 constraints:
-  -- C++ is hard to distribute, especially on older GHCs
-  -- See https://github.com/haskell/haskell-language-server/issues/3822
-  text -simdutf,
   ghc-check -ghc-check-use-package-abis,
   ghc-lib-parser-ex -auto,
   -- This is only present in some versions, and it's on by default since
   -- 0.14.5.0, but there are some versions we allow that need this
   -- setting
   stylish-haskell +ghc-lib,
-  -- Centos 7 comes with an old gcc version that doesn't know about
-  -- the flag '-fopen-simd', which blocked the release 2.2.0.0.
-  -- We want to be able to benefit from the performance optimisations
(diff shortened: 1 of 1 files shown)


crtschin:
CI looks okay, so putting it up for review assuming our windows jobs cover the `text` override

fendor:
The CI is not quite relevant here, it mostly matters for the release CI whether it works on all supported platforms.

We can try merging it and see what the release CI says, or we wait after the release finished and then merge.

crtschin:
I see, I'll try cherry picking it onto https://github.com/haskell/haskell-language-server/pull/5056 then.

PR #5078, opened 2026-09-07
Correctly detect file-local `NoUnicodeSyntax`

### LINKED ISSUE #5067: File local `{-# LANGUAGE NoUnicodeSyntax #-}` is mistakenly ignored by case-split plugin if the owning stanza in the .cabal file has `default-extensions: UnicodeSyntax`
Minimal example:
```cabal
cabal-version:   3.4
name:            app
version:         0.1.0.0

library
    build-depends: base
    other-modules: Foo
    default-language: GHC2024
    default-extensions: UnicodeSyntax
```
```cabal
packages: .
with-compiler: ghc-9.14.1
```
```haskell
{-# OPTIONS_GHC -Wall #-}
{-# OPTIONS_GHC -Wno-typed-holes #-}
{-# LANGUAGE NoUnicodeSyntax #-}
module Foo where

foo :: a
foo = case _ :: Bool of
```
The plugin completes with these lines
```haskell
  False → _
  True → _
```
which doesn't typecheck.

Aster89:
This change contains:

  1. a relatively small edit to take into account possible `{-# LANGUAGE UnicodeSyntax #-}`/`{-# LANGUAGE NoUnicodeSyntax #-}` in the file being edited, and to treat those with higher priority than the corresponding flags in (corresponding stanza in) the `.cabal` file,

  2. a test that looks a bit ugly right now;

Regarding 2, my intent is to test for the 9 combinations of `{-# LANGUAGE UnicodeSyntax #-}`/`{-# LANGUAGE NoUnicodeSyntax #-}`/neither × corresponding 3 alternatives in the `.cabal` file, and I've tried, but the code looks messy…

Suggestions?

Fix #5067.


plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -265,19 +268,21 @@ suggestCaseSplitProvider recorder state _ CodeActionParams{..}
...
-getArrowSyntax state nfp = do
-  (hsc_dflags . hscEnv -> dynFlags) <- runActionE "CaseSplit.GhcSessionDeps" state $ useE GhcSessionDeps nfp
-  pure $ if On Ext.UnicodeSyntax `elem` extensions dynFlags
-    then UnicodeSyntax
-    else NormalSyntax
+getArrowSyntax :: ParsedModule -> IdeState -> NormalizedFilePath -> ExceptT PluginError (HandlerM Config) IsUnicodeSyntax
+getArrowSyntax pm state nfp
+  | Ext.UnicodeSyntax `member` getExtensionsSet pm
+  = do (hsc_dflags . hscEnv -> dynFlags) <- runActionE "CaseSplit.GhcSessionDeps" state $ useE GhcSessionDeps nfp
+       pure $ if On Ext.UnicodeSyntax `elem` extensions dynFlags
```
fendor:
```suggestion
       pure $ if On Ext.UnicodeSyntax `member` getExtensionsSet dynFlags
```

No reason to convert the `Extension`s to a list.

plugins/hls-case-split-plugin/test/Main.hs
```diff
@@ -158,16 +160,106 @@ codeActionTests = testGroup
...
+              "common warnings",
+              "    ghc-options: -Wall",
+              "library",
+              "    import:             warnings",
+              "    exposed-modules:    TUnicodeArrow",
+              "    build-depends:      base",
+              "    hs-source-dirs:     .",
+              "    default-language:   GHC2024"
+              ]
+
```
fendor, line 182: It would have honestly been far easier to write a `direct` cradle test like this:

```
cradle:
  direct:
    arguments:
    - Module
    - -XGHC2024
```

Since you care about the base arguments to GHC, rather than that it is a cabal project.

We can merge as is, create an issue to convert it to a direct cradle test. 
Direct cradle tests are much faster since we don't need to load the cabal context.

fendor (approved):
LGTM, one nitpick and a suggestion that I am fine to do in a follow-up

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: Would you mind reviewing this once more?

The `getExtensionsSet` I'm using a few lines above on `pm :: ParsedModule` is from [`Development.IDE.GHC.Util`](https://hackage-content.haskell.org/package/ghcide-2.15.0.0/docs/Development-IDE-GHC-Util.html), and it tells me the set of extensions enabled _in the module_.

`extensions :: [OnOff Extension]`, instead, is one of the fields of [`DynFlags`](https://hackage-content.haskell.org/package/ghc-9.14.1/docs/GHC-Driver-DynFlags.html#t:DynFlags) defined in [`GHC.Driver.DynFlags`](https://hackage-content.haskell.org/package/ghc-9.14.1/docs/GHC-Driver-DynFlags.html), so it's not like _I_ am turning it into a list. It is a list to start with.

I don't think there is a `getExtensionsSet` for `DynFlags`.

Only now I see that `DynFlags` has another field, `extensionFlags :: EnumSet Extension`. However, considering that the module exports `DynFlags` with all its flags, i.e. both `extensions` and `extensionFlags`, I doubt they represent the same thing, otherwise one could make inconsistent alterations on them two.

Aster89: Oh, the source:
```haskell
  -- Don't change this without updating extensionFlags:
  -- Here we collect the settings of the language extensions
  -- from the command line, the ghci config file and
  -- from interactive :set / :seti commands.
  extensions            :: [OnOff LangExt.Extension],
  -- extensionFlags should always be equal to
  --     flattenExtensionFlags language extensions
  -- LangExt.Extension is defined in libraries/ghc-boot so that it can be used
  -- by template-haskell
  extensionFlags        :: EnumSet LangExt.Extension,
```

fendor: Is this the function we are looking for? https://hackage-content.haskell.org/package/ghc-9.14.1/docs/GHC-Driver-DynFlags.html#v:xopt

I guess I would just try to avoid the list `member` check if not necessary.

fendor: The `DynFlags` of a `ParsedModule` are not the same as the `DynFlags` of `useE GhcSessionDeps`. The former gives you the `DynFlags` including locally added `{-# LANGUAGE ... #-}`. The latter are the `DynFlags` of the session, if I am not mistaken :thinking:

Aster89: I've tried changing
```haskell
Ext.UnicodeSyntax `elem` extensions dynFlags
```
for this
```haskell
pure $ if Ext.UnicodeSyntax `xopt` dynFlags
```
and for this
```haskell
pure $ if Ext.UnicodeSyntax `member` extensionFlags dynFlags
```
and in both cases, the tests time out.

---

However, it seems to be a no-problem, because I hadn't understood that

> The former

also includes the sessions's flags, in case there are no

> locally added `{-# LANGUAGE ... #-}`

And this simplifies the whole thing, like, _a lot_, as per the new update I'm gonna push.

plugins/hls-case-split-plugin/test/Main.hs
Aster89: Ok, I'll create an issue for that.

To understand a bit more, do you mean that I would write
```haskell
      $ let cradleFile = [
              "cradle:",
              "  direct:",
              "    arguments:",
              "    - Module",
              "    - -XGHC2024",
              ]
```
and, similarly to what I've done, I'd add another line to mean _"add `UnicodeSyntax`"_ / _"add `NoUnicodeSyntax`"_?

Issue #5072, opened 2026-09-04
Allow some customization of case-split layout strategy


Aster89:
So far I thought of:

   - [ ] allow configuring the indentation w.r.t. the context (currently hardcoded to 2);
   - [ ] allow configuring the maximum number (possibly ∞) of underscores for the ctors' arguments before resorting to `{}` (currently hardcoded to 3).

Those two are already grouped in `data`:

https://github.com/haskell/haskell-language-server/blob/5ce2a847d255ce6f420da49771711a89eee78541/plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs#L682-L696

Originally considered during [my GSoC 2026 project](https://gist.github.com/Aster89/69a57c364ec2d0cc89f474cc3d194371#future-work--stretch-goals).

Aster89:
It might make sense to enforce that `maxUnderscore >= 2`. Without doing so we'd be forced to use prefix form when dealing with constructor operators, e.g.
```haskell
x = case _ :: String of
    [] -> _
    (:) {} -> _
```
which imho looks ugly.

See also #5066.

Aster89:
@fendor, [at least as far as `_ : _` vs `_:_` goes](https://github.com/haskell/haskell-language-server/issues/5066#issuecomment-5597811900), suggests that this sort of spacing decisions are up to the formatter, which I suppose assumes some user-config file that dictates the various choices.

But wouldn't this be true for all spacing related choices?

Furthermore, I would assume that invoking the formatter would at least require some action from the user, e.g. saving the file, no? And if the user wants `_:_` but we do `_ : _`, there's always at least a moment that user sees the formatting they don't like, which can possibly be distracting and break the flow of coding, no?

Anyway, worth having a discussion/doing a poll about this matter before making choices.

fendor:
> But wouldn't this be true for all spacing related choices?

Yes, hence I am on the team "do something sensible but don't care about anything else"

> Anyway, worth having a discussion/doing a poll about this matter before making choices.

I agree, we could post to discourse about it and ask for input. Note though that we also need to think about maintenance (which users have the luxury to ignore), and each configuration increases the testing surface considerably.

As such, my gut feeling is, resist the temptation to add configuration as much as possible. Configuration has a cost.

PR #5082, opened 2026-09-08
Lay an infix constructor properly

### LINKED ISSUE #5066: Case-split on list types mistakenly treats : as a prefix instead of infix constructor
While the plugin does provide completion for these
```haskell
x = case _ :: [Int] of
x = case _ :: String of
```
the completion is syntactically incorrect, because it doesn't take into account that the `:` constructor is infixed:
```haskell
    [] -> _
    : _ _ -> _
```
The correct completion is
```haskell
    [] -> _
    _ : _ -> _
```
In fact, another possibility is for the completion to be this:
```haskell
    [] -> _
    (:) _ _ -> _
```
which imho is a bit ugly, as nobody writes code like this, as far as I've seen.

See also #5072.

Aster89:
With this change, I'm checking if a data constructor is infix, and if it is, I'm parsing it into an `InfixCon` instead of `PrefixCon`.

Care is taken as regards spacing: prefix patterns should go like `Foo _ _`, i.e. spaces before all the `_` and  not before the constructor, whereas infix patterns should go like `_ : _`, i.e. spaces before the contructor and before all but the first `_`. (As I mention in #5066, this spacing policy might actually be customizable in the future. I plan to do that in #5072.)

A test is added.

Fixes #5066.


plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -631,26 +633,41 @@ parseSimpleConMatch :: IsUnicodeSyntax -> PmAltConApp -> Either String SimpleCon
...
+        underscore = WildPat NoExtField
+
+        rdrConName = nameRdrName dataConName
+
+  , Just (locatedCon, args) <- case (paca_ids, infixed) of
+                  -- Prefixed, laid out like @Foo _ _@
+                  (_, False) -> Just $ (-- leave no space before the constructor
+                                        L noSrcSpanA $ rdrConName,
+                                        -- leave one space before each argument
+                                        PrefixCon $ map (const $ L noAnnSrcSpanDP1 underscore) paca_ids)
```
fendor:
```suggestion
                  (_, False) -> Just  (-- leave no space before the constructor
                                        L noSrcSpanA $ rdrConName,
                                        -- leave one space before each argument
                                        PrefixCon $ map (const $ L noAnnSrcSpanDP1 underscore) paca_ids)
```

fendor (approved):
lovely addition!

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: Eagle-eyed.

PR #5090, opened 2026-09-16
Use stable cache key for unit cache directories 

### LINKED ISSUE #5089: Unnecessarily re-indexes the whole project on every start
### Your environment

Which OS do you use? fedora
Which version of GHC do you use and how did you install it? 9.12
How is your project built (alternative: link to the project)? It's HLS.
Which LSP client (editor/plugin) do you use? helix
Which version of HLS do you use and how did you install it? master
Have you configured HLS in any way (especially: a `hie.yaml` file)? no.

### Steps to reproduce

1. Open a module in the codebase. 
2. Look at logs, you'll find a lot of re-indexing log messages. 
3. Once HLS settles, close HLS. 
4. Reopen HLS with the exact same module, and observe the log again. 
5. It will again, be filled with re-indexing log messages.

The log message to look out for being `Re-indexing hie file for ...`, you'll need to pass `-d` to get debug logging.

### Expected behaviour

I would expect indexes to stay stable across restarts if nothing changed. 

### Actual behaviour

They are not.

crtschin:
Closes https://github.com/haskell/haskell-language-server/issues/5089.

When writing the response files for units in a multi-repl session, cabal uses temp directories. As we use cabal's GHC command line arguments to determine caching of units, we accidentally used these temp filepaths in the cache key, leading to impossible to hit cache directories across restarts. 

This PR fixes it by using (hashes of) the contents of the response files instead, and sorting the units as well.

---

- [X] I read and followed our [AI policy](https://github.com/haskell/haskell-language-server/tree/master/docs/contributing/ai_policy.md). This PR was assisted by a coding agent.


crtschin:
Apologies! Accidentally committed some other work I had, causing excessive pings

ghcide/session-loader/Development/IDE/Session.hs
```diff
@@ -221,10 +221,10 @@ data SessionLoadingOptions = SessionLoadingOptions
   -- If a 'hie.yaml' is given, use it to load the cradle.
   -- Otherwise, use the provided project root directory to determine the cradle type.
   , loadCradle             :: Recorder (WithPriority Log) -> Maybe FilePath -> FilePath -> IO (HieBios.Cradle Void)
-  -- | Given the project name and a set of command line flags,
+  -- | Given the key of a component's cache folder,
   --   return the path for storing generated GHC artifacts,
   --   or 'Nothing' to respect the cradle setting
-  , getCacheDirs           :: String -> Maybe B.ByteString -> [String] -> IO CacheDirs
+  , getCacheDirs           :: CacheKey -> IO CacheDirs
```
fendor: The previous implementation allowed users ghcide users (some time ago that was DAML, not sure that's still relevant) to switch out the getCacheDirs and hashing implementation without having to touch HLS. This implementation makes `getCacheDirs` redundant. Is the interface change necessary?

fendor (approved):
Very good catch, thank you very much! 

The interface change might be completely irrelevant, but if we change the implementation to be basically `XDG_CACHE_HOME </> ghcide </> cacheKey`, I am not sure it is worth keeping as an interface.

ghcide/session-loader/Development/IDE/Session.hs
crtschin: > Is the interface change necessary?

Not strictly no. I included it as cleanup/refactoring, as otherwise I'd pass around file contents, instead of only the hashes, which on second thought is probably harmless enough as it's not being retained. `getCacheDirs` isn't completely redundant as tests still use it to avoid `XDG_CACHE_HOME`.

It's not strictly necessary, so I'll revert the interface changes.

PR #5091, opened 2026-09-16
Even if `UnicodeSyntax` is enabled, use `NormalSyntax` syntax if more than half of the existing cases use it.

### LINKED ISSUE #5071: With UnicodeSyntax enabled, the case-split plugin should prefer ordinary syntax if all already existing patterns use it
Provided the `UnicodeSyntax` is enabled (not sure it'll ever become enabled by default) and not disabled locally to the source file being edited, the case-split plugin should prefer using `->` if already existing patterns, if any, use that instead of the unicode arrow.

In other words, calling the plugin on both `case`s below
```haskell
{-# OPTIONS_GHC -Wall #-}
{-# OPTIONS_GHC -Wno-typed-holes #-}
{-# LANGUAGE UnicodeSyntax #-}
module Foo where

foo :: a
foo = case _ :: Bool of
  False -> _

bar :: a
bar = case _ :: Bool of
```
currently results in this:
```haskell
-- as above
foo :: a
foo = case _ :: Bool of
  False -> _
  True → _

bar :: a
bar = case _ :: Bool of
  False → _
  True → _
```
should result in this (notice the `->` instead of `→` in the `True` alternative in `foo`'s definition):
```haskell
-- as above
foo :: a
foo = case _ :: Bool of
  False -> _
  True -> _

bar :: a
bar = case _ :: Bool of
  False → _
  True → _
```
Clearly, if there's more 2 or more existing patterns and they don't agree, a choice needs to be made (honor the last, honor the first, honor the most frequent, honor at least one unicode, ...).

Originally considered during [my GSoC 2026 project](https://gist.github.com/Aster89/69a57c364ec2d0cc89f474cc3d194371#future-work--stretch-goals).

Aster89:
Fixes https://github.com/haskell/haskell-language-server/issues/5071.

I'm making a choice. Others are possible, see e.g. the penultimate paragraph at #5071.


plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -373,10 +373,17 @@ graftMissingPatterns ps range missingPs arrowSyntax
...
-                     -- extract existing matches
+                     -- extract existing matches and most frequent syntax
                      let existingMatches = _matchGroup _expr
+
+                         dominantSyntax NormalSyntax = NormalSyntax
+                         dominantSyntax UnicodeSyntax = let (u, n) = map getSyntax (unLoc $ mg_alts existingMatches)
+                                                                   & partition (== UnicodeSyntax)
+                                    in if length u < length n
+                                              then NormalSyntax
+                                              else UnicodeSyntax
```
fendor: This could be a fold to avoid multiple traversals, e.g. `foldl' (\(!u, !n) val -> case getSyntax val of { UnicodeSyntax -> (u + 1, n); NormalSyntax -> (u, n + 1)}) (0,0) (unLoc $ mg_alts existingMatches)`

Only a one liner due to my laziness, the worker function could be a separate function for better readability.

fendor (approved):
LGTM :)

PR #5094, opened 2026-09-17
Qualify constructors when needed

### LINKED ISSUE #5084: Case split refers to out of scope patterns without adding imports or qualifying
This isn't really an issue for me currently (hence the contrived hide prelude example) but thought it was worth raising regardless. I think not using a qualified name if available is probably a bigger problem than not auto importing because you might be matching something with many patterns you don't want to bring into the normal scope. (Also beware of qualified operators.)

### Your environment

Windows
GHC 9.14.1 (from HLS)
Project build by cabal (but probably possible in a single file)
VSCode + vscode-haskell
HLS 2.15.0.0 via ghcup compile
Config - HLS managed by PATH, semantic tokens on

### Steps to reproduce

Complete the `\case`

```Haskell
module MyLib where
import Prelude (Maybe)
-- import qualified Data.Maybe -- doesn't fix the issue

test :: Maybe a -> ()
test = \case
```

### Expected behaviour

It will auto-import (somehow) similar to auto-completion, or use a qualified name if available (e.g. the various correction suggestions HLS will give after the change).

### Actual behaviour

It will refer to (out of scope) `Just` and `Nothing`

### Debug information

None

Aster89:
With this change, I'm retrieving the pretty-printing context, and then passing it down the call stack, until it reaches `parseSimpleConMatch`, which passes it to the new helper `qualifyIfNeeded` function, which takes the `Name` and the aforementioned context, and - if needed - it prepends the name of the module to the otherwise unqualified `Name`.

Fixes #5084


plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -264,6 +280,18 @@ suggestCaseSplitProvider recorder state _ CodeActionParams{..}
                        , _command     = Nothing
                        , _data_       = Nothing }
 
+-- | Retrieve the pretty printing context, which is used to determine whether
+-- the constructors of the patterns to be inserted need be qualified, and what
+-- the qualifier should be.
+getPprCtx :: IdeState -> NormalizedFilePath -> ExceptT PluginError (HandlerM Config) PrintUnqualified
+getPprCtx state nfp = do
+  (typechecked, hscEnvEq) <- runIdeActionE "ExplicitFields.InlayHintPosRec" (shakeExtras state) $ do
```
MangoIV: the action name looks wrong, probably a copy-pasta leftover?

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -264,6 +280,18 @@ suggestCaseSplitProvider recorder state _ CodeActionParams{..}
...
+    (typechecked, _) <- useWithStaleFastE TypeCheck nfp
+    (hscEnvEq, _) <- useWithStaleFastE GhcSession nfp
+    return (typechecked, hscEnvEq)
+  let reader = tcg_rdr_env (tmrTypechecked typechecked)
+  pure $ mkPrintUnqualifiedDefault (hscEnv hscEnvEq) reader
```
MangoIV, line 292: not yours but this is one hell of a function name :D

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: Oops, need to change this string.

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -687,7 +713,14 @@ parseSimpleConMatch arrow PACA{ paca_con = PmAltConLike con
...
                    , _conPat = conPat }
 
-parseSimpleConMatch _ paca = Left $ showSDocUnsafe $ ppr paca
+parseSimpleConMatch _ _ paca = Left $ showSDocUnsafe $ ppr paca
+
+-- | Given a 'PrintUnqualified' context and a 'Name', return the 'Qual'ified
+-- name or the 'Exact' name as needed.
+qualifyIfNeeded :: PrintUnqualified -> Name -> RdrName
+qualifyIfNeeded pprCtx name = case init $ T.split (== '.') $ printOutputableQualified pprCtx name of
+  [] -> Exact name
```
MangoIV: Why is this `Exact`? I don't think it's right, but then I also thought this must explode in colorful ways if this was wrong...

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -687,7 +713,14 @@ parseSimpleConMatch arrow PACA{ paca_con = PmAltConLike con
...
+  (T.intercalate "." -> moduleName) -> Qual (ModuleName $ mkFastString $ T.unpack moduleName) (occName name)
```
MangoIV: this should probably use `mkRdrQual`

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
MangoIV: is the contract with `printOutputableQualified` that it will only omit qualification if the given name is an exact name? If so, this should be stated.

MangoIV: can we have a testcase that exercises this code path?

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: Precisely 🤭

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
```diff
@@ -264,6 +280,18 @@ suggestCaseSplitProvider recorder state _ CodeActionParams{..}
...
                        , _data_       = Nothing }
 
+-- | Retrieve the pretty printing context, which is used to determine whether
+-- the constructors of the patterns to be inserted need be qualified, and what
+-- the qualifier should be.
+getPprCtx :: IdeState -> NormalizedFilePath -> ExceptT PluginError (HandlerM Config) PrintUnqualified
+getPprCtx state nfp = do
+  (typechecked, hscEnvEq) <- runIdeActionE "ExplicitFields.InlayHintPosRec" (shakeExtras state) $ do
+    (typechecked, _) <- useWithStaleFastE TypeCheck nfp
+    (hscEnvEq, _) <- useWithStaleFastE GhcSession nfp
```
crtschin: This is a manually prompted right, why not go with the standard `use` variants? 

I think users would prefer the right qualified prefix instead of an outdated one, even if it means waiting a little bit more for live results.

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: Wait a sec, which one? The function these lines belong to _is_ mine, `getPprCtx`, and I can change it to `getPprContext`, if you like. Or which one are you referring to?

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: > Why is this `Exact`?

From the `ConLike`, being it a `RealDataCon` or a `PatSynCon`, I get a `Name`, so I thought that `Exact` was the `RdrName`'s constructor to be used; now that I re-read the doc, I think that, even to create an `Exact`, I should have used `getRdrName` instead. If I'm not supposed to create an `Exact` at all, can you explain why?

> is the contract with `printOutputableQualified` that it will only omit qualification if the given name is an exact name?

I haven't worked that out. Here's `printOutputableQualified` and two functions used in its implementation:

https://github.com/haskell/haskell-language-server/blob/38a34d3593b4651d302e4462e99ec22bd495c4f6/ghcide/src/Development/IDE/GHC/Util.hs#L275-L277

`printOutputable'`

https://github.com/haskell/haskell-language-server/blob/38a34d3593b4651d302e4462e99ec22bd495c4f6/ghcide/src/Development/IDE/GHC/Util.hs#L279-L285

doesn't seem to have anything specific to `Unqual`/`Qual`/`Orig`/`Exact`.

`printSDocQualifiedUnsafe`,

https://github.com/haskell/haskell-language-server/blob/38a34d3593b4651d302e4462e99ec22bd495c4f6/ghcide/src/Development/IDE/GHC/Compat/Outputable.hs#L95-L101

takes already an `SDoc`.

That's to say nothing strikes me like the answer to your question. And there isn't much comments around.

> can we have a testcase that exercises this code path?

I would think all the other tests would exercise this path.

---

Anyway, I've change the code to this:
```haskell
-- | Given a 'PrintUnqualified' context and a 'Name', return the corresponding
-- 'RdrName', but qualified if needed.
qualifyIfNeeded :: PrintUnqualified -> Name -> RdrName
qualifyIfNeeded pprCtx name
  = let moduleName = printOutputableQualified pprCtx name
                   & T.split (== '.')
                   & init
                   & T.intercalate "."
    in case moduleName of
        "" -> nameRdrName name -- XXX Or mkRdrUnqual? Or, since I know the name, I should use getRdrName? Or what?
        _ -> mkRdrQual (ModuleName $ mkFastString $ T.unpack moduleName)
                       (occName name)
```

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
Aster89: I really just copy-and-pasted without even reading :no_mouth:

Changed.

plugins/hls-case-split-plugin/src/Ide/Plugin/CaseSplit.hs
MangoIV: as the doc explains `Exact` is for exact names which is a special concept in the compiler for names that don't work as the usual ones, i.e. are imported from somewhere or just come from the same module. They're used for instance to give the list constructors which have special syntax and are always in scope their names. They're can also be used to conjure up names from thin air in TH.

MangoIV: the `"" ->` case just *depends*; e.g. it depends on what kinda name you're actually getting, might be unqualified, might be exact.

Aster89: > the `"" ->` case just _depends_; e.g. it depends on what kinda name you're actually getting, might be unqualified, might be exact.

Are you suggesting I should use `mkRdrUnqual` for `Unqual`s, `getRdrName` for `Exact`s, and `mkOrig` for `Orig`s?

But how does using `mkRdrUnqual` vs `getRdrName` vs `mkOrig` make any difference, considering that I'm just using the outputs for ghc-exactprinting? I mean, aren't all of them writing a name without any `Leading.Qualifier.`?

---

As regards tests, I thought that

- this is proving the plugin treats `Exact`s correctly by ensuring that the built-ins `:` and `[]` are left qualified:

  https://github.com/haskell/haskell-language-server/blob/38a34d3593b4651d302e4462e99ec22bd495c4f6/plugins/hls-case-split-plugin/test/testdata/TColon.expected.hs#L5-L8

- this is proving the plugin treats `Orig`s correctly by ensuring that the constructors defined in the very module where `case` is are not qualified:

  https://github.com/haskell/haskell-language-server/blob/38a34d3593b4651d302e4462e99ec22bd495c4f6/plugins/hls-case-split-plugin/test/testdata/TNoPatternsNoBraces.expected.hs#L5-L19

- the test I'm adding in this change is proving that the plugin treats `Qual`s and `Unqual` correctly by ensuring that `Nothing` is qualified and `Just` is not:

  ```haskell
  test :: Maybe a -> ()
  test m = case m of
    Data.Maybe.Nothing -> _
    Just _ -> _
  ```

